PR #1080, opened 2017-12-07
Set idle stack size value to idle tcb

junmin-kim:
This prevents users from considering the idle task when using stack size


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1008, opened 2017-11-23
testcase/audio/utc: Adding tc's for new APIs

shivgarg:
TC for new apis in tinyalsa : pcm_drain and pcm_drop

PR #1081, opened 2017-12-07
tc/audio, framework/tinyalsa: Adding tc for new apis in tinyalsa.

shivgarg:
This PR has two commits, one containing extended tc for new apis in tinyalsa. The other commit fixes some bugs in tinyalsa which are necessary for this tc to run.

### DIFF
--- a/apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
+++ b/apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
@@ -597,6 +597,31 @@ static void utc_audio_pcm_format_to_bits_n(void)
 }
 
 /**
+* @testcase         audio_pcm_drop_p
+* @brief            drop all the buffers which are being processed and stop the device
+* @scenario         call pcm_drop after readi and check that audio after pcm_drop print statement is recorded.
+* @apicovered       pcm_drop
+* @precondition     NA
+* @postcondition    utc_audio_pcm_readi_p should be executed immediately after this function
+*/
+
+static void utc_audio_pcm_drop_p(void)
+{
+	int ret;
+	char *buffer = malloc(pcm_frames_to_bytes(g_pcm, pcm_get_buffer_size(g_pcm)));
+	TC_ASSERT_NEQ_CLEANUP("pcm_readi", buffer, NULL, clean_all_data(0, NULL));
+	printf("Staring pcm device to capture\n");
+	ret = pcm_readi(g_pcm, buffer, pcm_get_buffer_size(g_pcm));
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -596,6 +596,31 @@ static void utc_audio_pcm_format_to_bits_n(void)
...
+*/
+
+static void utc_audio_pcm_drop_p(void)
+{
+	int ret;
+	char *buffer = malloc(pcm_frames_to_bytes(g_pcm, pcm_get_buffer_size(g_pcm)));
+	TC_ASSERT_NEQ_CLEANUP("pcm_readi", buffer, NULL, clean_all_data(0, NULL));
+	printf("Staring pcm device to capture\n");
+	ret = pcm_readi(g_pcm, buffer, pcm_get_buffer_size(g_pcm));
+	TC_ASSERT_GEQ("pcm_drop", ret, 0);
```
sunghan-chang: not cleanup for every TC_ASSERT?

shivgarg: No, drop_p and readi_p tc's are closely linked to each other, so the cleanup function cannot be called here as pcm object is need by readi_p tc to have a successful result.

shivgarg:
#1008 Also works with the given tinyalsa change, we can choose both tc's which one to use

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
sunghan-chang: Then why do you call cleanup at above TC_ASSERT macro? and even if pcm is used at next tc, you should free allocated memory, buffer. Isn't it?

shivgarg: I missed freeing the buffer in error case, I will add that.I did cleanup in the first case because the error is caused by malloc(not a part we want to test) which may occur with high probability in the coming tc's.So I felt it is better to safely close the audio device and let the coming tcs fail too.

seinfra:
Code Rule Check OK.

shivgarg:
I have updated #1008 . You can choose from either of the two.

shivgarg:
#1008 was merged . So this PR is not needed anymore. Both #1008 and this PR where extending the TCs for the new apis in slightly different ways.

PR #1085, opened 2017-12-07
apps/testcase/i2c : i2c ITC rename

arvinmittal:
Rename of scenario ITCs of SystemIO i2c
TC Naming rule is <TC_Category>_<Module>_<SubModule>_<FunctionName>_<Positive/Negative>_<AdditionalInformation>
Current TC name donot have above naming convention

Signed-off-by: Arvin Mittal <arvin.mittal@samsung.com>


seinfra:
Code Rule Check OK.

PR #1088, opened 2017-12-08
framework/tinyalsa: Fixing memory leak in tinyalsa

shivgarg:
Freeing pcm struct in pcm_close when pcm structure passed is a bad pcm.
This is necessary because even in case of bad pcm, the pcm structure is allocated
by pcm_open.

### DIFF
--- a/framework/src/tinyalsa/tinyalsa.c
+++ b/framework/src/tinyalsa/tinyalsa.c
@@ -166,10 +166,6 @@ struct pcm {
 	unsigned int next_offset;
 };
 
-static struct pcm bad_pcm = {
-	.fd = -1,
-};
-
 static int oops(struct pcm *pcm, int e, const char *fmt, ...)
 {
 	va_list ap;
@@ -735,7 +731,8 @@ int pcm_close(struct pcm *pcm)
 		return -EINVAL;
 	}
 
-	if (pcm == &bad_pcm) {
+	if (pcm->fd < 0) {
+		free(pcm);
 		return 0;
 	}
 
@@ -766,9 +763,7 @@ int pcm_close(struct pcm *pcm)
 
 	mq_unlink(pcm->mqname);
 
-	if (pcm->fd >= 0) {
-		close(pcm->fd);
-	}
+	close(pcm->fd);
 
 	free(pcm);
 	return 0;


seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -733,6 +733,7 @@ int pcm_close(struct pcm *pcm)
 	}
 
 	if (pcm == &bad_pcm) {
+		free(pcm);
```
Taejun-Kwon: I think we don't have to keep bad_pcm, original tinyalsa asign pcm as bad_pcm if mem allocation failed.
it returns bad_pcm if malloc failed in pcm_open. but we return null instead of bad pcm.
Could you please remove bad_pcm structure & remove this condition too?

shivgarg: Could you also consider the case when open call fails in pcm_open, then allocated pcm struct is returned which has fd=-1 which makes it equivalent ot bad_pcm? Should I return NULL in that case too?

jsdosa: In my opinion, free at pcm_open and return null, or this seems to work.

Taejun-Kwon: If it allocated but fd is invalid, in this case it will be free at pcm_close.
In player level, it should check pcm_is_ready, and if its return value is false, it should call pcm_close immediately.

shivgarg: So, is this change fine?

Taejun-Kwon: I mean remove bad_pcm structure & related code will be fine.

shivgarg: If I remove bad_pcm, then the UNREGISTERMQ, FREEBUFFER and RELEASE ioctl operations will be called which may lead to errors because neither buffer or message queue were allocated. 
Or do you want me to not use bad_pcm and directly check the value of fd?

Taejun-Kwon: Yes. then 'if (pcm->fd >= 0)' can be removed.
Actually bad_pcm is static structurem and we do not set address of that structure to pcm anywhere.

shivgarg: Ok, I understood

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -735,7 +731,8 @@ int pcm_close(struct pcm *pcm)
 		return -EINVAL;
 	}
 
-	if (pcm == &bad_pcm) {
+	if (pcm->fd < 0) {
```
Taejun-Kwon: I already mentioned that checking condition in 766 line can be removed.

seinfra:
Code Rule Check OK.

PR #1092, opened 2017-12-08
Codecoverage: add corner cases for semaphore TC's 

thapav:
This patch adds left over cases for API's:

kernel/semaphore:
sem_setprotocol()
sem_tickwait()

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_semaphore.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_semaphore.c
@@ -19,15 +19,18 @@
 /// @file semaphore.c
 /// @brief Test Case Example for semaphore API
 #include <tinyara/config.h>
+#include <tinyara/clock.h>
+#include <tinyara/semaphore.h>
 #include <stdio.h>
 #include <errno.h>
 #include <sys/time.h>
 #include <sys/types.h>
 #include "tc_internal.h"
 
-#define PSHARED     0
-#define SEC_2       2
-#define LOOP_CNT    5
+#define PSHARED			0
+#define SEC_2			2
+#define LOOP_CNT		5
+#define SEM_PRIO_DEFAULT	3
 
 static sem_t g_empty;
 static sem_t g_full;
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@thapav Could you check that there are two commits which looks like similar?

apps/examples/testcase/le_tc/kernel/tc_semaphore.c
```diff
@@ -352,6 +356,92 @@ static void tc_semaphore_sem_destroy(void)
...
+	TC_ASSERT_EQ("sem_init", ret_chk, OK);
+
+	ret_chk = sem_setprotocol(&sem, SEM_PRIO_PROTECT);
+	TC_ASSERT_EQ("sem_setprotocol", ret_chk, ERROR);
+	TC_ASSERT_EQ("sem_setprotocol", errno, ENOSYS);
+
+	ret_chk = sem_setprotocol(&sem, SEM_PRIO_DEFAULT);
+	TC_ASSERT_EQ("sem_setprotocol", ret_chk, ERROR);
+	TC_ASSERT_EQ("sem_setprotocol", errno, EINVAL);
+
```
jeongchanKim: why didn't you checking SEM_PRIO_NONE and SEM_PRIO_INHERIT cases?

apps/examples/testcase/le_tc/kernel/tc_semaphore.c
```diff
@@ -352,6 +356,92 @@ static void tc_semaphore_sem_destroy(void)
...
+	ret_chk = sem_tickwait(&sem, clock_systimer() - SEC_2, 0);
+	TC_ASSERT_EQ("sem_tickwait", ret_chk, ERROR);
+
+	ret_chk = sem_tickwait(&sem, clock_systimer() - SEC_2, 1);
+	TC_ASSERT_EQ("sem_tickwait", ret_chk, ERROR);
+
+	ret_chk = sem_tickwait(&sem, clock_systimer() - SEC_2, 3);
+	TC_ASSERT_EQ("sem_tickwait", ret_chk, ERROR);
+
```
jeongchanKim: how about checking more cases?
for example, sem_tickwait will get semaphore after n-ticks.
(you checked negative cases only)

thapav: yes i will do that

seinfra:
Code Rule Check OK.

thapav:
@jeongchanKim 
modified as per review comments

seinfra:
Code Rule Check OK.

apps/examples/testcase/le_tc/kernel/tc_semaphore.c
```diff
@@ -19,20 +19,24 @@
...
+#define PSHARED			0
+#define SEC_2			2
+#define LOOP_CNT		5
+#define SEM_PRIO_DEFAULT	3
 
 static sem_t g_empty;
 static sem_t g_full;
 static sem_t g_sem;
 
+int sem_tickwait(FAR sem_t *sem, systime_t start, uint32_t delay);
```
jeongchanKim: why did you declare here?
this prototype is in tinyara/semaphore.h

sunghan-chang: Yesterday, that was added. Could you use latest version and include that header instead of this as Jeongchan said?

thapav:
updated the codebase, please review

seinfra:
Code Rule Check OK.

thapav:
In that case, if we make it compiler independent, we need to add for all TC files
For now i have added the include

seinfra:
Code Rule Check OK.

PR #1105, opened 2017-12-12
Add LWIP error dependency to network testcase

dr-venkman:
currently, network testcase does not have CONFIG_NET_LWIP_ERROR  automatically enabled, which causes problems in it's execution. The solution is to enforce a dependency of CONFIG_NET_LWIP_ERROR on the testcase example. This ensures that testcase can compile only if CONFIG_NET_LWIP_ERROR is enabled.

### DIFF
--- a/apps/examples/testcase/le_tc/network/Kconfig
+++ b/apps/examples/testcase/le_tc/network/Kconfig
@@ -6,6 +6,7 @@
 menuconfig EXAMPLES_TESTCASE_NETWORK
 	bool "Network TestCase Example"
 	default n
+	select NET_LWIP_ERROR if NET_LWIP
 	---help---
 		Enable the Network TestCase Example
 


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

apps/examples/testcase/le_tc/network/Kconfig
```diff
@@ -5,6 +5,7 @@
 
 menuconfig EXAMPLES_TESTCASE_NETWORK
 	bool "Network TestCase Example"
+	depends on NET_LWIP_ERROR
```
sunghan-chang: Is "depends on " right? not "select"?

dr-venkman: This is a dependency case, so it should be 'depends on' as far as I checked.

seinfra:
Code Rule Check OK.

PR #1108, opened 2017-12-12
le_tc/math: Cover corner TC' for few math API's

thapav:
This patch covers corner TC's for:
j0f()
j1f()
nextafter()
nextafterf()


seinfra:
Code Rule Check OK.

PR #1112, opened 2017-12-13
Remove unused configuration and references in audio.

btheosam:
Audio module is NOT yet ready for power management hence pm disabled
and alc5658 dont need to take references on audio pipleline buffer as
it just passes to lower layer.


seinfra:
Code Rule Check OK.

PR #1115, opened 2017-12-14
ttrace: make a dependancy between DEBUG_TTRACE and TTRACE

sunghan-chang:
DEBUG_TTRACE is valid when TTRACE is enabled.


seinfra:
Code Rule Check OK.

PR #1117, opened 2017-12-14
config/qemu: clean up

sunghan-chang:
1. Remove unnecessary defconfigs, tash_16m and tash_64k
2. The tc_64k runs on flash 256K, but many configs are enabled so that
it causes overflow. Let's disable some unnecessary configs.


seinfra:
Code Rule Check OK.

build/configs/qemu/tc_64k/defconfig
```diff
@@ -299,6 +261,7 @@ CONFIG_RAM_SIZE=65536
 # Board Selection
 #
 CONFIG_ARCH_BOARD_LM3S6965EK=y
+# CONFIG_ARCH_BOARD_ARTIK05X_FAMILY is not set
```
jeongchanKim: This defconfig is for qemu. Is it okay that ARTIK05X is showing?

PR #1119, opened 2017-12-14
ALC5658 codec to support shutdown properly

btheosam:
Shutdown logic for alc5658 has to clean up any pending buffers
in the pend queue.


seinfra:
Code Rule Check OK.

PR #1120, opened 2017-12-14
framework/tinyalsa: modifying pcm_stop operation

shivgarg:
pcm_stop operation is being changed to be a non-blocking operation as compared to current implementation. Now only the buffers from message queue will be removed . Rest of the buffers should be dequeued by lower driver layers.


shivgarg:
I will reopen this once i2s and alc layers are fixed. It was made to share the case when the audio tc crashes.

seinfra:
Code Rule Check OK.

PR #1123, opened 2017-12-15
Audio: Fix buffering delay issue

jarvisjang:
In tinyalsa, variable 'nbytes' be not frame unit but byte unit.
'nbytes' must be as byte unit.

### DIFF
--- a/framework/src/tinyalsa/tinyalsa.c
+++ b/framework/src/tinyalsa/tinyalsa.c
@@ -487,7 +487,7 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
 	pending = pcm_frames_to_bytes(pcm, frame_count);
 
 	while (pending > 0) {
-		nbytes = pending > pcm->buffer_size ? pcm->buffer_size : pending;
+		nbytes = pending > pcm_frames_to_bytes(pcm, pcm->buffer_size) ? pcm_frames_to_bytes(pcm, pcm->buffer_size) : pending;
 
 		if (pcm->buf_ptr < pcm->buffer_cnt) {
 			/* If we have empty buffers, fill them first */


seinfra:
Code Rule Check OK.

sunghan-chang:
@jarvisjang Please write details on your commit message.
```
framework/media: Fix buffering delay issue

In TizenRT tinyalsa, buffer size unit is different with original tiny alsa.
In original tinyalsa, buffer size unit is 'frame'. But in TizenRT, buffer size unit is 'byte'.
```

jsdosa:
This part is used to get the pcm buffer size and set the initial buffer size.
how about this comment~?

seinfra:
Code Rule Check OK.

framework/src/media/media_player.c
```diff
@@ -74,7 +74,7 @@ play_result_t media_play(int fd, media_format_t format)
 		return PLAY_ERROR_DRIVER;
 	}
 
-	g_pc.buffer_size = pcm_frames_to_bytes(g_pc.pcmout, pcm_get_buffer_size(g_pc.pcmout));
+	g_pc.buffer_size = pcm_get_buffer_size(g_pc.pcmout);
```
Taejun-Kwon: pcm_get_buffer_size returns # of frames.
Currently default setting is 16bit & stereo.
(Actually what you added in above default config value is set by tinyalsa, not neccessary here.. of course tinyalsa using 1024 frames as a default period size) 
Hence correct buffer size is that 2048 frames * 2 * 2 = 8192 bytes.
Now Audio subsystem using 8192 bytes as a 1 buffer, and there are 8buffers.
There is while loop in read_data_pcm_write which checking readed value, I guess delay comes due to recv size is too small..

sunghan-chang:
@jarvisjang You should write details at third line, not first. And you will add examples/media then, we can distinguish what it is for ```meda:```.

Taejun-Kwon:
Good job.

seinfra:
Code Rule Check OK.

PR #1125, opened 2017-12-18
frameworks/media: Fix buffer size issue on recording

jarvisjang:
Fix the issue which halt on recording.
Reason: Buffer size unit was different. One is byte, one is frame.

### DIFF
--- a/framework/src/media/media_recorder.c
+++ b/framework/src/media/media_recorder.c
@@ -268,7 +268,8 @@ int record_and_write(char *buffer)
 	int size;
 	char inst = ST_PROTO_RECORD_DATA;
 
-	frames = pcm_readi(g_rc.pcmin, buffer, g_rc.buffer_size);
+	frames = pcm_readi(g_rc.pcmin, buffer, pcm_bytes_to_frames(g_rc.pcmin, g_rc.buffer_size));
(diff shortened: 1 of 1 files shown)


framework/src/media/media_recorder.c
```diff
@@ -268,7 +268,8 @@ int record_and_write(char *buffer)
 	int size;
 	char inst = ST_PROTO_RECORD_DATA;
 
-	frames = pcm_readi(g_rc.pcmin, buffer, g_rc.buffer_size);
+	frames = pcm_readi(g_rc.pcmin, buffer, pcm_bytes_to_frames(g_rc.pcmin, g_rc.buffer_size));
+	printf("(*)frames: %d\n", frames);
```
sunghan-chang: It looks not a mandatory message. How about use debug message?

seinfra:
Code Rule Check OK.

framework/src/media/media_recorder.c
jarvisjang: Currently, this is mandatory.
Without printf, recv()/send() functions don't work out well. I guess this is related to the scheduling multi-threads mechanism. After this issue is fixed, I will delete this line.

Taejun-Kwon:
Do not forget remove printf when issue solved.

PR #1132, opened 2017-12-19
fs/poll: support poll for regular files and block devices

jeongarmy:
cherry-picked 027a446 from nuttx
poll: fix poll for regular files and block devices.  Open Group documentation tells that poll (and select) support regular files and that 'Regular files shall always poll TRUE for reading and writing'.


seinfra:
Code Rule Check OK.

PR #1135, opened 2017-12-19
testcase/filesystem : change the scenario of poll tc about polling regular file instead of stdin, stdout

jeongarmy:
If current tc is run with TASH simultanously, it can be affected by holding uart device.


seinfra:
Code Rule Check OK.

PR #1136, opened 2017-12-19
examples/media: Fix recording stop issue

jarvisjang:
When stop recording, server program would be caught in a infinite loop.
Reason: Sent wrong instruction number.


seinfra:
Code Rule Check OK.

PR #1140, opened 2017-12-20
fix deadlock for media_framework.

jsdosa:
a problem by unlocking the lock after pthread_join,
and I fixed that part.

### DIFF
--- a/framework/src/media/media_player.c
+++ b/framework/src/media/media_player.c
@@ -150,9 +150,12 @@ play_result_t media_stop_play(void)
 	}
 
 	g_pc.state = PLAY_STOPPING;
-	pthread_join(g_pc.pth, NULL);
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

framework/src/media/media_player.c
```diff
@@ -150,9 +150,12 @@ play_result_t media_stop_play(void)
 	}
 
 	g_pc.state = PLAY_STOPPING;
+	PLAY_UNLOCK();
+
 	pthread_join(g_pc.pth, NULL);
+	
+	PLAY_LOCK();
```
shivgarg: Why is it different in player and recorder ? You are locking and unlocking again in media_stop_play  to set pth variable to NULL but not in media_stop_record.

PR #1142, opened 2017-12-21
configs/artik05x: modify download script to download each partition

sunghan-chang:
Only "ALL" option is providing to flash all of partitions on "make download" command.
This patch makes user can select which partition should be flashed.
i.e.
make download OS BL2 ROMFS
make download SSSFW WLANFW


seinfra:
Code Rule Check OK.

PR #1146, opened 2017-12-22
Delete fd members of dumpParser object

junmin-kim:
Refactor dumpParser.py for encapsulation
Use 'with' statement to make sure fd is closed after use


seinfra:
Code Rule Check OK.

PR #1150, opened 2017-12-22
Register the basic TASH commands first

pradeepns85:
Since basic TASH commands are registered at the end, there is a
chance that help command would not get install itself in some
corner cases and thus it causes inconvenience to see all
the registered commands

This patch fixes this issue by always registering basic commands first
and then prints the error message. This will help developer to increase
the CONFIG_TASH_MAX_COMMANDS value

For ex:
1. cd os/tools
2. ./configure.sh artik053/iotivity
3. cd ..
4. make menuconfig
5. go to Application Configuration --> TASH
6. It shows MAX commands allowed as 132
7. Disable the tash option and save the config and exit
8. again run "make menuconfig" and enable the tash
9. It updates the default registered commands as 32 only.
10. Save the config and exit
11. build and download the binary
12. help command is not registered.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -281,8 +281,12 @@ int tash_execute_cmd(char **args, int argc)
 int tash_cmd_install(const char *str, TASH_CMD_CALLBACK cb, int thread_exec)
 {
 	int cmd_idx;
+	static int fail_cmd_count = 0;
 
 	if (TASH_MAX_COMMANDS == tash_cmds_info.count) {
+		printf("Allowed Max tash cmds: %d and Current tash cmd count: %d\n",
+				TASH_MAX_COMMANDS, tash_cmds_info.count + ++fail_cmd_count);
+		printf("Couldn't install cmd: (%s), Refer CONFIG_TASH_MAX_COMMANDS\n", str);
 		return -1;				/* MAX cmd count reached */
 	}
 
--- a/apps/system/init/init.c
+++ b/apps/system/init/init.c
@@ -130,13 +130,13 @@ int preapp_start(int argc, char *argv[])
 #endif
 
 #ifdef CONFIG_TASH
-	tash_register_cmds();
-
 	pid = tash_start();
 	if (pid <= 0) {
 		printf("TASH is failed to start, error code is %d\n", pid);
 		goto error_out;
 	}
+
+	tash_register_cmds();
 #endif
 
 #if defined(CONFIG_LIB_USRWORK) || defined(CONFIG_TASH)


seinfra:
Code Rule Check OK.

apps/shell/tash_command.c
```diff
@@ -281,8 +281,13 @@ int tash_execute_cmd(char **args, int argc)
 int tash_cmd_install(const char *str, TASH_CMD_CALLBACK cb, int thread_exec)
 {
 	int cmd_idx;
+	static int max_tashcmd = 0;
```
sunghan-chang: Do you use max_tashcmd to print a message once? I think you can print it always. Because that is a warning case.
How about show number of exceed to know what value is approprietary?

apps/system/init/init.c
```diff
@@ -63,6 +63,9 @@ extern void iotjs_register_cmds(void);
 #ifdef CONFIG_TASH
 static void tash_register_cmds(void)
 {
+	/* Register the basic commands first to always have "help" cmd */
+	tash_register_basic_cmds();
```
sunghan-chang: Let's call tash_init before calling tash cmds instead of this change.

apps/shell/tash_command.c
pradeepns85: Yes. It was displayed only once and error is shown for all the failed commands.   It can be modified to display the number of exceed as well.

apps/system/init/init.c
pradeepns85: Ok. Can be done.

seinfra:
Code Rule Check OK.

apps/system/init/init.c
sunghan-chang: Oh, I am so sorry, I mean I want moving tash_start, not init. That's my fault.
current
```
int preapp_start()
{
tash_register_cmds();
tash_start();
}
```
suggestion
```
int preapp_start()
{
tash_start();
tash_register_cmds();
}
```

apps/shell/tash_command.c
```diff
@@ -283,6 +283,8 @@ int tash_cmd_install(const char *str, TASH_CMD_CALLBACK cb, int thread_exec)
 	int cmd_idx;
 
 	if (TASH_MAX_COMMANDS == tash_cmds_info.count) {
+		printf("Allowed Max tash cmds: %d and Current tash cmd count: %d\n", TASH_MAX_COMMANDS, tash_cmds_info.count);
```
sunghan-chang: tash_cmds_info.count is not increasing due to return -1. How do you print as you show it?

pradeepns85: Sorry. local changes were not part of push..

seinfra:
Code Rule Check OK.

apps/system/init/init.c
```diff
@@ -130,6 +130,9 @@ int preapp_start(int argc, char *argv[])
 #endif
 
 #ifdef CONFIG_TASH
+	/* Initialize the tash */
+	tash_init();
```
sunghan-chang: Oh, I am so sorry, I mean I want moving tash_start, not init. That's my fault.
current
```
int preapp_start()
{
tash_register_cmds();
tash_start();
}
```
suggestion
```
int preapp_start()
{
tash_start();
tash_register_cmds();
}
```

pradeepns85:
@sunghan-chang , kindly review the latest changes https://github.com/Samsung/TizenRT/pull/1150/commits/e2668ead4cabb2a757db614fbbab4543157dbbe3

PR #1154, opened 2017-12-27
example/nettest: add receive broadcast example

jeongarmy:
add new example for showing how to receive broadcast data

### DIFF
--- a/apps/examples/nettest/nettest.c
+++ b/apps/examples/nettest/nettest.c
@@ -261,6 +261,71 @@ return_with_close:
 	close(sock);
 }
 
+/* ------------------------------------------------------------ */
+/*                                                              */
+/* Receive Broadcast example.                                   */
+/*                                                              */
+/* ------------------------------------------------------------ */
+void broadcast_receive(uint32_t num_packets)
+{
+	int sockfd;
+	int is_broadcast = 1;
+	int count;
+	int nbytes;
+	char databuf[256];
+	struct sockaddr_in recvaddr;
+	struct sockaddr_in sendaddr;
+	socklen_t addrlen = sizeof(struct sockaddr_in);
+
+	printf("[BRSERVER] broadcast receive start!\n");
+
+	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd < 0) {
+		printf("[BRSERVER] socket create fail err(%d)\n", errno);
+		return;
+	}
+
+	if (setsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &is_broadcast, sizeof(is_broadcast)) < 0) {
+		printf("[BRSERVER] set socket option fail err(%d)\n", errno);
+		goto return_with_close;
+	}
+
+	memset(&recvaddr, 0, sizeof(recvaddr));
+	recvaddr.sin_family = AF_INET;
+	recvaddr.sin_addr.s_addr = htonl(INADDR_ANY);
+	recvaddr.sin_port = htons(g_app_target_port);
+
+	if (bind(sockfd, (struct sockaddr *)&recvaddr, sizeof(recvaddr)) < 0) {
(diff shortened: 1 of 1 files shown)


jeongarmy:
@sangwon03 @pillip8282 @KimDongEon @dr-venkman @sunghan-chang Please review.

apps/examples/nettest/nettest.c
```diff
@@ -261,6 +261,71 @@ void broadcast_send(uint32_t num_packets, uint32_t sleep_time)
...
+
+		nbytes = recvfrom(sockfd, databuf, sizeof(databuf), 0, (struct sockaddr *)&sendaddr, (socklen_t *)&addrlen);
+		if (nbytes <= 0) {
+			printf("[BRSERVER] receive data fail err(%d)\n", errno);
+			goto return_with_close;
+		}
+		count++;
+		printf("[BRSERVER] received data from %s #%d : %s\n", inet_ntoa(sendaddr.sin_addr), count, databuf);
+
+		if (num_packets == 0) {
```
jsdosa: i have one question.
call broadcast_receive(0); // It seems to be an infinite loop situation....
is that right?

jsdosa: num_packets does not handle in loops.

jeongarmy: Right, 0 is value from user input to send/receive packets continousely.

apps/examples/nettest/nettest.c
```diff
@@ -261,6 +261,71 @@ void broadcast_send(uint32_t num_packets, uint32_t sleep_time)
...
+			continue;
+		}
+		if (count >= num_packets) {
+			printf("[BRSERVER] exit BRSERVER as received sufficient packets for testing\n");
+			break;
+		}
+	}
+
+return_with_close:
```
Taejun-Kwon: this is error case, 'error_with_fd' is better

apps/examples/nettest/nettest.c
```diff
@@ -807,6 +872,8 @@ int nettest_main(int argc, char *argv[])
 			tcp_server_thread(num_packets_to_process);
 		} else if (proto == NT_UDP) {
 			udp_server_thread(num_packets_to_process);
+		} else if (proto == NT_BROADCAST) {
```
Taejun-Kwon: Currently branches here is very simple but you had better use switch-case if there are more than 2 else if 
It is good habit. Also you already check mode above, below else if(mode == NETTEST_CLIENT_MODE) can be changed just else.

apps/examples/nettest/nettest.c
```diff
@@ -261,6 +261,71 @@ void broadcast_send(uint32_t num_packets, uint32_t sleep_time)
...
+			printf("[BRSERVER] receive data fail err(%d)\n", errno);
+			goto return_with_close;
+		}
+		count++;
+		printf("[BRSERVER] received data from %s #%d : %s\n", inet_ntoa(sendaddr.sin_addr), count, databuf);
+
+		if (num_packets == 0) {
+			continue;
+		}
+		if (count >= num_packets) {
```
Taejun-Kwon: How about below instead of 2 if states? count is always greater than 1
if (num_packets != 0 && num_packets <= count) {
    break;
}

jeongarmy: I will change all same things in this file via another commit

apps/examples/nettest/nettest.c
jeongarmy: I will change all same things in this file via another commit

apps/examples/nettest/nettest.c
jeongarmy: I will change all same things in this file via another commit

apps/examples/nettest/nettest.c
sunghan-chang: How can we stop this infinite loop?

PR #1157, opened 2017-12-28
Audio Tinyalsa: Modify and enable mmap APIs

kishore-sn:
Modified/fixed the mmap related apis in Tinyalsa and exposed them in tinyalsa.h file.


seinfra:
Code Rule Check Result:
framework/src/tinyalsa/tinyalsa.c:1077: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
framework/src/tinyalsa/tinyalsa.c:1302: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
framework/src/tinyalsa/tinyalsa.c:1319: ERROR: [SPC_M_KWD] space required before the open parenthesis '('

seinfra:
Code Rule Check OK.

framework/include/tinyalsa/tinyalsa.h
```diff
@@ -331,7 +331,7 @@ unsigned int pcm_bytes_to_frames(const struct pcm *pcm, unsigned int bytes);
  * @param[out] pcm         A PCM handle.
  * @param[in]  data        The audio sample array
  * @param[in]  frame_count The number of frames occupied by the sample array.
- * @return On success,  written number of frames returned. On failure, a negative number returned.
+ * @return On success, written number of frames returned. On failure, a negative number returned.
```
jsdosa: you should change permission all files.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1120,16 +1136,12 @@ int pcm_stop(struct pcm *pcm)
 	unsigned int size;
 	int prio;
 	struct timespec st_time;
-	while (pcm->buf_ptr > 0) {
+	while (1) {
```
Taejun-Kwon: why you rollback code here? Is there any reason?

framework/include/tinyalsa/tinyalsa.h
kishore-sn: Sorry, i dont understand your comment. Can you please explain?

jsdosa: currently, all files changed file permission 100644->100755

framework/src/tinyalsa/tinyalsa.c
kishore-sn: I have rolled back a few changes related to buf_ptr. 
When i introduced buf_ptr variable, it was meant to keep track of the buffers to be enqueued only at the start of playback or record. For example, if we have 4 buffers, buf_ptr will be incremented in writei or readi when these 4 buffers are enqueued the first time only. It is not supposed to be used to track the enque-deque of buffers later on. Using it in this way is causing a problem at another point. So i reverted it. I think this implementation works well enough.

kishore-sn: But, I think i will change this implementation slightly.

framework/include/tinyalsa/tinyalsa.h
kishore-sn: Fixed

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1210,71 +1223,137 @@ int pcm_drain(struct pcm *pcm)
...
-}
-
+/** Application request to access a portion of direct (mmap) area
+ * @param[in] pcm A PCM handle
+ * @param[out] areas Returned mmap channel areas
+ * @param[out] offset Returned mmap area offset in frames
+ * @param[out] frames Mmap area portion size in frames
+ * @returns On success, zero is returned. On failure, a negative number returned
+ * @ingroup libtinyalsa-pcm
+ */
```
Taejun-Kwon: already defined in header file, you can remove it.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: Got it.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1292,22 +1365,41 @@ int pcm_wait(struct pcm *pcm, int timeout)
...
+		if (msg.msgId == AUDIO_MSG_DEQUEUE) {
+			apb = (struct ap_buffer_s *)msg.u.pPtr;
+			apb->flags &= ~AUDIO_APB_OUTPUT_ENQUEUED;
+			apb->curbyte = 0;
+			count++;
+		}
+	}
+
+	/* If there were no buffers in the queue, wait for codec to put a buffer on the queue */
+	if (count == 0) {
```
Taejun-Kwon: How about below code?
/* Buffer is in the queue, hence we don't have to wait */
if (count > 0) {
    return 1;
}

if (timeout > 0){
............
} else {
............
}

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1320,54 +1412,48 @@ int pcm_wait(struct pcm *pcm, int timeout)
...
 		}
 		if (msg.msgId == AUDIO_MSG_DEQUEUE) {
+			/* If you find a dequeue message in the queue,
+			update the apb flags and curbyte and increment the count*/
 			apb = (struct ap_buffer_s *)msg.u.pPtr;
-			pcm->next_buf = apb;
-			pcm->next_size = apb->nbytes;
-			pcm->buf_ptr--;
+			apb->flags &= ~AUDIO_APB_OUTPUT_ENQUEUED;
+			apb->curbyte = 0;
```
Taejun-Kwon: if (msg.msgId == AUDIO_MSG_DEQUEUE) {
  .....
  return 1;
}
return oops;

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1210,71 +1217,137 @@ int pcm_drain(struct pcm *pcm)
...
+	/* Update buf_ptr. It will be used during pcm_drain */
+	if (pcm->buf_ptr < pcm->buffer_cnt) {
 		pcm->buf_ptr++;
-		pcm->next_buf = NULL;
 	}
 
 	return frames;
 }
 
+/** Application request to access a portion of direct (mmap) area
```
Taejun-Kwon: I think pcm_avail_update's role is that return number of frames which is ready to use

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1210,71 +1217,137 @@ int pcm_drain(struct pcm *pcm)
...
+	}
+
+	if (*frames > nframes) {
+		*frames = nframes;
 	}
 
 	return 0;
 }
 
+/** Application request to access a portion of direct (mmap) area
```
Taejun-Kwon: commit doing finish access & update position, right?

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Ok

framework/src/tinyalsa/tinyalsa.c
kishore-sn: True. 
But, I removed this description as we have defined it in header file. (as per your previous comment)

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Sure. I changed it.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Ok. I changed it.

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -561,7 +562,7 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 		return -EINVAL;
 	}
 
-	if (!(pcm->flags & PCM_IN)) {
+	if (!(pcm->flags & PCM_IN) || pcm->flags & PCM_MMAP) {
```
jsdosa: i think, pcm->flags & PCM_OUT is better.

jsdosa: i mean,
if ((pcm->flags & PCM_OUT) || pcm->flags & PCM_MMAP)

kishore-sn: Since it is pcm_read API, I think its better to check for PCM_IN. 
The if-statement is a bit confusing, but it works well.. and maybe more logically correct.

Taejun-Kwon: I think there's no change call pcm_readi with PCM_MMAP flags, you can remove second condition

kishore-sn: Ok. Done.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -439,7 +441,7 @@ static int pcm_mmap_transfer_areas(struct pcm *pcm, char *buf, unsigned int offs
 	while (size > 0) {
 		frames = size;
 		pcm_mmap_begin(pcm, &pcm_areas, &pcm_offset, &frames);
-		pcm_areas_copy(pcm, pcm_offset, buf, offset, frames);
+		pcm_areas_copy(pcm, pcm_offset, pcm_areas, buf, offset, frames);
```
shivgarg: I have a doubt here, shouldn't min(frames,size) be passed as argument to pcm_areas_copy , because if size is less than frames, then we can end up overwriting memory which is not allocated to this buffer.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1210,71 +1217,137 @@ int pcm_drain(struct pcm *pcm)
...
+		bufdesc.numbytes = pcm_frames_to_bytes(pcm, frames);
+	} else {
+		/* In record case, we will enqueue the buffer only after the application
+		reads all data present in buffer */
+		if (pcm_frames_to_bytes(pcm, offset + frames) < apb->nbytes) {
+			/* Application has not completely read the buffer. So, we only
+			update the offset in the buffer. The application will get the same buffer
+			when it calls mmap_begin() function again */
+			apb->curbyte = pcm_frames_to_bytes(pcm, offset + frames);
+			return frames;
```
shivgarg: This return value is not in accordance with the return values defined in doxygen comments.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1292,82 +1365,95 @@ int pcm_wait(struct pcm *pcm, int timeout)
...
-			int time = -1;
-
-			/* We will not support NOIRQ flag presently */
-#if 0
-			if (pcm->flags & PCM_NOIRQ)
-				time = (pcm->buffer_size - avail - pcm->mmap_control->avail_min)
-					   / pcm->noirq_frames_per_msec;
-#endif
-			err = pcm_wait(pcm, time);
+		if ((unsigned int)avail == 0) {
```
shivgarg: In the proposed change, 0 is also returned by pcm_avail_update when pcm object is null(line 1333), so I think we should change that to a negative value.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1385,19 +1471,33 @@ int pcm_mmap_transfer(struct pcm *pcm, const void *buffer, unsigned int bytes)
...
 	}
 
 	return 0;
 }
 
+/** Write data to the codec using mmap audio buffer
+ * @param[in] pcm A PCM handle
+ * @param[in] data Pointer to the buffer containing data to be written
+ * @param[in] count Size of the data in bytes
+ * @returns On success, written number of frames returned. On failure, a negative number returned
```
shivgarg: pcm_mmap_transfer either returns 0 or negative values, so the return values may not match to those in documentation.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1410,6 +1510,13 @@ int pcm_mmap_write(struct pcm *pcm, const void *data, unsigned int count)
 	return pcm_mmap_transfer(pcm, (void *)data, count);
 }
 
+/** Read data from codec using mmap audio buffers
+ * @param[in] pcm A PCM handle
+ * @param[in] data Pointer to the buffer to read into
+ * @param[in] count Size of the buffer in bytes
+ * @returns On success, read number of frames returned. On failure, a negative number returned
```
shivgarg: same as above

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1292,82 +1365,95 @@ int pcm_wait(struct pcm *pcm, int timeout)
...
 	unsigned int offset = 0, count;
 
+	/* No data to be transferred */
 	if (bytes == 0) {
 		return 0;
 	}
 
-	count = pcm_bytes_to_frames(pcm, bytes);
+	/* Start the PCM, if required */
+	if (pcm->flags & PCM_IN && !pcm->running) {
```
shivgarg: Should PCM_MMAP be checked here as we check for PCM_IN in pcm_readi

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed

framework/src/tinyalsa/tinyalsa.c
kishore-sn: We are checking in pcm_mmap_read and pcm_mmap_write, and this API is not exposed to user. 
So, no need to check here.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1292,82 +1365,95 @@ int pcm_wait(struct pcm *pcm, int timeout)
...
+	if (timeout > 0) {
+		/* Use the timeout given by application */
+		clock_gettime(CLOCK_REALTIME, &st_time);
+		st_time.tv_nsec += timeout * MILLI_TO_NANO;
+		size = mq_timedreceive(pcm->mq, (FAR char *)&msg, sizeof(msg), &prio, &st_time);
+	} else {
+		/* Application did not give a timeout value
+		The behavior here depends on whether mq was opened with O_NONBLOCK
+		This might block until a msg becomes available on the queue */
+		size = mq_receive(pcm->mq, (FAR char *)&msg, sizeof(msg), &prio);
```
shivgarg: In my opinion, we should check for pcm_avail_update, because if pcm_wait is called when no buffer is enqueued. (pcm_open(PCM_OUT | PCM_MMAP) -> pcm_wait(pcm,-1)) , the framework would wait indefinitely .

kishore-sn: True. Fixed.

seinfra:
Code Rule Check Result:
framework/src/tinyalsa/tinyalsa.c:1412: ERROR: [SPC_M_KWD] space required before the open parenthesis '('

kishore-sn:
I made few changes to to handle running and draining states. 
Please review again.

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1210,71 +1222,139 @@ int pcm_drain(struct pcm *pcm)
...
-		bufdesc.session = pcm->session;
+	bufdesc.session = pcm->session;
 #endif
-		bufdesc.numbytes = pcm->next_buf->nmaxbytes;
-		bufdesc.u.pBuffer = pcm->next_buf;
-		if (ioctl(pcm->fd, AUDIOIOC_ENQUEUEBUFFER, (unsigned long)&bufdesc) < 0) {
-			return oops(pcm, errno, "AUDIOIOC_ENQUEUEBUFFER ioctl failed");
-		}
+	bufdesc.u.pBuffer = apb;
+	if (ioctl(pcm->fd, AUDIOIOC_ENQUEUEBUFFER, (unsigned long)&bufdesc) < 0) {
```
shivgarg: Here we should set the curbyte and nbytes parameters to zero , the previous values will get passed to the lower layers and can lead unexpected behaviour.

kishore-sn:
I added a new commit for underrun handling. Kindly review.

seinfra:
Code Rule Check Result:
os/audio/audio.c:814: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/audio/audio.c:814: ERROR: [IDT_M_TAB] please, no spaces at the start of a line

seinfra:
Code Rule Check OK.

os/audio/audio.c
```diff
@@ -693,6 +693,42 @@ static inline void audio_dequeuebuffer(FAR struct audio_upperhalf_s *upper, FAR
...
+ *   Send an AUDIO_MSG_IOERR message to the client to indicate that the
+ *   an error has occured.  The lower-half driver initiates this
+ *   call via its callback pointer to our upper-half driver.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_AUDIO_MULTI_SESSION
+static inline void audio_handleerror(FAR struct audio_upperhalf_s *upper, FAR struct ap_buffer_s *apb, uint16_t status, FAR void *session)
+#else
+static inline void audio_handleerror(FAR struct audio_upperhalf_s *upper, FAR struct ap_buffer_s *apb, uint16_t status)
```
Taejun-Kwon: you should check status.
please use switch-case for it.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -506,6 +504,18 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
...
 				memcpy(apb->samp, (char *)data + offset, nbytes);
 				apb->flags &= ~AUDIO_APB_OUTPUT_ENQUEUED;
+			} else if (msg.msgId == AUDIO_MSG_XRUN) {
+				/* Handle underrun */
+				/* If PCM_NORESTART flag is set, just return an ERROR. Let the application handle it */
+				if (pcm->flags & PCM_NORESTART) {
+					return -EPIPE;
+				}
+
+				/* Try to restart the codec and resend the data */
```
Taejun-Kwon: preparing should be done by pcm_prepare.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -506,6 +504,18 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
 				apb = (struct ap_buffer_s *)msg.u.pPtr;
 				memcpy(apb->samp, (char *)data + offset, nbytes);
 				apb->flags &= ~AUDIO_APB_OUTPUT_ENQUEUED;
+			} else if (msg.msgId == AUDIO_MSG_XRUN) {
```
Taejun-Kwon: please use switch-case instead of if to improve readability.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: We'd better refer to tinyalsa's pcm.c
also AUDIOIOC_PREPARE should be defined in audio.h
We'd better do something in i2s & dma layer.
We will remove specific data from dma pool...

os/audio/audio.c
```diff
@@ -693,6 +693,42 @@ static inline void audio_dequeuebuffer(FAR struct audio_upperhalf_s *upper, FAR
...
+#endif
+{
+	struct audio_msg_s msg;
+
+	audvdbg("Entry\n");
```
Taejun-Kwon: It is useless here.

os/audio/audio.c
```diff
@@ -693,6 +693,42 @@ static inline void audio_dequeuebuffer(FAR struct audio_upperhalf_s *upper, FAR
...
+ *
+ * Description:
+ *   Send an AUDIO_MSG_IOERR message to the client to indicate that the
+ *   an error has occured.  The lower-half driver initiates this
+ *   call via its callback pointer to our upper-half driver.
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_AUDIO_MULTI_SESSION
+static inline void audio_handleerror(FAR struct audio_upperhalf_s *upper, FAR struct ap_buffer_s *apb, uint16_t status, FAR void *session)
```
Taejun-Kwon: How about audio_error_handler ??

os/audio/audio.c
```diff
@@ -693,6 +693,42 @@ static inline void audio_dequeuebuffer(FAR struct audio_upperhalf_s *upper, FAR
...
+	if (upper->usermq != NULL) {
+		/* We are always sending XRUN error msg since it is the
+		only type of error that we handle now. If new error scenario
+		is required, then we need to handle here */
+		msg.msgId = AUDIO_MSG_XRUN;
+		msg.u.pPtr = NULL;
+#ifdef CONFIG_AUDIO_MULTI_SESSION
+		msg.session = session;
+#endif
+		mq_send(upper->usermq, (FAR const char *)&msg, sizeof(msg), MQ_PRIO_MAX);
```
btheosam: since error notification happens at irq level. so this mq_send happens at interrupt context. would be great to know the following?
a. Did you test this funcitonality?
b. Did mq_send work properly from interrupt_context, any sideeffects of this?
b. Any rough esmitate on how much time it takes for the error notification from isr to reach the app queue?

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -673,6 +683,14 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 					}
 				}
 
+			} else if (msg.msgId == AUDIO_MSG_XRUN) {
```
btheosam: Shouldn't tinyalsa do something more, like idenfitying semi filled buffer, and remove if before continuing?

framework/src/tinyalsa/tinyalsa.c
btheosam: Same comment as in readi processing.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -614,7 +625,7 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 				if (ioctl(pcm->fd, AUDIOIOC_ENQUEUEBUFFER, (unsigned long)&bufdesc) < 0) {
 					return oops(pcm, errno, "failed to enque buffer after read");
 				}
-				pcm->buf_ptr++;
+				apb->flags |= AUDIO_APB_OUTPUT_ENQUEUED;
```
Taejun-Kwon: Should we do this in tinyalsa layer, not audio layer?????

kishore-sn: I think its better to do in tinyalsa. 
If we do in audio layer and there is some problem in the msg queue, then it wont be valid.
If we do at this point in tinyalsa, we are sure that enqueue/dequeue has completed successfully.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: According to tinyalsa code, when underrun happens -->
- Restart pcm
- Try to redo the read / write operation

This is what I am doing here. pcm_stop -> start will restart the pcm. This will clear all pending buffers and enqueue new buffers. Also, I am using continue keyword. So, in next iteration of the loop, we will retry to perform read operation.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Fixed

os/audio/audio.c
kishore-sn: Done

os/audio/audio.c
kishore-sn: Done

os/audio/audio.c
kishore-sn: Since we have only underrun and overrun errors, I am not checking the status. 
I am always sending XRUN msg and it is implicit that overrun during record and underrun during playback. 
In future, if we have some new error types, we can check status and pass msg accordingly. But, I am not expecting any new error msgs.

os/audio/audio.c
kishore-sn: a. No, I have not tested. I need audio module to test.
b. Not sure if it will work. But, in my opinion, we must not be doing this. I suggest that we do it on a worker thread. This must be done either in I2S driver or codec driver. audio.c is just a wrapper file and must not create new worker.
c. Not sure. But I am using MQ_PRIO_MAX so that the error msg will take priority over all others in the audio msg queue

framework/src/tinyalsa/tinyalsa.c
kishore-sn: We need to decide how to handle xrun scenario. 
How does the codec chip behave when an xrun happens? 
Do we need to restart it?
Or just do a sw reset of codec? 
Or nothing to be done?
In any case, I think we can use pcm_prepare instead of pcm_stop - pcm_start. And implement the required logic in codec side (since the behavior after xrun depends on each codec).

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Replied above.

kishore-sn: Sorry. But is it really needed to used switch for just 2 - 3 cases?
I feel it would make the code look complex since the switch will nest inside a while or for loop in some places. Lets keep current code for now?

os/audio/audio.c
btheosam: if you think, "But, in my opinion, we must not be doing this." then please this as a NOTE in main review comment. But, it is still worth to get metrics/statistics even if we forbid this. 

Regarding your suggestion; 
Can you elaborate further to know if that is feasibile ?

framework/src/tinyalsa/tinyalsa.c
btheosam: But what about the half filled buffer (in case)? How is it handled in tinyalsa? (just ignored?)

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: No, application(or media f/w) should decide that recorver or not.
So we should provide pcm_prepare instead of pcm_stop & pcm_start.
pcm_prepare should send some message to device layer through the ioctl, and then device will do something when it get message from upper layer.
@btheosam and I have a plan that handle this in codec & i2s layer. we will design it.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: In my opinion, handling buffer(change something in buffer structure) should be done by audio layer.
I think we can change it in device layer before send ret to upper layer.
Please double check this.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: In other words, return -EPIPE is enough for pcm_writei(or call when it started like a original tinyalsa)
also pcm_prepare should call ioctl(pcm->fd, AUIDIOIC_PREPARE);
Application or media f/w will call as a below.
if (pcm->flags & PCM_MMAP) {
     ret = pcm_mmap_write();
} else {
    ret = pcm_write();
}

if (ret != 0) {
   switch(ret) {
      case XRUN : 
          ret = pcm_prepare(pcm);
}
if (pcm->flags & PCM_MMAP) {
     ret = pcm_mmap_write();
} else {
    ret = pcm_write();
}

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -673,6 +683,14 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 					}
 				}
 
+			} else if (msg.msgId == AUDIO_MSG_XRUN) {
+				/* Handle underrun */
+				/* Try to restart the codec and continue reading data */
+				if (!pcm->draining) {
+					pcm_stop(pcm);
```
jsdosa: In my opinion, need to check for error cases.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -673,6 +683,14 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
...
+					pcm_start(pcm);
```
jsdosa: same as above.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Yes. I understand your point about setting the flags in audio layer. 
But, I am using the AUDIO_APB_OUTPUT_ENQUEUED flag to check if the buffer is available to tinyalsa for writing data. Suppose, the audio layer sets AUDIO_APB_OUTPUT_ENQUEUED. But the buffer is still in queue. Then, it will lead to mistakes. So, I want to update the flag only after removing the buffer from the queue.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: The error is XRUN. That is what we check with "else if (msg.msgId == AUDIO_MSG_XRUN)".
There is no sub-error codes withing XRUN.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: There is no sub-error codes withing XRUN.

os/audio/audio.c
kishore-sn: I think you can call the err_callback function from i2s_rx_worker or i2s_tx_worker. Since, we know that application only will respond to the xrun error, there is no point in handling it in interrupt context.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: Ok. 
But in Tinyalsa implementation, pcm_prepare is being called directly inside pcm_writei and pcm_readi when XRUN happens. But mmap API's just return -EPIPE. Do you want to continue this type of implementation OR always return -EPIPE during XRUN?

Taejun-Kwon: They check pcm->running before call pcm_prepare();
It means pcm_stopped due to some reasons, we can follow their policy.
In our case, if enqueue returns XRUN, we can just return -EPIPE. 
If you have more good idea, you can suggest but my opinion is that user should decice about recovery ...

os/audio/audio.c
kishore-sn: In i2s file.. create new function like err_cb_worker().
In this function, just call priv->err_cb().
In i2s_irq_handler, schedule err_cb_worker() instead of calling priv->err_cb().

How about this?

framework/src/tinyalsa/tinyalsa.c
kishore-sn: How about below implementation in tinyalsa?
In codec driver PREPARE ioctl can be handled in hw specific manner.

pcm_writei() {
...
	if(xrun) {
		if (pcm->flags & PCM_NORESTART) {
				return -EPIPE;
		} else {
			pcm_prepare();
			pending = pcm_frames_to_bytes(pcm, frame_count);
			offset = 0;
			continue;
		}
	}
}

pcm_readi() {
	...
	if (xrun) {
		if (!pcm->draining) {
			pcm_prepare();
		}
		continue;
	}
}

pcm_mmap_xxx_api() {
	...
	if (xrun) {
		return -EPIPE;
	}
}

btheosam: I totally agree with @Taejun-Kwon comment "No, application(or media f/w) should decide that recorver or not"

seinfra:
Code Rule Check OK.

kishore-sn:
As per offline discussion about XRUN error handling, I have made following changes:
- Return -EPIPE error from readi and writei 
- Add PREPARE ioctl in audio.h
- Add place holder code in alc driver for handling PREPARE ioctl
- Call prepare ioctl in pcm_prepare

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -476,7 +475,7 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
 		return -EINVAL;
 	}
 
-	if (pcm->flags & PCM_IN) {
+	if (!(pcm->flags & PCM_OUT)) {
```
Taejun-Kwon: There's no reason to change this.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: What I mean is that change state right before ioctl return & mqueue returns dequeue result.

os/audio/audio.c
Taejun-Kwon: please check this one more time, log is still here.

jsdosa:
framework/src/tinyalsa/tinyalsa.c 100644 → 100755

framework/src/tinyalsa/tinyalsa.c
kishore-sn: The description of the API says "This function is only valid for PCMs opened with the @ref PCM_OUT flag". So I think it is better to change like this.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: You mean that we have to change the state in audio.c, before return from ioctl. But, when we return from ioctl, the buffer is still in the queue. I want to change the state after removing the buffer from the audio msg queue. Only then, the buffer is available to write data.

os/audio/audio.c
kishore-sn: Sorry. I missed in my last commit. 
Fixed now.

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
Taejun-Kwon: I don't understand well, in your code, tinyalsa request enqueue through the ioctl, and ioctl request enqueueing buffer to i2s, whenever callback function called from i2s, it changed apb state.
Please check alc5658_enqueuebuffer that even if priv is not running, it returns ok..
In other words, AUDIO_APB_OUTPUT_ENQUEUED should be set after driver request i2s_send or i2s_receiver, on the other hands, AUDIO_APB_OUTPUT_ENQUEUED should be unset driver's call back called with valid buffer data.
Make sure that only driver knows it enqued or dequeued properly...

os/audio/audio.c
Taejun-Kwon: @btheosam What is your opinion?

os/drivers/audio/alc5658.c
```diff
@@ -974,6 +974,11 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
 
 	switch (cmd) {
 
+	case AUDIOIOC_PREPARE: {
+		/* Need to be implemented later */
```
Taejun-Kwon: Please add TODO or FIXME tag

kishore-sn: Done

seinfra:
Code Rule Check OK.

framework/src/tinyalsa/tinyalsa.c
kishore-sn: My previous comment is a bit confusing / misleading!! Sorry.
What I mean is this. I am using the AUDIO_APB_OUTPUT_ENQUEUED flag to check whether the buffer is with tinyalsa (means we can use it to read / write) OR the buffer is not with tinyalsa (its with codec or i2s or mqueue) and so we cannot use it. I dont think there is any need for driver to use this flag to track the buffer state. It will be used only in tinyalsa to check if the buffer is in tinyalsa or not. If its not in tinyalsa, then it does not matter if it is in codec or i2s.
That is why I set the flag on a successful return from ioctl(AUDIOIOC_ENQUEUEBUFFER) and reset it on successfully receiving AUDIO_MSG_DEQUEUE. This flag is not being used or checked anywhere else in the code.

Taejun-Kwon: What I am saying is that ioctl return ok without enqueing, in this case tinyalsa set buffer state as ENQUEUED.
That's why I ask to you double check.
If it used in tinyalsa to check buffer is enqueued or something, we should make an extra policy.
(For example, ioctl in driver returns size of buffer if enqueued. or define new constatnt (for example, AUDIOIOC_GET_BUFFERSTATE)to use extra ioctl to get buffer state...

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check Result:
apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c:934: ERROR: [BRC_M_FTN] open brace '{' following function declarations go on the next line
apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c:965: ERROR: [BRC_M_FTN] open brace '{' following function declarations go on the next line
apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c:995: ERROR: [BRC_M_FTN] open brace '{' following function declarations go on the next line
apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c:1035: ERROR: [BRC_M_FTN] open brace '{' following function declarations go on the next line
apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c:1125: ERROR: [BRC_M_FTN] open brace '{' following function declarations go on the next line

seinfra:
Code Rule Check OK.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -925,6 +922,259 @@ static void utc_audio_pcm_drain_n(void)
...
+/**
+* @testcase         audio_pcm_write_p
+* @brief            write some predetermined amout of data from file to pcm
+* @scenario         open a pcm with MMAP flag and write data using mmap mode api's
+* @apicovered       pcm_avail_update, pcm_wait, pcm_mmap_begin, pcm_mmap_commit
+* @precondition     NA
+* @postcondition    NA
+*/
+
+static void utc_audio_pcm_mmap_write_p(void)
```
Taejun-Kwon: You don't use mmap_write anywhere

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -925,6 +922,259 @@ static void utc_audio_pcm_drain_n(void)
...
+/**
+* @testcase         audio_pcm_read_p
+* @brief            read predetermined bytes of data from pcm and store in a file
+* @scenario         open pcm in mmap mode and read some data and store it in a file
+* @apicovered       pcm_avail_update, pcm_wait, pcm_mmap_begin, pcm_mmap_commit
+* @precondition     NA
+* @postcondition    NA
+*/
+
+static void utc_audio_pcm_mmap_read_p(void)
```
Taejun-Kwon: pcm_mm_read too.

framework/include/tinyalsa/tinyalsa.h
```diff
@@ -398,6 +398,75 @@ unsigned int pcm_format_to_bits(enum pcm_format format);
...
+* @brief Writes audio samples to the PCM using the mmap buffer
+*
+* @details @b #include <tinyalsa/tinyalsa.h>
+* @param[in] pcm A PCM handle
+* @param[in] data Pointer to the buffer containing data to be written
+* @param[in] count Size of the data in bytes
+* @returns On success, written number of frames returned. On failure, a negative number returned
+* @since Tizen RT v1.x
+*/
+int pcm_mmap_write(struct pcm *pcm, const void *data, unsigned int count);
```
Taejun-Kwon: utc should test api what you added here.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -925,6 +922,259 @@ static void utc_audio_pcm_drain_n(void)
...
+
+	bytes_per_frame = pcm_frames_to_bytes(g_pcm, 1);
+	remain = AUDIO_DEFAULT_RATE * AUDIO_RECORD_DURATION;
+
+	printf("Sample will be recorded for 3s, press any key to start(Total frame: %d)\n", remain);
+	fflush(stdout);
+	str = gets(input_str);
+	TC_ASSERT_NEQ_CLEANUP("pcm_mmap_read", str, NULL, clean_all_data(fd, NULL));
+
+	ret = pcm_start(g_pcm);
```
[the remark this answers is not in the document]
kishore-sn: We need to make pcm_start as open api. It is required for mmap scenario. 
pcm_start is open api in the original tinyalsa code. I will change this and commit.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Yes. Actually, I have written this UTC to test the complete mmap write scenario. Just like utc_audio_pcm_writei_p() or utc_audio_pcm_readi_p() which actually performs a proper audio write or read from file. I want to have similar type of API for mmap write / read.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Same as above

framework/include/tinyalsa/tinyalsa.h
kishore-sn: Yes. I understand. But, this mmap_write API is actually not useful. Internally it performs memcpy which we want to avoid. So, either we can remove this API, or we can just keep it and write an additional UTC for it. 

However, i suggest that this api must not be used by application. Application must use the other mmap api's like avail -> wait -> begin -> commit explicitly to take advantage of mmap.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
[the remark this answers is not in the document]
Taejun-Kwon: No, I checked some codes, but I couldn't found any code which call pcm_start directly,
If you found any code, could you please let me know?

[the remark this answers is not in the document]
kishore-sn: Please check the below alsa documentation for mmap api. 
https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m___direct.html#gac306bd13c305825aa39dd9180a3ad520

And the example code is below. Please look for direct_loop() method. 
https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2pcm_8c-example.html#a45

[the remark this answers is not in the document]
kishore-sn: However, if you think that we must not open pcm_start, then i can find a work around for this.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
Taejun-Kwon: Ok I understand your approach but role of utc is that checking open api works fine or not.
so we should call mm_write or mm_read in utc instead of same scenario.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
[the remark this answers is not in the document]
kishore-sn: I modified the implementation to avoid usage of pcm_start() by application.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Added testcases for all APIs.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Added testcases for all APIs.

framework/include/tinyalsa/tinyalsa.h
kishore-sn: Added testcases for all APIs.
Still, Application must use the other mmap api's like avail -> wait -> begin -> commit explicitly to take advantage of mmap.

seinfra:
Code Rule Check OK.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
Taejun-Kwon: All these apis should be opened?

kishore-sn: Yes. We need to open these API's.
avail, wait, begin, commit ---> actual mmap implementation to avoid memcpy
mmap_read, mmap_write ---> to be compatible with original alsa implementation.

os/audio/audio.c
btheosam: Rough estimate gives around 25ms delay for the error propogation from irq to user app (assuming mq_send is NOT interrupted & use app gets its time slice soon). With 8192 periods, this should be okay.

PR #1158, opened 2017-12-28
Fix : tash_start() prints incorrect error code

pradeepns85:
tash_start() api prints pid as error code, tash_start calls task_create()
to start tash, Incase of task_create fails to start a task, it sets the
error code in "errno" and returns either valid pid or ERROR(-1)

Also, there is is no need to print the error message inside tash_start as it
causes duplication inside preapp_start() api of apps/system/init/init.c

This commit does following things
1. Sets the valid error code inside tash_start
2. Removes duplicated printf message inside tash_start as it's already
printed in preapp_start()

### DIFF
--- a/apps/shell/tash_main.c
+++ b/apps/shell/tash_main.c
@@ -220,10 +220,13 @@ static int tash_main(int argc, char *argv[])
 int tash_start(void)
 {
 	int pid;
+	int errcode;
 
 	pid = task_create("tash", TASH_TASK_PRIORITY, TASH_TASK_STACKSIZE, tash_main, (FAR char *const *)NULL);
 	if (pid < 0) {
-		printf("TASH is not started, error code = %d\n", pid);
+		errcode = errno;
+		DEBUGASSERT(errcode > 0);
+		return -errcode;
 	}
 
 	return pid;


seinfra:
Code Rule Check OK.

apps/shell/tash_main.c
```diff
@@ -220,10 +220,13 @@ static int tash_main(int argc, char *argv[])
 int tash_start(void)
 {
 	int pid;
+	int errcode;
 
 	pid = task_create("tash", TASH_TASK_PRIORITY, TASH_TASK_STACKSIZE, tash_main, (FAR char *const *)NULL);
 	if (pid < 0) {
-		printf("TASH is not started, error code = %d\n", pid);
+		errcode = errno;
+		DEBUGASSERT(errcode > 0);
```
sunghan-chang: What's the purpose of this line?

pradeepns85: It validates that whether error code is junk value or not.  Because we return negative value of error code, its expected to be always greater than 0

sunghan-chang: Is it possible to be negative value?

pradeepns85: @sunghan-chang , logically by software its not possible. only in case of memory corruption or hardware errors like BITFLIP could make it

sunghan-chang: Do we need to concern hw error especially here?

pradeepns85: I could see that, It's been handled at other place as well . 
For ex. work_usrstart () in lib/libc/wqueue/work_usrthread.c  , Anyway if you think that it's not necessary, we can remove it

sunghan-chang: Thank you for letting me know the places. I agree and merged.

PR #1159, opened 2017-12-29
framework/src/media: modify media logic. 

YiHyunJin:
modify deadlock and error. modify server_linux.

### DIFF
--- a/apps/examples/media_test/server/server_example_linux.c
+++ b/apps/examples/media_test/server/server_example_linux.c
@@ -27,6 +27,7 @@
 #include <sys/socket.h>
 
 #define SEND_BUFFER_SIZE (1 << 16)
+#define RECV_BUFFER_SIZE (1 << 11)
 
 #define INST_PLAY         0x00
 #define INST_STOP         0x01
@@ -59,7 +60,8 @@ struct server_context_s {
 };
 
 struct server_context_s g_sc;
-char buffer[SEND_BUFFER_SIZE];
+char g_send_buffer[SEND_BUFFER_SIZE];
+char g_receive_buffer[RECV_BUFFER_SIZE];
 
 bool recv_data(void *ptr, size_t size)
 {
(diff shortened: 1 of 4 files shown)


seinfra:
Code Rule Check OK.

framework/src/media/media_player.c
```diff
@@ -323,7 +321,7 @@ int player_worker(void *args)
 	g_pc.pcmout = NULL;
 
 	PLAY_LOCK();
-	g_pc.state = PLAY_STOPPING;
+	g_pc.state = PLAY_ERROR;
```
YiHyunJin: I will fix all status.

jsdosa: when do you it~?

YiHyunJin: We need to discuss about status, and I will fix it on next commit.

framework/src/media/media_recorder.c
```diff
@@ -194,8 +194,10 @@ record_result_t media_stop_record(void)
 
 	g_rc.state = RECORD_STOPPING;
 	RECORD_UNLOCK();
+	
+	pthread_join(g_rc.pth, NULL);
 
-	pthread_join(g_rc.pth, NULL);	
+	printf("media_stop_record done.\n");
```
Taejun-Kwon: Actually it is framework, not application so please get rid of printf.
We should refine some debug command instead of printf.

apps/examples/media_test/server/server_example_linux.c
```diff
@@ -59,7 +60,8 @@ struct server_context_s {
 };
 
 struct server_context_s g_sc;
-char buffer[SEND_BUFFER_SIZE];
+char s_buffer[SEND_BUFFER_SIZE];
```
Taejun-Kwon: Correct coding style is g_send_buffer / g_receive_buffer.
you can use 'buf' instead of buffer.

seinfra:
Code Rule Check OK.

PR #1160, opened 2017-12-29
pthread, time: change sdbg to svdbg

sunghan-chang:
Some sdbg has just information message, not error message so that
debug message level is changed to verbose instead of error.


seinfra:
Code Rule Check OK.

PR #1162, opened 2017-12-29
Fix incorrect error code display in preapp_start()

pradeepns85:
mount operation sets the error number and returns either 0 or ERROR.
To get the actual error no, use get_errno()

### DIFF
--- a/apps/system/init/init.c
+++ b/apps/system/init/init.c
@@ -36,10 +36,9 @@
 #ifdef CONFIG_SYSTEM_INFORMATION
 #include <apps/system/sysinfo.h>
 #endif
-#ifdef CONFIG_BUILD_PROTECTED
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

sunghan-chang:
@pradeepns85 Could you modify one more?
FS_PROCFS has a dependancy with DISABLE_MOUNTPOINT already. We don't need to check DISABLE_MOUNTPOINT definition.
```
config FS_PROCFS
	bool "PROCFS File System"
	default n
	depends on !DISABLE_MOUNTPOINT
```
Let's remove it.
```#if !defined(CONFIG_DISABLE_MOUNTPOINT) && defined(CONFIG_FS_PROCFS)``` ==> ```#ifdef CONFIG_FS_PROCFS```

seinfra:
Code Rule Check OK.

pradeepns85:
@sunghan-chang , I have modified it as per the suggestions.. Kindly review it

PR #1163, opened 2017-12-29
Logging task scheduling info for armv7-m based board

pradeepns85:
Further enhancement of Reference PR#1141

This commits adds support for logging the task scheduling information
for armv7-m architecture and it's verified using qemu as it uses armv7-m


seinfra:
Code Rule Check OK.

PR #1166, opened 2018-01-02
Alc5658 driver changes

btheosam:
Alc5658 driver don't register for underrun i2s error notification

### DIFF
--- a/os/drivers/audio/alc5658.c
+++ b/os/drivers/audio/alc5658.c
@@ -674,8 +674,8 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
 	}
 
 	alc5658_setregs(priv);
-
 	alc5658_getregs(priv);
+
 	return ret;
 }
 
@@ -693,16 +693,15 @@ static int alc5658_shutdown(FAR struct audio_lowerhalf_s *dev)
 	DEBUGASSERT(priv);
 
 	/* First disable interrupts */
-
 	ALC5658_DISABLE(priv->lower);
 
 	alc5658_takesem(&priv->devsem);
-	dq_entry_t *tmp = NULL;
-	for (tmp = (dq_entry_t *)dq_peek(&priv->pendq); tmp; tmp = dq_next(tmp)) {
-		dq_rem(tmp, &priv->pendq);
+	sq_entry_t *tmp = NULL;
+	for (tmp = (sq_entry_t *)sq_peek(&priv->pendq); tmp; tmp = sq_next(tmp)) {
+		sq_rem(tmp, &priv->pendq);
 		audvdbg("(alcshutdown)removing tmp with addr 0x%x\n", tmp);
 	}
-	dq_init(&priv->pendq);
+	sq_init(&priv->pendq);
 	alc5658_givesem(&priv->devsem);
 
 	/* Now issue a software reset.  This puts all ALC5658 registers back in
@@ -711,6 +710,25 @@ static int alc5658_shutdown(FAR struct audio_lowerhalf_s *dev)
 
 	alc5658_hw_reset(priv);
 	return OK;
+}
+
+/****************************************************************************
+ * Name: alc5658_io_err_cb
+ *
+ * Description:
+ *   Callback function for io error
+ *
+ ****************************************************************************/
+static void alc5658_io_err_cb(FAR struct i2s_dev_s *dev, FAR void *arg, int flags)
+{
+	FAR struct alc5658_dev_s *priv = (FAR struct alc5658_dev_s *)arg;
+
+	/* Call upper callback, let it post msg to user q
+	 * apb is set NULL, okay? Rethink
+	*/
+	lldbg("alc5658_io_err_cb flags: 0x%x\n", flags);
+	priv->dev.upper(priv->dev.priv, AUDIO_CALLBACK_IOERR, NULL, flags);
+	
 }
 
 /****************************************************************************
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.c
```diff
@@ -713,6 +713,25 @@ static int alc5658_shutdown(FAR struct audio_lowerhalf_s *dev)
...
+ ****************************************************************************/
+static void alc5658_io_err_cb(FAR struct i2s_dev_s *dev, FAR void *arg, int flags)
+{
+	FAR struct alc5658_dev_s *priv = (FAR struct alc5658_dev_s *)arg;
+
+	/* Call upper callback, let it post msg to user q
+	 * apb is set NULL, okay? Rethink
+	*/
+	auddbg("Received error with flags: 0x%x\n", flags);
+	priv->dev.upper(priv->dev.priv, AUDIO_CALLBACK_IOERR, NULL, flags);
```
kishore-sn: I think the flags must not be passed directly to upper layer. The upper layers cannot make sense of the flags related to I2S. The codec is responsible for converting the flags into some form that the upper layers can understand. 
My suggestion is to add new error codes in audio.h file for overflow and underflow scenario. This function has to check the I2S error and send appropriate error codes to upper layer. 

Also, need to add code in audio.c for AUDIO_CALLBACK_IOERR.

os/drivers/audio/alc5658.c
```diff
@@ -729,41 +748,34 @@ static int alc5658_start(FAR struct audio_lowerhalf_s *dev)
...
-	if (entry == -1) {
-		return -EINVAL;
+	if (priv->running) {
+		goto alcstart_ok;
 	}
 
-	alc5658_exec_i2c_script(priv, g_sample_entry[entry].script, g_sample_entry[entry].size / sizeof(t_codec_init_script_entry));
-	alc5658_exec_i2c_script(priv, codec_init_inout_script2, sizeof(codec_init_inout_script2) / sizeof(t_codec_init_script_entry));
+	/* Register cb for io error */
+	I2S_ERR_CB_REG(priv->i2s, alc5658_io_err_cb, priv);
```
kishore-sn: Other than overflow and underflow, there can be a i2s timeout error also. I think that also needs to be handled.

os/drivers/audio/alc5658.c
```diff
@@ -912,7 +922,7 @@ static int alc5658_enqueuebuffer(FAR struct audio_lowerhalf_s *dev, FAR struct a
 
 	if (priv->inout) {		/* record */
 		ret = I2S_RECEIVE(priv->i2s, apb, alc5658_rxtxcallback, priv, CONFIG_ALC5658_I2S_TIMEOUT);
-	} else {					/* playback */
+	} else {				/* playback */
```
junmin-kim: Is this diff necessary and related with commit message?
It seems better to remove unnecessary diff chunck.

os/drivers/audio/alc5658.c
Taejun-Kwon: I have a same opinion as kishore's

os/drivers/audio/alc5658.c
```diff
@@ -729,41 +748,34 @@ static int alc5658_start(FAR struct audio_lowerhalf_s *dev)
...
-	int entry;
-	if (priv->running) {
-		return OK;
-	}
-
 	audvdbg(" alc5658_start Entry\n");
-	alc5658_exec_i2c_script(priv, codec_init_inout_script1, sizeof(codec_init_inout_script1) / sizeof(t_codec_init_script_entry));
+	alc5658_takesem(&priv->devsem);
 
-	entry = alc5658_get_sample_rate_script(priv->samprate);
```
Taejun-Kwon: If we remove code here, can we change sample rate??

os/drivers/audio/alc5658.c
```diff
@@ -729,41 +748,34 @@ static int alc5658_start(FAR struct audio_lowerhalf_s *dev)
...
-	if (entry == -1) {
-		return -EINVAL;
+	if (priv->running) {
+		goto alcstart_ok;
```
Taejun-Kwon: goto end_with_sem; or start_with_sem; or you can rename xxxx_with_sem.

os/drivers/audio/alc5658.c
```diff
@@ -985,10 +986,8 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
 #endif
 	}
 	break;
-
-
 	default:
-		audvdbg("Ignored\n");
+		auddbg("alc5658_ioctl received unkown cmd 0x%x\n", cmd);
```
Taejun-Kwon: good change for information. but verbose is correct, I think.

os/drivers/audio/alc5658.c
```diff
@@ -1274,28 +1265,25 @@ FAR struct audio_lowerhalf_s *alc5658_initialize(FAR struct i2c_dev_s *i2c, FAR
...
 		 */
-
 		priv->dev.ops = &g_audioops;
 		priv->lower = lower;
 		priv->i2c = i2c;
 		priv->i2s = i2s;
 
 		sem_init(&priv->devsem, 0, 1);
 		dq_init(&priv->pendq);
-		dq_init(&priv->doneq);
```
Taejun-Kwon: if we don't need doneq, why don't you remove doneq from structure in alc5658.h?(Actually I think we don't need it)

os/drivers/audio/alc5658.h
```diff
@@ -109,7 +109,6 @@ struct alc5658_dev_s {
 	FAR struct i2c_dev_s *i2c;	/* I2C driver to use */
 	FAR struct i2s_dev_s *i2s;	/* I2S driver to use */
 	struct dq_queue_s pendq;	/* Queue of pending buffers to be sent */
-	struct dq_queue_s doneq;	/* Queue of sent buffers to be returned */
```
Taejun-Kwon: oops you already removed. ignore previous comment

os/drivers/audio/alc5658.c
```diff
@@ -949,23 +959,14 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
...
 
 	case AUDIOIOC_HWRESET: {
-		/* REVISIT:  Should we completely re-initialize the chip?   We
-		 * can't just issue a software reset; that would puts all ALC5658
-		 * registers back in their default state.
-		 */
-
-		audvdbg("AUDIOIOC_HWRESET:\n");
+		/* This should put ALC5658 in default state, reconfiguration needed */
+		audvdbg("AUDIOIOC_HWRESET: No Action Would be taken now \n");
```
Taejun-Kwon: We have a alc5658_hw_reset(), but we don't have to use here?

os/drivers/audio/alc5658.c
```diff
@@ -674,8 +674,8 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
 	}
 
 	alc5658_setregs(priv);
-
 	alc5658_getregs(priv);
```
jsdosa: why do you setresgs and then getregs ?
Is it the code that checks to see if it is set?

os/drivers/audio/alc5658.c
jsdosa: I think it is better to keep it.
Currently unused, it can be used according to the driver.

os/drivers/audio/alc5658.c
Taejun-Kwon: After merge this & kishore's code #1157, I will update code regarding this.

os/drivers/audio/alc5658.c
btheosam: return value of enqueue request would address timeout error.

os/drivers/audio/alc5658.c
btheosam: But unknown cmd is a error.

os/drivers/audio/alc5658.c
btheosam: Already removed in .h file also as part of this commit itself.

os/drivers/audio/alc5658.c
btheosam: alc5658_getregs can be used to get the register values anytime for debugging alc5658.

os/drivers/audio/alc5658.c
btheosam: Right now, the error flags are defined in i2s.h. flags are populated with the values defined in i2s.h. But based on tinyalsa changes & changes in audio layer, this can be modified.

os/drivers/audio/alc5658.c
kishore-sn: No. Its not handled by enqueue request. 
Timeout happens much later after enqueue. It is indicated by result parameter of alc5658_rxtxcallback. But its being ignored.

kishore-sn: I think its better to just print an error message in alc5658_rxtxcallback when timeout error happens.

btheosam: Agreed

os/drivers/audio/alc5658.c
btheosam: agreed

os/drivers/audio/alc5658.c
Taejun-Kwon: No, it is not error case, it just undefined case for specific file operation(in this case audio operation of alc5658)
Any arg value can be sent to driver layer, so tizenrt usually using verbose for default value.

os/drivers/audio/alc5658.c
btheosam: Just the order is changed. It is needed for setting sample rate.

seinfra:
Code Rule Check OK.

PR #1168, opened 2018-01-03
Fix minor errata in audio utc

junmin-kim:
The value of fd can be 0.
Comparison of pointers should only use =, !=


seinfra:
Code Rule Check OK.

PR #1169, opened 2018-01-04
Check mkdir return value in audio_register

junmin-kim:
Concider mkdir fail case.
If it happens, audio_register failed


seinfra:
Code Rule Check OK.

PR #1172, opened 2018-01-04
Fix oops parameter in tinyalsa.c

junmin-kim:
Argument e of oops(struct pcm *pcm, int e, const char *fmt, ...)
must be >= 0


framework/src/tinyalsa/tinyalsa.c
```diff
@@ -442,7 +442,7 @@ static int pcm_mmap_transfer_areas(struct pcm *pcm, char *buf, unsigned int offs
 		pcm_areas_copy(pcm, pcm_offset, buf, offset, frames);
 		commit = pcm_mmap_commit(pcm, pcm_offset, frames);
 		if (commit < 0) {
-			return oops(pcm, commit, "failed to commit %d frames\n", frames);
+			return oops(pcm, -commit, "failed to commit %d frames\n", frames);
```
jsdosa: pcm_mmap_commit 

	if (pcm == NULL) {
		return -EINVAL;
	}

is that correct~?

seinfra:
Code Rule Check OK.

PR #1176, opened 2018-01-05
Avoid recursive abort by get_symbol in arm_assert.c

pradeepns85:
get_symbol() api does file system operations like open
seek, close etc. If there is a crash inside filesystem
it enters to abort mode and since get_symbol is used inside
abort_handler to print the symbol names infront of callstack,
It will call filesystem apis again. This would cause recursive
abort. This patch fixes this recursive abort.


seinfra:
Code Rule Check OK.

PR #1179, opened 2018-01-09
Fix problem with offset in wpa_receive_from not being updated correctly

kimusan:
In cases where a large amount of data is to be transferred over a non-UDP socket, the offset would not be updated correctly.
This patch makes sure that the offset is always updated according to the amount read.

### DIFF
--- a/external/wpa_supplicant/src/common/wpa_ctrl.c
+++ b/external/wpa_supplicant/src/common/wpa_ctrl.c
@@ -564,7 +564,7 @@ int wpa_ctrl_recvfrom(int sock, char *buf, size_t len
 	ssize_t amount = 0;
 	size_t recv_len;
 	char *pos;
-	int res;
+	int res = -1;
 #ifdef CONFIG_CTRL_IFACE_UDP
 	UNUSED(flags);
 #endif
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

external/wpa_supplicant/src/common/wpa_ctrl.c
```diff
@@ -568,23 +568,22 @@ int wpa_ctrl_recvfrom(int sock, char *buf, size_t len
 #ifdef CONFIG_CTRL_IFACE_UDP
 	UNUSED(flags);
 #endif
-	while ((amount >= 0) && ((amount + offset) < CTRL_HEADER_SIZE)) {
+	while ((offset += amount) < CTRL_HEADER_SIZE)) {
```
lokeshbv: check parenthesis mismatch remove `)` after `CTRL_HEADER_SIZE`

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

kimusan:
should be ready for pull now

seinfra:
Code Rule Check OK.

sunghan-chang:
@kimusan You made 5 commits for a topic. That is a wrong usage of git. Could you amend your original commit, ```Fix problem with offset in wpa_receive_from not being updated correctly```?

kimusan:
The change was made directly in github web editor so no way to amend. There is a "squash and merge" button when you take in the pull request. This will return the fix into being just one.

sunghan-chang:
I know that but it causes different pattern of merging. You can pull your patch-4 branch to latest and reset soft and then you can amend original one. Anyway this time I will merge this using squash option.

PR #1184, opened 2018-01-10
config/artik053s, artik055s: update defconfig to latest

sunghan-chang:
disable pm, modify ram_size, disable unnecessary disable_os_api


seinfra:
Code Rule Check OK.

PR #1196, opened 2018-01-15
docs: add new doc, HowToDebug.md

sunghan-chang:
This shows how to debug TizenRT.
1. how to use ramdump
2. describe debug utils like ps, stackmonitor and so on
More item will be added.


seinfra:
Code Rule Check OK.

PR #1197, opened 2018-01-15
armv7-m/up_assert: wrap CONFIG_ARMV7M_MPU to 'mpu_show_regioninfo'

lokeshbv:
MPU regions info should be displayed in asserts only when `CONFIG_ARMV7M_MPU` is enabled.


seinfra:
Code Rule Check OK.

PR #1198, opened 2018-01-16
framework/st_things : Resolves issue where time is reset on reboot

JeongJunSik:
Resolves issue where time is reset on reboot
Solution: Set the system time by taking the time from the NTP server

### DIFF
--- a/framework/src/st_things/things_stack/things_stack.c
+++ b/framework/src/st_things/things_stack/things_stack.c
@@ -51,6 +51,10 @@
 #include "utils/things_rtos_util.h"
 #endif
 
+#ifdef CONFIG_NETUTILS_NTPCLIENT
+#include <protocols/ntpclient.h>
+#endif
+
 #define TAG "[things_stack]"
 
 typedef void *(*pthread_func_type)(void *);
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

sunghan-chang:
Please check commit rule, 5th. You should fill commit description.

seinfra:
Code Rule Check OK.

framework/src/st_things/things_stack/things_stack.c
```diff
@@ -194,6 +220,12 @@ void things_wifi_sta_connected(wifi_manager_result_e res)
 	}
 	THINGS_LOG_D(THINGS_INFO, TAG, "T%d --> %s", getpid(), __FUNCTION__);
 
+#ifdef CONFIG_NETUTILS_NTPCLIENT
+	sync_time_from_ntp();
```
pillip8282: if things_wifi_sta_connected is the callback from wifi_manager then it would be better to call 'sync_time_from_ntp()' in other place. you should return the callback asap because it is called from wpa_supplicant task.
wpa supplicant task would be delayed.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check Result:
framework/src/st_things/things_stack/things_stack.c:192: ERROR: [SPC_M_SEP] space required after that close brace '}'

framework/src/st_things/things_stack/things_stack.c
```diff
@@ -173,8 +177,36 @@ int things_register_easysetup_state_func(things_get_easysetup_state_func_type fu
...
+#ifdef CONFIG_NETUTILS_NTPCLIENT
+
+#define MAX_NTP_SERVER_NUM 2
+
+static void ntp_link_error(void)
+{
+	THINGS_LOG_D(THINGS_INFO, TAG, "ntp_link_error() callback is called.");
+}
+
+void sync_time_from_ntp(void)
```
davidfather: This function may be static.

PR #1201, opened 2018-01-17
artik: Explain to erase OTA part for factory reset

rzr:
For a safer factory reset, it's safer to erase OTA partition.

I came to the case, where there was junk in partition,
instead of actual "Over the Air" image
And do_checkupdate failed on crc32 which caused a reset of uboot.

So to reduce unexpected cases, the partition is just erased,
then comparaison test on ${otapart} header will be skipped (==0xffffffff).

For more insights check file:
artik-05x/include/configs/artik05x.h

URL: https://developer.artik.io/forums/t/solved-how-to-factory-reset-artik055s/5254
Change-Id: I19de266fdfe134fe3cc9574f5cc2482215d9877a


seinfra:
Code Rule Check OK.

PR #1203, opened 2018-01-18
Initialize usermq only when first audio_open

junmin-kim:
usermq should not be initialize after first audio_open,
because previous thread may using that usermq


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1204, opened 2018-01-18
tools: Drop relic CONFIG_NSH_ROMFSETC variable

rzr:
This is comming from NuttX's NSH and no more used in TASH,
as there is no startup script on TizenRT.

Change-Id: I3b6fa61e6410dbd4f8eee1ac8145bab56db391d0


seinfra:
Code Rule Check OK.

os/tools/mkromfsimg.sh
```diff
@@ -82,7 +82,6 @@ if [ ! -r $topdir/.config ]; then
     exit 1
 fi
 
-romfsetc=`grep CONFIG_NSH_ROMFSETC= $topdir/.config | cut -d'=' -f2`
```
sunghan-chang: Here are many CONFIG_NSH_XX which are not used in TizenRT. I think we had better cleanup all of content in this file if we don't have any reason to remove this only immediately.

rzr:
this is introducing a regression, I am fixing it now, stay tunned

PR #1205, opened 2018-01-19
framework/wifi_manager: add wifi manager configuration get/set features

pillip8282:
Add wifi manager APIs to store the connected AP information in persistent storage

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test.c
+++ b/apps/examples/wifi_manager_sample/wm_test.c
@@ -201,6 +201,36 @@ static const wifi_manager_ap_crypto_type_e crypto_type_table[] = {
 	WIFI_MANAGER_CRYPTO_UNKNOWN,               /**<  unknown encryption             */
 };
 
+void
+print_wifi_ap_profile(wifi_manager_ap_config_s *config, char *title)
+{
+	printf("====================================\n");
+	if (title) {
+		printf("%s\n", title);
+	}
+	printf("------------------------------------\n");
+	printf("SSID: %s\n", config->ssid);
+	printf("SECURITY TYPE: %s\n", wifi_test_auth_method[config->ap_auth_type]);
+	if (config->ap_auth_type != WIFI_MANAGER_AUTH_OPEN) {
+		printf("PASSWORD: %s\n", config->passphrase);
+	}
+	printf("====================================\n");
+}
+
+void
+print_wifi_softap_profile(wifi_manager_softap_config_s *config, char *title)
+{
+	printf("====================================\n");
+	if (title) {
+		printf("%s\n", title);
+	}
+	printf("------------------------------------\n");
+	printf("SSID: %s\n", config->ssid);
+	printf("channel: %d\n", config->channel);
+	printf("PASSWORD: %s\n", config->passphrase);
+	printf("====================================\n");
+}
+
 
 wifi_manager_ap_auth_type_e
 get_auth_type(const char *method)
@@ -328,6 +358,60 @@ void wm_scan_done(wifi_manager_scan_info_s **scan_result, wifi_manager_scan_resu
  * Control Functions
  */
 void
+wm_reset_info(void *arg)
+{
+	WM_TEST_LOG_START;
+	wifi_manager_result_e res = wifi_manager_remove_config();
+	if (res != WIFI_MANAGER_SUCCESS) {
+		printf("Get AP configuration failed\n");
+		return;
+	}
+
+	WM_TEST_LOG_END;
+}
+
+
+void
+wm_get_info(void *arg)
+{
+	WM_TEST_LOG_START;
+	wifi_manager_ap_config_s apconfig;
+	wifi_manager_result_e res = wifi_manager_get_config(&apconfig);
+	if (res != WIFI_MANAGER_SUCCESS) {
+		printf("Get AP configuration failed\n");
+		return;
+	}
+	print_wifi_ap_profile(&apconfig, "Stored Wi-Fi Infomation");
+
+	WM_TEST_LOG_END;
+}
+
+void
+wm_set_info(void *arg)
+{
+	WM_TEST_LOG_START;
+	struct options *ap_info = (struct options *)arg;
+	wifi_manager_ap_config_s apconfig;
+	strncpy(apconfig.ssid, ap_info->ssid, 33);
+	apconfig.ssid_length = strlen(ap_info->ssid);
+	apconfig.ap_auth_type = ap_info->auth_type;
+	if (apconfig.ap_auth_type != WIFI_MANAGER_AUTH_OPEN) {
+		strncpy(apconfig.passphrase, ap_info->password, 64);
+		apconfig.passphrase_length = strlen(ap_info->password);
+		apconfig.ap_crypto_type = ap_info->crypto_type;
+	}
+
+	print_wifi_ap_profile(&apconfig, "Set AP Info");
+
+	wifi_manager_result_e res = wifi_manager_save_config(&apconfig);
+	if (res != WIFI_MANAGER_SUCCESS) {
+		printf("Save AP configuration failed\n");
+		return;
+	}
+	WM_TEST_LOG_END;
+}
+
+void
 wm_start(void *arg)
 {
 	WM_TEST_LOG_START;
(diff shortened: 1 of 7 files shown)


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -201,6 +201,36 @@ static const wifi_manager_ap_crypto_type_e crypto_type_table[] = {
...
 
+void
+print_wifi_ap_profile(wifi_manager_ap_config_s *config, char *title)
+{
+	printf("====================================\n");
+	if (title) {
+		printf("%s\n", title);
+	}
+	printf("------------------------------------\n");
+	printf("SSID: %s\n", config->ssid);
```
sunghan-chang: Don't you need to check null dereferece for ```config```?

framework/include/wifi_manager/wifi_manager.h
```diff
@@ -228,8 +228,35 @@ wifi_manager_result_e wifi_manager_disconnect_ap(void);
  */
 wifi_manager_result_e wifi_manager_scan_ap(void);
 
-#endif
+/**
+ * @brief Save the AP configuration at persistent storage
+ * @details @b #include <wifi_manager/wifi_manager.h>
+ * @param[in] AP configuration infomation should be given including ssid, channel, and passphrase.
+ * @return On success, WIFI_MANAGER_SUCCESS (i.e., 0) is returned. On failure, non-zero value is returned.
+ * @since Tizen RT v1.1
```
sunghan-chang: 1. Let's change ```v1.1``` to ```v1.x```. Same for all of new APIs
2. You should apply ACR.

framework/include/wifi_manager/wifi_manager.h
```diff
@@ -228,8 +228,35 @@ wifi_manager_result_e wifi_manager_disconnect_ap(void);
  */
 wifi_manager_result_e wifi_manager_scan_ap(void);
 
-#endif
+/**
+ * @brief Save the AP configuration at persistent storage
+ * @details @b #include <wifi_manager/wifi_manager.h>
+ * @param[in] AP configuration infomation should be given including ssid, channel, and passphrase.
```
sunghan-chang: When you use param[xx], you should write a name of argument first. That is a rule of doxygen.
```
* @param[in] config AP ......
```

framework/src/wifi_manager/wifi_profile.c
```diff
@@ -0,0 +1,238 @@
+#include <tinyara/config.h>
```
sunghan-chang: You should add Samsung copyright and license at top of file.

framework/src/wifi_manager/wifi_profile.c
```diff
@@ -0,0 +1,238 @@
...
+		sprintf(conv, "%d", tmp);				\
+		data = tmp;								\
+		pos += (strlen(conv) + DELI_LEN);		\
+	} while (0)
+
+/*
+ * Internal Functions
+ */
+
+int _wifi_profile_serialize(char *buf, uint32_t buf_size, wifi_manager_ap_config_s *config)
```
sunghan-chang: This looks internal API. If it is, let's add ```static```. Same for all of ```_xxx``` function in this file

framework/src/wifi_manager/wifi_profile.c
```diff
@@ -0,0 +1,238 @@
...
+		return WIFI_UTILS_FILE_ERROR;
+	}
+#endif
+	return WIFI_UTILS_SUCCESS;
+}
+
+wifi_utils_result_e wifi_profile_write(wifi_manager_ap_config_s *config)
+{
+	char buf[256];
+	int ret = 0, len = 0;
```
sunghan-chang: Please define a new variable in different line. That is a coding rule recommandation.

framework/src/wifi_manager/wifi_profile.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _WIFI_PROFILE_H__
```
sunghan-chang: Please add copyright and license

seinfra:
Code Rule Check OK.

apps/examples/wifi_manager_sample/wm_test.c
pillip8282: I think we don't need check. because it's simple print function.

framework/include/wifi_manager/wifi_manager.h
pillip8282: Can you give me a example? I just follow the above templete

framework/src/wifi_manager/wifi_profile.c
pillip8282: Ok I will add

framework/src/wifi_manager/wifi_profile.c
pillip8282: Ok I will fix

framework/src/wifi_manager/wifi_profile.c
pillip8282: Ok I'll fix

framework/src/wifi_manager/wifi_profile.h
pillip8282: Ok I'll fix

framework/include/wifi_manager/wifi_manager.h
```diff
@@ -228,8 +228,35 @@ wifi_manager_result_e wifi_manager_disconnect_ap(void);
...
+ * @return On success, WIFI_MANAGER_SUCCESS (i.e., 0) is returned. On failure, non-zero value is returned.
+ * @since Tizen RT v1.x
+ */
+
+wifi_manager_result_e wifi_manager_get_config(wifi_manager_ap_config_s *config);
 
+/**
+ * @brief Remove the AP configuration which was saved
+ * @details @b #include <wifi_manager/wifi_manager.h>
+ * @param[in] none
```
sunghan-chang: I think you should remove this line. It might print "none" as argument name without any description.

seinfra:
Code Rule Check OK.

framework/src/wifi_manager/wifi_manager.c
```diff
@@ -974,3 +982,40 @@ wifi_manager_result_e wifi_manager_scan_ap(void)
 	wifi_manager_result_e res = _handle_request(&msg);
 	return res;
 }
+
+
+wifi_manager_result_e wifi_manager_save_config(wifi_manager_ap_config_s *config)
+{
+	LOCK_WIFIMGR;
+	wifi_utils_result_e res = wifi_profile_write(config);
+	if (res != WIFI_UTILS_SUCCESS) {
```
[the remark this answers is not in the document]
davidfather: This option is also possible: just make 'wifi_manager_save_config' and 'wifi_manager_remove_config' apis into internal ones. That is, wifi-manager just maintains wifi profiles inside automatically even without any requests.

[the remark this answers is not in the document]
pillip8282: It only support single profile.
If we maintain the wifi profile inside, then a wifi profile is always set even though user doesn't want.
And maintaining inside requires file operation. so I think we should decide the policy very carefully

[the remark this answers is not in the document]
davidfather: Let me know when you make sure that only one profile is handled without any bugs.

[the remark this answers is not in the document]
pillip8282: I worried that if smaller data is insert then the previous data would be remain at the end of the file. But it's not true. fopen will erase previous data. So we don't need to care about it.
------------------------------------------------------------------------------------------------------
w      Truncate file to zero length or create text file for writing.  The stream is positioned at the beginning of the file.
w+     Open for reading and writing.  The file is created if it does not exist, otherwise it is truncated.  The stream is positioned at the beginning of the file.
------------------------------------------------------------------------------------------------------

framework/src/wifi_manager/wifi_manager.c
```diff
@@ -974,3 +982,40 @@ wifi_manager_result_e wifi_manager_scan_ap(void)
 	wifi_manager_result_e res = _handle_request(&msg);
 	return res;
 }
+
+
+wifi_manager_result_e wifi_manager_save_config(wifi_manager_ap_config_s *config)
+{
+	LOCK_WIFIMGR;
```
davidfather: This mutex is required to protect wifi_manager operations and related common variables.
The file, where the config is saved, can be protected by another mutex.

pillip8282: Ok, I will fix. I think that these functions doesn't require mutex. because it doesn't refer global variables

seinfra:
Code Rule Check OK.

PR #1207, opened 2018-01-22
framework/iotbus/gpio: set a dev to NULL on close

sunghan-chang:
When gpio is closed, even though dev resource is freed, dev value
is not set to NULL. It causes getting wrong response from gpio function
after calling close function.


seinfra:
Code Rule Check OK.

PR #1209, opened 2018-01-22
arch/src/armv7-m: Add function to show tasks info

lokeshbv:
- Adds 'up_showtasks' function in crash dump state
- 'up_showtasks' prints list of all tasks in the system
with PID and amount of stack used

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -179,6 +179,43 @@ static inline void up_registerdump(void)
 #endif
 
 /****************************************************************************
+ * Name: up_taskdump
+ ****************************************************************************/
+
+#ifdef CONFIG_STACK_COLORATION
+static void up_taskdump(FAR struct tcb_s *tcb, FAR void *arg)
+{
+	/* Dump interesting properties of this task */
+
+#if CONFIG_TASK_NAME_SIZE > 0
+	lldbg("%s: PID=%d Stack Used=%lu of %lu\n",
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -178,6 +178,43 @@ static inline void up_registerdump(void)
 #define up_registerdump()
 #endif
 
+/****************************************************************************
+ * Name: up_taskdump
+ ****************************************************************************/
+
+#ifdef CONFIG_STACK_COLORATION
+static void up_taskdump(FAR struct tcb_s *tcb, FAR void *arg)
```
sunghan-chang: Why is it taskdump? not stackinfodump?

sunghan-chang: Please forget my previous comment. But anyway ```taskdump``` is not an approprietary name I think. 
1. change a name to ```up_tasklist```?
2. add more information like tcb_s address and content because name is dump.

lokeshbv: It's copy of `up_taskdump` from below.  Do I need to change?
https://github.com/Samsung/TizenRT/blob/c356fc184471438d8f8b2b4d5b84fee396b814ad/os/arch/arm/src/armv7-r/arm_assert.c#L474

sunghan-chang: I don't think that is a good name. But it comes from Nuttx. Let's keep original name for maintainance.

PR #1213, opened 2018-01-24
ARMv7-M: Fix logic in up_sechedule_sigaction

lokeshbv:
- Fix logic that determines if there is a pending signal action before
	scheduling the next signal action.
- Both the test and the scheduling action need to be atomic.

This patch is back ported from Nuttx


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1214, opened 2018-01-24
armv7: Move log header of up_showtasks inside it's function definition

lokeshbv:
It is not relavant to print log header without actual log content.
So it is better to move log header of up_showtasks inside it's function definition.


seinfra:
Code Rule Check OK.

PR #1216, opened 2018-01-25
docs: add the byacc package in bash command window

sunghan-chang:
In some environment, byacc should be installed to use Kconfig-frontend.
We wrote that on description, "(or byacc if supported)".
But someone doesn't see it and just copy commands from bash window and execute.
To prevent an error, let's defaultly add that in bash window.


seinfra:
Code Rule Check OK.

PR #1217, opened 2018-01-25
Remove three ITC from itc_arastorage_main.c

bipin4321:
Remove below ITCs:
itc_arastorage_db_print_header_n_after_deinit
itc_arastorage_cursor_get_count_n_after_deinit
itc_arastorage_cursor_get_row_n_after_deinit

Signed-off-by: bipinkumar <bipin.k@samsung.com>

### DIFF
--- a/apps/examples/testcase/ta_tc/arastorage/itc/itc_arastorage_main.c
+++ b/apps/examples/testcase/ta_tc/arastorage/itc/itc_arastorage_main.c
@@ -293,28 +293,6 @@ void itc_arastorage_db_init_deinit_p_multi_time(void)
 }
 
 /**
-* @testcase             itc_arastorage_db_print_header_n_after_deinit
-* @brief                To check print of the related information : relation, attribute name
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

sunghan-chang:
@bipin4321 Please write the reason why you remove them on commit description.

bipin4321:
@manoj11588612 please review

seinfra:
Code Rule Check OK.

jeongarmy:
please update commit message about the reason why these testcase are invalid.
When we call db_deinit, real data of db will be deinitialized, not cursor.
That means we can get a cursor data after db_deinit and it is not a wrong operation.
When we try to access db data through cursor, we can meet an error.

samegautam:
@jeongarmy  : we have discussed with developer on same.
Developer response was "I think these are not needed. remains which call db functions are enough for testing 'after deinit'.
".

seinfra:
Code Rule Check OK.

PR #1219, opened 2018-01-25
libc: remove duplicated imaxabs function implementation

sunghan-chang:
The prototype of imaxabs function is in inttype.h, not stdlib.h
so that this file was moved from libc/stdlib to libc/inttypes.
But old one is not deleted and causes compilation warning as shown below:

make[1]: Entering directory `/home/sunghan/Work/TinyAra/TizenRT_forked/lib/libc'
Makefile:132: target `bin/lib_imaxabs.o' given more than once in the same rule.


seinfra:
Code Rule Check OK.

PR #1221, opened 2018-01-25
NTPC Daemon : Set default ntp server on ntpc daemon

JeonginKim:
ntp server address or interval time are not entered, it is possible to connect to the ntp server set in the system.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1222, opened 2018-01-29
NTPC Daemon : Set default ntp server on ntpc daemon

JeonginKim:
ntp server address or interval time are not endtered, it is possible to connect to the ntp server set in the system.


external/ntpclient/Kconfig.protocol
```diff
@@ -25,6 +25,18 @@ config NETUTILS_NTPCLIENT_SIGWAKEUP
...
+	string "NTP client default server"
+	default "0.ubuntu.pool.ntp.org"
+
+config NETUTILS_NTPCLIENT_DEFAULT_SERVER_PORT
+	int "NTP client default server port"
+	default 123
+
+config NETUTILS_NTPCLIENT_DEFAULT_INTERVAL_SECONDS
+	int "NTP client default interval seconds"
+	default 60
```
JeongJunSik: It seems to be too frequent in 60 seconds. How about once a day?

seinfra:
Code Rule Check OK.

external/ntpclient/Kconfig.protocol
JeonginKim: The interval time has been modified to a 24-hour period.

seinfra:
Code Rule Check OK.

PR #1223, opened 2018-01-29
iotbus/gpio: change a prameter type of iotbus_gpio_close to pointer of iotbus_gpio_context_h

jeongarmy:
iotbus_gpio_close closes the gpio context and frees all resources pointed by dev.
We need to pass the address of dev to set dev as NULL.


seinfra:
Code Rule Check OK.

sangwon03:
please check API change process

jeongarmy:
yes, I know. I will do that after this is merged

PR #1231, opened 2018-01-30
network/wifi_manager: fix copying mac address bug

pillip8282:
if mac address is "fe:00:..." then it will lose data after ":00:.."
because strncpy will stop copy when character is null.
so using memcpy() instead of strncpy() will fix the issue.

### DIFF
--- a/framework/src/wifi_manager/wifi_manager.c
+++ b/framework/src/wifi_manager/wifi_manager.c
@@ -627,9 +627,8 @@ wifi_manager_result_e _handler_on_uninitialized_state(_wifimgr_msg_s *msg)
 		return WIFI_MANAGER_FAIL;
 	}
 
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

framework/src/wifi_manager/wifi_manager.c
```diff
@@ -629,7 +629,7 @@ wifi_manager_result_e _handler_on_uninitialized_state(_wifimgr_msg_s *msg)
 
 	/* update wifi_manager info by being protected by w_info_mutex */
```
davidfather: Where is w_info_mutex?
You need to fix or remove the comment.

pillip8282: ok, I will fix

seinfra:
Code Rule Check OK.

PR #1237, opened 2018-01-31
framework/st_things : increase MTU size

JeongJunSik:
- increase MTU size for downloading data from server


seinfra:
Code Rule Check OK.

PR #1240, opened 2018-02-01
configs : update all board Make.defs to remove cxx include path

manoharahk:
include/cxx && include/uClibc++ directories are not present.
hence this patch updates all board configuration Make.defs to remove
these directory path from the include


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1247, opened 2018-02-05
Remove hardcoded directories in framework/Makefile

junmin-kim:
Using wildcard, all Make.defs are included


seinfra:
Code Rule Check OK.

PR #1249, opened 2018-02-05
NTPC Daemon : change default ntp server

JeonginKim:
changed default ntp server address to stable ntp server.


seinfra:
Code Rule Check OK.

PR #1251, opened 2018-02-05
framework/st_things : change default ntp server

JeongJunSik:
- changed default ntp server address to stable ntp server


seinfra:
Code Rule Check OK.

PR #1256, opened 2018-02-07
framework/st_things : Error return when wifi connection fails

JeongJunSik:
- In Station Mode, if the connection to the Home AP fails, return an error


seinfra:
Code Rule Check OK.

PR #1257, opened 2018-02-07
framework/st_things : Delete wifi profile information on reset

JeongJunSik:
- Delete stored wifi profile information

### DIFF
--- a/framework/src/st_things/things_stack/things_stack.c
+++ b/framework/src/st_things/things_stack/things_stack.c
@@ -317,6 +317,8 @@ bool try_turn_on_soft_ap()
 	wifi_manager_info_s info;
 	wifi_manager_get_info(&info);
 
+	wifi_manager_remove_config();
+
 	if (info.mode != SOFTAP_MODE) {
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

framework/src/st_things/things_stack/things_stack.c
```diff
@@ -317,6 +317,8 @@ bool try_turn_on_soft_ap()
 	wifi_manager_info_s info;
 	wifi_manager_get_info(&info);
 
+	wifi_manager_remove_config();
```
davidfather: When this function gets failed, what happens?

JeongJunSik: There is no big problem even if wifi_profile information is not erased.
st_things maintains provisioning data separately and requests join to sta home ap through the information.
Also, if Easy-Setup has been resumed after the reset, it will not be a problem to save the profile again.

seinfra:
Code Rule Check OK.

PR #1259, opened 2018-02-08
external/webclient : add chunked encoding and refactoring callback flow

JeonginKim:
Add chunked type encoding processing
Callback is called as soon as data is received


seinfra:
Code Rule Check OK.

external/webclient/webclient.c
```diff
@@ -1123,7 +1124,6 @@ static pthread_addr_t wget_base(void *arg)
 	int sockfd = -1;
 	int ret;
 	int buf_len, sndlen, len;
```
sangwon03: do we need 'buf_len' variable?

seinfra:
Code Rule Check OK.

external/webclient/webclient.c
JeonginKim: It is used when sending, but not when receiving. It's corrected.

seinfra:
Code Rule Check OK.

PR #1261, opened 2018-02-08
network/wifimanager: Enhance cancelling reconnect.

pillip8282:
API caller should waits interval time even though he want to cancel the reconnect process.
it doesn't matter when the interval is short, but it became a problem if interval time is long.
So instead of using sleep(), wifi_manager calls select() to terminate reconnect process when
disconnect API(cancel the reconnect) is invoked.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -556,6 +558,21 @@ wm_disconnect(void *arg)
...
+{
+	WM_TEST_LOG_START;
+	wifi_manager_result_e res = WIFI_MANAGER_SUCCESS;
+	/* Disconnect AP */
+	res = wifi_manager_disconnect_ap();
+	if (res != WIFI_MANAGER_SUCCESS) {
+		printf("disconnect fail (%d)\n", res);
+		return;
+	}
+	WM_TEST_LOG_END;;
```
sunghan-chang: Let's remove one semicolon.

pillip8282: Ok I will fix

seinfra:
Code Rule Check OK.

framework/src/wifi_manager/wifi_manager.c
```diff
@@ -28,14 +30,21 @@
 #include "wifi_utils.h"
 #include "wifi_profile.h"
 
-#define NUM_OF_STATE_HANDLER 8 // the number of states
+#undef ndbg
+#define ndbg printf
+#undef nvdbg
+#define nvdbg printf
+
+#define NUM_OF_STATE_HANDLER 11 // the number of states
```
davidfather: The number of states is 11.
but, the number of handlers is not 11, but 10.

seinfra:
Code Rule Check OK.

framework/src/wifi_manager/wifi_manager.c
pillip8282: Ok I will fix

seinfra:
Code Rule Check OK.

PR #1267, opened 2018-02-12
ALC5658's support for prepare through ioctl.

btheosam:
ALC5658 doesn't support prepare ioctl to handle xrun. This commit
provides changes in alc5658 to support (tinyalsa's) prepare.

### DIFF
--- a/os/drivers/audio/alc5658.c
+++ b/os/drivers/audio/alc5658.c
@@ -974,8 +974,31 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
 	switch (cmd) {
 
 	case AUDIOIOC_PREPARE: {
-		/* TODO: Need to be implemented later */
-		audvdbg("AUDIOIOC_PREPARE: No Action Would be taken now \n");
+		/* This is to aid audio prepare, this would stop dma, i2s and
+		cleanup all queues. Idea behind is to use this in xrun cases to stop
+		followed by (restart) but restart is not part of this */
+		audvdbg("AUDIOIOC_PREPARE: Stop dma,i2s and cleanup qs\n");
+
+		/* Take semaphore */
+		alc5658_takesem(&priv->devsem);
+
+		/* Stop dma, i2s */
+		I2S_STOP(priv->i2s);
+
+		/* Switch alc status */
+		priv->running = false;
+
+		/* Clean up alc queue */
+		sq_entry_t *tmp = NULL;
+		for (tmp = (sq_entry_t *)sq_peek(&priv->pendq); tmp; tmp = sq_next(tmp)) {
+			sq_rem(tmp, &priv->pendq);
+		}
+		sq_init(&priv->pendq);
+
+		/* TOCHECK: Do we need to run stop alc stop script? ??? for now no */
+
+		/* Give semaphore */
+		alc5658_givesem(&priv->devsem);
 	}
 	break;
 	case AUDIOIOC_HWRESET: {


seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.c
```diff
@@ -974,8 +974,31 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
 	switch (cmd) {
 
 	case AUDIOIOC_PREPARE: {
-		/* TODO: Need to be implemented later */
-		audvdbg("AUDIOIOC_PREPARE: No Action Would be taken now \n");
+		/* This is to aid Tinyalsa pcm prepare, this would stop dma, i2s and
```
Taejun-Kwon: You'd better remove comment regarding Tinyalsa, AUDIOIOC_PREPARE can be sent by any layer where using ioctl.

seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.c
```diff
@@ -974,8 +974,31 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
...
+
+		/* Clean up alc queue */
+		sq_entry_t *tmp = NULL;
+		for (tmp = (sq_entry_t *)sq_peek(&priv->pendq); tmp; tmp = sq_next(tmp)) {
+			sq_rem(tmp, &priv->pendq);
+		}
+		sq_init(&priv->pendq);
+
+		/* TOCHECK: Do we need to run stop alc stop script? ??? for now no */
+
```
kishore-sn: In my opinion, pcm_prepare is being done to restart the pcm.. so either we have to call codec_stop or atleast we have to call codec_reset_script ... but need to make sure that codec_reset_script only resets the codec without altering the codec configurations.. like sample rate, etc..

os/drivers/audio/alc5658.c
```diff
@@ -974,8 +974,31 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
...
+		/* Stop dma, i2s */
+		I2S_STOP(priv->i2s);
+
+		/* Switch alc status */
+		priv->running = false;
+
+		/* Clean up alc queue */
+		sq_entry_t *tmp = NULL;
+		for (tmp = (sq_entry_t *)sq_peek(&priv->pendq); tmp; tmp = sq_next(tmp)) {
+			sq_rem(tmp, &priv->pendq);
```
kishore-sn: This will cause problem if pcm_prepare is called in pcm_start in tinyalsa.
- i will remove pcm_prepare call from pcm_start in tinyalsa.  
- then pcm_prepare will be called onlly for underrun cases..
- if user calls pcm_prepare during normal operation, it is his responsibility to enqueue new buffers..

btheosam: @Taejun-Kwon  A clear definition on pcm_prepare and when it has to be used has to be defined. If we follow tinyalsa, it has to be called at pcm_start.

os/drivers/audio/alc5658.c
btheosam: NO. Codec reset script would reset codec that means all registers will have default values. It has to be reconfigured again. Not a good idea. But, if what you are saying is right, then using pcm_prepare to handle xrun may be wRONG. Instead tinyalsa has to stop and start again. Correct?

os/drivers/audio/alc5658.c
```diff
@@ -974,8 +974,31 @@ static int alc5658_ioctl(FAR struct audio_lowerhalf_s *dev, int cmd, unsigned lo
...
+
+		/* Take semaphore */
+		alc5658_takesem(&priv->devsem);
+
+		/* Stop dma, i2s */
+		I2S_STOP(priv->i2s);
+
+		/* Switch alc status */
+		priv->running = false;
+
```
kishore-sn: Please use I2S_PAUSE instead of STOP.
Do not set running state to false.

os/drivers/audio/alc5658.c
kishore-sn: As discussed offline, "prepare" will not do any buffer handling. 
Please remove buffer handling code.

os/drivers/audio/alc5658.c
kishore-sn: Please move codec init and configuration code from start() api to this place in "prepare".
After codec configuration, pls call I2S_RESUME.

btheosam:
This changes are moved as another PR & hence closing this.

PR #1268, opened 2018-02-12
build/artik05x: support erasing ota partition through make command

sunghan-chang:
Erasing ota partition is needed to support FOTA. This commit add
ERASE_OTA option on make command to do it.


seinfra:
Code Rule Check OK.

PR #1269, opened 2018-02-12
external/webclient : fix chunked data length

JeonginKim:
modifying the length value when to call callback function


seinfra:
Code Rule Check OK.

PR #1270, opened 2018-02-12
arch/arm: add new feature and rename an artik crash file

sunghan-chang:
1. arch/artik05x: rename artik053_crashdump to artik05x_crashdump
2. arch/arm: add autoreset on assert


seinfra:
Code Rule Check OK.

PR #1272, opened 2018-02-13
configs/artik053, sidk_s5jt200: disable ITCs on kernel and network de…

sunghan-chang:
…faultly

ITCs on kernel and network TC have some issues so that we disable them
defaultly.


seinfra:
Code Rule Check OK.

PR #1274, opened 2018-02-13
Revert permission for le_tc/kernel/tc_task.c

thapav:
This patch reverts the permission for the file:
le_tc/kernel/tc_task.c from 777 to 644


seinfra:
Code Rule Check OK.

PR #1276, opened 2018-02-15
libcxx : move libc++ source to external

manoharahk:
This patch moves the libc++ source from lib/libxx/libcxx to
external/libcxx

Current Approach:
===============================================================================
[1] Using only Basic C++ (libxx) :
	- Basic C++ support is present in lib/libxx
	- Headers are present in os/include/libcxx (since both libxx & libcxx
		headers are merged)
	- When selected it creates static library named *libcxx.a* and linked to
		final binary.
[2] Using LLVM libc++ (libcxx) :
	- It uses basic functionality from lib/libxx as well to provide the
		complete c++ support on tizenrt.
	- Libxx is also a placeholder to support the ABI related functionality
		for libcxx in tizenrt
	- llvm libc++ library is present in lib/libxx/libcxx
	- Headers are present in os/include/libcxx
	- When selected libcxx builds along with libxx and creates single static
		library named *libcxx.a* and linked to final binary

New Approach:
===============================================================================
- libcxx is moved to external directory.
[1] Using only Basic C++ (libxx) :
	- Basic C++ support is present in lib/libxx
	- When selected it creates static library named *libxx.a* and linked to
		final binary.
	- Headers are present in external/include/libcxx (since both libxx &
		libcxx headers are merged and also moved to external)
[2] Using LLVM libc++ (libcxx) :
	- It uses basic functionality from lib/libxx as well to provide the
		complete c++ support on tizenrt.
	- libxx is also a placeholder to support the ABI related functionality
		for libcxx in tizenrt
	- llvm libc++ library is present in external/libcxx
	- Headers are present in external/include/libcxx
	- When selected,
		- libxx builds and creates a static library named *libxx.a* and
			linked to final binary.
		- libcxx builds and creates a static library named *libcxx.a* and
			linked to final binary.


seinfra:
Code Rule Check OK.

sunghan-chang:
1. Why do you make llvm as sparated library from external even if it is in external?
2. Why do you rename libcxx to libxx for lib folder? What's the meaning of libxx? And last time we discussed about Nuttx's libcxx and we decided deleting it. What is remaining file on libxx?

sunghan-chang:
Does LLVM replace basic cxx or add addtional functionality from basic cxx?

manoharahk:
1. Because I dint want libc++ to be part of libexternal.a. I wanted it to be libcxx.a to clearly distinguish at the link time and as well to see the size information like how much libc++ is adding to the total binary.
2. It doesn't completely replace cxx(libxx) . Original libxx is still required when llvm libc++ is used. llvm libc++ adds more functionality especially STL.

sunghan-chang:
For first query and answer.
Then why don't you merge libcxx with libxx?
1. When you check a size of c++, you should sum them.
2. They are a same topic as c++ library. I can't distingush what are they only with name, libxx and libcxx. Could you let me know the reason why we keep two libraries for c++?

manoharahk:
If we have to make a single library, then the original existing libxx also to be moved to external. Is it okay?

sunghan-chang:
No. We can merge them from different folders. To do it, we don't need to move libxx to external. I don't like making a library from outside of folder, but in your change, external has different libraries.

manoharahk:
Ok. Please let me know how to merge and create a single library.
Then libcxx.a will be created by libxx Makefile, but how to merge the objects compiled in external libcxx to libcxx.a?

sunghan-chang:
@manoharahk I am sorry for delay. Could you refer Makefile of hello example?
```
ifeq ($(CONFIG_WINDOWS_NATIVE),y)
  BIN = ..\..\libapps$(LIBEXT)
else
ifeq ($(WINTOOL),y)
  BIN = ..\\..\\libapps$(LIBEXT)
else
  BIN = ../../libapps$(LIBEXT)
endif
endif

.built: $(OBJS)
	$(call ARCHIVE, $(BIN), $(OBJS))
	@touch .built
```

manoharahk:
How about compiling(make) external/libcxx as part of libxx/libcxx.a target and add objects to libxx/libcxx.a like below?
1. In os/LibTargets.mk,
 libxx$(DELIM)libcxx$(LIBEXT): context
        $(Q) $(MAKE) -C $(LIB_DIR)$(DELIM)libxx TOPDIR="$(TOPDIR)" libcxx$(LIBEXT) KERNEL=n
+ifeq ($(CONFIG_LIBCXX),y)
+       $(Q) $(MAKE) -C $(EXTDIR)$(DELIM)libcxx TOPDIR="$(TOPDIR)" all KERNEL=n
+endif

2. In external/libcxx/Makefile,
BIN = $(TOPDIR)/../lib/libxx/libcxx$(LIBEXT)

The result will be single static library libcxx.a combining both libxx and external/libcxx.

sunghan-chang:
@manoharahk It might work. But does not it cause confusing? The llvm is under external but it doesn't build by external's Makefile. Is it same as Kconfig?

manoharahk:
Dear Mr.Chang,
It is compiling using external/libcxx/Makefile directly but not via external/Makefile.
And we don't have separate Kconfig entires in external/libcxx/. 
Libc++ Kconfig entries are present in libxx/Kconfig.

manoharahk:
Dear @sunghan-chang ,

If the above is approach is okay, then shall i push the updated patch?

sunghan-chang:
@manoharahk I am sorry for delay.
In external, there are many modules which belongs to network, kernel, etc. But they are built as external, not each target. Why should we exclude llvm only from external and include to libc++? That's what I worry about. Why only for llvm?

manoharahk:
Dear @sunghan-chang ,

In case of llvm libc++, we need a separate target to invoke a external/libcxx/Makefile because to add libcxx objects to libc++.a library.

It is not going via external/Makefile, because external makefile creates libexternal.a and we dont want libc++ objects to be part of libexternal.a. In external/Makefile, it uses Make.defs from all the submodules and builds and adds it to libexternal.a.

samirsahu1947:
@manoharahk Why llvm libc++ objects cannot be a part of libexternal.a ?

manoharahk:
Dear @samirsahu1947 ,

Because, llvm libc++ is standard library and it is good to build a separate static or shared(not for TizenRT) instead of putting its objects in other non standard libraries. It also doesn't make sense to link libexternal.a where we require a c++ library. Also similar to libc.a, it is better to have libcxx.a(which is already building as part of core c++ source, lib/libxx and Add all llvm libc++ objects (from external) to this libcxx.a  . It is also good for measuring the size of the c++ library independently

sunghan-chang:
@manoharahk Ok. I agree your opinion.

manoharahk:
Thanks @sunghan-chang, I will push the updated patch set for this approach. Please review.

seinfra:
Code Rule Check OK.

manoharahk:
Okay. artik053/grpc and qemu/tc_16m_grpc are recently added. 
I will modify the patch and push it.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1278, opened 2018-02-19
Edit fix for CVE issues in wpa_supplicant

dr-venkman:
Edit a fix for CVE issues (commit id: 23fa1c75c26aa0fb400c7e2df6345e63a4cc2825). In this pull request, we perform the following edits to the above fix.

1. Rename the Kconfig parameter, allowing a default value of n.
2. Fix correct usage of the parameter to disable build of the affected files.
3. Document the CVE issues at affected code, and suggest the developer to fix the CVE issues before using the code.

### DIFF
--- a/external/wpa_supplicant/Kconfig
+++ b/external/wpa_supplicant/Kconfig
@@ -73,9 +73,12 @@ config DRIVER_T20
 	default y
 	depends on SCSC_WLAN
 
-config DISABLE_EAP_FOR_SUPPLICANT
-	bool "Disable EAP for WPA supplicant"
-	default y
+config ENABLE_EAP_FOR_SUPPLICANT
+	bool
+	default n
+	---help---
+		Enable EAP support for the supplicant. This is disabled by default,
+		because of pending CVE issues in some affected files.
 endif
 
 endmenu # wpa_supplicant
--- a/external/wpa_supplicant/src/eap_peer/eap_pwd.c
+++ b/external/wpa_supplicant/src/eap_peer/eap_pwd.c
@@ -4,8 +4,16 @@
  *
  * This software may be distributed under the terms of the BSD license.
  * See README for more details.
+ *
+ * NOTE: This file has a potential CVE issue (CVE-2015-5315 and CVE-2015-5316)
+ * at lines 673 (eap_pwd_perform_confirm_exchange function), 790 and 813 (both under
+ * eap_pwd_process function). However, we have disabled this file for build, by
+ * setting a default value n for CONFIG_ENABLE_EAP_FOR_SUPPLICANT.
+ *
+ * If you choose to build this file, you are advised to set CONFIG_ENABLE_EAP_FOR_SUPPLICANT=y
+ * in your config, and also fix the CVE issues described above.
  */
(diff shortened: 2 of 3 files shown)


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

external/wpa_supplicant/src/eap_peer/eap_pwd.c
```diff
@@ -4,8 +4,16 @@
...
+ * NOTE: This file has a potential CVE issue (CVE-2015-5315 and CVE-2015-5316)
+ * at lines 673 (eap_pwd_perform_confirm_exchange function), 790 and 813 (both under
+ * eap_pwd_process function). However, we have disabled this file for build, by
+ * setting a default value n for CONFIG_ENABLE_EAP_FOR_SUPPLICANT.
+ *
+ * If you choose to build this file, you are advised to set CONFIG_ENABLE_EAP_FOR_SUPPLICANT=y
+ * in your config, and also fix the CVE issues described above.
  */
-#ifndef CONFIG_DISABLE_EAP_FOR_SUPPLICANT
+#if CONFIG_ENABLE_EAP_FOR_SUPPLICANT
```
sunghan-chang: This is wrong. If you want to use ```#if```, not ```#ifdef```, when we don't want to use, it should be #define CONFIG_XX 0.
But in our system, when we don't want to use, it is not defined. So this will be ```#if ```. This is a wrong usage.
```#ifdef``` is correct.

File: external/wpa_supplicant/src/eap_server/eap_server_pwd.c [hunk 2]
```diff
@@ -4,8 +4,17 @@
...
+ * at lines 852 and 869 (eap_pwd_process function). However, we have
+ * disabled this file for build, by setting a
+ * default value n for CONFIG_ENABLE_EAP_FOR_SUPPLICANT.
+ *
+ * If you choose to build this file, you are advised to
+ * set CONFIG_ENABLE_EAP_FOR_SUPPLICANT=y in your config,
+ * and also fix the CVE issues described above.
  */
-#ifndef CONFIG_DISABLE_EAP_FOR_SUPPLICANT
+#if CONFIG_ENABLE_EAP_FOR_SUPPLICANT
```
sunghan-chang: same as previous.

external/wpa_supplicant/Kconfig
```diff
@@ -73,9 +73,9 @@ config DRIVER_T20
 	default y
 	depends on SCSC_WLAN
 
-config DISABLE_EAP_FOR_SUPPLICANT
-	bool "Disable EAP for WPA supplicant"
-	default y
+config ENABLE_EAP_FOR_SUPPLICANT
+	bool
+	default n
```
sunghan-chang: Many users (developers) do NOT see source codes. To control Kconfig, they will see only kconfig file or menuconfig. Why don't you add help sentences to describe this like source codes?

seinfra:
Code Rule Check OK.

sunghan-chang:
@pillip8282 Could you check this and merge?

PR #1284, opened 2018-02-20
 libc/spwan: modify checking logic of function arguments

sunghan-chang:
Nuttx checks them with DEBUGASSERT. But that is different with POSIX
standard and we don't want to make a crash even if it has invalid arguments.
This commit changes them returning errno values without crash.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1288, opened 2018-02-22
testcase/wifi-manager: fix callback function definition

pillip8282:
As definition of wifi-manager's callback is changed, utc need to be changed in UTC.

### DIFF
--- a/apps/examples/testcase/ta_tc/wifi_manager/utc/utc_wifi_manager_main.c
+++ b/apps/examples/testcase/ta_tc/wifi_manager/utc/utc_wifi_manager_main.c
@@ -40,7 +40,7 @@ do {										\
 	pthread_cond_wait(&g_wifi_manager_test_cond, &g_wifi_manager_test_mutex);	\
 	pthread_mutex_unlock(&g_wifi_manager_test_mutex);			\
 } while (0)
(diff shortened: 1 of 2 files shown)


apps/examples/testcase/ta_tc/wifi_manager/utc/utc_wifi_manager_main.c
```diff
@@ -40,7 +40,7 @@ do {										\
 	pthread_cond_wait(&g_wifi_manager_test_cond, &g_wifi_manager_test_mutex);	\
 	pthread_mutex_unlock(&g_wifi_manager_test_mutex);			\
 } while (0)
-void wifi_sta_connected(void);		// in station mode, connected to ap
+void wifi_sta_connected(wifi_manager_result_e);		// in station mode, connected to ap
```
sunghan-chang: [M11] In function prototypes, include parameter names with their data types and return type.

[Example-Do]
```
extern int function(struct type *ex);
```
[Example-Don't]
```
extern int function(struct type *);
```

pillip8282: Ok, I will fix

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1290, opened 2018-02-22
framework/st_things: Add rootCA in util for FOTA

KimDongEon:
RootCA is missing from util on FOTA. add.


seinfra:
Code Rule Check OK.

PR #1292, opened 2018-02-22
kernel/timer: add a logic to check timer is valid or not

sunghan-chang:
When timer is created or released, they update pt_flags
inside posix_timer_s struct. And set and get timer functions check
that flag and return appropriatary value.


seinfra:
Code Rule Check OK.

PR #1294, opened 2018-02-23
Add CXXOBJS recipe in frameworks

junmin-kim:
frameworks .cxx files can be compiled

### DIFF
--- a/framework/Makefile
+++ b/framework/Makefile
@@ -24,6 +24,7 @@ DELIM ?= $(strip /)
 DEPPATH = --dep-path .
 ASRCS =
 CSRCS =
+CXXSRCS =
 VPATH = .
 
 # Framework Directories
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

framework/Makefile
```diff
@@ -40,9 +41,10 @@ $(foreach BDIR, $(BUILDIRS), $(eval $(call Add_Framework,$(BDIR))))
 
 AOBJS = $(ASRCS:.S=$(OBJEXT))
 COBJS = $(CSRCS:.c=$(OBJEXT))
+CXXOBJS = $(CXXSRCS:.cxx=$(OBJEXT))
```
sunghan-chang: Why cxx? not cpp? What is common name for c++?

junmin-kim: We can support both cxx and cpp by using
`CXXOBJS = $(patsubst %.cxx,%$(OBJEXT),$(patsubst %.cpp,%$(OBJEXT),$(CXXSRCS)))`,
but what if CXXSRC have same name and different suffix?
ex) a.cpp a.cxx

seinfra:
Code Rule Check OK.

PR #1301, opened 2018-02-27
apps/shell: add a link to shell command list

sunghan-chang:
Make a link to shell command list document from shell readme.
Comands list is a good topic to show what there is in shell description.


seinfra:
Code Rule Check OK.

PR #1303, opened 2018-02-28
arm : fix stack memory base alignment

manoharahk:
Currently stack base is aligned to the stack size requested which causes more
fragmentation.

stack base is not required to be aligned, only stack top needs to be
aligned to max 8 bytes in case of armv7 which is already done.

Hence this patch removes aligning stack base during stack allocation from heap


seinfra:
Code Rule Check OK.

PR #1307, opened 2018-03-05
external/iotjs: get build options, target-arch and board from Kconfig

sunghan-chang:
When we try to build IoTjs, we should set --target-arch and --target-board
options on Makefile. They are fixed to arm and artik05x but they cause
difficult to find and wrong operation on IoTjs.
When TizenRT is built, CONFIG_ARCH and CONFIG_ARCH_BOARD are set and
they shows the information of selected board. User doesn't need to
set them in iotjs again.
This commit makes iotjs using those configurations.

### DIFF
--- a/external/Makefile
+++ b/external/Makefile
@@ -180,8 +180,8 @@ endif
 
 iotjs_build:
 ifeq ($(CONFIG_ENABLE_IOTJS),y)
-	python $(IOTJS_ROOT_DIR)/tools/build.py --target-arch=arm --target-os=tizenrt \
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@haesik @hs0225 @rzr Could you check this P.R.?

seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

haesik:
Currently, IoT.js doestn't support any other target, but I think it's reasonable to change it for future use.

rzr:
+1 , I have verified the change on artik055s and it works as expected,
once merged I'll might submit a PR to add profile to this rule (to enable i2c)...

sunghan-chang:
@rzr Thank you very much for verification.

PR #1308, opened 2018-03-06
framwork/st_things: apply string type on FIRMWARE_PROPERTY_UPDATE

YiHyunJin:
…tr type.

Modify int type to str type.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1309, opened 2018-03-06
add commands list for file system functionalities on README

sunghan-chang:
Add descriptions of cat, cd, df, ls, mkdir, mkrd, mksmartfs, rm, rmdir, pwd, mount, umount commands


seinfra:
Code Rule Check OK.

PR #1310, opened 2018-03-06
iotivity: Add support for ARTIK055S to iotivity_simpleserver

rzr:
Change-Id: I292ae9c3f1fc846e76e860e3dc04637d8a0c0885

### DIFF
--- a/apps/examples/iotivity_simpleserver/simpleServer.c
+++ b/apps/examples/iotivity_simpleserver/simpleServer.c
@@ -74,21 +74,6 @@
  * Definitions
  ****************************************************************************/
 
-#define IFHWADDRLEN 6
-
-#if LWIP_HAVE_LOOPIF == 1
-#define NET_DEVNUM 1
-#else
-#define NET_DEVNUM 0
-#endif
-
-#if defined(CONFIG_ENC28J60) || defined(CONFIG_WICED)
-#define NET_DEVNAME "en"
-#elif defined(CONFIG_ARCH_BOARD_SIDK_S5JT200) || defined(CONFIG_ARCH_BOARD_ARTIK053)
-#define NET_DEVNAME "wl"
-#else
-#error "undefined CONFIG_NET_<type>, check your .config"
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

apps/examples/iotivity_simpleserver/simpleServer.c
```diff
@@ -84,7 +84,7 @@
 
 #if defined(CONFIG_ENC28J60) || defined(CONFIG_WICED)
 #define NET_DEVNAME "en"
-#elif defined(CONFIG_ARCH_BOARD_SIDK_S5JT200) || defined(CONFIG_ARCH_BOARD_ARTIK053)
+#elif defined(CONFIG_ARCH_BOARD_SIDK_S5JT200) || defined(CONFIG_ARCH_BOARD_ARTIK053) || defined(CONFIG_ARCH_BOARD_ARTIK055S)
```
sunghan-chang: I think ```defined(CONFIG_ARCH_BOARD_ARTIK05X_FAMILY)``` is better than ```defined(CONFIG_ARCH_BOARD_ARTIK053) || defined(CONFIG_ARCH_BOARD_ARTIK055S)```.
Anyway I don't know why it has board-specific conditional instead of function conditional. It looks not good.

rzr: I can set CONFIG_ARCH_BOARD_ARTIK05X_FAMILY but I can only test on 55s I guess 53 worked before and we can assume others will.

I plan to make a cleanup patch to introduce CONFIG_NET_DEVNAME that will be set by kconfig to "en" or "wl", would that be ok ?

sunghan-chang: 055s and 053s are same on iotitivity so that family will work well.
CONFIG_NET_DEVNAME is better I think.

seinfra:
Code Rule Check OK.

PR #1311, opened 2018-03-08
Move media Kconfig to framework/src/media

junmin-kim:
Move os/media/Kconfig -> framework/src/media/Kconfig
Kconfig should be located with the relevant code


seinfra:
Code Rule Check OK.

PR #1315, opened 2018-03-09
Fetch IP Address for WiFi Station

bhargava-cs:
- Once device is connected in STA mode start DHCP client procedure

### DIFF
--- a/apps/examples/slsiwifi/slsiwifi_main.c
+++ b/apps/examples/slsiwifi/slsiwifi_main.c
@@ -28,6 +28,8 @@
 #include <math.h>
 #include <slsi_wifi/slsi_wifi_utils.h>
 #include "output_functions.h"
+#include <protocols/dhcpc.h>
+
 // the mm functions are badly guarded with CONFIG_DEBUG
 #ifndef CONFIG_DEBUG
 #undef CONFIG_EXAMPLES_SLSIDEMO_MEM_CHECK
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

bhargava-cs:
juitem

bhargava-cs:
Hi Pillip,

Thanks for the review. I agree with you. 
In addition to that, we may keep a single application for WiFi testing i.e wm_test.

However, this patch is required  to fix a build error.
https://github.com/Samsung/TizenRT/pull/1315/commits/8126fca0dd94a13969ea29b554d26621772501c3

Please suggest whether to raise separate PR for this?

pillip8282:
Hi Bhargava
Seperating PR would be good.

apps/examples/slsiwifi/slsiwifi_main.c
```diff
@@ -164,6 +170,23 @@ void sw_linkUpHandler(slsi_reason_t *reason)
...
+			return;
+		}
+		if (dhcpc_request(dhcp_handle, &state) != OK) {
+			dhcpc_close(dhcp_handle);
+			return;
+		}
+		netlib_set_ipv4addr(WLAN_INTF_NAME, &state.ipaddr);
+		netlib_set_ipv4netmask(WLAN_INTF_NAME, &state.netmask);
+		netlib_set_dripv4addr(WLAN_INTF_NAME, &state.default_router);
 	} else {
```
girishks2000: This setting is static for ipv4. Restricting it to only ipv4 is not good idea. Can the test case be made compatible to work for ipv4 and v6 both. You may have to provide another wrapper (or may be already exist)

bhargava-cs:
WM_TEST App gets IP address automatically. This change is not needed.

PR #1317, opened 2018-03-13
[AUDIO] Fix audio UTC failure issues

kishore-sn:
Below issues fixed:
Missing underrun handling code in some UTC's
Fix I2S stop / pause / resume API's which was causing failure in MMAP related audio UTC's

### DIFF
--- a/apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
+++ b/apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
@@ -48,6 +48,8 @@
 #define AUDIO_DEFAULT_FORMAT PCM_FORMAT_S16_LE
 
 #define AUDIO_RECORD_DURATION 3	//3sec
+
+#define PREPARE_RETRY_COUNT 10
 /****************************************************************************
  * Global Variables
  ****************************************************************************/
@@ -610,6 +612,7 @@ static void utc_audio_pcm_readi_p(void)
 	unsigned int bytes_per_frame;
 	int frames_read;
 	int remain;
+	int retry = PREPARE_RETRY_COUNT;
 
 	buffer = malloc(pcm_frames_to_bytes(g_pcm, pcm_get_buffer_size(g_pcm)));
 	TC_ASSERT_NEQ_CLEANUP("pcm_readi", buffer, NULL, clean_all_data(0, NULL));
@@ -629,13 +632,20 @@ static void utc_audio_pcm_readi_p(void)
 	while (remain > 0) {
 		frames_read = pcm_readi(g_pcm, buffer, pcm_get_buffer_size(g_pcm));
 		if (frames_read == -EPIPE) {
-			pcm_prepare(g_pcm);
-			continue;
+			if (retry--) {
+				pcm_prepare(g_pcm);
+				continue;
+			} else {
+				printf("ERROR: read failed due to multiple xrun \n");
+				TC_ASSERT_CLEANUP("pcm_readi", 0, clean_all_data(fd, buffer));
+			}
 		}
 
 		if (frames_read < 0) {
 			break;
 		}
+
+		retry = PREPARE_RETRY_COUNT;
 		remain -= frames_read;
 		ret = write(fd, buffer, bytes_per_frame * frames_read);
 		TC_ASSERT_EQ_CLEANUP("pcm_readi", ret, (bytes_per_frame * frames_read), clean_all_data(fd, buffer));
@@ -707,6 +717,7 @@ static void utc_audio_pcm_drop_p(void)
 	int size;
 	int total_frames;
 	int num_read;
+	int retry = PREPARE_RETRY_COUNT;
 
 	g_pcm = pcm_open(0, 0, PCM_IN, NULL);
 	TC_ASSERT_GT("pcm_drop", pcm_is_ready(g_pcm), 0);
@@ -728,9 +739,20 @@ static void utc_audio_pcm_drop_p(void)
 
 	while (remain > 0) {
 		frames_read = pcm_readi(g_pcm, buffer, pcm_get_buffer_size(g_pcm));
+		if (frames_read == -EPIPE) {
+			if (retry--) {
+				pcm_prepare(g_pcm);
+				continue;
+			} else {
+				printf("ERROR: readi failed due to multiple xrun \n");
+				TC_ASSERT_CLEANUP("pcm_drop", 0, clean_all_data(fd, buffer));
+			}
+		}
 		if (frames_read < 0) {
 			break;
 		}
+
+		retry = PREPARE_RETRY_COUNT;
 		remain -= frames_read;
 		ret = write(fd, buffer, bytes_per_frame * frames_read);
 		TC_ASSERT_EQ_CLEANUP("pcm_drop", ret, (bytes_per_frame * frames_read), clean_all_data(fd, buffer));
@@ -756,11 +778,25 @@ static void utc_audio_pcm_drop_p(void)
 
 	printf("Check now for recorded part. Nothing after drop should be heard\n");
 	total_frames = 0;
+	retry = PREPARE_RETRY_COUNT;
 	do {
-		num_read = read(fd, buffer, size);
+		if (retry == PREPARE_RETRY_COUNT) {
+			num_read = read(fd, buffer, size);
+		}
 		if (num_read > 0) {
-			total_frames += num_read;
 			ret = pcm_writei(g_pcm, buffer, pcm_bytes_to_frames(g_pcm, num_read));
+			if (ret == -EPIPE) {
+				if (retry--) {
+					pcm_prepare(g_pcm);
+					continue;
+				} else {
+					printf("ERROR: writei failed due to multiple xrun \n");
+					TC_ASSERT_CLEANUP("pcm_drop", 0, clean_all_data(fd, buffer));
+				}
+			}
+
+			retry = PREPARE_RETRY_COUNT;
+			total_frames += num_read;
 			TC_ASSERT_GEQ_CLEANUP("pcm_drop", ret, 0, clean_all_data(fd, buffer));
 		}
 	} while (num_read > 0);
@@ -810,6 +846,7 @@ static void utc_audio_pcm_writei_p(void)
 	char *buffer;
 	int num_read;
 	unsigned int size;
+	int retry = PREPARE_RETRY_COUNT;
 
 	/* use default config here */
 	g_pcm = pcm_open(0, 0, PCM_OUT, NULL);
@@ -825,15 +862,25 @@ static void utc_audio_pcm_writei_p(void)
 	printf("Playback start!!\n");
 
 	for (;;) {
-		num_read = read(fd, buffer, size);
+		if (retry == PREPARE_RETRY_COUNT) {
+			num_read = read(fd, buffer, size);
+		}
+
 		if (num_read <= 0) {
 			break;
 		}
 		ret = pcm_writei(g_pcm, buffer, pcm_bytes_to_frames(g_pcm, num_read));
 		if (ret == -EPIPE) {
-			pcm_prepare(g_pcm);
-			ret = pcm_writei(g_pcm, buffer, pcm_bytes_to_frames(g_pcm, num_read));
+			if (retry--) {
+				pcm_prepare(g_pcm);
+				continue;
+			} else {
+				printf("ERROR: writei failed due to multiple xrun \n");
+				TC_ASSERT_CLEANUP("pcm_writei", 0, clean_all_data(fd, buffer));
+			}
 		}
+
+		retry = PREPARE_RETRY_COUNT;
 		TC_ASSERT_GEQ_CLEANUP("pcm_writei", ret, 0, clean_all_data(fd, buffer));
 	}
 
@@ -861,7 +908,7 @@ static void utc_audio_pcm_writei_n(void)
 	ssize_t size;
 	char *buffer;
 
-	g_pcm = pcm_open(0, 0, PCM_IN, NULL);
+	g_pcm = pcm_open(0, 0, PCM_OUT, NULL);
 	TC_ASSERT_GT("pcm_writei", pcm_get_file_descriptor(g_pcm), 0);
 
 	size = pcm_frames_to_bytes(g_pcm, pcm_get_buffer_size(g_pcm));
@@ -979,7 +1026,7 @@ static void utc_audio_pcm_avail_update_n(void)
 static void utc_audio_pcm_avail_update_p(void)
 {
 	int ret;
-	unsigned int retry = 20;
+	unsigned int retry = PREPARE_RETRY_COUNT;
 
 	g_pcm = pcm_open(0, 0, PCM_IN | PCM_MMAP, NULL);
 	TC_ASSERT_GT("pcm_avail_update", pcm_is_ready(g_pcm), 0);
@@ -1120,7 +1167,7 @@ static void utc_audio_pcm_begin_p(void)
 	char *areas = NULL;
 	unsigned int offset;
 	unsigned int frames;
-	unsigned int retry = 20;
+	unsigned int retry = PREPARE_RETRY_COUNT;
 
 	g_pcm = pcm_open(0, 0, PCM_IN | PCM_MMAP, NULL);
 	TC_ASSERT_GT("pcm_mmap_begin",  pcm_is_ready(g_pcm), 0);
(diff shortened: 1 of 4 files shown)


seinfra:
Code Rule Check Result:
os/arch/arm/src/s5j/s5j_i2s.c:1719: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/arch/arm/src/s5j/s5j_i2s.c:1719: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/arch/arm/src/s5j/s5j_i2s.c:1721: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/arch/arm/src/s5j/s5j_i2s.c:1721: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/arch/arm/src/s5j/s5j_i2s.c:1722: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/arch/arm/src/s5j/s5j_i2s.c:1722: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/arch/arm/src/s5j/s5j_i2s.c:1723: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/arch/arm/src/s5j/s5j_i2s.c:1723: ERROR: [IDT_M_TAB] please, no spaces at the start of a line

seinfra:
Code Rule Check OK.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1600,18 +1600,20 @@ static int i2s_pause(struct i2s_dev_s *dev)
 {
 	struct s5j_i2s_s *priv = (struct s5j_i2s_s *)dev;
 	irqstate_t flags;
+	volatile u32 reg = 0;
 	DEBUGASSERT(priv);
 
+	reg = getreg32(priv->base + S5J_I2S_CON);
 #ifdef I2S_HAVE_TX_P
-	if (priv->txpenab) {
+	if (priv->txpenab && reg & I2S_CR_TXDMACTIVE) {
```
btheosam: Why this check of DMA active status? What happens if pause bit is set when dma is inactive? Any harm? Better you check the status within irqsave.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1600,18 +1600,20 @@ static int i2s_pause(struct i2s_dev_s *dev)
...
 		flags = irqsave();
 		modifyreg32(priv->base + S5J_I2S_CON, 0, I2S_CR_TXCHPAUSE);
 		irqrestore(flags);
 	}
 #endif
 
 #ifdef I2S_HAVE_RX
-	if (priv->rxenab) {
+	if (priv->rxenab && reg & I2S_CR_RXDMACTIVE) {
```
btheosam: Same as prev

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1650,9 +1654,13 @@ static int i2s_stop(struct i2s_dev_s *dev)
 	struct s5j_i2s_s *priv = (struct s5j_i2s_s *)dev;
 	irqstate_t flags;
 	struct s5j_buffer_s *bfcontainer;
+	volatile u32 reg;
 	DEBUGASSERT(priv);
 
 #ifdef I2S_HAVE_TX_P
+	modifyreg32(priv->base + S5J_I2S_CON, I2S_CR_TXDMACTIVE, 0);
```
btheosam: Why is this modified without disabling interrupts?

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1681,6 +1687,9 @@ static int i2s_stop(struct i2s_dev_s *dev)
 #endif
 
 #ifdef I2S_HAVE_RX
+	modifyreg32(priv->base + S5J_I2S_CON, I2S_CR_RXDMACTIVE, 0);
```
btheosam: same as above

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: We need to pause DMA channel, only if it is currently active.
Not doing this was resulting in UTC getting blocked in pcm_wait() which waits for a dequeue msg.

I modified to use irqsave and perform the entire pause / resume as one single atomic operation.

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: Same as above

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: Changed

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: Changed.

seinfra:
Code Rule Check OK.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1600,47 +1600,47 @@ static int i2s_pause(struct i2s_dev_s *dev)
...
 #ifdef I2S_HAVE_RX
-	if (priv->rxenab) {
-		flags = irqsave();
+	if (priv->rxenab && reg & I2S_CR_RXDMACTIVE) {
 		modifyreg32(priv->base + S5J_I2S_CON, 0, I2S_CR_RXCHPAUSE);
-		irqrestore(flags);
 	}
 #endif
+	irqrestore(flags);
 	return 0;
```
girishks2000: you don't have to explicitly save and restore the irq state when you use  modifyreg32. It internally does it. take care of this in all places

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1650,9 +1650,14 @@ static int i2s_stop(struct i2s_dev_s *dev)
 	struct s5j_i2s_s *priv = (struct s5j_i2s_s *)dev;
 	irqstate_t flags;
 	struct s5j_buffer_s *bfcontainer;
+	volatile u32 reg;
 	DEBUGASSERT(priv);
```
girishks2000: added this but I cant see the usage of this anywhere

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -728,6 +728,10 @@ static void utc_audio_pcm_drop_p(void)
 
 	while (remain > 0) {
 		frames_read = pcm_readi(g_pcm, buffer, pcm_get_buffer_size(g_pcm));
+		if (frames_read == -EPIPE) {
+			pcm_prepare(g_pcm);
+			continue;
+		}
 		if (frames_read < 0) {
```
girishks2000: Are you sure  it will succeed in reading / writing the second time.  There should be a way to check prepare status and try again.
Suggestion:
Retry mechanism has to be modified  based on some timeout mechanism. Or any better idea if you have

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -861,7 +869,7 @@ static void utc_audio_pcm_writei_n(void)
 	ssize_t size;
 	char *buffer;
 
-	g_pcm = pcm_open(0, 0, PCM_IN, NULL);
+	g_pcm = pcm_open(0, 0, PCM_OUT, NULL);
 	TC_ASSERT_GT("pcm_writei", pcm_get_file_descriptor(g_pcm), 0);
```
girishks2000: Why direction changed? just curious. Was it a mistake earlier?

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1600,47 +1600,47 @@ static int i2s_pause(struct i2s_dev_s *dev)
...
+	volatile u32 reg = 0;
 	DEBUGASSERT(priv);
 
+	reg = getreg32(priv->base + S5J_I2S_CON);
+	flags = irqsave();
 #ifdef I2S_HAVE_TX_P
-	if (priv->txpenab) {
-		flags = irqsave();
+	if (priv->txpenab && reg & I2S_CR_TXDMACTIVE) {
 		modifyreg32(priv->base + S5J_I2S_CON, 0, I2S_CR_TXCHPAUSE);
```
girishks2000: separate conditions with brackets. Logical and conditional expression evaluation will be easy to understand

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: It was a mistake earlier. Fixed it now.

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: It is used in line num 1717.

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: ok done

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: I want both rx and tx channels to pause / resume in one shot. 
So, I think  its better to disable irq explicitly.. Please let me know your opinion.

girishks2000: But both features are mutually exclusive. In a single codec either one will be active at any given point of time. So oneshot enable disable doesn't make sense.

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Fixed. 
Added retry mechanism

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: Yes true. But we had not planned for it. 
In that case, we will have to modify all our API's to handle tx and rx separately. 
I will do the changes and publish them.

kishore-sn: Fixed.

kishore-sn:
I modified I2S API's to accept channel information and operate only on that channel. 
Please review again.

seinfra:
Code Rule Check Result:
os/arch/arm/src/s5j/s5j_i2s.c:1683: ERROR: [BRC_M_SMT] that open brace { should be on the previous line

seinfra:
Code Rule Check OK.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1596,117 +1596,128 @@ static uint32_t i2s_txdatawidth(struct i2s_dev_s *dev, int bits)
...
-	s5j_dmastop(priv->txp.dma);
+	if (dir == I2S_TX) {
+		modifyreg32(priv->base + S5J_I2S_CON, I2S_CR_FTXURINTEN, 0);
+		modifyreg32(priv->base + S5J_I2S_CON, I2S_CR_TXDMACTIVE, 0);
+		s5j_dmastop(priv->txp.dma);
+
+		while (sq_peek(&priv->txp.pend) != NULL) {
+			flags = irqsave();
+			bfcontainer = (struct s5j_buffer_s *)sq_remfirst(&priv->txp.pend);
+			irqrestore(flags);
```
girishks2000: here also saving interrupt context is not required. You are working on a local copy of the data structure, so modification of when interrupt occurs should not affect. Same applies wherever local structure is modified

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
```diff
@@ -810,6 +844,7 @@ static void utc_audio_pcm_writei_p(void)
 	char *buffer;
 	int num_read;
 	unsigned int size;
+	int retry = 10;
```
Taejun-Kwon: #define PREPARE_RETRY_COUNT 10(Actually 10 is too large, I think )

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: As discussed, the queue is being updated in the irq context also. So we have to disable irq to protect the queue from corruption.

Taejun-Kwon:
@GalkinIvan , could you please check s5j code here?

apps/examples/testcase/ta_tc/audio/utc/utc_audio_main.c
kishore-sn: Done. 
I agree that retry count of 10 is big. But I think its ok for UTC. 
We can have a smaller count in media player.

seinfra:
Code Rule Check OK.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -1596,117 +1596,128 @@ static uint32_t i2s_txdatawidth(struct i2s_dev_s *dev, int bits)
...
-	while (sq_peek(&priv->rx.done) != NULL) {
-		flags = irqsave();
-		bfcontainer = (struct s5j_buffer_s *)sq_remfirst(&priv->rx.done);
-		irqrestore(flags);
-		apb_free(bfcontainer->apb);
-		i2s_buf_rx_free(priv, bfcontainer);
+
+	reg = getreg32(priv->base + S5J_I2S_CON);
+
+	if ((reg & (I2S_CR_TXDMACTIVE | I2S_CR_RXDMACTIVE)) == 0) {
```
btheosam: This condition would disable I2S port when either of RX/TX is NOT active. But I see you want to handle RX & TX seperately. Confirm if this condition is okay.

btheosam:
Not sure if this changes for I2S to add operations on RX/TX alone like i2s_stop(dev, RX ).
Coz, if ARTIK is the master, then to stop either RX/TX, the clocks had to be cut, which inturn would stop both RX & tX. Right now, alc5658 is the master, so stopping RX & TX at artik end might work, as the clocks still on from ALC end. Need a second opinion on this.  Unless these changes are generic and workable on both master & slave mode, this changes should be deferred.

os/arch/arm/src/s5j/s5j_i2s.c
```diff
@@ -297,9 +297,9 @@ static uint32_t i2s_samplerate(struct i2s_dev_s *dev, uint32_t rate);
 
 static uint32_t i2s_txdatawidth(struct i2s_dev_s *dev, int bits);
 static int i2s_send(struct i2s_dev_s *dev, struct ap_buffer_s *apb, i2s_callback_t callback, void *arg, uint32_t timeout);
-static int i2s_stop(struct i2s_dev_s *dev);
-static int i2s_pause(struct i2s_dev_s *dev);
-static int i2s_resume(struct i2s_dev_s *dev);
+static int i2s_stop(struct i2s_dev_s *dev, i2s_ch_dir_t dir);
```
btheosam: Would it be possibel to stop only RX or TX if HOST is master?

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: It disables I2S when both RX and TX are not active.

os/arch/arm/src/s5j/s5j_i2s.c
kishore-sn: Similar logic is applied in IRQ handler. 
So, I suppose this must work. 

Also, if stopping either RX or TX would completely stop the I2S clock, then why do we have separate register setting to stop the complete I2S core?

GalkinIvan:
Changes of os/arch/arm/src/s5j/s5j_i2s.c look fine.

os/arch/arm/src/s5j/s5j_i2s.c
btheosam: @kishore-sn Is it common for all chipset? (My knowledge is limited on other chipsets)
If you are sure that these changes are portable on Cypress board also, then I can dismiss the hold. 
But, if you are NOT sure, then better to wait for Ivan's agreement.

btheosam: @kishore-sn please dont scope only for ARTIK. Please check if this will work on others as well.

kishore-sn: I cannot be sure about how I2S works in all SOC's. In my opinion, there has to be a way to individually control the RX and TX channels. I have checked the I2S implementation in bcm43909 and also in an old android project based on Samsung SOC. They seem to have similar type of implementation. For example, checkout the below API's from bcm code:

static int i2s_stop_stream( platform_i2s_port_t port, platform_i2s_direction_t dir );

So, I suggest that we go ahead with this implementation for now. When we work on a new SOC, we would either need to find some way of implementing the similar feature or add a work around in our code. 

@Taejun-Kwon If you have any suggestions, please let me know.

Taejun-Kwon: I checked many i2s codes and I found seperate rx & tx with direction is common style.(I jsut checked android & zephyr's some h/w)
But I think benix's opinion also very important.
Let's discuss about it more. I wil invite you guys.

PR #1321, opened 2018-03-14
kdbg/heapinfo : Add missing pid assignment logic for pthread_create

jeongchanKim:
And Add tcb checking logic which tcb is alive or not.


seinfra:
Code Rule Check OK.

PR #1325, opened 2018-03-14
example/grpc : add async client examples in route_client

jeongarmy:
These are examples of asynchonous RPC calls in gPRC.


seinfra:
Code Rule Check OK.

apps/examples/grpc_route_client/route_guide_client.cc
```diff
@@ -252,40 +270,292 @@ class RouteGuideClient {
...
+      ret = cq.Next(&got_tag, &ok);
+      if (!ret || !ok || got_tag != (void*)1) {
+        std::cout << "Write Failed" << std::endl;
+        break;
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(
+        delay_distribution(generator)));
+    }
+
+    writer->WritesDone(got_tag);
```
dr-venkman: Is WritesDone necessary to be called? Doxygen says that Finish method can implicitly verify that client has sent all messages. Could you please verify?

apps/examples/grpc_route_client/route_guide_client.cc
```diff
@@ -252,40 +270,292 @@ class RouteGuideClient {
...
+          if (!thd_ret || !thd_ok || got_tag_write != (void*)1) {
+            std::cout << "Write Failed" << std::endl;
+            break;
+          }
+        }
+        stream->WritesDone(got_tag_write);
+        thd_ret = cq.Next(&got_tag_write, &thd_ok);
+        if (!thd_ret || !thd_ok || got_tag_write != (void*)1) {
+          std::cout << "WriteDone Failed" << std::endl;
+          return;
```
dr-venkman: maybe not needed, as the thread was exiting anyway. Please ignore comment if not the case.

apps/examples/grpc_route_client/route_guide_client.cc
jeongarmy: yes, WritesDone is nessasary for verify to finish Writes on this stream.
and then it is actually done when its CQ tag comes back.

dr-venkman: Thank you for verifying. Yes, WritesDoneis necessary. Please ignore the comment above.

seinfra:
Code Rule Check OK.

PR #1326, opened 2018-03-14
framework/st_things : Memory allocation and free modification

JeongJunSik:
- Pair alignment (things_malloc / things_calloc / things_free)


seinfra:
Code Rule Check OK.

juitem:
This is right direction. OIC _xxx codes and things_xxx codes are different layer.
It will help for estimating per module memory usage.

PR #1327, opened 2018-03-14
sem: add a flag to show initialization and check it on sem APIs

sunghan-chang:
1. POSIX standard describes return value as shown below:
Upon successful completion, a value of zero shall be returned.
Otherwise, a value of -1 shall be returned and errno set to indicate the error.
  But our sem APIs do not check sem initialization and do not return ERROR.
2. Without sem_init, almost sem APIs work. But it can cause abnormal
  operation.

The sem_s structure already has the flag element inside
and first bit is an inheritance flag. Let's use second bit as
initialization flag and check it at first operation of sem APIs.

### DIFF
--- a/lib/libc/semaphore/sem_getprotocol.c
+++ b/lib/libc/semaphore/sem_getprotocol.c
@@ -57,7 +57,7 @@
 #include <tinyara/config.h>
 
 #include <sys/types.h>
-#include <assert.h>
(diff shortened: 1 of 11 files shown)


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1332, opened 2018-03-16
network/socket_api: remove dependency between iotivity and socket apis

pillip8282:
socket apis(recvmsg, sendmsg) have a dependency to iotivity.
recvmsg and sendmsg will be built with other socket apis
but those apis are not fully supported. so it should be used carefully.


seinfra:
Code Rule Check OK.

PR #1335, opened 2018-03-16
Fix bug in framework/Makefile

junmin-kim:
CXXOBJS needed both .cxx and .cpp, and it is a bug
Divide CXXOBJS by CXXOBJS and CPPOBJS according to suffix


seinfra:
Code Rule Check OK.

PR #1345, opened 2018-03-20
kernel/init: call fs_initialize before clock_initialize to initialize semaphore for inode

jeongarmy:
clock_initialize uses the semahpore for inode, g_inode_sem before initialization.
A semaphore shouldn't be used without initialization.
So we need to call fs_initilize function which initializes inode semaphore before that.


seinfra:
Code Rule Check OK.

PR #1346, opened 2018-03-20
network/dhcpd: fix calling dhcpd join callback even it points invalid…

pillip8282:
… memory.

A callback that notifies providing an address to node should be protected by null check.

### DIFF
--- a/external/dhcpd/dhcpd.c
+++ b/external/dhcpd/dhcpd.c
@@ -100,11 +100,13 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <semaphore.h>
+#include <pthread.h>
 
 /****************************************************************************
  * Global Data
  ****************************************************************************/
 sem_t g_dhcpd_sem;
+
 
 /****************************************************************************
  * Private Data
@@ -321,7 +323,8 @@ static char DHCPD_IFNAME[IFNAMSIZ] = { 0, };
 static struct timeval g_select_timeout = { 1, 0 };
 #endif
 
-static dhcp_sta_joined g_dhcp_sta_joined;
+static dhcp_sta_joined g_dhcp_sta_joined = NULL;
+static pthread_mutex_t g_dhcpd_lock = PTHREAD_MUTEX_INITIALIZER; // protect a write operation in g_dhcpd_running
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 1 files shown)


seinfra:
Code Rule Check OK.

external/dhcpd/dhcpd.c
```diff
@@ -1637,7 +1639,7 @@ int dhcpd_run(void *arg)
 
 exit_with_error:
 	g_dhcpd_running = 0;
-
+    g_dhcp_sta_joined = 0;
```
sunghan-chang: 1. use a tab instead of spaces as indentation
2. Isn't it ```NULL```?

girishks2000: If callback initialization is handled in dhcpd_start function, you don't have to set it to NULL in run function. In start initialize by default to NULL, if the parameter is valid then it gets assigned below  or else it remains NULL.

pillip8282: Thanks, I'll fix it

pillip8282: Thanks, I will fix it

seinfra:
Code Rule Check OK.

external/dhcpd/dhcpd.c
```diff
@@ -321,7 +321,7 @@ static char DHCPD_IFNAME[IFNAMSIZ] = { 0, };
 static struct timeval g_select_timeout = { 1, 0 };
 #endif
 
-static dhcp_sta_joined g_dhcp_sta_joined;
+static dhcp_sta_joined g_dhcp_sta_joined = NULL;
```
girishks2000: Can keep this uninitialized and move initialize code 
g_dhcp_sta_joined = NULL;  to dhcpd_start function

pillip8282: Could you tell me what benefit can I get from that?

girishks2000: In current situation, if dhcpd_stop is called the above variable remains initialized to the callback even after stop. Next if  start is called with NULL then the above pointer still remains initialized to old call back. So if we move it to start function it would be good. Handling is done in one place, unlike  earlier in 2 places.

seinfra:
Code Rule Check OK.

pillip8282:
@girishks2000
Do you have any comments?

pillip8282:
@girishks2000 
As you said it's more beneficial. but If we don't have atomic operation now then it will take time to adapt it. So I think that merging this commit would be ok.
If you think so, approve it

girishks2000:
Reviewed and Acked

sunghan-chang:
@girishks2000 When you merge a commit, you should approve first and then merge it. SE team checks and counts number of merging without approval.

sunghan-chang:
@girishks2000 Thank you.

PR #1347, opened 2018-03-20
Modify protobuf makefile not to build two same files

HONGCHAEHEE:
- There are two common.cc files in Makefile, one should be removed.


seinfra:
Code Rule Check OK.

PR #1350, opened 2018-03-20
add dependency between configurations about grpc and protobuf

jeongarmy:
These examples can be executed with enabling related modules.
And protobuf should be enabled for gRPC usage.
So these have some dependency on each module.

### DIFF
--- a/apps/examples/protobuf/Kconfig
+++ b/apps/examples/protobuf/Kconfig
@@ -6,6 +6,7 @@
 config EXAMPLES_PROTOBUF
 	bool "Protocol Buffers example"
 	default n
+	depends on PROTOBUF
 	---help---
 		Enable the Protocol Buffers example
 
--- a/external/grpc/Kconfig.protocol
+++ b/external/grpc/Kconfig.protocol
@@ -6,7 +6,7 @@
 config GRPC
 	bool "gRPC"
 	default n
-	depends on NET
-	depends on HAVE_CXX
+	select PROTOBUF
+	depends on NET && LIBCXX
 	---help---
 		Enable support for the gRPC.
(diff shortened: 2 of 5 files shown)


seinfra:
Code Rule Check OK.

external/grpc/Kconfig.protocol
```diff
@@ -6,7 +6,6 @@
 config GRPC
 	bool "gRPC"
 	default n
-	depends on NET
-	depends on HAVE_CXX
+	depends on NET && HAVE_CXX && PROTOBUF
```
sunghan-chang: 1. Don't you need llvm? Is HAVE_CXX enough?
2. PROTOBUF is a simple config and doesn't have any dependancy. Let's use ```select PROTOBUF```.

apps/examples/protobuf/Kconfig
```diff
@@ -6,6 +6,7 @@
 config EXAMPLES_PROTOBUF
 	bool "Protocol Buffers example"
 	default n
+	depends on PROTOBUF
```
sunghan-chang: ```select PROTOBUF```

external/grpc/Kconfig.protocol
pillip8282: I'm confused that protobuf is a simple config. Is there any differences between HAVE_CXX and PROTOBUF. if there are differences, I think that we should guide the developers it first.

seinfra:
Code Rule Check OK.

external/grpc/Kconfig.protocol
```diff
@@ -6,7 +6,7 @@
 config GRPC
 	bool "gRPC"
 	default n
-	depends on NET
-	depends on HAVE_CXX
+	select PROTOBUF
+	depends on NET && HAVE_CXX && LIBCXX
```
sunghan-chang: Because LIBCXX is in HAVE_CXX conditional in Kconfig of libxx, you had better check only LIBCXX. Checking both is redundant.

seinfra:
Code Rule Check OK.

PR #1354, opened 2018-03-20
os/sem: fix missing value of SEM_INITIALIZER on commit 9d4e238

sunghan-chang:
The sem_s structure has the flags element always by commit 9d4e238.
But SEM_INITIALIZER misses adding that element.
This commit fixs commit 9d4e238 and 3b7eaca.


seinfra:
Code Rule Check OK.

PR #1357, opened 2018-03-21
external/grpc: refactoring transport security interface

an4967:
- Change process of handshake.
- Integrate server-side hs and client-side hs.
- Extract SSL context initialize function.
- Remove in_ssl_buffer (saved 16KB heap memory)


seinfra:
Code Rule Check OK.

sunghan-chang:
@an4967 You need to check [**our commit rule**](https://github.com/Samsung/TizenRT/wiki/Documentations#commit-rules)
1. You should fill commit description on each commit. There is no description on all of commit.
2. I can't know which module is changed for security from commit title.
  TLS itself? gRPC? or other module?
  I mean from commit title. I can know it from change content.

PR #1358, opened 2018-03-21
Refactoring on iotbus gpio

sunghan-chang:
1. framework/iotbus: remove wrong checking condition on iotbus_gpio_register_cb
2. examples/iotbus_test: add new configs to show gpio number and refactoring
3. configs/artik053,055s/minimal: update defconfig to disable iotbus


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

sunghan-chang:
@girishks2000 Thank you for reviewing. Could you merge this?

girishks2000:
I have a question regarding the GPIO numbers hardcoded in application level. Doesn't it make test cases specific to a chipset.  Is there any approach to read the GPIO number during runtime to keep test cases common across chipset. 
Testcases can be common and PIN numbers can be dynamically read, based on the "Help menu displayed for that chip"
It is just a suggestion. You can use better methods which can keep Tcs common across chipsets

seinfra:
Code Rule Check OK.

sunghan-chang:
1. I think common application across chipsets is no need to change source code to run it.
 Previously we should modify iotbus_test sample source code itself to run it on other board.
 That's why I make new configs. And it makes no need to check and modify source code.
2. It is difficult to give GPIO information. Because that can be huge information, different across chipsets and nobody knows which pin will be used.

Ideally I agree your suggestion. But I have no idea and think it difficult  to do it.

girishks2000:
OK @sunghan-chang, I will merge this. Will come with a approach later to consolidate the gpio app

sunghan-chang:
@girishks2000 Thank you for merging.

PR #1359, opened 2018-03-21
external/grpc: refactoring transport security interface

an4967:
- Change process of handshake.
- Integrate server-side hs and client-side hs.
- Extract SSL context initialize function.
- Remove in_ssl_buffer (saved 16KB heap memory)

### DIFF
--- a/external/grpc/Kconfig.protocol
+++ b/external/grpc/Kconfig.protocol
@@ -10,3 +10,19 @@ config GRPC
 	depends on HAVE_CXX
 	---help---
 		Enable support for the gRPC.
+
+if GRPC
+
+config TSI_LOG
+	bool "Enable TSI Layer Log"
+	default n
+	---help---
+		Show function calls process in TSI Layer
+
+config TSI_MBEDTLS_LOG
+	int "Set mbedTLS Log Level(0:NONE - 5:ALL)"
+	default 0
+	---help---
+		Set Log levels for mbedtls debugging log
+
+endif # GRPC
--- a/external/grpc/src/core/tsi/ssl_transport_security.cc
+++ b/external/grpc/src/core/tsi/ssl_transport_security.cc
@@ -71,19 +71,22 @@ extern "C" {
    SSL structure. This is what we would ultimately want though... */
 #define TSI_SSL_MAX_PROTECTION_OVERHEAD 100
 
-
 /* For mbedtls */
 static const char *pers = "TizenRT";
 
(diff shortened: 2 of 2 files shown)


seinfra:
Code Rule Check OK.

external/grpc/src/core/tsi/ssl_transport_security.cc
```diff
@@ -77,15 +77,25 @@ static const char *pers = "TizenRT";
 #define PEM_BEGIN_CRT           "-----BEGIN CERTIFICATE-----\n"
 #define PEM_END_CRT             "-----END CERTIFICATE-----\n"
 
-#define MBEDTLS_BUF_MAX_SIZE  128
+#define TSI_BUF_SIZE  128
 
 #define DEBUG_LEVEL 0
```
girishks2000: The external folder concept was brought to keep the imported code from other repository. This code might be maintained by others and keeps on growing with features and bug fixes. If we modify these imported files and modify it will be difficult to further upgrade to new code version. So 
either 
1, remove code from external, modify and maintain it as part of TizenRT
2. Donot modify the original cloned code, but create interface/wrapper with TizenRT. This will help in upgrading the external source easily

seinfra:
Code Rule Check OK.

external/grpc/src/core/tsi/ssl_transport_security.cc
pillip8282: I agree that we shouldn't modify sources as much as possible in an external folder. 
But we should modify grpc itself because tizenrt doesn't support full features that linux provides.
We should change security layer because original grpc uses openssl.
So I think that we should keep going on this way for grpc.

external/grpc/Kconfig.protocol
```diff
@@ -10,3 +10,21 @@ config GRPC
 	depends on HAVE_CXX
 	---help---
 		Enable support for the gRPC.
+
+if GRPC
+
+comment "gRPC Config Parameters"
```
sunghan-chang: "Config Parameters" is unnecessary comment. All of config are config parameters.
How about "gRPC Security Parameters"?

sunghan-chang: And to show those configs, how about making GRPC as menuconfig?

an4967: @sunghan-chang I removed the commend as you said. I will add new comment if there would be more than 2 options in the future,

PR #1362, opened 2018-03-21
[Audio] Remove pcm_decoder from audio path

kishore-sn:
Remove the usage of pcm decoder in kernel side audio code and corresponding
code in tinyalsa to insert pcm header during audio recording. PCM header insertion
will be handled by media player based on requirement.


build/configs/artik053/audio/defconfig
```diff
@@ -815,7 +815,7 @@ CONFIG_AUDIO_BUFSIZE=2048
 #
 # CONFIG_AUDIO_FORMAT_AC3 is not set
 # CONFIG_AUDIO_FORMAT_DTS is not set
-CONFIG_AUDIO_FORMAT_PCM=y
+# CONFIG_AUDIO_FORMAT_PCM is not set
```
Taejun-Kwon: please check these configs are needed or not.

os/arch/arm/src/artik05x/src/artik055_alc5658.c
```diff
@@ -296,19 +295,8 @@ int s5j_alc5658_initialize(int minor)
...
-
-		pcm = pcm_decode_initialize(alc5658);
-		if (!pcm) {
-			auddbg("ERROR: Failed create the PCM decoder\n");
-			ret = -ENODEV;
-			goto errout_with_alc5658;
-		}
 
-		ret = audio_register(devname, pcm);
+		ret = audio_register(devname, alc5658);
```
Taejun-Kwon: Then we can delete pcm_decoder.c, can't we?

build/configs/artik053/audio/defconfig
kishore-sn: I will remove CONFIG_AUDIO_FORMAT_PCM. But I will keep the other FORMAT_xxx configs. They might be required in future if h/w supports those formats.

os/arch/arm/src/artik05x/src/artik055_alc5658.c
kishore-sn: Yes. I will remove it.

seinfra:
Code Rule Check OK.

btheosam:
If pcm decoder is removed, then corresponding formats in audio.h makes no sense. Please remove that as well.

seinfra:
Code Rule Check OK.

PR #1364, opened 2018-03-21
apps/utils/kill(all): enable task termination with any configs

sunghan-chang:
The kill and killall functions can always run task termination, SIGKILL
even if signal and pthread are disabled.
So, remove a signal conditional on Kconfig and Makefile of kill and
make conditionals on source code for signal and pthread.
Kill and Killall is controled by only CONFIG_ENABLE_KILL(ALL).


seinfra:
Code Rule Check OK.

PR #1368, opened 2018-03-22
Should deinitialize database after removing relations (db_cleanup).

ahmet-celik:
Original code tries to remove the relation after deinitializing database. I observed kernel panic due to illegal memory access occasionally during the execution of `arastorage_itc` tests on ARTIK053 board. After this fix, I did not observe the kernel panic.


seinfra:
Code Rule Check OK.

sunghan-chang:
@ahmet-celik Thank you for contribution. Every developer checks commit, not pull request so that we think commit title and description are much important. Could you copy your pull request description into commit description like below?
```
testcase: Should deinitialize database after removing relations (db_cleanup).

Original code tries to remove the relation after deinitializing database. I observed kernel panic due to illegal memory access occasionally during the execution of arastorage_itc tests on ARTIK053 board. After this fix, I did not observe the kernel panic.
```

ahmet-celik:
I updated commit message.

seinfra:
Code Rule Check OK.

PR #1371, opened 2018-03-23
external/protobuf: Fix build script in a protobuf sample.

pillip8282:
protocol buffer sample program doesn't need grpc plugin build.
So all grpc plugin features in protobuf sample could be deleted.


seinfra:
Code Rule Check OK.

apps/examples/grpc_greeter_client/Kconfig
```diff
@@ -5,6 +5,7 @@
 
 config EXAMPLES_GREETER_CLIENT
 	bool "grpc client example"
+    depends on GRPC
```
sunghan-chang: Could you align new line? Tab instead of spaces makes it align

pillip8282: Thanks for your comment. I will fix

pillip8282: And it will be fixed by another pull request so I will reset it

seinfra:
Code Rule Check OK.

PR #1375, opened 2018-03-23
framework/st_things : change thread creation logic and thread names

JeongJunSik:
- add const
- change thread creation logic and thread names


seinfra:
Code Rule Check Result:
framework/src/st_things/things_stack/utils/things_rtos_util.c:64: ERROR: [SPC_M_OPR] space required after that ',' (ctx:VxV)

PR #1378, opened 2018-03-23
framework/st_things : Reduce max path length.

Hyunjun85:
We have decided the max size of path as 32.
Now, the full path of file cannot be over 32.


seinfra:
Code Rule Check OK.

PR #1379, opened 2018-03-23
framework/st_things : Change max length of path and ssid as 32.

Hyunjun85:
We have decided the max length of path and ssid as 32.
Now, the those size cannot be over 32.


seinfra:
Code Rule Check OK.

PR #1381, opened 2018-03-23
apps/shell: add new configs, DEBUG_TASH_ERROR and DEBUB_TASH_VERBOSE

sunghan-chang:
TASH has debug APIs, shdbg and shvdbg. They are splitted by debug level.
But it has only one config so that there is no difference between them.
This commit makes them working as original purpose of debug level.


seinfra:
Code Rule Check OK.

PR #1387, opened 2018-03-26
framework/st_things : Processing the result of st_things_start()

JeongJunSik:
- Logic processing based on the result of st_things_start()


seinfra:
Code Rule Check OK.

PR #1395, opened 2018-03-28
framework/st_things : change "malloc" to "util_malloc"

JeongJunSik:
- To measure memory usage by module, change malloc to util_malloc (st_things_util)
- Remove Tizen Define

### DIFF
--- a/framework/src/st_things/st_things_logger.c
+++ b/framework/src/st_things/st_things_logger.c
@@ -26,9 +26,9 @@
 #include <stdlib.h>
 
 #include "st_things_logger.h"
+#include "st_things_util.h"
 #include "string.h"
 
-#ifndef __TIZEN__
 static const uint16_t LINE_BUFFER_SIZE = 255;
 static const char *LEVEL[] __attribute__((unused)) = {
 	"DEBUG", "INFO", "WARNING", "ERROR", "FATAL"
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check OK.

framework/src/st_things/st_things_logger.c
```diff
@@ -45,7 +45,7 @@ void st_things_log_init()
 void st_things_log_shutdown()
 {
 	/* Free the memory allocated for p_version */
-	free(p_version);
+	util_free(p_version);
```
sunghan-chang: not ```things_free```? What is difference between things and util? Isn't ```util``` a part of st_things?

JeongJunSik: util_ is used in st_things SDK level
things_ is used in st_things stack inside.

sunghan-chang: 1. Are SDK and stack separated by folder? or Is it mentioned on file the difference?
2. Anyway SDK is a part of st_things as well. Is there any chance getting memory usages separately?

JeongJunSik: 1. yes. its separated by folder 
 - sdk : framework/st_things 
 -stack : framework/st_things/things_stack
2. Yes. So i call things_malloc(in util_malloc)
However, it can be separated later to distinguish it from the SDK (APP).

juitem: let's apply with unified rule with next commit

PR #1397, opened 2018-03-28
apps/builtin : Add omitted const identifier

jeongchanKim:
tash_taskinfo_list[] is declared when build time, so it cannot be changed.
so set to const.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1400, opened 2018-03-29
Audio: add support for changing volume settings for ALC5658 codec

btheosam:
ALC5658 thru 3.5mm jack at present works with fixed volume settings.
This commit allows to change the volume of alc5658 for 3.5mm jack ONLY.

### DIFF
--- a/os/drivers/audio/alc5658.c
+++ b/os/drivers/audio/alc5658.c
@@ -235,22 +235,6 @@ static uint16_t alc5658_modifyreg(FAR struct alc5658_dev_s *priv, uint16_t regad
 	return alc5658_readreg(priv, regaddr);
 }
 
-static void alc5658_setregs(struct alc5658_dev_s *priv)
-{
-	alc5658_writereg(priv, ALC5658_IN1, (0 + 16) << 8);
-	alc5658_writereg(priv, ALC5658_HPOUT, 0);
-	alc5658_writereg(priv, ALC5658_HPOUT_L, 0xd00);
-	alc5658_writereg(priv, ALC5658_HPOUT_R, 0x700);
-}
-
-static void alc5658_getregs(struct alc5658_dev_s *priv)
-{
-	audvdbg("MIC GAIN 0x%x\n", (uint32_t)alc5658_readreg(priv, ALC5658_IN1));
-	audvdbg("MUTE HPOUT MUTE %x\n", (uint32_t)alc5658_readreg(priv, ALC5658_HPOUT));
-	audvdbg("VOLL 0x%x\n", (uint32_t)alc5658_readreg(priv, ALC5658_HPOUT_L));
-	audvdbg("VOLR 0x%x\n", (uint32_t)alc5658_readreg(priv, ALC5658_HPOUT_R));
-}
-
 /************************************************************************************
  * Name: alc5658_exec_i2c_script
  *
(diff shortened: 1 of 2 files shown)


seinfra:
Code Rule Check Result:
os/drivers/audio/alc5658.c:306: ERROR: [SPC_M_KWD] space required before the open brace '{'
os/drivers/audio/alc5658.c:306: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/drivers/audio/alc5658.c:312: ERROR: [BRC_M_SMT] else should follow close brace '}'
os/drivers/audio/alc5658.c:576: ERROR: [SPC_M_KWD] space required before the open brace '{'
os/drivers/audio/alc5658.c:581: ERROR: [BRC_M_SMT] else should follow close brace '}'
os/drivers/audio/alc5658.h:76: ERROR: [IDT_M_TAB] please, no space before tabs
os/drivers/audio/alc5658.h:77: ERROR: [IDT_M_TAB] please, no space before tabs

btheosam:
@sunghan-chang All fixed,

os/drivers/audio/alc5658.h
```diff
@@ -119,7 +122,7 @@ struct alc5658_dev_s {
 #ifndef CONFIG_AUDIO_EXCLUDE_BALANCE
 	uint16_t balance;			/* Current balance level (b16) */
 #endif							/* CONFIG_AUDIO_EXCLUDE_BALANCE */
-	uint8_t volume;				/* Current volume level {0..63} */
+	uint16_t volume;				/* Current volume level {0..31} */
```
Taejun-Kwon: why type has changed? any reason?

os/drivers/audio/alc5658.c
```diff
@@ -672,10 +658,6 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
 	case AUDIO_TYPE_PROCESSING:
 		break;
 	}
-
-	alc5658_setregs(priv);
-	alc5658_getregs(priv);
```
Taejun-Kwon: I think getregs is useful for debuging, but definetly rename needed.

seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.h
btheosam: Yes, first it is a 16bit register (easy to store), second, the volume levels in Nuttx are upto 1000, if we have to stick to it, then 8bits wont be enough.

os/drivers/audio/alc5658.c
btheosam: Okay, will keep it.

os/drivers/audio/alc5658.c
```diff
@@ -832,7 +814,7 @@ static int alc5658_pause(FAR struct audio_lowerhalf_s *dev)
 		/* Disable interrupts to prevent us from suppling any more data */
 
 		priv->paused = true;
-		alc5658_setvolume(priv, priv->volume, true);
+		alc5658_setvolume(priv);
```
Taejun-Kwon: Mute is needed. for below case.
Pre-status : volume level is 20 and there are 2 songs in track
a.mp3 finished -> set mute to reduce noise -> unmute and play b.mp3

btheosam: Yes, Mute is definitely needed. But, mutes changes in many other places are also missing. Will add & fix mute in another commit.

seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.c
```diff
@@ -584,15 +588,13 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
...
 
 			uint16_t volume = caps->ac_controls.hw[0];
-			audvdbg("    Volume: %d\n", volume);
-
-			if (volume >= 0 && volume <= 1000) {
-				/* Scale the volume setting to the range {0.. 63} */
 
-				alc5658_setvolume(priv, (63 * volume / 1000), priv->mute);
-			} else {
+			if (volume >= 0 && volume <= 31) {
```
Taejun-Kwon: If 0 means min and 31 is max, then why don't you use #define??

Taejun-Kwon: Or you'd better add 2 variable inside of priv to maintain min & max

btheosam: This variable just stores the value set through configuration. So just one is enough.
But as you said, the min & max values can be changed to MACROs. Will do.

seinfra:
Code Rule Check OK.

os/drivers/audio/alc5658.h
girishks2000: volume level as per new comment is 31 only

PR #1404, opened 2018-03-29
framework/st_things : fix implicit memory leak.

Hyunjun85:
It doesn't free the memory which has allocated device and resource type information.
The memory was allocatee when stack started to build resource and data models.


seinfra:
Code Rule Check OK.

seinfra:
Code Rule Check OK.

PR #1405, opened 2018-03-30
framework/st_things : remove spec_ver & data_model_ver parsing logic

JeongJunSik:
- it does not need to set it by user
The Iotivity version must use the embedded version.


seinfra:
Code Rule Check OK.

PR #1407, opened 2018-03-30
framework/st_things : Changes OCRepSet/Add*AsOwner to OCRep*.

Hyunjun85:
OCRepSet/Add*AsOwner function doesn't allocated memory when build the response payload.
So, user should allocate the memory for payload but user can't free.

In order to ignore this issue , We can use OCRep* instead of OCRepSet/Add*AsOwner. Therefore we should care of the pointers to avoid memory leak.


seinfra:
Code Rule Check OK.

Hyunjun85:
Fix conflict.

PR #1409, opened 2018-04-02
framework/st_things : remove unnecessary variables & logic

JeongJunSik:
- remove unnecessary variables variable(is_set_device_property)
- change return value (int -> enum)
- set the device name from data_manager in esm_set_device_preperty

### DIFF
--- a/framework/src/st_things/things_stack/easy-setup/easysetup_manager.c
+++ b/framework/src/st_things/things_stack/easy-setup/easysetup_manager.c
@@ -36,6 +36,7 @@
 #include "easysetup.h"
 #include "logging/things_logger.h"
 #include "memory/things_malloc.h"
+#include "framework/things_data_manager.h"
 
 #include "utils/things_thread.h"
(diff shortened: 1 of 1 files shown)


framework/src/st_things/things_stack/easy-setup/easysetup_manager.c
```diff
@@ -277,14 +273,13 @@ esm_result_e esm_init_easysetup(int restart_flag, things_server_builder_s *serve
...
 
-	if (server_builder != NULL) {
-		g_server_builder = server_builder;
-	}
-
 	if (g_server_builder == NULL) {
 		THINGS_LOG_V_ERROR(THINGS_ERROR, TAG, "[Error] g_server_builder is NULL.");
 		return ESM_ERROR;
+	} else {
+		g_server_builder = server_builder;
```
Hyunjun85: Why did this logic move?
In this line, g_server_builder will be null if server_builder  was null.

PR #1410, opened 2018-04-02
s5j: Add Tick Suppression feature support

lokeshbv:
Adds Tick Suppression feature support for s5j platform.

> Tick suppression can generally be enabled and left enabled without any drawbacks.

> The way tick suppression works is that whenever the scheduler sees an expected idle time
 (no tasks to be scheduled) longer than the time remaining to the next scheduled timer interrupt, it reschedules the next tick/interval timer interrupt to a much later time (either when the next known tasks are scheduled to run or to a maximum possible tick timer value).

> Upon coming out of the idle state, whether the exit was caused by the scheduled interval timer interrupt or an asynchronous external interrupt, the system time and other state information is properly adjusted so that any running tasks are not affected – they see “the world” the same regardless if tick suppression happened or not.

This feature can be enabled through menuconfig as follows,

> Kernel Features > Clocks and Timers > [*] Support Tick Suppression
CONFIG_SCHED_TICKSUPPRESS=y

### DIFF
--- a/os/arch/arm/src/s5j/s5j_idle.c
+++ b/os/arch/arm/src/s5j/s5j_idle.c
@@ -119,6 +119,10 @@ static void up_idlepm(void)
 			break;
 
 		case PM_SLEEP:
+#ifdef CONFIG_SCHED_TICKSUPPRESS
+			/* Enable tick suppression */
+			up_start_ticksuppress();
+#endif
 			s5j_pmstandby();
 			break;
 
@@ -160,6 +164,11 @@ void up_idle(void)
 #else
 	/* Perform IDLE mode power management */
 	up_idlepm();
+
+#if !defined CONFIG_PM && defined CONFIG_SCHED_TICKSUPPRESS
+	/* Enable tick suppression */
+	up_start_ticksuppress();
+#endif
 
 	/* Sleep until an interrupt occurs to save power. */
 	asm("WFI");
--- a/os/arch/arm/src/s5j/s5j_ticksuppress.c
+++ b/os/arch/arm/src/s5j/s5j_ticksuppress.c
@@ -1,0 +1,273 @@
+/****************************************************************************
+ *
+ * Copyright 2018 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <tinyara/config.h>
+
+#include <debug.h>
+#include <queue.h>
+#include <tinyara/wdog.h>
+#include <tinyara/clock.h>
+#include <tinyara/time.h>
+#include <tinyara/wdog.h>
+#include <arch/board/board.h>
+
+#include "up_arch.h"
+#include "chip/s5jt200_rtc.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+/*
+ * The desired timer interrupt frequency is provided by the definition
+ * CLK_TCK (see include/time.h).  CLK_TCK defines the desired number of
+ * system clock ticks per second.  That value is a user configurable setting
+ * that defaults to 100 (100 ticks per second = 10 MS interval).
+ *
+ * The timer counts at the rate SYSCLK_FREQUENCY as defined in the board.h
+ * header file.
+ */
+#define SYSTICK_RELOAD ((SYSCLK_FREQUENCY / CLK_TCK) - 1)
+
+#define MAX_TCNT UINT32_MAX	/* Max 32-bit timer-tick value */
+
+/****************************************************************************
+ * Type Definitions
+ ****************************************************************************/
+enum {
+	TICKSUPPRESS_DISABLE,
+	TICKSUPPRESS_ENABLE
+} timertick_mode;
+
+enum {
+	START,
+	READ,
+	WRITE,
+	STOP
+} timer_states;
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+static uint8_t g_tick_mode;	/* timer-tick mode  */
+
+static uint32_t last_event = 0;
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+static uint32_t s5j_get_next_event(void)
+{
+	return (uint32_t)wd_getdelay();
+}
+
+static uint32_t s5j_timer_setup(int state, uint32_t value)
+{
+	switch (state) {
+		case START:
+			modifyreg32(S5J_RTC_RTCCON, RTC_RTCCON_TICCKSEL0_MASK, RTC_RTCCON_TICKEN0_ENABLE | RTC_RTCCON_TICCKSEL0_32768HZ);
+			break;
+		case READ:
+			return getreg32(S5J_RTC_CURTICCNT0);
+		case WRITE:
+			putreg32(SYSTICK_RELOAD * value, S5J_RTC_TICCNT0);
+			break;
+		case STOP:
+			/* Clear interrupt pending */
+
+			putreg32(RTC_INTP_TIMETIC0, S5J_RTC_INTP);
+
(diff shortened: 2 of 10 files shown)


os/arch/arm/src/s5j/s5j_ticksuppress.c
```diff
@@ -0,0 +1,273 @@
...
+	uint32_t next_event = 0;
+
+	/* Check whether the threshold value is reached or not
+	 * If threshold value is crossed, ticks are suppressed and
+	 * next expected event is scheduled.
+	 */
+
+	s5j_timer_setup(STOP, 0);
+
+	next_event = s5j_get_next_event();
```
pradeepns85: We may have to handle the scenario, when next_event is bigger than MAX_TCNT  . what's your opinion ?
for ex, if MAX_TCNT = 10 sec and next_event is 13 sec.

os/arch/arm/src/s5j/s5j_ticksuppress.c
```diff
@@ -0,0 +1,273 @@
...
+	if (g_tick_mode == TICKSUPPRESS_DISABLE) {
+		return;
+	}
+
+	/* If this function called from interrupt context,
+	 * read current value of the counter and update last event value
+	 */
+
+	if (up_interrupt_context()) {
+		last_event = ((uint32_t)s5j_timer_setup(READ, 0) / SYSTICK_RELOAD);
```
pradeepns85: does s5j provides a overflow counter to hold the number of times, the counter has been overflowed ? If it provides, we may have to consider handling that case as well

os/arch/arm/src/s5j/s5j_ticksuppress.c
```diff
@@ -0,0 +1,273 @@
...
+	if (!next_event) {
+		next_event = MAX_TCNT;
+	}
+	last_event = next_event;
+
+	g_tick_mode = TICKSUPPRESS_ENABLE;
+
+	/* Schedule next expected event */
+
+	s5j_set_next_event(next_event);
```
pradeepns85: As i observe, to change the timer to periodic, we are passing 0x1 as argument as below 
s5j_set_next_event(0x1); 
What if the next_event returned by s5j_get_next_event() is also 0x1 ? what is expected behavior?

os/include/tinyara/arch.h
```diff
@@ -1603,6 +1603,62 @@ int up_timer_cancel(FAR struct timespec *ts);
...
+ *	specific logic.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
+ * Assumptions:
+ *	Called from interrupt level handling, interrupts may need to be disabled
```
pradeepns85: rephrase the sentence. it's called from  ISR handler.

os/include/tinyara/arch.h
```diff
@@ -1603,6 +1603,62 @@ int up_timer_cancel(FAR struct timespec *ts);
...
+
+#ifdef CONFIG_SCHED_TICKSUPPRESS
+void up_start_ticksuppress(void);
+#endif
+
+/****************************************************************************
+ * Name: up_stop_ticksuppress
+ *
+ * Description:
+ *  Periodic timer tick interrupts interrupts are re-enabled.
```
pradeepns85: word "interrupts" used twice

os/include/tinyara/arch.h
```diff
@@ -1603,6 +1603,62 @@ int up_timer_cancel(FAR struct timespec *ts);
...
+ *  specific logic.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None
+ *
+ * Assumptions:
+ *  Called from interrupt level handling, interrupts may need to be disabled
```
pradeepns85: interrupt level => ISR

os/include/tinyara/wdog.h
```diff
@@ -180,6 +180,11 @@ int wd_start(WDOG_ID wdog, int delay, wdentry_t wdentry, int argc, ...);
 int wd_cancel(WDOG_ID wdog);
 int wd_gettime(WDOG_ID wdog);
 
+#ifdef CONFIG_SCHED_TICKSUPPRESS
+int wd_getdelay(void);
+void wd_updatedelay(int elapsed);
```
pradeepns85: What if a timer offers 64 bit registers, then the max threshold may change to 2 power 64.  should we consider that or assume that it's always 32 bit register and use int as its data type

os/kernel/wdog/wd_gettime.c
```diff
@@ -129,3 +129,53 @@ int wd_gettime(WDOG_ID wdog)
...
+ *
+ * Assumptions:
+ *
+ ********************************************************************************/
+
+void wd_updatedelay(int elapsed)
+{
+	if (g_wdactivelist.head) {
+		/* Update  wdog lag (delay) */
+		((FAR struct wdog_s *)g_wdactivelist.head)->lag -= elapsed;
```
pradeepns85: Are we not suppose to take care of updating the delay of all the members of  g_wdactivelist ?

os/arch/arm/src/s5j/s5j_idle.c
```diff
@@ -161,6 +165,11 @@ void up_idle(void)
 	/* Perform IDLE mode power management */
 	up_idlepm();
 
+#if !defined CONFIG_PM && defined CONFIG_SCHED_TICKSUPPRESS
```
pradeepns85: I don't understand, why CONFIG_PM flag has been checked  here ? whats the dependency of this feature with CONFIG_PM ?  I believe these are independent modules . isn't it ?

os/arch/arm/src/s5j/s5j_ticksuppress.c
```diff
@@ -0,0 +1,273 @@
...
+	WRITE,
+	STOP
+} timer_states;
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+static uint8_t g_tick_mode;	/* timer-tick mode  */
+
+static uint32_t last_event = 0;
```
pradeepns85: Am not sure, but re-writing these by changing the declaring order would save some bytes ?
static uint32_t last_event = 0;
static uint8_t g_tick_mode;

os/arch/arm/src/s5j/s5j_ticksuppress.c
lokeshbv: timer interrupt comes after 1 tick

os/include/tinyara/wdog.h
lokeshbv: As already defined in **include/tinyara/wdog.h**, wdog delay itself is limited by type '**int**' . https://github.com/Samsung/TizenRT/blob/8cb7c62c626e7f37c7ab72413444024785ca9942/os/include/tinyara/wdog.h#L152

os/kernel/wdog/wd_gettime.c
lokeshbv: I think not needed, as wdog active list holds members in order and one which would expire first will be at the head of the active list.

os/arch/arm/src/s5j/s5j_ticksuppress.c
lokeshbv: I'm also not sure, could you please explain how?

os/arch/arm/src/s5j/s5j_idle.c
lokeshbv: Existing PM framework is based on periodic timer tick interrupts. If we suppress timer tick interrupts, then PM algorithm doesn't work.
So I have handled that case as follows,
1) If ```CONFIG_PM``` is not enabled  and ```CONFIG_SCHED_TICKSUPPRESS```  is enabled
---> apply tick suppression immediately.

2) If ```CONFIG_PM``` is enabled and ```CONFIG_SCHED_TICKSUPPRESS```  is also enabled
---> apply tick suppression at the last state of PM states i.e. after PM algorithm completes.

os/arch/arm/src/s5j/s5j_ticksuppress.c
lokeshbv: I'm not sure about such overflow counter in s5j. I will check

os/arch/arm/src/s5j/s5j_ticksuppress.c
lokeshbv: yes, this case has to be handled. Thank you:)

lokeshbv: @pradeepns85 , I think this case will not occur as delay from user application is limited by ***unsigned int***. So max possible delay would be ***0xFFFFFFFF***. Am I right??

***unsigned int sleep(unsigned int seconds);***

seinfra:
Target : [f12e759f53fbb2f351d785cca9cc43523c5a1dfe] - Code Rule Check (C++) OK.

seinfra:
Target : [f12e759f53fbb2f351d785cca9cc43523c5a1dfe] -

PR #1414, opened 2018-04-03
framework/st_things : [defconfig] NFILE_DESCRIPTORS increased

JeongJunSik:
- change CONFIG_NFILE_DESCRIPTORS=16 to 64


sunghan-chang:
You modified build/configs, not framework/st_things.
```
configs/artik053(s)/st_thigns: change a default value of NFILE_DESCRIPTORS  to 64

write the reason
```

JeongJunSik:
Hello default config is 64.
so. i follow it

JeongJunSik:
- 64 is not an optimized value, but this config is used by other apps as Ref
It is changed in consideration of convenience.

juitem:
would you please write those reason as a commit comment?

PR #1416, opened 2018-04-03
mm/heapinfo : Add omitted header include

jeongchanKim:
for strncmp() and strlen, #include <string.h> is needed, but omitted


os/mm/mm_heap/mm_heapinfo.c
```diff
@@ -61,6 +61,7 @@
 #include <stdio.h>
 #ifdef CONFIG_HEAPINFO_GROUP
 #include <tinyara/mm/heapinfo_internal.h>
+#include <string.h>
```
sunghan-chang: Let's move this line above <tinyara/mm/heapinfo_internal.h>.
Order to include header is
1. public standard header
2. public os-specific header
3. local header

PR #1418, opened 2018-04-03
Audio: Add mute support for alc5658 audio codec

btheosam:
Add support to mute and unmute alc5658 codec for headphones ONLY

### DIFF
--- a/os/drivers/audio/alc5658.c
+++ b/os/drivers/audio/alc5658.c
@@ -78,6 +78,7 @@
 #include <tinyara/math.h>
 #include <math.h>
 #include <tinyara/i2c.h>
+#include <stdbool.h>
 
 #include "alc5658.h"
 #include "alc5658scripts.h"
@@ -595,6 +596,16 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
 				alc5658_setvolume(priv);
 			} else
 				ret = -EDOM;
+		}
+		break;
+
+		case AUDIO_FU_MUTE: {
+			/* Mute or unmute:  true(1) or false(0) */
+
+			bool mute = caps->ac_controls.b[0];
+			audvdbg("mute: 0x%x\n", mute);
+			priv->mute = mute;
+			alc5658_setvolume(priv);
 		}
 		break;
 #endif							/* CONFIG_AUDIO_EXCLUDE_VOLUME */
--- a/os/drivers/audio/alc5658.h
+++ b/os/drivers/audio/alc5658.h
@@ -73,7 +73,7 @@
 #define ALC5658_DEFAULT_NCHANNELS	2
 #define ALC5658_DEFAULT_BPSAMP		16
 #define FAIL						0xFFFF
-#define MUTE_HP_LR					0x8010
+#define MUTE_HP_LR					0x8080
 #define UNMUTE_HP_LR				0x0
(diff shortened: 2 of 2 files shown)


os/drivers/audio/alc5658.c
```diff
@@ -597,6 +597,21 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
...
 		}
 		break;
+
+		case AUDIO_FU_MUTE: {
+			/* Mute or unmute:  true(1) or false(0) */
+
+			bool mute = caps->ac_controls.hw[0];
+			auddbg("    mute: 0x%x\n", mute);
+
+			if (false == mute || true == mute) {
```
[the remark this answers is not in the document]
btheosam: No. Just mute and unmute only. the check further prevents unintentional mute thru static identifiers garbage value. Nothing else.

os/drivers/audio/alc5658.c
```diff
@@ -597,6 +597,21 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
...
+					priv->mute = mute;
+					alc5658_setvolume(priv);
+				}
+			else
```
pradeepns85: kindly use the flower brackets for empty else case as well. < coding guidelines >

os/drivers/audio/alc5658.c
```diff
@@ -597,6 +597,21 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
 				ret = -EDOM;
 		}
 		break;
+
+		case AUDIO_FU_MUTE: {
+			/* Mute or unmute:  true(1) or false(0) */
+
+			bool mute = caps->ac_controls.hw[0];
```
sunghan-chang: to use bool type, you should include <stdbool.h>

os/drivers/audio/alc5658.c
```diff
@@ -597,6 +597,21 @@ static int alc5658_configure(FAR struct audio_lowerhalf_s *dev, FAR const struct
...
+			auddbg("    mute: 0x%x\n", mute);
```
sunghan-chang: It looks information. Let's use audvdbg if it is.

os/drivers/audio/alc5658.c
sunghan-chang: Additionally, 
1. line 610, need to remove one tab
2. else should be at next of closing bracket
3. As Pradeep said, even though else has a single statemen, you recommand using brackets.

os/drivers/audio/alc5658.c
Taejun-Kwon: we can use b[0] as a uint8_t instead of uint16_t that's why it defined as union, I think.

os/drivers/audio/alc5658.c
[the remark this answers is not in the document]
Taejun-Kwon: I think 0 = false, none 0 = true is enough.

os/drivers/audio/alc5658.c
btheosam: @Taejun-Kwon can you elaborate? Yes, either b[0] or hw can be used, Do you mean uint8_t is compatible to char(bool)?
If so, then yes, it makes sense to modify. If anything else, please expalin.

btheosam:
You are right, this line dont add any value with gcc.

PR #1421, opened 2018-04-03
tidy up for cy4390x

sunghan-chang:
1. make a dependacy RAM_DDR_XXX with DDR
2. modify syntax of help
3. move location of DDR
4. add "cy4390x" on supported board / emulator of README

### DIFF
--- a/os/arch/Kconfig.chip
+++ b/os/arch/Kconfig.chip
@@ -743,6 +743,7 @@ config RAM_SIZE
 config RAM_DDR_START
 	hex "DDR RAM start address"
 	default 0x0
+	depends on DDR
 	--help--
 		If the board using external DDR RAM memory, you need to define
 		the start address of DDR RAM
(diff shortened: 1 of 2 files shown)


os/arch/Kconfig.chip
```diff
@@ -734,23 +729,30 @@ config RAM_VSTART
...
+	---help---
 		The size in bytes of the installed primary RAM. "Primary" RAM
 		refers to the RAM that you link program code into.  If program code
 		does not execute out of RAM but from FLASH, then you may designate
 		any block of RAM as "primary."
 
+config DDR
+	bool "Use DDR memory"
+	default n
+	depends on ARCH_CHIP_BCM4390X
```
pradeepns85: I wonder, whether we need to create this dependency with respect to specific chip? Because any board which supports external DDR can enable this.  In my opinion default n would be sufficient .

PR #1422, opened 2018-04-04
Enable Ethernet MTU size when making menuconfig

JeonginKim:
Change the Ethernet MTU size when making menuconfig


sunghan-chang:
@JeonginKim This change is not "Change default ethernet MTU size". Please write approprietary title which shows what you modify.

sunghan-chang:
@JeonginKim And, copy and paste for description is not allowed. Please write why you make this, what issue is.

PR #1424, opened 2018-04-04
fixed crash when scan done

CypressHJ:
when scan is executed, if scan result is  NULL, the crash has occurred. 

BR,  HJ Lee.


pradeepns85:
Dear HJ Lee,

If it's a bug fix, usually we write the module in which bug is fixed as part of commit HEADER

Can you please write something like this?
apps/examples/wifi_manager_sample : Fixed crash when scan done

Best Regards,
Pradeep

sunghan-chang:
@CypressHJ Hello.
As I told at offline, please find our commit rules [here](https://github.com/Samsung/TizenRT/wiki/Documentations#commit-rules).

1. Pradeep comment
2. don't use past tense
3. fill commit description

i.e.
```
apps/examples/wifi_manager_sample : Fix crash when scan done

when scan is executed, if scan result is NULL, the crash has occurred.
```

PR #1428, opened 2018-04-05
framework/st_things: Fix implicit bug to create collection resource.

Hyunjun85:
There can be showing a bug when using collection resource.
In previous version, It passes a uri instead of type.
This is an fault what is entering wrong parameters to create resource.


Hyunjun85:
Already done.

PR #1437, opened 2018-04-06
external/mbedtls : enable debugging function call in mbedtls

an4967:
- Remove '#if' statement for general debugging function

### DIFF
--- a/external/mbedtls/debug.c
+++ b/external/mbedtls/debug.c
@@ -81,16 +81,12 @@ static inline void debug_send_line(const mbedtls_ssl_context *ssl, int level, co
 	 * Since there is no portable way to get one, use the address of the ssl
 	 * context instead, as it shouldn't be shared between threads.
 	 */
-#if 0							/* For TinyAra */
 #if defined(MBEDTLS_THREADING_C)
 	char idstr[20 + DEBUG_BUF_SIZE];	/* 0x + 16 nibbles + ': ' */
 	mbedtls_snprintf(idstr, sizeof(idstr), "%p: %s", ssl, str);
 	ssl->conf->f_dbg(ssl->conf->p_dbg, level, file, line, idstr);
 #else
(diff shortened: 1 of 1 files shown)


external/grpc/src/core/tsi/ssl_transport_security.cc
```diff
@@ -154,7 +154,7 @@ typedef struct {
 /* Debug for mbedtls SSL process */
 static void grpc_tsi_debug(void *ctx, int level, const char *file, int line, const char *str)
 {
-  (void) ctx;
+  printf("%s(%d) : %s", file, line, str);
```
pillip8282: Calling the grpc debug method would be better.

an4967:
@pillip8282 I sperated the PR in mbedtls part and grpc part.
So, I want mbedtls fix would be reviewed first.

external/mbedtls/debug.c
```diff
@@ -81,17 +81,13 @@ static inline void debug_send_line(const mbedtls_ssl_context *ssl, int level, co
 	 * Since there is no portable way to get one, use the address of the ssl
 	 * context instead, as it shouldn't be shared between threads.
 	 */
-#if 0							/* For TinyAra */
 #if defined(MBEDTLS_THREADING_C)
```
pradeepns85: Is it really required to make this API as inline function ? it's been called from several functions and it just increases code space.  May not be directly linked to your changes but am just mentioning the overhead of removing #if 0

an4967: @pradeepns85 
if you see line 103 in debug.c, it always checks 'ssl->conf->f_dbg' to be registered by users.
As following this logic, even if an user registered f_dbg, it doesn't work and wastes code space. 
Moreover, if 'f_dbg' is not set, printing debugging logs fails.

seinfra:
Target : [ebcf4359e824045a0b1913ea245ea051ae3d3fc4] - Code Rule Check (C++) OK.

PR #1450, opened 2018-04-11
tools: Add new features to build configuration tool

lokeshbv:
- Execute build configuration tool from top directory.
- If build configuration is already done, then report the same to user.
- Add informational echo statements.

This patch is back-ported from Nuttx.

### DIFF
--- a/os/tools/configure.sh
+++ b/os/tools/configure.sh
@@ -49,7 +49,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #
 
-WD=`pwd`
+WD=`test -d ${0%/*} && cd ${0%/*}; pwd`
 TOPDIR="${WD}/.."
 USAGE="
 
@@ -58,7 +58,7 @@ USAGE: ${0} [-d] [-a <app-dir>] <board-name>/<config-name>
 Where:
   <board-name> is the name of the board in the configs directory
   <config-name> is the name of the board configuration sub-directory
-  <add-dir> is the path to the apps/ directory, relative to the tinyara directory
+  <app-dir> is the path to the apps/ directory, relative to the tinyara directory
 
 "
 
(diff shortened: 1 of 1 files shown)


os/tools/configure.sh
```diff
@@ -150,6 +150,12 @@ if [ ! -r "${src_config}" ]; then
   exit 6
 fi
 
+if [ -r ${dest_config} ]; then
```
sunghan-chang: I don't know that this is right. Actually no.
If I configured once without compilation, I can reconfigure. There is no issue. But this change makes error.

How about check ${dest_config} and .version file?
Because .version is made on compilation, existing of it means compiled.
Then, we had better guide "make distclean".

Could you give your opinion?

sunghan-chang:
@lokeshbv There are many tools folder. Please write "os/tools:" on your title to clearfy.
And this is not a new feature I think.

os/tools/configure.sh
lokeshbv: @sunghan-chang,  yes I agree with you. Checking for both ${dest_config} and .version file is necessary. I made changes as per your suggestions. Please check

os/tools/configure.sh
```diff
@@ -150,6 +150,14 @@ if [ ! -r "${src_config}" ]; then
...
+  if [ -x "${TOPDIR}/.version" ]; then
```
sunghan-chang: Is this -x? not -r? .version and .config have same usages which includes configured value. .config is checked with -r at above line. But this line checks with -x. It is little strange.

os/tools/configure.sh
```diff
@@ -150,6 +150,14 @@ if [ ! -r "${src_config}" ]; then
...
+    echo "Already configured!"
```
sunghan-chang: How about "Already configured and compiled!" to give exact information?

lokeshbv: yes, I Agree

os/tools/configure.sh
lokeshbv: typo mistake perhaps

PR #1453, opened 2018-04-12
Add protobuf test app using gtest

HONGCHAEHEE:
This directory contains Protocol Buffers unit test using Google Mock and Google Test.
 configuration prerequisites: HAVE_CXX, HAVE_CXXINITIALIZE, LIBCXX, PROTOBUF, GMOCK
 configuration settings specific to this example:
 CONFIG_GMOCK=y
 CONFIG_PROTOBUF=y
 CONFIG_EXAMPLES_PROTOBUF_TEST=y


HONGCHAEHEE:
@sunghan-chang @pillip8282 I updated on this new pull request. Please review this.

apps/examples/protobuf_test/gtest_main.cc
```diff
@@ -0,0 +1,110 @@
...
+		args.argv = argv;
+
+		pthread_t tid;
+		pthread_attr_t attr;
+		int ret;
+
+		/* Initialize the attribute variable */
+		if ((ret = pthread_attr_init(&attr)) != 0) {
+			printf("%s: pthread_attr_init failed, status=%d\n", __func__, ret);
+		}
```
pradeepns85: Why it's not returning when there is a error ?

apps/examples/protobuf_test/gtest_main.cc
```diff
@@ -0,0 +1,110 @@
...
+
+		/* set a stacksize */
+		if ((ret = pthread_attr_setstacksize(&attr, PROTOBUF_TEST_STACKSIZE)) != 0) {
```
pradeepns85: attempting to do these API when previous function "pthread_attr_init" fails, is an unexpected behavior

apps/examples/protobuf_test/gtest_main.cc
```diff
@@ -0,0 +1,110 @@
...
+			printf("%s: pthread_attr_setstacksize failed, status=%d\n", __func__, ret);
+		}
+
+		/* create pthread with entry function */
+		if ((ret = pthread_create(&tid, &attr, (pthread_startroutine_t) gtest_run, (void *)&args)) != 0) {
```
pradeepns85: same as above, if previous API fails , pthread_create should not be called.

apps/examples/protobuf_test/gtest_main.cc
```diff
@@ -0,0 +1,110 @@
...
+			printf("%s: pthread_create failed, status=%d\n", __func__, ret);
+		}
+
+		/* Wait for the threads to stop */
+		pthread_join(tid, NULL);
+
+		return 0;
```
pradeepns85: returning 0 means success, it's not true always. as there are so many situation, it may fail as above. kindly return using variable "ret"

seinfra:
Target : [dd3a0bd181fceac4ac167e2f64c4eb5dab2a1825] - Code Rule Check (C++) OK.

PR #1462, opened 2018-04-13
Fix build error in heapinfo util

lokeshbv:
- apps/system/utils: Fix build error in heapinfo


apps/system/utils/kdbg_heapinfo.c
```diff
@@ -71,8 +71,8 @@ static void kdbg_heapinfo_group_threadlist(void)
 	while (*ptr != '/') {
 		printf("%c", *ptr++);
 		if (ptr == end_list) {
-			ptr = CONFIG_HEAPINFO_USER_GROUP_LIST - 1;
-			break;
+			ptr = CONFIG_HEAPINFO_USER_GROUP_LIST;
+			return;
```
jeongchanKim: This should be `break`

PR #1466, opened 2018-04-16
framework/wifi_manager: wifi_manager bug fix

SeonghoByeon:
- Add logic to ensure callback for scannig (scan_ap_done) is declared
- Handling ENT_DEINIT in RECONNECTED state
- Add logic to ensure wifi_manager_info_s is not null when it is called


SeonghoByeon:
@sunghan-chang I applied your comments. New (separated) PRs will be added. 
Please check the modified one.

JeongJunSik:
The null check for the wifi_manager_get_info argument is fine.
When using Wifi Manager, Assert Error occurs when Wifi join when registering user callback as NULL at init.
We ask you to verify that the logic is normal.

PR #1470, opened 2018-04-17
framework/st_things : url safe base64 encoding for artik h/w certific…

JeongJunSik:
…ate uuid

problem : (cci1+6o5 -> cci1%2B6o5)
solution : Base64 Encode -> Base64 URL Safe Encode


framework/src/st_things/things_stack/framework/things_sss_manager.c
```diff
@@ -306,5 +306,15 @@ bool things_encrypt_artik_uuid(unsigned char *output)
 	memcpy(output, encode_buf2, 8);
 	THINGS_LOG_D(THINGS_DEBUG, TAG, "output len [%d][%s]", strlen((char *)output), output);
 
+	// 6. url safe
+	int i;
+	for (i = 0; i < 8; i++) {
+		if (output[i] == '+')
```
sunghan-chang: Please use braces even if it has single statement

JeongJunSik: Ok. Thanks.
i will fix and commit again.

PR #1474, opened 2018-04-17
Fix a build error in webclient

HONGCHAEHEE:
- webclient.c:1064:19: error: comparison of constant '-1' with boolean expression is always false [-Werror=bool-compare]
   if (read_finish == HTTP_ERROR) {
                   ^~
cc1: all warnings being treated as errors

- This patch is to change the data type of 'read_finish' as conditions compare boolean and '-1'

### DIFF
--- a/external/webclient/webclient.c
+++ b/external/webclient/webclient.c
@@ -935,7 +935,7 @@ static pthread_addr_t wget_base(void *arg)
 	struct wget_s ws;
 	struct http_client_request_t *param = (struct http_client_request_t *)arg;
 	struct http_client_response_t response = {0, };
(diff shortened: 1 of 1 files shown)


external/webclient/webclient.c
```diff
@@ -935,7 +935,7 @@ static pthread_addr_t wget_base(void *arg)
 	struct wget_s ws;
 	struct http_client_request_t *param = (struct http_client_request_t *)arg;
 	struct http_client_response_t response = {0, };
-	bool read_finish = false;
+	int read_finish = 0;
```
sunghan-chang: Let's keep original initial value, false.

PR #1475, opened 2018-04-18
framework/wifimanager: fix Wi-Fi manager Kconfig to include  exteranl…

pillip8282:
… library

Wi-Fi Manager enabled wpa_supplicant wrapper layer by default.
This patch will prevent above error. and wpa_supplicant wrapper API will be selected
as long as wpa_supplicant is selected.


seinfra:
Target : [e800e1579bb33d4039a8b91b65a1e6647d274768] - Code Rule Check OK.

PR #1479, opened 2018-04-18
Port NGHTTP2 on TizenRT platform

sahilpai1:
- Base Code from https://github.com/nghttp2/nghttp2
	- NGHTTP2 version 1.32.0
	- Use curl_test sample application for testing


sahilpai1:
@juitem , @bhargava-cs ,@sunghan-chung,
Please review NGHTTP Patch

seinfra:
Target : [76edc56e01cc80a58c879fe1d6b93eb00a15e92b] - Code Rule Check (C++) OK.

external/curl/Kconfig
```diff
@@ -18,6 +18,7 @@ if !DISABLE_HTTP
 config USE_NGHTTP2
 	bool "Use NGHTTP2"
 	default n
+	select ENABLE_NGHTTP2
```
HONGCHAEHEE: dosen't it need dependancy?

sunghan-chang:
I heard lwip which is RT's network stack does not support AF_UNIX. You should block HAVE_UNIX_SOCKRTS from curl_config.h.
'''
#ifndef CONFIG_NET_LWIP
#define HAVE_...
#endif
'''

external/nghttp2/Kconfig
```diff
@@ -0,0 +1,11 @@
+config ENABLE_NGHTTP2
+	bool "Enable NGHTTP2"
+	default n
+	---help---
+		nghttp2 is an implementation of HTTP/2 and its header compression algorithm HPACK.
+
+if USE_NGHTTP2
```
sunghan-chang: Is this "USE_"? not "ENABLE_"?

external/nghttp2/Makefile
```diff
@@ -0,0 +1,74 @@
...
+.depend: Makefile $(SRCS)
+	$(Q) $(MKDEP) $(DEPPATH) "$(CC)" -- $(CFLAGS) -- $(SRCS) >Make.dep
+	$(Q) touch $@
+
+depend: .depend
+
+clean:
+	$(call DELFILE, .built)
+	$(call CLEAN)
+	rm -f $(OBJS)
```
sunghan-chang: Remove this line. This is duplicated with above "call CLEAN".

external/include/nghttp2/nghttp2ver.h
```diff
@@ -0,0 +1,42 @@
...
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+#ifndef NGHTTP2VER_H
+#define NGHTTP2VER_H
+
+/**
+ * @macro
+ * Version number of the nghttp2 library release
+ */
+#define NGHTTP2_VERSION "1.32.0-DEV"
```
sunghan-chang: Is this public header? "external/include" is a path of public header.
It looks not needs these two definions to other module.

external/nghttp2/Kconfig
sahilpai1: yes, this is done

external/nghttp2/Makefile
sahilpai1: ok

external/include/nghttp2/nghttp2ver.h
sunghan-chang: If these are used in nghttp inside only, I want to move this to nghttp folder, not include. But as you said, they are used in other module like curl, you can keep it. No need to change.

sahilpai1: @sunghan-chang ok

PR #1480, opened 2018-04-19
Edit grpc and lwip files for compatibility

dr-venkman:
Lwip version 2.0 needs certain modifications to allow grpc to run over it on TizenRT.  Also, grpc module on TizenRT needs modifications to interface with LWIP 2.0. A summary of these modifications are as follows:

1. Disable GRPC_HAVE_UNIX_SOCKET on grpc, TizenRT does not support Unix domain sockets at present.

2. Under lwip 2.0, AF_INET6 gets the same enum value as AF_UNSPEC, when IPv6 is not configured.  Accordingly, disable checks for AF_INET6 and AF_UNSPEC under switch-case statements.

3. Grpc module features dualsocket creation. Accordingly, expose IPPROTO_IPV6, even if lwip does not configure IPv6.

4. Declare netent, protoent, servent and servent_data structures under netdb.h, for use by grpc

5. Declare sendmsg and recvmsg socket APIs for use by grpc, in socket.h


seinfra:
Target : [dea7b2148cc24eddf8611cd1d66b69d3db240066] - Code Rule Check OK.

PR #1492, opened 2018-04-20
apps: kill apps depend on signals

rzr:
Change-Id: I31c51dab2338cf3d6235c058d9033f79f196328a

### DIFF
--- a/apps/system/utils/kernelcmd.c
+++ b/apps/system/utils/kernelcmd.c
@@ -43,11 +43,13 @@ const static tash_cmdlist_t kdbg_cmds[] = {
 #if defined(CONFIG_ENABLE_IRQINFO)
 	{"irqinfo",   kdbg_irqinfo,      TASH_EXECMD_SYNC},
 #endif
+#if ! defined(CONFIG_DISABLE_SIGNALS)
 #if defined(CONFIG_ENABLE_KILL)
 	{"kill",     kdbg_kill,         TASH_EXECMD_SYNC},
 #endif
 #if defined(CONFIG_ENABLE_KILLALL)
 	{"killall",  kdbg_killall,      TASH_EXECMD_SYNC},
+#endif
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [fa8ea3949a1ac83b1c1fabeefacbd9490d1ba02d] - Code Rule Check OK.

seinfra:
Target : [fa8ea3949a1ac83b1c1fabeefacbd9490d1ba02d] - Code Rule Check (C++) OK.

sunghan-chang:
@rzr Could you let me know the reason why you make a dependancy with SIGNAL?

rzr:
Actually dependency was introduced in declaration and not implementation.

```
cat ./apps/system/utils/kdbg_commands.h

#ifndef CONFIG_DISABLE_SIGNALS
#if defined(CONFIG_ENABLE_KILL)
int kdbg_kill(int argc, char **args);
#endif

#if defined(CONFIG_ENABLE_KILLALL)
int kdbg_killall(int argc, char **args);
#endif
#endif  
```

I have tested the other patch, but it does much that my straightforward one, so I would recommend to merge this and rebase the other.

sunghan-chang:
@rzr You're right. We missed that point. I will modify kdbg_command.h. But I don't agree why we merge yours. "kill" command has two types execution. One is sending signal, another is terminating thread. Terminating thread does not treat CONFIG_DISABLE_SIGNAL. What is your straightforward point?

rzr:
I just built master without this patch, 

92760eeee331f18dd6cd0fe6bec386b56020453a seems to remove the dependencies in
./apps/system/utils/kdbg_commands.h

So I can align to it, by closing this unmerged PR. Is it ok ?

sunghan-chang:
@rzr Thank you for confirmation. If you don't have different opinion, it is ok to close this to me.

PR #1552, opened 2018-04-25
media: Use enum value to represent pcm format

junmin-kim:
Instead of hardcoded 0, use utils::PCM_FORMAT_S16_LE

### DIFF
--- a/framework/include/media/MediaUtils.h
+++ b/framework/include/media/MediaUtils.h
@@ -15,6 +15,72 @@
  * limitations under the License.
  *
  ******************************************************************/
+/****************************************************************************
+*
+* Developed by:
+*
+*   Copyright (C) 2013 Ken Pettit. All rights reserved.
+*   Author: Ken Pettit <pettitkd@gmail.com>
+*
+* With ongoing support:
+*
+*   Copyright (C) 2014 Gregory Nutt. All rights reserved.
+*   Author: Greory Nutt <gnutt@nuttx.org>
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions
+* are met:
+*
+* 1. Redistributions of source code must retain the above copyright
+*    notice, this list of conditions and the following disclaimer.
+* 2. Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in
+*    the documentation and/or other materials provided with the
+*    distribution.
+* 3. Neither the name NuttX nor the names of its contributors may be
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [9ae8636633fee36765a7e35dee1f8ab6958368c4] - Code Rule Check (C++) OK.

framework/include/media/MediaUtils.h
```diff
@@ -15,6 +15,72 @@
  * limitations under the License.
  *
  ******************************************************************/
+/****************************************************************************
```
Taejun-Kwon: please check comment below.

framework/include/media/MediaUtils.h
```diff
@@ -44,6 +110,40 @@ typedef enum audio_type_e {
 	AUDIO_TYPE_FLAC = 5
 } audio_type_t;
 
+/**
```
Taejun-Kwon: Actually we don't need most of things, I think we need only 2 things we can expand later.. please see below
```
in mediautils.h
typedef enum pcm_format_e {
     AUDIO_PCM_BIT_S16_LE = 0,
     AUDIO_PCM_BIT_S32_LE = 1
}

```
@chanijjani Audio manager should check this.
```
audio_manager_result_t set_audio_stream_in(int channels, int sample_rate, int format) {
     config.format = get_audio_pcm_format(format);
      ...
}

int  get_audio_pcm_format(int format)
{
   switch(format) {
    case AUDIO_PCM_BIT_S16_LE :
         return PCM_FORMAT_S16_LE;
    case AUDIO_PCM_BIT_S32_LE :
         return PCM_FORMAT_S32_LE;
     default :
         return AUDIO_MANAGER_INVALID_PARAM;
}

File: framework/include/media/MediaUtils.h [hunk 3]
```diff
@@ -44,6 +110,40 @@ typedef enum audio_type_e {
...
+ * @brief Audio sample format of a PCM.
+ * @details The first letter specifiers whether the sample is signed or unsigned.
+ * The letter 'S' means signed. The letter 'U' means unsigned.
+ * The following number is the amount of bits that the sample occupies in memory.
+ * Following the underscore, specifiers whether the sample is big endian or little endian.
+ * The letters 'LE' mean little endian.
+ * The letters 'BE' mean big endian.
```
sunghan-chang: Here is no information of doxygen as below:
```
  * @details @b #include <media/MediaUtils.h>
  * @since TizenRT v2.0 PRE
```

PR #1591, opened 2018-04-30
Testcase for C++ Library(LLVM Libc++)

manoharahk:
- Port LLVM Test Cases Sources to TizenRT
 - Development of TC on top of Ported LLVM Test Code


apps/examples/testcase/tc_main.c
```diff
@@ -71,6 +71,10 @@
 #endif
 #endif
 
+#ifdef CONFIG_LIBCXX_UTC
```
sunghan-chang: You should put this inside conditional.
```
#ifdef
#else
#ifdef CONFIG_LIBCXX_UTC
..
#endif
#endif
```

apps/examples/testcase/Kconfig
```diff
@@ -32,6 +32,7 @@ source "$APPSDIR/examples/testcase/ta_tc/wifi_manager/itc/Kconfig"
 source "$APPSDIR/examples/testcase/ta_tc/mqtt/utc/Kconfig"
 source "$APPSDIR/examples/testcase/ta_tc/mqtt/itc/Kconfig"
 source "$APPSDIR/examples/testcase/ta_tc/media/utc/Kconfig"
+source "$EXTERNALDIR/libcxx-test/Kconfig"
```
sunghan-chang: Do we need to call this from here?
For external, we don't need to follow our testcase. Testcase is for our os and framework.

apps/examples/testcase/tc_main.c
```diff
@@ -159,6 +167,9 @@ static const tash_cmdlist_t tc_cmds[] = {
 #endif
 #ifdef CONFIG_EXAMPLES_TESTCASE_MEDIA_UTC
 	{"media_utc", utc_media_main, TASH_EXECMD_ASYNC},
+#endif
+#ifdef CONFIG_LIBCXX_UTC
+	{"libcxx_utc", utc_libcxx_main, TASH_EXECMD_ASYNC},
```
sunghan-chang: Please add this by Arphabet order.

apps/examples/testcase/tc_main.c
```diff
@@ -334,6 +345,13 @@ int tc_main(int argc, char *argv[])
 	}
 #endif
 
+#ifdef CONFIG_LIBCXX_UTC
```
sunghan-chang: same as previous

external/libcxx-test/libcxx_tc_common.h
```diff
@@ -0,0 +1,41 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __EXTERNAL_LIBCXX_TEST_LIBCXX_TC_COMMON_H
+#define __EXTERNAL_LIBCXX_TEST_LIBCXX_TC_COMMON_H
+#include "tc_common.h"
```
sunghan-chang: Do you set testcase path to include this without any path?

external/libcxx-test/libcxx_tc_common.h
```diff
@@ -0,0 +1,41 @@
...
+
+/* Custom TC macros */
+
+#define TC_ASSERT_EXPR(expr)	\
```
sunghan-chang: This is same as ```TC_ASSERT```. Why do you define this even though you include "tc_common.h"?

external/libcxx-test/tc_common.h
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __EXTERNAL_LIBCXX_TEST_TC_COMMON_H
```
sunghan-chang: Oh. You copied content from original tc_common.h. But could you let me know the reason?

external/libcxx-test/libcxx_tc_common.h
sunghan-chang: I saw new tc_common.h at end of list. Could you check below query?

sunghan-chang:
@manoharahk I can't review this, because this PR has toooo many files. Could you let me know the file list which you modify or add?

apps/examples/testcase/tc_main.c
manoharahk: Under #ifdef CONFIG_TASH ? . Ok I will add it

apps/examples/testcase/tc_main.c
manoharahk: I think it is already under #ifdef CONFIG_TASH .

external/libcxx-test/libcxx_tc_common.h
manoharahk: It is not same as TC_ASSERT.
It has differed in two ways.
1. Expects only one argument (expression) same as assert. So it is to replace to original assert calls.
2. It returns return value, Because this return value will be used to check TC_ASSERT_FUNC.

external/libcxx-test/tc_common.h
manoharahk: Yes, tc_common.h in not common place. I could include it via <> or <tinyara/*>. If we can directly include , they copying this file is not required.

apps/examples/testcase/tc_main.c
manoharahk: Okay I will reorder

manoharahk: Dear @sunghan-chang,

Current list is not in alphabetical order. Where do you want me to place it? ..

manoharahk:
- external/libcxx-test: Port *   // These are the patches ported form llvm libc++ test to TizenRT with minor modifications.
- external/libcxx-test : Add UTC *   //These are the patches where we modified with TC macros and customized to work withTtizenRT libc++

apps/examples/testcase/tc_main.c
sunghan-chang: Oh, in different change, I reordered them, not yet merged. Ok, don't mind my comment.

manoharahk: Ok

seinfra:
Target : [062216275cb76caad67567854748d3484ae411d8] - Code Rule Check (C++) OK.

seinfra:
Target : [062216275cb76caad67567854748d3484ae411d8] - Code Rule Check OK.

external/libcxx-test/tc_common.h
sunghan-chang: This module is not in testcase. Originally only testcase which is in testcase folder needs tc_common.h. This is an exception of testcase so how about use CFLAGS to include original tc_common.h?

external/libcxx-test/utc_libcxx.c
```diff
@@ -0,0 +1,35 @@
...
+{
+
+#if defined(CONFIG_HAVE_CXXINITIALIZE)
+	up_cxxinitialize();
+#endif
+
+	if (tc_handler(TC_START, "Libc++ TC") == ERROR) {
+		return ERROR;
+	}
+
```
sunghan-chang: Where is main entry of libcxx test? How does it start?

sunghan-chang:
@manoharahk First of all, why do you follow our testcase for libcxx testsuit?

external/libcxx-test/utc_libcxx.c
manoharahk: tc_main will call this

manoharahk:
1. Whole llvm libc++ tests uses "assert()" to check pass/fail. So we added TC Macros to easily replace with and report pass/fail instead of really asserting the code.
2. Only tc_handler(TC_START/END ..) code is used similar to our test case, to keep the test report same similar to other module(similar to C library example)

manoharahk:
About tc_common.h, 
Only tc_handler() prototype is used from that header. 
Instead of adding a apps/example/testcase module CFLAGS, How about just use extern prototype for tc_handler?

external/libcxx-test/tc_common.h
manoharahk: Ok. I will include it using CFLAGS

manoharahk:
Changes:
1. Removed duplicate tc_common.h and included using CFLAGS

seinfra:
Target : [1e6ce963628e8ab307b40df6162d71db98d0b668] - Code Rule Check OK.

seinfra:
Target : [1e6ce963628e8ab307b40df6162d71db98d0b668] - Code Rule Check (C++) OK.

manoharahk:
Changes:
- Re-based to align commit 9249aa57c1a4bd90407ecef810a6fcf6a71610a9

seinfra:
Target : [d3a61443776d936f1bf65233dd1a1fb3c522617a] - Code Rule Check OK.

seinfra:
Target : [d3a61443776d936f1bf65233dd1a1fb3c522617a] - Code Rule Check (C++) OK.

manoharahk:
Changes:
Re based to latest

PR #1602, opened 2018-05-02
media: modify volume tc for media recorder

jsdosa:
modify logic and tc for volume(get,set)


jsdosa:
wrong commit,

PR #1617, opened 2018-05-03
init/os_bringup: Add return value check when creating app init task

jeongchanKim:
WID:5130364 Return value of a function 'task_create' called at os_bringup.c:255 is not checked, but it is usually checked for this function (6/7).


seinfra:
Target : [44518ec19fc94c91debe0e10a30e2be5d6bda90a] - Code Rule Check (C++) OK.

seinfra:
Target : [44518ec19fc94c91debe0e10a30e2be5d6bda90a] - Code Rule Check OK.

PR #180, opened 2017-06-13
Modify a Makefile in the external folder to remove mbedtls folders.

heejin-kim:

PR #1635, opened 2018-05-03
media: Clean duplicated code in audio_manager

Taejun-Kwon:
1. Merge find_in/out_audio_card to one function
  2. Remove useless code
  3. Clean duplicated code to get samplerate & volume at the same time
  4. Add missing configs

### DIFF
--- a/framework/src/media/audio/audio_manager.c
+++ b/framework/src/media/audio/audio_manager.c
@@ -51,6 +51,20 @@
 
 #define AUDIO_DEVICE_MAX_VOLUME 10
 
+#ifndef CONFIG_AUDIO_MAX_INPUT_CARD_NUM
+#define CONFIG_AUDIO_MAX_INPUT_CARD_NUM 2
+#endif
+
+#ifndef CONFIG_AUDIO_MAX_OUTPUT_CARD_NUM
+#define CONFIG_AUDIO_MAX_OUTPUT_CARD_NUM 2
+#endif
+
+#ifndef CONFIG_AUDIO_RESAMPLER_BUFSIZE
+#define CONFIG_AUDIO_RESAMPLER_BUFSIZE 4096
+#endif
+
+#define INVALID_ID -1
+
 /****************************************************************************
  * Private Declarations
  ****************************************************************************/
@@ -65,7 +79,7 @@ enum manufacturer_type_e {
 	AUDIO_CARD_REALTEK = 0
 };
 
-enum audio_device_type_e {
+enum audio_card_type_e {
 	INPUT = 0,
 	OUTPUT
 };
@@ -83,7 +97,7 @@ struct audio_resample_s {
 	uint32_t samprate_types;
 	uint32_t from;
 	uint32_t to;
-	uint16_t *readbuffer; // readbuffer for only pcm_readi
+	uint16_t *readbuffer;		// readbuffer for only pcm_readi
 	float ratio;
 	src_handle_t handle;
 };
@@ -95,7 +109,6 @@ struct audio_card_info_s {
 	struct audio_config_s config;
 	struct pcm *pcm;
 	struct audio_resample_s resample;
-	uint32_t supported_samprates;
 };
 
 struct audio_samprate_map_entry_s {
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [4c79a638f40ae1911ccf46a56b74ffa27fa52b4f] - Code Rule Check (C++) OK.

seinfra:
Target : [4c79a638f40ae1911ccf46a56b74ffa27fa52b4f] - Code Rule Check OK.

framework/src/media/audio/audio_manager.c
```diff
@@ -140,121 +151,72 @@ static const struct audio_samprate_map_entry_s g_audio_samprate_entry[] = {
...
-			g_audio_in_cards[card_id].status = AUDIO_CARD_READY;
-			g_audio_in_cards[card_id].device_id = device_id;
-			g_audio_in_card_num++;
-
-			ret = get_supported_sample_rate(INPUT, g_audio_in_cards[card_id].card_path);
-			ASSERT(ret == AUDIO_MANAGER_SUCCESS);
-
-			medvdbg("Found an input audio card, id=%d, count=%d,  card_path=%s\n", card_id, g_audio_in_card_num,
-					g_audio_in_cards[card_id].card_path);
+		if (card_type == INPUT) {
```
chanijjani: I think this line needs be changed into  "if (type != type_chr) {"

framework/src/media/audio/audio_manager.c
```diff
@@ -140,121 +151,72 @@ static const struct audio_samprate_map_entry_s g_audio_samprate_entry[] = {
...
-			g_audio_out_cards[card_id].status = AUDIO_CARD_READY;
-			g_audio_out_cards[card_id].device_id = device_id;
-			g_audio_out_card_num++;
-
-			ret = get_supported_sample_rate(OUTPUT, g_audio_out_cards[card_id].card_path);
-			ASSERT(ret == AUDIO_MANAGER_SUCCESS);
-
-			medvdbg("Found an output audio card, id=%d, count=%d,  card_path=%s\n", card_id, g_audio_out_card_num,
-					g_audio_out_cards[card_id].card_path);
+		if (type == type_chr) {
```
chanijjani: I think this "if" can be omitted if line 178 is modified as my comment.

framework/src/media/audio/audio_manager.c
```diff
@@ -140,121 +151,72 @@ static const struct audio_samprate_map_entry_s g_audio_samprate_entry[] = {
...
+			active_card_id++;
+			medvdbg("Found an audio card, total card : %d id : %d, card_path : %s\n", active_card_id, card_id, card[card_id].card_path);
+			get_hardware_params(&card[card_id], card_type);
 		}
+
 	}
 
-	if (!g_audio_out_card_num) {
-		meddbg("Wrong g_audio_out_card_num\n");
+	if (active_card_id == INVALID_ID) {
```
chanijjani: Hmm.. I think just "card_id" can be used instead of active_card_id..

framework/src/media/audio/audio_manager.c
```diff
@@ -504,7 +444,7 @@ static int get_avail_audio_in_card_id()
 {
 	int i;
 	for (i = 0; i < CONFIG_AUDIO_MAX_INPUT_CARD_NUM; i++) {
-		if (g_audio_in_cards[i].status == AUDIO_CARD_READY) {
+		if (g_audio_in_cards[i].status != AUDIO_CARD_IDLE) {
```
chanijjani: We also need to set initially cards as "AUDIO_CARD_IDLE" when found in "found_audio_card()".

framework/src/media/audio/audio_manager.c
```diff
@@ -517,7 +457,7 @@ static int get_avail_audio_out_card_id()
 {
 	int i;
 	for (i = 0; i < CONFIG_AUDIO_MAX_OUTPUT_CARD_NUM; i++) {
-		if (g_audio_out_cards[i].status == AUDIO_CARD_READY) {
+		if (g_audio_out_cards[i].status != AUDIO_CARD_IDLE) {
```
chanijjani: Same as the above.

framework/src/media/audio/audio_manager.c
```diff
@@ -934,93 +854,117 @@ audio_manager_result_t reset_audio_stream_out(void)
...
+	return AUDIO_MANAGER_SUCCESS;
+}
+
+static void get_hardware_params(audio_card_info_t *card, audio_card_type_t card_type)
+{
+	int fd;
+	int ret;
+	audio_config_t *config;
+
+	fd = open(card->card_path, O_RDONLY);
```
chanijjani: "card->card_path" needs to be opened with "O_WRONLY" for output cards.

framework/src/media/audio/audio_manager.c
```diff
@@ -140,121 +151,72 @@ static const struct audio_samprate_map_entry_s g_audio_samprate_entry[] = {
...
+	uint8_t card_id = INVALID_ID;
+	uint16_t device_id = INVALID_ID;
 	char type;
 	struct dirent *dir_entry;
 	DIR *dir_info;
 
+	audio_card_info_t *card = g_audio_out_cards;
+	int max_card_num = CONFIG_AUDIO_MAX_OUTPUT_CARD_NUM;
+	char type_chr = 'p';
+	int active_card_id = INVALID_ID;
```
jarvisjang: I think the role of 'active_card_id' is counting found cards.
So, the initial value can be as zero.
And... how about 'activated_card_count' or 'found_cards' for the variable name?

framework/src/media/audio/audio_manager.c
Taejun-Kwon: I don't understand, but type_chr is changed line #180 if card_type is INPUT, initial value is 'c' (see Line #165)

framework/src/media/audio/audio_manager.c
Taejun-Kwon: same as above, I put explanation addtionally,
I modified this function to use both of INPUT & OUTPUT case.
Basically some values initialized for OUTPUT, but if param is INPUT, it changes some values.
so flow is below
1. change values
2. find direnty in audio driver directory
3. if driver name is related to card type, it set values.
4. after find all cards, it check driver founded or not.
5. if founded, decide actual card.

Taejun-Kwon: Also please check it continue if card_id is greater than max_card_num(changed in line #178) or device id is grater than MAX_AUDIO_DEVICE_NUM this can not be added in if state.

framework/src/media/audio/audio_manager.c
Taejun-Kwon: If we set AUDIO_CARD_IDLE in found_audio_card, currently there's should be no active card.
I think we have to establish policy for this, I set AUDIO_CARD_IDLE in found_audio_card as you suggested, and I will put TODO above of this function & put basic logic.

framework/src/media/audio/audio_manager.c
Taejun-Kwon: same here

framework/src/media/audio/audio_manager.c
Taejun-Kwon: I think you confusing file permission and pcm operation.
Virtual File system consist of flash file system and various drivers.
Each drivers also follow file operation, open, read, write.
If flag is O_RDONLY, file system restrict write operation, for driver case, put some values into driver's priv data to change driver.
If flag is O_WRONLY, file system restrict read operation. This is used when user want get data from driver.
Ioctl doesn't call write function so O_RDONLY is correct.
Also please check this getter, not setter I mean even if we should use read or write, read is correct operation for getter

framework/src/media/audio/audio_manager.c
Taejun-Kwon: ok I will change.

seinfra:
Target : [9142d9b16f93b40d9ab63c121e8371b9dc93cc24] - Code Rule Check (C++) OK.

seinfra:
Target : [9142d9b16f93b40d9ab63c121e8371b9dc93cc24] - Code Rule Check OK.

PR #1636, opened 2018-05-04
os/Kconfig.debug: make a dependancy between DEBUG and core module

sunghan-chang:
DEBUG output message is valid only when core module is enabled.
As of now, it shows debug always even though core is disabled.


seinfra:
Target : [e00989f76caa39bb99d510d8c8d5c5427d310f3d] - Code Rule Check OK.

seinfra:
Target : [e00989f76caa39bb99d510d8c8d5c5427d310f3d] - Code Rule Check (C++) OK.

PR #1640, opened 2018-05-04
media: Modify samplerate value for resampler code coverage

jarvisjang:
Most audio card doesn't support 20000hz. So if we deliver the sample rate as 20000hz,
device would open as 22050hz, and resampling logic is working to convert to 22050 from 20000.


seinfra:
Target : [39ffc800f9ce83848dab06c6562167485020bb77] - Code Rule Check OK.

seinfra:
Target : [39ffc800f9ce83848dab06c6562167485020bb77] - Code Rule Check (C++) OK.

PR #1648, opened 2018-05-04
shell: modify cmdtask debug message

lokeshbv:
shell: modify cmdtask debug message

Unnecessary TASH debugging information continued to be output by
default. This patch changes the output level of that information.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -213,7 +213,7 @@ static int tash_launch_cmdtask(TASH_CMD_CALLBACK cb, int argc, char **args)
 	}
 #endif
 
-	printf("Command will be launched with pri (%d), stack size(%d)\n", pri, stack_size);
+	shvdbg("Command will be launched with pri (%d), stack size(%d)\n", pri, stack_size);
 
 	ret = task_create(args[0], pri, stack_size, cb, &args[1]);
 


seinfra:
Target : [4c87839fbed58eaa38f80c0f7036b0af8ac5cdad] - Code Rule Check OK.

seinfra:
Target : [4c87839fbed58eaa38f80c0f7036b0af8ac5cdad] - Code Rule Check (C++) OK.

apps/shell/tash_command.c
```diff
@@ -213,7 +213,7 @@ static int tash_launch_cmdtask(TASH_CMD_CALLBACK cb, int argc, char **args)
 	}
 #endif
 
-	printf("Command will be launched with pri (%d), stack size(%d)\n", pri, stack_size);
+	shdbg("Command will be launched with pri (%d), stack size(%d)\n", pri, stack_size);
```
sunghan-chang: 1. shvdgb is correct because this is an information sentence.
2. If we change this to debug output, when we execute a command which is set to ASYNC, we can not see any result. If it has low priority or it is silrently working, in my opinion, I wonder it whether it is executed or not. 
Is it ok to you?

seinfra:
Target : [92f7e193e77412902e60b23aff7da0f52a1f740e] - Code Rule Check OK.

seinfra:
Target : [92f7e193e77412902e60b23aff7da0f52a1f740e] - Code Rule Check (C++) OK.

apps/shell/tash_command.c
lokeshbv: 1. Yes, it's an informational log message, should be changed to `shvdbg`
2. But this log message doesn't say whether the ASYNC command is executed or not. It's just an info by TASH about launching of command.

sunghan-chang: 2. Right. But we can know that is launched and will be executed. I don't care when it will be scheduled. So, when I execute something, I want to know whether executed or not.

sunghan-chang: 1. not changed. It has ```shdbg``` yet.

sunghan-chang: 2. Because I think you miss-understood what I mean, I left previous comment. If you think that is ok (changed to shvdbg), let's carry on yours.

seinfra:
Target : [84bc1e670eea876dba17d1f8ee97e49c3fae4998] - Code Rule Check (C++) OK.

seinfra:
Target : [84bc1e670eea876dba17d1f8ee97e49c3fae4998] - Code Rule Check OK.

PR #1656, opened 2018-05-08
framework/st_things : separates object attribute handling logic as collection feature.

Hyunjun85:
Regarding smart things resource model, there is no resource which it has object array attribute.
So, the object array attribute logic never used. Now.
But these codes will be able to use when collection resource feature is activated.
In this case, object array's elements will representated a child resource.

### DIFF
--- a/framework/include/st_things/st_things_types.h
+++ b/framework/include/st_things/st_things_types.h
@@ -265,6 +265,7 @@ typedef struct _st_things_representation {
 	*/
 	bool (*get_double_array_value)(struct _st_things_representation *rep, const char *key, double **array, size_t *length);
 
+#ifdef CONFIG_ST_THINGS_COLLECTION
 	/**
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [e960d90dcc6bbb08f44c64ac162c818bd40170de] - Code Rule Check (C++) OK.

seinfra:
Target : [e960d90dcc6bbb08f44c64ac162c818bd40170de] - Code Rule Check OK.

seinfra:
Target : [eabfab7194b6c78e8d7c56f9ba502015ac7fa190] - Code Rule Check OK.

seinfra:
Target : [eabfab7194b6c78e8d7c56f9ba502015ac7fa190] - Code Rule Check (C++) OK.

JeongJunSik:
This is the public api.
Be careful of changes.
Let's think more.

PR #1657, opened 2018-05-08
os/tools/configure.sh: check existence of necessary package to build

sunghan-chang:
To build some configuration set like grpc, st_things and so on,
some packages are needed to build or download.
When we configure configuration set, we can know what is needed.

TizenRT/os/tools$ ./configure.sh artik053/grpc
  Copy build environment files

  CAUTION!! To build artik053/grpc, protoc should be installed
  Find details at https://github.com/Samsung/TizenRT/blob/master/external/protobuf/README.md

  Configuartion is Done!

### DIFF
--- a/os/tools/configure.sh
+++ b/os/tools/configure.sh
@@ -215,7 +215,7 @@ fi
 
 # Okay... Everything looks good.  Setup the configuration
 
-echo "  Copy files"
+echo "  Copy build environment files"
 install -m 644 "${src_makedefs}" "${dest_makedefs}" || \
   { echo "Failed to copy \"${src_makedefs}\"" ; exit 7 ; }
 if [ "X${have_setenv}" = "Xy" ]; then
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [d9a585a0a8a88e746a837b0513e802cf162f8767] - Code Rule Check (C++) OK.

seinfra:
Target : [d9a585a0a8a88e746a837b0513e802cf162f8767] - Code Rule Check OK.

os/tools/configure.sh
```diff
@@ -245,4 +245,45 @@ if [ "X${defappdir}" = "Xy" ]; then
     echo "CONFIG_APPS_DIR=\"$posappdir\"" >> "${dest_config}"
   fi
 fi
+
+# Check mandatory package to configuration set
+IOTIVITY=`grep CONFIG_ENABLE_IOTIVITY=y ${dest_config}`
+if [ ! -z ${IOTIVITY} ]; then
```
manoharahk: Just curious, Is ! -z {var} better that -n {var}  or same?

os/tools/configure.sh
```diff
@@ -245,4 +245,45 @@ if [ "X${defappdir}" = "Xy" ]; then
...
+if [ ! -z ${GRPC} ]; then
+  if ! which protoc >/dev/null; then
+    echo " "
+    echo "  CAUTION!! To build ${boardconfig}, protoc should be installed"
+    echo "  Find details at https://github.com/Samsung/TizenRT/blob/master/external/protobuf/README.md"
+    echo " "
+  fi
+fi
+
 echo "  Configuartion is Done!"
```
manoharahk: Can we please accommodate below change within this patch?
Configuartion -> Configuration ?

os/tools/configure.sh
sunghan-chang: No. same

os/tools/configure.sh
```diff
@@ -245,4 +245,45 @@ if [ "X${defappdir}" = "Xy" ]; then
...
+    echo "  Find details at https://github.com/Samsung/TizenRT/blob/master/docs/HowToUseROMFS.md"
+    echo " "
+  fi
+fi
+
+IOTJS=`grep CONFIG_ENABLE_IOTJS=y ${dest_config}`
+if [ ! -z ${IOTJS} ]; then
+  if ! which cmake >/dev/null; then
+    echo " "
+    echo "  CAUTION!! To build ${boardconfig}, cmake should be installed"
```
manoharahk: How about providing this guide - https://cmake.org/install/?

os/tools/configure.sh
sunghan-chang: Actually that is not in my change. Is it ok to include fixing?

manoharahk: Yeah.. It is not in your change. 
Since it is a small change , I thought we can accommodate in this patch. 
If not we can create a new PR.

sunghan-chang: ok. agree that is a small. I will.

os/tools/configure.sh
sunghan-chang: There is no specific README of iotjs to build it on TizenRT. That's why I only miss it.
But you want to provide a guidance. Right? Then, how about a generic README of iotjs? But it does not show how to do it on TizenRT.
I don't agree adding specific package installlation on this script.

manoharahk: Adding README for iotjs is good option. But if we don't have one then at least  it is better to point to generic cmake installation guide.

sunghan-chang: I added iotjs README.

seinfra:
Target : [3fbc7f496b633eff944acd4c5bd3360dfedec345] - Code Rule Check OK.

PR #1715, opened 2018-05-14
media : add logic for nullptr case in setObserver

jsdosa:
add missing logic for nullptr case in setObserver
1. first, mRecorderObserver is not null case
2. second, observer is not null case


seinfra:
Target : [65c266baf1fc0d975b7ac920350a10ecf8c14929] -

seinfra:
Target : [65c266baf1fc0d975b7ac920350a10ecf8c14929] - Code Rule Check (C++) OK.

PR #1725, opened 2018-05-14
kernel/pthread: Modify comment for pthread_key create and delete APIs

jeongchanKim:
In present implementation, pthread_key_delete and destructor for key deletion are supported.


seinfra:
Target : [2d095e7ba43a12454c0617d0c8520d817557d318] -

PR #1728, opened 2018-05-15
external/iotivity: fix build configurations

pillip8282:
fix that iotivity refers /usr/lib/include/
Iotivity on TizenRT should not refer compiler's header files
And iotivity refered header files that TizenRT doesn't support.
because thoes are in system include path.


seinfra:
Target : [7667e37e44645cfcafb59e2668694b928e4a39f2] - Code Rule Check OK.

seinfra:
Target : [7667e37e44645cfcafb59e2668694b928e4a39f2] - Code Rule Check (C++) OK.

PR #1736, opened 2018-05-15
external/libcxx-test: Update Thread Testcases

yashwanth686007:
Try lock may not always get succeeded.
Hence, few try lock tc's might fail sometimes.
So, few tc's are not required.


seinfra:
Target : [fa2a732dd9a58e9cde2ce381aaf5b32ab29d191a] -

PR #1750, opened 2018-05-17
framework/wifimanager: fix the overflow issue on wifi profile

pillip8282:
when wifi.conf has data more than 200 bytes then it makes crash.
So when it reads the wifi ap information from the storage. it will check
it is in valid range.


seinfra:
Target : [2e1adf05d1e9cf7fd09e22c982e871cefc5f53fd] - Code Rule Check (C++) OK.

seinfra:
Target : [2e1adf05d1e9cf7fd09e22c982e871cefc5f53fd] - Code Rule Check OK.

framework/src/wifi_manager/wifi_profile.c
```diff
@@ -232,18 +238,27 @@ wifi_utils_result_e wifi_profile_write(wifi_manager_ap_config_s *config)
...
+	// file size of wifi.conf should not exceed 107(+EOF)
+	// ssid_length and passphrase length are stored as character.
+	//
+	/////////////////////////////////////////////////////////////////////////////////////////
+	unsigned int readlen = 128;
 	ret = see_read_secure_storage((unsigned char *)buf, &readlen, WIFI_PROFILE_SS_INDEX);
 	if (ret != SEE_OK) {
 		return WIFI_UTILS_FILE_ERROR;
-	} else if (readlen == 1) {
+	} else if (readlen == 1 || readlen > 128) {
```
JeongJunSik: What does readlen == 1 mean?

pillip8282: I think that it is not needed anymore. as I remember, readlen == 1 points that read fail. but it's not valid anymore. so I'll remove it

seinfra:
Target : [d74f88009bd00f267fdacd06a4d2b7ebe53cbcde] - Code Rule Check (C++) OK.

JeongJunSik:
Don't you need to change wifi_profile_write to buf size 256 -> 128 also?

pillip8282:
@JeongJunSik 
Ok it would be better to change both buffer size

seinfra:
Target : [124025c768d515d8346b946d2de757274c55842e] - Code Rule Check (C++) OK.

seinfra:
Target : [124025c768d515d8346b946d2de757274c55842e] - Code Rule Check OK.

PR #1754, opened 2018-05-17
arch: move and rename configs to support additional heap regions

sunghan-chang:
By user decision or hw limitation, heap regions can be separated.
As of now, HEAPx_BASE and HEAPx_SIZE supports that.
But, config of first heap region is RAM_START and RAM_SIZE.
Because they have different naming rules and are at different locations,
it is difficult to know.
Let's move additional heap configurations to arch and rename them.
Because usually architecture determines heap regions.

### DIFF
--- a/os/arch/Kconfig.chip
+++ b/os/arch/Kconfig.chip
@@ -735,6 +735,31 @@ config RAM_SIZE
 		does not execute out of RAM but from FLASH, then you may designate
 		any block of RAM as "primary."
 
+config RAMx_START
+	string "List of start address for additional heap region"
+	default "0x00000000,"
+	depends on MM_REGIONS != 1
+	---help---
+		The start address list of the additional heap regions.
+		When you want to use separated heap region in a physical RAM or
+		two different physical RAMs, set MM_REGIONS first at memory
+		management, and then set hexa values at RAMx_START and RAMx_SIZE
+		to support several heap regions.
+		Each region is separated by ',' as shown in example.
+		i.e. "0x02000000,0x4000000"
+
+config RAMx_SIZE
+	string "List of size for additional heap region"
+	default "0,"
+	depends on MM_REGIONS != 1
+	---help---
+		The size list of additional heap region.
+		Refer RAMx_START content. One different thing is this has
+		decimal values (START has hexa values).
+		This should have same number of list element with RAMx_START's.
+		Each region is separated by ',' as shown in example.
+		i.e. "100,200"
+
 config DDR
 	bool "Use DDR memory"
 	default n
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [3e26a125312fa62bdc80e07571afa68f88c9e24d] - Code Rule Check OK.

seinfra:
Target : [3e26a125312fa62bdc80e07571afa68f88c9e24d] - Code Rule Check (C++) OK.

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -216,8 +216,9 @@ void up_addregion(void)
 	char *mem_size = CONFIG_HEAPx_SIZE;
 
 	for (region_cnt = 0; region_cnt < CONFIG_MM_REGIONS - 1; region_cnt++) {
-		if (*mem_start || *mem_size) {
-			dbg("CONFIG_HEAPx_BASE and CONFIG_HEAPx_SIZE are not defined properly\n");
+		if (!*mem_start || !*mem_size) {
+			dbg("Fail to add %dth heap region\n", region_cnt + 1);
+			break;
 		}
```
girishks2000: If checked outside loop, code will be more readable and avoids abrupt exit from loop.

os/arch/Kconfig.chip
```diff
@@ -735,6 +735,27 @@ config RAM_SIZE
...
+	default "0x00000000,"
+	depends on MM_REGIONS != 1
+	---help---
+		The start address list of the additional heap regions.
+		When you want to use separated heap region in a physical RAM or
+		two different physical RAMs, set MM_REGIONS first at memory
+		management, and then set RAMx_START and RAMx_SIZE to support
+		several heap regions.
+		Commas show each region.
+
```
girishks2000: Give an example of region definition. e.g. {0x20000000, 0x40000000}. This will give better picture and overall purpose of how it can be used. Same for SIZE.  "Commas show each region." This is confusing, Instead mention "Each region is separated by ',' as shown in example."

os/arch/arm/src/common/up_allocateheap.c
sunghan-chang: mem_start and mem_size are updated at every loop. This is a conditional to check valid data. It can not be at outside of loop.

os/arch/Kconfig.chip
sunghan-chang: I will. Thank you.

os/arch/arm/src/common/up_allocateheap.c
girishks2000: OK, understood after viewed the whole function

seinfra:
Target : [f5b0f307247e31944656897d1df0531cb1e26df0] - Code Rule Check (C++) OK.

seinfra:
Target : [f5b0f307247e31944656897d1df0531cb1e26df0] - Code Rule Check OK.

seinfra:
Target : [c69bb68d35eee1dc679ab0b4c62338673eb79ebd] - Code Rule Check OK.

seinfra:
Target : [c69bb68d35eee1dc679ab0b4c62338673eb79ebd] - Code Rule Check (C++) OK.

PR #1758, opened 2018-05-18
audio/resampler: interface update.

jiawei-xiang:
In the struct src_data_s,  should set data_out and out_buf_length by user.
@Taejun-Kwon

### DIFF
--- a/framework/src/media/audio/resample/samplerate.c
+++ b/framework/src/media/audio/resample/samplerate.c
@@ -99,7 +99,7 @@
  */
 struct resampler_s {
 	int16_t *in_buffer;     // internal input buffer
-	int16_t *out_buffer;    // internal output buffer
+	int16_t *out_buffer;    // pointer of the external output buffer
 	int size_in_bytes;      // input buffer size in bytes
 	int size_in_frames;     // input buffer size in frames
 	int outsize_in_frames;  // output buffer size in frames
@@ -402,9 +402,6 @@ int src_destroy(src_handle_t handle)
 	free(src->in_buffer);
 	src->in_buffer = NULL;
 
-	free(src->out_buffer);
-	src->out_buffer = NULL;
-
 	free(src);
 	return SRC_ERR_NO_ERROR;
 }
@@ -422,6 +419,7 @@ int src_simple(src_handle_t handle, src_data_t *src_data)
 {
 	resampler_t *src = (resampler_t *)handle;
 	RETURN_VAL_IF_FAIL((src != NULL && src_data != NULL), SRC_ERR_BAD_PARAMS);
+	RETURN_VAL_IF_FAIL((src_data->data_in != NULL && src_data->data_out != NULL), SRC_ERR_BAD_PARAMS);
 
 	int channels_num = src_data->channels_num;
 	RETURN_VAL_IF_FAIL((channels_num > 0 && channels_num <= SRC_MAX_CH), SRC_ERR_BAD_CHANNEL_COUNT);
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [5af5644373dd2df291a20df8c3846a9d886ed7cf] - Code Rule Check (C++) OK.

seinfra:
Target : [5af5644373dd2df291a20df8c3846a9d886ed7cf] - Code Rule Check OK.

framework/src/media/audio/resample/samplerate.c
```diff
@@ -493,8 +480,15 @@ int src_simple(src_handle_t handle, src_data_t *src_data)
 			   FRAMES_TO_BYTES(src->left_frames));
 		frames_num += src->left_frames;
 
+		//according to the outsize_in_frames, to calculate the adjust_size_in_frames
+		int ratio = new_sample_rate / old_sample_rate;
```
chanijjani: "ratio" is already calculated in line 433.
Couldn't this duplication be removed?

framework/src/media/audio/resample/samplerate.c
```diff
@@ -474,6 +458,9 @@ int src_simple(src_handle_t handle, src_data_t *src_data)
 		RETURN_VAL_IF_FAIL((src->lastnewformat == new_sample_width), SRC_ERR_NOT_SUPPORT);
 	}
 
+	src->out_buffer = src_data->data_out;
+	src->outsize_in_frames = src_data->out_buf_length / (BYTES_PER_SAMPLE(new_sample_width) * channels_num);
```
chanijjani: How about checking the validity of "src_data->out_buf_length" in a proper position?

framework/src/media/audio/resample/samplerate.c
jiawei-xiang: It has some difference.  
In the 433 line,  it calculate the src_ratio, it is a float value. 
And in this modification , ratio  is a int value ,ex: when  origin_sample_rate is 16000,desired_sample_width is 24000,  the src-ratio is 1.5,  but we need get a ratio value is 2.

zhouxinhe: similar with:
int ratio = lrintf(src_ratio + 0.5);
but there's still some difference:
2.0  -> 2  (if use lrintf, it return 3, not our expected result.)
2.1  -> 3  (lrintf return same value)
2.9 ->  3  (lrintf return same value)

framework/src/media/audio/resample/samplerate.c
zhouxinhe: out_buffer and out_buf_length are user input values now, 
resampler record them and convert length bytes to frames, 
it's easy for us to use frame conceptual value :), I think.

zhouxinhe: sure, validation checking is necessary, we will fix it.

seinfra:
Target : [2afc4b940cf331e371b773fb33c21f995b9e02a3] - Code Rule Check (C++) OK.

seinfra:
Target : [2afc4b940cf331e371b773fb33c21f995b9e02a3] - Code Rule Check OK.

framework/src/media/audio/resample/samplerate.c
chanijjani: I see. Thank you for good explanations.

framework/src/media/audio/resample/samplerate.c
```diff
@@ -493,8 +482,15 @@ int src_simple(src_handle_t handle, src_data_t *src_data)
 			   FRAMES_TO_BYTES(src->left_frames));
 		frames_num += src->left_frames;
 
+		//according to the outsize_in_frames, to calculate the adjust_size_in_frames
+		int ratio = new_sample_rate / old_sample_rate;
+		if ((new_sample_rate % old_sample_rate) != 0) {
+			ratio += 1;
+		}
+		int adjust_size_in_frames = MINIMUM(src->size_in_frames, src->outsize_in_frames/ratio);
```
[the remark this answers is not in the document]
jiawei-xiang: About the resample, it is a little complex. My explaination might be a little complex too.
1. when do resample, it must  havesome remaining data, because of the filter mechanism. Need to add these remainning data to the begining of next data_in. so actual required memory size might be bigger than 'src->size_in_frames' . It will affect the next 'src->outsize_in_frames' . It is difficuclt for user to calculate the exact 'src->outsize_in_frames' which we need.
2. Because of the pcm length is not fixed and TizenRT have memory limitation, we add a split mechanism. If size_in_frames is  bigger or 'src->outsize_in_frames'  is  smaller, user can call src_simple serveral times to complete the resample in their while loop.  In the while loop ,it use src->input_frames_used related condition.
3. When need to do resample, it have a src_init(int size).  We use this size to malloc memory and calculate how many frames can resampled.   It is not equal to  'src->size_in_frames' .  If  'src->size_in_frames' is bigger,  will run into the split mechanism.
Now, it is requested to modify the interface, and user malloc data_out memory, so we modify the code to make sure resample can run normally, even if the 'src->outsize_in_frames' is smaller than required.

So,   we can not use 'ratio * src->size_in_bytes / (BYTES_PER_SAMPLE(new_sample_width) * channels_num)'  to check the 'src->outsize_in_frames' .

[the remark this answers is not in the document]
chanijjani: Okay. I'll approve and Taejun kwon may finalize the review of this PR and and merge.

PR #1766, opened 2018-05-21
external/webclient: Modify WGET return value

JeonginKim:
There is an intermittent case where recv is called one more time after receiving all files.
In this case, returning 0 from the recv function is not an error.
If all data is received, the return value is changed to WGET_OK.


seinfra:
Target : [b4cad40d86730fbc86c9922682546498bef96b26] - Code Rule Check (C++) OK.

seinfra:
Target : [b4cad40d86730fbc86c9922682546498bef96b26] - Code Rule Check OK.

PR #1773, opened 2018-05-24
kernel/pthread: re-implement pthread key functions

sunghan-chang:
Because Nuttx's pthread key functions are not POSIX compatible,
re-implement all of pthread key functions,
pthread_key_create, pthread_key_delete, pthread_getspecific and
pthread_setspecific.

To support destructor feature on pthread key, modify pthread_exit
and pthread_cancel function as well.


seinfra:
Target : [0fb24be85e88f52cc9ab9f88231da6b75ef42f76] - Code Rule Check (C++) OK.

seinfra:
Target : [0fb24be85e88f52cc9ab9f88231da6b75ef42f76] - Code Rule Check OK.

seinfra:
Target : [61cda482f406870f587111d3e03c03f638805ff2] -

seinfra:
Target : [c2fd406567947f6c546831bda6c7e4709eb9f9a2] - Code Rule Check (C++) OK.

seinfra:
Target : [c2fd406567947f6c546831bda6c7e4709eb9f9a2] - Code Rule Check OK.

seinfra:
Target : [2d1d2f66b4b3b750c350a8795681636d04216669] - Code Rule Check OK.

seinfra:
Target : [2d1d2f66b4b3b750c350a8795681636d04216669] - Code Rule Check (C++) OK.

seinfra:
Target : [27cf824350a619b0ae62b66733c72ec921b003fc] - Code Rule Check (C++) OK.

seinfra:
Target : [27cf824350a619b0ae62b66733c72ec921b003fc] - Code Rule Check OK.

seinfra:
Target : [0e4a458efee6ed8d713b79dcd94de5525357a6a7] -

seinfra:
Target : [0e4a458efee6ed8d713b79dcd94de5525357a6a7] -

seinfra:
Target : [ee3a82efebe68c9d8ee2c21ed685b4144e16d858] -

seinfra:
Target : [ee3a82efebe68c9d8ee2c21ed685b4144e16d858] -

seinfra:
Target : [0078cd62be48b96a645fd884915173e1478815c9] - Code Rule Check OK.

seinfra:
Target : [0078cd62be48b96a645fd884915173e1478815c9] - Code Rule Check (C++) OK.

os/kernel/pthread/pthread_keydelete.c
```diff
@@ -111,13 +79,36 @@ int pthread_key_delete(pthread_key_t key)
...
+			/* No Key in the list */
+
+			return EINVAL;
+		}
+
+		/* Remove key from list */
+
+		sq_rem((sq_entry_t *)cur_key, &rtcb->key_list);
+
+		/* Free key data */
```
pradeepns85: Releasing the cur_key , without calling the corresponding destructor wouldn't cause issue ?

pradeepns85: Don't we need to reset the cur_key values before calling sched_kfree ?
		cur_key->data = NULL;
		cur_key->destructor = NULL;

os/kernel/pthread/pthread_keydelete.c
```diff
@@ -111,13 +79,36 @@ int pthread_key_delete(pthread_key_t key)
...
 
 	DEBUGASSERT(group && (rtcb->cmn.flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_PTHREAD);
 
-	if (key < PTHREAD_KEYS_MAX && group->tg_keys[key] == IN_USE) {
-		rtcb->pthread_data[key].data = NULL;
-		rtcb->pthread_data[key].destructor = NULL;
-		group->tg_keys[key] = NOT_IN_USE;
+	/* Check key validation */
+
+	if (key < PTHREAD_KEYS_MAX) {
```
pradeepns85: At some places,  key validation is done as
if (group->tg_nkeys == PTHREAD_KEYS_MAX) 

here key validation is done as  if (key < PTHREAD_KEYS_MAX) {

Though logic is correct,  I think it's better to mention in comments that, keys are starting from 0 and go until ( PTHREAD_KEYS_MAX - 1  ) or else it's confusing as same variable (PTHREAD_KEYS_MAX) used to validate 2 conditions

build/configs/artik053/cxxtest/defconfig
```diff
@@ -282,6 +282,7 @@ CONFIG_PTHREAD_MUTEX_TYPES=y
 CONFIG_PTHREAD_MUTEX_UNSAFE=y
 # CONFIG_PTHREAD_MUTEX_BOTH is not set
 CONFIG_NPTHREAD_KEYS=4
+CONFIG_NPTHREAD_DESTRUCTOR_ITERATIONS=4
```
pradeepns85: Should we create a proper dependency in the Kconfig entry ? like if  user selects CONFIG_NPTHREAD_KEYS, it should select CONFIG_NPTHREAD_DESTRUCTOR_ITERATIONS ?

build/configs/qemu/tc_64k/defconfig
```diff
@@ -306,7 +308,7 @@ CONFIG_PREALLOC_TIMERS=4
 #
 CONFIG_INIT_ENTRYPOINT=y
 CONFIG_RR_INTERVAL=200
-CONFIG_TASK_NAME_SIZE=32
+CONFIG_TASK_NAME_SIZE=16
```
pradeepns85: I guess setting to even number makes lot of memory wastage as the code inside OS uses 
include/tinyara/sched.h:588:    char name[CONFIG_TASK_NAME_SIZE + 1]; to  fit into 32 bytes ( 4 words ) 
I request to change it to  15

pradeepns85: And same may apply to other defconfigs as well , kindly reduce to 31

../build/configs/qemu/tc_16m/defconfig:314:CONFIG_TASK_NAME_SIZE=32
../build/configs/qemu/tc_16m_grpc/defconfig:314:CONFIG_TASK_NAME_SIZE=32
../build/configs/qemu/tc_1m/defconfig:314:CONFIG_TASK_NAME_SIZE=32

os/include/tinyara/sched.h
```diff
@@ -309,6 +305,8 @@ struct dspace_s {
 /* struct pthread_key_s **********************************************************/
 #if CONFIG_NPTHREAD_KEYS > 0
 struct pthread_key_s {
+	struct pthread_key_s *flink;
+	int key_index;
```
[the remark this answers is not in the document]
sunghan-chang: I change it ```pthread_key_t key```. This element shows this structure for this "key". So, this value is same as key argument of pthread_key_create. That's why I change it to ```pthread_key_t```. 
And nkeys means number of key. But this is an index of key. We don't need to set same type for them.
Anyway due to your comment, I found it is not appropriate. Thank you.

os/kernel/pthread/pthread.h
```diff
@@ -158,6 +150,11 @@ void pthread_enable_cancel(uint16_t oldstate);
 int pthread_mutexattr_verifytype(int type);
 #endif
 
+#if defined(CONFIG_NPTHREAD_KEYS) && CONFIG_NPTHREAD_KEYS > 0
```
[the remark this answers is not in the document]
sunghan-chang: This file is a header. Another file includes this file regardless of config. When pthead is disabled, CONFIG_NPTHREAD_KEY is not defined. It causes a compilation error.
There are two ways.
1. This change.
2. Remove conditional. Just leave prototype of functions without conditional.

Another prototypes are under conditional. That's why I choose a first way.

os/kernel/pthread/pthread_getspecific.c
```diff
@@ -119,18 +85,28 @@
...
+
+	if (key < PTHREAD_KEYS_MAX) {
+
+		/* Find key data */
+
+		cur_key = (struct pthread_key_s *)pthread_key_find(rtcb, key);
+		if (cur_key == NULL) {
 
-	DEBUGASSERT(group && (rtcb->cmn.flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_PTHREAD);
+			/* No Key in the list */
```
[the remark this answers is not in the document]
sunghan-chang: Done. Thank you.

os/kernel/pthread/pthread_keydelete.c
sunghan-chang: 1. Pthread executes destructor on exit, not delete.
```
An optional destructor function may be associated with each key value. At thread exit, if a key value has a non-NULL destructor pointer, and the thread has a non-NULL value associated with that key, the value of the key is set to NULL, and then the function pointed to is called with the previously associated value as its sole argument.
```
From POSIX, ```at thread exit``` is written.

2. There are two reasons.
    1. User can not acess key->data and key->destructor directly. Pthread uses them internally.
    2. Pthread can acess key structure through key single list. Because pthread remove deleted key from list, set to NULL is unnecessary.

os/kernel/pthread/pthread_keydelete.c
```diff
@@ -111,13 +79,36 @@ int pthread_key_delete(pthread_key_t key)
...
+			return EINVAL;
+		}
+
+		/* Remove key from list */
+
+		sq_rem((sq_entry_t *)cur_key, &rtcb->key_list);
+
+		/* Free key data */
+
+		sched_kfree(cur_key);
```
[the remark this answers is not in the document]
sunghan-chang: No specific reason. Just it is in kernel and a part of scheduler. And sched_kfree is not for interrupt context.
Should I use ```kmm_free```?
But pthread_create uses kmm_zalloc and sched_kfree as well.

os/kernel/pthread/pthread_keydestroy.c
```diff
@@ -0,0 +1,105 @@
...
+	struct task_group_s *group = tcb->cmn.group;
+	struct pthread_key_s *cur_key;
+	struct pthread_key_s *next_key;
+	int destr_count = 0;
+
+	for (cur_key = (struct pthread_key_s *)sq_peek(&tcb->key_list); cur_key; cur_key = next_key) {
+
+		/* Execute destructor until exceeding limitation */
+
+		if (destr_count < PTHREAD_DESTRUCTOR_ITERATIONS) {
```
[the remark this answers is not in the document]
sunghan-chang: That is written in POSIX.
```
If, after all the destructors have been called for all non-NULL values with associated destructors,
there are still some non-NULL values with associated destructors, then the process is repeated.
If, after at least {PTHREAD_DESTRUCTOR_ITERATIONS} iterations of destructor calls for outstanding non-NULL values,
there are still some non-NULL values with associated destructors, implementations may stop calling destructors,
or they may continue calling destructors until no non-NULL values with associated destructors exist,
even though this might result in an infinite loop.
```

os/kernel/pthread/pthread_keydestroy.c
```diff
@@ -0,0 +1,105 @@
...
+				cur_key->destructor(cur_key->data);
+				destr_count++;
+			}
+		}
+
+		next_key = sq_next(cur_key);
+
+		/* Free each key data */
+
+		sched_kfree(cur_key);
```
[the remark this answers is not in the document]
sunghan-chang: Same as above

build/configs/artik053/cxxtest/defconfig
sunghan-chang: Already there it is.
```
config NPTHREAD_DESTRUCTOR_ITERATIONS
	int "Maximum number of calling pthread destructor"
	default 4
	depends on NPTHREAD_KEYS != 0
```

build/configs/qemu/tc_64k/defconfig
sunghan-chang: I think that is a different kind of change. We need to change all of defconfig for that.

seinfra:
Target : [5a21162f19992fc90e78a2a80e5f1b6f12dee2e7] - Code Rule Check OK.

seinfra:
Target : [5a21162f19992fc90e78a2a80e5f1b6f12dee2e7] - Code Rule Check (C++) OK.

build/configs/qemu/tc_64k/defconfig
sunghan-chang: I've contained changes of tc and configs in this PR because of build break with pthread key changes.
I have no plan to add another changes.

os/include/pthread.h
```diff
@@ -204,6 +204,12 @@
 #define _PTHREAD_MFLAGS_INCONSISTENT  (1 << 1)	/* Mutex is in an inconsistent state */
 #define _PTHREAD_MFLAGS_NRECOVERABLE  (1 << 2)	/* Inconsistent mutex has been unlocked */
 
+/*
+ * Maximum values of pthread key operation
+ */
+#define PTHREAD_KEYS_MAX              CONFIG_NPTHREAD_KEYS
+#define PTHREAD_DESTRUCTOR_ITERATIONS CONFIG_NPTHREAD_DESTRUCTOR_ITERATIONS
```
pradeepns85: Is this name should be changed to PTHREAD_DESTRUCTOR_ITERATIONS_MAX ? just like above PTHREAD_KEYS_MAX ?

os/kernel/pthread/pthread_keydelete.c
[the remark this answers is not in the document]
pradeepns85: Ok. I thought sched_kfree needs to be used if we need to release memory from interrupt context by referring to below comment inside kernel/sched/sched.h

201 /* These are lists of dayed memory deallocations that need to be handled
202  * within the IDLE loop or worker thread.  These deallocations get queued
203  * by sched_kufree and sched_kfree() if the OS needs to deallocate memory
204  * while it is within an interrupt handler.
205  */
206
207 extern volatile sq_queue_t g_delayed_kufree;
208
209 #if (defined(CONFIG_BUILD_PROTECTED) || defined(CONFIG_BUILD_KERNEL)) && \
210          defined(CONFIG_MM_KERNEL_HEAP)
211 extern volatile sq_queue_t g_delayed_kfree;
212 #endif

You can ignore this comment if it's not appropriate.

os/include/pthread.h
sunghan-chang: PTHREAD KEY MAX and PTHREAD DESTRUCTOR ITERATIONS are written in POSIX. I just got them from POSIX. You can find them at my previous comment.

sunghan-chang:
@juitem @pradeepns85 @sangwon03 Could you merge this?

os/kernel/pthread/pthread_keydelete.c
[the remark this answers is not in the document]
sunghan-chang: It suppors freeing on interrupt context. But it supports normal freeing as well.
Above content is a description on interrupt context. That dose not mean not using on normal context.
So, no problem. Thank you.

PR #1781, opened 2018-05-25
kernel/environ: Fix an error in the duplication of the child tasks environment

lokeshbv:
Fix an error in the duplication of the child tasks environment in the special case
where the parent's environment was created, but then all of the variables were unset.
In that case, there is still an allocation in place but the size of the allocation is zero.
This case was not being handled correctly when a child task attempts to create its environment
and inherit the zero-size parent environment. Noted by Anthony Merlino.

`This patch is back-ported from Nuttx.`

### DIFF
--- a/os/kernel/environ/env_dup.c
+++ b/os/kernel/environ/env_dup.c
@@ -103,7 +103,7 @@ int env_dup(FAR struct task_group_s *group)
 	size_t envlen;
 	int ret = OK;
 
-	DEBUGASSERT(group && ptcb && ptcb->group);
+	DEBUGASSERT(group != NULL && ptcb != NULL && ptcb->group != NULL);
 
 	/* Pre-emption must be disabled throughout the following because the
 	 * environment may be shared.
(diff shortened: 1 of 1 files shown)


lokeshbv:
But when `envlen = 0`,  it should not return `-ENOMEM`. `envlen = 0` means that zero size parent's environment, which is not an error.

os/kernel/environ/env_dup.c
```diff
@@ -113,18 +113,39 @@ int env_dup(FAR struct task_group_s *group)
...
+			} else {
+				/* Duplicate the parent environment. */
+
+				memcpy(envp, ptcb->group->tg_envp, envlen);
+			}
 		}
+
+		/* Save the size and child environment allocation. */
+
+		group->tg_envsize = envlen;
```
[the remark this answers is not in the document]
lokeshbv: 1. yes, it is redundant when `envlen = 0`, but this is for the case when `envlen != 0`.
2. I don't get why no parent env is a problem. Can you please elaborate the problem.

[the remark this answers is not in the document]
sunghan-chang: 1. No. This line will be executed even if envlen is zero.
2. This function is duplication. Some logic calls this function to duplicate parent env. But regardless of reason, duplication is failed. If it is not a problem, there is no reason to be ENOENT or ENOANO or ENXIO these kinds errno.

seinfra:
Target : [86ca89f52f7cea19a07cabc0304fca99538a43a7] - Code Rule Check (C++) OK.

seinfra:
Target : [86ca89f52f7cea19a07cabc0304fca99538a43a7] - Code Rule Check OK.

os/kernel/environ/env_dup.c
```diff
@@ -113,17 +113,38 @@ int env_dup(FAR struct task_group_s *group)
...
+		if (envlen > 0) {
+			/* There is an environment, duplicate it */
+
+			envp = (FAR char *)kumm_malloc(envlen);
+			if (envp == NULL) {
+				/* The parent's environment can not be inherited due to a
+				 * failure in the allocation of the child environment.
+				 */
+
+				envlen = 0;
```
sunghan-chang: same as previous

os/kernel/environ/env_dup.c
```diff
@@ -113,17 +113,38 @@ int env_dup(FAR struct task_group_s *group)
...
 
 		envlen = ptcb->group->tg_envsize;
-		envp   = (FAR char *)kumm_malloc(envlen);
-		if (!envp) {
-			ret = -ENOMEM;
-		} else {
-			group->tg_envsize = envlen;
-			group->tg_envp = envp;
-			memcpy(envp, ptcb->group->tg_envp, envlen);
+		envp   = NULL;
```
sunghan-chang: redundant. This line is necessary on Nuttx change. Not in here.

lokeshbv: removed

seinfra:
Target : [87b1a941cf16faefc3de4f607533d04958266679] - Code Rule Check (C++) OK.

seinfra:
Target : [87b1a941cf16faefc3de4f607533d04958266679] - Code Rule Check OK.

PR #1794, opened 2018-05-28
apps/builtin: Use shell default values of priority and stacksize

sunghan-chang:
When builtin feature is enabled and priority and stacksize is not
configured in each application, let's use default values of
priority and stacksize which is defined in shell Kconfig instead
of hard-coded in apps/Make.defs.

### DIFF
--- a/apps/Make.defs
+++ b/apps/Make.defs
@@ -51,17 +51,22 @@
 #
 ############################################################################
 
+include $(TOPDIR)/.config
+
 # Builtin Registration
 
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [fe593dd49b5b20dc5191a818958758620a2e9025] - Code Rule Check (C++) OK.

seinfra:
Target : [fe593dd49b5b20dc5191a818958758620a2e9025] - Code Rule Check OK.

PR #1796, opened 2018-05-29
apps/example/iperf: Set priority and stacksize

SeonghoByeon:
- Since Iperf uses more stack size than that of a typical example,
an ASSERT can occur, when the builtin default is changed.


seinfra:
Target : [25ea4c29bc2e8d308e0dca110d093957d6ec2145] - Code Rule Check OK.

seinfra:
Target : [25ea4c29bc2e8d308e0dca110d093957d6ec2145] - Code Rule Check (C++) OK.

PR #1799, opened 2018-05-29
libc++: allocate thread_specific_ptr per task

manoharahk:
There is problem due to static constructor.  current libc++ uses pthread key for thread local storage which is constructed on static storage space which resides in global space.

It works on Linux, because each process will get a separate global symbols, where as in tizenrt global symbols are shared across all application/task(due to flat memory model) and global symbols wont reset even after task exit.              
                                                                                 
Hence in tizenrt case, it fails because of global symbols cannot be reset upon task exit  and even though we tried to reset only selected symbols during exit, concurrent c++ tasks using the same global symbols(tls) would crash.

To avoid this, as a workaround for tizenrt, We have stored the TLS object(key) pointer in the task group on per task basis instead of static object. So each  task would create a separate key and all its 
threads use the same keys.

Also there is one more patch to avoid registration of lot of static destructors which are not useful incase of tizenrt

### DIFF
--- a/external/include/libcxx/thread
+++ b/external/include/libcxx/thread
@@ -159,7 +159,6 @@ class __thread_specific_ptr
      // Only __thread_local_data() may construct a __thread_specific_ptr
      // and only with _Tp == __thread_struct.
     static_assert((is_same<_Tp, __thread_struct>::value), "");
-    __thread_specific_ptr();
     friend _LIBCPP_FUNC_VIS __thread_specific_ptr<__thread_struct>& __thread_local_data();
 
     __thread_specific_ptr(const __thread_specific_ptr&);
@@ -169,6 +168,7 @@ class __thread_specific_ptr
 public:
     typedef _Tp* pointer;
 
+    __thread_specific_ptr();
     ~__thread_specific_ptr();
 
     _LIBCPP_INLINE_VISIBILITY
@@ -213,7 +213,10 @@ __thread_specific_ptr<_Tp>::set_pointer(pointer __p)
 {
     _LIBCPP_ASSERT(get() == nullptr,
                    "Attempting to overwrite thread local data");
-    __libcpp_tls_set(__key_, __p);
+    if (__libcpp_tls_set(__key_, __p) != 0)
+    {
+        delete static_cast<pointer>(__p);
+    }
 }
 
 class _LIBCPP_TYPE_VIS thread;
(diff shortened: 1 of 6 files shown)


seinfra:
Target : [21c4dcb29d889c71d0b4dd77d06ea337308a4ab6] - Code Rule Check OK.

seinfra:
Target : [21c4dcb29d889c71d0b4dd77d06ea337308a4ab6] - Code Rule Check (C++) OK.

sunghan-chang:
I dont understand why not use pthread key function.

manoharahk:
May be title & summary is quite confusing. 
It is not exactly the pthread key, It is an instance of __thread_specific_ptr class which has pthread key as a member. For key creation, it is already using pthread_key_create. 
Please check https://github.com/Samsung/TizenRT/blob/0991ff02609e3788dd803a228bed4960931cd51b/external/libcxx/thread.cxx#L163

external/include/libcxx/thread
```diff
@@ -400,6 +420,15 @@ template <class _Fp>
 void* __thread_proxy_cxx03(void* __vp)
 {
     std::unique_ptr<_Fp> __p(static_cast<_Fp*>(__vp));
+#if defined(__TINYARA__)
```
pradeepns85: Is it reqd to keep redundant code ? can we keep this logic in another func and call from both apis?

manoharahk: Ok. I will add a new function for common code.

seinfra:
Target : [596918ad4810d19aa6e4d2f011dff91033685f79] - Code Rule Check OK.

seinfra:
Target : [596918ad4810d19aa6e4d2f011dff91033685f79] - Code Rule Check (C++) OK.

sunghan-chang:
@manoharahk 
1. Please check your commit description. There are many typos.
2. Is it enough to save per a group? not per a thread? Does all of threads which is in same group use a same pointer?

seinfra:
Target : [7a6656c1087b4fcbe85d2e96d5624633dc50766e] - Code Rule Check OK.

seinfra:
Target : [c3aecca9ed4f4cdb496230e7000ff51596df1297] - Code Rule Check (C++) OK.

seinfra:
Target : [c3aecca9ed4f4cdb496230e7000ff51596df1297] - Code Rule Check OK.

seinfra:
Target : [c3aecca9ed4f4cdb496230e7000ff51596df1297] - Code Rule Check OK.

seinfra:
Target : [c3aecca9ed4f4cdb496230e7000ff51596df1297] - Code Rule Check (C++) OK.

seinfra:
Target : [c3aecca9ed4f4cdb496230e7000ff51596df1297] - Code Rule Check (C++) OK.

PR #1800, opened 2018-05-29
framework/wifi_manager: Unification of SSID length 

SeonghoByeon:
- Add WIFIMGR's string length definition.
- AP configuration structure should have a SSID with a length of WIFIMGR_SSID_LEN.
Some of them have a length of 33, which causes SVACE criticals.

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test.c
+++ b/apps/examples/wifi_manager_sample/wm_test.c
@@ -399,11 +399,13 @@ void wm_set_info(void *arg)
 	WM_TEST_LOG_START;
 	struct options *ap_info = (struct options *)arg;
 	wifi_manager_ap_config_s apconfig;
-	strncpy(apconfig.ssid, ap_info->ssid, 33);
+	strncpy(apconfig.ssid, ap_info->ssid, WIFIMGR_SSID_LEN-1);
 	apconfig.ssid_length = strlen(ap_info->ssid);
+	apconfig.ssid[WIFIMGR_SSID_LEN-1] = '\0';
 	apconfig.ap_auth_type = ap_info->auth_type;
 	if (apconfig.ap_auth_type != WIFI_MANAGER_AUTH_OPEN) {
-		strncpy(apconfig.passphrase, ap_info->password, 64);
+		strncpy(apconfig.passphrase, ap_info->password, WIFIMGR_PASSPHRASE_LEN-1);
+		apconfig.passphrase[WIFIMGR_PASSPHRASE_LEN-1] = '\0';
 		apconfig.passphrase_length = strlen(ap_info->password);
 		apconfig.ap_crypto_type = ap_info->crypto_type;
 	}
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [d8769200ff7b3e12e0efeb8968733b7fe6cad093] - Code Rule Check OK.

seinfra:
Target : [d8769200ff7b3e12e0efeb8968733b7fe6cad093] - Code Rule Check (C++) OK.

seinfra:
Target : [ec9e38a13ddd4daf942aaa476a3488ac2d0234b8] - Code Rule Check (C++) OK.

seinfra:
Target : [ec9e38a13ddd4daf942aaa476a3488ac2d0234b8] - Code Rule Check OK.

apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -623,17 +623,17 @@ void wm_auto_test(void *arg)
...
 	softap_config.ssid[strlen(info->ssid)] = '\0';
-	strncpy(softap_config.passphrase, info->softap_password, 64);
+	strncpy(softap_config.passphrase, info->softap_password, WIFIMGR_PASSPHRASE_LEN);
 	softap_config.passphrase[strlen(info->password)] = '\0';
 	softap_config.channel = 1;
 
 	/* Set AP Configuration */
 	wifi_manager_ap_config_s ap_config;
-	strncpy(ap_config.ssid, info->ssid, 33);
+	strncpy(ap_config.ssid, info->ssid, WIFIMGR_SSID_LEN);
```
sunghan-chang: Even though ssid element has ```WIFIMGR_SSID_LEN``` size, you copy it with same size. Is it ok not saving ```'\0'``` at end of array?
Usaually in this case, set array size as ```XXX + 1``` and copy it with ```XXX``` size.

SeonghoByeon: Hi @sunghan-chang, Basically, the length of SSID should be less than WIFIMGR_SSID_LEN. If the SSID entered by the user is shorter than WIFIMGR_SSID_LEN, there is no need to insert a null because the SSID variable in 'info' structure ends with null. In case that a user sets SSID of AP (or SoftAP) longer than WIFIMGR_SSID_LEN, WI-Fi Manager will return 'fail' due to length limit, that is, there is no Wi-Fi connection and no SoftAP mode using the SSID longer than WIFIMGR_SSID_LEN.
For safety reasons, if NULL is inserted at the end, only SSID of the cutted 31 bytes will be utilized when the user enters a long SSID. Which way could be a better example? We might use strcpy, instead of strncpy.

SeonghoByeon: @sunghan-chang Modification completed.

seinfra:
Target : [4d497f0de0e73f308d2a54bcd0ec304104cebd8a] - Code Rule Check OK.

seinfra:
Target : [4d497f0de0e73f308d2a54bcd0ec304104cebd8a] - Code Rule Check (C++) OK.

PR #1810, opened 2018-05-31
init : do up_cxxinitialize during application init

manoharahk:
up_cxxinitialize is to perform initialization of the
static C++ class instances. All the static initializers
are put in init section(sinit <--> _einit) during build time
and before application start all these initializers must be
invoked by parsing init section.

But in case of Flat address space, all application's initializers are
put into init section of a single blob.

Hence up_cxxinitialize need not to be performed in each application start,
instead it can invoked during application init in os bringup.

This patch makes up_cxxinitialize to perfrom initialization only once and
any subsequent invokations, are ingored.

### DIFF
--- a/apps/platform/gnu/gnu_cxxinitialize.c
+++ b/apps/platform/gnu/gnu_cxxinitialize.c
@@ -123,6 +123,14 @@ extern uint32_t _etext;
 void up_cxxinitialize(void)
 {
 	initializer_t *initp;
+#if !defined(CONFIG_BUILD_KERNEL)
+	static bool cxxinitialized = false;
+
+	if (cxxinitialized)
+		return;
+
+	cxxinitialized = true;
+#endif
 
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [0bb514c8e37b6d968d06d9bcf1a4d6d214fdbb17] - Code Rule Check (C++) OK.

seinfra:
Target : [0bb514c8e37b6d968d06d9bcf1a4d6d214fdbb17] - Code Rule Check OK.

manoharahk:
up_cxxintialize() are required in the application in case of BUILD_KERNEL mode, where each app has to do it explicitly. To make it common it should be called in user mode stub function( task_startup : libc/sched/task_startup.c). But i am not sure how to decide c application or c++ application. So i intentionally left till it gets cleared

manoharahk:
It is taken care here,
https://github.com/Samsung/TizenRT/blob/1ba7aa083736a3024dc8ad36a3bedf69de24d864/apps/platform/gnu/Kconfig#L9

PR #1815, opened 2018-06-01
fs/stdio : change return value as same as close if flush returns -ENOSPC

Taejun-Kwon:
If flush failed during fclose, it doesn't return close's result.
But if disk is full, there's no way write contents anymore, so
File should be closed


seinfra:
Target : [760749f6cc3234a5bd1096a8052a3537b3fd79f1] - Code Rule Check (C++) OK.

seinfra:
Target : [760749f6cc3234a5bd1096a8052a3537b3fd79f1] - Code Rule Check OK.

PR #1827, opened 2018-06-04
init : call cxxinit initializers from idle thread

manoharahk:
Previous commit which did cxxinitialize during os_do_appstart()
would cause crash in few scenarios.

If a task does cxxinitialize and creates a new c++ application task
and exits immediately then in newly created C++ task if iostream are used,
the task would crash.
It is because during cxxinitialize iostreams are initialized (ios_base::Init)
and during iostream initialization they refer console streams
(Ex: stdout, which are defined per task group: group->tg_streamlist)
and set those pointers in iostreams internal file pointers members.
Once the task which did cxxinitialize exits, it would destroy the streams as well
Hence if the newly created c++ task uses any iostreams(Ex: cout) would crash.

os_do_appstart can run as init kthread and would exit after application launch
and the above scenario would occur.

Hence this patch moves cxxinitialize to idle thread which never exits.

### DIFF
--- a/os/kernel/init/os_bringup.c
+++ b/os/kernel/init/os_bringup.c
@@ -250,9 +250,6 @@ static inline void os_do_appstart(void)
 
 	svdbg("Starting application init thread\n");
 
-#if !defined(CONFIG_BUILD_KERNEL) && defined(CONFIG_HAVE_CXXINITIALIZE)
-	up_cxxinitialize();
-#endif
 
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [195698f047e2464a5a9b809c75b2340298c79408] - Code Rule Check (C++) OK.

seinfra:
Target : [195698f047e2464a5a9b809c75b2340298c79408] - Code Rule Check OK.

PR #1830, opened 2018-06-04
Add read/write APIs to testcase driver

kishore-sn:
Add read/write APIs to testcase driver so that UTC apps can
open /dev/testcase to pass ioctls

### DIFF
--- a/os/drivers/testcase/testcase_drv.c
+++ b/os/drivers/testcase/testcase_drv.c
@@ -30,7 +30,8 @@
  ****************************************************************************/
 
 static int testdrv_ioctl(FAR struct file *filep, int cmd, unsigned long arg);
-
+static ssize_t testdrv_read(FAR struct file *filep, FAR char *buffer, size_t len);
+static ssize_t testdrv_write(FAR struct file *filep, FAR const char *buffer, size_t len);
 /****************************************************************************
  * Private Data
  ****************************************************************************/
@@ -38,8 +39,8 @@ static int testdrv_ioctl(FAR struct file *filep, int cmd, unsigned long arg);
 static const struct file_operations testdrv_fops = {
 	0,							/* open */
 	0,							/* close */
-	0,							/* read */
-	0,							/* write */
+	testdrv_read,						/* read */
+	testdrv_write,						/* write */
 	0,							/* seek */
 	testdrv_ioctl						/* ioctl */
 #ifndef CONFIG_DISABLE_POLL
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [f8c51390b2884cd42d4d8e87fc75555c4b782f0d] - Code Rule Check (C++) OK.

seinfra:
Target : [f8c51390b2884cd42d4d8e87fc75555c4b782f0d] - Code Rule Check Result:
os/drivers/testcase/testcase_drv.c:93: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/drivers/testcase/testcase_drv.c:93: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/drivers/testcase/testcase_drv.c:98: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/drivers/testcase/testcase_drv.c:98: ERROR: [IDT_M_TAB] please, no spaces at the start of a line

sunghan-chang:
@kishore-sn I don't think this should be merged now. When you implement function inside, I can merge it.

kishore-sn:
@sunghan-chang 
Power team is implementing testcases for os/pm using this method. 
So they need the testcase driver.
I have raised a separate PR for this since we did not want it to be part of the Power PR.

seinfra:
Target : [d21edb4064a7d4ebbe43b102af4fd04ecbea4b3b] - Code Rule Check OK.

seinfra:
Target : [d21edb4064a7d4ebbe43b102af4fd04ecbea4b3b] - Code Rule Check (C++) OK.

sunghan-chang:
@kishore-sn Is it enough code to work (to use) testcase driver? return 0 and return len?

kishore-sn:
Yes. 
We will never read or write any data /dev/testcase file. 
We will only pass ioctl to indicate which testcase is to be run. 
So, I have added dummy read / write API to allow the app to open the /dev/testcase with proper permissions like O_RDONLY.

sunghan-chang:
@kishore-sn I am sorry, but I don't understand.
```
We will never read or write any data /dev/testcase file.
We will only pass ioctl to indicate which testcase is to be run.
```
Then, why do you need to merge this?

kishore-sn:
Before this change, the file properties are as below:

TASH>>ls -l /dev/testcase
 crw-rw-rw-       0 /dev/testcase


After adding these API, it is changed to :

kishore-sn:
Before this change, the file properties are as below

PR #1838, opened 2018-06-04
network/tc: Domain, type, and protocol matching for socket API

SeonghoByeon:
- Remove socket protocols (and tc) not supported by lwip
- Provide a proper socket_argument_validation function to match
  domain, type, and protocol correctly
- Modify tc contents (positive/negative) according to the socket protocol


seinfra:
Target : [829da1eb50836162032a0b0848d7fab4d7831345] - Code Rule Check Result:
os/net/socket/bsd_socket_api.c:176: ERROR: [IDT_M_SCH] switch and case should be at the same indent

seinfra:
Target : [829da1eb50836162032a0b0848d7fab4d7831345] - Code Rule Check (C++) OK.

PR #1846, opened 2018-06-05
kernel/task: Move calling atexit and onexit for supporting them in ta…

jeongchanKim:
…sk_restart

If moving them to task_recover, axexit and onexit can be called when restart the task.
Then, we can use atexit or onexit for resource deallocation callback when restart the task.


seinfra:
Target : [99d2c60686a8d6e342c6596472d84699b91ff3a2] - Code Rule Check (C++) OK.

seinfra:
Target : [99d2c60686a8d6e342c6596472d84699b91ff3a2] - Code Rule Check OK.

PR #1852, opened 2018-06-07
kernel/task: Add calling atexit and onexit in task_restart

jeongchanKim:
If atexit and onexit are in task restart, we can hook function when restart the task, such as resource deallocation callback

### DIFF
--- a/os/kernel/task/task_restart.c
+++ b/os/kernel/task/task_restart.c
@@ -58,6 +58,9 @@
 
 #include <sys/types.h>
 #include <sched.h>
+#if defined(CONFIG_SCHED_ATEXIT) || defined(CONFIG_SCHED_ONEXIT)
+#include <stdlib.h>
+#endif
 #include <errno.h>
 
 #include <tinyara/arch.h>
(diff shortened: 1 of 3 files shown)


os/kernel/task/task_restart.c
```diff
@@ -164,6 +167,18 @@ int task_restart(pid_t pid)
 			return ERROR;
 		}
 
+#if defined(CONFIG_SCHED_ATEXIT) || defined(CONFIG_SCHED_ONEXIT)
+		/* If exit function(s) were registered, call them now before we do any un-
+		 * initialization.
+		 */
+
+		task_atexit((FAR struct tcb_s *)tcb);
```
sunghan-chang: tcb is defined as struct tcb_s *. I think we don't need to type cast.

seinfra:
Target : [a528b9754676068600f304152fc448a38c692d68] - Code Rule Check OK.

seinfra:
Target : [a528b9754676068600f304152fc448a38c692d68] - Code Rule Check (C++) OK.

os/kernel/task/task_restart.c
jeongchanKim: no. it is  task_tcb_s type. so casting is needed.

seinfra:
Target : [332f769e395ea6addc3aa7a1be3e9fbb8a7a63e5] - Code Rule Check (C++) OK.

seinfra:
Target : [332f769e395ea6addc3aa7a1be3e9fbb8a7a63e5] - Code Rule Check OK.

sunghan-chang:
@girishks2000 @manoharahk Could you review this? When we call restart, it can cause memory leak. That is same as task termination. But at termination, atexit can be called and it can free allocated memory whenever task is terminated. 
As Nuttx wrote that restart is terminate and start again, and there is no side effect and restart is not a POSIX-compliant already, we can execute atexit at restart case also.
Anyway, our purpose of this PR is providing a way to release allocated memory on restart case.

PR #1877, opened 2018-06-21
Fix problem with up_progmem_ispageerased on s5j

kimusan:
The implementation of the up_progmem_ispageerased() function checked for changes in 4 pages instead of just one. This could return false results. It also checks against the wrong erased flash value (correct is all 0xFFFFFFFF).

### DIFF
--- a/os/arch/arm/src/s5j/s5j_sflash.c
+++ b/os/arch/arm/src/s5j/s5j_sflash.c
@@ -149,8 +149,8 @@ ssize_t up_progmem_ispageerased(size_t page)
 
 	bwritten = 0;
 	addr = up_progmem_getaddress(page);
-	for (count = up_progmem_pagesize(page); count; count--) {
-		if (getreg32(addr) != 0xffffff) {
+	for (count = (up_progmem_pagesize(page)/sizeof(int)); count; count--) {
+		if (getreg32(addr) != 0xffffffff) {
 			bwritten++;
 		}
 		addr += sizeof(int);


seinfra:
Target : [e4999ec16b4db082755fd97b58f53a2379c9d5c1] - Code Rule Check (C++) OK.

seinfra:
Target : [e4999ec16b4db082755fd97b58f53a2379c9d5c1] - Code Rule Check OK.

os/arch/arm/src/s5j/s5j_sflash.c
```diff
@@ -149,8 +149,8 @@ ssize_t up_progmem_ispageerased(size_t page)
 
 	bwritten = 0;
 	addr = up_progmem_getaddress(page);
-	for (count = up_progmem_pagesize(page); count; count--) {
-		if (getreg32(addr) != 0xffffff) {
+	for (count = (up_progmem_pagesize(page)/sizeof(int)); count; count--) {
```
Taejun-Kwon: Please check below 2 things.
1. Please use 'number_of_bit(uint32_t) >> 3' instead of sizeof(int)
2. Your code will check every 4bytes, then count should be considered. count += (number_of_bit(uint32_t) >> 3)

kimusan: 1) I just used sizeof() to match the other code in the file. If it should be made differently then I think you should look through the entire code base as sizeof(int) (or similar) is used in quite a few places (especially like malloc(sizeof(uint8_t *4))) 

2) counting +=4 or dividing the count with 4 gives the same result.  since up_progmem_pagesize returns a static number and sizeof(int) is also static, the compiler will optimize this away and put a static value instead. 

I can change it but really dont see any good reasons for doing so.

PR #1880, opened 2018-06-22
external/libcxx-test/itc : add ITC testcase for llbm module

manoj11588612:
New testcase's :
1.itc_libcxx_algorithms_alg_min_max_max -> To check the max functionality
2.itc_libcxx_algorithms_alg_min_max_max_comp -> To check max between two integers in different scenarios of comparison using greater
3.itc_libcxx_containers_associative_map_access -> To check access functionality in different scenarios for map
4.itc_libcxx_containers_associative_map_cons -> To check cons based scenarios in a map
5.itc_libcxx_containers_associative_map_modifiers -> To check scenarios based on modification in a map like if you erase an element
6.itc_libcxx_containers_multimap_cons -> To check cons based scenarios in a multimap
7.itc_libcxx_containers_multimap_modifier -> To check scenarios based on modification in a map like if you erase an element
8.itc_libcxx_containers_multimap_ops -> To check scenarios for associative multimap after insertion with multiple same keys having distict value pair

Signed-off-by: manoj <manoj.g2@samsung.com>


manoharahk:
What is that extra thing covered in "itc*" ?. I could see the same llvm test sources are used.

manoj11588612:
@manoharahk 
In ITC we tried to create scenario Testcase. Likw with mixing of multiple API's we created Testcase.
Unlike UTC verifying single operation of libcxx library.
please review

manoharahk:
I dont see any logical additions to ITC source files except naming changes. The same tc sources are copied and renamed.  If that is case instead of copying why dont we use the same source even for ITC?. It avoids the duplicated code.

manoj11588612:
@manoharahk 
Have you gone through all testcase.
AS i again recheck for few Testcase we just change input value and for few Testcase like for associative module we even combined multiple scenario in one Testcase. Please check.

manoharahk:
yes.
associative/map/map.access/ are combined and input values are changed(algorithms/max_* )
1. Why input values are changed?, Was the previous values are wrong? Because it was picked directly from llvm test source. Reason is, Is there a scope to avoid duplicate code.
2. Are these map access test cases  are just combined and modified arguments OR any new test snippet added? If any new test snippet added please point me to then location.

sunghan-chang:
@manoj11588612 As I left comment on other itc PR,
1. Please share test result
2. Please share what is your scenario and what is difference against utc.

seinfra:
Target : [68ee347f96a0e7b6f8ecde115a26403059f1bca2] - Code Rule Check OK.

seinfra:
Target : [68ee347f96a0e7b6f8ecde115a26403059f1bca2] - Code Rule Check (C++) OK.

samegautam:
@manoj11588612 @bipin4321  :  please update the status of PR.

PR #1902, opened 2018-07-03
example/protobuf_enc_itc : add ITC for proto buf

manoj11588612:
New Testcase's:
1.ITCNegTestUint32f -> this as a negative testcase in which we are giving float input for comparing with 32 bit unsigned integer value
2.ITCNegTestBool -> this as a negative testcase in which we are giving float input for comparing with bool value
3.ITCNegTestInt32 -> this as a negative testcase in which we are giving float input for comparing with 32 bit  integer value
4.ITCNegTestInt64 -> this as a negative testcase in which we are giving float input for comparing with 64 bit integer value
5.ITCNegTestUint32 -> this as a negative testcase in which we are giving signed input for comparing with 32 bit unsigned integer value

Signed-off-by: manoj.g2 <manoj.g2@samsung.com>


seinfra:
Target : [feb9972a6e8a166783e00eecf7ca28d8aae14302] - Code Rule Check (C++) OK.

manoj11588612:
@pillip8282 please review

pillip8282:
@manoj11588612  This ITC is not different to prototest_main. So I think the ITC just duplicate the test cases.

manoj11588612:
@pillip8282 
we developed 5 testcase  with negative scenario
These are not similar to ITC
1.ITCNegTestUint32f -> this as a negative testcase in which we are giving float input for comparing with 32 bit unsigned integer value
 2.ITCNegTestBool -> this as a negative testcase in which we are giving float input for comparing with bool value
 3.ITCNegTestInt32 -> this as a negative testcase in which we are giving float input for comparing with 32 bit integer value
 4.ITCNegTestInt64 -> this as a negative testcase in which we are giving float input for comparing with 64 bit integer value
 5.ITCNegTestUint32 -> this as a negative testcase in which we are giving signed input for comparing with 32 bit unsigned integer value

pillip8282:
@manoj11588612 
I think those negative cases are similar. I think they're negative UTCs. I think that template of ITC and UTC should be different

PR #1904, opened 2018-07-04
framework/st_things: Fix bug for mnid, vid value

webispy:
This patch fix mnid and vid setting bugs. When loading device
information from device_def.json, the mnid and vid are not stored to
variables due to invalid NULL check logic.

For example, if mnid is not NULL, copy mnid->valuestring to
g_device->mnid is not work. The vid also same.

Signed-off-by: Inho Oh <webispy@gmail.com>


seinfra:
Target : [00977fcc82de3d8d19891133217a19394eaa5712] - Code Rule Check OK.

seinfra:
Target : [00977fcc82de3d8d19891133217a19394eaa5712] - Code Rule Check (C++) OK.

PR #1912, opened 2018-07-09
Static object reset & handle atexit/onexit registration dynamically

manoharahk:
1. Reset static object after destruction, to allow object construnction again.
2. There is a limitations in the current implementation that
    only a single atexit/onexit function can be registered unless
    CONFIG_SCHED_ATEXIT_MAX/CONFIG_SCHED_ONEXIT_MAX/ defines a
    larger number. To overcome this limitation this patch removes the support of static
    array of callbacks and keeps the callbacks in the dynamically allocated
    space at runtime using dqueue.

### DIFF
--- a/os/kernel/Kconfig
+++ b/os/kernel/Kconfig
@@ -727,38 +727,11 @@ config SCHED_ATEXIT
 	---help---
 		Enables the atexit() API
 
-config SCHED_ATEXIT_MAX
-	int "Max number of atexit() functions"
-	default 1
-	depends on SCHED_ATEXIT && !SCHED_ONEXIT
-	---help---
-		By default if SCHED_ATEXIT is selected, only a single atexit() function
-		is supported. That number can be increased by defined this setting to
-		the number that you require.
-
-		If both SCHED_ONEXIT and SCHED_ATEXIT are selected, then atexit() is built
-		on top of the on_exit() implementation.  In that case, SCHED_ONEXIT_MAX
-		determines the size of the combined number of atexit(0) and on_exit calls
-		and SCHED_ATEXIT_MAX is not used.
-
 config SCHED_ONEXIT
 	bool "Enable on_exit() API"
 	default n
 	---help---
 		Enables the on_exit() API
-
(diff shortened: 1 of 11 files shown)


seinfra:
Target : [d4904623dc08d834818671b44c9f14c8b9edc93b] - Code Rule Check OK.

seinfra:
Target : [d4904623dc08d834818671b44c9f14c8b9edc93b] - Code Rule Check (C++) OK.

seinfra:
Target : [02fce54381ca8c728054de2f2e1a1fa09a392619] - Code Rule Check OK.

seinfra:
Target : [02fce54381ca8c728054de2f2e1a1fa09a392619] - Code Rule Check (C++) OK.

os/kernel/Kconfig
```diff
@@ -727,39 +727,12 @@ config SCHED_ATEXIT
 	---help---
 		Enables the atexit() API
 
-config SCHED_ATEXIT_MAX
```
sunghan-chang: In POSIX description of atexit, below is written.
```
The application should call sysconf() to obtain the value of {ATEXIT_MAX},
the number of functions that can be registered.
There is no way for an application to tell
 how many functions have already been registered with atexit().
```

Do we need to remove this?

manoharahk: 1. I dont think we have sysconf() support.
2. SCHED_ATEXIT_MAX  was used for allocating fixed size atexit callback function array,  since we planning to use linked list, we dont need it anymore.
Do you see any problem with this approach?

samirsahu1947: Just FYI .
This fix is related to media_utc memory leak and repeated execution failure.

1. Memory Leak:
We have two sub issues,
1. Leak in libc++
2. Leak due to inode_alloc. This would help to free the memory allocated by inode.

2. Repeated execution of media_utc doesn’t work:
It is due to static object issue. If you run media_utc for the second time, it won’t work because static objects are not getting constructed for the second and onward case.

Detailed history can be found at WCTIAPP-2707

PR #1920, opened 2018-07-11
external/webserver: Fix strncat() with calculating the remaining length

JeonginKim:
strncat was being used without calculating the remaining length of the destination string buffer.
Prevent buffer overflow by calculating the remaining length of the destination string buffer.


seinfra:
Target : [5a71a88eb9ed66633a8d4379f9fe7c77ddf266ab] - Code Rule Check (C++) OK.

seinfra:
Target : [5a71a88eb9ed66633a8d4379f9fe7c77ddf266ab] - Code Rule Check OK.

PR #1928, opened 2018-07-12
framework/st_things] fixes a minor bug what it doesn't stop ntp daemon.

Hyunjun85:
It has a bug what it does not stop ntp daemon when the things was disconnected from wifi.

### DIFF
--- a/framework/src/st_things/things_stack/utils/things_network.c
+++ b/framework/src/st_things/things_stack/utils/things_network.c
@@ -304,6 +304,9 @@ void things_wifi_sta_connected(wifi_manager_result_e res)
 void things_wifi_sta_disconnected(void)
 {
 	THINGS_LOG_V(TAG, "T%d --> %s", getpid(), __FUNCTION__);
+#ifdef CONFIG_NETUTILS_NTPCLIENT
+	ntpc_stop();
+#endif
 	things_wifi_changed_call_func(0, NULL, NULL);
 }
 


seinfra:
Target : [6b27d13bc4725a8ba3ed927968691eacc7141255] - Code Rule Check OK.

seinfra:
Target : [6b27d13bc4725a8ba3ed927968691eacc7141255] - Code Rule Check (C++) OK.

framework/src/st_things/things_stack/utils/things_network.c
```diff
@@ -304,6 +304,9 @@ void things_wifi_sta_connected(wifi_manager_result_e res)
 void things_wifi_sta_disconnected(void)
 {
 	THINGS_LOG_V(TAG, "T%d --> %s", getpid(), __FUNCTION__);
+#ifdef CONFIG_NETUTILS_NTPCLIENT
```
Taejun-Kwon: We don't have to check CONFIG_DISABLE_SIGNALS??

JeongJunSik:
I do not think I need to call ntpc_stop every disconnect.
Ntp is a Deamon Start Trigger only, it takes more overhead when it comes down during frequent connections and uploads.
Would not it be right to call ntpc_stop from things_stop?

framework/src/st_things/things_stack/utils/things_network.c
JeongJunSik: @Taejun-Kwon  
Good Point. 
I found some bugs in ntpclient about this MACRO.
i fixed it at PR #1930 
Thanks.

JeongJunSik:
Please chage commit title "]" to ":"

framework/src/st_things/things_stack/utils/things_network.c
Hyunjun85: @Taejun-Kwon, 
We aready define and use CONFIG_NETUTILS_NTPCLIENT in things sdk layer.
Can use CONFIG_DISABLE_SIGNALS, but user is able to feel any strange things.

Why are related functions binded each different definitions CONFIG_NETUTILS_NTPCLIENT and CONFIG_DISABLE_SIGNALS?

Can call sync_time_from_ntp when defined CONFIG_NETUTILS_NTPCLIENT?
Can call ntpc_stop when defined CONFIG_DISABLE_SIGNALS?

In my think,  we need to change things as use only one at least in same file.

After we had desided any way, I will reopen it.

PR #1949, opened 2018-07-19
framework/task_manager: Fix stop and exit cb NULL check and move loca…

jeongchanKim:
…tion of pthread data free

1. If TM_STOP_CB or TM_EXIT_CB is not set, we should skip calling its callback
2. Move the location of pthread data free logic
3. fix coding rule
4. add typecasting for setting cb
5. add missing '=' operation for checking current handle count


seinfra:
Target : [525b2e3f8a7ec168f079ffcae8dec673a4c56420] - Code Rule Check (C++) OK.

seinfra:
Target : [525b2e3f8a7ec168f079ffcae8dec673a4c56420] - Code Rule Check OK.

PR #1950, opened 2018-07-19
framework/media : Correct wrong casting type.

dev-hue:
- SocketOutputDataSource::write return type is ssize_t

### DIFF
--- a/framework/src/media/SocketOutputDataSource.cpp
+++ b/framework/src/media/SocketOutputDataSource.cpp
@@ -100,7 +100,7 @@ bool SocketOutputDataSource::isPrepare()
 ssize_t SocketOutputDataSource::write(unsigned char* buf, size_t size)
 {
 	if (!buf) {
-		return (size_t)0;
+		return (ssize_t)0;
 	}
 
 	return send(mSockFd, buf, size, 0);
--- a/framework/src/media/SocketOutputDataSource.cpp
+++ b/framework/src/media/SocketOutputDataSource.cpp
@@ -100,7 +100,7 @@ bool SocketOutputDataSource::isPrepare()
 ssize_t SocketOutputDataSource::write(unsigned char* buf, size_t size)
 {
 	if (!buf) {
-		return (ssize_t)0;
+		return EOF;
 	}
 
 	return send(mSockFd, buf, size, 0);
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [dd3f0810c28247e93ee3271e783e9fe9b6a1831f] - Code Rule Check OK.

seinfra:
Target : [dd3f0810c28247e93ee3271e783e9fe9b6a1831f] - Code Rule Check (C++) OK.

framework/src/media/SocketOutputDataSource.cpp
```diff
@@ -100,7 +100,7 @@ bool SocketOutputDataSource::isPrepare()
 ssize_t SocketOutputDataSource::write(unsigned char* buf, size_t size)
 {
 	if (!buf) {
-		return (size_t)0;
+		return (ssize_t)0;
```
sunghan-chang: Is typecasting needed in this case?

Hyunjun85: typedef uint16_t size_t;
typedef int16_t ssize_t;

It looks like to execute a downcasting.
Isn't there any possibility about data missing?

JeongJunSik: i think

"return EOF;"
This seems to be right.

dev-hue: I think so "return EOF" and typecasting does not seem necessary.

How about FileOutputDataSource::write and BufferOutputDataSource::write?
FileOutputDataSource::write returns `(ssize_t)EOF`
BufferOutputDataSource::write returns `EOF`

Taejun-Kwon: I think type casting is not neccessary, EOF is -1.
Also those of 3 datasources should return EOF.
(Of course we have to define some error types in future)

gcjjyy:
size_t is just a parameter's type. Return value doesn't need to be same type with a parameter's type. Isn't it? About data loss, the function has to control it. This is not a type casting problem.

framework/src/media/SocketOutputDataSource.cpp
gcjjyy: size_t is just a parameter's type. Return value doesn't need to be same type with a parameter's type. About data loss, The function logic has to control it. This is not a type casting problem I think.

seinfra:
Target : [b1e68108f972dd54f282148affa701311808a8a8] - Code Rule Check OK.

PR #1951, opened 2018-07-19
apps/utils/date: add messages to show limitation of year, month and week

sunghan-chang:
The date command allows year within [1970,2106] and month, the day
of week allows 3 abbreviation characters.
It it difficult to know these limitations without messages.

### DIFF
--- a/apps/system/utils/kdbg_date.c
+++ b/apps/system/utils/kdbg_date.c
@@ -325,30 +325,29 @@ int kdbg_date(int argc, char **args)
 #endif
 		ret = date_settime(argc, args);
 	} else {
-#if !(defined(CONFIG_LIBC_LOCALTIME) || defined(CONFIG_TIME_EXTENDED))
+
 		printf("\nUsage: date\n");
 		printf("   or: date [-s FORMAT]\n");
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [268650e2f610696d04b30b6962086330e7a9bc4e] - Code Rule Check OK.

seinfra:
Target : [268650e2f610696d04b30b6962086330e7a9bc4e] - Code Rule Check (C++) OK.

apps/system/utils/kdbg_date.c
```diff
@@ -325,30 +325,29 @@ int kdbg_date(int argc, char **args)
 #endif
 		ret = date_settime(argc, args);
 	} else {
-#if !(defined(CONFIG_LIBC_LOCALTIME) || defined(CONFIG_TIME_EXTENDED))
```
btheosam: Why is this removed from here and moved to line 329?

sunghan-chang: Because 5 lines from start are same between ```#if``` and ```#else```.
Those conditional affects only FORMAT. We don't need to keep same 5 lines at each case.

PR #1959, opened 2018-07-23
frameworn/st_things: Cannot device reset when try multiple times.

Hyunjun85:
The things does not try to reset, if tries reset repeatedly.
It is not changed the reset flag when the reset was completed.
So after reset, if user try reset again, it just returns and doesn't try.

### DIFF
--- a/framework/src/st_things/things_stack/things_stack.c
+++ b/framework/src/st_things/things_stack/things_stack.c
@@ -456,7 +456,7 @@ GOTO_OUT:
 int things_stop_stack(void)
 {
 	THINGS_LOG_D(TAG, THINGS_FUNC_ENTRY);
-	pthread_mutex_lock(&g_things_stop_mutex);	
+	pthread_mutex_lock(&g_things_stop_mutex);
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [ad4e2b578dcdccfa3f18ca422207399dde33a7c0] - Code Rule Check (C++) OK.

seinfra:
Target : [ad4e2b578dcdccfa3f18ca422207399dde33a7c0] - Code Rule Check OK.

framework/src/st_things/things_stack/things_stack.c
```diff
@@ -757,6 +757,7 @@ static void *__attribute__((optimize("O0"))) t_things_reset_loop(reset_args_s *a
 	pthread_mutex_lock(&m_thread_oic_reset);
 	h_thread_things_reset = 0;
 	b_thread_things_reset = false;
+	b_reset_continue_flag = false;
```
davidfather: The flag name is strange to me. For simplicity, b_reset_flag is enough.

davidfather:
Typo in the PR title.

PR #1966, opened 2018-07-24
testcase: register all of testcase at Makefile when builtin enabled

sunghan-chang:
The builtin supports a functionality to set priority and stacksize
to each ASYNC command. But tash does not do it on API side.
To allow setting each different value, register all of testcase
at Makefile when builtin is enabled.

### DIFF
--- a/apps/examples/testcase/Makefile
+++ b/apps/examples/testcase/Makefile
@@ -15,7 +15,6 @@
 # language governing permissions and limitations under the License.
 #
 ###########################################################################
-
 ############################################################################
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [4a0a6f62da7ed23d3e5546f81fc5e1608a4ff73d] - Code Rule Check OK.

seinfra:
Target : [4a0a6f62da7ed23d3e5546f81fc5e1608a4ff73d] - Code Rule Check (C++) OK.

PR #1977, opened 2018-07-26
configs: change a build log level, V to 3

sunghan-chang:
V=1 means enabling echo of commands. But it is not needed at
normal build so that let's set it 3 to disable verbose logs.
FYI, 2 means enabling bug/verbose options in tools and scripts.


seinfra:
Target : [89a71f0752140d3a59c8a9d10142c07470ac208d] - Code Rule Check OK.

seinfra:
Target : [89a71f0752140d3a59c8a9d10142c07470ac208d] - Code Rule Check (C++) OK.

PR #1983, opened 2018-07-30
net/lwip: Fix bug of handling LWIP_IPV6

SeonghoByeon:
- Since LWIP_IPV6 is always defined to differenciate IP version, ifdef is not allowed.
- Fix a problem of missing #if LWIP_IPV6 before using IPADDR6_INIT macro.

### DIFF
--- a/os/net/lwip/src/api/sockets.c
+++ b/os/net/lwip/src/api/sockets.c
@@ -409,7 +409,9 @@ int copy_socket(void *arg)
 {
 	int i;
 	int num_copy = 0;
+#if LWIP_IPV6
 	const ip_addr_t ip6_addr_any = IPADDR6_INIT(0ul, 0ul, 0ul, 0ul);
+#endif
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [952d7d831b640df760ad46752790b4f9957b4cde] - Code Rule Check (C++) OK.

seinfra:
Target : [952d7d831b640df760ad46752790b4f9957b4cde] - Code Rule Check OK.

KimDongEon:
It seems to me the right change.

os/net/lwip/src/api/sockets.c
```diff
@@ -438,7 +440,7 @@ int copy_socket(void *arg)
 				}
 				if (netconn_getaddr(sockets[i].conn, &sock_info->remote_ip, &sock_info->remote_port, 0)) {
 					sock_info->remote_port = 0;
-#ifdef LWIP_IPV6
+#if LWIP_IPV6
```
KimDongEon: It seems to me the right change.

PR #2008, opened 2018-08-07
framework/media : fix a logical error at MediaWorker::startWorker()

davidfather:
Even though a worker thread creation fails, the flag of 'mIsRunning' is set to true.
Fix this logical error by moving the 'mIsRunning' update logic right after the thread creation successfully.

### DIFF
--- a/framework/src/media/MediaWorker.cpp
+++ b/framework/src/media/MediaWorker.cpp
@@ -45,6 +45,7 @@ void MediaWorker::startWorker()
 		if (ret != OK) {
 			medvdbg("Fail to create worker thread, return value : %d\n", ret);
 			--mRefCnt;
+			mIsRunning = false;
 			return;
 		}
 		pthread_setname_np(mWorkerThread, mThreadName);


seinfra:
Target : [90171bab9966542263c87df69dc325e267f014c0] - Code Rule Check OK.

seinfra:
Target : [90171bab9966542263c87df69dc325e267f014c0] - Code Rule Check (C++) OK.

framework/src/media/MediaWorker.cpp
```diff
@@ -40,13 +40,13 @@ void MediaWorker::startWorker()
 		pthread_attr_t attr;
 		pthread_attr_init(&attr);
 		pthread_attr_setstacksize(&attr, mStacksize);
-		mIsRunning = true;
```
gcjjyy: The mIsRunning variable is used in the thread loop for checking the thread is alive. This line should be remain.

framework/src/media/MediaWorker.cpp
```diff
@@ -40,13 +40,13 @@ void MediaWorker::startWorker()
...
 		ret = pthread_create(&mWorkerThread, &attr, static_cast<pthread_startroutine_t>(MediaWorker::mediaLooper), this);
 		if (ret != OK) {
 			medvdbg("Fail to create worker thread, return value : %d\n", ret);
 			--mRefCnt;
 			return;
 		}
+		mIsRunning = true;
```
gcjjyy: This line should be removed. And "mIsRunning = false" should be inserted inside of the "if" failure case.

davidfather: You're right. I will  fix.

jsdosa: If mRefCnt is 1, it is always set to true.
So, you can set it false or not.

framework/src/media/MediaWorker.cpp
```diff
@@ -45,6 +45,7 @@ void MediaWorker::startWorker()
 		if (ret != OK) {
 			medvdbg("Fail to create worker thread, return value : %d\n", ret);
 			--mRefCnt;
+			mIsRunning = false;
```
sunghan-chang: Why do you use the variable to decide thread running?
You can create pthread without any variable and call pthread_cancel when you want to stop it.
In yours,
```
set the variable to true
-> create thread
-> check the variable always during thread is alive
-> set the variable to false when you want to stop the thread
-> check the variable and then stop
```
If you use pthread_cancel, it can work like below
```
create thread without any variable setting
-> run loop without any variable check
-> call pthread_cancel when you want to stop
```

sunghan-chang: My comment is not for @davidfather 's change. But if this module will be changed, this PR is not needed so that I mention above here.

seinfra:
Target : [e3c431d1a1b4881d3c50d08129a27fcb03da9dca] - Code Rule Check (C++) OK.

seinfra:
Target : [e3c431d1a1b4881d3c50d08129a27fcb03da9dca] - Code Rule Check OK.

framework/src/media/MediaWorker.cpp
gcjjyy: @sunghan-chang In that case, how to free memory which allocated local in the thread function?

sunghan-chang: There are several ways to do it.
1. use atexit if it has only one thread or if it can release allocated memory later
2. use SIGKILL signal and callback function
3. use task manager with exit / cancel cb

davidfather: If possible, it is better to change the thread termination process with another PR.

jsdosa: after merged. i will check for pthread_cancel.

PR #2021, opened 2018-08-10
network/wifi_manager: fix build issue on wifi_profile

pillip8282:
if CONFIG_WIFI_PROFILE_SECURESTORAGE is disabled then it can't refer readlen
which is used only above config so it causes a build error.

### DIFF
--- a/framework/src/wifi_manager/wifi_profile.c
+++ b/framework/src/wifi_manager/wifi_profile.c
@@ -118,7 +118,6 @@ static void _wifi_profile_deserialize(wifi_manager_ap_config_s *config, char *bu
 	if (config->ap_auth_type == WIFI_MANAGER_AUTH_OPEN) {
 		return;
 	}
-	
 	DECODE_INTEGER(buf, config->passphrase_length, pos);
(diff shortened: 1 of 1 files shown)


framework/src/wifi_manager/wifi_profile.c
```diff
@@ -262,13 +262,13 @@ wifi_utils_result_e wifi_profile_read(wifi_manager_ap_config_s *config)
 	if (ret != SEE_OK) {
 		return WIFI_UTILS_FILE_ERROR;
 	}
+	nvdbg("readed data len(%d)\n", readlen);
```
manoharahk: Shouldn't it be _**read**_ instead of **_readed_**?

JeongJunSik: It is better to print unsigned int as %u.

seinfra:
Target : [df4d50f2099e87eb50a0b8cc6c9509f645eed268] - Code Rule Check OK.

seinfra:
Target : [df4d50f2099e87eb50a0b8cc6c9509f645eed268] - Code Rule Check (C++) OK.

PR #2032, opened 2018-08-11
add pthread_tryjoin_np

sunghan-chang:
1. pthread: add pthread_tryjoin_np function as syscall API
2. testcase/kernel/pthread: add tc for pthread_tryjoin_np
3. testcase/kernel/pthread: refactoring


seinfra:
Target : [8ccd946b6d29f85f147ec13b83062c31b80a456b] - Code Rule Check (C++) OK.

seinfra:
Target : [8ccd946b6d29f85f147ec13b83062c31b80a456b] - Code Rule Check OK.

seinfra:
Target : [e3aec637856722100574828444a69bcc40d18163] - Code Rule Check OK.

seinfra:
Target : [e3aec637856722100574828444a69bcc40d18163] - Code Rule Check (C++) OK.

sunghan-chang:
@jeongchanKim @manoharahk @lokeshbv @pradeepns85 Let me hear your opinion about location of pthread_tryjoin_np prototype.
Glibc has it in pthread.h but it is not a POSIX API.
Where is appropriate location, pthread.h like glibc or tinyara/pthread.h for non-POSIX API?

davidfather:
This is not a POSIX API, so that it's not good for portability.
Please refer to the postfix of  '_np' which stands for 'non-portable'.
Are you going to open this as a public API?

manoharahk:
@sunghan-chang 
In my opinion, 
1. since this function is a **_nonstandard GNU extension_**, we will add only if it is very much required.  I don't  see newlibc implements this.
2. Regarding the placement of prototype,  musl & glibc both keep it in **_<pthread.h>_** but under **__GNU_SOURCE_**. So it is better to keep it in <pthread.h> in similar fashion. Keeping in tinyara/pthread.h would make it internal and it is not a good idea for application to include <tinyara/pthread.h.>

sunghan-chang:
@davidfather In my opinion, there is no reason not to make it a public API.
1. TizenRT has many non-Posix API and it is difficult to make an app without non-POSIX apis.
2. We don't make a standard and a reference so that we don't need to concern non-portable.
3. Because Linux supports it, it makes us easy to get linux based module.

sunghan-chang:
@manoharahk
1. It is needed to do our current job.
2. I don't use any GNU_SOURCE to add this. Do we need to cover it with GNU_SOURCE?

os/kernel/pthread/pthread_tryjoin_np.c
```diff
@@ -0,0 +1,239 @@
...
+ *
+ *    The caller's task/thread must belong to the same "task group" as the
+ *    pthread is (or was) a member of.  The thread may or may not still
+ *    be running.
+ *
+ * Parameters:
+ *   thread
+ *   pexit_value
+ *
+ * Return Value:
```
davidfather: EBUSY is missing.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -633,6 +633,55 @@ static void tc_pthread_pthread_create_exit_join(void)
...
+	ret_chk = pthread_cancel(pid);
+	TC_ASSERT_EQ("pthread_cancel", ret_chk, OK);
+
+	/* To make sure thread is terminated */
+	sleep(SEC_1);
+
+	ret_chk = pthread_tryjoin_np(pid, &pexit_value);
+	TC_ASSERT_EQ("pthread_tryjoin_np", ret_chk, ESRCH);
+
+	ret_chk = pthread_create(&pid, NULL, do_nothing_thread, NULL);
```
davidfather: do_nothing_thread is removed in your another commit. Please check this.

manoharahk:
@sunghan-chang ,
1. Ok. 
2. All these **__np_** functions are GNU extensions to <pthread.h> provided by gnulib and other C libraries cover it with GNU macro. But for TizenRT case, i am not sure whether to cover it with GNU or not .

os/kernel/pthread/pthread_tryjoin_np.c
```diff
@@ -0,0 +1,239 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ *
+ *   Copyright (C) 2007, 2008, 2011, 2013 Gregory Nutt. All rights reserved.
+ *   Author: Gregory Nutt <gnutt@nuttx.org>
```
manoharahk: Is it back ported from NuttX?, if not shall we remove this?

sunghan-chang: Actually no. But it is almost same as phtread_join. The difference between pthread_join and pthread_tryjoin_np is blockable. When thread is not terminated, pthread_join blocks this context. But pthread_tryjoin_np just returns with error. So, I just remove blocking logic from pthread_join. Except removing, all are same as pthread_join. That's why I left it.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
sunghan-chang: I added this at 2nd commit and I removed do_nothing_thread at 3rd commit. I refacored it at 3rd commit with 2nd commit added line together. So it is right. At 2nd commit, another thread is not perfectly matched to my usages.

seinfra:
Target : [758e14da0fd2a3bbed8bf915e6e20ea5f7cfd168] - Code Rule Check OK.

seinfra:
Target : [758e14da0fd2a3bbed8bf915e6e20ea5f7cfd168] - Code Rule Check (C++) OK.

seinfra:
Target : [8e3a3260d5158a907a3c000e50fb127171185ad7] - Code Rule Check OK.

seinfra:
Target : [8e3a3260d5158a907a3c000e50fb127171185ad7] - Code Rule Check (C++) OK.

PR #2037, opened 2018-08-13
system: remove mounting procfs and reading it for sysinfo

sunghan-chang:
1. mount procfs
 CONFIG_FS_AUTOMOUNT_PROCFS does it on fs_auto_mount() so that
 mounting it on preapp_start() should be removed.
2. read version info from procfs
 Even if CONFIG_BUILD_PROTECTED is enabled, version.h can be
 read from anywhere. That is not a value of kernel context.

### DIFF
--- a/apps/system/init/init.c
+++ b/apps/system/init/init.c
@@ -102,14 +102,6 @@ int preapp_start(int argc, char *argv[])
 	int pid;
 #endif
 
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [8427254d3fded43752cf6046468beabad2f6ee08] - Code Rule Check (C++) OK.

seinfra:
Target : [8427254d3fded43752cf6046468beabad2f6ee08] - Code Rule Check OK.

PR #2052, opened 2018-08-17
framework/wifi_manager: Add a public API to extract the connected AP config

SeonghoByeon:
-Wi-Fi Manager saves the connected AP config as a file, and provides the public API to extract
-Wi-Fi profile now can differentiate an internal file link


seinfra:
Target : [b90dfa2425f18defc79208b4cf4e1e485a988b5d] - Code Rule Check OK.

PR #2066, opened 2018-08-20
README: add AdvancedEnv tab

sunghan-chang:
That shows necessary installation to use some features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ export PATH=<Your Toolchain PATH>:$PATH
 ```
 Be aware that recommanded toolchain is fully working on 64bits machine.
 
+NOTE: See [Additional Requirements](docs/InstallAdReq.md) for gRPC, ROMFS, OCF and IoT.js.
+
 ### Getting the sources
 
 ```bash
--- a/docs/InstallAdReq.md
+++ b/docs/InstallAdReq.md
@@ -1,0 +1,27 @@
+# Additional Requirements
+
+To use below features, each prerequisites are necessary.  
+The configure.sh also shows missed package to build configured as show below:
+```bash
+os$ ./tools/configure.sh artik053/grpc
+  Copy build environment files
+ 
+  CAUTION!! To build artik053/grpc, protoc should be installed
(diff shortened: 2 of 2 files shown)


sunghan-chang:
When we support docker to build TizenRT, this document will be an appendix.

README.md
```diff
@@ -25,6 +25,8 @@ export PATH=<Your Toolchain PATH>:$PATH
 ```
 Be aware that recommanded toolchain is fully working on 64bits machine.
 
+NOTE: See [Advanced Environments](docs/InstallAdvEnv.md) to use gRPC, ROMFS, IoTivity, IoTjs and so on.
```
davidfather: Additional requirements might be better than 'advanced environments':
Additional requirements for gRPC, ROMFS, IoTivity, and IoT.js.
Please remove 'and so on'.

docs/InstallAdvEnv.md
```diff
@@ -0,0 +1,27 @@
...
+To use below features, each prerequisites are necessary.  
+The configure.sh also shows missed package to build configured as show below:
+```bash
+os$ ./tools/configure.sh artik053/grpc
+  Copy build environment files
+ 
+  CAUTION!! To build artik053/grpc, protoc should be installed
+  Find details at https://github.com/Samsung/TizenRT/blob/master/external/protobuf/README.md
+ 
+  Configuration is Done!
```
davidfather: In this case, protoc is not installed yet.
So is it reasonable to say that configuration is done?

README.md
davidfather: When you say IoT.js, don't forget the dot between 'IoT' and 'js'.

davidfather: Regarding feature-wise description, OCF is the proper term instead of IoTivity.

docs/InstallAdvEnv.md
sunghan-chang: Regardless existance of packages, there is no problem on configuration.
Even if he installs missed packages, he does not need to re-configure because it affects only build.
That's why it says "configuration is done".

The previous step of build is this configuration so that this time we check and say missed packages. But actually it is not a step of configuration.

PR #2070, opened 2018-08-21
testcase/kernel/sched: remove unncessary logic and refactor

sunghan-chang:
1. Because the testcase of waitid has a wrong test,
   let's remove it.
[tc_sched_waitid] FAIL [Line : 372] waitid : Values (ret_chk == 0x0) and (-1 == 0xffffffff) are not equal

2. Because the function name should have what it does,
   let's rename it.
3. Let's remove unnecessary definitions.


seinfra:
Target : [52d3ae4c25ac5c76b116b132ec239fd92d46a780] - Code Rule Check OK.

PR #2072, opened 2018-08-21
media: Check result using PLAYER_OK

junmin-kim:
PLAYER_ERROR will be split.
So check whether result is PLAYER_OK


seinfra:
Target : [10f9031f2bb16cd8cda6274edd8acdd10161730d] - Code Rule Check OK.

PR #2074, opened 2018-08-21
media: Check invalid param early in MediaRecorder

junmin-kim:
Check whether vol is nullptr before enqueuing command


seinfra:
Target : [cf9f9659d97983887a50921632dfb97353ec6a34] - Code Rule Check OK.

seinfra:
Target : [cf9f9659d97983887a50921632dfb97353ec6a34] - Code Rule Check (C++) OK.

jsdosa:
I think it is correct to first check whether the worker is alive in the current structure.
Let's think about this part as a whole.

Taejun-Kwon:
good point.

PR #2078, opened 2018-08-22
add defense code for player TC

jsdosa:
Add error handling logics
for filesystem and file open file cases

### DIFF
--- a/apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
+++ b/apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
@@ -17,6 +17,7 @@
  ****************************************************************************/
 
 #include <memory>
+#include <string.h>
 #include <media/MediaPlayer.h>
 #include <media/FileInputDataSource.h>
 #include "tc_common.h"
@@ -51,8 +52,14 @@ void EmptyObserver::onPlaybackPaused(media::MediaPlayer &mediaPlayer)
 static void SetUp(void)
 {
 	FILE *fp = fopen(dummyfilepath, "w");
-	fputs("dummydata", fp);
-	fclose(fp);
+
+	if (fp != nullptr) {
+		int ret = fputs("dummydata", fp);
+		if (ret != (int)strlen("dummydata")) {
+			printf("MediaPlayer SetUp Failed\n");
+		}
+		fclose(fp);
+	}
 }
 
 static void TearDown()


seinfra:
Target : [e9bacaeb2ed7e9f235f8c924f2b41ef55ca8377d] - Code Rule Check (C++) OK.

seinfra:
Target : [e9bacaeb2ed7e9f235f8c924f2b41ef55ca8377d] - Code Rule Check OK.

juitem:
Would you please describe more details about defense code?.
Add error handling codes for filesystem or file open fail cases

jsdosa:
@juitem ok thanks., i will change that,

seinfra:
Target : [be9546b3832f06db22b8cb82f071df4e7b4a1c22] - Code Rule Check OK.

seinfra:
Target : [be9546b3832f06db22b8cb82f071df4e7b4a1c22] - Code Rule Check (C++) OK.

apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
```diff
@@ -48,13 +48,21 @@ void EmptyObserver::onPlaybackPaused(media::MediaPlayer &mediaPlayer)
...
 }
 
-static void SetUp(void)
+static int SetUp(void)
 {
 	FILE *fp = fopen(dummyfilepath, "w");
-	fputs("dummydata", fp);
-	fclose(fp);
+	if (fp != nullptr) {
+		if (fputs("dummydata", fp) != EOF) {
```
Taejun-Kwon: below is correct.
```
size_t ret = fputfs("dummydata", fp);
fclose(fp);
if (ret != strlen("dummydata") {
     return TRUE;
}
return false;
```

apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
```diff
@@ -409,7 +417,10 @@ static void utc_media_MediaPlayer_setVolume_n(void)
 
 int utc_media_MediaPlayer_main(void)
 {
-	SetUp();
+	if (SetUp() != 0) {
```
Taejun-Kwon: I think let it work will be better than return -1 because in this case tc result will treat it as all succedded

jsdosa: In case of failure, do not proceed tc.

apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
jsdosa: int fputs(FAR const char *s, FAR FILE *stream)
{
	int nput;
	int ret;

	/* Make sure that a string was provided. */

#ifdef CONFIG_DEBUG				/* Most parameter checking is disabled if DEBUG is off */
	if (!s) {
		set_errno(EINVAL);
		return EOF;
	}
#endif

	/* Write the string.  Loop until the null terminator is encountered */

	for (nput = 0; *s; nput++, s++) {
		/* Write the next character to the stream buffer */

		ret = lib_fwrite(s, 1, stream);
		if (ret <= 0) {
			return EOF;
		}

		/* Flush the buffer if a newline was written to the buffer */

		if (*s == '\n') {
			ret = lib_fflush(stream, true);
			if (ret < 0) {
				return EOF;
			}
		}
	}

	return nput;
}

Taejun-Kwon: EOF is error, but we should check size of written data.

apps/examples/testcase/ta_tc/media/utc/utc_media_mediaplayer.cpp
Taejun-Kwon: No, TC_ASSERT increase fail count and TC_SUCCESS_RESULT will increase success count.
So if we do not test all, fail count is not increased.
This doesn't give correct info to user.

jsdosa: you right. so i remove return type and just will add notice

seinfra:
Target : [8e76c8214f0335486f00e8e6429c6d86f4555420] - Code Rule Check OK.

seinfra:
Target : [8e76c8214f0335486f00e8e6429c6d86f4555420] - Code Rule Check (C++) OK.

PR #2091, opened 2018-08-22
framework/task_manager: Fix bug for broadcast when user data is NULL

jeongchanKim:
if user_data is NULL, size var should be set to zero.


seinfra:
Target : [a3708d2327db427665148301ab1b453d7e15859b] - Code Rule Check (C++) OK.

seinfra:
Target : [a3708d2327db427665148301ab1b453d7e15859b] - Code Rule Check OK.

PR #2104, opened 2018-08-24
framework/st_things: fix memory leak

sushant-23:
In cloud_manager, log_in_out_to_cloud makes a call to make_cloud_address
with ci_addr as NULL. Hence, we can free ipport.

### DIFF
--- a/framework/src/st_things/things_stack/cloud/cloud_manager.c
+++ b/framework/src/st_things/things_stack/cloud/cloud_manager.c
@@ -1557,9 +1557,11 @@ int push_notification_to_cloud(const char *uri, OCRepPayload *payload)
 static char *make_cloud_address(char *ip, char *port, const char *ci_addr)
 {
 	char *ipport = (char *)ci_addr;
+	bool is_ci_addr_null = true;
 	THINGS_LOG_D(TAG, "ip=%s, port=%s, ci_addr=%s", ip, port, ci_addr);
 
 	if (ci_addr != NULL && strlen(ci_addr) > 0) {	// ci address.
+		is_ci_addr_null = false;
 		char *point = strstr(ci_addr, DEFAULT_COAP_TCP_HOST);
 		if (point) {
 			ipport = point + strlen(DEFAULT_COAP_TCP_HOST);
@@ -1571,13 +1573,23 @@ static char *make_cloud_address(char *ip, char *port, const char *ci_addr)
 	// Update Cloud Address.
 	memset(g_cloud_address, 0, MAX_CI_ADDRESS);
 	if (things_strcat(g_cloud_address, MAX_CI_ADDRESS, DEFAULT_COAP_TCP_HOST) == NULL) {
+		if (is_ci_addr_null) {
+			things_free(ipport);
+		}
 		THINGS_LOG_E(TAG, "things_strcat() is failed.");
 		return NULL;
 	}
 
 	if (things_strcat(g_cloud_address, MAX_CI_ADDRESS, ipport) == NULL) {
+		if (is_ci_addr_null) {
+			things_free(ipport);
+		}
 		THINGS_LOG_E(TAG, "things_strcat() is failed.");
 		return NULL;
+	}
+
+	if (is_ci_addr_null) {
+		things_free(ipport);
 	}
 
 	return g_cloud_address;


seinfra:
Target : [8bed6c677ee00191648ef4750eec7179b7a10295] - Code Rule Check OK.

seinfra:
Target : [8bed6c677ee00191648ef4750eec7179b7a10295] - Code Rule Check (C++) OK.

Hyunjun85:
Could you please change titile as 'framework/st_things : ~~~~~'.

framework/src/st_things/things_stack/cloud/cloud_manager.c
```diff
@@ -1580,6 +1582,11 @@ static char *make_cloud_address(char *ip, char *port, const char *ci_addr)
 		return NULL;
 	}
 
+	if (is_ci_addr_null == true) {
```
SenthilKumarGS: Minor: Can we change it to **if(is_ci_addr_null)** ?

sushant-23:
@Hyunjun85 @SenthilKumarGS  did the necessary changes.

seinfra:
Target : [13d52cbb29863e3f7edb4538b5cf4166c1d1c8f3] - Code Rule Check (C++) OK.

seinfra:
Target : [13d52cbb29863e3f7edb4538b5cf4166c1d1c8f3] - Code Rule Check OK.

framework/src/st_things/things_stack/cloud/cloud_manager.c
Hyunjun85: @sushant-23 Could you please leave your opinion.
Our policy is that to reply every comment whether it is fix or not.
If you don't want to change then please leave why.

sushant-23:
hi @Hyunjun85 @SenthilKumarGS , that was a mistake on my side. I updated `is_ci_addr_null` in my local repo but forgot to update the PR. Made the necessary changes.

seinfra:
Target : [3fb58efcd6a50b93c03eae53fcdace4b1c0ed0c1] - Code Rule Check (C++) OK.

seinfra:
Target : [3fb58efcd6a50b93c03eae53fcdace4b1c0ed0c1] - Code Rule Check OK.

SenthilKumarGS:
@YiHyunJin ,
@1569, In else block, we call make_ip_port() which allocates memory, is_ci_addr_null will remain as true for indicate the necessity of de-allocation before returning to caller.  If the control doesn't come to else block, then is_ci_addr_null flag will be false and memory allocation/deallocation won't happen at all.
I kindly request for further clarification.

sushant-23:
hi @YiHyunJin , `is_ci_addr_null` is being assigned false at line 1564 only when `ci_addr` is NULL. 
So de allocating `ipport` when ci_addr is not NULL gives an error since both refer to the same memory location.

framework/src/st_things/things_stack/cloud/cloud_manager.c
```diff
@@ -1571,15 +1573,25 @@ static char *make_cloud_address(char *ip, char *port, const char *ci_addr)
 	// Update Cloud Address.
 	memset(g_cloud_address, 0, MAX_CI_ADDRESS);
 	if (things_strcat(g_cloud_address, MAX_CI_ADDRESS, DEFAULT_COAP_TCP_HOST) == NULL) {
+		if (is_ci_addr_null) {
+			things_free(ipport);
```
Hyunjun85: Why is it freed here? ipport is assigned 1560 as a parameter.
My thinking, ci_addr will free when returns NULL.

sushant-23: As we are returning from the function, and we dont require `ipport` since `ci_addr` is NULL, so, we can free it.

Hyunjun85: @sushant-23 Right. But It looks not good. 
You'd better it frees in caller function.
Any function or module has allocated a memory then the function or module will free the memory.

SenthilKumarGS: @Hyunjun85, 
Thanks for approving.
I would like to clarify. We have to de-allocate memory which was allocated @ line: 1569 before returning. This memory allocation is done only when 'ci_addr' which is passed by the caller is NULL. So the 'if' block will be executed only when the memory has been allocated.

PR #2105, opened 2018-08-24
Fix issue with concurrent socket select and socket close

dr-venkman:
Fix issue with concurrent socket select and socket close.

Scenario: 2 threads 1 and 2.

Thread 1 select-waits on a socket fd, while Thread 2 closes the same socket fd. 

At present: TizenRT does not de-allocate the said socket fd.
Fix: Exit the select wait, and also de-allocate the socket fd.

### DIFF
--- a/os/net/lwip/src/api/sockets.c
+++ b/os/net/lwip/src/api/sockets.c
@@ -524,6 +524,27 @@ struct socket *tryget_socket(int s)
 	return &sockets[s];
 }
 
+ /**
+  * Check closed socket which has non-zero select_waiting value
+  *
+  * @param s externally used socket index
+  * @return struct lwip_sock for the socket with non-zero select_waiting value,
+  *         or NULL if not found
+  */
+struct socket *trycheck_selwait_socket(int s)
+{
+	s -= LWIP_SOCKET_OFFSET;
+	if ((s < 0) || (s >= NUM_SOCKETS)) {
+		return NULL;
+	}
+
+	if (!sockets[s].conn && sockets[s].select_waiting > 0) {
+		return &sockets[s];
+	}
+
+	return NULL;
+}
+
 /**
  * Allocate a new socket for a given netconn.
  *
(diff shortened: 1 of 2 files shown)


os/net/lwip/src/api/sockets.c
```diff
@@ -1835,7 +1861,17 @@ int lwip_poll(int fd, struct pollfd *fds, bool setup)
 
 	sock = tryget_socket(fd);
 	if (!sock) {
-		return -EBADF;
+		/* Need to check select_waiting counter on socket ... */
+		sock = trycheck_selwait_socket(fd);
```
SeonghoByeon: Would not it be better to call trycheck_selwait_socket, when the socket deallocation is called (in other thread)? Please let me know your opinion.

dr-venkman: Thanks for the comment. Actually, de-allocation is in two parts, 1. removing socket netconn structure (which happens in the closing thread), and 2. Decrementing socket select_waiting variable, which happens under poll (poll_teardown specifically). When both are 0, the same socket structure can be re-used. However, at present, part 2 happens after the condition check above. Also, the above condition check is in the right context, i.e. when another thread might not know that the socket fd is closed, and might try to select wait on it. This is where such cases are filtered out. Using the function above in the closing thread will most likely miss this context. Having mentioned so, there might be better alternatives to fixing this, but this is what I gather from the patch.

SeonghoByeon: Thanks for the kind reply. Now, I understand the point. Then, what do you think if lwip_poll_teardown is called right after filling zeros in event variables, and then return within this if statement (if (!sock) at line 1863)? It might be more clear to recognize the problem dealt with by this PR happened.

seinfra:
Target : [42a239361f6bd8a9fb126443eff06e4993e9c20d] - Code Rule Check OK.

seinfra:
Target : [42a239361f6bd8a9fb126443eff06e4993e9c20d] - Code Rule Check (C++) OK.

PR #2115, opened 2018-08-27
framework/task_manager: Add pthread_detach before stopping a pthread

jeongchanKim:
if not detaching, joininfo is not destroyed when cancelling the pthread


seinfra:
Target : [cd9f14befa6c50e97425a97bccb647447195ad43] - Code Rule Check OK.

PR #2123, opened 2018-08-28
framework/task_manager: Change unicast about freeing user data

jeongchanKim:
When user sends unicast msg, user should free the user data.
With this change, user don't need to free the user data. It will be freed automatically.


seinfra:
Target : [031aca7f675b3d59534278fa2a197cf0f2bde2bc] - Code Rule Check OK.

seinfra:
Target : [031aca7f675b3d59534278fa2a197cf0f2bde2bc] - Code Rule Check (C++) OK.

PR #2130, opened 2018-08-29
mm/mm_realloc: Fix copysize when allocate a new block

jeongchanKim:
1. The "oldsize" variable is consists of chunk size and data size.
But when copying oldmem to newmem, we should do only for data.
Realloc has a memory corruption problem because of this fault as show below:
1) heap status
ALLOC NODE 1
ALLOC NODE 2
ALLOC NODE 3
FREE NODE
2) realloc memory of ALLOC NODE 2 (NEIGHBOR_EXTENSION is disabled, or it's impossible to extend), then heap status change like this:
ALLOC NODE 1
ALLOC NODE 2 // 2nd: to be freed
ALLOC NODE 3
ALLOC NODE 4 // 1st: new allocated
FREE NODE
3) memcpy from ALLOC NODE 2 to NODE 4
if `data size` of NODE 4 is less than `total size` of NODE 2, FREE NODE area will be overwritten.

2. mm/realloc: Exclude reallocation for smaller size from disabling neighbor extension and modify the config name


seinfra:
Target : [4e5abf8f1cb337ad20852821e7ab37f771ffec77] - Code Rule Check OK.

seinfra:
Target : [4e5abf8f1cb337ad20852821e7ab37f771ffec77] - Code Rule Check (C++) OK.

os/mm/mm_heap/mm_realloc.c
```diff
@@ -381,7 +381,13 @@ FAR void *mm_realloc(FAR struct mm_heap_s *heap, FAR void *oldmem, size_t size)
 		newmem = (FAR void *)mm_malloc(heap, size);
 #endif
 		if (newmem) {
-			memcpy(newmem, oldmem, oldsize);
+			size_t copysize = oldsize - SIZEOF_MM_ALLOCNODE;
+#ifdef CONFIG_DISABLE_REALLOC_NEIGHBOR_EXTENSION
+			if (newsize <= oldsize) {
```
juitem: what is relation with NEIGHBOR EXTENTION case, it seems should be re-calculate again for all cases

zhouxinhe: @juitem 
`NEIGHBOR_EXTENSION enable and newsize <= oldsize` case is handled at line 158,
we need only consider `newsize <= oldsize` when NEIGHBOR_EXTENSION is disabled here.

jeongchanKim: @juitem, zhouxinhe is right. for newsize <= oldsize case, it will be returned at line 158.

sunghan-chang: Without this code, just use ```size```. no?

zhouxinhe: @sunghan-chang 
Without this code, I face a heap overwritten issue.
1. heap status
ALLOC NODE 1
ALLOC NODE 2
ALLOC NODE 3
FREE NODE
2. realloc memory of ALLOC NODE 2 (NEIGHBOR_EXTENSION is disabled, or it's impossible to extend), then heap status change like this:
ALLOC NODE 1
ALLOC NODE 2 // 2nd: to be freed
ALLOC NODE 3
ALLOC NODE 4 // 1st: new allocated
FREE NODE
3. memcpy from ALLOC NODE 2 to NODE 4
if `data size` of NODE 4 is less than `total size` of NODE 2, FREE NODE area will be overwritten...

seinfra:
Target : [40226c104e1c484686e938527140d7151a45d326] -

seinfra:
Target : [245508d729472dae1ae98aac896c52065a5e9a46] - Code Rule Check (C++) OK.

seinfra:
Target : [bddefc7375be898893094a7fe87f43086d3815e3] - Code Rule Check OK.

seinfra:
Target : [bddefc7375be898893094a7fe87f43086d3815e3] - Code Rule Check (C++) OK.

os/mm/mm_heap/mm_realloc.c
sunghan-chang: Thank you for information. But we considered overall of realloc.
1. CONFIG_DISABLE_REALLOC_XXX_EXTENSION
 This feature is disabling extension. That means new size should be bigger than old size. But we blocked (newsize < oldsize) case as well. But there is no issue on smaller size and is not matched with config name and intention. We will not adapt it on smaller size.
2. With 1st change, no need to check smaller case at that location. Just subtract ALLOCNAME is enough and it is more simple.

zhouxinhe: Welcome! Yes, that's a good idea! ^^

sunghan-chang:
@jeongchanKim For 2nd commit, important content is adapting it only for extension. not modifying name and moving location.
```
mm/realloc: exclude reallocation for smaller size from disabling ne...xx extension and modify the config name
```

os/mm/mm_heap/mm_realloc.c
sunghan-chang: @zhouxinhe  Anyway you found our big problem and gave a solution. Thank you very much.

sunghan-chang:
@jeongchanKim And please fill 1st commit description with @zhouxinhe 's example.
```
The "oldsize" variable is consists of chunk size and data size.
But when copying oldmem to newmem, we should do only for data.
Realloc has a memory corruption problem because of this fault as show below:
1. heap status
ALLOC NODE 1
ALLOC NODE 2
ALLOC NODE 3
FREE NODE
2. realloc memory of ALLOC NODE 2 (NEIGHBOR_EXTENSION is disabled, or it's impossible to extend), then heap status change like this:
ALLOC NODE 1
ALLOC NODE 2 // 2nd: to be freed
ALLOC NODE 3
ALLOC NODE 4 // 1st: new allocated
FREE NODE
3. memcpy from ALLOC NODE 2 to NODE 4
if `data size` of NODE 4 is less than `total size` of NODE 2, FREE NODE area will be overwritten...
```

seinfra:
Target : [b7255ce729570bd5c4910dccaf48c05eca9d955e] - Code Rule Check OK.

seinfra:
Target : [b7255ce729570bd5c4910dccaf48c05eca9d955e] - Code Rule Check (C++) OK.

PR #2131, opened 2018-08-29
framework/media: Init the audio codec when data source calls open().

jiawei-xiang:
1. Fix the mConfig return false issue.  
2. Add Decoder::create()/ Encoder::create() interface, and call it in data source open() fucntion.
@Taejun-Kwon


seinfra:
Target : [76c24c7791a226072e3f053dfd81e5b4d05c81b9] - Code Rule Check (C++) OK.

PR #2134, opened 2018-08-29
media/voice: Add Speech detector prototype and Software EPD solution

gcjjyy:
Add SpeechDetector prototype for co-work.
Most functions are not implemented yet.

Add Software EPD solution(Fixed float) at the External.


seinfra:
Target : [65f638ad02e52fa3182b424505bd8b9fbd56a530] - Code Rule Check (C++) OK.

framework/src/media/voice/SpeechDetector.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace media {
+namespace voice {
+
+static SpeechDetector *g_instance = nullptr;
+
+SpeechDetectorInterface *SpeechDetectorInterface::instance()
+{
+	if (!g_instance) {
+		g_instance = new SpeechDetector();
```
junmin-kim: To implement singleton, this way can be problem.
How about using `static SpeechDector instance`?

gcjjyy: You're right. I'll fix this!

seinfra:
Target : [57f1f391349e968a459f5f66a3ed714467ff6aec] - Code Rule Check OK.

gcjjyy:
@sunghan-chang You're right. I'll follow the rules!

PR #2155, opened 2018-09-04
framework/task_manager: Modify task_manager_set_broadcast_cb to alloc…

jeongchanKim:
… for callback data

1. when setting broadcast callback, user can pass callback data. it will be allocated, not only has pointer.
2. Change its TC for passing callback data with tm_msg_t type


seinfra:
Target : [66f776151dde947cb44ac248f0c079b6d43cb96f] - Code Rule Check (C++) OK.

seinfra:
Target : [66f776151dde947cb44ac248f0c079b6d43cb96f] - Code Rule Check OK.

PR #2165, opened 2018-09-05
Fix Error in Tinyalsa during audio playback

kishore-sn:
During audio playback, the media player tries to call pcm_drain API
at the end of playback to flush the data and close the device. However,
since we have stopped enqueuing new buffers to the codec and waiting
for dequeue msg, the codec generates underrun errors in the meantime.
Since we are in the phase of closing the audio device, we can ignore
any underrun related errors. Hence, modify tinyasla to ignore underrun
during pcm_drain.


seinfra:
Target : [f6cef94dab59e0d532e251091538bd3406b044c2] - Code Rule Check (C++) OK.

seinfra:
Target : [f6cef94dab59e0d532e251091538bd3406b044c2] - Code Rule Check OK.

PR #2171, opened 2018-09-07
external/grpc: Configure pthread size for gRPC modules

dr-venkman:
external/grpc: Configure pthread size for gRPC modules.

Presently, grpc pthreads use a hard-coded pthread size during runtime. For ease of configuration, a new menuconfig parameter CONFIG_GRPC_PTHREAD_SIZE is added, with a default size of 10240 bytes.


seinfra:
Target : [ecb52d53dc91268be1ea46d062fb8991d4a307fd] -

seinfra:
Target : [dfccd3d0ac802ff3e5de8cf0522451b378fe4a48] - Code Rule Check OK.

seinfra:
Target : [7efb6144059945a94ca95ed69258fc04aa339835] - Code Rule Check OK.

seinfra:
Target : [7efb6144059945a94ca95ed69258fc04aa339835] - Code Rule Check (C++) OK.

external/grpc/README.md
```diff
@@ -110,8 +110,9 @@ Please refer to `grpc_greeter_client` and `grpc_route_client` applications under
 
 An important consideration when running gRPC on TizenRT is the stack size allocated for its threads.
 From our initial verfication, we observe that a minimum thread stack size of 16384 bytes is necessary for the main gRPC thread to run successfully on TizenRT.
-Additionally, this gRPC main thread engine uses `pthread_create` internally for its run-time procedures, for which we recommend setting a stack size of 10240 bytes or higher.
-This can be configured via menuconfig at the location `Kernel Features -> Stack size information -> Default pthread stack size`.
+This can be achieved by either using the `pthread_attr_setstacksize` API or configuring the default pthread stacksize via menuconfig, at the location `Kernel Features -> Stack size information -> Default pthread stack size`.
```
sunghan-chang: This configuration affects all of thread including not gRPC threads. Why don't you use gRPC specific config?

external/grpc/Kconfig.protocol
```diff
@@ -13,6 +13,12 @@ config GRPC
 
 if GRPC
 
+config GRPC_PTHREAD_SIZE
```
sunghan-chang: Is this gRPC main thread? or sub-threads made by gRPC main thread?

external/grpc/README.md
dr-venkman: The line mentions either approaches as options. The grpc example application uses the pthread API to set the stack size.

seinfra:
Target : [d66b3f8e7024b1655ca049c61675dcb599378b6e] - Code Rule Check OK.

external/grpc/Kconfig.protocol
dr-venkman: sub-threads created for grpc core modules.

seinfra:
Target : [d66b3f8e7024b1655ca049c61675dcb599378b6e] - Code Rule Check (C++) OK.

external/grpc/README.md
dr-venkman: This is a result of using std::thread in certain C++ applications (for example, the two grpc applications). While we can use `pthread_attr_setstacksize`, and then pass this attr to `pthread_create`, there is no corresponding option for `std::thread`. Hence, we are faced with the option of having to raise the default stack size. I agree this is not memory-efficient, but this is the solution we have at present. Please let me know if you find a better alternative.

seinfra:
Target : [6f0f19ef424b11965e361cc11a54efbf382ca8f7] - Code Rule Check OK.

seinfra:
Target : [6f0f19ef424b11965e361cc11a54efbf382ca8f7] - Code Rule Check (C++) OK.

external/grpc/Kconfig.protocol
sunghan-chang: As you know, usually developer does not see documentation. How about mark this with "sub-threads" or something else which we can know what this configures?
And how about add description how to set gRPC main thread stack size in ---help--- tab.

PR #2203, opened 2018-09-12
 framework/task_manager: Add procedures for NULL setting after free

jeongchanKim:
1. If not set to NULL after free, callback msg and callback data can be treated as having data, even they are empty.
2. testcase/task_manager: Modify return check for stop task and Add semaphore for callback flag


seinfra:
Target : [e710c8453b08b243d804de52a5d4ad11eb784467] - Code Rule Check OK.

seinfra:
Target : [e710c8453b08b243d804de52a5d4ad11eb784467] - Code Rule Check (C++) OK.

PR #2206, opened 2018-09-12
framework/task_manager : Add NULL setting after free and Modify its TCs

jeongchanKim:
1. framework/taskmanager: Add procedures for NULL setting after free
2.  testcase/task_manager: Modify return check for stop task and Add semaphore for callback flag
3. testcase/task_manager: Add sem_destroy in utc_task_manager_set_stop_cb_p


seinfra:
Target : [db0d4fcd6567515b9777e8315f874b453b27f403] - Code Rule Check (C++) OK.

PR #2210, opened 2018-09-12
apps/examples/network_tc: Add a missing socket close

SeonghoByeon:
- Add a logic to close the socket before return in getpeername_client function


seinfra:
Target : [433c0dfb41e21099817e876006fb902308503dd2] - Code Rule Check OK.

PR #2218, opened 2018-09-12
Correct time initialization for the case where time is initialized fr…

sunghan-chang:
…om settings in the .config file.  The correct range for month (tm_mon) internally is 0-11, not 1-12 as it is in the .config file.


seinfra:
Target : [8b94e62fba9ff5fa5962ca4e7ed49af499a0762f] - Code Rule Check OK.

PR #2231, opened 2018-09-14
media/utc: Restore volume after test

junmin-kim:
Before testing setVolume, save previous volume and restore after test

### DIFF
--- a/apps/examples/testcase/ta_tc/media/utc/utc_media_mediarecorder.cpp
+++ b/apps/examples/testcase/ta_tc/media/utc/utc_media_mediarecorder.cpp
@@ -105,12 +105,20 @@ static void utc_media_MediaRecorder_setDataSource_n(void)
 
 static void utc_media_MediaRecorder_setVolume_p(void)
 {
+	uint8_t prev, volume;
 	MediaRecorder mr;
 	unique_ptr<FileOutputDataSource> dataSource = unique_ptr<FileOutputDataSource>(new FileOutputDataSource(channels, sampleRate, pcmFormat, filePath));
 	mr.create();
 	mr.setDataSource(std::move(dataSource));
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [9a29802c3988912ac40e82aae41c8f7c658c3afd] - Code Rule Check OK.

seinfra:
Target : [9a29802c3988912ac40e82aae41c8f7c658c3afd] - Code Rule Check (C++) OK.

apps/examples/testcase/ta_tc/media/utc/utc_media_mediarecorder.cpp
```diff
@@ -105,26 +105,39 @@ static void utc_media_MediaRecorder_setDataSource_n(void)
...
 	TC_ASSERT_EQ("utc_media_mediarecorder_setVolume", mr.setVolume(10), RECORDER_ERROR_NOT_ALIVE);
 	mr.create();
 	mr.setDataSource(std::move(dataSource));
 	mr.prepare();
-	TC_ASSERT_EQ("utc_media_mediarecorder_setVolume", mr.setVolume(11), RECORDER_OK);
+	mr.getVolume(&prev);
+	TC_ASSERT_EQ_CLEANUP("utc_media_mediarecorder_setVolume", mr.setVolume(11), RECORDER_OK, goto cleanup);
+
+cleanup:
+	mr.setVolume(prev);
```
YiHyunJin: @junmin-kim Does it need to do set prev in situations where it does not set?

junmin-kim: This test try to set volume as 11, but set 10(max value).
So it is needed to restore volume

PR #2236, opened 2018-09-14
 framework/wifi_manager: Add configurations for autoconnect operation

SeonghoByeon:
framework/wifi_manager: 
1. WIFIMGR_DISABLE_EXTERNAL_AUTOCONNECT, to disable an external autoconnect operation defined in wifi dirver
2. WIFIMGR_ENABLE_INTERNAL_AUTOCONNECT, to enable an internal autoconnect operation defined in Wi-Fi Manager
3. WIFIMGR_INTERNAL_AUTOCONNECT_INTERVAL, to set an interval of internal autoconnect operation

external/slsi_wifi: 
1. Add definition of WPA_AUTOCONNECT to enable/disable WPA's AUTOCONNECT operation

### DIFF
--- a/framework/src/wifi_manager/Kconfig
+++ b/framework/src/wifi_manager/Kconfig
@@ -70,5 +70,16 @@ config WIFIMGR_DISABLE_AUTO_GET_IP
 	bool "disable auto get ip (ipv4 dhcp)"
 	default n
 
+config DISABLE_EXTERNAL_AUTOCONNECT
+	bool "disable external autoconnect"
+	default n
+
+if DISABLE_EXTERNAL_AUTOCONNECT
+config WIFIMGR_INTERNAL_AUTOCONNECT
+	bool "enable internal auto(re)connect"
+	default y
+
+endif #WIFIMGR_DISABLE_EXTERNAL_AUTOCONNECT
+
 endif #WIFI_MANAGER
 
--- a/external/include/slsi_wifi/slsi_wifi_api.h
+++ b/external/include/slsi_wifi/slsi_wifi_api.h
@@ -421,6 +421,15 @@ int8_t WiFiSetCountryCode(const char *country_code);
  */
 int8_t WiFiSaveConfig(void);
 
+/**
+ * Request to set the autoconnect of external driver.
+ *	@check		1 for enabling and 0 for disabling external autoconnect
+ * Return: Completed successfully or failed
+ *
+ *   Will set external autoconnect enabled or disabled
+ */
+int8_t WiFiSetAutoconnect(uint8_t check);
+
 #undef EXTERN
 #ifdef  __cplusplus
 }
--- a/external/slsi_wifi/slsi_wifi_api.c
+++ b/external/slsi_wifi/slsi_wifi_api.c
@@ -3125,10 +3125,10 @@ static int8_t slsi_init(WiFi_InterFace_ID_t interface_id, const slsi_ap_config_t
 			slsi_set_updateconfig();
 			slsi_set_scan_interval(SLSI_SCAN_INTERVAL);
 			slsi_set_bss_expiration();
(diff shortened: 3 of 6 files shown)


seinfra:
Target : [32f9805d955e988de191d3882a45201e874ea833] - Code Rule Check (C++) OK.

seinfra:
Target : [32f9805d955e988de191d3882a45201e874ea833] - Code Rule Check OK.

framework/src/wifi_manager/Kconfig
```diff
@@ -70,5 +70,22 @@ config WIFIMGR_DISABLE_AUTO_GET_IP
 	bool "disable auto get ip (ipv4 dhcp)"
 	default n
 
+config WIFIMGR_DISABLE_EXTERNAL_AUTOCONNECT
```
pillip8282: TizenRT supports various wi-fi board. if we add this configuration then other boards also provide this feature. So I think removing dependency of the configuration would be better.

framework/src/wifi_manager/Kconfig
```diff
@@ -70,5 +70,22 @@ config WIFIMGR_DISABLE_AUTO_GET_IP
...
+	bool "disable external auto(re)connect of wifi driver"
+	default n
+
+if WIFIMGR_DISABLE_EXTERNAL_AUTOCONNECT
+config WIFIMGR_ENABLE_INTERNAL_AUTOCONNECT
```
pillip8282: Do we need this 'WIFIMGR_DISABLE_EXTERNAL_AUTOCONNECT'. I think WIFIDRIVER_SUPPORT_AUTOCONNECT can do same thing.

framework/src/wifi_manager/Kconfig
SeonghoByeon: @pillip8282 Recently, I have found that autoconnect of the Wi-Fi driver needs be avoided in certain applications. For example, if the autoconnect takes a long time, reconnecting only at the driver-level will prevent the application from making the necessary reconnection (extreme case, DHCP). On the other hand, the autoconnect from the driver sometimes makes the reconnection very easy and straightforward. Therefore, my recommandation is to guide the vender to provide this configuration. Please let me know your opinion. Thanks!

framework/src/wifi_manager/Kconfig
SeonghoByeon: @pillip8282 That is right. However, WIFIDRIVER_SUPPORT_AUTOCONNECT is defined in wifi_manager.c which is an internal variable and should be mapped to CONFIG_xxx, if it is controlled by the menuconfig.

s-bhuwalka:
I tested with the above patch. It is working fine without any issues.

SeonghoByeon:
Dear reviewers, @s-bhuwalka, @pillip8282, @Hyunjun85,
I have modified the structure of this PR, which removes dependency on the types of board. Specifically, instead of using config options utilized on both Wi-Fi Manager and external Wi-Fi driver, wifi_utils layer defines a new API to enable/disable the external autoconnect, which is directly called by Wi-Fi Manager.

I have tested this PR, and obtained successful results. @s-bhuwalka could you test the new version of PR once more in your environment? 

Thanks.

seinfra:
Target : [99e702d5805316be334947b9efde46284a706d5f] -

seinfra:
Target : [99e702d5805316be334947b9efde46284a706d5f] -

seinfra:
Target : [f9ef660bac13a0d0a17f727fba270cfa721fc390] - Code Rule Check OK.

seinfra:
Target : [f9ef660bac13a0d0a17f727fba270cfa721fc390] - Code Rule Check (C++) OK.

s-bhuwalka:
Ok I am checking

s-bhuwalka:
I tested with the new commit.It is working fine

PR #2247, opened 2018-09-17
driver/mtd: add winbond w25 driver

sunghan-chang:
This driver is back ported from NuttX.

Reason:
1. There is no separate winbond w25 driver
2. Current winbond W25Q64FV support in m25px driver is not good
   to continue in terms of both modularity and maintainability

Additional changes:
1. There is an assert during mtd read due to below additional code
   for checking write completion status.
   status = w25_waitwritecomplete(priv);
   DEBUGASSERT((status & (W25_SR_WEL | W25_SR_BP_MASK)) == 0);
   Hence this additional debug assert statement is removed.

2. There is an infinite travers of mtd nodes in mtd_register if
   the whole mtd is registered first in the driver and then
   if we try to register mtd partitions.
   - w25 driver is trying to register the whole mtd as below.
        mtd_register(&priv->mtd, "w25");
   - board initialization code would try to register mtd partitions as below.
        if (mtd) {
                mtd_register(mtd, "master");
        }

        if (mtd_fs) {
                mtd_register(mtd_fs, "FS");
        }

   Hence to keep it same as m25px driver "mtd_register(&priv->mtd, "w25");"
   call is removed.

### DIFF
--- a/os/fs/driver/Make.defs
+++ b/os/fs/driver/Make.defs
@@ -112,6 +112,10 @@ ifeq ($(CONFIG_MTD_M25P),y)
 CSRCS_DRIVER += mtd/m25px/m25px.c
 endif
 
+ifeq ($(CONFIG_MTD_W25),y)
+CSRCS_DRIVER += mtd/w25.c
+endif
+
 endif
 
 CSRCS += $(notdir $(CSRCS_DRIVER))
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [4a3d0c871fb49e8da0e5f95571ce84530f6aa14b] - Code Rule Check OK.

seinfra:
Target : [4a3d0c871fb49e8da0e5f95571ce84530f6aa14b] - Code Rule Check (C++) OK.

PR #2250, opened 2018-09-18
framework/wifi_manager: Security type and phy mode management

SeonghoByeon:
- Add enterprise security types into wifi_manager and wifi_utils
- Modify phy_mode to denote the supported standard version

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test_stress.c
+++ b/apps/examples/wifi_manager_sample/wm_test_stress.c
@@ -118,8 +118,9 @@ void wm_scan_done(wifi_manager_scan_info_s **scan_result, wifi_manager_scan_resu
 	}
 	wifi_manager_scan_info_s *wifi_scan_iter = *scan_result;
 	while (wifi_scan_iter != NULL) {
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [e695dffdfeed8030b0c76ea08a263159b956f3ec] - Code Rule Check (C++) OK.

seinfra:
Target : [e695dffdfeed8030b0c76ea08a263159b956f3ec] - Code Rule Check OK.

sunghan-chang:
@SeonghoByeon New modification, new commit makes dummy commit on history. When you get changes from another repo, and they should be synced each other, then you should use cherry pick instead of making new commit.

SeonghoByeon:
@sunghan-chang Thanks for the comment. In this case, there are lots of commits involved in the current unsynchronization, so the process of merge after porting each cherry-pick (step by step) is very complicated. Here, I tried to modify the code to avoid conflict with another repo as possible as I can, so please let me know if conflict occurs. Sorry for the delayed work for synchronization.

PR #2258, opened 2018-09-18
media: Modify media stream buffer config

junmin-kim:
Intergrate stream buffer config and add HANDLER prefix


seinfra:
Target : [973c8f7c4f54efff914ddac832350ef09ad0a274] - Code Rule Check OK.

seinfra:
Target : [973c8f7c4f54efff914ddac832350ef09ad0a274] - Code Rule Check (C++) OK.

build/configs/artik055s/audio/defconfig
```diff
@@ -412,13 +412,6 @@ CONFIG_AUDIO_MAX_INPUT_CARD_NUM=2
 CONFIG_AUDIO_MAX_OUTPUT_CARD_NUM=2
 CONFIG_AUDIO_MAX_DEVICE_NUM=3
 # CONFIG_AUDIO_PROCESSING_FEATURES is not set
-
-#
-# Audio Processing Features
-#
-# CONFIG_AUDIO_SPEECH_DETECT_FEATURES is not set
-# CONFIG_AUDIO_KEYWORD_DETECT is not set
-# CONFIG_AUDIO_ENDPOINT_DETECT is not set
```
chanijjani: Are these configs not used anymore..?

junmin-kim: I updated defconfig using `make menuconfig`. I don't care why this configs are removed.
Anyway these config are not set, so we don't consider them.

framework/src/media/Kconfig
```diff
@@ -78,24 +78,20 @@ config AUDIO_RESAMPLER_BUFSIZE
 	---help---
 		Buffer size for resampler
 
-config FILE_DATASOURCE_STREAM_BUFFER_SIZE
-	int "File DataSource stream buffer size"
+config CONFIG_HANDLER_STREAM_BUFFER_SIZE
```
sunghan-chang: 1. This causes `CONFIG_CONFIG_XXX` but you use `CONFIG_XXX` in source code. I am sure you don't want this.
2. From config name, I can't image where it belongs. How about add MEDIA prefix at name?

PR #2285, opened 2018-09-23
os/mm: use printf instead of echo

manoharahk:
In some shell configurations, newline character(\n) may
not work properly with echo command.

When HEAPINFO_USER_GROUP is enabled, it could generate
macros without newline in heapinfo_internal.h and resulting
in build error

This commit uses "printf" instead of "echo" in Makefile


seinfra:
Target : [9632e09563b0f9e2413903f421803669ac22b397] - Code Rule Check (C++) OK.

seinfra:
Target : [9632e09563b0f9e2413903f421803669ac22b397] - Code Rule Check OK.

PR #2288, opened 2018-09-27
os/net/lwip: Add a logic to handle broadcast address

SeonghoByeon:
- Add a logic to handle two types of broadcast address when assigning the network interface in lwip;
1) general baddr 0xffffffff
2) specific baddr belonging to subnet

Note that lwip fails to match a proper network interface without this modification, expecially when the NIC is mappted into the several (virtual) interfaces such as wl1 and wl2.


seinfra:
Target : [20c8f9392e6cd99fbd4fb0b1a078fb9769fea1c6] - Code Rule Check (C++) OK.

seinfra:
Target : [20c8f9392e6cd99fbd4fb0b1a078fb9769fea1c6] - Code Rule Check OK.

PR #2299, opened 2018-10-01
net, fs: fix compilation errors

sunghan-chang:
- Check definition of CONFIG_NET_LWIP first because LWIP_ARP and LWIP_ETHERNET are defined in lwip
CC:  net/lib_etherntoa.c
In file included from net/lib_etherntoa.c:60:0:
os/include/net/ethernet.h:101:5: error: "LWIP_ARP" is not defined [-Werror=undef]
 #if LWIP_ARP || LWIP_ETHERNET
     ^~~~~~~~
os/include/net/ethernet.h:101:17: error: "LWIP_ETHERNET" is not defined [-Werror=undef]
 #if LWIP_ARP || LWIP_ETHERNET
                 ^~~~~~~~~~~~~

- Cover them with CONFIG_NET_SOCKET conditioanl because they are valid with enabled it
CC:  vfs/fs_ioctl.c
In file included from /home/sunghan/Work/TinyAra/Forked_TizenRT/os/include/net/if.h:60:0,
                 from vfs/fs_ioctl.c:64:
os/include/sys/socket.h:306:36: error: 'struct msghdr' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 ssize_t recvmsg(int sockfd, struct msghdr *msg, int flags);
                                    ^~~~~~
os/include/sys/socket.h:307:36: error: 'struct msghdr' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 ssize_t sendmsg(int sockfd, struct msghdr *msg, int flags);
                                    ^~~~~~

- Initialize the errcode variable with invalid value, 0
CC:  vfs/fs_mkdir.c
In file included from vfs/fs_mkdir.c:61:0:
vfs/fs_mkdir.c: In function 'mkdir':
os/include/errno.h:132:33: error: 'errcode' may be used uninitialized in this function [-Werror=maybe-uninitialized]
 #define set_errno(e) do { errno = (int)(e); } while (0)
                                 ^
vfs/fs_mkdir.c:116:6: note: 'errcode' was declared here
  int errcode;
      ^~~~~~~

- Include <tinyara/spi/spi.h> which includes definition of 'struct spi_dev_s'
CC:  driver/mtd/mtd_config.c
In file included from driver/mtd/mtd_config.c:80:0:
os/include/tinyara/fs/mtd.h:269:50: error: 'struct spi_dev_s' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 FAR struct mtd_dev_s *m25p_initialize(FAR struct spi_dev_s *dev);
                                                  ^~~~~~~~~
os/include/tinyara/fs/mtd.h:280:49: error: 'struct spi_dev_s' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 FAR struct mtd_dev_s *w25_initialize(FAR struct spi_dev_s *dev);
                                                 ^~~~~~~~~


seinfra:
Target : [b080b42fef6a0c0efb42824c035dcde2a63f5f1d] - Code Rule Check OK.

seinfra:
Target : [b080b42fef6a0c0efb42824c035dcde2a63f5f1d] - Code Rule Check (C++) OK.

seinfra:
Target : [17eb15ed508476f089d7b6528e5f7ee8878c54d9] - Code Rule Check OK.

seinfra:
Target : [17eb15ed508476f089d7b6528e5f7ee8878c54d9] - Code Rule Check (C++) OK.

os/include/sys/socket.h
```diff
@@ -303,8 +303,10 @@ int getsockname(int sockfd, FAR struct sockaddr *addr, FAR socklen_t *addrlen);
 */
 int getpeername(int s, struct sockaddr *name, socklen_t *namelen);
 
+#ifdef CONFIG_NET_SOCKET
```
dr-venkman: In `os/net/recvmsg.c`, the `#ifdef` condition is based on `CONFIG_NET`. If you believe `CONFIG_NET_SOCKET` is the correct check, then recvmsg.c also needs modification.

os/include/sys/socket.h
```diff
@@ -303,8 +303,10 @@ int getsockname(int sockfd, FAR struct sockaddr *addr, FAR socklen_t *addrlen);
...
 ssize_t recvmsg(int sockfd, struct msghdr *msg, int flags);
 ssize_t sendmsg(int sockfd, struct msghdr *msg, int flags);
```
dr-venkman: Similar comment as above, except that the file is sendmsg.c.

os/include/sys/socket.h
sunghan-chang: 1. CONFIG_NET_SOCKET can be enabled when CONFIG_NET is enabled. That means even if CONFIG_NET is enabled, CONFIG_NET_SOCKET can be disabled.
2. In os/net/socket/Make.defs which determines compilation of source code, recvmsg.c file is built when CONFIG_NET_SOCKET is enabled.
```
ifeq ($(CONFIG_NET_SOCKET),y)
SOCK_CSRCS += net_sockets.c net_close.c net_dupsd.c net_dupsd2.c
SOCK_CSRCS += net_clone.c net_vfcntl.c bsd_socket_api.c
SOCK_CSRCS += recvmsg.c sendmsg.c
endif
```
In my opinion, CONFIG_NET_SOCKET is correct and ```#ifdef CONFIG_NET``` inside recvmsg.c is redundant because of Make.defs.

os/include/sys/socket.h
SeonghoByeon: CONFIG_NET_SOCKET depends on CONFIG_NET_LWIP. That means source files in os/net/socket are dependent on LWIP. On the other hand, prototypes defined in this include/sys/socket.h provide general function formats regardless of the type of network stack, I believe. My point is that in case recvmsg and sendmsg are dependent on LWIP, their prototype should be defined in os/net/socket.h not here. However, if recvmsg and sendmsg are also general types of function, then it should not be dependend on CONFIG_NET_SOCKET. Please let me know your opinion. Thank you!
In my opinion, those two functions recvmsg and sendmsg should be moved into socket.h in

sunghan-chang: I am not a network engineer so that it is difficult to say for me.
But
1. This commit fixs compilation error with keeping current network structure.
2. The recvmsg function just calls lwip API. It looks having strong dependancy with LWIP.
3. In POSIX standard, the recvmsg is in <sys/socket.h> like ours.
  If you want to support recvmsg, you must keep it here, not lwip's socket header.

PR #2321, opened 2018-10-10
driver/mtd : correct W25Q128JV configuration in m25px driver

manoharahk:
In some platforms m25px driver is used for winbond serial flash.
and M25P128 configuration is not compatible with Q type W25Q128JV.

This patch adds a valid configuration for W25Q128x serial flash

### DIFF
--- a/os/fs/driver/mtd/m25px/m25px.c
+++ b/os/fs/driver/mtd/m25px/m25px.c
@@ -113,6 +113,10 @@
 #define M25P_M25P128_CAPACITY      0x18	/* 128 M-bit */
 #define M25P_M25P256_CAPACITY      0x19	/* 256 M-bit */
 
+/* Winbond flash & Memory types */
+#define WINBOND_SFLASH_ID          0xEF
+#define WINBOND_W25Q128_TYPE       0x40
+
 /*  M25P1 capacity is 131,072 bytes:
  *  (4 sectors) * (32,768 bytes per sector)
  *  (512 pages) * (256 bytes per page)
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [e03b2dfe0bb7f11fd3ed3f358455e553d0a523b6] - Code Rule Check (C++) OK.

seinfra:
Target : [e03b2dfe0bb7f11fd3ed3f358455e553d0a523b6] - Code Rule Check OK.

os/fs/driver/mtd/m25px/m25px.c
```diff
@@ -176,6 +180,22 @@
 #define M25P_M25P128_PAGE_SHIFT    8	/* Page size 1 << 8 = 256 */
 #define M25P_M25P128_NPAGES        65536
 
+/* W25Q128JV  capcity :
```
thapav: capacity*

manoharahk: Sorry It was a typo. Thanks for pointing out.

seinfra:
Target : [2d13cee90c39df1fc43221399673b47318cebab3] - Code Rule Check OK.

seinfra:
Target : [2d13cee90c39df1fc43221399673b47318cebab3] - Code Rule Check (C++) OK.

os/fs/driver/mtd/m25px/m25px.c
```diff
@@ -413,10 +433,26 @@ static inline int m25p_readid(struct m25p_dev_s *priv)
 		} else if (capacity == M25P_M25P128_CAPACITY) {
 			/* Save the FLASH geometry */
 
-			priv->sectorshift = M25P_M25P128_SECTOR_SHIFT;
-			priv->nsectors = M25P_M25P128_NSECTORS;
-			priv->pageshift = M25P_M25P128_PAGE_SHIFT;
-			priv->npages = M25P_M25P128_NPAGES;
+			if (manufacturer == WINBOND_SFLASH_ID &&
```
sunghan-chang: Line 377,
```
	if (manufacturer == M25P_MANUFACTURER && memory == M25P_MEMORY_TYPE) {
```
This covers here, I think this never calls. Could you check it?

manoharahk: As you know, it is to support winbond flash inside M25px driver.
M25P_MANUFACTURER  & MEMORY_TYPE are configurable. 
Lets say it is set to M25P128 chip, manufacture = C2, memory = 20, then for capacity 128 we need to M25P128 configurations. 
For W25Q128 case, we explicitly check for winbond id & mem type and use W25Q128 configurations.

sunghan-chang: I don't understand what you mean. Anyway, Line 377 and line 436 use same variables, `manufacturer` and `memory`. How do they have different value? That is not a bit comparision.

sunghan-chang: Oh ok. I understood. Line 377 is a comparision between read values from SPI and configured values but line 436 is a comparision between read values and winbond values.
It looks little strange on readability. Because of definitions of M25P_MANUFACTURER and WINBOND_SFLASH_ID, I thought they are definitely different values.

If we support w25 in m25 feature, do we need to keep w25.c?

manoharahk: Yes. This supports only a single winbond flash in m25px driver whereas w25.c is for all winbond family flash's.
As you know, this requirement came from one of our issue in our private repo where we have used m25px driver for winbond flash(for some reason).  For code sync, I had to push this patch here.

PR #2331, opened 2018-10-12
Get the current location from script file path on shell scripts

sunghan-chang:
Many sh files get a current path using PWD command. But it
is changed where you run it.
For example, if you run a shell script on aa/bb folder, PWD is aa/bb.
But if you run it like bb/xx.sh on aa folder, PWD is aa.
The location of executing shell script affects working of shell script.

This commit gets the current path from location of shell script which
is fixed. Wherever you run it, it works.

### DIFF
--- a/build/configs/artik05x/artik05x_download.sh
+++ b/build/configs/artik05x/artik05x_download.sh
@@ -20,10 +20,13 @@
 # File   : artik05x_download.sh
 # Description : Download script for ARTIK 05X
 
-source .config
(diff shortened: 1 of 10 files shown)


seinfra:
Target : [84cb07e27220b91be40ae45fa70ffd4f15ebdf0c] - Code Rule Check (C++) OK.

seinfra:
Target : [84cb07e27220b91be40ae45fa70ffd4f15ebdf0c] - Code Rule Check OK.

PR #2333, opened 2018-10-12
media: Return 0 when OutputDatasource size is 0

junmin-kim:
0 length fwrite(FileOutputDataSource) or array(BufferOutputSource)
are not verified. So just return 0 if size is 0

### DIFF
--- a/framework/src/media/BufferOutputDataSource.cpp
+++ b/framework/src/media/BufferOutputDataSource.cpp
@@ -78,6 +78,10 @@ bool BufferOutputDataSource::isPrepare()
 
 ssize_t BufferOutputDataSource::write(unsigned char *buf, size_t size)
 {
+	if (size == 0) {
+		return 0;
+	}
+
 	if (!isPrepare()) {
 		return EOF;
 	}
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [38439d823913aa9c0fe4b4645f68e61a99b48721] - Code Rule Check OK.

seinfra:
Target : [38439d823913aa9c0fe4b4645f68e61a99b48721] - Code Rule Check (C++) OK.

framework/src/media/FileOutputDataSource.cpp
```diff
@@ -113,6 +113,10 @@ ssize_t FileOutputDataSource::write(unsigned char *buf, size_t size)
 		return EOF;
 	}
 
+	if (size == 0) {
```
davidfather: Why don't you add this fix to SocketOutputDataSource as well?

framework/src/media/BufferOutputDataSource.cpp
```diff
@@ -86,6 +86,10 @@ ssize_t BufferOutputDataSource::write(unsigned char *buf, size_t size)
 		return EOF;
 	}
 
+	if (size == 0) {
```
Taejun-Kwon: Move this  state to top of write function.
Cheapest first, It is good habit.

seinfra:
Target : [c7ea659ad6f99df300b3e3ebc0967d458f55b1bd] - Code Rule Check OK.

seinfra:
Target : [c7ea659ad6f99df300b3e3ebc0967d458f55b1bd] - Code Rule Check (C++) OK.

PR #2340, opened 2018-10-16
framework/task_manager: Add state update logic before getinfo

jeongchanKim:
Sometimes task state is not updated instantly. So before getting task information, task state should be updated.
i.e. Normal termination of task will not update the task state itself.


seinfra:
Target : [7cac7703e1f745213e4ffef516615f9f37b6cda4] - Code Rule Check (C++) OK.

seinfra:
Target : [7cac7703e1f745213e4ffef516615f9f37b6cda4] - Code Rule Check OK.

PR #2347, opened 2018-10-17
Resize number_buffer array in cJSON.c to solve the buffer overrun issue.

HONGCHAEHEE:
iperf app uses cJSON library sending and receiving app results including the bytes sent.
When you send the large bytes to a server, especially using 5GHz WiFi,
the 'number_buffer' array, which is temporary used to print the number in cJSON, can be overrun.
That will cause sending results error, and you will see the log : "error - unable to send results".
The size of 'number_buffer' array was set to 27 to solve the error, that is enough to get the large bytes.
If you encounter the same problem, however, check the 'number_buffer' array size again in cJSON.

### DIFF
--- a/apps/system/iperf/README.txt
+++ b/apps/system/iperf/README.txt
@@ -1,0 +1,13 @@
+iperf
+^^^^^
+
+Warning:
+
+When you are running iperf app on a client mode, if you get the sending results error, check the 'number_buffer' array size in cJSON.c.
+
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [62ff8c5a70571e812e04d44bf97bb369f4aa64f8] - Code Rule Check OK.

seinfra:
Target : [62ff8c5a70571e812e04d44bf97bb369f4aa64f8] - Code Rule Check (C++) OK.

PR #2350, opened 2018-10-17
framework/media: Modify comparison condition of MediaRecorder

dev-hue:
Modify comparison condition to guarantee the re-setting duration of MediaRecorder.

At first, Set the duration of MediaRecorder to 3 seconds.
And then set it back to 0 for endless recording.
But, Under the current comparison condition, it can't be back to 0.

### DIFF
--- a/framework/src/media/MediaRecorderImpl.cpp
+++ b/framework/src/media/MediaRecorderImpl.cpp
@@ -187,7 +187,7 @@ void MediaRecorderImpl::prepareRecorder(recorder_result_t& ret)
 		return notifySync();
 	}
 
-	if (mDuration > 0) {
+	if (mDuration >= 0) {
 		mTotalFrames = mDuration * source->getSampleRate();
 	}
 
@@ -552,8 +552,8 @@ void MediaRecorderImpl::setRecorderDuration(int second, recorder_result_t& ret)
 		ret = RECORDER_ERROR_INVALID_STATE;
 		return notifySync();
 	}
-	if (second > 0) {
-		medvdbg("second is greater than zero, set limit : %d\n", second);
+	if (second >= 0) {
+		medvdbg("second is greater than or eqaul to zero, set limit : %d\n", second);
 		mDuration = second;
 	}
 


seinfra:
Target : [b8239ff1a9de09d2a41225e425082878033a4344] -

seinfra:
Target : [12fd5917c6c39e2e8f050bef158db62e8b7d50aa] - Code Rule Check (C++) OK.

seinfra:
Target : [12fd5917c6c39e2e8f050bef158db62e8b7d50aa] - Code Rule Check OK.

framework/src/media/MediaRecorderImpl.cpp
```diff
@@ -187,7 +187,7 @@ void MediaRecorderImpl::prepareRecorder(recorder_result_t& ret)
 		return notifySync();
 	}
 
-	if (mDuration > 0) {
+	if (mDuration >= 0) {
```
Taejun-Kwon: Instead of this, we had better return error if duration is lower than 1.

dev-hue: @Taejun-Kwon But, zero means endless recording. I think that error return should be reconsidered.
And when I think about agian, actually, eqaul condition is meaningless.
Because After calling `prepare()`, `setDuration()` is useless.

Taejun-Kwon: I mean if setDuration(0) comes, return error.

dev-hue: @Taejun-Kwon 
First, Returning error when setDuration(0) come must be reconsidered.

On current version, As I mentioned that zero means endless(or infinite) recording.

Let's think about below case.

> setDuration(3);
> setDuration(0);
> prepare();
> start();

Above mediarecorder should record for 3 seconds.
Because of the current comparison condition.
To change from 3 seconds to 0 seconds on current version, you need to do :

> setDuration(3);
> prepare();
> unprepare();
> prepare();
> start();

This case does not occur in static situation(or code), but in dynamic.
For supporting to set 0 sec(actually, infinite recording) without prepare()&unprepare(), we should change the comparison condition of setRecorderDuration.

Second, This change is meaningless. 
Because this line is the prepare() code. If mDuration is 0, mTotalFrame is also 0. 
This line will be meaningfuly only if mDuration is greater than 0.

Taejun-Kwon: I think there was miscommunication between you and I.
So you want to set 0 as a deactivates limitation, right?
what I meaning is that 
Not 'if (mDuration ==0)', I mean 'if (second == 0)'
In your above case, of course it should play for 3 secs but when it calls setduration(0), it will return error regarding invalid param directly.
But I think disable limit values also good options.

dev-hue: @Taejun-Kwon 
I made a new PR that contains new APIs and contents of this review.

dev-hue:
New PR is registered.

PR #2381, opened 2018-10-23
framework/task_manager: Add NULL check procedure

tyha1115:
When user uses task_manager_set_broadcast_cb() with 'NULL' value as the cb_data, an assert problem occurs.
In order to prevent this problem, NULL check procedure is added.


seinfra:
Target : [81d057d850defdfc0e689a085901f68bb859904d] - Code Rule Check OK.

seinfra:
Target : [81d057d850defdfc0e689a085901f68bb859904d] - Code Rule Check (C++) OK.

PR #2397, opened 2018-10-29
media: Initialize audio_caps_desc_s

junmin-kim:
audio_caps_desc_s should be initialized as zero before read/write

### DIFF
--- a/framework/src/media/audio/audio_manager.c
+++ b/framework/src/media/audio/audio_manager.c
@@ -317,6 +317,8 @@ static audio_manager_result_t get_supported_sample_rate(audio_io_direction_t dir
 	if (ret != AUDIO_MANAGER_SUCCESS) {
 		return ret;
 	}
+
+	memset(&caps_desc, 0, sizeof(struct audio_caps_desc_s));
 	if (direct == INPUT) {
 		card = &g_audio_in_cards[g_actual_audio_in_card_id];
 		caps_desc.caps.ac_type = AUDIO_TYPE_INPUT;
@@ -535,6 +537,7 @@ static audio_manager_result_t get_audio_volume(audio_io_direction_t direct)
 	audio_config_t *config;
 	char card_path[AUDIO_DEVICE_FULL_PATH_LENGTH];
 
+	memset(&caps_desc, 0, sizeof(struct audio_caps_desc_s));
 	caps_desc.caps.ac_len = sizeof(struct audio_caps_s);
 	caps_desc.caps.ac_type = AUDIO_TYPE_FEATURE;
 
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [5acee402a4f5a7d7910f6fed86e79b14536e1497] - Code Rule Check (C++) OK.

seinfra:
Target : [5acee402a4f5a7d7910f6fed86e79b14536e1497] - Code Rule Check OK.

junmin-kim:
If driver code use caps like below, `caps->member` will be a garbage
```
caps->member |= value;
```

framework/src/media/audio/audio_manager.c
```diff
@@ -1429,6 +1434,8 @@ audio_manager_result_t find_stream_in_device_with_process_type(device_process_ty
 
 	process_type = get_process_type_audio_param_value(type);
 	subprocess_type = get_subprocess_type_audio_param_value(subtype);
+
+	memset(&caps_desc, 0, sizeof(struct audio_caps_desc_s));
```
Taejun-Kwon: I think memset is not neccessary because this is local structure.
If there are any problem relatedto cpas_desc, you should check line 1469 first.

Taejun-Kwon: I don't agree with you, please think about role of caller & callee one more time.
caps_dec is kind of object to get data from driver, so it doesn't know what is default value. 
So driver should handle this.
```
static int DRIVER_getcaps(..., struct audio_caps_s *caps)
{
caps->ac_control.b[0] = AUDIO_PU_UNDEF;
if (foo) {
    caps->ac_controls.b[0] |= AUDIO_PU_SPEECH_DETECT;
}
...
}
```

junmin-kim: I got it. Then I'll fix the driver code

junmin-kim: I have a question.
What if `spi driver` control `AUDIO_TYPE_PROCESSING ac_type` case, but `i2s driver` doesn't?
Should all our driver code have codes that control all Audio Device Types?

```cpp
static int spi_getcpas()
{
    switch(caps->ac_types) {
    case AUDIO_TYPE_PROCESSING:
        caps->ac_control.b[0] = AUDIO_PU_SPEECH_DETECT;
        break;
    default:
        break;
    }
}

static int i2s_getcpas()
{
    switch(caps->ac_types) {
    /* Below case doesn't exist */
    /* But caps->ac_controls.b[0] should be set */
    //case AUDIO_TYPE_PROCESSING:
    //    break;
    default:
        break;
    }
}

```

junmin-kim: In AUDIOIOC_GETCAPS case,
we can control like below. How do you think?
```cpp
if (caps_desc.caps.ac_subtype == 0 && caps_desc.caps.ac_channels == 0) {
		meddbg("Unsupported ac_type\n");
		caps_desc.caps.ac_subtype = process_type;
		close(fd);
		pthread_mutex_unlock(&(card->card_mutex));
		continue;
	}
```

junmin-kim: I think getcaps should return -1 and set errno as EINVAL.
How do you think?

Taejun-Kwon: I already defined UNDEF, please check audio.h

junmin-kim: Could you explain more?
I can't catch it

PR #2418, opened 2018-11-01
audio: Check AUDIO_MSG_COMPLETE

junmin-kim:
Add logic for AUDIO_CALLBACK_COMPLETE

### DIFF
--- a/framework/src/tinyalsa/tinyalsa.c
+++ b/framework/src/tinyalsa/tinyalsa.c
@@ -671,6 +671,8 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 			} else if (msg.msgId == AUDIO_MSG_XRUN) {
 				/* Underrun to be handled by client */
 				return -EPIPE;
+			} else if (msg.msgId == AUDIO_MSG_COMPLETE) {
+				return pcm_bytes_to_frames(pcm, pcm_frames_to_bytes(pcm, frame_count) - pending);
 			} else {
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [648f7816eacfd86ed365b907bcb4881f6c0f42d2] - Code Rule Check OK.

seinfra:
Target : [648f7816eacfd86ed365b907bcb4881f6c0f42d2] - Code Rule Check (C++) OK.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -671,6 +671,8 @@ int pcm_readi(struct pcm *pcm, void *data, unsigned int frame_count)
 			} else if (msg.msgId == AUDIO_MSG_XRUN) {
 				/* Underrun to be handled by client */
 				return -EPIPE;
+			} else if (msg.msgId == AUDIO_MSG_COMPLETE) {
```
Taejun-Kwon: after this time and if one more request comes, we should return specific error, -ESTRPIPE
so below is needed
1. add additional definition in audio.h 
ex) AUDIO_MSG_DISCONNECTED
2. add definition in audio_manager
ex) AUDIO_MANAGER_DEVICE_SUSPENDED 
3. Tinyalsa should handle msg from audio driver here.
4. handle -ESTRPIPE(if it is not in errno.h then we have to add), return AUDIO_MANAGER_DEVICE_SUSPENDED 
5. Media Recorder should handle AUDIO_MANAGER_DEVICE_SUSPENDED.
    If it is returned, stop automatically(do not treat it as a failure)
6. If device is already closed, or some thread already destroyed, return -EDEV
    In this case we can treat it as a fail case.

PR #2437, opened 2018-11-09
C library: modify sscanf to read more than 128byte

shadaksharikp:
C library: modify sscanf to read more than 128byte string

Modify C library vsscanf() as described below.
1) Due to MAXLN of 128 bytes, sscanf scan only upto 128 bytes string. This is issue reported by
ST things team since ST things SDK needs to scan more than 2K data.
2) So remove the 128 byte upper limit for string case, Hence it can scan any string range
3)Test procedure:
--------------------------
char *buf = "ABCDXXXX..... . .. .Z"(say 555 char)
char tmp[124];

ret = sscanf(buf,"%s", tmp)
printf("Post scan: %s\n", tmp);
--------------------------
a) O/P (without fix) : 128 characters only
b) O/P (with fix) : All 555 chars

Signed-off-by: Shadakshari K P <shari.kps@samsung.com>

### DIFF
--- a/lib/libc/stdio/lib_sscanf.c
+++ b/lib/libc/stdio/lib_sscanf.c
@@ -70,7 +70,6 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-
 #define MAXLN 128
 
 #ifndef MIN
@@ -452,7 +451,6 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
 						width  = fwidth;
 					}
 
-					width = MIN(sizeof(tmp) - 1, width);
 
 					/* Copy the string (if we are making an assignment) */
 
@@ -511,7 +509,6 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
 						width  = fwidth;
 					}
 
-					width = MIN(sizeof(tmp) - 1, width);
 
 					/* Copy the string (if we are making an assignment) */
 


seinfra:
Target : [2ca5bef72544a75fe7d16ed4f595e8475e243b3e] - Code Rule Check OK.

seinfra:
Target : [2ca5bef72544a75fe7d16ed4f595e8475e243b3e] - Code Rule Check (C++) OK.

thapav:
Please add CONFIG_DYNAMIC_ALLOC_ENABLE to the Kconfig. 
It is not present currently.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -70,8 +70,11 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+#define CONFIG_DYNAMIC_ALLOC_ENABLE
```
sunghan-chang: Let's make it in Kconfig. If you want to define this in here, please remove `CONFIG_`. But I think this is difficult to find and to control.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -70,8 +70,11 @@
...
 
+#ifndef CONFIG_DYNAMIC_ALLOC_ENABLE
```
sunghan-chang: As I gave my opinion to Sangam, how about use both, static 128 buffer and dynamic buffer?
Almost sscanf is used with small data, if we support only dynamic way, it causes performace down always.

If there is a way to measure the size of input, we can use dynamic way only when size is bigger than 128.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -452,7 +466,9 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
 						width  = fwidth;
 					}
 
+#ifndef CONFIG_DYNAMIC_ALLOC_ENABLE
 					width = MIN(sizeof(tmp) - 1, width);
```
sangam-swami: I guess the above width limitation (128 byte) causing the string copy failure to tv buffer. Can we modify MIN check and allow long buffer copy.
We can keep tmp check and tmp buffer conversion same other than in this %s case.

kishore-sn: As mentioned by Sangam, the only cause of failure seems to be the statement on line 470.

width = MIN(sizeof(tmp) - 1, width);

This check is not required, since the succeeding strncpy does not copy into the variable tmp. 
If we remove this line, the issue with "%s" will be fixed. 
All other cases are already working properly and hence all other changes done under CONFIG_DYNAMIC_ALLOC_ENABLE can be reverted. 

Also there is no requirement to perform malloc in any case.
Hence, please make the above change and revert all other previous changes.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -511,7 +527,9 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
...
+#endif
```
kishore-sn: This is similar to the "%s" case. In this case we are using regex based operations to determine the length of the input characters to be scanned. In any case, the code in line 531 needs to be removed, as we are not copying data into the variable "tmp". Instead we copy data into the variable "tv".

seinfra:
Target : [2347c4b10e4918754467287437840c44e404636b] -

seinfra:
Target : [2347c4b10e4918754467287437840c44e404636b] -

seinfra:
Target : [78e56975ad20ff59310cbc1e8a3adda052dbee0e] - Code Rule Check (C++) OK.

seinfra:
Target : [78e56975ad20ff59310cbc1e8a3adda052dbee0e] - Code Rule Check OK.

seinfra:
Target : [34bb96d9f663d5e5c604d083e2061dff637aa4ab] - Code Rule Check (C++) OK.

seinfra:
Target : [34bb96d9f663d5e5c604d083e2061dff637aa4ab] - Code Rule Check OK.

lib/libc/stdio/lib_sscanf.c
shadaksharikp: Thanks
filtered other changes except string handling case.

lib/libc/stdio/lib_sscanf.c
shadaksharikp: Thanks
filtered other changes except string handling case.

seinfra:
Target : [9ee2a8c4baada9a7d09bd70daf9ea77c76ca4018] - Code Rule Check OK.

seinfra:
Target : [9ee2a8c4baada9a7d09bd70daf9ea77c76ca4018] - Code Rule Check (C++) OK.

PR #2454, opened 2018-11-19
testcase/task_manager: Reverse the wrong param for task_manager_regis…

jeongchanKim:
…ter_task TC

The 4th param of task_manager_register_task should be entry point for task.


seinfra:
Target : [5fce934b79021cedfa6d7d13734e1cb09bdefc0e] - Code Rule Check OK.

PR #2474, opened 2018-11-22
mm/realloc: Fix realloc when alloc to another heap and Fix mm_get_hea…

jeongchanKim:
…pindex

1. When realloc to another heap, we should free the previous mem
2. If mem is NULL in mm_get_heapindex, we cannot find the heapindex. In this case, mm_get_heapindex returns 0(first heap index).


seinfra:
Target : [288dbee9dcfef559beda4bf62670372270946f52] - Code Rule Check (C++) OK.

seinfra:
Target : [288dbee9dcfef559beda4bf62670372270946f52] - Code Rule Check OK.

PR #2493, opened 2018-11-26
examples/slsiwifi: Fix build error for new heap

jeongchanKim:
mm_get_heap_info() was changed to pass the memory address for checking heap index
heapinfo_parse() was changed to pass 3-params

### DIFF
--- a/apps/examples/slsiwifi/slsiwifi_main.c
+++ b/apps/examples/slsiwifi/slsiwifi_main.c
@@ -104,7 +104,7 @@ static struct mm_heap_s *g_user_heap;
 int g_memstat_total = 0;
 static int getMemUsage(void)
 {
-	g_user_heap = mm_get_heap_info();
+	g_user_heap = mm_get_heap_info(GET_BASE_HEAP_ADDR);
 	printf("\n\tCurrent memory usage (total): %d bytes\n", g_user_heap->total_alloc_size);
 	return g_user_heap->total_alloc_size;
 }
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [056121c64b38e213ed85e75cf41bb3bf7aee19d7] - Code Rule Check OK.

apps/examples/slsiwifi/slsiwifi_main.c
```diff
@@ -104,7 +104,7 @@ static struct mm_heap_s *g_user_heap;
 int g_memstat_total = 0;
 static int getMemUsage(void)
 {
-	g_user_heap = mm_get_heap_info();
+	g_user_heap = mm_get_heap_info(NULL);
```
sunghan-chang: NULL? Why NULL? Even if it is working with NULL, it looks strange. How about using new definition instead of NULL?

seinfra:
Target : [54edaec28bc98b46345447781c7fcacf8c12eaea] - Code Rule Check (C++) OK.

seinfra:
Target : [54edaec28bc98b46345447781c7fcacf8c12eaea] - Code Rule Check OK.

sunghan-chang:
@jeongchanKim I don't agree adding `GET_BASE_HEAP_ADDR` on mm.h. That's a quick fix of slsiwifi. That is not good for public support. Let's move it into slsiwifi and use it inside slsiwifi only.

PR #2511, opened 2018-11-30
external/iotivity: Force delete libcoap.a from external/libcoap for 'make clean'.

Hyunjun85:
The 'make clean' will show below message and wait the user confirm.
'rm: remove write-protected regular file '*/external/iotivity/iotivity_1.2-rel/extlibs/libcoap/libcoap.a'?'

But it doesn't need and it will not be showing when user uses './dbulid.sh menu' to clean.
So the rm command is able to use '-f' option.


Hyunjun85:
@SenthilKumarGS Hi. 
When I an trying build and clean using make command instead of './dbuild.sh menu'.
I can see a message which 'rm: remove write-protected~~'.
How about this? Could you please let me know your opinion.

SenthilKumarGS:
@Hyunjun85 It can be merged.

PR #2557, opened 2018-12-19
external/mbedtls: update mbedtls patch

an4967:
### 1. artik05x: Remove SHA2_224 hw support
sssfw does not support SHA2_224, but there is SHA2_224 in source code.

### 2. mbedtls : memory leak (x509_crl)
When fail to parse crl, do not free pem resource.

reference :
https://www.bountysource.com/issues/40064024-memory-leak-in-mbedtls_x509_crl_parse

### 3. net/tls: Fix implementation of VERIFY_OPTIONAL verification mode
This commit changes the behaviour of mbedtls_ssl_parse_certificate
to make the two authentication modes MBEDTLS_SSL_VERIFY_REQUIRED and
MBEDTLS_SSL_VERIFY_OPTIONAL be in the following relationship:

    Mode == MBEDTLS_SSL_VERIFY_REQUIRED <=> Mode == MBEDTLS_SSL_VERIFY_OPTIONAL + check verify result

Also, it changes the behaviour to perform the certificate chain
verification even if the trusted CA chain is empty. Previously, the
function failed in this case, even when using optional verification,
which was brought up in Samsung#864.
(Samsung#864 ARMmbed/mbedtls#864)

### DIFF
--- a/external/mbedtls/pk_wrap.c
+++ b/external/mbedtls/pk_wrap.c
@@ -623,9 +623,6 @@ int hw_rsa_sign_wrap(mbedtls_rsa_context *ctx, mbedtls_md_type_t md_alg, const u
 	case MBEDTLS_MD_SHA1:
 		rsa_sign.alg_type = SHA1_160 | (padding ^ 0x1);	/* 0x110X */
 		break;
-	case MBEDTLS_MD_SHA224:
-		rsa_sign.alg_type = SHA2_224 | (padding ^ 0x1);	/* 0x220X */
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [d3a190a5cbd256613f18a2f693247234208aa223] - Code Rule Check (C++) OK.

seinfra:
Target : [d3a190a5cbd256613f18a2f693247234208aa223] - Code Rule Check OK.

PR #2587, opened 2019-01-08
framework/media: Fix issues regarding wave header

zhouxinhe:
1. fix memory leaked issue.
2. fix strcpy overflow issue, using safe strncpy.
Above issues are reported by SVACE
3. move fclose from createWavHeader() to FileOutputDataSource::open()
4. ffush stream buffer before ftell to get correct file size
@Taejun-Kwon @YiHyunJin

### DIFF
--- a/framework/src/media/FileOutputDataSource.cpp
+++ b/framework/src/media/FileOutputDataSource.cpp
@@ -72,6 +72,12 @@ bool FileOutputDataSource::open()
 		case AUDIO_TYPE_WAVE:
 			if (!utils::createWavHeader(mFp)) {
 				meddbg("wav header create failed\n");
+				if (fclose(mFp) == OK) {
+					mFp = nullptr;
+				} else {
+					meddbg("file close failed error : %d\n", errno);
+				}
+				return false;
 			}
 			break;
 		default:
@@ -88,17 +94,19 @@ bool FileOutputDataSource::open()
 bool FileOutputDataSource::close()
 {
 	switch (getAudioType()) {
-		case AUDIO_TYPE_WAVE:
-			unsigned int fileSize;
-			fileSize = ftell(mFp);
-			if (fileSize < 0) {
-				meddbg("file size could not be found\n");
+		case AUDIO_TYPE_WAVE: {
+			fflush(mFp);
+			long ret = ftell(mFp);
+			if (ret < 0) {
+				meddbg("file size could not be found errno : %d\n", errno);
 				break;
 			}
+			unsigned int fileSize = (unsigned int)ret;
 			if (!utils::writeWavHeader(mFp, getChannels(), getSampleRate(), getPcmFormat(), fileSize)) {
 				meddbg("wav header write to failed\n");
 			}
 			break;
+		}
 		default:
 			/* Don't set any encoder for unsupported formats */
 			break;
--- a/framework/src/media/utils/MediaUtils.cpp
+++ b/framework/src/media/utils/MediaUtils.cpp
@@ -527,9 +527,6 @@ bool createWavHeader(FILE *fp)
 	header = (struct wav_header_s *)malloc(sizeof(struct wav_header_s));
 	if (!header) {
 		meddbg("fail to malloc buffer\n");
-		if (fclose(fp) == 0) {
-			fp = nullptr;
-		}
 		return false;
 	}
 
@@ -538,15 +535,16 @@ bool createWavHeader(FILE *fp)
 	ret = fwrite(header, sizeof(unsigned char), WAVE_HEADER_LENGTH, fp);
 	if (ret < 0) {
 		meddbg("file write failed error %d\n", errno);
+		free(header);
 		return false;
 	}
 	if (fseek(fp, WAVE_HEADER_LENGTH, SEEK_SET) != 0) {
 		meddbg("file seek failed error\n");
+		free(header);
 		return false;
 	}
-	if (header != NULL) {
-		free(header);
-	}
+
+	free(header);
 	return true;
 }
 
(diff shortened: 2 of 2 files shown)


seinfra:
Target : [9c5311a45f41067723f8c59e5e0187dc662e8587] - Code Rule Check OK.

seinfra:
Target : [9c5311a45f41067723f8c59e5e0187dc662e8587] - Code Rule Check (C++) OK.

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -538,15 +535,16 @@ bool createWavHeader(FILE *fp)
 	ret = fwrite(header, sizeof(unsigned char), WAVE_HEADER_LENGTH, fp);
 	if (ret < 0) {
```
JeongJunSik: i think `ret != WAVE_HEADER_LENGTH ` it's better.

JeongJunSik: I modify and upload another PR #2588

framework/src/media/FileOutputDataSource.cpp
```diff
@@ -89,13 +95,13 @@ bool FileOutputDataSource::close()
 {
 	switch (getAudioType()) {
 		case AUDIO_TYPE_WAVE:
-			unsigned int fileSize;
+			long fileSize;
```
Taejun-Kwon: uint16_t?

framework/src/media/FileOutputDataSource.cpp
```diff
@@ -89,13 +95,13 @@ bool FileOutputDataSource::close()
...
 			fileSize = ftell(mFp);
 			if (fileSize < 0) {
 				meddbg("file size could not be found\n");
 				break;
 			}
-			if (!utils::writeWavHeader(mFp, getChannels(), getSampleRate(), getPcmFormat(), fileSize)) {
+			if (!utils::writeWavHeader(mFp, getChannels(), getSampleRate(), getPcmFormat(), (unsigned int)fileSize)) {
```
Taejun-Kwon: we had better modify parameter of writeWavHeader, itself

framework/src/media/FileOutputDataSource.cpp
zhouxinhe: At old line 94, expression `fileSize < 0` is always false.
We need long type variable to accept return value of `ftell`.

zhouxinhe: How about this?
```c
long ret = ftell(mFp);
if (ret < 0) {
    meddbg(...);
    break;
}
unsigned int fileSize = (unsigned int)ret;
if (!utils::writeWavHeader(mFp, getChannels(), getSampleRate(), getPcmFormat(), fileSize)) {
    ...
}
```

framework/src/media/utils/MediaUtils.cpp
zhouxinhe: good

framework/src/media/FileOutputDataSource.cpp
Taejun-Kwon: I think it is good idea.

zhouxinhe: Fixed.

seinfra:
Target : [86a383ed6797345b4802bf06886d71db3ae2c21b] - Code Rule Check (C++) OK.

seinfra:
Target : [86a383ed6797345b4802bf06886d71db3ae2c21b] - Code Rule Check OK.

framework/src/media/FileOutputDataSource.cpp
```diff
@@ -88,17 +94,18 @@ bool FileOutputDataSource::open()
 bool FileOutputDataSource::close()
 {
 	switch (getAudioType()) {
-		case AUDIO_TYPE_WAVE:
-			unsigned int fileSize;
-			fileSize = ftell(mFp);
-			if (fileSize < 0) {
+		case AUDIO_TYPE_WAVE: {
+			long ret = ftell(mFp);
```
Taejun-Kwon: As I know, bofore call fsync(when sector is filled with data), ftell will return wrong value.
How about maintain variable and update it when write called?

zhouxinhe: `USER -> fwrite -> STDIO-BUFFER -> fflush -> SYSTEM-BUFFER -> fsync -> FLASH`
There may be data cached in STDIO-BUFFER has not be flushed into SYSTEM-BUFFER,
and `ftell` get position(file size) from SYSTEM-BUFFER, so the size may be less than USER write.
Is it what you mean?

zhouxinhe: @Taejun-Kwon  what's your opinion?

Taejun-Kwon: fflush write remained data in buffer, but sync is called if next sector is needed because of data is bigger than remained area in current sector.
But As per you wrote in above, ftell using position of stream buffer, I think it is enough.

Taejun-Kwon:
Could you Please modify title?
ex. Fix issues regarding wave header..blahblah

framework/src/media/FileOutputDataSource.cpp
```diff
@@ -88,17 +94,19 @@ bool FileOutputDataSource::open()
...
 {
 	switch (getAudioType()) {
-		case AUDIO_TYPE_WAVE:
-			unsigned int fileSize;
-			fileSize = ftell(mFp);
-			if (fileSize < 0) {
+		case AUDIO_TYPE_WAVE: {
+			fflush(mFp);
+			long ret = ftell(mFp);
+			if (ret < 0) {
```
Taejun-Kwon: please add more info for error case, 
meddbg("file size could not be found errno : %d\n", errno);

seinfra:
Target : [b87a3e2530831f21fabf123662aac057aa2fe7b6] - Code Rule Check OK.

seinfra:
Target : [b87a3e2530831f21fabf123662aac057aa2fe7b6] - Code Rule Check (C++) OK.

framework/src/media/FileOutputDataSource.cpp
zhouxinhe: Fixed, thanks.

PR #2592, opened 2019-01-11
framework/media: Add media_init

junmin-kim:
media_init initialize media framework

### DIFF
--- a/apps/system/init/init.c
+++ b/apps/system/init/init.c
@@ -39,6 +39,9 @@
 #ifdef CONFIG_EVENTLOOP
 #include <tinyara/eventloop.h>
 #endif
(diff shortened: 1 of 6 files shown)


apps/system/init/init.c
```diff
@@ -133,6 +139,13 @@ int preapp_start(int argc, char *argv[])
 	}
 #endif
 
+#ifdef CONFIG_MEDIA
+	ret = media_init();
```
sunghan-chang: media_init funtion is built with CONFIG_AUDIO. But here, you add it with CONFIG_MEDIA. Please check it

junmin-kim: Thanks for review. I fixed it in #2593

PR #2607, opened 2019-01-18
examples/sensorbd_demo: use sleep()/usleep() instead of up_mdelay()

sunghan-chang:
The up_mdelay() should be only used inside kernel context,
user should not.
Let's use sleep() and usleep() instead of it.


seinfra:
Target : [acd15262df3d6586e598457198c33aeed2015f0b] - Code Rule Check OK.

seinfra:
Target : [acd15262df3d6586e598457198c33aeed2015f0b] - Code Rule Check (C++) OK.

PR #2610, opened 2019-01-21
testcase/kernel: Change signal for waking up a receiver thread in mqe…

jeongchanKim:
…ue tc.

Signo 9 is for terminating the task or pthread, not for waking up.
If pthread_kill(PID, 9) is sent, receiver will be terminated, so returns for pthread_cancel and pthread_join will be changed unexpected.


seinfra:
Target : [03d47b2834744095d59b85e063d4822dcd07609e] - Code Rule Check OK.

seinfra:
Target : [03d47b2834744095d59b85e063d4822dcd07609e] - Code Rule Check (C++) OK.

PR #2631, opened 2019-01-25
kernel/init: allow CONFIG_USER_ENTRYPOINT to be NULL

sunghan-chang:
Originally CONFIG_USER_ENTRYPOINT can not be undefined with
normal usages. But someone modifies defconfig directly (actually
it is wrong usages). For that case, mkconfig.c has a conditional.
If it is undefined, mkconfig.c defines it as "main".
But we can't guarantee it is right so that this commit removes it
to be a compilation error.

Additionally TizenRT allows to execute an application from TASH.
That means someone can use TizenRT without user entry point
which is started automatically at bootup.
When CONFIG_TASH is enabled, this commit allows that
CONFIG_USER_ENTRYPOINT is undefined.
At that case, TizenRT just starts TizenRT's services only.
User application should be launched from TASH.

### DIFF
--- a/os/kernel/init/os_bringup.c
+++ b/os/kernel/init/os_bringup.c
@@ -104,7 +104,7 @@
 #if defined(CONFIG_INIT_ENTRYPOINT)
 /* Initialize by starting a task at an entry point */
 
-#ifndef CONFIG_USER_ENTRYPOINT
+#if !defined(CONFIG_USER_ENTRYPOINT) && !defined(CONFIG_TASH)
 /* Entry point name must have been provided */
 
 #error CONFIG_USER_ENTRYPOINT must be defined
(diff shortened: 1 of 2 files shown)


os/kernel/init/os_bringup.c
```diff
@@ -104,7 +104,7 @@
 #if defined(CONFIG_INIT_ENTRYPOINT)
 /* Initialize by starting a task at an entry point */
 
-#ifndef CONFIG_USER_ENTRYPOINT
+#if !defined(CONFIG_USER_ENTRYPOINT) && !defined(CONFIG_TASH)
```
kishore-sn: Since tash_start() happens in preapp_start(), would it be better to check for CONFIG_SYSTEM_PREAPP_INIT also?

sunghan-chang: How about removing `SYSTEM_PREAPP_INIT` config? Because there is no way to use TASH without preapp init. But if there is nothing to launch some services, we don't need to make a preapp task. Then, we need it... Do you have any suggestion for that?

sunghan-chang: or making a dependancy between preapp config and tash config. When preapp config is enabled, user can enable tash?? Is it bad?

PR #2633, opened 2019-01-28
framework/arastorage: set a maximum value of tuples defined in db_options.h

jeongarmy:
DB_TUPLE_LIMIT is a maximum number of tuples in a relation.
This value should be set according to use case.
A arastorage utc has test cases with more than 800 tuples.
So DB_TUPLE_LIMIT should be bigger than this value.

### DIFF
--- a/framework/src/arastorage/db_options.h
+++ b/framework/src/arastorage/db_options.h
@@ -165,7 +165,7 @@
 
 /* The maximum number of tuples in a relation. */
 #ifndef DB_TUPLE_LIMIT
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [24d2b5a8ee8a0867bd574f52704439a116c53d06] - Code Rule Check (C++) OK.

seinfra:
Target : [24d2b5a8ee8a0867bd574f52704439a116c53d06] - Code Rule Check OK.

PR #2637, opened 2019-01-29
Remove unnecessary compilation options, NXFLATs and files

sunghan-chang:
Makefile: remove NXFLAT compilation options 
configs/esp*/scripts: remove unnecessary Make.defs


sunghan-chang:
@xiarihf I removed `NXFLAT` options and Make.defs from `build/configs/esp*/scripts` folder. If there is any usages of it, please let me know.

seinfra:
Target : [9658df3adf259c7728015283696004e675df775d] - Code Rule Check OK.

seinfra:
Target : [9658df3adf259c7728015283696004e675df775d] - Code Rule Check (C++) OK.

xiarihf:
@sunghan-chang 

Esp32 would not use it. We keep it because we see other part will use it.
Thanks for informing us.

sunghan-chang:
@xiarihf 
> We keep it because we see other part will use it.

Could you let me know more details? What part? for what? And do you mean `NXFLAT`? or `Make.defs` in scripts folder?

sunghan-chang:
@xiarihf Is it used to make a linker script output file?

xiarihf:
> @xiarihf
> 
> > We keep it because we see other part will use it.
> 
> Could you let me know more details? What part? for what? And do you mean `NXFLAT`? or `Make.defs` in scripts folder?

I see you remove the places.
I have checked it, you can remove it in ESP32.

PR #2639, opened 2019-01-30
dbuild.sh: Add ESP32 board image type for updating esp32 binary files.

xiarihf:
Add ESP32 board image type for updating esp32 binary files.

Signed-off-by: xiarihf qiang3.zhang@samsung.com

### DIFF
--- a/os/dbuild.sh
+++ b/os/dbuild.sh
@@ -24,6 +24,7 @@ TOPDIR="${OSDIR}/.."
 BUILDDIR="${TOPDIR}/build"
 BINDIR="${BUILDDIR}/output/bin"
 BINFILE="${BINDIR}/tinyara.bin"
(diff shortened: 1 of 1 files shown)


xiarihf:
@sunghan-chang 
ESP32 tools need us to keep the binary file name as tinyara.elf.bin, so i make such modification. If it's not appropriate, please let me know. Thanks.

sunghan-chang:
@xiarihf As you said, some boards support different type of file to flash. I can't block that.
But I just want to keep same operation code even defining can be done by each board.
i.e.
```
if board == xx
FILES = aa
elif board == yy
FILES = bb

flash()
{
 download FILES
}
```
I don't want to modify many lines to support new files. Just a line is enough.

sunghan-chang:
@xiarihf When I make a PR, I will let you know.

PR #2640, opened 2019-01-30
build/configs: Disable secure storage for st_things.

Hyunjun85:
ST app will try to save some certification files to secure storage.
In this case, SmartThings application can show some error about certification saving.

When the certification size is bigger than the storage storage's max size(8192).
then the stack show some error and fail to start stack.
Currently, feature which is saving secure storage cannot be default option until the policy will be prepared.

For the future, we should decide how to handle big files to save on it.


seinfra:
Target : [19930c267d70992a7d666170a8717a2445224fc3] - Code Rule Check OK.

seinfra:
Target : [19930c267d70992a7d666170a8717a2445224fc3] - Code Rule Check (C++) OK.

sunghan-chang:
@Hyunjun85 Location of modification is `build/configs`, not 'framework/st_things`.

Hyunjun85:
@sunghan-chang Thanks. 
fixed.

seinfra:
Target : [a6dcf2393c495e45568c110e9f0fb8d5fb1ae2c3] - Code Rule Check OK.

JeongJunSik:
I think WIFI Profile information and Cloud Token are very important information.
The data over 8kb is  SVR DB only., don't you?
Then, How about turning off SVR DB feature only?

Hyunjun85:
@JeongJunSik  Thank you. Very good point. 

But as far as I know, There is no guide or policy about this. 
An information will save secure storage and the another information will save smartfs. Like this case, we can give some confuse to user. 

In my thinking, It is good for use only smart file system before decide policy or prepare some logic to care it, now.

JeongJunSik:
@Hyunjun85 I respect your opinion.
When ARTIK_HW_CERT is used, SVR DB Size is reduced to 5kb or less.
So in this case, what about connecting CONFIG_SVR_DB_SECURESTORAGE to CONFIG_ST_THINGS_ARTIK_HW_CERT_KEY dependencies?

Hyunjun85:
@JeongJunSik  Thank you for advice. 

Currently personal account user can be developeing thing using without permission as manufacturer. and hardwere certification key feature is hard to use for common user. 

Moreover, It looks that Secure storage feature doesn't have relation with h/w cert key in directly per useablility perspective.

If I am wrong, could you please let me know any other options?

build/configs/artik053s/st_things/defconfig
```diff
@@ -248,7 +248,7 @@ CONFIG_ARTIK05X_AUTOMOUNT_SSSRW_MOUNTPOINT="/sss"
 CONFIG_ARTIK05X_AUTOMOUNT_ROMFS=y
 CONFIG_ARTIK05X_AUTOMOUNT_ROMFS_DEVNAME="/dev/mtdblock9"
 CONFIG_ARTIK05X_AUTOMOUNT_ROMFS_MOUNTPOINT="/rom"
-CONFIG_SVR_DB_SECURESTORAGE=y
```
davidfather: I agree with you in ARTIK053.
However, unlike ARTIK053, ARTIK053s has build-in secure storage, intending to save information in a safe way.
So, it might be better to have the secure storage config option by default.

Hyunjun85: I agree.
I will modify default config of 053s.
H/W KEY CERT FEATURE is good to get dependency to SECURE STORAGE only for 053s.

seinfra:
Target : [4e120852fe562508e7cb0443df02c555fe028fe9] - Code Rule Check OK.

seinfra:
Target : [4e120852fe562508e7cb0443df02c555fe028fe9] - Code Rule Check (C++) OK.

PR #2647, opened 2019-02-01
dbuild: define BINFILE to say output file of each board and tidy up

sunghan-chang:
1. Use the BINFILE variable to check output file of each board
  Each board uses specific output file to program on a board.
  This commit makes a line change to support new board binary type
  by separating operation codes and a definition which can be
  different by each board.

2. Show the download option
  FIND_BINFILE function checks existing of output file and updates
  STATUS so that we don't need to check it again to show the download
  option. Let's use STATUS for that.

3. Update STATUS after build
  Let's use UPDATE_STATUS function which we already support.

4. Rename STATUS function to UPDATE_STATUS
  That function checks and updates the STATUS variable so that
  UPDATE is better than CHECK as a function name.

### DIFF
--- a/os/dbuild.sh
+++ b/os/dbuild.sh
@@ -23,8 +23,6 @@ CONFIGFILE="${OSDIR}/.config"
 TOPDIR="${OSDIR}/.."
 BUILDDIR="${TOPDIR}/build"
 BINDIR="${BUILDDIR}/output/bin"
-BINFILE="${BINDIR}/tinyara.bin"
-BINFILE_ESP32="${BINDIR}/tinyara.elf.bin"
 CONFIGDIR="${BUILDDIR}/configs"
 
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [7c98793e77566cc8da1ca49f2ecd2529ec55638a] - Code Rule Check (C++) OK.

seinfra:
Target : [7c98793e77566cc8da1ca49f2ecd2529ec55638a] - Code Rule Check OK.

seinfra:
Target : [917e6c7de19407c0878f19036f8c4cd711ab10e5] - Code Rule Check (C++) OK.

seinfra:
Target : [917e6c7de19407c0878f19036f8c4cd711ab10e5] - Code Rule Check OK.

PR #2681, opened 2019-02-20
mm_heap/mm_malloc: remove redundant logic when getting nodelist index.

davidfather:
In mm_size2ndx(), there is the logic that check if the request size is bigger than MM_MAX_CHUNK. So it's not necessary to have the same logic right before entering mm_size2ndx().


seinfra:
Target : [d6777406f2fb518993d4f88e09f5e27e8f56410c] - Code Rule Check OK.

seinfra:
Target : [d6777406f2fb518993d4f88e09f5e27e8f56410c] - Code Rule Check (C++) OK.

PR #2686, opened 2019-02-20
Fix the network testcase 'tc_net_ioctl_fionread_n'

HONGCHAEHEE:
The ioctl command FIONREAD is defined twice both in tinyara/fs/ioctl.h and net/lwip/sockets.h,
and lwip_ioctl fuction uses the command defined in tinyara/fs/ioctl.h.
We disable FIONREAD in net/lwip/sockets.h to avoid conflict.

The testcase 'tc_net_ioctl_fionread_n' was passed because the parameter FIONREAD which is defined in net/lwip/sockets.h didn't match that used in lwip_ioctl function.
So we change this testcase to a positive case because it works well with FIONREAD command.

### DIFF
--- a/os/include/net/lwip/sockets.h
+++ b/os/include/net/lwip/sockets.h
@@ -412,9 +412,6 @@ typedef struct ip_mreq {
 #define _IOW(x, y, t)     (IOC_IN | (((long)sizeof(t) & IOCPARM_MASK) << 16) | ((x) << 8) | (y))
 #endif							/* !defined(FIONREAD) || !defined(FIONBIO) */
 
-#ifndef FIONREAD
-#define FIONREAD    _IOR('f', 127, unsigned long)	/* get # bytes to read */
-#endif
 #ifndef FIONBIO
 #define FIONBIO     _IOW('f', 126, unsigned long)	/* set/clear non-blocking i/o */
 #endif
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [242eaaa36d3828e8c049575f64f8b8114076dc93] - Code Rule Check (C++) OK.

seinfra:
Target : [242eaaa36d3828e8c049575f64f8b8114076dc93] - Code Rule Check OK.

os/include/net/lwip/sockets.h
```diff
@@ -412,9 +412,13 @@ typedef struct ip_mreq {
 #define _IOW(x, y, t)     (IOC_IN | (((long)sizeof(t) & IOCPARM_MASK) << 16) | ((x) << 8) | (y))
 #endif							/* !defined(FIONREAD) || !defined(FIONBIO) */
 
+/* FIONREAD is already defined in tinyara/fs/ioctl.h.
+   To avoid conflict, we disable this command. */
+#if 0
```
drashti304: Can't we just remove the lines instead of adding #if 0 ?

HONGCHAEHEE: I'll remove the lines because it's not used.

seinfra:
Target : [c8d406b00cc81dd633fc4fa4ac7978003a03ea96] - Code Rule Check (C++) OK.

seinfra:
Target : [c8d406b00cc81dd633fc4fa4ac7978003a03ea96] - Code Rule Check OK.

PR #2699, opened 2019-02-22
kernel/signal : Change sig_dispatch about delivering the signal to th…

sunghan-chang:
…e specified pid

sig_dispatch runs like below
1. if pid task/pthread is alive and unblock the signo, it can get the signal first even HAVE_GROUP_MEMBERS is enabled.
2. if pid task/pthread is dead or block the signo, then one of the members in group will receive the signal when HAVE_GROUP_MEMBERS is enabled.


seinfra:
Target : [e4b1ed6b6399066ac715aaaeb6d679c2253a483c] - Code Rule Check OK.

seinfra:
Target : [e4b1ed6b6399066ac715aaaeb6d679c2253a483c] - Code Rule Check (C++) OK.

PR #2720, opened 2019-02-26
apps/examples: Delete examples mtdpart

KimDongEon:
This is an example of an api that the user does not use.


seinfra:
Target : [6f450d268d9afe384e51125aaf42901e15518e1f] - Code Rule Check (C++) OK.

seinfra:
Target : [6f450d268d9afe384e51125aaf42901e15518e1f] - Code Rule Check OK.

KimDongEon:
@sunghan-chang 
Is it not necessary to provide this in protected build mode because it makes a kernel context call?
If so, how do you think about registering the command to the fs driver and calling it using ioctl?

sunghan-chang:
@KimDongEon You need to determine whether there is an usecase or not. If it is a part of kernel task, and it just shows how to use it for kernel, then you don't need to provide it with VFS.

KimDongEon:
If you rely on !BUILD_PROTECTED and only show the way you use it in the kernel, there is no problem with your current code. Therefore, I will close this PR.

KimDongEon:
The filesystem tc was made dependent on the "!BUILD_PROTECTED" on another PR.

PR #2734, opened 2019-02-27
board/artik05x: remove stale elf file after protected build

manoharahk:
board/artik05x/userspace/tinyara_user.elf is copied to
build/output/bin/ and no longer required.

Verification:
run git status after build

### DIFF
--- a/os/board/artik05x/userspace/Makefile
+++ b/os/board/artik05x/userspace/Makefile
@@ -94,6 +94,7 @@ OBJS  = $(COBJS)
 # Targets:
 
 all: $(USER_BINPATH)$(DELIM)tinyara_user.elf $(USER_BINPATH)$(DELIM)User.map
+	$(call DELFILE, tinyara_user.elf)
 .PHONY: tinyara_user.elf depend clean distclean
 
 $(COBJS): %$(OBJEXT): %.c


os/board/artik05x/userspace/Makefile
```diff
@@ -94,6 +94,7 @@ OBJS  = $(COBJS)
 # Targets:
 
 all: $(USER_BINPATH)$(DELIM)tinyara_user.elf $(USER_BINPATH)$(DELIM)User.map
+	$(Q) rm -f tinyara_user.elf
```
sunghan-chang: Let's use `DELFILE` instead of `rm -f`. Actually both are same but I want to keep same usages for Makefile.

seinfra:
Target : [3fcf350aef95ba9e5d89046285b8cfcb3fc1daf4] - Code Rule Check OK.

os/board/artik05x/userspace/Makefile
manoharahk: OK. I will update

seinfra:
Target : [3fcf350aef95ba9e5d89046285b8cfcb3fc1daf4] -

os/board/artik05x/userspace/Makefile
sunghan-chang: But I don't know this is necessary or not. After getting approval from @kishore-sn , I will merge this.

seinfra:
Target : [9f553a6977c1f0e2569ed3279dffe6c491670394] - Code Rule Check (C++) OK.

seinfra:
Target : [9f553a6977c1f0e2569ed3279dffe6c491670394] - Code Rule Check OK.

os/board/artik05x/userspace/Makefile
manoharahk: OK

os/board/artik05x/userspace/Makefile
```diff
@@ -94,6 +94,7 @@ OBJS  = $(COBJS)
 # Targets:
 
 all: $(USER_BINPATH)$(DELIM)tinyara_user.elf $(USER_BINPATH)$(DELIM)User.map
+	$(call DELFILE, tinyara_user.elf)
```
sunghan-chang: 1. `$(USER_BINPATH)` is needed.
2. That file will be deleted when we call `CLEAN`. Is this necessary to delete here? Then, Why do you need to support deleting it in CLEAN?

manoharahk: After make, if we do git status , it shows `board/artik05x/userspace/tinyara_user.elf`
Its not good to show these unwanted build artifacts. 
Either hide these by adding it to `.gitignore` OR delete. 
Since this elf already moved to `build/output/bin/`, it is no more useful, hence deleted the file.

Yes, if it is deleted after the build, no need to do the same in clean. I will remove it from clean target

kishore-sn:
@sunghan-chang, As  @manoharahk mentioned, its better to remove the elf file after build since it shows up in "git status" until we perform clean.

sunghan-chang:
@manoharahk @kishore-sn Then, why only user elf? Why not include it into gitignore?

kishore-sn:
@sunghan-chang 
If the file is useful for any future operations then we can keep it and add the file to gitignore. 
But, in this case, the binary has already been copied to its final destination. This elf file has no other usage. Hence, its better to delete it.

PR #2744, opened 2019-03-06
framework/st_things: Add error handling logic for data manager.

Hyunjun85:
There can be a case which device has not enough memory to allocate.
And things_malloc can return NULL.


seinfra:
Target : [feb7e5138d1a2259171369cf0cd567791262af78] - Code Rule Check (C++) OK.

seinfra:
Target : [feb7e5138d1a2259171369cf0cd567791262af78] - Code Rule Check OK.

PR #2748, opened 2019-03-08
apps/examples/testcase: Add dependency configuration filesystem tc

KimDongEon:
I made filesystem TC dependent on the !BUILD_PROTECTED setting.

reason1 : The file system TC has an API that calls the kernel context. Therefore, it is disabled in the BUILD_PROTECTED state.

reason2 : Because the file system TC is created for testing purposes, not for actual code, disabling BUILD_PROTECTED does not cause any problems


seinfra:
Target : [9e51fc39068584a475792964cf703e84ccab895b] - Code Rule Check (C++) OK.

seinfra:
Target : [9e51fc39068584a475792964cf703e84ccab895b] - Code Rule Check OK.

PR #2754, opened 2019-03-08
board/common: change a location of sub-title in Kconfig

sunghan-chang:
A sub-title of flash partition was in board/Kconfig and it
prints sub-title before sourcing board/common/Kconfig.
Because of this, when it is not available, only sub-title
is come up without any config as shown below:

(4096) Page size of external flash memory (in bytes)
    *** Board-Partition Options ***

To hidden it when it is not available, move it inside of
board/common/Kconfig and cover it using conditionals.

### DIFF
--- a/os/board/Kconfig
+++ b/os/board/Kconfig
@@ -268,7 +268,6 @@ endif
 if ARCH_BOARD_IMXRT1050_EVK
 source board/imxrt1050-evk/Kconfig
 endif
-comment "Board-Partition Options"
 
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [c8cce64678744fa10b4962e448842923d1f033d2] - Code Rule Check OK.

seinfra:
Target : [c8cce64678744fa10b4962e448842923d1f033d2] - Code Rule Check (C++) OK.

PR #2757, opened 2019-03-11
Add bootprep.sh script.

btheosam:
This script generates a c file with partition table so that
it can be compiled and stored in flash to aid bootloader in
image upgrade.


btheosam:
Accidently generated this PR (incomplete)

PR #2761, opened 2019-03-11
os/binfmt : fix coding rules

manoharahk:

### DIFF
--- a/os/include/elf32.h
+++ b/os/include/elf32.h
@@ -74,41 +74,41 @@
 
 /* Values for Elf32_Ehdr::e_type */
 
-#define ET_NONE           0       /* No file type */
-#define ET_REL            1       /* Relocatable file */
-#define ET_EXEC           2       /* Executable file */
-#define ET_DYN            3       /* Shared object file */
-#define ET_CORE           4       /* Core file */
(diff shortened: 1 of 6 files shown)


seinfra:
Target : [80bdd7f58b695c6d38a0ffe7278fb684f5f01cfd] - Code Rule Check (C++) OK.

seinfra:
Target : [80bdd7f58b695c6d38a0ffe7278fb684f5f01cfd] - Code Rule Check OK.

os/include/elf32.h
```diff
@@ -74,41 +74,41 @@
 
 /* Values for Elf32_Ehdr::e_type */
 
-#define ET_NONE           0       /* No file type */
```
sunghan-chang: Before is not a violation. After can be. Because there are spaces before tab. Let's revert this change.

manoharahk: This change was done by cformarter.sh .. 
Is it wrong?

sunghan-chang: As far as I know, if there is a space before a tab. CRC recognizes it a violation. But I can be incorrect. And I actually prefer spaces than tabs for alignment.

PR #2766, opened 2019-03-12
external/mbedtls: Remove redundant heap memory check logic

SeonghoByeon:
- Remove a comparison logic of mallinfo fordblks and MBEDTLS_MAXIMUM_HANDSHAKE_MEMORY_USAGE
1) It increases unnecessary time overhead
2) fordblks shows the whole (discontinuous) heap size
3) Memory allocation failure will be returned if it fails

### DIFF
--- a/external/mbedtls/ssl_tls.c
+++ b/external/mbedtls/ssl_tls.c
@@ -7019,16 +7019,6 @@ int mbedtls_ssl_handshake( mbedtls_ssl_context *ssl )
 {
     int ret = 0;
 
-#if defined(MBED_TIZENRT)
-    struct mallinfo data = mallinfo();
-
-    if( data.fordblks < MBEDTLS_MAXIMUM_HANDSHAKE_MEMORY_USAGE )
-    {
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [6d0e4f5084ea2e218a37ceaca928d0687616b65d] - Code Rule Check (C++) OK.

seinfra:
Target : [6d0e4f5084ea2e218a37ceaca928d0687616b65d] - Code Rule Check OK.

sunghan-chang:
@SeonghoByeon In memory usages, it's unnecessary because allocation already checks and returns it. But if operation before allocation is too huge (longer than time of mallinfo) so that it is good to check it first, we can keep memory checking here. I think alloc and free is enough instead of mallinfo. mallinfo searchs whole memory so that it spends much time.

SeonghoByeon:
@sunghan-chang Thanks for the comment. I agree with you. Memory check logic here might reduce some overhead caused by the lack of memory during tls session. However, 
1. This memory check logic is too naive (to search the whole memory)
2. The remaining memory identified in the start phase is variable
3. It is not possible to set MBEDTLS_MAXIMUM_HANDSHAKE_MEMORY_USAGE to the correct value. It depends on the application, server, and etc

Thanks for the feedback, again.

PR #2771, opened 2019-03-12
os/mm: Fix mm_get_heapindex logic

zhouxinhe:
- Check the given memory address in all regions of user heaps.
- Refer the logic in `bool kmm_heapmember(FAR void *mem)`

@xixidodo @sunghan-chang

### DIFF
--- a/os/mm/mm_heap/mm_getheap.c
+++ b/os/mm/mm_heap/mm_getheap.c
@@ -81,9 +81,19 @@ int mm_get_heapindex(void *mem)
 		return 0;
 	}
 	for (heap_idx = 0; heap_idx < CONFIG_MM_NHEAPS; heap_idx++) {
-		if (mem < (void *)(g_mmheap[heap_idx].mm_heapstart + g_mmheap[heap_idx].mm_heapsize) && mem >= (void *)g_mmheap[heap_idx].mm_heapstart) {
-			return heap_idx;
+		int region = 0;
+#if CONFIG_MM_REGIONS > 1
+		for (; region < g_mmheap[heap_idx].mm_nregions; region++)
+#endif
+		{
+			/* A valid address from the user heap for this region would have to lie
+			 * between the region's two guard nodes.
+			 */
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [447944f1ed595f7c5119757d28562f8ea9ce15e2] - Code Rule Check OK.

seinfra:
Target : [447944f1ed595f7c5119757d28562f8ea9ce15e2] - Code Rule Check (C++) OK.

os/mm/mm_heap/mm_getheap.c
```diff
@@ -81,9 +81,19 @@ int mm_get_heapindex(void *mem)
...
-			return heap_idx;
+		int region = 0;
+#if CONFIG_MM_REGIONS > 1
+		for (; region < g_mmheap[heap_idx].mm_nregions; region++)
+#endif
+		{
+			/* A valid address from the user heap for this region would have to lie
+			 * between the region's two guard nodes.
+			 */
+			if ((mem > (void *)g_mmheap[heap_idx].mm_heapstart[region]) && (mem < (void *)g_mmheap[heap_idx].mm_heapend[region])) {
```
drashti304: Isn't it should be >= and <= ?

zhouxinhe: Thanks for your review!
```text
Region *:
A(addr)      B            C      D            E      F     ...      Y            Z
+------------+------------+------+------------+------+--------------+------------+
| guard node | alloc node + mem1 | alloc node + mem2 |     ...      | guard node |
+--------------------------------------------------------------------------------+
```
My understanding is, the input `mem` should be an pointer to the allocated memory, 
like `C->mem1` and `E->mem2` above. So that, it must be between the address  `B` and `Y`.
I think `> and <` can work fine. Of course, `>= and <=` is also correct.

jeongchanKim: I think `> and <` is enough.

PR #2799, opened 2019-03-18
os/fs/procfs: add exceptional protection

SRCN-LiQin:
add exceptional protection in proc_entry_stat,
when reading the heapinfo, pointer from mm_get_heap
maybe NULL, like the idle task, whose stack is not in the heap.
@jeongchanKim @xixidodo

### DIFF
--- a/os/fs/procfs/fs_procfsproc.c
+++ b/os/fs/procfs/fs_procfsproc.c
@@ -386,6 +386,9 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	hash_pid = PIDHASH(tcb->pid);
 	heap = mm_get_heap(tcb->stack_alloc_ptr);
+	if (heap == NULL) {
+		heap = mm_get_heap_with_index(0);
+	}
 	if (heap->alloc_list[hash_pid].pid == tcb->pid) {
 		curr_heap = heap->alloc_list[hash_pid].curr_alloc_size;
 		peak_heap = heap->alloc_list[hash_pid].peak_alloc_size;


seinfra:
Target : [92e331a7bf923da9b2d8cde12c0f1210d2757a7b] - Code Rule Check OK.

seinfra:
Target : [92e331a7bf923da9b2d8cde12c0f1210d2757a7b] - Code Rule Check (C++) OK.

sunghan-chang:
@SRCN-LiQin Without this PR, do you meet a crash there? Even it shows it is not a heap region, why should we run it with index 0? why not stop? We will check it more and then let you know.
@jeongchanKim Please let us know a test result.

SRCN-LiQin:
> @SRCN-LiQin Without this PR, do you meet a crash there? Even it shows it is not a heap region, why should we run it with index 0? why not stop? We will check it more and then let you know.
> @jeongchanKim Please let us know a test result.

yes, there is crash issue without this PR, because idle task return NULL for mm_get_heap because it's stack is not in the heap. this solution handled all null pointer case under the consideration that: if we can't find heap with the sp ptr, then we use the basic heap in mm which have index 0 to check heap alloc info.

possible alternative solution is that: we handle idle task as known case and get heap with index 0 directly. and other NULL case we can treat as abnormal case and do not get heapinfo for it.

plz kindly share your opinion

jeongchanKim:
@SRCN-LiQin, Hello. I checked that mm_get_heap returns NULL when Idle task's stack ptr is passed. So I agree that it can be crashed. But I think returning 0th heap for NULL case is not good, because returning NULL can happen with another case. Then we cannot distinguish.

 And there is only one case that idle's stack ptr is passed to mm_get_heap(). Only in fs_procfsproc.c. 

So I think we'd better handle this idle task's stack case separately in fs_procfsproc.c
And add error checking logic for returning NULL from mm_get_heap().

I will update these with another PR. Please close this PR. Thanks.

sunghan-chang:
Please find #2819 .

SRCN-LiQin:
OK I find your PR include abnormal protection for other mm_get_heap function call, I will close this PR, thanks.

PR #2805, opened 2019-03-19
framework/st_things: fix bug which it can generate wrong hash value.

Hyunjun85:
We need some hash prefix based device uuid.
and it can have NULL('\0') character middle of string.

In previous, It has been using strlen to get length of hash
and it can return wrong length why the hash can get NULL in middle of string.

Now, It will use fixed length(32) for device UUID.

### DIFF
--- a/framework/src/st_things/things_stack/framework/things_sss_manager.c
+++ b/framework/src/st_things/things_stack/framework/things_sss_manager.c
@@ -41,6 +41,8 @@
 
 #define TAG "THINGS_SSS"
 
+#define HASH_LENGTH 32
+
 static OicSecKey_t cacert;
 static OicSecKey_t subca;
 static OicSecKey_t devicecert;
@@ -276,8 +278,8 @@ bool things_encrypt_artik_uuid(unsigned char *output)
 	 */
 
 	// 1. sha256(CertUUID)
-	unsigned char uuid_hash[32 + 1] = { 0, };
-	memset(uuid_hash, 0, sizeof(uuid_hash));
+	unsigned char uuid_hash[HASH_LENGTH + 1] = { 0, };
+
 	mbedtls_sha256((const unsigned char *)uuid, uuid_len, uuid_hash, 0);
 
 	// 2. base64(uuid_hash)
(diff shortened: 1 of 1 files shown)


framework/src/st_things/things_stack/framework/things_sss_manager.c
```diff
@@ -41,6 +41,8 @@
 
 #define TAG "THINGS_SSS"
 
+#define THINGS_DEVICE_UUID_LENGTH 32
```
sunghan-chang: Are UUID and UUID_HASH same?

seinfra:
Target : [aa5efa43e80ab88e5eaec174fdd64e68e28df4dd] - Code Rule Check (C++) OK.

framework/src/st_things/things_stack/framework/things_sss_manager.c
```diff
@@ -276,7 +278,7 @@ bool things_encrypt_artik_uuid(unsigned char *output)
 	 */
 
 	// 1. sha256(CertUUID)
-	unsigned char uuid_hash[32 + 1] = { 0, };
+	unsigned char uuid_hash[THINGS_DEVICE_UUID_LENGTH + 1] = { 0, };
 	memset(uuid_hash, 0, sizeof(uuid_hash));
```
sijupunnnoose: Should we use memset again? Initialize to zero in previous step?

seinfra:
Target : [aa5efa43e80ab88e5eaec174fdd64e68e28df4dd] -

seinfra:
Target : [03bbf95ab39db8a44cca6b94e76d6d764d644571] -

seinfra:
Target : [6b660aba18ac3b919ac6fb295b36d189b01535bf] - Code Rule Check OK.

seinfra:
Target : [6b660aba18ac3b919ac6fb295b36d189b01535bf] - Code Rule Check (C++) OK.

Hyunjun85:
@SeonghoByeon  Yes it can be. 
But currently, regarding easy-setup guide(by mail), we have to  use only sha256 ans base64 encoder only.

PR #2828, opened 2019-03-22
testcase/Kconfig : Add if condition for avoiding to enabling with wro…

jeongchanKim:
…ng conditions

If a tc depends on some configs, it should not be selected by enabling default tc config.


seinfra:
Target : [37e7c51802ae476d888e89291709765d81a771ac] - Code Rule Check (C++) OK.

seinfra:
Target : [37e7c51802ae476d888e89291709765d81a771ac] - Code Rule Check OK.

PR #2869, opened 2019-04-02
mm_heap/mm_addfreechunk : sorting the free nodelist

davidfather:
in ascending order of physical address when the size is the same.
This can prevent higher address memory from being allocated
when lower address memory is available in the free nodelist.
Therefore, memory fragmentation can be mitigated.


seinfra:
Target : [19d290d47597db3e4ed15fa82524bdbc861a17a4] - Code Rule Check (C++) OK.

seinfra:
Target : [19d290d47597db3e4ed15fa82524bdbc861a17a4] - Code Rule Check OK.

PR #2874, opened 2019-04-03
imxrt : Fix kernel debug logs

kishore-sn:
Kernel debug logs are not coming since some implementation for lldbg and dbg is missing. 
In this PR, I have done following:
- Fix implementation of lldbg
- Redirect dbg to lldbg (This is temporary and will be reverted when dbg implementation in imxrt is fixed)
- Enabled CONFIG_DEBUG and CONFIG_DEBUG_ERROR

### DIFF
--- a/os/arch/arm/src/imxrt/imxrt_lowputc.h
+++ b/os/arch/arm/src/imxrt/imxrt_lowputc.h
@@ -68,6 +68,7 @@
 #include "chip.h"
 #include "imxrt_config.h"
 
+
 /****************************************************************************
  * Public Types
  ****************************************************************************/
@@ -122,7 +123,7 @@ int imxrt_lpuart_configure(uint32_t base, FAR const struct uart_config_s *config
  *
  ************************************************************************************/
 
-#if defined(HAVE_LPUART_DEVICE) && defined(CONFIG_DEBUG_FEATURES)
+#if defined(HAVE_LPUART_DEVICE) && defined(CONFIG_DEBUG)
 void imxrt_lowputc(int ch);
 #else
 #define imxrt_lowputc(ch)
--- a/os/arch/arm/src/imxrt/imxrt_serial.c
+++ b/os/arch/arm/src/imxrt/imxrt_serial.c
@@ -105,34 +105,42 @@
 #define CONSOLE_DEV         g_uart1port	/* LPUART1 is console */
 #define TTYS0_DEV           g_uart1port	/* LPUART1 is ttyS0 */
 #define UART1_ASSIGNED      1
+#define HAVE_SERIAL_CONSOLE 1
 #elif defined(CONFIG_LPUART2_SERIAL_CONSOLE)
 #define CONSOLE_DEV         g_uart2port	/* LPUART2 is console */
 #define TTYS0_DEV           g_uart2port	/* LPUART2 is ttyS0 */
 #define UART2_ASSIGNED      1
+#define HAVE_SERIAL_CONSOLE 1
 #elif defined(CONFIG_LPUART3_SERIAL_CONSOLE)
 #define CONSOLE_DEV         g_uart3port	/* LPUART3 is console */
 #define TTYS0_DEV           g_uart3port	/* LPUART3 is ttyS0 */
 #define UART3_ASSIGNED      1
+#define HAVE_SERIAL_CONSOLE 1
 #elif defined(CONFIG_LPUART4_SERIAL_CONSOLE)
 #define CONSOLE_DEV         g_uart4port	/* LPUART4 is console */
 #define TTYS0_DEV           g_uart4port	/* LPUART4 is ttyS0 */
 #define UART4_ASSIGNED      1
+#define HAVE_SERIAL_CONSOLE 1
 #elif defined(CONFIG_LPUART5_SERIAL_CONSOLE)
 #define CONSOLE_DEV         g_uart5port	/* LPUART5 is console */
 #define TTYS0_DEV           g_uart5port /* LPUART5 is ttyS0 */
 #define UART5_ASSIGNED      1
+#define HAVE_SERIAL_CONSOLE 1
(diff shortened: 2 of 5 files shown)


seinfra:
Target : [32d40db2d3bf765351eeeb6af873b60bd944d54f] - Code Rule Check (C++) OK.

seinfra:
Target : [32d40db2d3bf765351eeeb6af873b60bd944d54f] - Code Rule Check OK.

os/include/debug.h
```diff
@@ -153,8 +153,14 @@ Once LOGM is approved, each module should have its own index
  * @details @b #include <debug.h>
  * @since TizenRT v1.0
  */
```
sangam-swami: Can you please make separate commit for this temporary change, we can revert after the issue fix

kishore-sn: Done.

seinfra:
Target : [a09895e7a85af519aa3d06c2028dc04d71bd5ba4] - Code Rule Check OK.

seinfra:
Target : [a09895e7a85af519aa3d06c2028dc04d71bd5ba4] - Code Rule Check (C++) OK.

os/arch/arm/src/imxrt/imxrt_lowputc.h
```diff
@@ -68,6 +68,10 @@
 #include "chip.h"
 #include "imxrt_config.h"
 
+#ifdef CONFIG_DEBUG
+#define CONFIG_DEBUG_FEATURES
```
sunghan-chang: Could you let me know the purpose of new definition?
I think we need to use `CONFIG_DEBUG`.

os/include/debug.h
```diff
@@ -190,8 +196,13 @@ Once LOGM is approved, each module should have its own index
...
+#ifdef CONFIG_ARCH_CHIP_IMXRT
```
sunghan-chang: I don't want to add board-specific temporary feature in common standard header.
How about redefine it in arch header?
If there is a common header which almost imxrt source codes includes, we can add below
```
#undef llwdbg
#define llwdbg wdbg
```

sunghan-chang:
@kishore-sn
Your last change, `imxrt1050-evk/nxp_demo: Enable kernel debug logs`, is not a kernel-specific change.
Logs of all modules are enabled.

sunghan-chang:
@kishore-sn Could you update this?

kishore-sn:
> t1050-evk/nxp_

I changed the commit message.

os/arch/arm/src/imxrt/imxrt_lowputc.h
kishore-sn: True. I will change it.

os/include/debug.h
kishore-sn: I understand your concern. But, it is not possible to redefine in arch specific header. In that case our change will reflect only in arch specific files. As you know, dbg is not working properly and so, where ever dbg is used in entire TizenRT code, I want it to be replaced with lldbg. 

We will try to find a fix for the dbg issue. And revert this change as soon as possible.

seinfra:
Target : [ea68d0b77aabaa4473a2d83ccd31180c2463b21e] - Code Rule Check (C++) OK.

seinfra:
Target : [ea68d0b77aabaa4473a2d83ccd31180c2463b21e] - Code Rule Check OK.

os/arch/arm/src/imxrt/imxrt_serial.c
```diff
@@ -1520,12 +1528,11 @@ int up_putc(int ch)
...
 	if (ch == '\n') {
 		/* Add CR */
-
-		up_lowputc('\r');
+		imxrt_lowputc('\r');
 	}
 
-	up_lowputc(ch);
-	imxrt_restoreuartint(priv, intset);
+	imxrt_lowputc(ch);
```
sunghan-chang: Here is not in CONFIG_DEBUG conditional. But at above, definition of this function is in it. When debug is disabled, is it ok?

os/include/debug.h
sunghan-chang: ok. Could you add a comment to say this is a temporary fix?

kishore-sn: Done.

os/arch/arm/src/imxrt/imxrt_serial.c
kishore-sn: We have this code in imxrt_lowputc.h. So it will be ok if CONFIG_DEBUG is disabled.

`#if defined(HAVE_LPUART_DEVICE) && defined(CONFIG_DEBUG)
'void imxrt_lowputc(int ch);
'#else
'#define imxrt_lowputc(ch)
'#endif
`

kishore-sn: We have this code in imxrt_lowputc.h. So it will be ok if CONFIG_DEBUG is disabled.

#if defined(HAVE_LPUART_DEVICE) && defined(CONFIG_DEBUG)
void imxrt_lowputc(int ch);
#else
#define imxrt_lowputc(ch)
#endif

seinfra:
Target : [0b4bf76081dbd48076c2f7ef87ca05ec0f061dd9] - Code Rule Check OK.

seinfra:
Target : [0b4bf76081dbd48076c2f7ef87ca05ec0f061dd9] - Code Rule Check (C++) OK.

os/arch/arm/src/imxrt/imxrt_serial.c
sunghan-chang: In compilation point of view, it's ok. But is there any reason to block it with debug disabled?

kishore-sn: I dont think it needs to depend on CONFIG_DEBUG. I will remove the condition for CONFIG_DEBUG and keep the check for HAVE_LPUART_DEVICE.

PR #2915, opened 2019-04-10
Fix esp32 WiFi's Make.defs

ChinaTizenRT:
If archiving objects without option "P", one object will be
replaced with other one with the same file name, then causing
some unexpected issue, such as build error.

Below is our experience.
when buiding  library: libexternal.a, there are 2 objects with same name,
1) TizenRT/external/mbedtls/sha1.o
2) TizenRT/external/esp_idf_port/wpa_supplicant/src/crypto/sha1.o

then we found in TizenRT/build/output/libraries/libexternal.a,
object 2) is replaced with 1).

And if we add "P" for AR command, this issue fixed.
xtensa-esp32-elf-ar: 
_r[ab][f][u] - replace existing or insert new file(s) into the archive
[P] - use full path names when matching_


seinfra:
Target : [cfd5d98401ae282e5669927db2951e0c882ec7c8] - Code Rule Check OK.

seinfra:
Target : [cfd5d98401ae282e5669927db2951e0c882ec7c8] - Code Rule Check (C++) OK.

sunghan-chang:
How about title as shown below:
```
configs/esp32 : Add 'P' GNU AR option in WiFi's Make.defs
```
I can't know what you do to fix Make.defs from title so that I should go into commit change.

ChinaTizenRT:
changed with the concrete title.

seinfra:
Target : [9e35c5aae7bf8083e72b4bbd8193c39dc70d2724] - Code Rule Check (C++) OK.

seinfra:
Target : [9e35c5aae7bf8083e72b4bbd8193c39dc70d2724] - Code Rule Check OK.

PR #2955, opened 2019-04-16
arch/xtensa: fix build error when CONFIG_RR_INTERVAL is not defined

glb1013:
If CONFIG_RR_INTERVAL is not defined, there are build errors.

### DIFF
--- a/os/arch/xtensa/src/xtensa/xtensa_blocktask.c
+++ b/os/arch/xtensa/src/xtensa/xtensa_blocktask.c
@@ -140,7 +140,9 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 			 */
 
 			rtcb = this_task();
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [3f535939c6b1c0b37e2bb11b4a647dfb982a56fc] - Code Rule Check (C++) OK.

seinfra:
Target : [3f535939c6b1c0b37e2bb11b4a647dfb982a56fc] - Code Rule Check OK.

sunghan-chang:
@glb1013 Because this does not affect any operation change, I will merge as of now.
But later, please write what you change in a title. NOT what result is.
For example, this case can be written like below
```
arch/xtensa: cover the timeslice element of tcb structure with CONFIG_RR_INTERVAL

This resolves build break, because blah blash..
```

PR #2967, opened 2019-04-17
arch/arm:  Enable MPU for imxrt1050-evk

sangam-swami:
- Add MPU for imxrt in protected build
- Enable MPU setting


seinfra:
Target : [28d886f59ddd0f973c30536b0025883b50cc5f80] - Code Rule Check OK.

seinfra:
Target : [28d886f59ddd0f973c30536b0025883b50cc5f80] - Code Rule Check (C++) OK.

os/arch/arm/src/imxrt/imxrt_mpuinit.c
```diff
@@ -118,18 +134,11 @@ void imxrt_mpu_initialize(void)
...
-	mpu_user_flash(USERSPACE->us_textstart, USERSPACE->us_textend - USERSPACE->us_textstart);
-
-	datastart = MIN(USERSPACE->us_datastart, USERSPACE->us_bssstart);
-	dataend = MAX(USERSPACE->us_dataend, USERSPACE->us_bssend);
-
-	DEBUGASSERT(dataend >= datastart);
+	int i;
 
-	mpu_user_intsram(datastart, dataend - datastart);
+	for (i = 0; i < (sizeof(regions_info) / sizeof(struct mpu_region_info)); i++) {
```
NilabhShant: How about: 
int i_max;
i_max = (sizeof(regions_info) / sizeof(struct mpu_region_info));

then use i_max in for loop. This will be increase overhead of a variable but more optimize as it  doesn't call sizeof function over an over again.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -146,6 +148,7 @@
 #define MPU_RASR_S              (1 << 18)	/* Bit 18: Shareable */
 #define MPU_RASR_TEX_SHIFT    (19)	/* Bits 19-21: TEX Address Permisson */
 #define MPU_RASR_TEX_MASK     (7 << MPU_RASR_TEX_SHIFT)
+#define MPU_RASR_TEX(n)		((uint32_t)(n) << MPU_RASR_TEX_SHIFT)
```
NilabhShant: Please aline macros.

os/arch/arm/include/armv7-m/irq.h
```diff
@@ -91,6 +91,23 @@
...
+#define MPU_REG3       (3)
+#define MPU_REG4       (4)
+#define MPU_REG5       (5)
+#define MPU_REG6       (6)
+#define MPU_REG7       (7)
+#if (CONFIG_ARMV7M_MPU_NREGIONS == 12)
+#define MPU_REG8       (8)
+#define MPU_REG9       (9)
+#define MPU_REG10      (10)
+#define MPU_REG11      (11)
```
yashwanth686007: What if the number of mpu regions is greater than 12 (or equal to 16 as mentioned in defconfig)?

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: there is a single tab for the change made

os/arch/arm/src/imxrt/imxrt_mpuinit.c
sangam-swami: this is resolved at compile time, I don't think it will have any overhead

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -319,13 +312,15 @@ static inline void mpu_control(bool enable, bool hfnmiena, bool privdefena)
...
 #if defined(CONFIG_ARMV7M_HAVE_ICACHE) || defined(CONFIG_ARMV7M_DCACHE)
-static inline void mpu_priv_stronglyordered(uintptr_t base, size_t size)
+static inline void mpu_priv_stronglyordered(uint32_t region, uintptr_t base, size_t size)
 {
-	unsigned int region = mpu_allocregion();
 	uint32_t regval;
 	uint8_t l2size;
 	uint8_t subregions;
 
+#if defined(CONFIG_ARMV7M_MPU_NREGIONS)
```
kishore-sn: In my opinion we dont need to check this. NREGION is always set to some value and we always need to verify that reion number that we pass is less than REGION. Please change in all the API's.

os/arch/arm/src/imxrt/imxrt_mpuinit.c
```diff
@@ -80,6 +81,26 @@
 #define MIN(a, b) a < b ? a : b
 #endif
 
+#if defined(CONFIG_BUILD_PROTECTED)
+const struct mpu_region_info regions_info[] = {
+	{
+		&mpu_userintsram_wb, 0x0, 0x80000000, MPU_REG_ENTIRE_MAP,
```
kishore-sn: Here the size of the default regions is not correct. Please refer to the memory map in os/arch/arm/src/imxrt/chip/imxrt105x_memorymap.h and set the size.

os/arch/arm/src/imxrt/imxrt_mpuinit.c
```diff
@@ -150,7 +159,7 @@ void imxrt_mpu_initialize(void)
 #ifdef CONFIG_BUILD_PROTECTED
 void imxrt_mpu_uheap(uintptr_t start, size_t size)
```
kishore-sn: This API will no longer be required. Please delete the entire API.

os/arch/arm/include/armv7-m/irq.h
```diff
@@ -91,6 +91,23 @@
 #define CONFIG_SYS_NNEST 2
 #endif
 
+/* MPU Region Definitions */
```
kishore-sn: These changes are not required. We dont use them. Please remove.

os/arch/arm/src/imxrt/mpu-reg.h
```diff
@@ -0,0 +1,58 @@
...
+#ifndef __ASSEMBLY__
+typedef void (*fptr)(uint8_t region, uintptr_t base, size_t size);
+struct mpu_region_info {
+	fptr	call;
+	unsigned int	base;
+	unsigned int	size;
+	unsigned int	rgno;
+};
+
+#ifdef CONFIG_BUILD_PROTECTED
```
yashwanth686007: Does it require no dependency on CONFIG_ARMV7M_MPU ?

os/arch/arm/include/armv7-m/irq.h
sangam-swami: modified and added changes for more mpu regions in case of M7

build/configs/imxrt1050-evk/nxp_demo_protected/defconfig
```diff
@@ -875,6 +892,7 @@ CONFIG_EXAMPLES_NXP_DEMO=y
 # CONFIG_EXAMPLES_SPEECH_DETECTOR_TEST is not set
 # CONFIG_EXAMPLES_ST_THINGS is not set
 # CONFIG_EXAMPLES_TESTCASE is not set
+# CONFIG_EXAMPLES_UPDATE_TEST is not set
```
kishore-sn: Please enable the configs for MPU TEST

sangam-swami: Can we separate feature and test config, it can be used only during testing

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: changed

os/arch/arm/src/imxrt/imxrt_mpuinit.c
sangam-swami: deleted

os/arch/arm/src/imxrt/mpu-reg.h
sangam-swami: changed

os/arch/arm/src/imxrt/imxrt_mpuinit.c
sangam-swami: cleaned up unnecessary changes

os/arch/arm/src/imxrt/mpu-reg.h
```diff
@@ -0,0 +1,58 @@
...
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+/*
+ * Region definitions for our platform
+ */
+#define MPU_REG_KERN_CODE	MPU_REG0
+#define MPU_REG_KERN_DATA	MPU_REG1
+#define MPU_REG_USER_CODE	MPU_REG2
+#define MPU_REG_USER_DATA	MPU_REG3
```
sunghan-chang: Should be they fixed?

sangam-swami: It is better to reserve region for robustness as we have enough regions, else it will create one more dependency in kernel for mapping and managing the regions.

kishore-sn: @sunghan-chang 
It is necessary to have fixed regions. 
When we add code for MPU in app separation scenario, the fixed regions becomes important. 
We will have a single region fixed for the app memory protection. Whenever there is a context switch, we need to update this MPU region with the new app's memory range.

sunghan-chang: Then, could you leave comments to say necessary of this and why?

And one more thing.
Is this file, os/arch/arm/src/imxrt/mpu-reg.h chip-specific?
Can't we make it chip-common file?

sunghan-chang: From your comment, I think they should be fixed regardless of chip or board. And it looks it does not have any chip-specific contents. Then why not common file?

kishore-sn: The 4 basic regions like kernel and user code and data are common to all chips. 
But there might be other regions which are specific to the chip. 
For example imxrt has dtcm, itcm and ocram which we might decide to use for some purpose. Then we need to configure separate regions for each of them. 
Other chips might not have these regions or might have some other memory regions to be configured. So, I think this is chip specific.

sangam-swami: yes it is chip specific, as I remember in broadcom case also there were reserved memory regions set. So the available regions might depend on chip

sunghan-chang: That means those 4 regions can be another numbers. Anyway we need 4 fixed regions for them but number can be changed. Am I right?

sangam-swami: yes, depends on reserved regions for chip

sangam-swami: updated

PR #2970, opened 2019-04-18
os/se/sss: Remove duplicate hal wrapper code

SeonghoByeon:
- Remove security_hal_wrapper.c in os/arch/arm/src/s5j/sss/
- Modify isp_custom.h path for header incldue
- Add seclink_drv.h in security_hal_wrapper.c

- Add TLS_WITH_SSS select option in Kconfig of SE
- Add SECURITY_LINK_DRV select option in Kconfig of SE


os/se/sss/security_hal_wrapper.c
```diff
@@ -47,7 +47,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include "isp_custom.h"
+#include "../../arch/arm/src/s5j/sss/isp_custom.h"
```
Hyunjun85: It is not good to me. 
But after this PR, security_hal module will move into os/driver. and It will be modified.

seinfra:
Target : [8d48a7e601a3ef832f69dc59e126801023ffb248] - Code Rule Check OK.

seinfra:
Target : [8d48a7e601a3ef832f69dc59e126801023ffb248] - Code Rule Check (C++) OK.

os/se/sss/security_hal_wrapper.c
SeonghoByeon: All isp code used in s5j/sss should be refactored soon. Thanks.

PR #2975, opened 2019-04-19
security: Add security HAL test cases

pillip8282:
Add HAL test case to verify HAL features defined by SE vendor.
It's prohibited to developers to call these API directly.
Developers who want to use TizenRT security feature have to call
Security APIs that will be supported later.
Furthermore It would work if binary separation feature is disabled.
So checking binary separation on Kconfig will be applied after it is
merged to public git


seinfra:
Target : [84cab45276fdd9b738728d4a581fb7134a6c7789] - Code Rule Check OK.

seinfra:
Target : [84cab45276fdd9b738728d4a581fb7134a6c7789] - Code Rule Check (C++) OK.

seinfra:
Target : [d02e173f66f0e4474bead7c4786420f205dd5a6b] - Code Rule Check (C++) OK.

seinfra:
Target : [d02e173f66f0e4474bead7c4786420f205dd5a6b] - Code Rule Check OK.

Taejun-Kwon:
1. Lower Device operation is not open API, why tc needed?
2. Lower Device operation should be hided to User
3. You have to disable unsupported function, return -ENOTTY or -ENOSYS, please apply our device code style
```
	if (lower->ops->fun == NULL) {
		return -ENOSYS;
	}
```

pillip8282:
@Taejun-Kwon 
I wrote the reason why did I add this API to commit message.
furthermore I think that this testcases make the vendor who want to port their solution to TizenRT to understand the HAL layer better.
And I think your suggestion is right when I write the driver code. but it's testcase code. so null checking could be skip in testcase code.

os/Directories.mk
```diff
@@ -250,3 +250,8 @@ KERNDEPDIRS += logm
 endif
 CLEANDIRS += logm
 
+ifeq ($(CONFIG_LOGM),y)
```
sunghan-chang: LOGM?

pillip8282: it's mistake. I'll fix it

PR #2977, opened 2019-04-21
os/se/sss: Modify printf to vdbg in security_hal_warpper

SeonghoByeon:
- Change the definition of HWRAP_LOG from printf to vdbg

### DIFF
--- a/os/se/sss/security_hal_wrapper.c
+++ b/os/se/sss/security_hal_wrapper.c
@@ -51,7 +51,12 @@
 
 #define ISP_CHECKBUSY() while (isp_get_status()) {}
 
-#define HWRAP_LOG printf
+#ifdef CONFIG_HAL_WRAPPER_DEBUG
+#define HWRAP_LOG(format, ...) printf(format, ##__VA_ARGS__)
+#else
+#define HWRAP_LOG(a, ...) (void)0
+#endif
+
 #define HWRAP_TAG "[HAL_WRAPPER]"
 #define HWRAP_ENTER                                                         \
 	do {                                                                    \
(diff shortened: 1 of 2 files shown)


os/se/sss/security_hal_wrapper.c
```diff
@@ -51,7 +51,7 @@
 
 #define ISP_CHECKBUSY() while (isp_get_status()) {}
 
-#define HWRAP_LOG printf
+#define HWRAP_LOG vdbg
```
sunghan-chang: Why do you use vdbg? This is not enough important to print with CONFIG_DEBUG_VERBOSE enabled.
Please use se-specific debug option.

SeonghoByeon: Thanks for the comment. Made a specific debug option and related code for security hal wrapper layer. Could you review the code if available.
Thanks again!

os/se/sss/security_hal_wrapper.c
```diff
@@ -51,7 +51,12 @@
 
 #define ISP_CHECKBUSY() while (isp_get_status()) {}
 
-#define HWRAP_LOG printf
+#ifdef CONFIG_HAL_WRAPPER_DEBUG
+#define HWRAP_LOG(format, ...) printf(format, ##__VA_ARGS__)
```
sunghan-chang: Please don't use `printf` in kernel. Please use xxdbg like other kernel module.

SeonghoByeon: Thanks for the comment. I will update soon.

PR #2985, opened 2019-04-23
os/include/debug.h : Revert dbg commands re-direction.

yashwanth686007:
* First patch adds dependency on Protected build on the following TC's: lldbg, llvdbg, lowsyslog, lowvsyslog. These API's call up_lowputc which is not accessible from userspace.
* Second patch eliminates the redirection of dbg to lldbg commands.
* Also, "dbg" cannot be called prior to driver registration since it uses write system call. So, this patch changes "dbg" to "lldbg" prior to driver registration.

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
@@ -257,6 +257,7 @@ static void tc_libc_misc_dbg(void)
 	TC_SUCCESS_RESULT();
 }
 
+#ifndef CONFIG_BUILD_PROTECTED
 /**
  * @fn                  :tc_libc_misc_lldbg
  * @brief               :Identical to [a-z]dbg() except this is uses special interfaces provided by architecture-specific logic to \n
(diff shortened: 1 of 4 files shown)


sunghan-chang:
@yashwanth686007 Could you let me know who calls up_lowputc?

yashwanth686007:
@sunghan-chang, lib_lowoutstream calls up_putc which internally calls lowputc.

sunghan-chang:
@yashwanth686007 With some specific conditionals, it calls up_putc. And it is already in protected build conditional. I don't think that blocking it regardless of another conditionals is right.

sunghan-chang:
@yashwanth686007 I agree changes of 2nd commit.

sunghan-chang:
@yashwanth686007 Could you let me know your opinion about my 1st commit's comment?

kishore-sn:
@sunghan-chang 
lowoutstream_putc enables up_putc only in case of KERNEL side libc code. 
`defined(CONFIG_BUILD_PROTECTED) && defined(__KERNEL__)`

We had planned this design because user side code never needs to use low level debug. It is required only by kernel side code during early board bring up.

sunghan-chang:
@kishore-sn I agree what you say. But is it related with this PR?
Source codes are implemented as what you say. Because of that, up_putc is not called when only protected build is enabled. Why do we need to block tc even though it is already in proper conditional? That's my comment.

PR #2986, opened 2019-04-23
framework/media: fix mp3 header parsing

jsdosa:
The logic to find header of mp3 is incorrect

Signed-off-by: jsdosa <jsdosa@gmail.com>

### DIFF
--- a/framework/src/media/utils/MediaUtils.cpp
+++ b/framework/src/media/utils/MediaUtils.cpp
@@ -307,24 +307,35 @@ bool header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel, uns
 	unsigned char *header;
 	unsigned char tag[2];
 	bool isHeader;
+	int ret;
 
 	switch (audioType) {
 	case AUDIO_TYPE_MP3:
 		isHeader = false;
-		while (fread(tag, sizeof(unsigned char), 1, fp) > 0) {
-			if (tag[0] == 0xFF) {
+		while (fread(tag, sizeof(unsigned char), 2, fp) == 2) {
+			/* 12 bits for MP3 Sync Word(the beginning of the frame) */
+			if ((tag[0] == 0xFF) && ((tag[1] & 0xF0) == 0xF0)) {
 				isHeader = true;
 				break;
+			} else {
+				/* If isn't the header information, go back 1byte and then check 2bytes again */
+				ret = fseek(fp, -1, SEEK_CUR);
+				if (ret != OK) {
+					meddbg("file seek failed errno : %d\n", errno);
+					free(header);
+					return false;
+				}
 			}
 		}
 		if (isHeader) {
 			header = (unsigned char *)malloc(sizeof(unsigned char) * (MP3_HEADER_LENGTH + 1));
+			
 			if (header == NULL) {
 				meddbg("malloc failed error\n");
 				return false;
 			}
-			int ret;
-			ret = fseek(fp, -1, SEEK_CUR);
+
+			ret = fseek(fp, -2, SEEK_CUR);
 			if (ret != OK) {
 				meddbg("file seek failed errno : %d\n", errno);
 				free(header);
@@ -420,7 +431,8 @@ bool header_parsing(unsigned char *buffer, unsigned int bufferSize, audio_type_t
 		isHeader = false;
 		headPoint = 0;
 		while (headPoint < bufferSize) {
-			if (buffer[headPoint] == 0xFF) {
+			/* 12 bits for MP3 Sync Word(the beginning of the frame) */
+			if ((buffer[headPoint]) == 0xFF && ((buffer[headPoint + 1] & 0xF0) == 0xF0)) {
 				isHeader = true;
 				break;
 			}


framework/src/media/utils/MediaUtils.cpp
```diff
@@ -307,14 +307,22 @@ bool header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel, uns
...
 	bool isHeader;
+	int ret;
 
 	switch (audioType) {
 	case AUDIO_TYPE_MP3:
 		isHeader = false;
-		while (fread(tag, sizeof(unsigned char), 1, fp) > 0) {
-			if (tag[0] == 0xFF) {
+		while (fread(tag, sizeof(unsigned char), 2, fp) > 0) {
+			if ((tag[0] == 0xFF) && ((tag[1] & 0xE0) == 0xE0)) {
```
Taejun-Kwon: Please add simple description of below logic & mean of 0xE0(line 430 too). Then I will apply it.

jsdosa: i added description.

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -420,7 +428,8 @@ bool header_parsing(unsigned char *buffer, unsigned int bufferSize, audio_type_t
 		isHeader = false;
 		headPoint = 0;
 		while (headPoint < bufferSize) {
-			if (buffer[headPoint] == 0xFF) {
+			/* 11 bits for MP3 Sync Word(the beginning of the frame) */
+			if (buffer[headPoint] == 0xFF && ((buffer[headPoint + 1] & 0xE0) == 0xE0)) {
```
Taejun-Kwon: add one bracket set for first condition.

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -307,14 +307,23 @@ bool header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel, uns
...
 		isHeader = false;
-		while (fread(tag, sizeof(unsigned char), 1, fp) > 0) {
-			if (tag[0] == 0xFF) {
+		while (fread(tag, sizeof(unsigned char), 2, fp) > 0) {
+			/* 11 bits for MP3 Sync Word(the beginning of the frame) */
+			if ((tag[0] == 0xFF) && ((tag[1] & 0xE0) == 0xE0)) {
 				isHeader = true;
 				break;
+			} else {
+				ret = fseek(fp, -1, SEEK_CUR);
```
Taejun-Kwon: you had read 2 but move back to '-1'..
Isn't '-2' is correct?

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -323,8 +332,7 @@ bool header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel, uns
 				meddbg("malloc failed error\n");
 				return false;
 			}
-			int ret;
-			ret = fseek(fp, -1, SEEK_CUR);
+			ret = fseek(fp, -2, SEEK_CUR);
```
Taejun-Kwon: I think calling fseek before malloc is more comfortable.
If malloc is failed, it will skip header forever.

framework/src/media/utils/MediaUtils.cpp
jsdosa: -1 is correct.
because of,  this case is 0x00 0xFF 0xF3

framework/src/media/utils/MediaUtils.cpp
jsdosa: currently, If either of them is fail, the result is the same.
I will cleanup code later.

framework/src/media/utils/MediaUtils.cpp
jsdosa: i added bracket

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -307,24 +307,35 @@ bool header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel, uns
...
 	unsigned char tag[2];
 	bool isHeader;
+	int ret;
 
 	switch (audioType) {
 	case AUDIO_TYPE_MP3:
 		isHeader = false;
-		while (fread(tag, sizeof(unsigned char), 1, fp) > 0) {
-			if (tag[0] == 0xFF) {
+		while (fread(tag, sizeof(unsigned char), 2, fp) == 2) {
```
Taejun-Kwon: If fread failed because of internal problem, we should stop below all
```
ssize_t ret;
while(1) {
    ret = fread();
    if (ret != 2) {
        error_log(errno());
        return ERROR;
    }
    if (check_sync_word) {
         is_header_true;
    } else {
         seek();
   }
}
```

seinfra:
Target : [b5b99ee787243ee2e6b80980486c6b4e7403e00d] - Code Rule Check (C++) OK.

framework/src/media/utils/MediaUtils.cpp
jsdosa: you right.
but, many places use fseek and fread. so, i will cleanup later.

Taejun-Kwon: got it. please do not forget it.

seinfra:
Target : [b5b99ee787243ee2e6b80980486c6b4e7403e00d] - Code Rule Check OK.

seinfra:
Target : [b5b99ee787243ee2e6b80980486c6b4e7403e00d] - Code Rule Check (C++) OK.

seinfra:
Target : [b5b99ee787243ee2e6b80980486c6b4e7403e00d] - Code Rule Check (C++) OK.

framework/src/media/utils/MediaUtils.cpp
jsdosa: yes

jsdosa:
> I think that you need to change the header check of aac.
> I recommend that you check the 2 bytes combination or 12bit from the front for syncword.

yes, you right.
I will update to the next commit.

PR #2987, opened 2019-04-23
apps/examples/wifi_manager_sample: Supporting various auth/crypto

SeonghoByeon:
- Add various authentication and crypto methods defined in wifi_manager
- AUTH: open, wep_shared, wpa, wpa2, wpa12, wps, and enterprise
- CRPYTO: none, wep_shared 64/128, aes, tkip, mixed, and enterprise

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test.c
+++ b/apps/examples/wifi_manager_sample/wm_test.c
@@ -49,6 +49,11 @@
 	"	 wm_test mode\n\n"												\
 	"\n set a profile:\n"											\
 	"	 wm_test set [ssid] [security mode] [password]\n"			\
+	"	 security mode examples : open, wep_shared \n"						\
+	"	               wpa_aes, wpa_tkip, wpa_mixed  \n"			\
+	"	               wpa2_aes, wpa2_tkip, wpa2_mixed  \n"			\
+	"	               wpa12_aes, wpa12_tkip, wpa12_mixed  \n"		\
+	"	               (*_ent for enterprise)  \n"					\
 	"\n get a profile:\n"											\
 	"	 wm_test get\n"												\
 	"\n remove a profile:\n"										\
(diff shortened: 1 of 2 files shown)


apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -262,29 +295,54 @@ void print_wifi_softap_profile(wifi_manager_softap_config_s *config, char *title
...
+	result[0] = strtok_r(data, "_", &next_ptr);
+	result[1] = strtok_r(NULL, "_", &next_ptr);
+	result[2] = strtok_r(NULL, "_", &next_ptr);
+
 	int i = 0;
 	int list_size = sizeof(wifi_test_auth_method)/sizeof(wifi_test_auth_method[0]);
 	for (; i < list_size; i++) {
 		if (strcmp(method, wifi_test_auth_method[i]) == 0) {
-			return auth_type_table[i];
+			if (strcmp(result[2], "ent") == 0) {
```
Taejun-Kwon: 1. result[1] is not used here, is it explicit approach?
2. 'else if' part also checking 'result[2]' and 'ent', then below is more good.
```
if ((strcmp(method, wifi_test_auth_method[i]) == 0) || (strcmp(result[0], wifi_test_auth_method[i]) == 0)) {
    if (stcmp(result[2], "ent") == 0) {
         return auth_type_table[i + 2];
    }
    return auth_type_table[i];
}

File: apps/examples/wifi_manager_sample/wm_test.c [hunk 2]
```diff
@@ -262,29 +295,54 @@ void print_wifi_softap_profile(wifi_manager_softap_config_s *config, char *title
...
 	}
 	return WIFI_MANAGER_AUTH_UNKNOWN;
 }
 
 wifi_manager_ap_crypto_type_e get_crypto_type(const char *method)
 {
+	char data[20];
+	strcpy(data, method);
+
+	char *result[2];
```
Taejun-Kwon: actually result is not necessary.

apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -599,6 +662,11 @@ void wm_connect(void *arg)
 		apconfig.passphrase_length = strlen(ap_info->password);
 		apconfig.ap_crypto_type = ap_info->crypto_type;
 	}
+	else{
```
Taejun-Kwon: coding rule.

apps/examples/wifi_manager_sample/wm_test.c
SeonghoByeon: > 1. result[1] is not used here, is it explicit approach?
That's right. In this case, result[1] exists only for handle the dummy input. For example, when an input string is 'wpa_aes_ent', result[1] will be never used in this function, because it requires only 'wpa' and 'ent'. It might be possible to use result[1] = strtok_r(NULL, "_", &next_ptr); double times, but it does look not good for readibility, in my opinion.

> 2. 'else if' part also checking 'result[2]' and 'ent', then below is more good.
Sounds perfect. Thanks for the comment. Applied.

apps/examples/wifi_manager_sample/wm_test.c
SeonghoByeon: Thanks for the comment. That's right, by calling strtok_r(NULL, "_", &next_ptr); again. However thought it does not good for readability and result[1] might be used in the future, when more types of auth/crypto algorithms are added.

apps/examples/wifi_manager_sample/wm_test.c
SeonghoByeon: Completed. Thanks!

seinfra:
Target : [d50956fb09e1e05069e9dc405a7ab2039acb9ea1] - Code Rule Check Result:
apps/examples/wifi_manager_sample/wm_test.c:1318: ERROR: [SPC_M_SEP] space prohibited before that close parenthesis ')'
apps/examples/wifi_manager_sample/wm_test.c:1320: ERROR: [SPC_M_SEP] space prohibited before that close parenthesis ')'
apps/examples/wifi_manager_sample/wm_test.c:1349: ERROR: [SPC_M_SEP] space prohibited before that close parenthesis ')'
apps/examples/wifi_manager_sample/wm_test.c:1351: ERROR: [SPC_M_SEP] space prohibited before that close parenthesis ')'

seinfra:
Target : [d50956fb09e1e05069e9dc405a7ab2039acb9ea1] - Code Rule Check (C++) OK.

seinfra:
Target : [d50956fb09e1e05069e9dc405a7ab2039acb9ea1] - Code Rule Check (C++) OK.

seinfra:
Target : [5f87be85b59675abe7e0e4c598be79838339d058] - Code Rule Check (C++) OK.

seinfra:
Target : [5f87be85b59675abe7e0e4c598be79838339d058] - Code Rule Check OK.

PR #2989, opened 2019-04-23
framework/media: Modify SW EPD logic

junmin-kim:
EPD doesn't detect speech -> no-speech, but detects no-speech.
Add holdTime to detect no-speech for a period of time.

### DIFF
--- a/framework/src/media/voice/SoftwareEndPointDetector.cpp
+++ b/framework/src/media/voice/SoftwareEndPointDetector.cpp
@@ -22,9 +22,7 @@ namespace media {
 namespace voice {
 
 SoftwareEndPointDetector::SoftwareEndPointDetector() :
-	mState(nullptr),
-	mPreviousVAD(0),
-	mVAD(0)
+	mState(nullptr)
 {
 	sem_init(&mSem, 0, 0);
 }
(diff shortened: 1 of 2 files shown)


Taejun-Kwon:
Let's think about given condition first.
1. Upper layer of media f/w doesn't know which type of epd selected.
2. They will check epd based on period of silence time.
   - That is, they will call 'detect' again & again.
   - This given condition regarding 'period' is not opened to lower layer, because upper layer will call it by itself.
3. So s/w based epd's role is that decide silence level based on current sound data. see below example.
```
upper_process_detect()
{
   bool silence;
   short *data = getAudioData();
   int dataSize = getAudioDataSize();
   int framesize = channel * (bitRate>>3);
   int epdProcessSize = mSpeechDetector.getProcessSize(); //size of snr checking routine in f/w, can be changed to config value
   /* If size of data is greater than process size, we check it in loop */
   for(i = 0; i < dataSize; i++) {
       silence = mSpeechDetector.detect(data + i, (dataSize/frameSize)); //send data & give Size of frame
       if (mTotalSilenceFrames > EPD_LIMIT) { // We can configure EPD_LIMIT value, in cofig file in upper.
           handleSilence();
           break;
       }
   }
   ....
}

bool SoftwareEndPointDetector::detectEndPoint(short *sample, int numSample)
{
     return ( speex_preprocess_run()== 0 ? true : false);
}
```
4. To use H/W epd, we have a 2 option, see below

 a) change EPD_LIMIT to smallest value(0  or 1??)
```
bool HardwareEndPointDetector::detectEndPoint(short *sample, int numSample)
{
    do {
        result = get_device_process_handler_message(mSdCard, mSdDevice, &msgId);
        switch(msgId) {
            case AUDIO_DEVICE_SPEECH_DETECT_EPD:
                  return true;
            default :
                  return false;
        }
}
```
 b) Driver Change its mechanism. Now it notify to audiomanager, but now it just maintain its state.

```
bool HardwareEndPointDetector::detectEndPoint(short *sample, int numSample)
{
       result = get_device_process_state(mSdCard, mSdDevice);
       switch(result) {
       case AUDIO_DEVICE_SPEECH_DETECT_EPD:
            return true;
       default :
            return false;
        }
}

get_device_process_state(int card, int device)
{
    int fd = open(path);
    long *args = AUDIO_DEVICE_SPEECH_DETECT_EPD;
    ret = ioctl(fd, AUDIOIOC_GETCAPS, args);
    if (ret == AUDIO_MANAGER_SUCESS) {
       return AUDIO_DEVICE_SPEECH_DETECT_EPD;
   }
}

static int audio_getcaps(FAR struct audio_lowerhalf_s *dev, int type, FAR struct audio_caps_s *caps)
{
    ...
    case AUDIO_DEVICE_SPEECH_DETECT_EPD;
          if ((priv->running) || (priv->state != SPEECH_STATE_NONE)) { /* running means absolutely recording state, if it wating KD or epd*/
             return -EAGAIN; //try again
          }
          return OK;
    ...
}
```

seinfra:
Target : [b890f88a646eee8ec2691fc624fadcef91fb476a] - Code Rule Check (C++) OK.

seinfra:
Target : [b890f88a646eee8ec2691fc624fadcef91fb476a] - Code Rule Check (C++) OK.

seinfra:
Target : [d226757f99642d012998e21400d6892aa06b74f8] - Code Rule Check (C++) OK.

seinfra:
Target : [d226757f99642d012998e21400d6892aa06b74f8] - Code Rule Check Result:
framework/src/media/voice/SoftwareEndPointDetector.h:42: ERROR: [SPC_M_OPR] spaces prohibited around that ':' (ctx:WxW)

PR #2993, opened 2019-04-24
Exit build on undefined symbols in elf

sijupunnnoose:
Add verification of elf using nm command,  since the elf are relocatable undefined symbols will not be considered as error during linking. With this check build will stop if any undefined symbols is observed during build.

### DIFF
--- a/apps/examples/elf/micomapp/Makefile
+++ b/apps/examples/elf/micomapp/Makefile
@@ -84,12 +84,12 @@ LDLIBS += $(LIBGCC)
 endif
 
 BIN = micomapp
-
+BIN_CHECK = $(shell nm -u $(PWD)/$(BIN)/$(BIN) | wc -l)
 SRCS = micomapp.c
 OBJS = $(SRCS:.c=$(OBJEXT))
 
 all: $(BIN)
-.PHONY: clean install
+.PHONY: clean install verify
 
 $(OBJS): %$(OBJEXT): %.c
 	@echo "CC: $<"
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [0994d9c657c332c609af2535b7ff71e28a3cd1fb] - Code Rule Check OK.

seinfra:
Target : [0994d9c657c332c609af2535b7ff71e28a3cd1fb] - Code Rule Check (C++) OK.

seinfra:
Target : [0994d9c657c332c609af2535b7ff71e28a3cd1fb] - Code Rule Check (C++) OK.

apps/examples/elf/micomapp/Makefile
```diff
@@ -107,3 +107,13 @@ clean:
 install:
 	$(Q) mkdir -p $(SYSTEM_BIN_DIR)
 	$(Q) install $(BIN) $(SYSTEM_BIN_DIR)/$(BIN)
+
+verify:
+ifneq ($(BIN_CHECK),0)
+	@echo "Undefined Symbols"
+	$(Q) $(NM) -u -l $(BIN)
+	$(call DELFILE, $(SYSTEM_BIN_DIR)/$(BIN))
```
kishore-sn: Is it necessary to delete the bin file here?
We could just return after printing the error message. 
It is up to the user to perform clean. 
The elf file might be required by the user to check and fix the error messages.

sijupunnnoose: Since we are verifying it after the linking, object and binary would be generated already. If  developer/user build again without resolving undefined error, build will pass and there wont be any build error. To avoid it we delete object/binary if there is undefined symbols.

kishore-sn: ok

PR #3036, opened 2019-05-01
config/loadable_elf_apps: remove unnecessary CONFIG_USER_ENTRYPOINT

davidfather:
When binaries can be loaded at runtime, we no longer need a starting point from kernel to user apps
because the binary manager takes this responsibility.


seinfra:
Target : [0b2fe98d175a62b5b99ce0ddfe443b12edda721b] - Code Rule Check (C++) OK.

seinfra:
Target : [0b2fe98d175a62b5b99ce0ddfe443b12edda721b] - Code Rule Check OK.

PR #3045, opened 2019-05-02
kernel/init: start the preapp task regardless of binary manager

sunghan-chang:
The preapp task which launches TASH and system tasks does not
have any dependency with binary manager.
So, it should be outside of binary manager conditional.
This commit fixes an issue that TASH is not come up.


seinfra:
Target : [4e6091b72ce72186719d6576e9864317c996ef45] - Code Rule Check (C++) OK.

seinfra:
Target : [4e6091b72ce72186719d6576e9864317c996ef45] - Code Rule Check OK.

PR #3054, opened 2019-05-03
Add MPU setting for SD-RAM kernel heap

kishore-sn:
If SD RAM is enabled, then we allocate kernel heap region on SD RAM.
This patch adds MPU protection for this heap region. Also, we will
use enum based region type to region number mapping.


seinfra:
Target : [6524102bb258b9f738c20fba8eb5bef5e3140306] - Code Rule Check (C++) OK.

seinfra:
Target : [6524102bb258b9f738c20fba8eb5bef5e3140306] - Code Rule Check OK.

PR #3070, opened 2019-05-07
drivers/analog: update poll event for ADC

an4967:
- update poll notify function for ADC.

### DIFF
--- a/os/drivers/analog/adc.c
+++ b/os/drivers/analog/adc.c
@@ -90,6 +90,10 @@ static ssize_t adc_read(FAR struct file *fielp, FAR char *buffer,
 static int     adc_ioctl(FAR struct file *filep, int cmd, unsigned long arg);
 static int     adc_receive(FAR struct adc_dev_s *dev, uint8_t ch,
 			   int32_t data);
+static void    adc_notify(FAR struct adc_dev_s *dev);
+#ifndef CONFIG_DISABLE_POLL
+static int     adc_poll(FAR struct file *filep, struct pollfd *fds, bool setup);
+#endif
 
 /****************************************************************************
  * Private Data
@@ -102,7 +106,7 @@ static const struct file_operations g_adc_fops = {
 	0,		/* seek */
 	adc_ioctl,	/* ioctl */
 #ifndef CONFIG_DISABLE_POLL
-	0,		/* poll */
+	adc_poll,		/* poll */
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [8a439118e74d34a2ec9fbc8cdb9233207b986990] - Code Rule Check (C++) OK.

seinfra:
Target : [8a439118e74d34a2ec9fbc8cdb9233207b986990] - Code Rule Check Result:
os/drivers/analog/adc.c:457: ERROR: [IDT_M_TAB] code indent should use tabs where possible
os/drivers/analog/adc.c:457: ERROR: [IDT_M_TAB] please, no space before tabs
os/drivers/analog/adc.c:457: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/drivers/analog/adc.c:491: ERROR: [BRC_M_SMT] that open brace { should be on the previous line
os/drivers/analog/adc.c:495: ERROR: [IDT_M_TAB] please, no spaces at the start of a line
os/drivers/analog/adc.c:504: ERROR: [IDT_M_TAB] please, no spaces at the start of a line

seinfra:
Target : [f23f6f54fe2ae7e74d09aa815a16d80d66e52bf0] - Code Rule Check OK.

seinfra:
Target : [f23f6f54fe2ae7e74d09aa815a16d80d66e52bf0] - Code Rule Check (C++) OK.

seinfra:
Target : [ad6fc6ee1d4848615bb27716d275842bf83ba69a] - Code Rule Check (C++) OK.

seinfra:
Target : [ad6fc6ee1d4848615bb27716d275842bf83ba69a] - Code Rule Check OK.

os/drivers/analog/adc.c
```diff
@@ -381,16 +385,129 @@ static int adc_receive(FAR struct adc_dev_s *dev, uint8_t ch, int32_t data)
...
+			if (!dev->fds[i]) {
+				/* Bind the poll structure and this slot */
+
+				dev->fds[i] = fds;
+				fds->priv   = &dev->fds[i];
+				break;
+			}
+		}
+
+		if (i >= CONFIG_ADC_NPOLLWAITERS) {
```
SeonghoByeon: After running for loop above,
'i' will never be larger than CONFIG_ADC_NPOLLWAITER,
but becomes CONFIG_ADC_NPOLLWAITER, when no room for the new poll waiter. 
It might be better to manage a counter to follow up the number of registered poll waiters, and check it before running for loop.

os/drivers/analog/adc.c
```diff
@@ -381,16 +385,129 @@ static int adc_receive(FAR struct adc_dev_s *dev, uint8_t ch, int32_t data)
...
+	} else if (fds->priv) {
+		/* This is a request to tear down the poll. */
+
+		struct pollfd **slot = (struct pollfd **)fds->priv;
+
+		/* Remove all memory of the poll setup */
+
+		*slot                = NULL;
+		fds->priv            = NULL;
+	}
```
SeonghoByeon: In case 'setup' is 0, and fds->priv is null, returning value becomes 0, which is not appropriate. It would be better to add 'else' logic.

an4967: @sunghan-chang 
In case, setup is 0, there is nothig to do in this function.
'else if (fds->priv)' means just clean the 'fds->priv' only when setup is 0

os/drivers/analog/adc.c
an4967: @sunghan-chang 
the pollfds is an array, it can be checked empty fd after loop.
'>='  just emphasizes the given npollwaiter is overflowed.

PR #3084, opened 2019-05-09
security: change the logic of getting HAL data in security API

pillip8282:
security API which is in user space can't access kernel memory.
so if hal which is in kernel creates memory then security API can't
access it. instead of allocating memory in kernel space, security API
allocates memory and give it to hal to write result.
So inside the security handle, there are couple of memory to get data
from HAL and assign memory space when security_init is called. vice
versa it will be removed on security_deinit


seinfra:
Target : [61e154129d9d5c4b595d356dfb725ff80c71bc44] - Code Rule Check (C++) OK.

seinfra:
Target : [61e154129d9d5c4b595d356dfb725ff80c71bc44] - Code Rule Check OK.

framework/src/security/security_keymgr.c
```diff
@@ -114,7 +117,7 @@ int keymgr_get_key(security_handle hnd, security_key_type algo, const char *key_
 		SECAPI_FREE(pubkey_x);
 		SECAPI_RETURN(SECURITY_ALLOC_ERROR);
 	}
-	SECAPI_DATA_DCOPY(h_pubkey, pubkey_x);
+	SECAPI_PRIV_DCOPY(h_pubkey, pubkey_y);
```
chanijjani: Is SECAPI_PRIV_DCOPY correct rather than SECAPI_DATA_DCOPY?

framework/include/security/security_common.h
```diff
@@ -21,10 +21,10 @@
 
 #include <stdint.h>
 
-#define SECURITY_MAX_BUF              1024
 #define SECURITY_MAX_KEY_BUF          256
 #define SECURITY_MAX_CERT_BUF         2048
 #define SECURITY_MAX_SS_BUF           4096
+#define SECURITY_MAX_BUF              SECURITY_MAX_SS_BUF
```
chanijjani: Where SECURITY_MAX_SS_BUF is defined..? Is SECURITY_MAX_SS_BUF 1024?

framework/src/security/security_keymgr.c
pillip8282: they copy different member.

framework/include/security/security_common.h
pillip8282: line 26 it's defined.

PR #3090, opened 2019-05-10
os/fs/procfs: Add up current and peak heap in all heaps

zhouxinhe:
In current logic, it only get the usage from one heap.
It's not exact, when user configure multiple heaps and
the task/thread may allocate memory from different heaps.

### DIFF
--- a/os/fs/procfs/fs_procfsproc.c
+++ b/os/fs/procfs/fs_procfsproc.c
@@ -366,6 +366,7 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	struct mm_heap_s *heap;
 	pid_t hash_pid;
+	int heap_idx;
 #endif
 	if (tcb->pid == 0) {
 		tcb->adj_stack_size = CONFIG_IDLETHREAD_STACKSIZE;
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [a4035cc4ea1908e400c925465f846a4e05c67be4] -

seinfra:
Target : [a4035cc4ea1908e400c925465f846a4e05c67be4] -

os/fs/procfs/fs_procfsproc.c
```diff
@@ -381,24 +382,24 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
...
-		/* Idle task normally uses heap with index 0. */
-		heap = mm_get_heap_with_index(0);
-	} else {
-		heap = mm_get_heap(tcb->stack_alloc_ptr);
-	}
-	if (heap == NULL) {
-		return -1;
-	}
-
+	curr_heap = 0;
```
SRCN-LiQin: this 0 initialization could refine the logic of += using in for loop, but it's changing the default -1 value when there is no alloc_size for current tcb. how about initialize it to -1 as before, and check it after for loop to revise the -1

zhouxinhe: If CONFIG_DEBUG_MM_HEAPINFO is not defined, the default value is -1 (feature not support). But if it's defined, -1 is an illegal value (the minimal value is 0).

seinfra:
Target : [0d0714d007103d4de03696b6a940bc92c0f6715e] - Code Rule Check OK.

seinfra:
Target : [0d0714d007103d4de03696b6a940bc92c0f6715e] - Code Rule Check (C++) OK.

zhouxinhe:
@jeongchanKim Could you please help to review?

PR #3129, opened 2019-05-15
binfmt/loadbinary: Set signal mask of task loaded by binary manager

jeongarmy:
The task loaded by binary manager has same signal mask with binary manager.
That's because when a task or thread is created, it inherits signal mask from parent task.
Binary manager blocks all signals excluding SIGKILL so we need to clean it.

### DIFF
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -25,6 +25,9 @@
 #include <string.h>
 #include <debug.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [89c110e5aea0c3915d1ba0e7f22ea0bc0148b1d7] - Code Rule Check OK.

seinfra:
Target : [89c110e5aea0c3915d1ba0e7f22ea0bc0148b1d7] - Code Rule Check (C++) OK.

PR #3150, opened 2019-05-16
os/arch/xtensa: change function board_app_initialize() defination to …

glb1013:
argument in board_app_initialize() defination is different with its prototype in os\include\tinyara\board.h;  esp_err.h may be include more than one time.

### DIFF
--- a/os/arch/xtensa/src/esp32/esp_err.h
+++ b/os/arch/xtensa/src/esp32/esp_err.h
@@ -16,6 +16,10 @@
  *
  ******************************************************************/
 
+#ifndef _ESP_ERR_H_
+#define _ESP_ERR_H_ 1
+
+#include "stdint.h"
 
 typedef int32_t esp_err_t;
 #define ESP_OK          0		/*!< esp_err_t value indicating success (no error) */
@@ -35,3 +39,5 @@ typedef int32_t esp_err_t;
 
 #define ESP_ERR_WIFI_BASE       0x3000	/*!< Starting number of WiFi error codes */
 #define ESP_ERR_MESH_BASE       0x4000	/*!< Starting number of MESH error codes */
+
+#endif //_ESP_ERR_H_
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [513b5469bd5d29a7a09e3776d1a9ca5b0abed4d5] -

seinfra:
Target : [513b5469bd5d29a7a09e3776d1a9ca5b0abed4d5] -

os/arch/xtensa/src/esp32/esp_err.h
```diff
@@ -16,6 +16,10 @@
  *
  ******************************************************************/
 
+#ifndef _ESP_ERR_H_
+#define _ESP_ERR_H_ 1
+
+#include "stdint.h"
```
zhouxinhe: how about using `#include <stdint.h>` ?

ChinaTizenRT:
these two code refine seems  irrelevant, how about spilt to 2 commit?

os/arch/xtensa/src/esp32/esp_err.h
xixidodo: #define _ESP_ERR_H_ is enough

and besides, we can remove duplicate 'esp_err.h' in esp_idf_port, right?

os/arch/xtensa/src/esp32/esp_err.h
```diff
@@ -16,6 +16,10 @@
  *
  ******************************************************************/
 
+#ifndef _ESP_ERR_H_
+#define _ESP_ERR_H_ 1
```
xiarihf: why if not define A then define B,  but #endif comment is A.

os/arch/xtensa/src/esp32/esp_err.h
xiarihf: Agree with xixidodo :)

xiarihf:
Agree with ChinaTizenRT, It's better to split to 2 commits.

PR #3174, opened 2019-05-20
os/include/debug.h: Remove re-direction of dbg to lldbg

yashwanth686007:
* This patch eliminates the re-direction of dbg to lldbg
commands.
* Also, "dbg" cannot be called prior to driver registration
since it uses write system call. So, this patch changes
"dbg" to "lldbg" prior to driver registration.


seinfra:
Target : [c333e547b6079b5c11a746b38ba37f6e2222b2f4] - Code Rule Check (C++) OK.

seinfra:
Target : [c333e547b6079b5c11a746b38ba37f6e2222b2f4] - Code Rule Check OK.

PR #3176, opened 2019-05-20
kernel/binary_manager: Remove the TCB from the task list associated with the state in recovery

jeongarmy:
When binary manager handles a fault, All children created in a same binary are excluded from scheduling.
At this time, we need to remove the TCB from the task list associated with the state.


seinfra:
Target : [b170efc4a4eac2ce5cf1ccd183490f78e312a705] - Code Rule Check (C++) OK.

seinfra:
Target : [b170efc4a4eac2ce5cf1ccd183490f78e312a705] - Code Rule Check OK.

PR #3188, opened 2019-05-21
Fix compilation issue for artik053

drashti304:
Fix compilation issue for artik053/elf config

Add IMXRT dependency for mpu_user_intsram_context function, as it is not implemented for ARTIK053 to solve the compilation issue


seinfra:
Target : [c192908f72d61ed7c9cadf3d3b1ff10526b1136d] - Code Rule Check (C++) OK.

seinfra:
Target : [c192908f72d61ed7c9cadf3d3b1ff10526b1136d] - Code Rule Check OK.

os/binfmt/binfmt_exec.c
```diff
@@ -211,10 +211,9 @@ int exec(FAR const char *filename, FAR char *const *argv, FAR const struct symta
 	tcb->ram_start = (uint32_t)start_addr;
 
 	/* Initialize the MPU registers in tcb with suitable protection values */
-#ifdef CONFIG_ARMV7M_MPU
+#if defined(CONFIG_ARMV7M_MPU) && defined(CONFIG_ARCH_CHIP_IMXRT)
```
yashwanth686007: i think chip specific code should not be present here.

sunghan-chang: I agree @yashwanth686007 's opinion.
If this is not chip specific, you had better turn on the config of disabling binfmt in ARTIK.

I already asked @kishore-sn that why binfmt is enabled by default even though app_separation is not a default.
If we change `CONFIG_BINFMT_DISABLE` to `CONFIG_BINFTM_ENABLE`, there is no problem. I think.

build/configs/artik053/elf/defconfig
```diff
@@ -876,6 +876,8 @@ CONFIG_BUILTIN_APPS=y
 # CONFIG_EXAMPLES_EEPROM_TEST is not set
 CONFIG_EXAMPLES_ELF=y
 CONFIG_EXAMPLES_ELF_FULLYLINKED=y
+CONFIG_MPU_TEST_KERNEL_CODE_ADDR=0x00000000
+CONFIG_MPU_TEST_APP_ADDR=0x00000000
```
sunghan-chang: Why do they have 0x0?

sunghan-chang:
@drashti304 Could you update this?

sunghan-chang:
@drashti304 As we discuss, let's close this and make that config as by defulat disabled.

PR #3193, opened 2019-05-22
security/utc: Fix security UTC bug

pillip8282:
There are several issues in security UTC.
1) confuse asymmetric key list and symmetric key list which are
defined at UTC. so it causes unexpected errors
2) there is no logic that checks whether seconds public key is
exist. so it allocate unused memory for key and make an error.


seinfra:
Target : [ef1cb8c9ed0ce941ea1b919b63f7ae8605474466] - Code Rule Check (C++) OK.

seinfra:
Target : [ef1cb8c9ed0ce941ea1b919b63f7ae8605474466] - Code Rule Check OK.

PR #3199, opened 2019-05-23
arch/imxrt/gpt: set proper divider with clock source in imxrt_gpt_init

sunghan-chang:
There are two functions of setting clock divider, imxrt_gpt_setclockdivider and
imxrt_gpt_setoscclockdivider.
When clock source is kGPT_ClockSource_Osc, imxrt_gpt_setoscclockdivider should
be used to set.
But imxrt_gpt_init function only uses setclockdivider so that it is wrong with
OSC clock.
This commit adds a conditional to check clock source and uses proper function.


seinfra:
Target : [0be321cd65062f6c7e2c7b52180a196040dbb849] - Code Rule Check (C++) OK.

seinfra:
Target : [0be321cd65062f6c7e2c7b52180a196040dbb849] - Code Rule Check OK.

PR #3213, opened 2019-05-27
kernel/binary_manager: Fix wrong dependency conditions

jeongarmy:
FLASH_PART_SIZE, FLASH_PART_TYPE and FLASH_PART_NAME have a string type, not bool type.
So we need to make a condition that these values are not NULL.


seinfra:
Target : [0957b1a7a3ddf27fb757f97734180cd4c66591ce] - Code Rule Check OK.

seinfra:
Target : [0957b1a7a3ddf27fb757f97734180cd4c66591ce] - Code Rule Check (C++) OK.

PR #3218, opened 2019-05-27
framework/st_things: Increase sleep time to apply Wi-Fi state changing.

Hyunjun85:
Error has been showing during Wi-Fi state changing.
It guesses that side effect of IPC changes. 

And now, there is no callback api about Wi-Fi state changes like "stop_softap_done".
So, It is one of refactoring point for st_things. 

In future, sleep will be changed to some callback to get result of Wi-Fi state changing.


seinfra:
Target : [f3a8019d6b23ae324c4cab3b4b835082f25321d2] - Code Rule Check (C++) OK.

seinfra:
Target : [f3a8019d6b23ae324c4cab3b4b835082f25321d2] - Code Rule Check OK.

PR #3232, opened 2019-05-28
arch/arm/src/armv7-m : Add stack overflow check for nested interrupts

drashti304:
This patch adds the support of stack overflow check for nested interrupt.
If stack grows more than the configured interrupt stack size then
this check will assert the system.

### DIFF
--- a/os/arch/Kconfig
+++ b/os/arch/Kconfig
@@ -584,8 +584,15 @@ config ARCH_NESTED_INTERRUPT
 	---help---
 		Enables the nested interrupt support. If defined, higher priority interrupt
 		can pre-empt the lower priority interrupt.
-endif
 
+config ARCH_NESTED_INTERRUPT_STACKCHECK
+	bool "Nested interrupt stack check support"
+	depends on ARCH_NESTED_INTERRUPT && STACK_COLORATION
+	default n
+	---help---
+		Enables the nested interrupt stack check support. It checks every time before
+		serving the interrupt, if the stack overflows, it results in panic()
+endif
 config ARCH_HAVE_HIPRI_INTERRUPT
 	bool
 	default n
--- a/os/arch/arm/src/armv7-m/up_doirq.c
+++ b/os/arch/arm/src/armv7-m/up_doirq.c
@@ -92,6 +92,7 @@ uint32_t *up_doirq(int irq, uint32_t *regs)
 {
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
 	irqstate_t flags;
+	uint32_t stack_remain;
 #endif
 	board_led_on(LED_INIRQ);
 #ifdef CONFIG_SUPPRESS_INTERRUPTS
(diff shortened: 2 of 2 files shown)


seinfra:
Target : [03bd9de87503d2df7ae1db596438037634dfeeb1] - Code Rule Check OK.

seinfra:
Target : [03bd9de87503d2df7ae1db596438037634dfeeb1] - Code Rule Check (C++) OK.

seinfra:
Target : [a1b3bad23fe1fa1e8b4a1d6e9a79170e594e0846] - Code Rule Check OK.

seinfra:
Target : [a1b3bad23fe1fa1e8b4a1d6e9a79170e594e0846] - Code Rule Check (C++) OK.

os/arch/Kconfig
```diff
@@ -584,8 +584,15 @@ config ARCH_NESTED_INTERRUPT
 	---help---
 		Enables the nested interrupt support. If defined, higher priority interrupt
 		can pre-empt the lower priority interrupt.
-endif
 
+config ARCH_NESTED_INTERRUPT_STACKCHECK
+	bool "Nested interrupt stack check support"
+	depends on ARCH_NESTED_INTERRUPT
```
yashwanth686007: Add dependency on CONFIG_STACK_COLORATION as well.

drashti304: done

seinfra:
Target : [21e8b112c806d0334a460670a14fdcd910440325] - Code Rule Check (C++) OK.

seinfra:
Target : [21e8b112c806d0334a460670a14fdcd910440325] - Code Rule Check OK.

os/arch/arm/src/armv7-m/up_doirq.c
```diff
@@ -92,13 +92,25 @@ uint32_t *up_doirq(int irq, uint32_t *regs)
...
 	PANIC();
 #else
 
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
+
+#ifdef CONFIG_ARCH_NESTED_INTERRUPT_STACKCHECK
+
+	/* check remaining stack available, assert if stack overflows */
+
+	stack_remain = (CONFIG_ARCH_INTERRUPTSTACK & ~3) - up_check_intstack();
```
sijupunnnoose: Should we check like below instead?
        uint32_t sp = up_getsp();
	uint32_t istackbase = (uint32_t)&g_intstackbase;
	uint32_t istacksize = (CONFIG_ARCH_INTERRUPTSTACK & ~3);
	if (sp <= istackbase && sp > istackbase - istacksize) {
		PANIC();
	}
stack_remain is unsigned int, if stack is already overflowed above statement would result in a positive number which would be higher than 8. I feel better check the stack is in the range.

drashti304: Hi Siju,

up_check_intstack() is using coloration method to check how much stack is being utilized. 
so if the stack is over flowed then this function will return the full size of stack.
so it will result for ex. in existing implementation . 2048 -2048 =0 

I have tested with remian_size =0 and it is working fine.

So i think we can use this method.

Please let me know your opinion on this.

sijupunnnoose: Ok

sunghan-chang: You can call `up_check_intstack_remain` instead of `(CONFIG_ARCH_INTERRUPTSTACK & ~3) - up_check_intstack()`.

sunghan-chang: And how much more time does it take against without this? I mean I want to know the performance degraded.

drashti304: yeah, but when I checked the stack_remain value, up_check_intstack_remain was taking extra 8 bytes , so I though to got with stack_remain = (CONFIG_ARCH_INTERRUPTSTACK & ~3) - up_check_intstack(); instead of up_check_intstack_remain().

like if I use, up_check_intstack_remain(), this function --> stack_remain is 64
and if I use (CONFIG_ARCH_INTERRUPTSTACK & ~3) - up_check_intstack(); -- > stack_remain is 72.

So I though to use later.

drashti304: > And how much more time does it take against without this? I mean I want to know the performance degraded.

I did not check this as I though we will enable this feature in case of any system failure to see what caused the issue, by default it is disable only.

I will update the performance degradation after measuring soon.

PR #3234, opened 2019-05-28
os/drivers: Modify PWM driver to support lower ioctl

SeonghoByeon:
- PWMIOC_SETCHARACTERISTICS is now provided by calling lower's ioctl
- Modify imxrt_pwm to be properly initialized

### DIFF
--- a/os/arch/arm/src/imxrt/imxrt_pwm.c
+++ b/os/arch/arm/src/imxrt/imxrt_pwm.c
@@ -238,6 +238,9 @@ status_t imxrt_pwm_init(PWM_Type *base, pwm_submodule_t subModule, const pwm_con
 		base->SM[subModule].CTRL2 |= PWM_CTRL2_FORCE(1U);
 	}
 
+	base->SM[subModule].DISMAP[0] = 0;
+	base->SM[subModule].DISMAP[1] = 0;
+
 	return kStatus_Success;
 }
 
@@ -933,7 +936,7 @@ static int imxrt_pwm_shutdown(FAR struct pwm_lowerhalf_s *dev)
 	imxrt_pwm_stop(dev);
 
 	/* Then put the GPIO pins back to the default state */
-	imxrt_pwm_deinit(priv->base,kPWM_Control_Module_0 | kPWM_Control_Module_1 | kPWM_Control_Module_2);
+	imxrt_pwm_deinit(priv->base, kPWM_Control_Module_0 | kPWM_Control_Module_1 | kPWM_Control_Module_2);
 
 	return OK;
 }
@@ -956,7 +959,23 @@ static int imxrt_pwm_shutdown(FAR struct pwm_lowerhalf_s *dev)
 static int imxrt_pwm_ioctl(FAR struct pwm_lowerhalf_s *dev, int cmd,
 						 unsigned long arg)
 {
-	return -ENOTTY;
+	int ret = -ENOTTY;
+	switch (cmd) {
+	case PWMIOC_SETCHARACTERISTICS: {
+		FAR const struct pwm_info_s *info = (FAR const struct pwm_info_s *)((uintptr_t)arg);
+		DEBUGASSERT(info != NULL);
+		//todo
+	}
+	break;
+	case PWMIOC_IDLE:
+		//todo
+	break;
+	default:
+		//todo
+	break;
+	}
+
+	return ret;
 }
 
 /****************************************************************************
@@ -979,7 +998,7 @@ static void imxrt_pwm_reset(FAR struct pwm_lowerhalf_s *dev, int timer)
 {
 	FAR struct imxrt_pwmtimer_s *priv = (FAR struct imxrt_pwmtimer_s *)dev;
 
-	imxrt_pwm_stop(dev);
+	imxrt_pwm_stop(priv);
 }
 
 /****************************************************************************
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [d4990430d6d85944e26acc058a6cad68183ac046] - Code Rule Check Result:
os/arch/arm/src/imxrt/imxrt_pwm.c:939: ERROR: [SPC_M_OPR] space required after that ',' (ctx:VxV)
os/arch/arm/src/imxrt/imxrt_pwm.c:1150: ERROR: [IDT_M_SCH] switch and case should be at the same indent

seinfra:
Target : [d4990430d6d85944e26acc058a6cad68183ac046] - Code Rule Check (C++) OK.

os/drivers/pwm.c
```diff
@@ -445,27 +445,40 @@ static int pwm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			ret = lower->ops->stop(lower);
+			if (ret < 0) {
+				return ret;
+			}
+			upper->started = false;
 #ifdef CONFIG_PWM_PULSECOUNT
-			ret = lower->ops->start(lower, &upper->info, upper);
-#else
-			ret = lower->ops->start(lower, &upper->info);
+			if (upper->waiting) {
```
Taejun-Kwon: As I know basic flow is below,
open -> ioctl(PWMIOC_SETCHARACTERISTICS) -> ioctl(PWMIOC_START)..
pwm_start is being called by ioctl(PWMIOC_START), and please notice that this flag is only for upper's state..

os/drivers/pwm.c
```diff
@@ -445,27 +445,40 @@ static int pwm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+				upper->waiting = FALSE;
+			}
 #endif
+			DEBUGASSERT(lower->ops->ioctl != NULL);
+			ret = lower->ops->ioctl(lower, cmd, info);
+			if (ret < 0) {
+				return ret;
+			}
+			/* Restart the pulse train */
+			ret = pwm_start(upper, filep->f_oflags);
```
Taejun-Kwon: Also, when you change upper, then you should consider all lowers(s5j, stm32,...and so on..)

os/arch/arm/src/imxrt/imxrt_pwm.c
```diff
@@ -1350,7 +1369,8 @@ FAR struct pwm_lowerhalf_s *imxrt_pwminitialize(int timer)
 		return NULL;
 	}
 
-	imxrt_pwm_reset(lower, timer);
+	imxrt_pwm_setup(lower);
```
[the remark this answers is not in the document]
SeonghoByeon: You are right. imxrt_pwm_setup and imxrt_pwm_start below are mapped to lower ops, therefore an application using pwm should 1) open the driver to setup, and 2) call driver's ioctl with PWMIOC_START. Thanks for the comment.

os/drivers/pwm.c
```diff
@@ -445,27 +445,40 @@ static int pwm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
-
-		/* Save the pulse train characteristics */
-
-		memcpy(&upper->info, info, sizeof(struct pwm_info_s));
-
 		/* If PWM is already running, then re-start it with the new characteristics */
-
 		if (upper->started) {
+			DEBUGASSERT(lower->ops->stop != NULL);
+			ret = lower->ops->stop(lower);
```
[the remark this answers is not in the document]
SeonghoByeon: PWMIOC_SETCHARACTERISTICS sets the characteristics of the next pulsed output, but it might not change the characteristics on the fly. So for safety, it firstly stops the current pulse, and then, after changing the characteristics such as duration and period, the pulse will be restarted below.

os/drivers/pwm.c
SeonghoByeon: The upper->waiting is used for CONFIG_PWM_PULSECOUNT mode. It means that if the number of pulse does not reach to the specified value (that is, the upper still waits the pulse hit the specified counter value,) it can set the characteristics by stopping the original, but restart the next pulse with new parameters. Thanks for the comment.

SeonghoByeon: One thing to note is that
PWM characteristics can be changed during the start phase, that is;
open -> set char -> start -> set char (stop->set char->restart) ->...

os/drivers/pwm.c
SeonghoByeon: The previous implementation was wrong, in my opinion. Calling start ops here might conflict with CONFIG_PWM_PULSECOUNT, so it should call pwm_start which is defined by following the configuration (CONFIG_PWM_PULSECOUNT or not). Thanks for the comment!

seinfra:
Target : [28e70a88b21ab98da49706439d8d681f36f5a185] - Code Rule Check (C++) OK.

seinfra:
Target : [28e70a88b21ab98da49706439d8d681f36f5a185] - Code Rule Check OK.

seinfra:
Target : [37511ec6b79d71aa4b8116d69d2e55111413144d] - Code Rule Check (C++) OK.

seinfra:
Target : [37511ec6b79d71aa4b8116d69d2e55111413144d] - Code Rule Check OK.

os/drivers/pwm.c
Taejun-Kwon: I mean it is only for upper's state, and it will clear when it expired or stopped.
So I think lower should handle callback mechanism to clear this.

SeonghoByeon: Got your point. As you said, when CONFIG_PWM_PULSECOUNT is set, pwm_start() will waits until pwm_expired() callback is called. That is the right process. However, If other app or thread wants to make PWM stop, it can call ioctl to stop it, therefore upper can directly control here.

PR #3244, opened 2019-05-29
add tc_protected in imxrt1020-evk and fix build warning

sunghan-chang:
1. configs/imxrt1020-evk: add tc_protected configurations set
2. kernel/task: fix wrong including messaging header


seinfra:
Target : [f9bc5089acdefb61dd2cdda5827179ef9f5068d8] - Code Rule Check (C++) OK.

seinfra:
Target : [f9bc5089acdefb61dd2cdda5827179ef9f5068d8] - Code Rule Check OK.

build/configs/imxrt1020-evk/tc_protected/defconfig
```diff
@@ -0,0 +1,1081 @@
...
+CONFIG_APPS_DIR="../apps"
+CONFIG_FRAMEWORK_DIR="../framework"
+CONFIG_TOOLS_DIR="../tools"
+# CONFIG_BUILD_FLAT is not set
+CONFIG_BUILD_PROTECTED=y
+# CONFIG_APP_BINARY_SEPARATION is not set
+CONFIG_BUILD_2PASS=y
+CONFIG_PASS1_TARGET="all"
+CONFIG_PASS1_OBJECT=""
+CONFIG_TINYARA_USERSPACE=0x60040000
```
btheosam: This should NOT be fixed like this. It depends on CONFIG_FLASH_PART_SIZE="256,256,4096. Otherwise user binary will be flashed based on CONFIG_FLASH_PART_SIZE="256  and this fixed value will be  used in Code. Lot of trouble. Either fix CONFIG_FLASH_PART_SIZE="256,  (make it NOT configurable) and use fixed value or calculate the start address from partititon and use it.

sunghan-chang: I agree. But not in this PR.

btheosam: Then exclude this file from this commit. Fix only the build warning. IF tc for protected is to  be added, then it has to be fixed.

sunghan-chang: No I don't think so. This is adding tc_protected using current implementation. If we remove that config, it does not boot up. I agree that we should modify to be set automatically from partition information. But this PR has totally different purpose.

build/configs/imxrt1020-evk/tc_protected/defconfig
```diff
@@ -0,0 +1,1081 @@
...
+# CONFIG_DDR is not set
+# CONFIG_ARCH_HAVE_SDRAM is not set
+
+#
+# Board Selection
+#
+CONFIG_ARCH_BOARD_IMXRT1050_EVK=y
+# CONFIG_ARCH_BOARD_ARTIK05X_FAMILY is not set
+# CONFIG_ARCH_BOARD_ESP32_FAMILY is not set
+CONFIG_ARCH_BOARD="imxrt1050-evk"
```
yashwanth686007: imxrt1050 -> imxrt1020

sunghan-chang: oops. right. I will update again

build/configs/imxrt1020-evk/tc_protected/defconfig
btheosam: I did n't mean to remove the config CONFIG_TINYARA_USERSPACE. I mean remove these set of files from this commit and push after making correct changes. Why do you want to commit with a known bug?

sunghan-chang: That is NOT a bug. It just causes inconvenient for user. It's right code of now. As I said before, to avoid mistake, we had better modify that logic. BUT not in this PR. This does NOT have that purpose. If you want to merge that with this PR, we should fix all of issue where you modify even if you don't modify. That's not good.

btheosam: This should be tracked and fixed later. For now, removed the hold.

seinfra:
Target : [ccb036a58c131af136887dddc2a2edc527391b08] - Code Rule Check (C++) OK.

seinfra:
Target : [ccb036a58c131af136887dddc2a2edc527391b08] - Code Rule Check OK.

PR #3265, opened 2019-05-30
os/arch/arm/src: Fix performance issue caused by MPU settings

kishore-sn:
Enabling the MPU and adding MPU regions is causing a degradation
of system performance. Each additional MPU region setting results
in a reduction of performance. Hence, we need to make sure that
we are using the minimum required MPU regions.

This commit contains following changes:
- Remove Kernel MPU regions, since Kernel has access to complete address space
- Enable USER_CODE and USER_DATA region only if user app is enabled
- Fix issues that happen during task start and context switch

### DIFF
--- a/os/arch/arm/src/armv7-m/mpu.h
+++ b/os/arch/arm/src/armv7-m/mpu.h
@@ -663,7 +663,11 @@ static inline void mpu_peripheral(uint32_t region, uintptr_t base, size_t size)
 #if defined(CONFIG_APP_BINARY_SEPARATION)
 static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 {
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [ea77078597be45799e411c58c996374116fe9783] - Code Rule Check (C++) OK.

seinfra:
Target : [ea77078597be45799e411c58c996374116fe9783] - Code Rule Check OK.

seinfra:
Target : [eba7176a2a4169264db07f6b2590e2c1a4f02c54] - Code Rule Check OK.

seinfra:
Target : [eba7176a2a4169264db07f6b2590e2c1a4f02c54] - Code Rule Check (C++) OK.

PR #3273, opened 2019-06-01
Modify NOR Flash function for iMXRT1020

HarrisonChoi750405:


seinfra:
Target : [c0c1552a93c93c1abcb041d54f93d3bd35778936] - Code Rule Check (C++) OK.

seinfra:
Target : [c0c1552a93c93c1abcb041d54f93d3bd35778936] - Code Rule Check OK.

os/arch/arm/src/imxrt/imxrt_mtd_flash.c
```diff
@@ -178,6 +178,12 @@ static inline int imxrt_readid(struct imxrt_dev_s *priv)
 	}
 	fvdbg("Vendor ID: 0x%x\r\n", vendorID);
 
+	/* Enter quad mode. */
+	status = imxrt_flexspi_nor_enable_quad_mode(IMXRT_FLEXSPI);
+	if (status != kStatus_Success) {
+		fdbg("flexspi_nor_enable_quad_mode error!!");
```
chanijjani: May need to add "return status;".

os/arch/arm/src/imxrt/imxrt_flash.h
```diff
@@ -136,6 +136,7 @@ extern "C" {
 #define EXTERN extern
 #endif
 
+#if !defined(CONFIG_ARCH_CHIP_FAMILY_IMXRT102x)
```
chanijjani: Why did you add this..? This function seems to be used only for imxrt1020 board...

os/arch/arm/src/imxrt/imxrt_norflash.c
```diff
@@ -725,6 +702,7 @@ static inline void flexspi_clock_init(void)
 /************************************************************************************
  * Public Functions
  ************************************************************************************/
+#if !defined(CONFIG_ARCH_CHIP_FAMILY_IMXRT102x)
```
chanijjani: Why did you add this..? This function seems to be used only for imxrt1020 board...

os/arch/arm/src/imxrt/imxrt_norflash.c
```diff
@@ -839,6 +818,8 @@ status_t imxrt_flexspi_nor_flash_page_program(FLEXSPI_Type *base, uint32_t addre
 	}
 
 	status = imxrt_flexspi_nor_wait_bus_busy(base);
+
+	arch_flush_dcache(address + IMXRT_FLASH_BASE, address + IMXRT_FLASH_BASE + IMXRT_FLASH_PAGE_SIZE);
```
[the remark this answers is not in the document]
Taejun-Kwon: Generally, Writing sequence follow below;
1. irq save
2. wait until previous writing is ended.(previous will change registery to disable write end of writing)
3. enable writing
4. write
5. disable write
6. irq restore.

My question is that, imxrt_flexspi_nor_wait_bus_busy only checks 'busy status', is it enough in multi thread environment?

os/arch/arm/src/imxrt/imxrt_flash.h
Taejun-Kwon: If all of 105x allows only hyperflash, and 102x allows only nor, then this is acceptable, but if there's chance that use another flash on 102x or 105x, then you should make config for flash itself.
Assume that you made configs for these flashes as a below.
CONFIG_XXXX_HYPER_FLASH_YYYY... / CONFIG_XXXX_NOR_FLASH_YYYYY...

Then you can enable it in each of defconfig, also use should enable it if they want make their own .config without defconfig

chanijjani:
@Taejun-Kwon All your valuable comments will be considered and implemented in the next PR after those comments are summarized and sent to Harrison and our members as an e-mail.

PR #3283, opened 2019-06-03
build/configs: Fix reset issue at imxrt flash init

sangam-swami:
Modify linker scripts to fix flexspi flash configuration issue
In case of XIP, flexispi and hyperflash objects run on top of flash,
if we copy these in flash itself then unexpected behavior results.
So these are copied to data section to execute the flash init and config
from outside.


seinfra:
Target : [3acc97f1221cde28e3454aa8ee3951b62ea9f606] - Code Rule Check OK.

seinfra:
Target : [3acc97f1221cde28e3454aa8ee3951b62ea9f606] - Code Rule Check (C++) OK.

PR #3296, opened 2019-06-04
os/mm/umm_heap: Fix memory leak in umm_free

kishore-sn:
umm_free can fail under the below scenario:
- Two tasks are running
- Task 1 exits. Task 2 is now active
- The task 1 clean up code calls kumm_free to free its stack
- umm_free() tries to get heap using mm_get_heapindex() API
- Since, task 2 is now active, this API returns -1

The issue is fixed by replacing mm_get_heapindex() with mm_get_heap()
which has the logic to search all the app heaps to find a match for
the given address.

### DIFF
--- a/os/mm/umm_heap/umm_free.c
+++ b/os/mm/umm_heap/umm_free.c
@@ -82,10 +82,13 @@
 
 void free(FAR void *mem)
 {
-	int heap_idx;
-	heap_idx = mm_get_heapindex(mem);
-	if (heap_idx != INVALID_HEAP_IDX) {
-		mm_free(&USR_HEAP[heap_idx], mem);
+	struct mm_heap_s *heap;
+	heap = mm_get_heap(mem);
+	if (heap) {
+		mm_free(heap, mem);
+		return;
 	}
+
+	mdbg("Failed to free address 0x%x\n", mem);
 }
 


seinfra:
Target : [ee58831dbe7cccf99d3e057c3f25b46f9b2b5eeb] - Code Rule Check (C++) OK.

seinfra:
Target : [ee58831dbe7cccf99d3e057c3f25b46f9b2b5eeb] - Code Rule Check OK.

os/mm/umm_heap/umm_free.c
```diff
@@ -82,10 +82,12 @@
...
 void free(FAR void *mem)
 {
-	int heap_idx;
-	heap_idx = mm_get_heapindex(mem);
-	if (heap_idx != INVALID_HEAP_IDX) {
-		mm_free(&USR_HEAP[heap_idx], mem);
+	struct mm_heap_s *heap;
+	heap = mm_get_heap(mem);
+	if (heap) {
+		return mm_free(heap, mem);
```
sunghan-chang: `free` and `mm_free` have void as return.
```
return mm_free( );
```
Does it causes any warning?

jeongchanKim: this function's return type is void, so you'd better change like below:
```
if (heap) {
    mm_free(xx,x);
    return;
}
```

kishore-sn: It does not cause warning. But i changed it.

seinfra:
Target : [31d532115cb994f30ed2eb6623f8b74da1a4ebdd] - Code Rule Check OK.

seinfra:
Target : [31d532115cb994f30ed2eb6623f8b74da1a4ebdd] - Code Rule Check (C++) OK.

os/mm/umm_heap/umm_free.c
kishore-sn: fixed.

PR #3317, opened 2019-06-05
external/dhcp: Running LwIP DHCP over ioctl

SeonghoByeon:
- Modify DHCP application to use ioctl when the application uses LwIP's DHCPC/D

### DIFF
--- a/os/net/netdev/netdev_dhcpc.c
+++ b/os/net/netdev/netdev_dhcpc.c
@@ -1,0 +1,285 @@
+/****************************************************************************
+ *
+ * Copyright 2019 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * net/netdev/netdev_dhcpc.c
+ *
+ *   Copyright (C) 2007, 2009, 2011-2012 Gregory Nutt. All rights reserved.
+ *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *
+ * Based heavily on portions of uIP:
+ *
+ *   Author: Adam Dunkels <adam@dunkels.com>
+ *   Copyright (c) 2005, Swedish Institute of Computer Science
+ *   All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <sys/types.h>
+
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <time.h>
+#include <errno.h>
+#include <debug.h>
+#include <net/if.h>
+
+#include <protocols/dhcpc.h>
+#include <netutils/netlib.h>
+#include <net/lwip/netif.h>
+#include <net/lwip/netifapi.h>
+#include "netdev/netdev.h"
+
+/****************************************************************************
+ * Definitions
+ ****************************************************************************/
+#define DHCPC_SET_IP4ADDR(intf, ip, netmask, gateway)	        \
+	do {							\
+		int res = -1;					\
+		res = _netlib_set_ipv4addr(intf, &ip);		\
+		if (res == -1) {				\
+			nvdbg("[DHCPC] set ipv4 addr error\n");	\
+		}						\
+		res = _netlib_set_ipv4netmask(intf, &netmask);	\
+		if (res == -1) {				\
+			nvdbg("[DHCPC] set netmask addr error\n");	\
+		}						\
+		res = _netlib_set_dripv4addr(intf, &gateway);	\
+		if (res == -1) {				\
+			nvdbg("[DHCPC] set route addr error\n");	\
+		}						\
+	} while (0)
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+static void _ioctl_setipv4addr(FAR in_addr_t *outaddr, FAR const struct sockaddr *inaddr)
+{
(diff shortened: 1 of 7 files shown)


seinfra:
Target : [b399ff00600148fad8d39184113a575e10a379cd] - Code Rule Check (C++) OK.

seinfra:
Target : [b399ff00600148fad8d39184113a575e10a379cd] - Code Rule Check OK.

seinfra:
Target : [44a248b9be71af68fc3a754591ee9a3819335f68] - Code Rule Check (C++) OK.

seinfra:
Target : [44a248b9be71af68fc3a754591ee9a3819335f68] - Code Rule Check OK.

os/net/netdev/netdev_ioctl.c
```diff
@@ -1149,6 +1149,47 @@ static int free_addrinfo(struct addrinfo *ai)
...
+	mqd_t md = mq_open(DHCPD_MQ_NAME, O_RDWR | O_CREAT, 0666, &attr);
+	if (!md) {
+		printf("mq open fail%s %d\n", __FUNCTION__, errno);
+		mq_close(md);
+		return;
+	}
+
+	char *msg = "dhcpd_join";
+	int mq_ret = mq_send(md, msg, DHCPD_MQ_LEN, 100);
+	printf("[shbyeon] send mq %s\n", msg);
```
Taejun-Kwon: I already know who you are. also it is kernel, do not use printf.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -92,15 +87,85 @@
...
+		char msg[DHCPD_MQ_LEN];
+		memset(msg, 0, DHCPD_MQ_LEN);
+		int prio = 0;
+		int nbytes = mq_receive(md, msg, DHCPD_MQ_LEN, &prio);
+		if (nbytes <= 0) {
+			if (errno != EAGAIN) {
+				ndbg("mq receive none (errno %d)\n", errno);
+				break;
+			}
+		} else {
```
Taejun-Kwon: This is not good, why don't you define message and check returned message?
I think Callback mechanism is much more better than your current logic.....

external/dhcpd/dhcpd_lwip.c
```diff
@@ -109,66 +174,86 @@
...
 	}
-	if (cur_netif->dhcps_pcb == NULL) {
-		ndbg("[DHCPS] stop dhcpd fail: no pcb\n");
-		return -1;
+
+	memset(&req, 0, sizeof(req));
+	req.type = DHCPDSTOP;
+	req.host_name = intf;
+
+	ret = ioctl(sockfd, SIOCLWIP, (unsigned long)&req);
```
Taejun-Kwon: I gave to you review already but how application layer knows ip layer's feature?? Of course file name including 'lwip'
But it is ioctl's command definition, this should be handled above of ip layer, I think
You can define SIOC_DHCPSTART / SIOC_DHCPSTOP and so on in ioctl.h
then socket should check current ip type and call functions properly.

os/net/netdev/netdev_dhcpc.c
```diff
@@ -0,0 +1,285 @@
...
+	while (netifapi_dhcp_address_valid(cur_netif) != 0) {
+		usleep(100000);
+		timeleft -= 100;
+		if (timeleft <= 0) {
+			ndbg("[DHCPC] dhcpc_start timeout\n");
+			netifapi_dhcp_stop(cur_netif);
+			return -1;
+		}
+	}
+	int ret = _netlib_get_ipv4addr(intf, &ip_check);
```
Taejun-Kwon: you can remove ret here.

os/net/netdev/netdev_dhcpc.c
```diff
@@ -0,0 +1,285 @@
...
+	DHCPC_SET_IP4ADDR(intf, in, in, in);
+	struct netif *cur_netif;
+	cur_netif = netif_find(intf);
+	if (cur_netif == NULL) {
+		ndbg("[DHCPC] No network interface for dhcpc\n");
+		return;
+	}
+	netifapi_dhcp_stop(cur_netif);
+	nvdbg("[DHCPC] dhcpc_stop -release IP address (lwip)\n");
+	return;
```
Taejun-Kwon: useless return here.

os/net/netdev/netdev_dhcpd.c
```diff
@@ -0,0 +1,165 @@
...
+	if (dhcp_join_cb) {
+		if (dhcps_register_cb(dhcp_join_cb) != ERR_OK) {
+			ndbg("[DHCPS] link callback fail\n");
+			return -1;
+		}
+	}
+
+	if (netifapi_dhcps_start(cur_netif) == ERR_OK) {
+		ndbg("[DHCPS] started successfully (LWIP)\n");
+		return OK;
```
Taejun-Kwon: You using OK here, then you can use ERROR instead of '-1' because it means you already include type.h

os/net/netdev/netdev_dhcpd.c
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return -1;
+}
+
+/****************************************************************************
+ * Name: dhcps_server_stop
+ ****************************************************************************/
+int netdev_dhcp_server_stop(char *intf)
```
Taejun-Kwon: same as above.

os/net/netdev/netdev_dhcpd.c
```diff
@@ -0,0 +1,165 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+/****************************************************************************
+ * Global Functions
+ ****************************************************************************/
+/****************************************************************************
+ * Name: dhcps_server_status
+ ****************************************************************************/
+int netdev_dhcp_server_status(char *intf)
```
Taejun-Kwon: You using like a below
```
		ret = netdev_dhcp_server_status((char *)in_arg->host_name);
		if (ret != 0) {
			printf("stop dhcpd fail\n");
		}
```
If server is opened, it returns '1' is it correct????

external/dhcpd/dhcpd_lwip.c
```diff
@@ -92,15 +87,85 @@
...
+	int sockfd = 0;
+	struct req_lwip_data req;
+	int ret = ERROR;
+	struct mq_attr attr;
+	attr.mq_maxmsg = DHCPD_MQ_MAX_LEN;
+	attr.mq_msgsize = DHCPD_MQ_LEN;
+	attr.mq_flags = 0;
+	attr.mq_curmsgs = 0;
+
+	mqd_t md = mq_open(DHCPD_MQ_NAME, O_RDWR | O_NONBLOCK |O_CREAT, 0666, &attr);
```
Taejun-Kwon: please add space between | and O_CREAT

external/dhcpd/dhcpd_lwip.c
```diff
@@ -92,15 +87,85 @@
...
+	req.type = DHCPDSTART;
+	req.host_name = data->intf;
+
+	ret = ioctl(sockfd, SIOCLWIP, (unsigned long)&req);
+	if (ret == ERROR) {
+		ndbg("ioctl() failed with errno: %d\n", errno);
+		goto go_out;
+	}
+
+	while (!g_dhcpd_term) {
```
Taejun-Kwon: You'd better classify code between related things for all the time (in this case handle socket first and then mq)

Also if it is possible, timed receive is more stable than mq_receive

external/dhcpd/dhcpd_lwip.c
```diff
@@ -109,66 +174,86 @@
...
-		return -1;
-	}
-	if (dhcp_join_cb) {
-		if (dhcps_register_cb(dhcp_join_cb) != ERR_OK) {
-			ndbg("[DHCPS] link callback fail\n");
-			return -1;
-		}
-	}
+	dhcp_join_data_s *data = (dhcp_join_data_s *)malloc(sizeof(dhcp_join_data_s));
+	data->intf = (char *)malloc(strlen(intf)+1);
```
Taejun-Kwon: numm clech is needed for these 2.

os/net/netdev/netdev_ioctl.c
```diff
@@ -1218,6 +1259,40 @@ int lwip_func_ioctl(int cmd, void *arg)
 		in_arg->req_res = ret;
 		ret = 0;
 		break;
+#endif
+#if defined(CONFIG_NET_LWIP_DHCP)
+	case DHCPCSTART:
```
Taejun-Kwon: Here is end node of ioctl, you should return one of error in errno.h
Then vfs will set it as errno...

external/dhcpd/dhcpd_lwip.c
SeonghoByeon: I totally agree with you. However, here there is a timing issue. After running LwIP DHCP through ioctl, there might be sending msg operation from LwIP DHCP, before mq is opened. Thanks for the comment.

external/dhcpd/dhcpd_lwip.c
SeonghoByeon: I agree with you, but we want to separate kernel and user space strictly, so callback might not be appropriate in this logic. I just added a logic to check whether the received msg is proper or not. Thanks for the comment.

external/dhcpd/dhcpd_lwip.c
SeonghoByeon: I guess it is 'null check'. Applied. Thanks for the comment.

external/dhcpd/dhcpd_lwip.c
SeonghoByeon: Thanks for the comment. This is an application, but it is kind of system utils. Based on the configuration, this application can use either 1) external dhcp or 2) LwIP (kernel) 's dhcp. For considering DHCP of LwIP, this structure is inevitable. As you know, LwIP has very exceptional architecture, so there is no other options, without changing the whole structure.

os/net/netdev/netdev_dhcpc.c
SeonghoByeon: Applied. Thanks for the comment.

os/net/netdev/netdev_dhcpc.c
SeonghoByeon: Applied. Thanks for the comment.

os/net/netdev/netdev_dhcpd.c
SeonghoByeon: This is wrong. It should return 0 when DHCP is opend. Thanks.

os/net/netdev/netdev_dhcpd.c
SeonghoByeon: Applied. Thanks for the comment.

os/net/netdev/netdev_ioctl.c
SeonghoByeon: Applied. Thanks ^^;;

os/net/netdev/netdev_dhcpd.c
SeonghoByeon: Applied. Thanks for the comment.

os/net/netdev/netdev_ioctl.c
SeonghoByeon: You are right. Applied. Thanks.

PR #3355, opened 2019-06-10
os/arch/arm/src: Fix performance issue caused by MPU settings

kishore-sn:
Enabling the MPU and adding MPU regions is causing a degradation
of system performance. Each additional MPU region setting results
in a reduction of performance. Hence, we need to make sure that
we are using the minimum required MPU regions.

This commit contains following changes:
- Remove Kernel MPU regions, since Kernel has access to complete address space
- Enable USER_CODE and USER_DATA region only if user app is enabled
- Fix issues that happen during task start and context switch

### DIFF
--- a/os/arch/arm/src/armv7-m/mpu.h
+++ b/os/arch/arm/src/armv7-m/mpu.h
@@ -663,7 +663,11 @@ static inline void mpu_peripheral(uint32_t region, uintptr_t base, size_t size)
 #if defined(CONFIG_APP_BINARY_SEPARATION)
 static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 {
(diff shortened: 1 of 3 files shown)


kishore-sn:
I am resubmitting this PR. I have removed the changes which were previously causing memfault in the Messaging fwk test.

seinfra:
Target : [69b507fefd2f0a74823bcb2da819d93b8b3fe7ca] - Code Rule Check (C++) OK.

seinfra:
Target : [69b507fefd2f0a74823bcb2da819d93b8b3fe7ca] - Code Rule Check OK.

sunghan-chang:
@kishore-sn You removed changes on up_svcall.c against first one. Does it cause abnormal behavior?

kishore-sn:
The changes in up_svcall.c is not required for now. But if we disable User binary (PREAPP_INIT) then it will cause abnormal behavior. I am investigating it and i will raise a separate PR for it.

PR #3371, opened 2019-06-12
Add userspace object to loadable application binaries

kishore-sn:
This PR contains 2 commits as below:

**commit 6597add05af1b83a8a4d78874bd9ec7294dfc292**

    The userspace object is used to share some information from the
    user space binary to the kernel. In case of loadable apps, we require
    task_startup(), pthread_startup() and signal_handler() api's to be
    shared from app to kernel. These api's are shared as pointers in the
    userspace object.

    If the userspace object is not included in the app binary, then the
    kernel would use the api's from the common user binary instead of the
    app binary. Then the call sequence would be kernel -> user binary -> app binary.
    This is wrong since we dont want calls to go through the common user binary.

    In this new implementation the userspace object will be placed at the beginning
    of the app binary at an offset of 4 bytes. The first 4 bytes contain a pointer to
    the app's heap object. This is followed by the userspace object and then the text
    section of the app binary.

**commit 4fdcb4e1e8476852c94e946ccc5902557b782a53**

    os/board: Move the userspace file to common folder

    The userspace.c file is not board specific. So, move it
    from the board specific folder into common folder.


seinfra:
Target : [6597add05af1b83a8a4d78874bd9ec7294dfc292] - Code Rule Check (C++) OK.

seinfra:
Target : [6597add05af1b83a8a4d78874bd9ec7294dfc292] - Code Rule Check OK.

seinfra:
Target : [7fe25488b49430c076efdd250070068e684c0cf5] - Code Rule Check OK.

seinfra:
Target : [7fe25488b49430c076efdd250070068e684c0cf5] - Code Rule Check (C++) OK.

os/arch/arm/src/armv7-m/up_svcall.c
```diff
@@ -350,7 +350,22 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
...
 
-		regs[REG_PC] = (uint32_t)USERSPACE->task_startup;
+#ifdef CONFIG_APP_BINARY_SEPARATION
+		/* While starting loadable apps, we cannot go through the
+		* USERSPACE->task_startup method. Instead we pick the PC value
+		* from the app's userspace object stored in its tcb.
+		*
+		* Here, we check if this task is a loadable app (non-zero ram_start)
+		*/
+		if (((struct tcb_s*)sched_self())->ram_start) {
```
jeongchanKim: Please add space between `tcb_s` and `*`.

kishore-sn: fixed

seinfra:
Target : [b5348024f1090e17765b5246762dd234e4137ae6] - Code Rule Check (C++) OK.

seinfra:
Target : [b5348024f1090e17765b5246762dd234e4137ae6] - Code Rule Check OK.

PR #3397, opened 2019-06-15
build/tools/esp32: include romfs when download ALL

zhouxinhe:
.dbuild.sh download ALL should also download romfs.
This behavior is consistent with artik's.


seinfra:
Target : [e57e2b52c1c162a1e0c4b80f1665d799f6adce50] - Code Rule Check OK.

seinfra:
Target : [e57e2b52c1c162a1e0c4b80f1665d799f6adce50] - Code Rule Check (C++) OK.

build/tools/esp32/esptool_py/Makefile
```diff
@@ -74,7 +74,7 @@ $(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
 
 ESPTOOL_ALL_FLASH_ARGS += $(ESPTOOL_BOOTLOADER_FLASH_ARGS)
 
-ALL:flash
+ALL:flash romfs
```
ChinaTizenRT: it looks fine. it keeps the function "make download ALL" same as other arch, like (artik).
and also if no romfs image for current configuration, it does not take effect.

PR #3425, opened 2019-06-20
framework/st_things: remove unnecessary logs.

Hyunjun85:
It has been getting bad effect to performance.
and now, the logs doesn't need.

### DIFF
--- a/framework/src/st_things/things_stack/framework/things_req_handler.c
+++ b/framework/src/st_things/things_stack/framework/things_req_handler.c
@@ -197,17 +197,12 @@ static OCEntityHandlerResult convert_ap_infor(things_resource_s *target_resource
 	} else {
 		rep = target_resource->rep;
 	}
+
+	THINGS_LOG_D(TAG, "Create response for accesspoints resource, WiFi list count is: %d", list_cnt);
 	things_representation_s **child_rep = (things_representation_s **)things_malloc(list_cnt * sizeof(things_representation_s *));
 	access_point_info_s *wifi_scan_iter = p_list;
 	int iter;
 	for (iter = 0; iter < list_cnt; ++iter) {
-#ifdef CONFIG_DEBUG_ST_THINGS_DEBUG
-		THINGS_LOG_V(TAG, "e_ssid : %s", wifi_scan_iter->e_ssid);
-		THINGS_LOG_V(TAG, "signal_level : %s", wifi_scan_iter->signal_level);
-		THINGS_LOG_V(TAG, "bss_id : %s", wifi_scan_iter->bss_id);
-		THINGS_LOG_V(TAG, "sec_type : %s", wifi_scan_iter->sec_type);
-		THINGS_LOG_V(TAG, "enc_type : %s", wifi_scan_iter->enc_type);
-#endif               // CONFIG_DEBUG_ST_THINGS_DEBUG
 		child_rep[iter] = things_create_representation_inst(NULL);
 		child_rep[iter]->things_set_value(child_rep[iter], SEC_ATTRIBUTE_AP_MACADDR, wifi_scan_iter->bss_id);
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [12f8ddb45fa207b881b92e2c1575a0ceb35cc9ae] - Code Rule Check (C++) OK.

seinfra:
Target : [12f8ddb45fa207b881b92e2c1575a0ceb35cc9ae] - Code Rule Check OK.

framework/src/st_things/things_stack/framework/things_req_handler.c
```diff
@@ -197,17 +197,12 @@ static OCEntityHandlerResult convert_ap_infor(things_resource_s *target_resource
...
 	}
+
+	THINGS_LOG_D(TAG, "Create response for accesspoints resource, WiFi list count is: %d", list_cnt);
 	things_representation_s **child_rep = (things_representation_s **)things_malloc(list_cnt * sizeof(things_representation_s *));
 	access_point_info_s *wifi_scan_iter = p_list;
 	int iter;
 	for (iter = 0; iter < list_cnt; ++iter) {
-#ifdef CONFIG_DEBUG_ST_THINGS_DEBUG
-		THINGS_LOG_V(TAG, "e_ssid : %s", wifi_scan_iter->e_ssid);
-		THINGS_LOG_V(TAG, "signal_level : %s", wifi_scan_iter->signal_level);
```
JeongJunSik: Would not it be better to put the rssi value at the debug level?

framework/src/st_things/things_stack/framework/things_req_handler.c
```diff
@@ -197,17 +197,12 @@ static OCEntityHandlerResult convert_ap_infor(things_resource_s *target_resource
 	} else {
 		rep = target_resource->rep;
 	}
+
+	THINGS_LOG_D(TAG, "Create response for accesspoints resource, WiFi list count is: %d", list_cnt);
```
JeongJunSik: i think this log is not need.
because in file of things_req_handler.c[Line 236], already printed `THINGS_LOG_V(TAG, "Get Access points list!!! %d ", list_cnt);` message.
Please check duplicate message.

framework/src/st_things/things_stack/framework/things_req_handler.c
Hyunjun85: This log doesn't need. It spend some time to pring log.

framework/src/st_things/things_stack/framework/things_req_handler.c
Hyunjun85: It is different message.
It(201) will be printed  when received get request about accesspoints resource. and 236 will print when ap list scanning is done.  both messges need for debugging.

JeongJunSik: I don't think so.
this logic is part of `convert_ap_infor()` function.
`convert_ap_infor()` function is called by `get_ap_scan_result()[line 237]` with 3rd parameter only.
So, it's always the same result. 
Please check it.

framework/src/st_things/things_stack/framework/things_req_handler.c
JeongJunSik: This is the only logic that runs in DEBUG mode.
In my opinion, debugging mode, it does not seem to have much relation to performance.
I think need more information for debugging (rssi signal) than the time spent in print

Hyunjun85: When enable ST_THIGNS_DEBUG option. 
Those log printing logic will  over a  seconds only for two printing logics about scanning result and creating response. 

Actually in last performance check, I know something that If DEBUG was enabled  the easy-setup time will increase more 20s. And It will increase spending time if there is ap device more. 

And now we can get ap list and infomations during do easy-setup. ST APP(since 1.7.33-26) will support this function and user must select ap to use.

PR #3429, opened 2019-06-20
framework/security: fix security bug

pillip8282:
if it doesn't allocate memory in security layer it'll cause crash
becuase of dangling pointer. furthermore sl_init function doesn't
deallocate handle when it fails to call ioctl or to get error message
from SE driver. handle should be freed both cases

### DIFF
--- a/framework/src/seclink/seclink.c
+++ b/framework/src/seclink/seclink.c
@@ -55,16 +55,9 @@
 
 #ifdef LINUX
 extern int sl_post_msg(int fd, int cmd, unsigned long arg);
+#define ioctl sl_post_msg
+#endif
 
-#define SL_CALL(hnd, code, param)										\
-	do {																\
-		int i_res = sl_post_msg(hnd->fd, code, (unsigned long)((uintptr_t)&param)); \
-		if (i_res < 0) {												\
-			SL_ERR(i_res);												\
-			return SECLINK_ERROR;										\
-		}																\
-	} while (0)
-#else
 #define SL_CALL(hnd, code, param)										\
 	do {																\
 		int i_res = ioctl(hnd->fd, code, (unsigned long)((uintptr_t)&param)); \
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [4155e1f563849e852be996792b15b23f4cb8ccf9] - Code Rule Check (C++) OK.

seinfra:
Target : [4155e1f563849e852be996792b15b23f4cb8ccf9] - Code Rule Check OK.

framework/src/seclink/seclink.c
```diff
@@ -55,26 +55,35 @@
...
+		int i_res = ioctl(hnd->fd, code, (unsigned long)((uintptr_t)&param)); \
 		if (i_res < 0) {												\
 			SL_ERR(i_res);												\
 			return SECLINK_ERROR;										\
 		}																\
 	} while (0)
-#else
-#define SL_CALL(hnd, code, param)										\
+
+#define SL_CALL2(hnd, code, param, error)								\
```
intom: How about changing to a meaningful name instead of SL_CALL2?

framework/src/seclink/seclink.c
```diff
@@ -124,7 +139,7 @@ int sl_init(sl_ctx *hnd)
 	handle->fd = fd;
 
 	struct seclink_req req = {.req_type.comm = NULL, 0};
-	SL_CALL(handle, SECLINKIOC_INIT, req);
+	SL_CALL2(handle, SECLINKIOC_INIT, req, SL_FREE_HANDLE(handle));
```
dr-venkman: As `SL_CALL2` can potentially return on error, shouldn't we also `close(fd)` inside the macro definition before returning? Please ignore if not necessary.

framework/src/seclink/seclink.c
```diff
@@ -55,26 +55,35 @@
...
 	} while (0)
-#else
-#define SL_CALL(hnd, code, param)										\
+
+#define SL_CALL2(hnd, code, param, error)								\
 	do {																\
 		int i_res = ioctl(hnd->fd, code, (unsigned long)((uintptr_t)&param)); \
 		if (i_res < 0) {												\
 			SL_ERR(i_res);												\
+			error;														\
```
dr-venkman: perhaps give it a meaningful name as well, such as `restore_on_error` or `error_cleanup`?

framework/src/security/security_utils.c
```diff
@@ -267,11 +267,8 @@ int secutils_convert_path_s2h(const char *path, uint32_t *slot)
 	if (!strncmp(path, SS_PATH, sizeof(SS_PATH) - 1)) {
 		*slot = atoi(&path[3]);
 		return 0;
-	} else if (!strncmp(path, FACTORY_PATH, sizeof(FACTORY_PATH) - 1)) {
```
dr-venkman: Is this change within the scope of the commit description?

pillip8282: it cause build fail

framework/src/seclink/seclink.c
pillip8282: I think that SL_CALL2 is enough

framework/src/seclink/seclink.c
pillip8282: I think error is understandable.

framework/src/seclink/seclink.c
sangwon03: There is a SL_CALL. Why we need SL_CALL2?

framework/src/seclink/seclink.c
Hyunjun85: SL_CALL2 is not good to understand.
There is "SL_CALL_NORET" on below, like this, Would you change like 'SL_CALL_RET'?

PR #3434, opened 2019-06-20
build/configs/imxrt: Modify CONFIG_RR_INTERVAL

yashwanth686007:
This patch modifies RR Interval to 100msec.


seinfra:
Target : [2a52bf36fd8925644237074124e40c8f37cab88e] - Code Rule Check OK.

seinfra:
Target : [2a52bf36fd8925644237074124e40c8f37cab88e] - Code Rule Check (C++) OK.

NilabhShant:
Don't you think same change will be applied for imxrt1020-evk/tc defconfig also.

yashwanth686007:
@NilabhShant , I cant find any such config on master branch.

NilabhShant:
> @NilabhShant , I cant find any such config on master branch.

Ok, got it. I thought you want to merged in protection.

build/configs/imxrt1020-evk/nxp_demo/defconfig
```diff
@@ -345,7 +345,7 @@ CONFIG_PREALLOC_TIMERS=8
 # Tasks and Scheduling
 #
 CONFIG_INIT_ENTRYPOINT=y
-CONFIG_RR_INTERVAL=0
+CONFIG_RR_INTERVAL=100
```
sijupunnnoose: You are enabling this to avoid failure with timer example, how about creating a dependency with CONFIG_EXAMPLES_TIMER??

sunghan-chang: @sijupunnnoose @yashwanth686007 Thank you for explaining the reason of this. I also thought this suggestion. But I want to know is it really needed? Without RR, isn't it working?

yashwanth686007: @sijupunnnoose @sunghan-chang , I believe its not only with timer example, its a generic round robin functionality that doesn't take place without this interval.
By the current default zero value, round robin scheduling will be disabled. If RR interval is set to zero, the task stays on the head of the ready-to-run list until it completes. It doesn't give any chance to other tasks of equal priority to run.

sijupunnnoose: @sunghan-chang, in master timer example fails with message "failed to set policy"

ret = pthread_attr_setschedpolicy(&attr, SCHED_RR);
if (ret != 0) {
	fprintf(stderr, "failed to set policy(%d)\n", ret);
	return -ret;
}

sunghan-chang: @yashwanth686007 We support FIFO as well as RR. Someone could want to work with FIFO. To work those defconfig, is RR necessary? I don't think so.
@sijupunnnoose  Current timer example makes it failed with FIFO. But it can't be the reason. I think we can use FIFO to run. Timer should be working with FIFO also. I think our timer example was implemented wrongly. We don't need to force to turn it on.

yashwanth686007: @sunghan-chang , yes, user can either of the scheduling policies. For this defconfig, RR is not necessary. Since we followed RR in remaining defconfigs, I have made the changes. If not required, i will revert.

sunghan-chang: @yashwanth686007 I think this is unnecessary. If you have the another reason, please let me know.

yashwanth686007: Okay.. I'll close the PR. Thanks.

PR #3486, opened 2019-07-01
framework/media: Correct http head parsing fault

zhouxinhe:
Abstract Content-Type value between `Content-Type: ` and `\r`.

### DIFF
--- a/framework/src/media/HttpInputDataSource.cpp
+++ b/framework/src/media/HttpInputDataSource.cpp
@@ -257,8 +257,11 @@ size_t HttpInputDataSource::HeaderCallback(char *data, size_t size, size_t nmemb
 	size_t totalsize = size * nmemb;
 	std::string header(data, totalsize);
 	medvdbg("%s\n", header.c_str());
-	if (header.find(TAG_CONTENT_TYPE) != std::string::npos) {
-		source->mContentType = header.substr(TAG_CONTENT_TYPE.length());
+	auto pos = header.find(TAG_CONTENT_TYPE);
+	if (pos != std::string::npos) {
+		pos = header.find_first_not_of(' ', pos + TAG_CONTENT_TYPE.length());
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [27d62100839c478c5187a11bbad6e4b0abaef57d] - Code Rule Check OK.

seinfra:
Target : [27d62100839c478c5187a11bbad6e4b0abaef57d] - Code Rule Check (C++) OK.

zhouxinhe:
@Taejun-Kwon @jsdosa 
Could you please help to review?

framework/src/media/HttpInputDataSource.cpp
```diff
@@ -257,8 +257,11 @@ size_t HttpInputDataSource::HeaderCallback(char *data, size_t size, size_t nmemb
 	size_t totalsize = size * nmemb;
 	std::string header(data, totalsize);
 	medvdbg("%s\n", header.c_str());
-	if (header.find(TAG_CONTENT_TYPE) != std::string::npos) {
-		source->mContentType = header.substr(TAG_CONTENT_TYPE.length());
+	auto pos = header.find(TAG_CONTENT_TYPE);
+	if (pos != std::string::npos) {
+		pos = header.find_first_not_of(' ', pos + TAG_CONTENT_TYPE.length());
+		auto end = header.find((char)0x0d, pos); // CR: 0x0d
```
xixidodo: Though '\r\n' shall always exist, but do we need to check 'end' for safe?

jsdosa: CURL already check the '\r\n'.If not, callback is not called

xixidodo: libcurl uses '\n' as delimiter while parsing HTTP header in Curl_http_readwrite_headers.
‘\r’ is not guaranteed here.

Anyway,  we can skip safety check for such really rare malicious HTTP server behavior.

PR #3500, opened 2019-07-02
Remove ARCH_HAVE_NESTED_INTERRUPT macro

drashti304:
Removing ARCH_HAVE_NESTED_INTERRUPT macro for nested interrupt
    as not all arm architecture supports it


seinfra:
Target : [24cf84be6d6ecd93fdc82ae741147350bf19a1be] - Code Rule Check (C++) OK.

seinfra:
Target : [24cf84be6d6ecd93fdc82ae741147350bf19a1be] - Code Rule Check OK.

build/configs/qemu/tc_64k/defconfig
```diff
@@ -246,7 +246,6 @@ CONFIG_BOARD_LOOPSPERMSEC=4531
 #
 CONFIG_ARCH_HAVE_INTERRUPTSTACK=y
 CONFIG_ARCH_INTERRUPTSTACK=0
-CONFIG_ARCH_HAVE_NESTED_INTERRUPT=y
```
sijupunnnoose: Should we remove this?? Did you test on qemu?

sunghan-chang:
@drashti304 As @sijupunnnoose said, I also think this should not be deleted. But this should be selected by architecture which supports nvic.
I mean
for example, if it is implemented on cortex-m7 only, cortex-m7 config only have `select ARCH_HAVE_NESTED_INTERRUPT`. Currently it is selected by arm, but you didn't implement the nvic on all of arm we support.

drashti304:
> @drashti304 As @sijupunnnoose said, I also think this should not be deleted. But this should be selected by architecture which supports nvic.
> I mean
> for example, if it is implemented on cortex-m7 only, cortex-m7 config only have `select ARCH_HAVE_NESTED_INTERRUPT`. Currently it is selected by arm, but you didn't implement the nvic on all of arm we support.

@sunghan-chang , I understand your point, but in TizenRT current implementation, armv7r architecture also includes CONFIG_ARMV7M_MPU, So we can not differentiate architectures by ARMV7M_MPU/ARMV7R_MPU.
That is why either we can let the current implementation as it is or we can remove this ARCH_HAVE_NESTED_INTERRUPT MACRO.

What are your opinion on this?

sunghan-chang:
@drashti304 
1. I am sorry but I don't know why you mention MPU in NVIC commit. Is there any relationship between MPU and NVIC? 
2. We could make new config for CONFIG_ARMV7R_MPU if it has a different logic.

seinfra:
Target : [12ccab31a8a76babffc4133e46dca59f0c77004d] - Code Rule Check (C++) OK.

seinfra:
Target : [12ccab31a8a76babffc4133e46dca59f0c77004d] - Code Rule Check OK.

sunghan-chang:
Could you push this into master first?

PR #3519, opened 2019-07-06
loadable_apps: tidyup

sunghan-chang:


seinfra:
Target : [db44a1ad083b07ff5daeddaa8132e81c85724027] - Code Rule Check OK.

seinfra:
Target : [db44a1ad083b07ff5daeddaa8132e81c85724027] - Code Rule Check (C++) OK.

seinfra:
Target : [0c62397ce0bfbc2822f0af39baee64a71252337a] - Code Rule Check (C++) OK.

seinfra:
Target : [0c62397ce0bfbc2822f0af39baee64a71252337a] - Code Rule Check OK.

seinfra:
Target : [c36ff8b42cbcd43f0cc5099efe5124c7669f0e6a] - Code Rule Check OK.

seinfra:
Target : [c36ff8b42cbcd43f0cc5099efe5124c7669f0e6a] - Code Rule Check (C++) OK.

seinfra:
Target : [0166dd8b4522ce1f53cb4d5d2429bd94ee071986] - Code Rule Check OK.

seinfra:
Target : [0166dd8b4522ce1f53cb4d5d2429bd94ee071986] - Code Rule Check (C++) OK.

seinfra:
Target : [9e3af3fd21873d67a07488650bb91f741afe88d6] - Code Rule Check OK.

seinfra:
Target : [9e3af3fd21873d67a07488650bb91f741afe88d6] - Code Rule Check (C++) OK.

loadable_apps/Makefile
```diff
@@ -53,53 +52,50 @@
...
-# Verify each program for undefined symbols
+verify: $(foreach BDIR, $(CONFIGURED), $(BDIR)_verify)
+else # !CONFIG_APP_BINARY_SEPARATION
 
-verify: $(foreach DIR, $(BUILD_SUBDIRS), $(DIR)_verify)
-else
-all:
+all: build
+.PHONY: all build clean distclean depend preconfig
+endif # CONFIG_APP_BINARY_SEPARATION
```
kishore-sn: Do we need to do anything in the #else part?
As I understand, the contents of the loadable_apps must be built only if app separation is enabled.

loadable_apps/loadable_sample/micomapp/Makefile
```diff
@@ -85,58 +72,35 @@ KERNEL_VER = 2.0
...
-BLOCK_SIZE = 0
-endif
+include $(TOPDIR)/../loadable_apps/loadable.mk
 
-APPDEFINE = ${shell $(TOPDIR)/tools/define.sh "$(CC)" __APP_BUILD__}
+###################################
+# Cases of Flat and Protected build
+# Merge into libapps.a
+###################################
+else
```
kishore-sn: I am wondering if this is correct? Do we need to merge these loadable apps into libapps?

loadable_apps/Makefile
sunghan-chang: @kishore-sn As I told on chat, I don't want to use this folder only for separated application. When app separation is enabled, it has to be separated binary. But with another build ways, it has to be a part of binary. It is more convenient for user. For this, we had better rename the folder name with another PR.

loadable_apps/loadable_sample/micomapp/Makefile
sunghan-chang: This is an else case of app separation. Without app separation, this is also a part of app so that I think this is right.

os/Kconfig
```diff
@@ -17,6 +17,10 @@ config LIBDIR
 	string
 	option env="LIBDIR"
 
+config LOADABLEDIR
```
btheosam: To be clear, this commit introduces CONFIG_LOADABLEDIR in addition to CONFIG_APPS_DIR and apps inside CONFIG_LOADABLEDIR dir alone will be loadable. Is it right?

loadable_apps/micomapp/Makefile
```diff
@@ -53,21 +53,12 @@
 -include $(TOPDIR)/.config
 -include $(TOPDIR)/Make.defs
 
-ifeq ($(CONFIG_BUILD_PROTECTED),y)
+ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
 LINKLIBS = $(patsubst $(LIBRARIES_DIR)/%, %, $(USERLIBS))
 LDLIBS = $(patsubst %.a, %, $(patsubst lib%,-l%,$(LINKLIBS)))
-endif
-
-OBJEXT ?= .o
```
btheosam: Generally makefile will have MACROS for everything. Coz changing would be easy later. For example to build on windows DELIM should a '\' backslash whereas for linux it is '/'. If this MACRO is not used. then later to change the build enviroment individual lines should be tracked and changed.

sunghan-chang: Right. But we already use '/' only. And we don't support Windows officially. When we support Windows, we had better modify all of Makefile with the same way. I don't want to use it different way in each file.

sunghan-chang:
@btheosam I resolved conflicts. Could you merge this?

seinfra:
Target : [5f8582eaced25b08a50ec8362ca653ddcac1bf23] - Code Rule Check (C++) OK.

seinfra:
Target : [5f8582eaced25b08a50ec8362ca653ddcac1bf23] - Code Rule Check OK.

loadable_apps/micomapp/Makefile
```diff
@@ -136,11 +135,11 @@ endif
...
-	$(call DELFILE, $(BIN))
-	$(call CLEAN)
-	exit 1
+ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
+	$(Q) if [ "`nm -u $(BIN) | wc -l`" != "0" ]; then \
+		echo "Undefined Symbols"; \
+		nm -u -l $(BIN); \
+		rm $(BIN); \
+		exit 1; \
+	fi
```
manoharahk: Instead of adding this check block in every loadable app Makefile, Is it not good to put this block into a shell script inside tools directory and invoke from every loadable app Makefile?

sunghan-chang: @manoharahk Hello. Yeah, as you told, it is moved into loadable.mk file. Please find the last commit.

loadable_apps/micomapp/Makefile
btheosam: @sunghan-chang if we dont support windows officially, then why the code is here?
REmvoe all related to windows.  Un-used code should NOT be committed as per coding rule. I am blocking this PR.

os/Kconfig
sunghan-chang: LOADABLEDIR and APPDIR are different.
LOADABLEDIR for loadable_apps folder
APPDIR for apps folder

They are same level of folder, not subdirectory.

PR #3521, opened 2019-07-06
arch/arm/assert : Display used stack size of idle task when asserted.

jeongchanKim:
Idle task's tcb has stack information now, so we can display the current used stack size when asserted.


seinfra:
Target : [b2c7535f5f5d5e70c4d0db5c7302ea61b925affb] - Code Rule Check (C++) OK.

seinfra:
Target : [b2c7535f5f5d5e70c4d0db5c7302ea61b925affb] - Code Rule Check OK.

PR #3528, opened 2019-07-08
apps/iperf: fix direct access to errno

SeonghoByeon:
- Use set/get_errno() APIs instead of direct access

### DIFF
--- a/apps/system/iperf/iperf_net.c
+++ b/apps/system/iperf/iperf_net.c
@@ -128,7 +128,7 @@ int netdial(int domain, int proto, char *local, int local_port, char *server, in
 	}
 
 	((struct sockaddr_in *)server_res->ai_addr)->sin_port = htons(port);
-	if (connect(s, (struct sockaddr *)server_res->ai_addr, server_res->ai_addrlen) < 0 && errno != EINPROGRESS) {
+	if (connect(s, (struct sockaddr *)server_res->ai_addr, server_res->ai_addrlen) < 0 && get_errno() != EINPROGRESS) {
 		close(s);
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [05a66909c34ba69ce303c0741d9a8bc3181d8428] - Code Rule Check (C++) OK.

seinfra:
Target : [05a66909c34ba69ce303c0741d9a8bc3181d8428] - Code Rule Check OK.

apps/system/iperf/iperf_net.c
```diff
@@ -238,7 +238,7 @@ int Nread(int fd, char *buf, size_t count, int prot)
 	while (nleft > 0) {
 		r = read(fd, buf, nleft);
 		if (r < 0) {
-			if (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK) {
+			if (get_errno() == EINTR || get_errno() == EAGAIN || get_errno() == EWOULDBLOCK) {
```
manoharahk: Better to load the get_errno() return value to an auto  variable and check inside if block to avoid multiple system calls incase of user/kernel mode

SeonghoByeon: That is good idea. But it requires to modify the whole system including network stack. At this moment, please let me use this simple logic temporarily. Thanks for the comment.

seinfra:
Target : [4549285e0c103a2fdd57194f74905c492ac1f7bc] - Code Rule Check (C++) OK.

seinfra:
Target : [4549285e0c103a2fdd57194f74905c492ac1f7bc] - Code Rule Check OK.

PR #3537, opened 2019-07-09
libc/time: fix wrong calculation and remove unnecessary codes

sunghan-chang:
1. fix wrong calculation of month in mktime()
 The month index starts from 0, not from 1 so that
mktime has not to add 1 for clock_calendar2utc().

2. remove unncessary codes for mktime()
 There are two mktime() functions with CONFIG_GREGORIAN_TIME.
 But they are same so that let's remove one.

3. fix wrong calculation of leap year in clock_calendar2utc
 To calculate the days of leap years, it has to subtract 1969
 from year value. But it misses ( ) so that it does 1971.

### DIFF
--- a/lib/libc/time/lib_calendar2utc.c
+++ b/lib/libc/time/lib_calendar2utc.c
@@ -199,7 +199,7 @@ time_t clock_calendar2utc(int year, int month, int day)
 
 	/* Add in the extra days for the leap years prior to the current year. */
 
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [d21ab3770e6900b03fe96b61531e798f78bc072e] - Code Rule Check (C++) OK.

seinfra:
Target : [d21ab3770e6900b03fe96b61531e798f78bc072e] - Code Rule Check OK.

PR #3539, opened 2019-07-10
[DO NOT MERGE] Provide read permission to App for GPT register area 

kishore-sn:
Since GPT register access through ioctl is taking more time, we are
provideing RO access permission for app to directly access the GPT
registers. This results in addition of a new region in the MPU. All
applications will have RO permission to access the GPT registers.

### DIFF
--- a/os/arch/arm/src/armv7-m/mpu-reg.h
+++ b/os/arch/arm/src/armv7-m/mpu-reg.h
@@ -46,6 +46,7 @@ enum {
 	MPU_REG_USER_DATA,
 #endif
 	MPU_REG_APP,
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [81f06dae18a0aa18211bba4ce02e9508a5011b67] - Code Rule Check OK.

seinfra:
Target : [81f06dae18a0aa18211bba4ce02e9508a5011b67] - Code Rule Check (C++) OK.

kishore-sn:
It is not possible to access peripheral memory address in unprevilaged mode. This is resulting in hard fault. We cannot continue with this approach. Hence closing the PR.

PR #3540, opened 2019-07-10
build/tools: Remove erase entire flash command.

ChinaTizenRT:
Remove erase entire flash command before downloading all images,
as it will take side effect for applications involves SPIRAM.

### DIFF
--- a/build/tools/esp32/esptool_py/Makefile
+++ b/build/tools/esp32/esptool_py/Makefile
@@ -74,7 +74,7 @@ $(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
 
 ESPTOOL_ALL_FLASH_ARGS += $(ESPTOOL_BOOTLOADER_FLASH_ARGS)
 
-ALL:erase_flash flash romfs
+ALL: flash romfs
 
 OS:app-flash
 


seinfra:
Target : [2e703c9c2118ff286ad72775e5a2bf8074f9a3aa] - Code Rule Check (C++) OK.

seinfra:
Target : [2e703c9c2118ff286ad72775e5a2bf8074f9a3aa] - Code Rule Check OK.

sunghan-chang:
@ChinaTizenRT I can't find where you modify. Let's write `build/tools/esp32`.
And as I remember, you added that recently. Could you let me know what side effect is?

ChinaTizenRT:
yes,  it is added recently. that time we met a strange issue,  that is 
romfs is mounted sucessfully, there is setting files in it but smart things application
 can not read it. so we erase all chip, this issue is fixed, so request to merge.

after a period, we found that system can not boot if add "erase all",
but if removing it when downling image, the system boots okay, 
so I think it will take some side effect, and not very sure  why it leads boot fail.

I will continue to check it.



but after merge it, i found that

sunghan-chang:
@ChinaTizenRT Could you let me know whether I merge this now or not? Do you find the root cause of boot failure?

ChinaTizenRT:
this PR should be merged,  as the second bootloader also resides in the flash,
if erasing entire all chipset, the second bootloader will be damaged.

Currently we can not reproduce the issue that romfs can not access, 
but that is another issue,  currently it needs to roll back.

Regarding the root cause of boot failure, we are asking espressif.

sunghan-chang:
@ChinaTizenRT Ok. then could you check my first comment?
```
@ChinaTizenRT I can't find where you modify. Let's write build/tools/esp32.
```

ChinaTizenRT:
Let's write build/tools/esp32.
  -- done.

seinfra:
Target : [7298f66c68194de7928d90a09e3390d5a3cb9087] - Code Rule Check OK.

seinfra:
Target : [7298f66c68194de7928d90a09e3390d5a3cb9087] - Code Rule Check (C++) OK.

PR #3555, opened 2019-07-15
arch/arm/src/armv7-m: Disabling MPU shareablity for user memory.

sijupunnnoose:
Shareability is not required for applications, it is used only when multiprocessor or DMA operation is required.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_mpu.c
+++ b/os/arch/arm/src/armv7-m/up_mpu.c
@@ -338,7 +338,6 @@ void mpu_user_intsram_context(uint32_t region, uintptr_t base, size_t size, uint
 	regval = MPU_RASR_ENABLE |      /* Enable region */
 			MPU_RASR_SIZE_LOG2((uint32_t)l2size) | /* Region size   */
 			((uint32_t)subregions << MPU_RASR_SRD_SHIFT) | /* Sub-regions   */
-			MPU_RASR_S |                   /* Shareable     */
 			MPU_RASR_C |                   /* Cacheable     */
 			MPU_RASR_AP_RWRW;              /* P:RW   U:RW   */
 	regs[2] = regval;


seinfra:
Target : [90da1113de5f788d6e81ddc59b24cc1e1faeb02a] - Code Rule Check OK.

seinfra:
Target : [90da1113de5f788d6e81ddc59b24cc1e1faeb02a] - Code Rule Check (C++) OK.

sunghan-chang:
@sijupunnnoose Is it possible to disable using Kconfig?

sijupunnnoose:
> @sijupunnnoose Is it possible to disable using Kconfig?

@Chang, i don't think we need to add it in config as it is not feature, it is just a register setting.

sunghan-chang:
@sijupunnnoose `@chang` is another guy :)
os/arch/arm/src/armv7-m/up_mpu.c is a common file of all boards which uses armv7-m architecture.
If another board uses that region, I think config is mandatory.

manoharahk:
IMO, It is good to append the function name with __non_shareable_ (if at all we need to support shareable version in future)

kishore-sn:
This API was specially created to support NXP app separation. So, I think its ok to modify this without any flags or change in API name. In future, if another board needs it, then it can create a new API with Shared region.

sunghan-chang:
@kishore-sn Is it only for NXP? Then, why do you add it in common file?
1. If it is valid only for NXP, let's move into nxp.
2. But if it could be common for all of armv7-m except reg_val, we had better make configs to support all of  boards with different value.

kishore-sn:
It is common for all Arm cortex M processors. But, currently its used only for NXP. 
In future, if a different board required Shared FLAG, then we can add a config to support it. In my opinion, its not required to add the config right now. Even in future boards, the Shared option will be required only in case of multicore scenario or when DMA will be used to access this memory region.

PR #3567, opened 2019-07-16
fs/poll: Clear slot for the poll structure reference when closing file

jeongarmy:
Check if file pointer is registered in a list of waiters for polling.
For example, when task A is blocked by calling poll and task B try to terminate task A, a pollfd of A remains in this list.
If it is, it should be cleared.


seinfra:
Target : [84a28e8327af57a1d022d8deba27e2598ebb8ca5] - Code Rule Check (C++) OK.

seinfra:
Target : [84a28e8327af57a1d022d8deba27e2598ebb8ca5] - Code Rule Check OK.

PR #3582, opened 2019-07-18
os/arch: Enable bluetooth null driver

amitpurwar:
Add bluetooth null driver to test the blueooth feature

Signed-off-by: amit purwar <amit.purwar@samsung.com>

PR #3586, opened 2019-07-22
os/arch: Enable bluetooth null driver

dh79pyun:
Add bluetooth null driver to test the blueooth feature

Signed-off-by: amit purwar <amit.purwar@samsung.com>

### DIFF
--- a/os/arch/arm/src/common/up_initialize.c
+++ b/os/arch/arm/src/common/up_initialize.c
@@ -62,6 +62,9 @@
 #include <tinyara/fs/fs.h>
 #include <tinyara/syslog/ramlog.h>
 #include <tinyara/syslog/syslog_console.h>
+#if defined(CONFIG_BLUETOOTH) && defined(CONFIG_BLUETOOTH_NULL)
+#include <tinyara/bluetooth/bt_null.h>
+#endif
 
 #include <arch/board/board.h>
 
@@ -206,6 +209,10 @@ void up_initialize(void)
 
 #if defined(CONFIG_DEV_NULL)
 	devnull_register();			/* Standard /dev/null */
+#endif
+
+#if defined(CONFIG_BLUETOOTH) && defined(CONFIG_BLUETOOTH_NULL)
+	btnull_register();    /* bluetooth bt_null */
 #endif
 
 #ifdef CONFIG_DEV_URANDOM


seinfra:
Target : [5220d83e9fb50b457915ae8c9929c24a5d68e3f5] - Code Rule Check (C++) OK.

seinfra:
Target : [5220d83e9fb50b457915ae8c9929c24a5d68e3f5] - Code Rule Check OK.

seinfra:
Target : [5b758cda0651955e8e9b22490aecffb983847bcc] - Code Rule Check (C++) OK.

seinfra:
Target : [5b758cda0651955e8e9b22490aecffb983847bcc] - Code Rule Check OK.

jeongchanKim:
There is same PR(#3582). Please check this.

dh79pyun:
> There is same PR(#3582). Please check this.

We will remove #3582, because the developer can't modify the patchset in now.
So please review this PR instead of #3582.

dh79pyun:
> please change commit title

Changed

os/arch/arm/src/common/up_initialize.c
```diff
@@ -208,6 +211,10 @@ void up_initialize(void)
 	devnull_register();			/* Standard /dev/null */
 #endif
 
+#if defined(CONFIG_BLUETOOTH_NULL)
```
Taejun-Kwon: We initialize common driver only here, this should be moved to other place.(board's bring up code)
also it should be done by below..
```

#ifdef CONFIG_WIRELESS_BLUETOOTH
#ifdef CONFIG_BLUETOOTH_XXXX // config for bt's lower driver
    dev = bt_XXXX_register(); 
    bt_netdev_register(dev);
#elif defined (CONFIG_BLUETOOTH_NULL)
    btnull_register();
#endif

seinfra:
Target : [eec0238c9c901ff36daafe50f4f77443f1ee9946] - Code Rule Check (C++) OK.

seinfra:
Target : [eec0238c9c901ff36daafe50f4f77443f1ee9946] - Code Rule Check OK.

os/arch/arm/src/common/up_initialize.c
dh79pyun: Can I know the current's NXP board's bring up code? I will move this null driver register code in there. And.. for the normal driver case. We will not use bt_XXXX_register directly. It will be called in each driver code when USB or UART is probing. And bt_netdev_register also will not used.

PR #3599, opened 2019-07-25
driver/otp: Add otp driver to access one time program area

pillip8282:
To provide otp service to applications and kernel modules otp driver
need to be defined. Applications can access otp by refering
"tinyara/otp.h" header. but it's not allowed to write otp. however,
kernel modules can write otp to store device specific data.

### DIFF
--- a/os/drivers/otp/otp.c
+++ b/os/drivers/otp/otp.c
@@ -1,0 +1,296 @@
+/****************************************************************************
+ *
+ * Copyright 2019 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <tinyara/config.h>
+
+#include <stdio.h>
+#include <semaphore.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <tinyara/kmalloc.h>
+#include <tinyara/otp.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+#define OTP_DEV_PATH "/dev/otp"
+
+#define OU_LOCK(priv)										\
+	do {													\
+		int oulock_ret = sem_wait(&priv->ou_exclsem);		\
+		if (oulock_ret < 0) {								\
+			lldbg("ERROR: ou_lock failed: %d\n", oulock_ret);	\
+			return oulock_ret;										\
+		}													\
+	} while (0)
+
+#define OU_UNLOCK(priv)									\
+	do {													\
+		int oulock_ret = sem_post(&priv->ou_exclsem);		\
+		if (oulock_ret < 0) {								\
+			lldbg("ERROR: ou unlock failed: %d\n", oulock_ret);	\
+			return oulock_ret;										\
+		}													\
+	} while (0)
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+static int     otp_open(struct file *filep);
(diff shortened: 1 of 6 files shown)


seinfra:
Target : [4c88724fb09d78ace698ef23efee2e0d0dd29853] - Code Rule Check (C++) OK.

seinfra:
Target : [4c88724fb09d78ace698ef23efee2e0d0dd29853] - Code Rule Check OK.

os/drivers/otp/otp.c
```diff
@@ -0,0 +1,240 @@
...
+
+/****************************************************************************
+ * Name: otp_write
+ *
+ * Description:
+ *    This function is called when you handle the write() API in
+ *    File system.
+ *
+ ****************************************************************************/
+static ssize_t otp_write(struct file *filep, const char *buffer,
```
Taejun-Kwon: Not allowed to user then please assign null instead otp_write in your ops.

os/drivers/otp/otp.c
```diff
@@ -0,0 +1,240 @@
...
+
+/****************************************************************************
+ * Name: get_otp_dev
+ *
+ * Description:
+ *   Get otp lower device.
+ *   Programs(SE) in a kernel space can access otp by this function
+ *
+ ****************************************************************************/
+otp_dev_s *get_otp_dev(void)
```
Taejun-Kwon: It is very dangerous approach, I explained to your group many times.
1. Upper never knows which lower is attached.
2. Lower should be controlled strictly, because only upper is registered in iNode structure because lower is private area of upper..
    It means that upper should maintain state of lower too.
    Please assume that there are 2 thread which use this function and read & write simultaneously.
    Each of data will be merged if upper do not lock it.

os/drivers/otp/otp.c
```diff
@@ -0,0 +1,240 @@
...
+ ****************************************************************************/
+static ssize_t otp_read(struct file *filep, char *buffer,
+			 size_t buflen)
+{
+	FAR struct inode *inode = filep->f_inode;
+	FAR struct otp_upperhalf_s *priv = inode->i_private;
+	FAR struct otp_lowerhalf_s *lower = priv->ou_lower;
+
+	// setting address should be updated
+	int res = lower->ops->read(lower, 0, (int8_t *)buffer, &buflen);
```
Taejun-Kwon: lower should return one of err number in errno.h
vfs will set errno with returned value of each file operation.
please see fs_read.c

os/drivers/otp/otpnull.c
```diff
@@ -0,0 +1,45 @@
...
+	dbg("%s(%s:%d)\n", __FUNCTION__, __FILE__, __LINE__);
+	return 0;
+}
+static int on_lock(struct otp_lowerhalf_s *dev)
+{
+	dbg("%s(%s:%d)\n", __FUNCTION__, __FILE__, __LINE__);
+	return 0;
+}
+
+static struct otp_ops_s g_on_ops = {
```
Taejun-Kwon: If there there's should be only one lower, then global is not needed.

os/drivers/otp/otpnull.c
```diff
@@ -0,0 +1,45 @@
...
+	on_lock,
+};
+
+struct otp_lowerhalf_s g_on_lower = {
+	&g_on_ops,
+	NULL,
+	NULL,
+};
+
+int create_otpnull(void)
```
Taejun-Kwon: FAR struct otp_lowerhalf_s *otp_null_initialize() {
   FAR struct otp_lower_half_s *dev;
   ...
   return dev;
}
```

File: os/drivers/otp/otpnull.c [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+{
+	otp_register(&g_on_lower);
+	return 0;
+}
+
+int destroy_otpnull(void)
```
Taejun-Kwon: I don't understand why destroy is needed??could you please explain me?

Taejun-Kwon:
I saw your description of commit right before, unfortunately we don't have a privilege now.
So you had better follow usage of TizenRT's driver.
(both of app & kernel should control file operation, app should use only read, kernel can access everything)
You can apply privilege later when it added(maybe root?)

seinfra:
Target : [975f88fad195854b67d5eda6ffc94369610c9133] - Code Rule Check OK.

seinfra:
Target : [975f88fad195854b67d5eda6ffc94369610c9133] - Code Rule Check (C++) OK.

pillip8282:
@Taejun-Kwon 
I'll follow your recommendation after TizenRT has privilege.

seinfra:
Target : [223642702618641b742a5590ac3791c82c5799d6] - Code Rule Check (C++) OK.

seinfra:
Target : [223642702618641b742a5590ac3791c82c5799d6] - Code Rule Check OK.

PR #3609, opened 2019-07-31
examples/mediaplayer: Add CmdView

junmin-kim:
Extract UI component from mediaplayer_main

### DIFF
--- a/apps/examples/mediaplayer/mediaplayer_main.cpp
+++ b/apps/examples/mediaplayer/mediaplayer_main.cpp
@@ -37,6 +37,7 @@
 #include <limits.h>
 
 #include "WiFiConnector.h"
+#include "CmdView.h"
 
 using namespace std;
 using namespace media;
@@ -46,16 +47,14 @@ using namespace media::stream;
 // Name: mediaplayer_main
 //***************************************************************************/
 
-static const int TEST_PCM = 0;
-static const int TEST_BUFFER = 1;
-static const int TEST_HTTP = 2;
+static const int TEST_PCM = 1;
+static const int TEST_BUFFER = 2;
+static const int TEST_HTTP = 3;
 
 static char TEST_FILE_PATH[128] = "/rom/44100.pcm";
 // We don't provide any song's URL, to avoid license issue.
 // Please fill a valid URL to `TEST_HTTP_URL` for testing!
 static const std::string TEST_HTTP_URL = "";
-
-static const int TEST_COMMAND_NUM = 8;
 
 enum test_command_e {
 	APP_OFF = 0,
@@ -67,10 +66,6 @@ enum test_command_e {
 	VOLUME_UP,
 	VOLUME_DOWN
 };
-
-extern "C" {
-static int list_dir_entries(const char *dirpath, char **filelist, int max);
-}
 
 class MyMediaPlayer : public MediaPlayerObserverInterface,
 					  public FocusChangeListener,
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [db89ce00c92a711f58008e0a29c8a8b47ec7c0ba] - Code Rule Check (C++) OK.

seinfra:
Target : [db89ce00c92a711f58008e0a29c8a8b47ec7c0ba] - Code Rule Check Result:
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that ':' (ctx:VxO)
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that ':' (ctx:OxV)
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that '<' (ctx:VxV)
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that ':' (ctx:VxO)
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that ':' (ctx:OxV)
apps/examples/mediaplayer/CmdView.h:31: ERROR: [SPC_M_OPR] spaces required around that '>' (ctx:VxW)

apps/examples/mediaplayer/mediaplayer_main.cpp
```diff
@@ -321,22 +316,25 @@ class MediaPlayerController
 	void start()
 	{
 		while (true) {
-			auto test = selectSource();
-			if (test < 0) {
+			vector<string> sourceList = {"Exit APP", "Test PCM", "Test BUFFER", "Test HTTP"};
+			listDirEntries("/rom", sourceList);
```
chanijjani: It seems that "sourceList" is loaded only once, isn't it..?

chanijjani: And how about defining "/rom" as a constant variable or macro..?

apps/examples/mediaplayer/mediaplayer_main.cpp
```diff
@@ -321,22 +316,25 @@ class MediaPlayerController
...
+			auto test = view.selectSource(sourceList);
+			if (test == 0) {
 				break;
+			} else if (test >= 4) {
```
chanijjani: How about replacing "4" as the size of "sourceList"?

junmin-kim: sourceList was appended in listDirEntries. So the size of sourceList is not 4

apps/examples/mediaplayer/mediaplayer_main.cpp
junmin-kim: sourceList can be changed, so it should be re-initialized

junmin-kim: The "/rom" is used only once. Is there any advantage if we use constant variable?

chanijjani: /rom is used partially in line 54, (static char TEST_FILE_PATH[128] = "/rom/44100.pcm";).
But replacing it is optional and up to you.

apps/examples/mediaplayer/mediaplayer_main.cpp
chanijjani: Hmm.. 4 can be stored to an integer variable at line 320 similar to line 404 (int offset = 4;) of the old source code.

apps/examples/mediaplayer/mediaplayer_main.cpp
chanijjani: Hmm.. it seems that all files under "/rom" are uploaded only once when flashing and no files can be created during runtime since "/rom" is read-only.
When and how sourceList can be changed..?

apps/examples/mediaplayer/mediaplayer_main.cpp
Taejun-Kwon: @chanijjani As per my understanding, 4 means that basic sourcelist, sourcelist append contents while it read contents in directory..
This will be enough answer for you.

PR #3634, opened 2019-08-08
kernel/task_monitor : Implement task monitor for checking if register…

jeongchanKim:
…ed tasks/pthreads is alive

Task Monitor checks registered tasks/pthreads alive flag on every interval.
If there is an inactive task/pthread, task monitor resets the system.
Alive flag can be updated through context switching or user api(task_monitor_update_status), and
it will be reset on every interval.

User API
1) void task_monitor_register(void) : Register to task monitor, Task monitor only checks registered one.
2) void task_monitor_update_status(void) : Update the alive flag

### DIFF
--- a/os/kernel/task/task_prctl.c
+++ b/os/kernel/task/task_prctl.c
@@ -73,6 +73,10 @@
 #include "sched/sched.h"
 #include "task/task.h"
 
+#ifdef CONFIG_TASK_MONITOR
+#include "task_monitor/task_monitor_internal.h"
+#endif
+
 /************************************************************************
  * Private Functions
  ************************************************************************/
@@ -237,6 +241,24 @@ int prctl(int option, ...)
 #endif
 	}
 	break;
+#ifdef CONFIG_TASK_MONITOR
+	case PR_MONITOR_REGISTER:
+	{
+		task_monitor_update_list(getpid(), MONITORED);
+	}
+	break;
+	case PR_MONITOR_UPDATE:
+	{
+		struct tcb_s *tcb;
+		tcb = sched_gettcb(getpid());
+		if (tcb == NULL) {
+			sdbg("Fail to update the status in task monitor.\n");
+			goto errout;
+		}
+		tcb->is_active = true;
+	}
+	break;
+#endif
 	default:
 		sdbg("Unrecognized option: %d\n", option);
 		err = EINVAL;
--- a/os/kernel/task_monitor/task_monitor.c
+++ b/os/kernel/task_monitor/task_monitor.c
@@ -1,0 +1,80 @@
+/****************************************************************************
+ *
+ * Copyright 2019 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <tinyara/config.h>
+#include <sched.h>
+#include <sys/boardctl.h>
+#include <tinyara/sched.h>
+#include "task_monitor_internal.h"
+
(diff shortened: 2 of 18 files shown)


seinfra:
Target : [c9b4f4eb421cb7a9d47143161e7e0b39e1a1ce03] - Code Rule Check (C++) OK.

seinfra:
Target : [c9b4f4eb421cb7a9d47143161e7e0b39e1a1ce03] - Code Rule Check OK.

seinfra:
Target : [084ed583bb87e1dcd693133dca762bfc677d69de] - Code Rule Check (C++) OK.

seinfra:
Target : [084ed583bb87e1dcd693133dca762bfc677d69de] - Code Rule Check OK.

os/kernel/task/task_prctl.c
```diff
@@ -237,6 +241,32 @@ int prctl(int option, ...)
...
+		tcb = sched_gettcb(getpid());
+		if (tcb == NULL) {
+			sdbg("Fail to update the status in task monitor.\n");
+			goto errout;
+		}
+		tcb->is_active = true;
+	}
+	break;
+#else
+	case PR_MONITOR_REGISTER:
```
Taejun-Kwon: if config wrapped these constant itself, you don't have to make #else part...because caller can't use these constant.
It will go to default directly.

os/kernel/task_monitor/task_monitor.c
```diff
@@ -0,0 +1,75 @@
...
+{
+	int pid_idx;
+	struct tcb_s *tcb;
+
+	/* Initialize the monitored tasks list */
+	for (pid_idx = 0; pid_idx < CONFIG_MAX_TASKS; pid_idx++) {
+		monitored_tasks_list[pid_idx] = NOT_MONITORED;
+	}
+
+	while (1) {
```
Taejun-Kwon: How about consider expandability??
1. Make kernel demon.
2. Attach each monitor to demon.
Then you can maintain only one demon(one thread)

sijupunnnoose:
Why dont we use watchdogs instead, which is commonly used to monitor task?? Here the time is fixed, it may requires to monitor task with different reporting interval which is not possible with current design. We can do it with a single s/w watchdog timer, each task can have different reporting intervals, timer can be initialized with lowest interval value, and monitor task check on every bark and reset the device based on it.

jeongchanKim:
> Why dont we use watchdogs instead, which is commonly used to monitor task?? Here the time is fixed, it may requires to monitor task with different reporting interval which is not possible with current
Yes, you are right. I will modify it to use workqueue for different interval on each tasks. 
We can use h/w watchdog, but there is a limitation about wdog number.

sijupunnnoose:
I think we can handle it with single watchdog, configure the watchdog for lowest reporting time or divisible value. Let say 2sec, 3sec  and 4sec are reporting interval for T1, T2 and T3, configure watchdog for 1sec and every bark we have to check elapse timer and reset the device.

jeongchanKim:
> I think we can handle it with single watchdog, configure the watchdog for lowest reporting time or divisible value. Let say 2sec, 3sec and 4sec are reporting interval for T1, T2 and T3, configure watchdog for 1sec and every bark we have to check elapse timer and reset the device.

Let's assume like what you said. There are 3 interval as 2sec, 3sec, and 4sec. You told me that configuring wdog for 1sec. TaskA(2sec) can be alive after 1.1sec, but in this case, system will be reset(even not the reset case...). It is almost same to set the lowest reporting time.

And I have another worries, because reporting(registering) can happen every time, not simultaneously. In this case, it is not easy to configure properly.
So I will change this logic to workqueue.

Please tell me your opinion.

sijupunnnoose:
Time granularity is an issues with single timer, Let say T1 interval is 2 sec, in that case reset has to be after 2 barks from registering which means reset can happen anywhere between 2sec~3sec. Yes i agree, it will not be exact time specified but higher than it. I didn't get work-queue implementation, if you give more details it would be great, or we will discuss once you come up with initial implementation.

os/kernel/task_monitor/task_monitor.c
jeongchanKim: Yes. It is a good idea. I will expand it with another commit.

seinfra:
Target : [3cbeefdef59da3188d40bafb9a3d6c1e5543fb82] - Code Rule Check (C++) OK.

seinfra:
Target : [3cbeefdef59da3188d40bafb9a3d6c1e5543fb82] - Code Rule Check OK.

seinfra:
Target : [04a675b9747e2772f8deb005a692d7fc1c918e7d] - Code Rule Check (C++) OK.

seinfra:
Target : [04a675b9747e2772f8deb005a692d7fc1c918e7d] - Code Rule Check OK.

PR #3665, opened 2019-08-26
board/imxrt1050-evk/imxrt_boot.c : Add watchdog initialization

jeongchanKim:
Add imxrt 1050 watchdog initialization in boot code.
There are two watchdogs in imxrt1050, but now we use only one temporarily.(WDOG1)

### DIFF
--- a/os/board/imxrt1050-evk/src/imxrt_boot.c
+++ b/os/board/imxrt1050-evk/src/imxrt_boot.c
@@ -85,6 +85,11 @@
 #include "imxrt_semc_sdram.h"
 #endif
 
+#ifdef CONFIG_WATCHDOG
+#include <tinyara/watchdog.h>
+#include "imxrt_wdog.h"
+#endif
+
 #ifdef CONFIG_ANALOG
 #include "imxrt_adc.h"
 #endif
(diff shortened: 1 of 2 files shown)


seinfra:
Target : [549445b1ae9ecfbab27a1823c5bb8b851e7b9014] - Code Rule Check OK.

seinfra:
Target : [549445b1ae9ecfbab27a1823c5bb8b851e7b9014] - Code Rule Check (C++) OK.

os/board/imxrt1050-evk/src/imxrt_boot.c
```diff
@@ -260,6 +265,10 @@ void board_initialize(void)
 
 	imxrt_spi_initialize();
 
+#ifdef CONFIG_WATCHDOG
+	imxrt_wdog_initialize(CONFIG_WATCHDOG_DEVPATH, IMXRT_WDOG1);
```
Taejun-Kwon: Actually we provide up_wdginitialize as a common interface to initialize watchdog.
In this case, imxrt should inplmenet up_wdginitialize, and then up_wdginitialize will be called in up_initialize after timer initialized as a common feature, like devnull, timer, and so on.
But if you do not want follow this rule, please remove up_wdginitialize from watchdog.h

seinfra:
Target : [f2b966d448136e788d67675722a38ee8a85f8eae] - Code Rule Check (C++) OK.

seinfra:
Target : [f2b966d448136e788d67675722a38ee8a85f8eae] - Code Rule Check OK.

os/board/imxrt1050-evk/src/imxrt_boot.c
jeongchanKim: I removed up_wdinitialize in watchdog.h
Some modules, like gpio, timer, and so on, can be used from another, but wdinitialize is only used for board specific code.
Thanks.

PR #3668, opened 2019-08-26
os/crypto/random_pool.c:Typecast 'void *' to 'size_t *' for correct calculations

drashti304:
When using void pointers in calculations, the behavior is undefined.
Arithmetic operations on 'void *' is a GNU C extension,
which defines the 'sizeof(void)' to be 1.

Note: This patch solve SonarQube issue

### DIFF
--- a/os/crypto/random_pool.c
+++ b/os/crypto/random_pool.c
@@ -345,7 +345,7 @@ static void rng_buf_internal(FAR void *bytes, size_t nbytes)
 		blake2s_update(&g_rng.blake2xs.ctx, g_rng.blake2xs.out_root, sizeof(g_rng.blake2xs.out_root));
 		blake2s_final(&g_rng.blake2xs.ctx, bytes, block_size);
 
-		bytes += block_size;
+		bytes = (size_t *)bytes + block_size;
 		nbytes -= block_size;
 	}
 }


seinfra:
Target : [1917c3df087a96125e2a9df019065947c292c1b1] - Code Rule Check OK.

seinfra:
Target : [1917c3df087a96125e2a9df019065947c292c1b1] - Code Rule Check (C++) OK.

os/crypto/random_pool.c
```diff
@@ -345,7 +345,7 @@ static void rng_buf_internal(FAR void *bytes, size_t nbytes)
 		blake2s_update(&g_rng.blake2xs.ctx, g_rng.blake2xs.out_root, sizeof(g_rng.blake2xs.out_root));
 		blake2s_final(&g_rng.blake2xs.ctx, bytes, block_size);
 
-		bytes += block_size;
+		(size_t *)bytes += block_size;
```
junmin-kim: I tested this code in MSVC, but it makes build error(C2106).
How about this?
```c
bytes = (size_t *)bytes + block_size;
```

drashti304: sure, thanks.
Updated

seinfra:
Target : [837f11df20445ee3147d0d6375eb7e2f0bf244d5] - Code Rule Check (C++) OK.

seinfra:
Target : [837f11df20445ee3147d0d6375eb7e2f0bf244d5] - Code Rule Check OK.

PR #3675, opened 2019-08-29
app/testcase: Modify network tc

SeonghoByeon:
- The first input of inet_aton should not be NULL, because it will cause a crash. Instead, Change it to '-1' which is never used for network addr.
- Send socket operation is able to send NULL msg to the target socket. Therefore, itc_net_send_n_msg is actually, positive case, expecting return 0, not -1.
- strlen(NULL) is not allowed


seinfra:
Target : [d42de0036939f721d9b80dc6735eaf67d45dfae9] - Code Rule Check (C++) OK.

seinfra:
Target : [d42de0036939f721d9b80dc6735eaf67d45dfae9] - Code Rule Check OK.

PR #3684, opened 2019-09-02
network: modify unsafe libc function

pillip8282:
change sprintf which is unsafe to snprintf

### DIFF
--- a/lib/libc/net/lib_inetntop.c
+++ b/lib/libc/net/lib_inetntop.c
@@ -114,7 +114,7 @@ static int inet_ipv4_ntop(FAR const void *src, FAR char *dest, socklen_t size)
 	}
 
 	ptr = (FAR uint8_t *)src;
-	sprintf(dest, "%u.%u.%u.%u", ptr[0], ptr[1], ptr[2], ptr[3]);
+	snprintf(dest, INET_ADDRSTRLEN, "%u.%u.%u.%u", ptr[0], ptr[1], ptr[2], ptr[3]);
 	return OK;
 }
 #endif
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [dbc4b67200ec7a6bc697a53cb8782b31157f5d1f] - Code Rule Check (C++) OK.

seinfra:
Target : [dbc4b67200ec7a6bc697a53cb8782b31157f5d1f] - Code Rule Check OK.

lib/libc/net/lib_inetntop.c
```diff
@@ -114,7 +114,7 @@ static int inet_ipv4_ntop(FAR const void *src, FAR char *dest, socklen_t size)
 	}
 
 	ptr = (FAR uint8_t *)src;
-	sprintf(dest, "%u.%u.%u.%u", ptr[0], ptr[1], ptr[2], ptr[3]);
+	snprintf(dest, INET_ADDRSTRLEN, "%u.%u.%u.%u", ptr[0], ptr[1], ptr[2], ptr[3]);
```
sangwon03: what is the correct buffer size?  "socklen_t size" or "INET_ADDRSTRLEN"

pillip8282: yes it's correct. refer below description
*   dest - The dest argument points to a buffer where the function stores
 *          the resulting text string; it shall not be NULL.
 *   size - The size argument specifies the size of this buffer, which must
 *          be large enough to hold the text string (INET_ADDRSTRLEN
 *          characters for IPv4, INET6_ADDRSTRLEN characters for IPv6).

PR #3688, opened 2019-09-03
Make crashdump for common

jeongchanKim:
1.    arch/arm/src : Add up_watchdog_disable
    up_watchdog_disable calls board specific watchdog disable function.

2.     board/crashdump : Make crashdump.c for common
    For using crashdump with another board, make crashdump.c in board/common, and remove artik specific crashdump.c

### DIFF
--- a/os/arch/arm/src/common/up_watchdog.h
+++ b/os/arch/arm/src/common/up_watchdog.h
@@ -1,0 +1,42 @@
+/****************************************************************************
+ *
+ * Copyright 2019 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+#ifndef __ARCH_ARM_SRC_ARM_WATCHDOG_H
+#define __ARCH_ARM_SRC_ARM_WATCHDOG_H
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: up_watchdog_disable
+ *
+ * Description:
+ *   Disable board specific watchdog
+ *
+ * Cautions:
+ *   This can be only used if we cannot use driver structure like assert.
+ *
+ ****************************************************************************/
+void up_watchdog_disable(void);
+
+#endif							/* __ARCH_ARM_SRC_ARM_WATCHDOG_H */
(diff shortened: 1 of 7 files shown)


seinfra:
Target : [44e77cf4e94b2f9229f83c6e460eb7482340710a] - Code Rule Check (C++) OK.

seinfra:
Target : [44e77cf4e94b2f9229f83c6e460eb7482340710a] - Code Rule Check OK.

os/include/tinyara/watchdog.h
```diff
@@ -273,6 +273,8 @@ void watchdog_unregister(FAR void *handle);
  * Architecture-specific Application Interfaces
  ****************************************************************************/
 
+void up_watchdog_disable(void);
```
Taejun-Kwon: all functions here is for upperhalf(like a register, unregister), you have to find another way.
Lower should be controlled by only upper. 
In other words, ioctl, I mean

sunghan-chang: It's for internal usages of kernel side and it is required by scheduler or assert regardless of user. Currently it is used in assert only. In assert, we can't use lower/upperhalf structure.

Taejun-Kwon: I don't agree, there are serveral reasons why do not allow this code.
Also I don't mean that you have to use lower/upperhalf, you have to use file operation.(ioctl I mean) Please don't access lower driver's function directly. see below.

1. each of module.h's definition including upper's interface such as register, unregister.
   Also define lower half's operation, such as watchdog_ops_s.
   All real device should inherits this operation only. So virtual driver(upper) doesn't know addtional function except these function set.
  So, that is, all device don't have to implement function what junchan has added, because it is not mandatory function, just breaked rule.

2. virtual driver allow to access exclusively, so please assume that it is under controlling by the another place normally.
  (for example, there's counting driver, and app is counting number from driver. but system reset counting number before current working is not finished. if count is initialized as a 0 and app trying to mod with 0, then??)

sunghan-chang: What you say is for normal status. In assert status, we have not to use file operation.

Taejun-Kwon: Will explain you in private channel

seinfra:
Target : [a446472bbe3e0678c01b240c74fe144c225a4b04] - Code Rule Check OK.

seinfra:
Target : [a446472bbe3e0678c01b240c74fe144c225a4b04] - Code Rule Check (C++) OK.

seinfra:
Target : [598879a22fd47c08e30bed4dbcd0372a5cdf4f71] - Code Rule Check (C++) OK.

seinfra:
Target : [598879a22fd47c08e30bed4dbcd0372a5cdf4f71] - Code Rule Check OK.

seinfra:
Target : [252b8d6a8b1d165b79a88c11cda3bca235d961d1] - Code Rule Check (C++) OK.

seinfra:
Target : [252b8d6a8b1d165b79a88c11cda3bca235d961d1] - Code Rule Check OK.

seinfra:
Target : [51a5251e270852f299935fa70968cd5b10673efd] - Code Rule Check (C++) OK.

seinfra:
Target : [51a5251e270852f299935fa70968cd5b10673efd] - Code Rule Check OK.

seinfra:
Target : [08767503b1ec83a01c6bb29333501fe49f83e8de] - Code Rule Check (C++) OK.

seinfra:
Target : [08767503b1ec83a01c6bb29333501fe49f83e8de] - Code Rule Check OK.

PR #3710, opened 2019-09-10
lib/libc/lib_localtime : Remove unnecessary NULL check

jeongchanKim:
p is assigned to TZDIR, and TZDIR cannot be NULL.
So there is no needed to check that p is NULL.


seinfra:
Target : [6126fe9a17686bbec1c6e4c1f01ef3aab7ad0ba7] - Code Rule Check (C++) OK.

seinfra:
Target : [6126fe9a17686bbec1c6e4c1f01ef3aab7ad0ba7] - Code Rule Check OK.

PR #3757, opened 2019-09-18
timer_main : Change timer_main to specific timer_xx_main

jeongchanKim:
timer_main is very common name, so it can be a problem if there are same func name.

Signed-off-by: jc_.kim <jc_.kim@samsung.com>


seinfra:
Target : [8d5aa5197b02f4017700efa28bba8df51c7e19dc] - Code Rule Check OK.

seinfra:
Target : [8d5aa5197b02f4017700efa28bba8df51c7e19dc] - Code Rule Check (C++) OK.

apps/examples/testcase/le_tc/kernel/kernel_tc_main.c
```diff
@@ -179,7 +179,7 @@ int tc_kernel_main(int argc, char *argv[])
 #endif
 
 #ifdef CONFIG_TC_KERNEL_TIMER
-	timer_main();
+	timer_tc_main();
```
sijupunnnoose: Why this change only for timer test, if so we should make the convention same for all the tc's

sunghan-chang: Because it has too common name, it easily causes the same problem we think. That's why we modify it together. Another tc also needs to be modified later.

PR #3758, opened 2019-09-18
apps/testcase/messaging : Add ommitted NULL initialization and NULL t…

jeongchanKim:
…estcase

1. After free, its pointer will not be NULL, just deallocate the memory.
For testing NULL case, pointer should be set to NULL.
2. For testing coverage, add testcase for NULL check in messaging_send API

Signed-off-by: jc_.kim <jc_.kim@samsung.com>


seinfra:
Target : [3767141a03d3bebef155811a22c94e826fce7eaf] - Code Rule Check (C++) OK.

seinfra:
Target : [3767141a03d3bebef155811a22c94e826fce7eaf] - Code Rule Check OK.

PR #3759, opened 2019-09-18
apps/examples/testcase/le_tc/filesystem: Add MTD path depending on specific board

KimDongEon:
Add MTD path depend on imxrt1020

### DIFF
--- a/apps/examples/testcase/le_tc/filesystem/fs_main.c
+++ b/apps/examples/testcase/le_tc/filesystem/fs_main.c
@@ -134,7 +134,13 @@
 #define INV_FD -3
 
 #define MTD_CONFIG_PATH "/dev/config"
+#if defined(CONFIG_ARCH_BOARD_CY4390x)
+#define MTD_FTL_PATH "/dev/mtdblock4"
+#elif defined(CONFIG_ARCH_BOARD_IMXRT1020_EVK)
+#define MTD_FTL_PATH "/dev/smart0p6"
+#else
 #define MTD_FTL_PATH "/dev/mtdblock1"
+#endif
 #define BUF_SIZE 4096
 
 #define DEV_PATH "/dev"


seinfra:
Target : [00b38dafc5d549cb8a4aaf834880bc6f4a99485f] - Code Rule Check (C++) OK.

seinfra:
Target : [00b38dafc5d549cb8a4aaf834880bc6f4a99485f] - Code Rule Check OK.

apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -134,7 +134,13 @@
 #define INV_FD -3
 
 #define MTD_CONFIG_PATH "/dev/config"
+#if defined(CONFIG_ARCH_BOARD_CY4390x)
+#define MTD_FTL_PATH "/dev/mtdblock4"
+#elif defined(CONFIG_ARCH_BOARD_IMXRT1020_EVK)
+#define MTD_FTL_PATH "/dev/smart0p6"
```
sunghan-chang: why 6?

KimDongEon: Userfs is registered as 6 during the partitioning process.

sunghan-chang: In specific defconfig, it is. But it does not guarantee all of defconfig has smartfs with 6 index. To make this working, you should enable automount with mountpoint.

KimDongEon: In this code, the MTD_FTL_PATH define is used as a test path for reading and writing directly to the MTD block dev.

Therefore, to run tc, you need the dev path before mounting.

However, registration numbers are registered in the order defined in CONFIG_FLASH_PART_TYPE in the defconfig file. Is there a way to unify this order with other defconfigs?

PR #3770, opened 2019-09-19
external/curl: fix logical build error.

Hyunjun85:
The curl has a dependency for mbedtls or openssl to use https protocols.
But the dependency was ignored.

### DIFF
--- a/external/curl/Kconfig
+++ b/external/curl/Kconfig
@@ -123,7 +123,7 @@ if SSL_SUPPORT
 config MBEDTLS
 	bool "Mbed TLS"
 	default y
-	depends on !NET_SECURITY_TLS
+	depends on NET_SECURITY_TLS
 
 endif #SSL_SUPPORT
 
--- a/external/include/mbedtls/config.h
+++ b/external/include/mbedtls/config.h
@@ -2701,9 +2701,9 @@
  *
  * This module is required for X.509 CRL parsing.
  */
-#if defined(MBEDTLS_OCF_PATCH)
+// #if defined(MBEDTLS_OCF_PATCH)
 #define MBEDTLS_X509_CRL_PARSE_C
-#endif
+// #endif
 
 /**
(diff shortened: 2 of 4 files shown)


seinfra:
Target : [0ce3cb20a723b9f877c96e91dd6c756a2af21563] - Code Rule Check (C++) OK.

seinfra:
Target : [0ce3cb20a723b9f877c96e91dd6c756a2af21563] - Code Rule Check OK.

external/curl/Kconfig
```diff
@@ -123,7 +123,7 @@ if SSL_SUPPORT
 config MBEDTLS
 	bool "Mbed TLS"
 	default y
-	depends on !NET_SECURITY_TLS
```
SeonghoByeon: Does this enable mbedtls ported into TizenRT? NET_SECURITY_TLS seems a flag to select which mbedtls is built.

external/include/mbedtls/config.h
```diff
@@ -2701,9 +2701,9 @@
  *
  * This module is required for X.509 CRL parsing.
  */
-#if defined(MBEDTLS_OCF_PATCH)
```
SeonghoByeon: This modification in public repo might incur a potential problem for other applications and/or security framework. Or I think X509 option is independent of OCF Patch. Please verify that the source code related width MBEDTLS_X509_CRL_PARSE_C is independent on MBEDTLS_OCF_PATCH.

external/include/mbedtls/config.h
```diff
@@ -2882,7 +2882,7 @@
 #undef MBEDTLS_ERROR_STRERROR_DUMMY
 #undef MBEDTLS_GENPRIME
 
-#undef MBEDTLS_FS_IO
```
SeonghoByeon: This might be conflict with our file system. It should be verified before disabled.

PR #3813, opened 2019-09-25
framework/araui: Apply partial display update to rotate and scale APIs.

tyha1115:
Previously, partial display update algorithm was not applied to the rotate and scale APIs.
(ui_widget_set_rotation(), ui_widget_set_scale())

From now on, partial display update algorithm is usable when rotate and scale APIs are called.

In order to provide this feature, the below modifications are applied.
1. transformation matrix is stored to the widget.
2. update transform matrix step is added before rendering widgets.
3. ui_window_add_redraw_list() is only called at the core.
(except ui_widget_set_position(). For this case, the previous location also needed to be update.)

### DIFF
--- a/framework/src/araui/core/ui_core.c
+++ b/framework/src/araui/core/ui_core.c
@@ -266,26 +266,21 @@ static ui_error_t _ui_render_widget_recur(ui_widget_body_t *widget, ui_rect_t dr
 	}
 
 	if (widget->visible) {
-		ui_renderer_push_matrix();
-
-		ui_renderer_translate((float)widget->local_rect.x, (float)widget->local_rect.y);
-		ui_renderer_rotate(widget->degree);
-		ui_renderer_scale(widget->scale_x, widget->scale_y);
+		new_vp = ui_rect_intersect(draw_area, widget->global_rect);
+		ui_dal_set_viewport(new_vp.x, new_vp.y, new_vp.width, new_vp.height);
 
 		if (widget->render_cb) {
 			new_vp = ui_rect_intersect(draw_area, widget->global_rect);
 			ui_dal_set_viewport(new_vp.x, new_vp.y, new_vp.width, new_vp.height);
 
 			widget->render_cb((ui_widget_t)widget, dt);
-			
+
 			ui_dal_set_viewport(draw_area.x, draw_area.y, draw_area.width, draw_area.height);
 		}
 
 		vec_foreach(&widget->children, child, iter) {
 			_ui_render_widget_recur(child, draw_area, dt);
 		}
-
-		ui_renderer_pop_matrix();
 	}
 
 	return UI_OK;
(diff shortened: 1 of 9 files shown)


seinfra:
Target : [42196a85793e26bf5f8fa29c2e845f1d8be82309] - Code Rule Check (C++) OK.

seinfra:
Target : [42196a85793e26bf5f8fa29c2e845f1d8be82309] - Code Rule Check OK.

seinfra:
Target : [5b11a00b0ec0edb7fa2e0a52ff6a06cff3e0cdbf] - Code Rule Check OK.

seinfra:
Target : [5b11a00b0ec0edb7fa2e0a52ff6a06cff3e0cdbf] - Code Rule Check (C++) OK.

seinfra:
Target : [fa04cd22a9ed997518bbd0f178c56850a161ef3c] - Code Rule Check OK.

seinfra:
Target : [fa04cd22a9ed997518bbd0f178c56850a161ef3c] - Code Rule Check (C++) OK.

framework/src/araui/core/ui_core.c
```diff
@@ -330,6 +325,48 @@ static void _ui_redraw(uint32_t dt)
...
+		ui_renderer_scale(&widget->trans_mat, widget->scale_x, widget->scale_y);
+
+		// Update new area
+		ui_widget_update_global_rect(widget);
+		ui_window_add_redraw_list(widget->global_rect);
+		widget->update_flag = false;
+		update_flag = true;
+	}
+
+	vec_foreach(&widget->children, child, iter) {
```
Taejun-Kwon: I think you can change this to loop. We have a small memory, we have to reduce stack as possible as we can.

framework/src/araui/core/ui_core.c
```diff
@@ -330,6 +325,48 @@ static void _ui_redraw(uint32_t dt)
...
+	update_flag = false;
+}
+
+static void _ui_update_redraw_list(void)
+{
+	ui_window_body_t *window;
+
+	window = ui_window_get_current();
+	if (window) {
+		_ui_update_redraw_list_recur(window->root);
```
Taejun-Kwon: you don't have to check validation of root?

framework/src/araui/core/ui_core.c
tyha1115: Thank you for your opinion. Without recursive function call is much better than current one. I'll modify it.

tyha1115: Hmm. I'll try to change this to the loop.

framework/src/araui/core/ui_core.c
tyha1115: Thanks for your opinion. If window does not have `root`, then it won't operate correctly. Therefore, I'll add assert for this case.

seinfra:
Target : [61fdcaeb51ab196367d0c9c8479d31e6bbdd3b3d] - Code Rule Check (C++) OK.

seinfra:
Target : [61fdcaeb51ab196367d0c9c8479d31e6bbdd3b3d] - Code Rule Check OK.

PR #3819, opened 2019-09-26
kernel/binary_manager : Remove unnecessary return value

jeongchanKim:
Binary manager always runs even with a request failure.
So we don't need to get return value from request.


seinfra:
Target : [1bf33c58a3db22008ce09aa6fb59a1f4583180e7] - Code Rule Check (C++) OK.

seinfra:
Target : [1bf33c58a3db22008ce09aa6fb59a1f4583180e7] - Code Rule Check OK.

PR #3847, opened 2019-09-30
framework/araui: Add return check procdures

tyha1115:
- In order to prevent unpredicted operations, add return value check procedures.
- When return value of the queue is NULL, it means error case (i.e, memory violations). Therefore it considered as an error case.


seinfra:
Target : [82161890faf2344346f53524afd9327cc893ebd6] - Code Rule Check OK.

seinfra:
Target : [82161890faf2344346f53524afd9327cc893ebd6] - Code Rule Check (C++) OK.

PR #3848, opened 2019-10-01
coap: fix coap build on Artik053

pillip8282:
TLS in coap used easy tls module that isn't supported anymore.
so TizenRT disable it until easy tls is replaced.

### DIFF
--- a/external/include/protocols/libcoap/config.h
+++ b/external/include/protocols/libcoap/config.h
@@ -37,11 +37,14 @@
 #undef NDEBUG
 #endif
 
-#ifdef CONFIG_NET_SECURITY_TLS
-#define WITH_MBEDTLS
-#else
-#undef WITH_MBEDTLS
-#endif
+/* coap refers easy_tls not supported anymore
+ * So WITH_MBEDTLS is disabled until change easy_tls
+ */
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [9063329e483689354159cf6d0cf0615191ffed2d] - Code Rule Check (C++) OK.

seinfra:
Target : [9063329e483689354159cf6d0cf0615191ffed2d] - Code Rule Check OK.

external/include/protocols/libcoap/config.h
```diff
@@ -37,11 +37,14 @@
...
 
-#ifdef CONFIG_NET_SECURITY_TLS
-#define WITH_MBEDTLS
-#else
-#undef WITH_MBEDTLS
-#endif
+/* coap refers easy_tls not supported anymore
+ * So WITH_MBEDTLS is disabled until change easy_tls
+ */
+/* #ifdef CONFIG_NET_SECURITY_TLS */
```
SeonghoByeon: Should we apply another config instead of CONFIG_NET_SECURITY_TLS, to enable mbedtls?

pillip8282: it use easy tls which is not supported anymore. so disabling CONFIG_NET_SECURITY_TLS is not solution

PR #3866, opened 2019-10-03
framework/messaging : Unify the error return case as goto

jeongchanKim:
There are same error return statements. They can be unified as goto.


seinfra:
Target : [480de0ed402c57eaec7054ae3e6fb857e05e2f51] - Code Rule Check (C++) OK.

seinfra:
Target : [480de0ed402c57eaec7054ae3e6fb857e05e2f51] - Code Rule Check OK.

PR #3870, opened 2019-10-07
Add strftime testcase and add missing explanation of strftime

hasw7569:
1.
In certain cases strftime function returns a different value.
For example, "%a" case can return "Day" or a three letter abbreviation for the day of the week depending on whether the particular CONFIG value is on off or not.
So I added #if line to handle this.

2.
libc_strftime.c have no explanation of %a and %A cases.
so I added explanation of this two cases.


seinfra:
Target : [cfe579ef09bea7dcf0cbbe09594a2ffc1d0b86eb] - Code Rule Check (C++) OK.

seinfra:
Target : [cfe579ef09bea7dcf0cbbe09594a2ffc1d0b86eb] - Code Rule Check OK.

PR #3891, opened 2019-10-08
fixup! framework/messaging : Unify the error return case as goto

sunghan-chang:
Commit e7f8067 unified the error return case using goto.
If there is no error, it has to finish without freeing memory allocated.
So, return should be located before goto.
This commit fixes the wrong location of return.


seinfra:
Target : [a35b3fbe895e4ef6c12d81aee0263d53ff20e45c] - Code Rule Check (C++) OK.

seinfra:
Target : [a35b3fbe895e4ef6c12d81aee0263d53ff20e45c] - Code Rule Check OK.

PR #3902, opened 2019-10-10
drivers/task_manager : Add NULL checking after sched_gettcb

jeongchanKim:
sched_gettcb can return NULL when cannot find the tcb.
In this case, we should not use the tcb in the below.


seinfra:
Target : [01be2d9e07d78d3d75e2f69c1dc12bf0e3dc9745] - Code Rule Check OK.

seinfra:
Target : [ed4f6722b8f802c7464024cbb8e853f5ab73a4fc] - Code Rule Check (C++) OK.

seinfra:
Target : [ed4f6722b8f802c7464024cbb8e853f5ab73a4fc] - Code Rule Check OK.

PR #3918, opened 2019-10-10
media/audiomanager: Fix resampler bug

Taejun-Kwon:
Capture should resample data based on user input.(Contrary case for playback)


seinfra:
Target : [9addae4ac66fae99e388a7985f401e8d690f2e7b] - Code Rule Check OK.

PR #3931, opened 2019-10-11
os/binfmt: Read symbol table and relocation table into RAM

kishore-sn:
Allocate memory from kernel heap for symbol table and reloc table.
Read complete symbol table and relocation table into RAM before
starting the relocation process. All accesses to the symbol and
relocation table will be done using the RAM instead of Flash. This
will reduce the relocation time.

### DIFF
--- a/os/binfmt/libelf/libelf_bind.c
+++ b/os/binfmt/libelf/libelf_bind.c
@@ -65,6 +65,7 @@
 #include <tinyara/elf.h>
 #include <tinyara/binfmt/elf.h>
 #include <tinyara/binfmt/symtab.h>
+#include <tinyara/kmalloc.h>
 
 #include "libelf.h"
 
@@ -99,6 +100,28 @@
  ****************************************************************************/
 
 /****************************************************************************
+ * Name: elf_readreltab
+ *
+ * Description:
+ *   Read the relocation table into memory.
+ *
+ ****************************************************************************/
+static inline void elf_readreltab(FAR struct elf_loadinfo_s *loadinfo, FAR const Elf32_Shdr *relsec)
+{
+	loadinfo->reltab = (uintptr_t)kmm_malloc(relsec->sh_size);
+
+	if (!loadinfo->reltab) {
+		berr("ERROR: Failed to allocate space for relocation table. Size = %u\n", relsec->sh_size);
+		return;
+	}
+
+	if (elf_read(loadinfo, (FAR uint8_t *)loadinfo->reltab, relsec->sh_size, relsec->sh_offset) < 0) {
+		berr("ERROR: Failed to read relocation table into memory\n");
+	}
+}
+
+
+/****************************************************************************
  * Name: elf_readrel
  *
  * Description:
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [b5eb9da29b1fd17cd1aff14f57eaeb051886d9dd] - Code Rule Check (C++) OK.

seinfra:
Target : [b5eb9da29b1fd17cd1aff14f57eaeb051886d9dd] - Code Rule Check OK.

os/binfmt/libelf/libelf_symbols.c
```diff
@@ -203,6 +203,32 @@ int elf_findsymtab(FAR struct elf_loadinfo_s *loadinfo)
...
+ * Name: elf_readsymtab
+ *
+ * Description:
+ *   Read the ELF symbol table into memory.
+ *
+ * Input Parameters:
+ *   loadinfo - Load state information
+ *
+ ****************************************************************************/
+void elf_readsymtab(FAR struct elf_loadinfo_s *loadinfo)
```
jeongchanKim: This is almost same as elf_readreltab. How about making common function for avoiding duplications? I mean not for whole code, just duplicated code.

kishore-sn: I dont think its a good idea to combine the two functions. Even if we combine them, most of the lines in the function need to be repeated according to the data structure to be used. For example, elf_readsym() and elf_readrel() are performing similar operation, but still they are maintained as separate API's.

sijupunnnoose: Better to keep two separate function

os/binfmt/libelf/libelf_symbols.c
```diff
@@ -232,13 +258,25 @@ int elf_readsym(FAR struct elf_loadinfo_s *loadinfo, int index, FAR Elf32_Sym *s
...
 
-	offset = symtab->sh_offset + sizeof(Elf32_Sym) * index;
+		offset = sizeof(Elf32_Sym) * index;
 
-	/* And, finally, read the symbol table entry into memory */
+		/* And, finally, read the symbol table entry into memory */
 
-	return elf_read(loadinfo, (FAR uint8_t *)sym, sizeof(Elf32_Sym), offset);
+		memcpy(sym, loadinfo->symtab + offset, sizeof(Elf32_Sym));
+		return 0;
```
sijupunnnoose: return OK

kishore-sn: ok

os/binfmt/libelf/libelf_bind.c
```diff
@@ -117,13 +139,24 @@ static inline int elf_readrel(FAR struct elf_loadinfo_s *loadinfo, FAR const Elf
...
+	if (loadinfo->reltab) {
+		/* Get the file offset to the symbol table entry */
+
+		offset = sizeof(Elf32_Rel) * index;
+
+		/* And, finally, read the symbol table entry into memory */
 
-	offset = relsec->sh_offset + sizeof(Elf32_Rel) * index;
+		memcpy(rel, loadinfo->reltab + offset, sizeof(Elf32_Rel));
+		return 0;
```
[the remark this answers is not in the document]
kishore-sn: ok

seinfra:
Target : [0d80d941c2bdcdb3f63b260974f883223d2082d2] - Code Rule Check (C++) OK.

seinfra:
Target : [0d80d941c2bdcdb3f63b260974f883223d2082d2] - Code Rule Check OK.

os/binfmt/libelf/libelf_bind.c
```diff
@@ -98,6 +98,28 @@
...
+/****************************************************************************
+ * Name: elf_readreltab
+ *
+ * Description:
+ *   Read the relocation table into memory.
+ *
+ ****************************************************************************/
+static inline void elf_readreltab(FAR struct elf_loadinfo_s *loadinfo, FAR const Elf32_Shdr *relsec)
+{
+	loadinfo->reltab = (uintptr_t)kmm_malloc(relsec->sh_size);
```
sunghan-chang: need to include header for kmm_malloc

kishore-sn: Ok. done.

sunghan-chang:
@kishore-sn Could you check my comment above?
```
Could you let me know the location of freeing symtab and reltab with normal operation? I don't mean freeing by error.
```

seinfra:
Target : [8c5fa3d0ca3bab5818f36877b0eb7069f702455e] - Code Rule Check (C++) OK.

kishore-sn:
@sunghan-chang The symtab and reltab are getting freed at the end of the API for both normal case and error case. In both cases, the return from the function happens from the end only.

PR #3960, opened 2019-10-16
os/tools: make global variable, line, static

sunghan-chang:
Files in the os/tools folder are used when TizenRT is built.
They are not running on a device. That's why there is no problem
to be static.


seinfra:
Target : [6ed5958b376dacf689426fdc5ad08d60758231bd] - Code Rule Check (C++) OK.

seinfra:
Target : [6ed5958b376dacf689426fdc5ad08d60758231bd] - Code Rule Check Result:
os/tools/cfgdefine.c:271: ERROR: [IDT_M_TAB] please, no spaces at the start of a line

PR #3977, opened 2019-10-24
os/drivers/scsc: Implement probe request to send target SSID

SeonghoByeon:
- Wildcard SSID in probe request forces for the surrounding APs to send probe response,
which is not efficient and makes it possible to miss the target AP's probe response when there are many near by AP with high traffic ratio.
This implementation inserts a specific SSID into probe requenst not wildcard SSID,
which reduces the scanning duration by an average value of 250 ms.


seinfra:
Target : [edb884814109a05ad1dfc587846ba33cfd0dc0f6] - Code Rule Check (C++) OK.

seinfra:
Target : [edb884814109a05ad1dfc587846ba33cfd0dc0f6] - Code Rule Check OK.

PR #3978, opened 2019-10-24
framework/araui: fix text widget issues

SRCN-LiQin:
1. fix memory overwrite issue when text size is too big, and also
the memset is unnecessary here.
2. fix building issue of ui_render_quad_uv function call
@gcjjyy


seinfra:
Target : [f6564fd835940270b8c5acf9d99458988064d8ea] - Code Rule Check (C++) OK.

PR #3980, opened 2019-10-25
os/wqueue: Adjust elapsed time when awakened earlier by a signal

jeongarmy:
That elapsed is relative to the the current time, not the time of beginning of this queue processing pass.
So it may need an adjustment when awakened earlier by a signal.


seinfra:
Target : [e493f88283c588e2b9e2e1ab0b3f63aeb9b62882] - Code Rule Check (C++) OK.

seinfra:
Target : [e493f88283c588e2b9e2e1ab0b3f63aeb9b62882] - Code Rule Check OK.

PR #3984, opened 2019-10-26
tools/ramdump: Fixed some possible incorrect comparison with null war…

dyyim741:
To ensure that PowerShell performs comparisons correctly, the $null element should be on the left side of the operator.

There are a number of reasons why this should occur:

$null is a scalar. When the input (left side) to an operator is a scalar value, comparison operators return a Boolean value. When the input is a collection of values, the comparison operators return any matching values, or an empty array if there are no matches in the collection. The only way to reliably check if a value is $null is to place $null on the left side of the operator so that a scalar comparison is performed.
PowerShell will perform type casting left to right, resulting in incorrect comparisons when $null is cast to other scalar types.

Ref:
https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation/PossibleIncorrectComparisonWithNull.md

### DIFF
--- a/tools/ramdump/Ramdump_windows.ps1
+++ b/tools/ramdump/Ramdump_windows.ps1
@@ -230,7 +230,7 @@ function ramdump_main {
 	catch [System.Exception] {
 		Write-Error ("Failed to connect : " + $_)
 		$error[0] | Format-List -Force
-		if ($port -ne $null) {
+		if ($null -ne $port) {
 			$port.Close()
 		}
 		exit 1
@@ -258,7 +258,7 @@ function ramdump_main {
 
 	Write-Output "`nRamdump received successfully..!"
 
-	if ($port -ne $null) {
+	if ($null -ne $port) {
 		$port.Close()
 	}
 }


seinfra:
Target : [219b4c1fd2986eb6715185f89b2f91054a62e5a8] - Code Rule Check OK.

seinfra:
Target : [219b4c1fd2986eb6715185f89b2f91054a62e5a8] - Code Rule Check (C++) OK.

sunghan-chang:
@dyyim741 Thank you for good PR. But could you use present tense for commit title?
```
fixed -> fix
```
Because it's our commit rules.

sunghan-chang:
And could you leave the commit description what you wrote in PR description?
```
tools/ramdump: Fix some possible incorrect comparison with null warnings

To ensure that PowerShell performs comparisons correctly, the $null element should be on the left side of the operator.

There are a number of reasons why this should occur:

$null is a scalar. When the input (left side) to an operator is a scalar value, comparison operators return a Boolean value. When the input is a collection of values, the comparison operators return any matching values, or an empty array if there are no matches in the collection. The only way to reliably check if a value is $null is to place $null on the left side of the operator so that a scalar comparison is performed.
PowerShell will perform type casting left to right, resulting in incorrect comparisons when $null is cast to other scalar types.

Ref:
https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation/PossibleIncorrectComparisonWithNull.md
```

sunghan-chang:
It makes TizenRT history better I think.

dyyim741:
@sunghan-chang Good comment.
I reflected your opinion.

seinfra:
Target : [41049876bf2057f26b6a733631862a73e1e76c6e] - Code Rule Check OK.

seinfra:
Target : [41049876bf2057f26b6a733631862a73e1e76c6e] - Code Rule Check (C++) OK.

PR #3992, opened 2019-10-29
kernel/wdog/wd_start.c : Remove unnecessary code

hasw7569:
If delay value is negative then wd_start return ERROR.
so we don't have to consider negative case.
also in case that ++delay is positive, we must decrease the delay value by one because we already increased delay value.
However I think this overflow check logic is not necessary.
so I remove whole elseif line.


seinfra:
Target : [46d513885545b767110544385823b0485f008767] - Code Rule Check (C++) OK.

seinfra:
Target : [46d513885545b767110544385823b0485f008767] - Code Rule Check OK.

jeongchanKim:
I think left code and right code is working different. Please check which is correct.

PR #4007, opened 2019-11-06
kernel/allocateheap : Remove unnecessary heap initialization checking

jeongchanKim:
We can check that heap is initialized or not with heap size. Because, if it is initialized, the size of its heap will not be zero.
So we don't need to use heapx_is_init array anymore.


seinfra:
Target : [cb3ec41a2ee400cb38bf42d1a02ab8013ad8c5fa] - Code Rule Check (C++) OK.

seinfra:
Target : [cb3ec41a2ee400cb38bf42d1a02ab8013ad8c5fa] - Code Rule Check OK.

PR #4023, opened 2019-11-12
build/tools: Update the Makefile of esptool.

xiarihf:
1. Add esptool_py submodule checking.
	2. Change some downloading peremeters to capital letter form.

Signed-off-by: xiarihf qiang3.zhang@samsung.com

### DIFF
--- a/build/tools/esp32/esptool_py/Makefile
+++ b/build/tools/esp32/esptool_py/Makefile
@@ -52,6 +52,16 @@ ESPTOOLPY_SRC := $(CURRENTDIR)/esptool_py/esptool/esptool.py
 ESPTOOLPY := $(PYTHON) $(ESPTOOLPY_SRC) --chip esp32
 ESPTOOLPY_SERIAL := $(ESPTOOLPY) --port $(ESPPORT) --baud $(ESPBAUD) --before $(CONFIG_ESPTOOLPY_BEFORE) --after $(CONFIG_ESPTOOLPY_AFTER)
 
+#Check submodule
+ESPTOOLPY_SUBMODULE_PATH := esptool_py/esptool
+ESPTOOLPY_SUBMODULE := esptool_modules
+
+$(ESPTOOLPY_SUBMODULE):
+ifneq ($(ESPTOOLPY_SRC), $(wildcard $(ESPTOOLPY_SRC)))
+	@echo "Get submodule $(ESPTOOLPY_SUBMODULE_PATH) firstly:"
+	git submodule update --init $(ESPTOOLPY_SUBMODULE_PATH)
+endif
+
 ESPTOOL_FLASH_OPTIONS := --flash_mode $(ESPFLASHMODE) --flash_freq $(ESPFLASHFREQ) --flash_size $(ESPFLASHSIZE)
 ifdef CONFIG_ESPTOOLPY_FLASHSIZE_DETECT
 ESPTOOL_WRITE_FLASH_OPTIONS := --flash_mode $(ESPFLASHMODE) --flash_freq $(ESPFLASHFREQ) --flash_size detect
@@ -68,7 +78,7 @@ ESPTOOL_ALL_FLASH_ARGS += $(APP_OFFSET) $(APP_BIN)
 # non-secure boot (or bootloader), both these files are the same
 APP_BIN_UNSIGNED ?= $(APP_BIN)
 
-$(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
+$(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SUBMODULE)
 	$(ESPTOOLPY) elf2image $(ESPTOOL_FLASH_OPTIONS) $(ESPTOOL_ELF2IMAGE_OPTIONS) -o $@ $<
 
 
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [350f5f3852bdc6af620be1e734abd54fb3181f1d] - Code Rule Check OK.

seinfra:
Target : [350f5f3852bdc6af620be1e734abd54fb3181f1d] - Code Rule Check (C++) OK.

build/tools/esp32/esptool_py/Makefile
```diff
@@ -74,19 +74,32 @@ $(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
...
-ROMFS:romfs
+OS: check-modules app-flash
+
+ROMFS: check-modules romfs
+
+BOOTLOADER: check-modules bootloader
+
+ERASE_ALL: check-modules erase_flash
+
+check-modules:
```
sunghan-chang: 1. Do we need to make a new target and need to call it at every target? If this is mandatory for all of activities, how about removing the target?
2. When esptool.py is not existed, how about executing pulling submodule automatically through this file?
  If it is not good, how about gibing guidance how to get submodule?

build/tools/esp32/esptool_py/Makefile
```diff
@@ -74,19 +74,32 @@ $(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
...
+ifeq ($(ESPTOOLPY_SRC), $(wildcard $(ESPTOOLPY_SRC)))
+	@echo "$(ESPTOOLPY_SRC) exist"
```
SRCN-LiQin: how about only echo when it's not exist.
and it's better to display the git command of getting submodule

build/tools/esp32/esptool_py/Makefile
```diff
@@ -74,19 +74,32 @@ $(APP_BIN_UNSIGNED): $(APP_ELF) $(ESPTOOLPY_SRC)
 
 ESPTOOL_ALL_FLASH_ARGS += $(ESPTOOL_BOOTLOADER_FLASH_ARGS)
 
-ALL: flash romfs
+ESPTOOLPY_SUBMODULE := esptool_py/esptool
```
zhouxinhe: Seemingly, this ESPTOOLPY_SUBMODULE is unnecessary, and it's not the exact submodule path of esptool.

build/tools/esp32/esptool_py/Makefile
xiarihf: Yes, good suggestion. The new target is not a good way. 
Thanks.

seinfra:
Target : [51ee5ce97d841dbd4eb2a0226be7dbf4014b889c] - Code Rule Check OK.

seinfra:
Target : [51ee5ce97d841dbd4eb2a0226be7dbf4014b889c] - Code Rule Check (C++) OK.

build/tools/esp32/esptool_py/Makefile
xiarihf: @sunghan-chang 
Dear Mr. Chang
  1.  I have checked the dependency of all the targets in this Makefile, and I think it's better to add a new target and make it to be the first dependency of all the targets, That's because all the target depends on this esptool.py file, so i change this dependency to the new target.
   2. The new target will get the esptool.py submodule automatically, and i have made the verification on both docker and local build. It works.

If you have any other suggestion, please inform me. Thanks.

build/tools/esp32/esptool_py/Makefile
xiarihf: yes, now it could get submodule automatically. 
Thanks for your suggestion.

build/tools/esp32/esptool_py/Makefile
xiarihf: Fixed, thanks.

seinfra:
Target : [1de3e87ce51af92544b8eda84327d63a46c52f69] - Code Rule Check OK.

seinfra:
Target : [1de3e87ce51af92544b8eda84327d63a46c52f69] - Code Rule Check (C++) OK.

seinfra:
Target : [8e0e90175c0afd4f0e4ab56faefbf627f03c8fe2] - Code Rule Check OK.

seinfra:
Target : [8e0e90175c0afd4f0e4ab56faefbf627f03c8fe2] - Code Rule Check (C++) OK.

sunghan-chang:
Actually I have no idea why new target is necessary. But it is not important. Let's merge.

build/tools/esp32/esptool_py/Makefile
xiarihf: @sunghan-chang  
I have tried to define a function, but i need ifeq to check the exist of  esptool.py, it would call a issue. Because function will be called in target content and a tab is needed before calling this function as:
romfs:
[tab]$(call function)
It will call a error due to ifeq needs no blank before it, but well have a tab before calling functions in a target.
if we remove the ifeq to each target, it will bring in too much duplicate code. And also if we add the ifeq checking in every target, it would bring in new redundant operations, because targets have their dependency and the dependency will be executed firstly. 

So, I think a new target is better in simple code and less operations.

PR #4024, opened 2019-11-12
build/configs: Add meta files to support esp32 boards on RT IDE

zhouxinhe:
To support build and flash project with Tizen Studio RT IDE on esp32 boards.
(co-work with PR #4023)

### DIFF
--- a/build/configs/esp32_DevKitC/.buildSpec.xml
+++ b/build/configs/esp32_DevKitC/.buildSpec.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2019 Samsung Electronics All Rights Reserved.
+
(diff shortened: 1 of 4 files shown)


seinfra:
Target : [9343847c2257e3e454845e9514bf3512cd5c9f07] - Code Rule Check OK.

seinfra:
Target : [9343847c2257e3e454845e9514bf3512cd5c9f07] - Code Rule Check (C++) OK.

sunghan-chang:
@zhouxinhe Regardless of this PR, is there any way to debug like gdb in ARM?

zhouxinhe:
> @zhouxinhe Regardless of this PR, is there any way to debug like gdb in ARM?

We can use gdb-debugging on esp32 boards, so it is possible.
Actually, I tried to do that, but it's difficult to use debug funtion of RT IDE for me now, so I can not verify it. I plan to enable it in future.

PR #4026, opened 2019-11-12
framework/preference: Support CRC to verify preference data

jeongarmy:
Caculate and Write CRC of them when setting preference data.
And then check it when getting preference data.

### DIFF
--- a/os/include/tinyara/preference.h
+++ b/os/include/tinyara/preference.h
@@ -55,6 +55,7 @@ enum preference_result_error_e {
 	PREFERENCE_OUT_OF_MEMORY = -5,
 	PREFERENCE_OPERATION_FAIL = -6,
 	PREFERENCE_NOT_REGISTERED = -7,
+	PREFERENCE_INVALID_DATA = -8,
 };
 
 #define PRIVATE_PREFERENCE     0
@@ -71,6 +72,7 @@ enum preference_result_error_e {
 #define PREFERNENCE_CBMSG_MAX       32
 
 struct value_attr_s {
+	uint32_t crc;
 	int type;
 	int len;
 };
--- a/os/include/tinyara/preference.h
+++ b/os/include/tinyara/preference.h
@@ -70,21 +70,17 @@ enum preference_result_error_e {
 #define PREFERENCE_CBMQ_LEN         16
 #define PREFERNENCE_CBMSG_MAX       32
 
-struct value_data_s {
+struct value_attr_s {
 	int type;
 	int len;
-	union {
-		int i;
-		double d;
-		char *s;
-	};
 };
-typedef struct value_data_s value_data_t;
+typedef struct value_attr_s value_attr_t;
 
 struct preference_data_s {
 	char *key;
 	int type;
-	value_data_t value;
+	value_attr_t attr;
+	void *value;
 };
 typedef struct preference_data_s preference_data_t;
 
--- a/os/kernel/preference/preference_read.c
+++ b/os/kernel/preference/preference_read.c
@@ -22,6 +22,7 @@
 #include <debug.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <crc32.h>
 #include <tinyara/preference.h>
 
 /****************************************************************************
@@ -31,6 +32,7 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
 {
 	int fd;
 	int ret;
+	uint32_t check_crc;
 	value_attr_t attr;
 
 	fd = open(path, O_RDONLY, 0666);
@@ -43,7 +45,7 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
 		return PREFERENCE_IO_ERROR;
 	}
 
-	/* Read and Verify attributes of data : type, len */
+	/* Read and Verify attributes of data : crc, type, len */
 	ret = read(fd, (FAR uint8_t *)&attr, sizeof(value_attr_t));
 	if (ret != sizeof(value_attr_t)) {
 		prefdbg("Failed to read attribute, errno %d\n", errno);
@@ -70,6 +72,15 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
 	if (ret != data->attr.len) {
 		prefdbg("Failed to read key value, errno %d\n", errno);
 		ret = PREFERENCE_IO_ERROR;
+		goto errout_with_free;
+	}
+
+	/* Calculate and Verify the checksum */
+	check_crc = crc32((uint8_t *)&data->attr.type, sizeof(value_attr_t) - sizeof(uint32_t));
+	check_crc = crc32part((uint8_t *)data->value, data->attr.len, check_crc);
+	if (check_crc != attr.crc) {
+		prefdbg("Invalid checksum, read crc : %u, calculated crc : %u\n", attr.crc, check_crc);
+		ret = PREFERENCE_INVALID_DATA;
 		goto errout_with_free;
 	}
 
--- a/os/kernel/preference/preference_read.c
+++ b/os/kernel/preference/preference_read.c
@@ -31,8 +31,7 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
 {
 	int fd;
 	int ret;
-	int read_size;
-	value_data_t value_data;
+	value_attr_t attr;
 
 	fd = open(path, O_RDONLY, 0666);
 	PREFERENCE_FREE(path);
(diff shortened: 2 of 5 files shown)


seinfra:
Target : [bcb93357c5091c0d52f32fa1d45fa81f0952e9ff] - Code Rule Check OK.

seinfra:
Target : [bcb93357c5091c0d52f32fa1d45fa81f0952e9ff] - Code Rule Check (C++) OK.

sunghan-chang:
@jeongarmy Could you split the changes for adding CRC and others (refactoring)?

os/kernel/preference/preference_read.c
```diff
@@ -44,59 +46,62 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
...
 	}
 
-	/* Read and Verify the type and len of data value */
-	read_size = 2 * sizeof(int);
-	ret = read(fd, (void *)&value_data, read_size);
-	if (ret != read_size) {
+	check_crc = 0;
+
+	/* Read and Verify attributes of data : type, len */
+	ret = read(fd, (FAR uint8_t *)&attr, sizeof(value_attr_t));
```
Taejun-Kwon: Below is simple logic & struct to handle read. You may understand write case after read this.
```
struct preference_entry_s {
    char *key;
    uint8_t type;
    uint32_t datalen; //is it enough size??
    uint32_t crc32;
}

# case read
int fd;
int ret;
struct preference_entry_s *entry;

fd = open(path, O_RDONLY); //0666 is not necessary, actually.
if (fd < 0) {
    dbg("error, errno : %d\n", errno);
    return error;
}
/* Verify Entry first */
int size = sizeof(struct preference_entry_s);
ret = read(fd, entry, size); 

if (ret < size) {
   dbg("read error, ret : %d errno : %d\n", ret, errno);
    return error;
}

if (type_from_user != entry->type) {
    dbg("error....blah blah\n");
    return error;
}
... //check entry here.

/* Now read data */
char *buf = malloc(entry->datalen);
ret = read(fd, buf, entry->datalen);
if (ret != entry->datalen) {
   blahblah
}

/* Check crc by compare crc value in entry and calculated read data */
if (entry->crc32 != crc32part()) {
   dbg("error");
   return error;
}
```

Anyway, now you calling read 3 times here, Please make sure that  I/O  always takes more much of time than any other operation.

seinfra:
Target : [f2df6c9481342c7c16cf1ec52aeeffa61a315fc7] - Code Rule Check OK.

seinfra:
Target : [f2df6c9481342c7c16cf1ec52aeeffa61a315fc7] - Code Rule Check (C++) OK.

seinfra:
Target : [1cedfbec2986480d101f77ee56cf7d0a068e2597] - Code Rule Check OK.

seinfra:
Target : [1cedfbec2986480d101f77ee56cf7d0a068e2597] - Code Rule Check (C++) OK.

os/include/tinyara/preference.h
```diff
@@ -70,21 +71,18 @@ enum preference_result_error_e {
 #define PREFERENCE_CBMQ_LEN         16
 #define PREFERNENCE_CBMSG_MAX       32
 
-struct value_data_s {
+struct value_attr_s {
```
Taejun-Kwon: In my opinion something should be changed,
1. I think this struct is not necessary.
 You have to define APIs first, for example,
 - pref_create_subkey(char *subkey) // Create directory for key. c.f /mnt/prefkey/appA 
 - pref_open_key(char* key);
 - pref_get_key_value(char* key, char *buf, ssize_t len);
 ....
 Then user will use these apis without this struct.
 also for file system, only key & length is needed to get whole of key value.
 If there's any reason to keep any of structure, then provide handler which is hide from user.
2. Will you open whole of structure & definition to user? I think many things here, can be hided from user.

os/kernel/preference/preference_read.c
```diff
@@ -44,59 +45,54 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
...
+	/* Read value data */
+	ret = read(fd, (void *)data->value, data->attr.len);
+	if (ret != data->attr.len) {
+		prefdbg("Failed to read key value, errno %d\n", errno);
 		ret = PREFERENCE_IO_ERROR;
+		goto errout_with_free;
+	}
+
+	/* Caculate and Verify the checksum */
+	check_crc = crc32part((uint8_t *)&data->attr.type, sizeof(value_attr_t) - sizeof(uint32_t), check_crc);
```
Taejun-Kwon: Below is more looks good.
uint32_t crc;
crc = crc32((uint8_t *)&data->attr.type, sizeof(value_attr_t) - sizeof(uint32_t));
crc = crc32part((uint8_t *)data->value, data->attr.len, crc);

os/kernel/preference/preference_read.c
```diff
@@ -31,8 +32,8 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
 {
 	int fd;
 	int ret;
-	int read_size;
-	value_data_t value_data;
+	uint32_t check_crc;
```
Taejun-Kwon: preference_read_fs_key is not good naming.
How about align with tizen's api?

os/include/tinyara/preference.h
jeongarmy: 1. I understand what you consider. I will move preference operations to driver to support them in kernel as you said. And then it would be implemented with general usages like buf, len.

2. I think those are needed because user can't call preference functions directly. If not, all functions should be registered to syscall or user needs to pass all parameters to functions respectively.

os/kernel/preference/preference_read.c
jeongarmy: Modified

seinfra:
Target : [db10ec07f5b8657120e79dc750d8367282fe2cdd] - Code Rule Check OK.

seinfra:
Target : [db10ec07f5b8657120e79dc750d8367282fe2cdd] - Code Rule Check (C++) OK.

os/kernel/preference/preference_read.c
jeongarmy: The name of internal functions are _read_data and _write_data in Tizen. I think it is different from our naming rules for internal functions. I will think more about function name.

os/include/tinyara/preference.h
```diff
@@ -70,21 +70,17 @@ enum preference_result_error_e {
...
 };
-typedef struct value_data_s value_data_t;
+typedef struct value_attr_s value_attr_t;
 
 struct preference_data_s {
 	char *key;
 	int type;
-	value_data_t value;
+	value_attr_t attr;
+	void *value;
```
junmin-kim: If `preference_data_t` has `void *value` and `value` is allocated during `preference_read_fs_key`,
user have to free `value` to avoid memory leak.

User will free `preference_data_t`, but will they free `value`?
I suggest that preference supports destroy function which frees `value` and `preference_data_t`.
How do you think?

jeongarmy: Result values are freed except string after assigning them to user variables.
For example, private_preference.c:173
  *value = *(double *)data.value;
  PREFERENCE_FREE(data.value);
So user need to free string value only.

User just calls preference APIs with key, value data and they are packed as preference_data_t and processed internally.
It means user don't use preference_data_t directly so free functions is not needed.

os/include/tinyara/preference.h
jeongarmy: 1. Sorry, I was confused. I think driver is not appropriate for preference.
I will update preference functions for kernel and general usage as you said.

os/kernel/preference/preference_read.c
```diff
@@ -44,59 +45,52 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
...
 		ret = PREFERENCE_IO_ERROR;
+		goto errout_with_free;
 	}
-errout:	
+
+	/* Caculate and Verify the checksum */
+	check_crc = crc32((uint8_t *)&data->attr.type, sizeof(value_attr_t) - sizeof(uint32_t));
+	check_crc = crc32part((uint8_t *)data->value, data->attr.len, check_crc);
+	if (check_crc != attr.crc) {
+		prefdbg("Invalid checksum, read crc : %u, calcuated crc : %u\n", attr.crc, check_crc);
```
jeongchanKim: calcuated -> calculated

os/kernel/preference/preference_read.c
```diff
@@ -44,59 +45,52 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
...
 		goto errout;
-	} else if (value_data.type < 0 || value_data.len < 0) {
-		prefdbg("Invalid data %d\n", value_data.len);
+	} else if (attr.type < 0 || attr.len < 0) {
+		prefdbg("Invalid data : type %d, len %d\n", attr.type, attr.len);
 		ret = PREFERENCE_OPERATION_FAIL;
 		goto errout;
-	} else if (value_data.type != data->value.type) {
-		prefdbg("Invalid type. request type:%d, read type:%d\n", data->value.type, value_data.type);
+	} else if (attr.type != data->attr.type) {
```
jeongchanKim: `data` can be dereferencing.

os/kernel/preference/preference_write.c
```diff
@@ -122,26 +123,26 @@ static int preference_write_fs_key(char *path, preference_data_t *data)
...
-		/* Write type, len */
-		write_size = 2 * sizeof(int);
-		ret = write(fd, (void *)&(data->value), write_size);
-		if (ret == write_size) {
-			/* Write value */
-			write_size += data->value.len;
-			ret += write(fd, (void *)data->value.s, data->value.len);
-		}
+	/* Calculate checksum of attributes, type, len and value */
+	crc_value = crc32((uint8_t *)&data->attr.type, sizeof(value_attr_t) - sizeof(uint32_t));
```
jeongchanKim: Same as above.

os/kernel/preference/preference_write.c
```diff
@@ -122,26 +123,26 @@ static int preference_write_fs_key(char *path, preference_data_t *data)
...
+	ret = write(fd, (void *)&data->attr, sizeof(value_attr_t));
+	if (ret != sizeof(value_attr_t)) {
+		prefdbg("Failed to write key value, errno %d\n", errno);
+		goto errout_with_close;
 	}
-	if (ret != write_size) {
-		ret = PREFERENCE_IO_ERROR;
+
+	/* Write value data */
+	ret = write(fd, (void *)data->value, data->attr.len);
```
Taejun-Kwon: You already checked validation and everything calculated.
Then you can write all just once. Internally, calling write twice is take bunch of processes.

jeongarmy: I called write function twice to avoid allocating buffer as a size of whole data and copying them because actual value is in another address, not continuous address region with another data.

os/kernel/preference/preference_write.c
jeongarmy: Already checked by above functions

os/kernel/preference/preference_read.c
jeongarmy: Already checked by above functions

seinfra:
Target : [64ef4252a4047942bf3e8db6bfe830e51eb31f43] - Code Rule Check OK.

seinfra:
Target : [64ef4252a4047942bf3e8db6bfe830e51eb31f43] - Code Rule Check (C++) OK.

os/kernel/preference/preference_read.c
```diff
@@ -44,59 +45,52 @@ static int preference_read_fs_key(char *path, preference_data_t *data)
...
+	/* Read value data */
+	ret = read(fd, (void *)data->value, data->attr.len);
+	if (ret != data->attr.len) {
+		prefdbg("Failed to read key value, errno %d\n", errno);
 		ret = PREFERENCE_IO_ERROR;
+		goto errout_with_free;
 	}
-errout:	
+
+	/* Caculate and Verify the checksum */
```
jeongchanKim: Caculate -> Calculate

seinfra:
Target : [4b110e7e2a3efc7b5ba55bb35236a7796d7bab23] - Code Rule Check OK.

seinfra:
Target : [4b110e7e2a3efc7b5ba55bb35236a7796d7bab23] - Code Rule Check (C++) OK.

PR #4035, opened 2019-11-13
kernel/semaphore: Remove unnecessary variable

jeongarmy:
Call directly without assigning rtcb variable because it is used when enabling some configurations only.


seinfra:
Target : [19ca10ef7e3f97799bd6b5d443650293c3ca445c] - Code Rule Check (C++) OK.

seinfra:
Target : [19ca10ef7e3f97799bd6b5d443650293c3ca445c] - Code Rule Check OK.

PR #4043, opened 2019-11-14
external/openssl: add deppath for subdir

SRCN-LiQin:
1. Add deppath to make sure making dependency normally
2. redefine mbedtls user config macro to avoid mkdep failure because
of unrecognize definition in makefile
there is "#include expects "FILENAME" or <FILENAME>" issue if define the macro in makefile directly


seinfra:
Target : [514ec31f78ac693ac4e2f597054394ea82404bb8] - Code Rule Check OK.

seinfra:
Target : [514ec31f78ac693ac4e2f597054394ea82404bb8] - Code Rule Check (C++) OK.

build/configs/esp_wrover_kit/STDK/defconfig
```diff
@@ -1145,6 +1145,7 @@ CONFIG_STDK_MQTT_USE_SSL=y
 CONFIG_STDK_IOT_CORE_USE_MBEDTLS=y
 # CONFIG_STDK_IOT_CORE_CRYPTO_SUPPORT_RSA is not set
 CONFIG_STDK_IOT_CORE_CRYPTO_SUPPORT_ED25519=y
+CONFIG_MBEDTLS_USER_FILE="/root/tizenrt/external/stdk/tizenrt-wrapper/mbedtls_config.h"
 # CONFIG_STDK_IOT_CORE_FS_SW_ENCRYPTION is not set
```
ChinaTizenRT: how this subpath "/root/tizenrt" works?

build/configs/artik053/STDK/defconfig
```diff
@@ -1246,6 +1243,7 @@ CONFIG_STDK_MQTT_USE_SSL=y
 CONFIG_STDK_IOT_CORE_USE_MBEDTLS=y
 # CONFIG_STDK_IOT_CORE_CRYPTO_SUPPORT_RSA is not set
 CONFIG_STDK_IOT_CORE_CRYPTO_SUPPORT_ED25519=y
+CONFIG_MBEDTLS_USER_FILE="/root/tizenrt/external/stdk/tizenrt-wrapper/mbedtls_config.h"
```
zhouxinhe: What would happen if we are not using docker?
Is it possible to use relative path here?

build/configs/esp_wrover_kit/STDK/defconfig
SRCN-LiQin: it's the absolute path in docker env.
as we use this Kconfig string type, which always have "", so we can't use a global path with <>, and could only use such kind of absolute full path.

build/configs/artik053/STDK/defconfig
SRCN-LiQin: as we use this Kconfig string type, which always have "", so we can't use a global path with <>, and could only use such kind of absolute full path.
if we use normal make command, this path need to be set as the absolute path in build env. , maybe we can add such description in Kconfig help

seinfra:
Target : [3ed5048c50a66bba2d1fae45960df8f5b0ed75f2] - Code Rule Check OK.

seinfra:
Target : [3ed5048c50a66bba2d1fae45960df8f5b0ed75f2] - Code Rule Check (C++) OK.

build/configs/artik053/STDK/defconfig
sunghan-chang: I think relative path is better and there is no problem.

seinfra:
Target : [3b8a85d1e2071ec0db150215dbefe4b11c4e8aea] - Code Rule Check OK.

seinfra:
Target : [3b8a85d1e2071ec0db150215dbefe4b11c4e8aea] - Code Rule Check (C++) OK.

build/configs/artik053/STDK/defconfig
SRCN-LiQin: updated and verified with relative path, misundertood here before, thanks to you all

seinfra:
Target : [3726f3409456f31faaac9c3ce73fe5076efa1c69] - Code Rule Check OK.

seinfra:
Target : [3726f3409456f31faaac9c3ce73fe5076efa1c69] - Code Rule Check (C++) OK.

external/stdk/Kconfig
```diff
@@ -116,6 +116,12 @@ config STDK_IOT_CORE_CRYPTO_SUPPORT_ED25519
 	---help---
 		Support Ed25519 based sign, encryption functions.
 
+config MBEDTLS_USER_FILE
```
sunghan-chang: This config is used in mbedtls, but definition is in stdk. It loosk strange.

build/configs/artik053/STDK/Make.defs
```diff
@@ -120,7 +120,7 @@ ARCHDEFINES =
 ARCHPICFLAGS = -fpic -msingle-pic-base -mpic-register=r10
 
 CFLAGS = $(ARCHCFLAGS) $(ARCHWARNINGS) $(ARCHOPTIMIZATION) $(ARCHCPUFLAGS) $(ARCHINCLUDES) $(ARCHDEFINES) $(EXTRADEFINES) -pipe -ffunction-sections -fdata-sections
-CFLAGS += -DMBEDTLS_USER_CONFIG_FILE=\"$(TOPDIR)/../external/stdk/tizenrt-wrapper/mbedtls_config.h\"
+CFLAGS += -DMBEDTLS_USER_CONFIG_FILE=CONFIG_MBEDTLS_USER_FILE
```
sunghan-chang: This is used only in mbedtls. Could you let me know why you added it in top of Make.defs?

external/stdk/Kconfig
sunghan-chang: Except the change in `external/openssl/Makefile`, it is not `add deppath for subdir`.

SRCN-LiQin: for 1# the reason is that it's the USER CONFIG for STDK module, which needs to use mbedtls.
for 2#, here is the consequent modification for adding deppath for openssl, because actually openssl is a wrapper impl. based on mbedtls, so mbedtls/config.h is also included in openssl, and during make depend for openssl, it needs to parse this including file path.

build/configs/artik053/STDK/Make.defs
SRCN-LiQin: because it's an customized config for stdk module, so I think this compling variable definition should be put in the build config for stdk. can you share your opinion? thanks

SRCN-LiQin: and because of it's a customized config for stdk, I thought it could be added in makefile of external/stdk , or in this top of Make.defs,  and as you known this config.h is include in mbedtls and openssl, so there will be error if add it in external/stdk, then I put it here in this Make.defs

external/stdk/Kconfig
SRCN-LiQin: how about make 2 commits in this PR ? each for deppath and mbedtls user config seperately

build/configs/artik053/STDK/Make.defs
sunghan-chang: It's just preset configuration like example. To use stdk, configuring `STDK` is not mandatory. We should make it enable to set from another defconfig. In architecture point of view, related definition or module should be located in its module. This CFLAGS should be in mbedtls.

external/stdk/Kconfig
ChinaTizenRT: I search the whole TizenRT, no other module need to tune mbedtls  default configuration when using mbedtls. so it is new situation for us.

generally, we prefer to modify our module(STDK) instead of existing module (like medtls). 
but as you said, it looks strange that STDK module includes other modules configuration and Kconfig(MBEDTLS_USER_FILE) name.

so I suggest directly add Kconfig and customed config file( in external/include/mbedtls)
in  mbedtls module,  anyway, it is new issue for us, it need to fix reasonablely.

SRCN-LiQin:
Alternaltive solution with PR 4061 has been merged, close this PR.

PR #4066, opened 2019-11-21
external/stdk: refine obtain time check condition

SRCN-LiQin:
1. remove unnecessary check condition to _obtain_time after set STA mode
   (some board may have future time as default system time)
2. refine the hard code of 2016-1900 year check condition

### DIFF
--- a/external/stdk/tizenrt-wrapper/bsp/iot_bsp_wifi_common.c
+++ b/external/stdk/tizenrt-wrapper/bsp/iot_bsp_wifi_common.c
@@ -19,7 +19,7 @@
 #include <tinyara/config.h>
 #include <stdlib.h>
 #include <string.h>
-#include <time.h>
+#include <tinyara/time.h>
 #include <wifi_manager/wifi_manager.h>
 #include "iot_debug.h"
 #include "iot_bsp_wifi.h"
@@ -108,8 +108,8 @@ static void _obtain_time(void)
 		return;
 	}
 
-	// wait for time to be set
-	while (timeinfo.tm_year < (2016 - 1900) && ++retry < retry_count) {
+	// wait for time to be set, system time initialized in _initialize_ntp
+	while (timeinfo.tm_year <= (EPOCH_YEAR - TM_YEAR_BASE) && ++retry < retry_count) {
 		IOT_INFO("Waiting for system time to be set... (%d/%d)", retry, retry_count);
 		IOT_DELAY(3000);
 		time(&now);
@@ -226,8 +226,6 @@ int _iot_bsp_sta_mode(void)
 iot_error_t iot_bsp_wifi_set_mode(iot_wifi_conf *conf)
 {
 	int str_len = 0;
-	time_t now;
-	struct tm timeinfo;
 	wifi_manager_result_e res;
 	wifi_manager_softap_config_s sap_config;
 	wifi_manager_ap_config_s apconfig;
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [3ab746054017d8f9257594d453eb1ec33affec5e] - Code Rule Check OK.

seinfra:
Target : [3ab746054017d8f9257594d453eb1ec33affec5e] - Code Rule Check (C++) OK.

external/stdk/Kconfig
```diff
@@ -15,6 +15,12 @@ config STDK_IOT_CORE
 
 if STDK_IOT_CORE
 
+config STDK_CURRENT_TIME_YEAR
```
xixidodo: This works, but a bit confusing.
As discussed just now, please discuss with @SeonghoByeon on new solution.

external/stdk/tizenrt-wrapper/bsp/iot_bsp_wifi_common.c
```diff
@@ -109,7 +110,7 @@ static void _obtain_time(void)
 	}
 
 	// wait for time to be set
-	while (timeinfo.tm_year < (2016 - 1900) && ++retry < retry_count) {
+	while (timeinfo.tm_year < byear && ++retry < retry_count) {
```
zhouxinhe: The time has been reset to year 1970  when starting NTP (in function _initialize_ntp), so we can check condition `timeinfo.tm_year > TM_YEAR_BASE` to judge if the time is synchronized or not.

external/stdk/Kconfig
```diff
@@ -15,6 +15,12 @@ config STDK_IOT_CORE
...
+	int "Current year setting"
+	default 2019
+	---help---
+		Current year used to check if system time is updated by ntp client.
```
ChinaTizenRT: this patch just work for now or for future?

external/stdk/tizenrt-wrapper/bsp/iot_bsp_wifi_common.c
zhouxinhe: Maybe the time should be saved before reset, and restored in case of NTP syncing failed...
Please consider this.

external/stdk/Kconfig
SRCN-LiQin: yes, this Kconfig also seems weired, I will remove this and try to check the time with initialized value.

external/stdk/tizenrt-wrapper/bsp/iot_bsp_wifi_common.c
SRCN-LiQin: time is reset to 1970, and TM_BASE_YEAR is 1900, here let's check tm_year <= (EPOCH_YEAR - TM_YEAR_BASE) for invalid time and delay to wait for system time set

external/stdk/Kconfig
SRCN-LiQin: this value could work even for future, but as discussed above, this kconfig also seems weired, I will remove this , and apply different solution for time delay

seinfra:
Target : [5637643fa88d47e2b56bd2294b6d67436805be8d] - Code Rule Check (C++) OK.

seinfra:
Target : [5637643fa88d47e2b56bd2294b6d67436805be8d] - Code Rule Check OK.

PR #4091, opened 2019-12-05
Add external ramdump support for imxrt1050 board

thapav:
This patch adds external ramdump to existing ramdump tool for imxrt1050 board

### DIFF
--- a/os/board/common/crashdump.c
+++ b/os/board/common/crashdump.c
@@ -105,28 +105,67 @@ static int ramdump_via_uart(void)
 	/* Send number of memory regions to HOST */
 	up_lowputc(CONFIG_MM_REGIONS);
 
+#if defined(CONFIG_MM_KERNEL_HEAP)
+	/* Send 1 to HOST if kernel heap exists */
+	up_lowputc('1');
+
+	/* Send number of memory regions to HOST */
+	up_lowputc(CONFIG_KMM_REGIONS);
+
+	/* Send number of memory regions to HOST */
+	up_lowputc((CONFIG_MM_REGIONS + CONFIG_KMM_REGIONS));
+
+	/* Send memory region address, size & heap index to HOST */
+	for (x = 0; x < (CONFIG_MM_REGIONS + CONFIG_KMM_REGIONS); x++) {
+#else
 	/* Send memory region address, size & heap index to HOST */
 	for (x = 0; x < CONFIG_MM_REGIONS; x++) {
+#endif
 
-		/* Send RAM address */
-		ptr = (uint8_t *)&regionx_start[x];
-		for (i = 0; i < sizeof(regionx_start[x]); i++) {
-			up_lowputc((uint8_t)*ptr);
-			ptr++;
+		if (x < CONFIG_MM_REGIONS) {
+
+			/* Send RAM address */
+			ptr = (uint8_t *)&regionx_start[x];
+			for (i = 0; i < sizeof(regionx_start[x]); i++) {
+				up_lowputc((uint8_t)*ptr);
+				ptr++;
+			}
+
+			/* Send RAM size */
+			ptr = (uint8_t *)&regionx_size[x];
+			for (i = 0; i < sizeof(regionx_size[x]); i++) {
+				up_lowputc((uint8_t)*ptr);
+				ptr++;
+			}
+
+			/* Send Heap Index */
+			up_lowputc(regionx_heap_idx[x]);
 		}
+#if defined(CONFIG_MM_KERNEL_HEAP)
+		else if (x >= CONFIG_MM_REGIONS) {
+			int y =  x - CONFIG_MM_REGIONS;
 
-		/* Send RAM size */
-		ptr = (uint8_t *)&regionx_size[x];
-		for (i = 0; i < sizeof(regionx_size[x]); i++) {
-			up_lowputc((uint8_t)*ptr);
-			ptr++;
+			/* Send Kernel region address */
+			ptr = (uint8_t *)&kregionx_start[y];
+			for (i = 0; i < sizeof(kregionx_start[y]); i++) {
+				up_lowputc((uint8_t)*ptr);
+				ptr++;
+			}
+
+			/* Send Kernel region size */
+			ptr = (uint8_t *)&kregionx_size[y];
+			for (i = 0; i < sizeof(kregionx_size[y]); i++) {
+				up_lowputc((uint8_t)*ptr);
+				ptr++;
+			}
+
+			/* Send Kernel Heap Index */
+			up_lowputc(regionx_kheap_idx[y]);
 		}
-
-		/* Send Heap Index */
-		up_lowputc(regionx_heap_idx[x]);
+#endif
 	}
 
-	/* Receive number of regions to be dumped from HOST */
+	/* Receive number of user memory regions to be dumped from HOST */
 	do {
 		if ((ch = up_getc()) != -1) {
 			host_reg[0] = ch;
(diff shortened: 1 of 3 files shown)


seinfra:
Target : [03f1ffc547ba22c274ce2d0e2ba16aee2d0d4103] - Code Rule Check OK.

seinfra:
Target : [03f1ffc547ba22c274ce2d0e2ba16aee2d0d4103] - Code Rule Check (C++) OK.

os/board/imxrt1050-evk/src/Makefile
```diff
@@ -88,7 +88,7 @@ CSRCS += partitions.c
 endif
 
 ifeq ($(CONFIG_BOARD_CRASHDUMP),y)
-CSRCS += crashdump.c
+CSRCS += imxrt_crashdump.c
```
sangam-swami: why have to move from common crash dump to board specific?

thapav: All boards might not support an external ram which needs to be dumped. Also we have used board specific configs for external ram

seinfra:
Target : [8635ba40fc5e94e20dbabce1e199b28cc5d78697] - Code Rule Check (C++) OK.

seinfra:
Target : [8635ba40fc5e94e20dbabce1e199b28cc5d78697] - Code Rule Check OK.

os/board/imxrt1050-evk/src/Makefile
sangam-swami: It was moved to common for making generic implementation.
(CONFIG_IMXRT_SEMC_SDRAM), CONFIG_IMXRT_SDRAM_START, CONFIG_IMXRT_SDRAM_SIZE
I see these configs are imxrt specific, can we rename or find method to make it common name and use by all boards?

seinfra:
Target : [1b9cc6b02392ac92a1c964da5275d6f222bac7ce] - Code Rule Check Result:
tools/ramdump/ramdump_tool.c:306: ERROR: [IDT_M_TAB] code indent should use tabs where possible
tools/ramdump/ramdump_tool.c:310: ERROR: [SPC_M_KWD] space required before the open parenthesis '('

seinfra:
Target : [1b9cc6b02392ac92a1c964da5275d6f222bac7ce] - Code Rule Check (C++) OK.

os/board/imxrt1050-evk/src/Makefile
sunghan-chang: As @sangam-swami said, please make it as common.

thapav: Updated in common crashdump file itself using CONFIG_IMXRT_SEMC_SDRAM

seinfra:
Target : [f3ad96113ec08456ee7609ceeab5de5e76b825a0] - Code Rule Check (C++) OK.

seinfra:
Target : [f3ad96113ec08456ee7609ceeab5de5e76b825a0] - Code Rule Check OK.

thapav:
Dump with internal   external RAM

thapav:
Dump with only internal RAM

thapav:
Artik case

sunghan-chang:
@thapav Thank you for updating. But we need to clean our source codes.

1. This is not common codes, it's imxrt specific. To make it common, you need to remove imxrt content in crashdump.c and then you should call common function with imxrt values from imxrt code.
2. Regardless of crashdump, I want to tidy up the heap management.
Currently we have two parts of heap management, one is os_start and allocheap for flat and protected build, another is mm_partition_mgr for app bin separation. But I can't find any reason why we keep the mm_partition_mgr additionally. We must avoid the confusion of heap usages so that I think we should remove mm_partition_mgr and merge it into original. Then, this PR is unnecessary. 

@kishore-sn @jeongchanKim Could you check my 2nd opinion?

sunghan-chang:
@thapav @kishore-sn was modifying imxrt heap management so that we don't need to merge this. Instead of this change, could you add dumpping the kmm_heap_regionx?

sunghan-chang:
@thapav I merged @kishore-sn 's PR which removes mm_partition_mgr.c. Please modify this PR as well.

thapav:
@sunghan-chang Sure i will do that

seinfra:
Target : [4fa776ffdab3e1788a7836ab2b4bee414279b3c5] - Code Rule Check Result:
os/board/common/crashdump.c:118: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/board/common/crashdump.c:138: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/board/common/crashdump.c:181: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/board/common/crashdump.c:193: ERROR: [SPC_M_KWD] space required before the open parenthesis '('

thapav:
@sunghan-chang @sijupunnnoose  I have updated the chnages as per KMM regions.
Kindly review the same.

Dump all
Dump kernel MM
Dump User MM

seinfra:
Target : [4fa776ffdab3e1788a7836ab2b4bee414279b3c5] - Code Rule Check (C++) OK.

seinfra:
Target : [bd9c3e34ec2affe4d2916627226fdc68564a8a5c] - Code Rule Check OK.

seinfra:
Target : [bd9c3e34ec2affe4d2916627226fdc68564a8a5c] - Code Rule Check (C++) OK.

os/board/common/crashdump.c
```diff
@@ -201,7 +249,7 @@ void board_crashdump(uint32_t cur_sp, void *tcb, uint8_t *filename, int lineno)
 	 * watchdog reset might stall the system, causing crashdump hang.
 	 * Hence let's disable watchdog reset during crashdump process.
 	 */
-	up_watchdog_disable();
+//	up_watchdog_disable();
```
sunghan-chang: Could you let me know the reason of commenting out?

os/board/common/crashdump.c
```diff
@@ -105,28 +106,58 @@ static int ramdump_via_uart(void)
...
+			up_lowputc(regionx_heap_idx[x]);
 		}
 
-		/* Send Heap Index */
-		up_lowputc(regionx_heap_idx[x]);
+		else if (x >= CONFIG_MM_REGIONS) {
+			y =  x - CONFIG_MM_REGIONS;
+
+			/* Send Kernel region address */
+			ptr = (uint8_t *)&kregionx_start[y];
```
sunghan-chang: kregionx_start is valid with CONFIG_MM_KERNEL_HEAP enabled. But here is no conditional for that. Could you confirm this is working on CONFIG_MM_KERNEL_HEAP disabled like ARTIK?

sunghan-chang:
@thapav In screen shot, 
```
Please enter desired ramdump option as below:
    1 for ALL
    2 for Region 0
    25 for Region 0 & 3
Please enter your input
```
What is the `25`? Isn't it `3`? What's the meaning of `Region 0 & 3`?

thapav:
> What is the `25`? Isn't it `3`? What's the meaning of `Region 0 & 3`?
> I think below is correct.
> 
> ```
>    1 for ALL
>    2 for Region 0
>    3 for Region 1
> ```

We can dump multiple regions also at once instead of always dumping all regions.  This option provides that feature. 
For example in the below case to dump region 0 & 3 user has to input value 25
>    1 for ALL
>    2 for Region 0
>    3 for Region 1
>    4 for Region 2
>    5 for Region 3

os/board/common/crashdump.c
thapav: You're right, CONFIG_MM_KERNEL_HEAP dependency needs to be added s build breaks with protected kernel heap config disabled. I will add it

os/board/common/crashdump.c
thapav: This API is defined under the config CONFIG_WATCHDOG_FOR_IRQ (default n when CONFIG_WATCHDOG is enabled) in file  "arch/arm/src/imxrt/**imxrt_wdog**.c", whereas this API in crashdump file is using CONFIG_WATCHDOG conditional compilation only.

Should I add CONFIG_WATCHDOG_FOR_IRQ check as well for it ??

seinfra:
Target : [87bff356674e4adb5f9c188ab1d7bc457988ba1f] - Code Rule Check OK.

seinfra:
Target : [87bff356674e4adb5f9c188ab1d7bc457988ba1f] - Code Rule Check (C++) OK.

os/board/common/crashdump.c
thapav: @jeongchanKim what do you think about above dependency ??

thapav: @sunghan-chang @sijupunnnoose Please check this dependency

PR #4097, opened 2019-12-11
compression : Change default compression algorithm to Miniz

jeongchanKim:
Miniz has lower compression ratio, but has big advantage of decompressing compared with LZMA.
So select the default compression algorithm as Miniz.


seinfra:
Target : [d4b62abca2c7a3f94244a465ff38cd88a8989b5c] - Code Rule Check (C++) OK.

seinfra:
Target : [d4b62abca2c7a3f94244a465ff38cd88a8989b5c] - Code Rule Check OK.

PR #4105, opened 2019-12-16
fs/vfs/fs_open.c:  Fix error in vopen() function

junmin-kim:
'va_start' is used in function with fixed args.
'va_start'/'va_end' should not be used in vopen function,
because it has fixed arguments.


seinfra:
Target : [5da4517c541b02f8c5f76698f873831f4ead06d3] - Code Rule Check OK.

seinfra:
Target : [5da4517c541b02f8c5f76698f873831f4ead06d3] - Code Rule Check (C++) OK.

PR #4107, opened 2019-12-16
framework/araui: Change the FPS calculate procedure.

tyha1115:
Previously, when a FPS of certain frame belows a maximum FPS which was set through menuconfig, the compensation occurs in the next frame.
(i.e., if the maximum FPS is set as 30 and certain frame's FPS is 29, then the next frame's FPS compensated with 31 (if possible).)
Therefore, the compensated frame's FPS exceeds the maximum FPS value.

In this pull request, above problem is solved.

### DIFF
--- a/framework/src/araui/core/ui_core.c
+++ b/framework/src/araui/core/ui_core.c
@@ -447,7 +447,6 @@ static void *_ui_core_thread_loop(void *param)
 	uint32_t dt;
 
 #if (CONFIG_UI_MAXIMUM_FPS > 0)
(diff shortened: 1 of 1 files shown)


seinfra:
Target : [767b06bb46d282eb73f2ab43b401ba7a0631e261] - Code Rule Check OK.

seinfra:
Target : [767b06bb46d282eb73f2ab43b401ba7a0631e261] - Code Rule Check (C++) OK.

PR #4113, opened 2019-12-19
framework/wifi_manager: Modify wifi_manager_deinit API

SeonghoByeon:
- When a user calls wifi_manager_deinit() in sta_connected state, disconnect event will be called prior to deinit operation.
This modification covers the case when the disconnect callback is handled very fast.


seinfra:
Target : [8a8d2ed797b6712065c998722293777f48f34f1e] - Code Rule Check (C++) OK.

seinfra:
Target : [8a8d2ed797b6712065c998722293777f48f34f1e] - Code Rule Check OK.

PR #4179, opened 2020-01-31
os/binfmt: Modify loadable app memory regions to reduce memory wastage

kishore-sn:
If CONFIG_OPTIMIZE_APP_RELOAD_TIME is enabled, then we define 3 MPU regions
for loadable apps. The existing implementation for this is to allocate one large
memory region to cover text, ro, data and heap requirements of the applications.
This large memory region is configured as RW memory in the MPU. Additionally, we
define two MPU regions for text and ro areas which will overlap the large RW region.
Allocating memory regions in this way can sometimes result in wastage of memory
due to the address alignment and region size requirements of the MPU. Specifically,
memory wastage occurs due to resizing the large RW memory region according to the
MPU requirements.

In order to overcome this problem, we modify the memory regions as follows:
- Allocate 2 separate regions for text and ro
- Allocate a seprate memory region for data + heap and configure it as RW region
- All 3 regions will follow the size and address alignment requirements of the MPU

### DIFF
--- a/os/arch/arm/src/armv7-m/up_svcall.c
+++ b/os/arch/arm/src/armv7-m/up_svcall.c
@@ -375,9 +375,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		* USERSPACE->task_startup method. Instead we pick the PC value
 		* from the app's userspace object stored in its tcb.
 		*
-		* Here, we check if this task is a loadable app (non-zero ram_start)
+		* Here, we check if this task is a loadable app (non-zero uspace)
 		*/
-		if (((struct tcb_s *)sched_self())->ram_start) {
+		if (((struct tcb_s *)sched_self())->uspace) {
 			regs[REG_PC] = (uint32_t)((struct userspace_s *)(((struct tcb_s *)sched_self())->uspace))->task_startup;
 		} else
 		/* If its a normal non-loadable user app, then follow the default method */
@@ -422,9 +422,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		* USERSPACE->task_startup method. Instead we pick the PC value
 		* from the app's userspace object stored in its tcb.
 		*
-		* Here, we check if this task is a loadable app (non-zero ram_start)
+		* Here, we check if this task is a loadable app (non-zero uspace)
 		*/
-		if (((struct tcb_s *)sched_self())->ram_start) {
+		if (((struct tcb_s *)sched_self())->uspace) {
 			regs[REG_PC] = (uint32_t)((struct userspace_s *)(((struct tcb_s *)sched_self())->uspace))->pthread_startup;
 		} else
 		/* If its a normal non-loadable user app, then follow the default method */
@@ -477,9 +477,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		* USERSPACE->task_startup method. Instead we pick the PC value
 		* from the app's userspace object stored in its tcb.
 		*
-		* Here, we check if this task is a loadable app (non-zero ram_start)
+		* Here, we check if this task is a loadable app (non-zero uspace)
 		*/
-		if (rtcb->ram_start) {
+		if (rtcb->uspace) {
 			regs[REG_PC] = (uint32_t)((struct userspace_s *)(rtcb->uspace))->signal_handler;
 		} else
 		/* If its a normal non-loadable user app, then follow the default method */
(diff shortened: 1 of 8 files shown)


seinfra:
Target : [9a1201c11dd28496d147fa23dbf9266275324f4d] - Code Rule Check (C++) OK.

seinfra:
Target : [9a1201c11dd28496d147fa23dbf9266275324f4d] - Code Rule Check OK.

os/arch/arm/src/armv7-m/up_svcall.c
```diff
@@ -375,9 +375,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		* USERSPACE->task_startup method. Instead we pick the PC value
 		* from the app's userspace object stored in its tcb.
 		*
-		* Here, we check if this task is a loadable app (non-zero ram_start)
+		* Here, we check if this task is a loadable app (non-zero uspace)
```
jeongchanKim: I think we can separate this to different PR.

os/binfmt/elf.c
```diff
@@ -306,7 +306,7 @@ static int elf_loadbinary(FAR struct binary_s *binp)
 	binp->alloc[4] = (FAR void *)loadinfo.dataalloc;
 	binp->textsize = loadinfo.textsize;
 	binp->rosize = loadinfo.rosize;
-	binp->datasize = loadinfo.datasize;
+	binp->datasize = loadinfo.datasize - binp->bsssize;
```
jeongchanKim: Could you remain comments why we subtract bsssize?

os/tools/mkbinheader.py
```diff
@@ -206,8 +206,12 @@ def get_static_ram_size(bin_type):
         sys.exit(1)
 
     static_ram_size = get_static_ram_size(bin_type)
-    binary_ram_size = int(static_ram_size) + int(dynamic_ram_size)
-    binary_ram_size = roundup_power_two(binary_ram_size)
+    cfg_path = os.getenv('TOPDIR') + '/.config'
+    if check_optimize_config(cfg_path) == True:
+        binary_ram_size = int(dynamic_ram_size)
```
jeongchanKim: For optimization case, static ram size is not needed?
If you think static is also allocated in heap, user cannot think dynamic size should include static size.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -151,37 +151,37 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
 
 #ifdef CONFIG_APP_BINARY_SEPARATION
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	uint32_t datamemsize = loadinfo->datasize + loadinfo->binp->ramsize;
```
jeongchanKim: I think we'd better remain comments why we calculate like below.

os/binfmt/elf.c
kishore-sn: Done.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Done

os/tools/mkbinheader.py
kishore-sn: I understand that you want to give an indication to the user about exact memory requirement for the application. 
For this purpose, i can add some print statement. But, for putting the value in the binary header, I need to put only the dynamic ram. 

In previous scenario, we were allocating one big partition containing text + ro + data + bss + heap. So, I was able to calculate the static and dynamic size here and pass the total size in the binary header. The app would read this size and directly allocate a partition of the size. 

But now, the app allocates memory in 3 parts: round_up(text) , round_up(ro + data) and round_up(data + bss + heap). There is no way for me to pass this information in the header.

seinfra:
Target : [10c5c05f186c3dfc329034e0986206667bed3c8f] - Code Rule Check (C++) OK.

seinfra:
Target : [10c5c05f186c3dfc329034e0986206667bed3c8f] - Code Rule Check OK.

os/tools/mkbinheader.py
jeongchanKim: I think if we need to pass much data, we can expand the header.
What do you think about this?

kishore-sn: I think the header must be compact. It must contain only the information which the loader cannot get from any other source. Since, the section information can be obtained from the elf, it must not be part of the header. Also, adding section information will make the header tightly coupled with the elf format. If any changes in elf sections or any new format needs to be used, then header will need to be modified. That is why I suggest that the header must contain only the dynamic ram size. If you want to provide an indication to user about the exact ram requirement, it can be given in the form of print logs from the script file or from the runtime logs in the loader which will indicate exact amounts of memory allocated.

PR #4183, opened 2020-02-03
os/Kconfig: Remove CONFIG_TINYARA_USERSPACE configuration variable

kishore-sn:
TizenRT protected build generates 2 binaries for kernel and user.
    The build system makes use of the linker variables defined in memory.ld
    file to link the binary to the proper memory addresses. The linker script
    always places the userspace object at the start of the user binary. Hence
    we can assume that the linker variable __uflash_segment_start__ always
    points to the userspace object in flash. Hence, this variable can be used
    instead of CONFIG_TINYARA_USERSPACE to obtain the address of the userspace
    object. The memory regions in memory.ld file can either be auto-generated
    based on the flash partition configuration in CONFIG_FLASH_PART_SIZE, or
    may be manually generated. In either case, the CONFIG_TINYARA_USERSPACE
    is no longer required.


seinfra:
Target : [280c201a4d74268fc6f3626090f85442b7dbda3a] - Code Rule Check (C++) OK.

seinfra:
Target : [280c201a4d74268fc6f3626090f85442b7dbda3a] - Code Rule Check OK.

PR #4193, opened 2020-02-11
build/configs/esp_wrover_kit: modify readme.md

Hyunjun85:
I have faced an error when tried to setup env. to build.
the 'xvjf' means that the compressed file is bzip2.
but actually the file is not bzip file. so It can show some error messages.


seinfra:
Target : [67946911003c040d50a9294f4e71e0208b6b1baa] - Code Rule Check OK.

seinfra:
Target : [67946911003c040d50a9294f4e71e0208b6b1baa] - Code Rule Check (C++) OK.

PR #4200, opened 2020-02-13
Update up_disable_watchdog() dependency on CONFIG_WATCHDOG_FOR_IRQ in imxrt

thapav:
This dependency causes build break in imxrt when ramdump is enabled.
This patch updates that dependency
It also updates crashdump file to send no Kernel MM heap notification to the ramdump tool in case only USer MM regions are present.


seinfra:
Target : [b60ac0fa384d6668a3cc7fdbffd8aaae304dbce5] - Code Rule Check (C++) OK.

seinfra:
Target : [b60ac0fa384d6668a3cc7fdbffd8aaae304dbce5] - Code Rule Check OK.

PR #4234, opened 2020-03-19
os/syscall : Remove syscall for telldir and pthread_setcancelstate

kishore-sn:
The API's for telldir and pthread_setcancelstate have been completely
implemented in the c library. Hence they are not part of system call.


seinfra:
Target : [4c1480723345a80357ba2c63394887cbd0d60621] - Code Rule Check (C++) OK.

seinfra:
Target : [4c1480723345a80357ba2c63394887cbd0d60621] - Code Rule Check Result:
os/syscall/syscall_lookup.h:352: ERROR: [IDT_M_TAB] please, no space before tabs

os/syscall/syscall.csv
```diff
@@ -85,7 +85,6 @@
 "pthread_mutex_trylock", "pthread.h", "!defined(CONFIG_DISABLE_PTHREAD)", "int", "FAR pthread_mutex_t*"
 "pthread_mutex_unlock", "pthread.h", "!defined(CONFIG_DISABLE_PTHREAD)", "int", "FAR pthread_mutex_t*"
 "pthread_mutex_consistent", "pthread.h", "!defined(CONFIG_DISABLE_PTHREAD) && !defined(CONFIG_PTHREAD_MUTEX_UNSAFE)", "int", "FAR pthread_mutex_t*"
-"pthread_setcancelstate", "pthread.h", "!defined(CONFIG_DISABLE_PTHREAD)", "int", "int", "FAR int*"
```
jeongchanKim: `pthread_setcancelstate` uses `task_setcancelstate`, but `task_setcancelstate` is kernel API.

kishore-sn: Yes, but task_setcancelstate is included in  the syscall interface.

sangam-swami:
> Target : [[4c14807](https://github.com/Samsung/TizenRT/commit/4c1480723345a80357ba2c63394887cbd0d60621)] - Code Rule Check Result:
> os/syscall/syscall_lookup.h:352: ERROR: [IDT_M_TAB] please, no space before tabs

you can remove the extra space

os/syscall/syscall.csv
jeongchanKim: Yes, but with your perspective, we can remove another syscall APIs.
For example, `pthread_setschedparam`. It is almost wrapper of `sched_setscheduler`.

kishore-sn: I am removing these two syscalls  because they are causing problem for the shared library build.. they are coming as undefined symbols. But, there is no such issue with other syscalls, so I think we can keep the others.

seinfra:
Target : [e937286d8c13a515dabf7233f11575e09bb73f90] - Code Rule Check OK.

seinfra:
Target : [e937286d8c13a515dabf7233f11575e09bb73f90] - Code Rule Check (C++) OK.

os/syscall/syscall.csv
jeongchanKim: Okay, but then, we'd better find why only those two APIs have problem.

kishore-sn: These two APIs are cause problem because:
- I am including libuc and libproxies in the common binary
- I am using --whole-archive linker flag to include all symbols from each library in the common binary
- The above two API's are defined in both the libraries
- Hence I get linker error as below for these two API's. None of the other syscalls are defined in two places. They are defined only in libproxies.

../build/output/libraries/libproxies.a(PROXY_pthread_setcancelstate.o): In function `pthread_setcancelstate':
/home/kishore/tinyara/TizenRT_kishore/os/syscall/proxies/PROXY_pthread_setcancelstate.c:10: multiple definition of `pthread_setcancelstate'
../build/output/libraries/libuc.a(pthread_setcancelstate.o):/home/kishore/tinyara/TizenRT_kishore/lib/libc/pthread/pthread_setcancelstate.c:96: first defined here
../build/output/libraries/libproxies.a(PROXY_telldir.o): In function `telldir':
/home/kishore/tinyara/TizenRT_kishore/os/syscall/proxies/PROXY_telldir.c:10: multiple definition of `telldir'
../build/output/libraries/libuc.a(lib_telldir.o):/home/kishore/tinyara/TizenRT_kishore/lib/libc/dirent/lib_telldir.c:95: first defined here
Makefile.unix:452: recipe for target 'pass1' failed

jeongchanKim: Okay. I understood. You mean that API is in libc and if it is syscall, it can be a problem.
If we want to remove from syscall list, we should modify the description about that api in header file(There is a description that a API is syscall or not.). 
And I'm not sure we can remove easily, because of Tizen ACR(Api Change Request Process).

PR #4235, opened 2020-03-19
Update Video Driver, NULL Driver for video and Camera app features.

amogh-samsung:
os/drivers/video - NULL driver
- Fix NULL Driver file read problems.
- Add delay setting function for worker in NULL Video Driver.
apps/examples/camera - camera main
- Change Camera app frame saving directory.
os/drivers/video - video.c
- Change checking point for return value from set_buf() in change_video_state()

### DIFF
--- a/os/drivers/video/video.c
+++ b/os/drivers/video/video.c
@@ -307,7 +307,7 @@ static enum video_state_e estimate_next_video_state(FAR video_upperhalf_t *priv,
 
 static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e next_state)
 {
-	int ret = OK;
+	int ret;
 	enum video_state_e current_state = priv->video_inf.state;
 	enum video_state_e updated_next_state = next_state;
 	FAR vbuf_container_t *dma_container;
@@ -317,6 +317,10 @@ static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e ne
 		dma_container = video_framebuff_get_dma_container(&priv->video_inf.bufinf);
 		if (dma_container) {
 			ret = video_devops->set_buf(priv->dev, V4L2_BUF_TYPE_VIDEO_CAPTURE, dma_container->buf.m.userptr, dma_container->buf.length);
+			if (ret != OK) {
+				videodbg("Unable to set buffer for state change.\n");
+				return ret;
+			}
 		} else {
 			updated_next_state = VIDEO_STATE_STREAMON;
 		}
@@ -326,11 +330,8 @@ static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e ne
 		}
 	}
 
-	if (ret == OK) {
-		priv->video_inf.state = updated_next_state;
-	}
-
-	return ret;
+	priv->video_inf.state = updated_next_state;
+	return OK;
 }
 
 static bool is_taking_still_picture(FAR video_upperhalf_t *priv)
--- a/os/drivers/video/video.c
+++ b/os/drivers/video/video.c
@@ -709,6 +709,11 @@ static int video_set_fmt(FAR video_upperhalf_t *priv, FAR struct v4l2_format *fm
 		return -EINVAL;
 	}
 
+	if (priv->video_inf.state == VIDEO_STATE_STREAMON || priv->still_inf.state == VIDEO_STATE_STREAMON) {
+		videodbg("Device currently streaming, unable to change parameters.\n");
+		return -EBUSY;
+	}
+
 	video_devops = priv->dev->ops;
 
 	if ((video_devops == NULL) || (video_devops->set_format == NULL)) {
@@ -747,6 +752,11 @@ static int video_set_parm(FAR video_upperhalf_t *priv, FAR struct v4l2_streampar
 
 	if (priv == NULL) {
 		return -EINVAL;
+	}
+
+	if (priv->video_inf.state == VIDEO_STATE_STREAMON || priv->still_inf.state == VIDEO_STATE_STREAMON) {
+		videodbg("Device currently streaming, unable to change parameters.\n");
+		return -EBUSY;
 	}
 
 	video_devops = priv->dev->ops;
@@ -1116,6 +1126,11 @@ static int video_set_ext_ctrls(FAR video_upperhalf_t *priv, FAR struct v4l2_ext_
 
 	if ((priv == NULL) || (ctrls == NULL)) {
 		return -EINVAL;
+	}
+
+	if (priv->video_inf.state == VIDEO_STATE_STREAMON || priv->still_inf.state == VIDEO_STATE_STREAMON) {
+		videodbg("Device currently streaming, unable to change parameters.\n");
+		return -EBUSY;
 	}
 
 	video_devops = priv->dev->ops;
--- a/os/drivers/video/video_null_driver.c
+++ b/os/drivers/video/video_null_driver.c
@@ -35,6 +35,7 @@
 #include <tinyara/wqueue.h>
 
 #include <video/video_halif.h>
+#include <video/video.h>
 
 #include "video_null_driver.h"
 
@@ -49,6 +50,36 @@
 #define MAX_FRAMES 7
 /* Used only in case of discrete */
 #define MAX_INTERVAL 5
+
+/* Normalized Frame Interval numerators based on fps */
+#define FINT_30_NUM 333333
+#define FINT_20_NUM 500000
+#define FINT_15_NUM 666666
+#define FINT_10_NUM 1000000 /* Two denominators defined for 10fps transfer, refer "v4l2_format_s supported_formats[]" */
+#define FINT_10a_NUM 999999 /* We regard both as same for setting the worker delay */
+
+/* Normalized 8 digit denominator for simplified mathematical calculations
+   of up to 6 decimal places with aforementioned Frame Interval Numerators
+*/
+#define FINT_DEN 10000000
+
+/* Standard delay (in ms) for worker thread callback at 30fps for various resolutions*/
+#define S_DELAY_QVGA 32
+#define S_DELAY_VGA 28
+#define S_DELAY_QUADVGA 15
+#define S_DELAY_HD 19
+#define S_DELAY_FULLHD 4
+#define S_DELAY_5M 0
+#define S_DELAY_3M 0
+
+/* Excess delay (in ms) for worker thread callback for lower fps */
+#define E_DELAY_FPS_30 0
+#define E_DELAY_FPS_20 17
+#define E_DELAY_FPS_15 33
+#define E_DELAY_FPS_10 67
+
+/* Default Worker Delay, QVGA Video Frames at 30 fps */
+#define W_DELAY_DEFAULT 32
 
 #define CHECK_RANGE(value, min, max, step) do { \
 												if ((value < min) || \
@@ -144,8 +175,14 @@ struct dummy_null_priv_s {
 	struct v4l2_format_s *def_format;	/* Reference to default format */
 	struct v4l2_format_s *cur_format;	/* Reference to current format */
 	struct v4l2_frames_s *cur_frame;	/* Reference to current frame */
+	struct v4l2_fract cur_fintvl;		/* Current Frame Interval, Default 30 fps*/
+	uint8_t w_delay;			/* Delay, in clock ticks, for the worker*/
 };
 typedef struct dummy_null_priv_s null_priv_t;
+
+/* Array of worker thread time delays(in ms) for streaming at 30fps */
+/* TODO: Proper worker delays for 3M and 5M resolutions to be calculated */
+uint8_t std_w_delays[] = {S_DELAY_QVGA, S_DELAY_VGA, S_DELAY_QUADVGA, S_DELAY_HD, S_DELAY_FULLHD, S_DELAY_5M, S_DELAY_3M};
 
 struct video_devops_s dummy_null_video_ops = {
 	.open = video_null_open,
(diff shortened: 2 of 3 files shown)


seinfra:
Target : [5f2964734566f8dd450e1161c61f2c17efb1fef3] - Code Rule Check Result:
os/drivers/video/video_null_driver.c:285: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/drivers/video/video_null_driver.c:298: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/drivers/video/video_null_driver.c:319: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/drivers/video/video_null_driver.c:1110: ERROR: [SPC_M_KWD] space required before the open parenthesis '('
os/drivers/video/video_null_driver.c:1283: ERROR: [SPC_M_KWD] space required before the open parenthesis '('

seinfra:
Target : [5f2964734566f8dd450e1161c61f2c17efb1fef3] - Code Rule Check (C++) OK.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+ *   video frames-per-second required.
+ *
+ * Input Paramaters:
+ *   priv - pointer to null_priv_t object holding the frame
+ *   resolution attributes.
+ *
+ * Returned Values:
+ *   OK on success, appropriate error code otherwise.
+ ****************************************************************************/
+static int video_null_set_worker_delay(FAR null_priv_t *priv)
```
amogh-samsung: This function calculates the delay as follows - 
1. Based on the frame resolution, a standard delay has been calculated assuming a required transfer at 30fps.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+			break;
+		}
+	}
+
+	if (i == format->framecount) {
+		videodbg("No matching standard frame resolution found, kindly check");
+		return -EINVAL;
+	}
+
+	priv->w_delay = std_w_delays[i];
```
amogh-samsung: 2. The current format is compared and the standard delay is assigned.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+	videodbg("Frame Interval = %d/%d\n", priv->cur_fintvl.numerator, priv->cur_fintvl.denominator);
+	/* Adding the excess delay required for transfer at less than 30fps */
+
+	switch(priv->cur_fintvl.numerator) {
```
amogh-samsung: 3. Delay is increased by a certain amount if transfer is required slower than 30fps.

seinfra:
Target : [d970761f67821041ba025ab4ed283216065f5c0d] - Code Rule Check OK.

seinfra:
Target : [d970761f67821041ba025ab4ed283216065f5c0d] - Code Rule Check (C++) OK.

os/drivers/video/video_null_driver.c
amogh-samsung: This function calculates the delay as follows -

1. Based on the frame resolution, a standard delay has been calculated assuming a required transfer at 30fps.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+			break;
+		}
+	}
+
+	if (i == format->framecount) {
+		videodbg("No matching standard frame resolution found, kindly check");
+		return -EINVAL;
+	}
+
+	priv->w_delay = std_w_delays[i];
```
amogh-samsung: 2. The current format is compared and the standard delay is assigned.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+	videodbg("Frame Interval = %d/%d\n", priv->cur_fintvl.numerator, priv->cur_fintvl.denominator);
+	/* Adding the excess delay required for transfer at less than 30fps */
+
+	switch (priv->cur_fintvl.numerator) {
```
amogh-samsung: 3. Delay is increased by a certain amount if transfer is required slower than 30fps.

os/drivers/video/video.c
```diff
@@ -317,6 +317,9 @@ static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e ne
 		dma_container = video_framebuff_get_dma_container(&priv->video_inf.bufinf);
 		if (dma_container) {
 			ret = video_devops->set_buf(priv->dev, V4L2_BUF_TYPE_VIDEO_CAPTURE, dma_container->buf.m.userptr, dma_container->buf.length);
+			if (ret < OK) {
```
thapav: can there be a case when return value is not updated and set_buf still fails ??
In that case can w ehave the default ret value in the start of this API as ERROR ?

amogh-samsung: Could you please point out where such a case might arise?
`set_buf` is expected to always return an ERROR value in case of failure and OK on success.

os/drivers/video/video_null_driver.c
```diff
@@ -429,18 +533,18 @@ int video_null_init_transfer(FAR null_priv_t *priv)
...
 	videodbg("Video File Name: %s\n", filename);
 
-#if 0
 	fd = open(filename, O_RDOK);
 	if (fd < 0) {
 		return -ENOENT;
 	}
 
-	ret = read(fd, priv->trfbuff, priv->reqsize);
+	ret = read(fd, priv->trfbuff, priv->trfsize);
```
Taejun-Kwon, line 541: if (ret != priv->trfsize)

os/drivers/video/video_null_driver.c
```diff
@@ -1481,6 +1600,11 @@ int video_null_initialize(const char *devpath)
 	priv->def_format = &priv->fmt[0];
 	priv->cur_format = &priv->fmt[0];
 	priv->cur_frame = &priv->fmt[0].frames[0];
+	/* Set the default frame interval */
+	priv->cur_fintvl.numerator = 333333;
```
Taejun-Kwon, line 1604: Please do not use constant number as possible as you can.

os/drivers/video/video.c
thapav: Also if you could make the default return value (at initialization) as ERROR so once this call succeeds, it will be updated to OK, else ERROR value will be returned

os/drivers/video/video_null_driver.c
```diff
@@ -144,9 +162,16 @@ struct dummy_null_priv_s {
...
 	struct v4l2_format_s *cur_format;	/* Reference to current format */
 	struct v4l2_frames_s *cur_frame;	/* Reference to current frame */
+	struct v4l2_fract cur_fintvl;		/* Current Frame Interval, Default 30 fps*/
+	uint8_t w_delay;			/* Delay, in clock ticks, for the worker*/
 };
 typedef struct dummy_null_priv_s null_priv_t;
 
+/* Array of worker thread time delays for streaming at 30fps */
+/* TODO: Worker delays for 3M and 5M resolutions to be calculated */
+/* Current restriction due to memory limitations on IMXRT1050-evk board */
```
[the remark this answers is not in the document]
amogh-samsung: The time delays for higher resolution frames couldn't be calculated due to inability to test on the imxrt1050-evk board. These shall be obtained with a board having higher memory availability.
There is no board specific dependence of the NULL driver.

os/drivers/video/video_null_driver.c
amogh-samsung: We need not confirm that `ret == priv->trfsize` since the number of bytes may very well be less than priv->trfsize.
trfsize calculated as 2 x height x frame (eg. for VGA Frames: - 2 x 320 x 240) is the maximum size of the frame that may be read. Frame size may be smaller than this for certain formats.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+ ****************************************************************************/
+static int video_null_set_worker_delay(FAR null_priv_t *priv)
+{
+	int i;
+	struct v4l2_format_s *format;
+	if (priv == NULL) {
+		videodbg("Unable to set worker delay duration!!!\n");
+	}
+
+	format = priv->cur_format;
```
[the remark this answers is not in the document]
amogh-samsung: In case a format change change request is made while the device is already active (streaming), the device stream first has to be stopped before any format or frame parameters for the stream can be changed. So we need not worry about format changes after the lower driver has already started work.

os/drivers/video/video_null_driver.c
amogh-samsung: Numerator = 333333 and Denominator = 10000000 have no significance in relation to settings for the NULL Driver. These are values used for initialization only for mathematical convenience with fractions, for ease of calculations and to obtain accuracy in the final frame interval calculated.
Similar to Line 1270 `timeperframe.denominator = 10000000;`

seinfra:
Target : [785a4d8c9b03623c2e85568d7189f3d327d286b6] - Code Rule Check (C++) OK.

seinfra:
Target : [785a4d8c9b03623c2e85568d7189f3d327d286b6] - Code Rule Check OK.

os/drivers/video/video_null_driver.c
```diff
@@ -237,6 +262,87 @@ struct v4l2_mapping_s ctrl_mapping[] = {
...
+ *
+ * Returned Values:
+ *   OK on success, appropriate error code otherwise.
+ ****************************************************************************/
+static int video_null_set_worker_delay(FAR null_priv_t *priv)
+{
+	int i;
+	struct v4l2_format_s *format;
+	if (priv == NULL) {
+		videodbg("Unable to set worker delay duration!!!\n");
```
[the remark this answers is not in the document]
amogh-samsung: Agreed. Changes made

os/drivers/video/video.c
amogh-samsung: Changes made.

os/drivers/video/video.c
```diff
@@ -326,11 +329,8 @@ static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e ne
 		}
 	}
 
-	if (ret == OK) {
-		priv->video_inf.state = updated_next_state;
-	}
-
-	return ret;
+	priv->video_inf.state = updated_next_state;
```
Taejun-Kwon, line 332: it is state changing, lock mechanism should be there.

os/drivers/video/video_null_driver.c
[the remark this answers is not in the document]
Taejun-Kwon: You don't have to care about board here, null driver is just for testing without any device.
Suggest below.
1. Remove comment regarding 1050 boards.
2. Define all possible values & Provide KConfig for resolution if it needed.

os/drivers/video/video_null_driver.c
```diff
@@ -429,18 +534,18 @@ int video_null_init_transfer(FAR null_priv_t *priv)
...
 
-	ret = read(fd, priv->trfbuff, priv->reqsize);
+	ret = read(fd, priv->trfbuff, priv->trfsize);
 	if (ret < 0) {
+		close(fd);
 		return ret;
 	}
-#endif
 
+	close(fd);
```
Taejun-Kwon, line 548: can be cleaned.

os/drivers/video/video_null_driver.c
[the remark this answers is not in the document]
Taejun-Kwon: Video layer provides some apis to changes format, for example video_set_parm, it seems that can be called in the middle of streaming.

os/drivers/video/video_null_driver.c
Taejun-Kwon: same opinion as I wrote line 172, null driver should test none-fixed parameter. please provide KConfig or something.

os/drivers/video/video_null_driver.c
```diff
@@ -1002,6 +1107,10 @@ static int video_null_set_format(FAR void *video_private, FAR struct v4l2_format
 	}
 
 	ret = video_null_v4l2_try_format(priv, fmt, 1);
+	/* Format changed, set worker function delay accordingly */
+	if (ret > 0) {
```
Taejun-Kwon, line 1111: ret != OK is correct, in my opinion.

os/drivers/video/video_null_driver.c
```diff
@@ -1160,11 +1269,22 @@ static int video_null_set_frameinterval(FAR void *video_private, FAR struct v4l2
...
 	timeperframe.numerator = frameInterval;
 	timeperframe.denominator = 10000000;
-	video_null_reduce_fraction(&timeperframe.numerator, &timeperframe.denominator);
 
+	/* Set current frame interval details according to the unreduced numerator and denominator */
+	priv->cur_fintvl.numerator = timeperframe.numerator;
+	priv->cur_fintvl.denominator = timeperframe.denominator;
+
+	video_null_reduce_fraction(&timeperframe.numerator, &timeperframe.denominator);
+	videodbg("Numerator = %d, denominator = %d\n", timeperframe.numerator, timeperframe.denominator);
```
Taejun-Kwon, line 1278: verbose type is correct, I think.

os/drivers/video/video.c
sijupunnnoose: change_video_state function is called always by caller with adequate locking, hence no need to do inside the function.

os/drivers/video/video_null_driver.c
amogh-samsung: The value 'ret' returned by `video_null_v4l2_try_format` can be an error value (negative), the number of bytes transferred during the format trial (positive) or OK (zero).
Only in the case where transfer of non-zero bytes with the new format is successful, the format of the null driver will be changed and we need to reset the worker's delay.
Hence, only checking for `ret > 0`.

os/drivers/video/video_null_driver.c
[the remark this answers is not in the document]
amogh-samsung: Done

os/drivers/video/video_null_driver.c
amogh-samsung: Done

os/drivers/video/video_null_driver.c
amogh-samsung: Done

os/drivers/video/video_null_driver.c
amogh-samsung: Done

os/drivers/video/video.c
```diff
@@ -317,6 +317,9 @@ static int change_video_state(FAR video_upperhalf_t *priv, enum video_state_e ne
 		dma_container = video_framebuff_get_dma_container(&priv->video_inf.bufinf);
 		if (dma_container) {
 			ret = video_devops->set_buf(priv->dev, V4L2_BUF_TYPE_VIDEO_CAPTURE, dma_container->buf.m.userptr, dma_container->buf.length);
+			if (ret != OK) {
+				return ret;
```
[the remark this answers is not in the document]
amogh-samsung: Done

os/drivers/video/video_null_driver.c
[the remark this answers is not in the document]
amogh-samsung: Pending, re-checking.

PR #4241, opened 2020-03-27
change the file permission to 644 for *.c, *.h, *.png, defconfig

sunghan-chang:
They are read-only files so that the permission should be 644.


seinfra:
Target : [9610c2849afaca5da97a47628e68620c54e254d3] - Code Rule Check (C++) OK.

PR #4248, opened 2020-04-05
Support MPU Stack Overflow Protection

sangam-swami:
Implement task's stack overflow protection with MPU support.
Stack guard memory(min of 32 bytes) will be reserved at the end of
each task's stack memory. The Stack guard memory will be marked as RO,
so that stack access above the permissible limit will result into
memory fault with MPU.

### DIFF
--- a/os/arch/arm/Kconfig
+++ b/os/arch/arm/Kconfig
@@ -271,6 +271,28 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config MPU_STACK_OVERFLOW_PROTECTION
+	bool "MPU Stack Overflow Protection"
+	depends on BUILD_PROTECTED
+	depends on ARM_MPU
+	default n
+	---help---
+		Enables stack overflow protection.
+		It protects each task's stack overflow outside its max allocated stack memory.
+		This is applicable to user mode tasks created in protected build.
+
+if MPU_STACK_OVERFLOW_PROTECTION
+
+config MPU_STACK_GUARD_SIZE
+	int "Stack Guard size configuration range(32-1024)"
+	default 32
+	range 32 1024
+	---help---
+		Set size for the stack guard. It must be set with 32 byte aligned values.
+		The size can be set in the range of 32 to 1024 bytes.
+		Setting large stack guard size will be a wastage of memory.
+endif
+
 if ARCH_CORTEXM3 || ARCH_CORTEXM4 || ARCH_CORTEXM7
 source arch/arm/src/armv7-m/Kconfig
 endif
--- a/os/arch/arm/src/armv7-m/mpu.h
+++ b/os/arch/arm/src/armv7-m/mpu.h
@@ -687,6 +687,33 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 #define up_set_mpu_app_configuration(x)
 #endif
 
+/****************************************************************************
+ * Name: up_set_mpu_stack_guard
+ *
+ * Description:
+ *   Configure task's stack guard region
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+static inline void up_set_mpu_stack_guard(struct tcb_s *rtcb)
+{
+	/* We update the MPU registers only if:
+	 * This is not a kernel thread AND
+	 * It has a non zero value of base address (This ensures valid MPU setting)
+	 */
+	if ((rtcb->flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_KERNEL) {
+		return;
+	}
+
+	if (rtcb->stack_mpu_regs[REG_RBAR]) {
+		putreg32(rtcb->stack_mpu_regs[REG_RNR], MPU_RNR);
+		putreg32(rtcb->stack_mpu_regs[REG_RBAR], MPU_RBAR);
+		putreg32(rtcb->stack_mpu_regs[REG_RASR], MPU_RASR);
+	}
+}
+#endif
+
 #undef EXTERN
 #if defined(__cplusplus)
 }
--- a/os/arch/arm/src/armv7-m/up_blocktask.c
+++ b/os/arch/arm/src/armv7-m/up_blocktask.c
@@ -169,6 +169,9 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/mpu.h
+++ b/os/arch/arm/src/armv8-m/mpu.h
@@ -687,6 +687,34 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 #define up_set_mpu_app_configuration(x)
 #endif
 
+/****************************************************************************
+ * Name: up_set_mpu_stack_guard
+ *
+ * Description:
+ *   Configure task's stack guard region
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+static inline void up_set_mpu_stack_guard(struct tcb_s *rtcb)
+{
+	/* We update the MPU registers only if:
+	 * This is not a kernel thread AND
+	 * It has a non zero value of base address (This ensures valid MPU setting)
+	 */
+	if ((rtcb->flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_KERNEL) {
+		return;
+	}
+
+	if (rtcb->stack_mpu_regs[REG_RBAR]) {
+		putreg32(rtcb->stack_mpu_regs[REG_RNR], MPU_RNR);
+		putreg32(rtcb->stack_mpu_regs[REG_RBAR], MPU_RBAR);
+		putreg32(rtcb->stack_mpu_regs[REG_RASR], MPU_RASR);
+	}
+
+}
+#endif
+
 #undef EXTERN
 #if defined(__cplusplus)
 }
--- a/os/arch/arm/src/common/up_createstack.c
+++ b/os/arch/arm/src/common/up_createstack.c
@@ -72,6 +72,10 @@
 #include "up_arch.h"
 #include "up_internal.h"
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+#include <tinyara/mpu.h>
+#endif
+
 /****************************************************************************
  * Pre-processor Macros
  ****************************************************************************/
@@ -158,6 +162,8 @@
 
 int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 {
+	size_t stack_alloc_size = stack_size;
+
 	/* Is there already a stack allocated of a different size?  Because of
 	 * alignment issues, stack_size might erroneously appear to be of a
 	 * different size.  Fortunately, this is not a critical operation.
@@ -180,12 +186,17 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		/* Use the kernel allocator if this is a kernel thread */
 
 		if (ttype == TCB_FLAG_TTYPE_KERNEL) {
-			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_size);
+			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_alloc_size);
 		} else
 #endif
 		{
 			/* Use the user-space allocator if this is a task or pthread */
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION	/* address must be aligned to 32 byte for MPU region */
+			stack_alloc_size = stack_alloc_size + CONFIG_MPU_STACK_GUARD_SIZE;
+			tcb->stack_alloc_ptr = (uint32_t *)kumm_memalign(CONFIG_MPU_STACK_GUARD_SIZE, stack_alloc_size);
+#else
 			tcb->stack_alloc_ptr = (uint32_t *)kumm_malloc(stack_size);
+#endif
 		}
 
 #ifdef CONFIG_DEBUG
@@ -209,7 +220,8 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		 * on the stack are referenced as positive word offsets from sp.
 		 */
 
-		top_of_stack = (uint32_t)tcb->stack_alloc_ptr + stack_size - 4;
+		top_of_stack = (uint32_t)tcb->stack_alloc_ptr + stack_alloc_size - 4;
+
 
 		/* The ARM stack must be aligned; 4 byte alignment for OABI and
 		 * 8-byte alignment for EABI. If necessary top_of_stack must be
@@ -226,6 +238,10 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 
 		size_of_stack = top_of_stack - (uint32_t)tcb->stack_alloc_ptr + 4;
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		/* Adjust stack size after guard_size calculationn */
+		size_of_stack = size_of_stack - CONFIG_MPU_STACK_GUARD_SIZE;
+#endif
 		/* Save the adjusted stack values in the struct tcb_s */
 
 		tcb->adj_stack_ptr = (uint32_t *)top_of_stack;
@@ -243,6 +259,12 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+		mpu_get_register_value(&tcb->stack_mpu_regs[0], STACKGUARD_MPU_REGION_NUM,
+			(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
+#endif
+
 		return OK;
 	}
 
--- a/os/include/tinyara/mpu.h
+++ b/os/include/tinyara/mpu.h
@@ -1,0 +1,52 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+///@file tinyara/mpu.h
+
+#ifndef __INCLUDE_TINYARA_MPU_H
+#define __INCLUDE_TINYARA_MPU_H
+
+/********************************************************************************
+ * Included Files
+ ********************************************************************************/
+
+#include <tinyara/config.h>
+
+/********************************************************************************
+ * Pre-processor Definitions
+ ********************************************************************************/
+#if (defined(CONFIG_ARM_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+enum {
+	REG_RNR,
+	REG_RBAR,
+	REG_RASR,
+	REG_MAX
+};
+
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#define MPU_NUM_REGIONS		3
+#else
+#define MPU_NUM_REGIONS		1
+#endif
+#endif
+
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+#define STACKGUARD_MPU_REGION_NUM		3	/* Region number 3 is reserved for stack protection */
+#endif
+
+#endif
--- a/os/include/tinyara/sched.h
+++ b/os/include/tinyara/sched.h
@@ -80,25 +80,10 @@
 #include <tinyara/mm/shm.h>
 #include <tinyara/fs/fs.h>
 #include <tinyara/net/net.h>
+#include <tinyara/mpu.h>
 
 #include <arch/arch.h>
 
-/********************************************************************************
- * Pre-processor Definitions
- ********************************************************************************/
-#if defined(CONFIG_APP_BINARY_SEPARATION) && defined(CONFIG_ARM_MPU)
-enum {
-	REG_RNR,
-	REG_RBAR,
-	REG_RASR
-};
-
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
-#define MPU_NUM_REGIONS		3
-#else
-#define MPU_NUM_REGIONS		1
-#endif
-#endif
 /* Configuration ****************************************************************/
 /* Task groups currently only supported for retention of child status */
 
@@ -635,6 +620,10 @@ struct tcb_s {
 #ifdef CONFIG_ARM_MPU
 	uint32_t mpu_regs[3 * MPU_NUM_REGIONS];	/* We need 3 register values to configure each MPU region */
 #endif
+#endif
+
+#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+uint32_t stack_mpu_regs[REG_MAX]; /* need 3 MPU registers to configure a region */
 #endif
 
 #if CONFIG_TASK_NAME_SIZE > 0
--- a/os/tools/mkconfig.c
+++ b/os/tools/mkconfig.c
@@ -305,6 +305,16 @@ int main(int argc, char **argv, char **envp)
 	printf("# undef CONFIG_DEBUG_SPI\n");
 	printf("# undef CONFIG_DEBUG_HEAP\n");
 	printf("#endif\n\n");
+	printf("#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)\n");
+	printf("# if (CONFIG_MPU_STACK_GUARD_SIZE != 32) && \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 64) && \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 128) && \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 256) && \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 512) && \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 1024) \n");
+	printf("#  define CONFIG_MPU_STACK_GUARD_SIZE 32\n");
+	printf("# endif\n");
+	printf("#endif\n\n");
 	printf("#endif /* __INCLUDE_CONFIG_H */\n");
 	fclose(stream);
 
--- a/os/arch/arm/src/armv7-m/up_mpu.c
+++ b/os/arch/arm/src/armv7-m/up_mpu.c
@@ -306,15 +306,15 @@ uint32_t mpu_subregion(uintptr_t base, size_t size, uint8_t l2size)
 }
 
 /****************************************************************************
- * Name: mpu_configure_app_regs
+ * Name: mpu_get_register_value
  *
  * Description:
  *   Configure the user application SRAM mpu settings into the tcb variables
  *
  ****************************************************************************/
 
-#if defined(CONFIG_ARMV7M_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)
-void mpu_configure_app_regs(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute)
+#if (defined(CONFIG_ARM_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+void mpu_get_register_value(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute)
 {
 	uint32_t regval;
 	uint8_t l2size;
--- a/os/arch/arm/src/armv7-m/up_releasepending.c
+++ b/os/arch/arm/src/armv7-m/up_releasepending.c
@@ -133,6 +133,9 @@ void up_release_pending(void)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv7-m/up_reprioritizertr.c
+++ b/os/arch/arm/src/armv7-m/up_reprioritizertr.c
@@ -184,6 +184,9 @@ void up_reprioritize_rtr(struct tcb_s *tcb, uint8_t priority)
 				/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 				up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+				up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 				/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv7-m/up_svcall.c
+++ b/os/arch/arm/src/armv7-m/up_svcall.c
@@ -262,6 +262,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 		up_set_mpu_app_configuration(tcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		up_set_mpu_stack_guard(tcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 		/* Update tcb active flag for monitoring. */
@@ -300,6 +303,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 		up_set_mpu_app_configuration(tcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		up_set_mpu_stack_guard(tcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 		/* Update tcb active flag for monitoring. */
--- a/os/arch/arm/src/armv7-m/up_unblocktask.c
+++ b/os/arch/arm/src/armv7-m/up_unblocktask.c
@@ -155,6 +155,9 @@ void up_unblock_task(struct tcb_s *tcb)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/up_blocktask.c
+++ b/os/arch/arm/src/armv8-m/up_blocktask.c
@@ -169,6 +169,9 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/up_mpu.c
+++ b/os/arch/arm/src/armv8-m/up_mpu.c
@@ -306,15 +306,15 @@ uint32_t mpu_subregion(uintptr_t base, size_t size, uint8_t l2size)
 }
 
 /****************************************************************************
- * Name: mpu_configure_app_regs
+ * Name: mpu_get_register_value
  *
  * Description:
  *   Configure the user application SRAM mpu settings into the tcb variables
  *
  ****************************************************************************/
 
-#if defined(CONFIG_ARMV8M_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)
-void mpu_configure_app_regs(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute)
+#if (defined(CONFIG_ARM_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+void mpu_get_register_value(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute)
 {
 	uint32_t regval;
 	uint8_t l2size;
--- a/os/arch/arm/src/armv8-m/up_releasepending.c
+++ b/os/arch/arm/src/armv8-m/up_releasepending.c
@@ -133,6 +133,9 @@ void up_release_pending(void)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/up_reprioritizertr.c
+++ b/os/arch/arm/src/armv8-m/up_reprioritizertr.c
@@ -184,6 +184,9 @@ void up_reprioritize_rtr(struct tcb_s *tcb, uint8_t priority)
 				/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 				up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+				up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 				/* Update rtcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/up_svcall.c
+++ b/os/arch/arm/src/armv8-m/up_svcall.c
@@ -262,6 +262,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 		up_set_mpu_app_configuration(tcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		up_set_mpu_stack_guard(tcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 		/* Update tcb active flag for monitoring. */
@@ -300,6 +303,9 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 		up_set_mpu_app_configuration(tcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		up_set_mpu_stack_guard(tcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 		/* Update tcb active flag for monitoring. */
--- a/os/arch/arm/src/armv8-m/up_unblocktask.c
+++ b/os/arch/arm/src/armv8-m/up_unblocktask.c
@@ -155,6 +155,9 @@ void up_unblock_task(struct tcb_s *tcb)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV8M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(rtcb);
+#endif
 #endif
 #ifdef CONFIG_TASK_MONITOR
 			/* Update rtcb active flag for monitoring. */
--- a/os/binfmt/binfmt_execmodule.c
+++ b/os/binfmt/binfmt_execmodule.c
@@ -83,7 +83,6 @@
 
 #ifdef CONFIG_ARM_MPU
 extern uint32_t g_mpu_region_nr;
-void mpu_configure_app_regs(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute);
 #endif
 
 /****************************************************************************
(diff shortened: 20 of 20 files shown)


seinfra:
Target : [c42c6f89ca3d21484a4841000e7b58c8755c870a] - Code Rule Check (C++) OK.

seinfra:
Target : [c42c6f89ca3d21484a4841000e7b58c8755c870a] - Code Rule Check Result:
os/arch/arm/src/common/up_createstack.c:260: ERROR: [SPC_M_SEP] space prohibited after that open parenthesis '('
os/arch/arm/src/common/up_createstack.c:260: ERROR: [SPC_M_SEP] space prohibited before that close parenthesis ')'

os/Kconfig
```diff
@@ -241,6 +241,15 @@ config AUTOGEN_MEMORY_LDSCRIPT
 
 endif # BUILD_PROTECTED
 
+config MPU_STACKGUARD
+	bool "Stack Overflow Protection"
+	depends on APP_BINARY_SEPARATION
```
sunghan-chang, line 246: Could you let me know why this has a dependency with APP_BINARY_SEPARATION?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +252,18 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACKGUARD
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+
+		tcb->stack_guard =	(uint32_t)tcb->stack_alloc_ptr + 0x40;
```
sunghan-chang, line 258: Could you check the alignment?

os/kernel/task/task_setup.c
```diff
@@ -468,6 +468,11 @@ static int thread_schedsetup(FAR struct tcb_s *tcb, int priority, start_t start,
 		for (; i < 3 * MPU_NUM_REGIONS; i += 3)
 #endif
 		{
+#ifdef CONFIG_MPU_STACKGUARD
+			if (i == 9) {	/* skip already assigned stack protection */
```
sunghan-chang, line 472: Why is this 9?

os/Kconfig
```diff
@@ -241,6 +241,15 @@ config AUTOGEN_MEMORY_LDSCRIPT
 
 endif # BUILD_PROTECTED
 
+config MPU_STACKGUARD
```
sunghan-chang, line 244: This implementation is ARM-specific. Please move this in Kconfig of ARM.

os/Kconfig
kishore-sn, line 246: I think this does not depend on APP_BINARY_SEPARATION, but it depends on PROTECTED build. 
In flat build it is not possible to provide protection since everything runs in previleged mode and we cannot restrict permission using MPU. But, in portected and app separation cases, we can protect the stack of application tasks by using this method.

os/Kconfig
```diff
@@ -241,6 +241,15 @@ config AUTOGEN_MEMORY_LDSCRIPT
...
+	bool "Stack Overflow Protection"
+	depends on APP_BINARY_SEPARATION
+	depends on ARCH_USE_MPU
+	default y
+	---help---
+		Enables stack overflow protection.
+		It protects each task's stack overflow outside its max allocated stack memory.
```
kishore-sn, line 251: Please mention that this is applicable only to user mode tasks.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -672,7 +672,7 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 		return;
 	}
 
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#if defined(CONFIG_OPTIMIZE_APP_RELOAD_TIME) || defined(CONFIG_MPU_STACKGUARD)
```
kishore-sn, line 675: Since, this is applicable to protected build also, its better to make a separate API for stack protection setting, and then call this new api from all the context switch places.

os/arch/arm/src/common/up_createstack.c
kishore-sn, line 258: It would be better to define a new alignment directive like STACK_ALIGN_UP to perform the proper 32 byte alignment.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +252,18 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+		tcb->stack_guard = ((uint32_t)tcb->stack_guard & MPU_RBAR_ADDR_MASK);
+		tcb->guard_size = ((uint32_t)tcb->stack_guard - (uint32_t)tcb->stack_alloc_ptr);
```
kishore-sn, line 260: The guard size will not change based on the guard start address. It must always be equal to 32 bytes.
So, instead of using a variable like guard_size, we can use a MACRO to give the size.

kishore-sn, line 260: I think we need to update the tcb->adj_stack_size based on this calculation.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +252,18 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+		mpu_configure_app_regs(&tcb->mpu_regs[9], g_mpu_region_nr + 3, tcb->stack_guard, tcb->guard_size, true, false);
+#else
+		mpu_configure_app_regs(&tcb->mpu_regs[3], g_mpu_region_nr + 1, tcb->stack_guard, tcb->guard_size, true, false);
+#endif
```
kishore-sn, line 265: The use of g_mpu_region_nr to hold the region number is becoming too much complex and difficult to maintain. It would be better if we can declare an enum with the different regions and use the enum names here. However, since this change will impact stack protection, app separation and common binary implementations, I think we must merge the current changes and then make a new PR for having the enum implementation.

os/arch/arm/src/common/up_createstack.c
sangam-swami: updated

sangam-swami: We need to take care of 32 byte aligned address as well as 32 byte size for the region. 
The aligned address will result into size from 32 byte to max 64 byte.

os/Kconfig
sangam-swami: I have developed the feature initially with loadable apps configs to verify the task protection inside apps. Now I have modified to support the protected build binary, so it will intern support the loadable configs.

os/arch/arm/src/common/up_createstack.c
sangam-swami: done

os/kernel/task/task_setup.c
sangam-swami: It is the register address to be updated for MPU region number 3. 
Each MPU region will be enabled with changing minimum of 3 registers.
I have modified this as generic in my latest changes submitted

os/Kconfig
sangam-swami: done

os/Kconfig
sangam-swami: done

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: changed the function for the mpu setting.

os/arch/arm/src/common/up_createstack.c
sangam-swami: added new local alignment directive

os/arch/arm/src/common/up_createstack.c
sangam-swami: I have changed the stack overflow protection to be generic.  
We will create new PR which will include modification of loadable apps MPU configuration to generic method

os/include/tinyara/sched.h
```diff
@@ -86,18 +86,24 @@
 /********************************************************************************
  * Pre-processor Definitions
  ********************************************************************************/
-#if defined(CONFIG_APP_BINARY_SEPARATION) && defined(CONFIG_ARM_MPU)
+#if defined(CONFIG_ARM_MPU) && (defined(CONFIG_APP_BINARY_SEPARATION) || defined (CONFIG_MPU_STACKGUARD))
+#define REGS_PER_REGION 3
```
sunghan-chang, line 90: One question.
Because the enum structure has 3 elements, is this defined as 3?
If it is,
```
enum {
    REG_RNR = 0,
    REG_RBAR,
    REG_RASR,
    REG_MAX
};
```
let's use `REG_MAX` instead of new definition. I don't want to add duplicated manual definition .

os/include/tinyara/sched.h
```diff
@@ -86,18 +86,24 @@
...
 enum {
 	REG_RNR,
 	REG_RBAR,
 	REG_RASR
 };
 
+#if defined(CONFIG_MPU_STACKGUARD)
+#define STACKGUARD_MPU_NUM		3	/* Region number 3 is reserved for stack protection */
```
sunghan-chang, line 98: Should this be hard-coded? If someone uses region 3 with another purpose, then what will be happened?
@kishore-sn @sangam-swami I've requested you to remove the manual allocation of region number for mpu functions at previous PR and you said you would do that. I don't accept the hard-coded mpu region at somewhere.

Regardless of above, how about moving mpu definitions at mpu.h, not in sched.h?

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -687,6 +687,29 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
...
+
+	if (rtcb->mpu_regs[i + REG_RBAR]) {
+		putreg32(rtcb->mpu_regs[i + REG_RNR], MPU_RNR);
+		putreg32(rtcb->mpu_regs[i + REG_RBAR], MPU_RBAR);
+		putreg32(rtcb->mpu_regs[i + REG_RASR], MPU_RASR);
+	}
+
+}
+#else
+#define up_set_mpu_region_configuration(x, y)
```
sunghan-chang, line 710: When you want to call this function without the conditional, this is right. But you call this with conditional at every locations, then this is unnecessary.

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -687,6 +687,29 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 #define up_set_mpu_app_configuration(x)
 #endif
 
+#if defined(CONFIG_MPU_STACKGUARD)
+static inline void up_set_mpu_region_configuration(struct tcb_s *rtcb, int region_no)
```
sunghan-chang, line 691: 1. Could you rename this function? I can't know what the difference against other mpu function.
2. Other mpu function has the region number as 1st argument. But this isn't. And it has different argument name for same purpose. Could you use the same function name policy?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -106,6 +110,12 @@
 #define STACK_ALIGN_DOWN(a) ((a) & ~STACK_ALIGN_MASK)
 #define STACK_ALIGN_UP(a)   (((a) + STACK_ALIGN_MASK) & ~STACK_ALIGN_MASK)
 
+#ifdef CONFIG_MPU_STACKGUARD
+#define STACK_GUARD_ADDR_MASK	0xffffffe0	/* Bits N-31:  Used for MPU Region base addrese */
+extern uint32_t g_mpu_region_nr;
+void mpu_configure_app_regs(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute);
```
sunghan-chang, line 116: Let's move the definition of prototype in header and let's just include the header.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +253,16 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACKGUARD
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+		tcb->stack_guard = (uint32_t)tcb->stack_alloc_ptr + 0x40;
```
sunghan-chang, line 258: Could you let me know the meaning of `0x40`? Could you use the definition to know what it is easily instead of the value?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +253,16 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+		tcb->stack_guard = ((uint32_t)tcb->stack_guard & STACK_GUARD_ADDR_MASK);
```
sunghan-chang, line 259: Where is the defintion of `stack_guard` in tcb structure?

os/include/tinyara/sched.h
```diff
@@ -626,15 +632,20 @@ struct tcb_s {
...
+#if defined(CONFIG_APP_BINARY_SEPARATION)
 	uint32_t ram_start;		/* Start address of RAM partition for this app */
 	uint32_t ram_size;		/* Size of RAM partition for this app */
 	uint32_t uspace;		/* User space object for app binary */
 	uint32_t uheap;			/* User heap object pointer */
-
 #ifdef CONFIG_ARM_MPU
-	uint32_t mpu_regs[3 * MPU_NUM_REGIONS];	/* We need 3 register values to configure each MPU region */
+#ifdef CONFIG_MPU_STACKGUARD
+	uint32_t mpu_regs[3 * (MPU_NUM_REGIONS + 1)]; /* We need 3 register values to configure each MPU region */
```
sunghan-chang, line 642: This has the same comment as below, but it has different the element number of array. Please write the right comment.

os/include/tinyara/sched.h
```diff
@@ -626,15 +632,20 @@ struct tcb_s {
...
+#else
+	uint32_t mpu_regs[3 * MPU_NUM_REGIONS]; /* We need 3 register values to configure each MPU region */
+#endif
 #endif
+#elif defined(CONFIG_MPU_STACKGUARD)
```
sunghan-chang, line 647: Why is this case necessary?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +253,16 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+		tcb->guard_size = ((uint32_t)tcb->stack_guard - (uint32_t)tcb->stack_alloc_ptr);
+		mpu_configure_app_regs(&tcb->mpu_regs[STACKGUARD_MPU_NUM * REGS_PER_REGION], STACKGUARD_MPU_NUM, tcb->stack_guard, tcb->guard_size, true, false);
```
sunghan-chang, line 261: Is this added to set MPU region for stack guard? What's the difference between calling this and calling up_set_mpu_region_configuration?

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,15 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config MPU_STACKGUARD
+	bool "Stack Overflow Protection"
```
kishore-sn, line 275: @sunghan-chang 
I had seen your comment yesterday. But, i was wondering why this is ARM specific config? It could be supported on any arch having MPU, right?

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -672,7 +672,7 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 		return;
 	}
 
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#if defined(CONFIG_OPTIMIZE_APP_RELOAD_TIME)
```
kishore-sn, line 675: We can revert this change.

os/arch/arm/src/armv8-m/mpu.h
kishore-sn, line 691: If we are going to use this API only for stack protection, then we can rename it as up_set_mpu_stack_gaurd

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -687,6 +687,29 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
 #define up_set_mpu_app_configuration(x)
 #endif
 
+#if defined(CONFIG_MPU_STACKGUARD)
+static inline void up_set_mpu_region_configuration(struct tcb_s *rtcb, int region_no)
```
kishore-sn, line 691: If we are going to use this API only for stack protection, then we can rename it as up_set_mpu_stack_gaurd

os/arch/arm/src/common/up_createstack.c
kishore-sn, line 258: I am not sure why we need to do `tcb->stack_alloc_ptr + 0x40`.
But, I have an alternative suggestion to have 32 byte aligned address. We can use memalign instead of malloc as follows:

`tcb->stack_alloc_ptr = (uint32_t *)kmm_memalign(32, stack_size);`

os/arch/arm/src/common/up_createstack.c
kishore-sn, line 259: This line is not required since we are doing this in mpu_configure_app_regs() at line num 336 in up_mpu.c.

os/arch/arm/src/common/up_createstack.c
kishore-sn: Right. Size must always be 32 bytes. 
But, according to above code, size will be `((uint32_t)tcb->stack_guard - (uint32_t)tcb->stack_alloc_ptr)` which is not always 32 bytes.

os/include/tinyara/sched.h
```diff
@@ -86,18 +86,24 @@
...
-#if defined(CONFIG_APP_BINARY_SEPARATION) && defined(CONFIG_ARM_MPU)
+#if defined(CONFIG_ARM_MPU) && (defined(CONFIG_APP_BINARY_SEPARATION) || defined (CONFIG_MPU_STACKGUARD))
+#define REGS_PER_REGION 3
 enum {
 	REG_RNR,
 	REG_RBAR,
 	REG_RASR
 };
 
+#if defined(CONFIG_MPU_STACKGUARD)
```
kishore-sn, line 97: We already checked for CONFIG_MPU_STACKGUARD at line 89. So, this check is not required.

os/include/tinyara/sched.h
kishore-sn, line 98: This would get overwritten by below defines.

os/include/tinyara/sched.h
```diff
@@ -626,15 +632,20 @@ struct tcb_s {
 
 	struct xcptcontext xcp;		/* Interrupt register save area        */
 
-#ifdef CONFIG_APP_BINARY_SEPARATION
+#if defined(CONFIG_APP_BINARY_SEPARATION)
```
kishore-sn, line 635: Why is this change required?

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes size will always be >= 32, to make both address and size aligned to 32 byte

sangam-swami: calculation is changed after adding kumm_memalign

os/include/tinyara/sched.h
sangam-swami: changed

os/include/tinyara/sched.h
sangam-swami: Hi @sunghan-chang ,
I agree for the above request.
Lets this PR will be for stack protection?
I will be modifying the above request with new commit/PR separately. 
It will include the changes in mpu interface related APP_BINARY_SEPARATION MPU  as well as STACKGUARD.

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: done

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: done

os/arch/arm/src/common/up_createstack.c
sangam-swami: done

os/arch/arm/src/common/up_createstack.c
sangam-swami: This is for the memory address alignment needed for MPU region. I have modified with macro now. Reserved stack allocation with 32 byte align. Adding the minimum size required for mpu allocation now (32 bytes).

os/arch/arm/src/common/up_createstack.c
sangam-swami: Added comments in tcb structure..is it the one you referring?

os/include/tinyara/sched.h
sangam-swami: changed

os/include/tinyara/sched.h
sangam-swami: It was added to support Protected build ..simplified with latest changes

os/arch/arm/src/common/up_createstack.c
sangam-swami: mpu_configure_app_regs will just keep the mpu data structures ready in the tcb structure variable.
up_set_mpu_region_configuration will actually set those values in MPU (It will be done during  context switch, scheduling task)

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: done

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: changed

os/arch/arm/src/common/up_createstack.c
sangam-swami: Modified to reserve memory with kmm_memalign. After this the size will be fixed to 32 byte now.

os/arch/arm/src/common/up_createstack.c
sangam-swami: after aligned memory reserved for stack, this change is not needed, so removed.

os/include/tinyara/sched.h
sunghan-chang: ok. thank you

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +260,14 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACKGUARD
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+		tcb->stack_guard = (uint32_t)tcb->stack_alloc_ptr + MPU_REGION_SIZE_ALIGNMENT;
```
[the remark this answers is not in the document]
sunghan-chang: Does `MPU_REGION_SIZE_ALIGNMENT` have two usages?
1. to align for MPU setting
2. to give the size of stack guard
If it is, let's define another definition for stack guard size even it could be the same as alignment.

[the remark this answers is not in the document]
sunghan-chang: And don't you need to malloc that area to prevent allocation and used?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +260,14 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+		mpu_configure_app_regs(&tcb->mpu_regs[STACKGUARD_MPU_NUM * REG_MAX], STACKGUARD_MPU_NUM, tcb->stack_guard, MPU_REGION_SIZE_ALIGNMENT, true, false);
+		/* Adjust the adj_stack_size after guard_size calculationn */
+		tcb->adj_stack_size = tcb->adj_stack_size - MPU_REGION_SIZE_ALIGNMENT;
```
sunghan-chang, line 268: Why do you subtract `MPU_REGION_SIZE_ALIGNMENT`?

os/include/tinyara/sched.h
```diff
@@ -626,14 +631,18 @@ struct tcb_s {
...
 	uint32_t ram_size;		/* Size of RAM partition for this app */
 	uint32_t uspace;		/* User space object for app binary */
 	uint32_t uheap;			/* User heap object pointer */
+#endif
 
-#ifdef CONFIG_ARM_MPU
-	uint32_t mpu_regs[3 * MPU_NUM_REGIONS];	/* We need 3 register values to configure each MPU region */
+#if defined(CONFIG_ARM_MPU) && (defined(CONFIG_APP_BINARY_SEPARATION) || defined(CONFIG_MPU_STACKGUARD))
+#ifdef CONFIG_MPU_STACKGUARD
+	uint32_t mpu_regs[3 * (STACKGUARD_MPU_NUM + 1)]; /* Using 4 mpu regions,as stackguard region is fixed temporarily*/
```
sunghan-chang, line 643: When we configure as below,
1. CONFIG_ARM_MPU and CONFIG_MPU_STACKGUARD are enabled
2. CONFIG_APP_BINARY_SEPARATION is disabled

Do we use 3 mpu regions regardless of mpu stackguard?
Oh, that's why you add calling `mpu_configure_app_regs` function at up_create_stack function. 
But is there any pros using 3 MPU region with protected build without APP_BINARY_SEPARATION?

sunghan-chang:
@sangam-swami Could you check the first commit? Description is not separated from title.

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -687,6 +699,35 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
...
+	if ((rtcb->flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_KERNEL) {
+		return;
+	}
+
+	if (rtcb->mpu_regs[i + REG_RBAR]) {
+		putreg32(rtcb->mpu_regs[i + REG_RNR], MPU_RNR);
+		putreg32(rtcb->mpu_regs[i + REG_RBAR], MPU_RBAR);
+		putreg32(rtcb->mpu_regs[i + REG_RASR], MPU_RASR);
+	}
+
```
intom, line 727: I think this new line does not need.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -106,6 +110,11 @@
 #define STACK_ALIGN_DOWN(a) ((a) & ~STACK_ALIGN_MASK)
 #define STACK_ALIGN_UP(a)   (((a) + STACK_ALIGN_MASK) & ~STACK_ALIGN_MASK)
 
+#ifdef CONFIG_MPU_STACKGUARD
+#define STACK_GUARD_ADDR_MASK	0xffffffe0	/* Bits N-31:  Used for MPU Region base addrese */
+#define MPU_REGION_SIZE_ALIGNMENT 32			/* Min MPU region is 32 bytes,regions must be 32 byte aligned address */
```
intom, line 115: You should adjust tab for comment.

sangam-swami: done

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: done

sangam-swami:
> @sangam-swami Could you check the first commit? Description is not separated from title.

Modified

os/include/tinyara/sched.h
sangam-swami: in the case mentioned above it was protected build enabled, we use total 3 MPU regions including stack guard memory protection region (1.mpu_userflash, 2.mpu_userintsram, 3.stack guard mpu region).
The region number for stack guard is fixed now in this PR. 
When there is dynamic mpu regions support is added, it will not be hard coded and this part of structure will change.

os/arch/arm/src/common/up_createstack.c
sangam-swami: The exact stack size available for task will be reduced by 32 byte because of guard memory used for protection.

os/arch/arm/src/common/up_createstack.c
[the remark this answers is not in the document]
sangam-swami: I removed new member in tcb and added the fixed value.

Does MPU_REGION_SIZE_ALIGNMENT have two usages? => yes one is for MPU region min size and other is for 32 byte address alignment. I defined differently now.

And don't you need to malloc that area to prevent allocation and used? =>stack guard will be part of the stack memory allocated for the task.

os/arch/arm/src/common/up_createstack.c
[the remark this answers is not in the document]
sangam-swami: changed

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,15 @@ config ARCH_DABORTSTACK
...
+	depends on BUILD_PROTECTED
+	default y
```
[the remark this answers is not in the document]
sangam-swami: ok

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,15 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config MPU_STACKGUARD
```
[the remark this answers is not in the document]
sangam-swami: 1. Changed the name
2. added ARM_MPU as dependency.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -215,6 +215,18 @@ uint8_t mpu_log2regionfloor(uintptr_t base, size_t size);
...
 
+/****************************************************************************
+ * Name: mpu_configure_app_regs
+ *
+ * Description:
+ *   Configure the user application SRAM mpu settings into the tcb variables
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_ARMV7M_MPU) && (defined(CONFIG_APP_BINARY_SEPARATION) || defined(CONFIG_MPU_STACKGUARD))
```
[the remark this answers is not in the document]
sangam-swami: yes.. removed

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -160,6 +160,12 @@
 #define MPU_RASR_AP_RORO      (6 << MPU_RASR_AP_SHIFT)	/* P:RO   U:RO   */
 #define MPU_RASR_XN             (1 << 28)	/* Bit 28: Instruction access disable */
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+#define STACK_GUARD_ADDR_MASK	0xffffffe0	/* Bits N-31:  Used for MPU Region base addrese */
```
kishore-sn, line 164: STACK_GUARD_ADDR_MASK is not used anywhere. Please remove it.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -160,6 +160,12 @@
...
+#define MPU_REGION_ADDRESS_ALIGNMENT 32	/* MPU regions must be 32 byte address aligned */
```
kishore-sn, line 165: Can you pls rename these 2 defines to indicate it is related to stack gaurd? Maybe like MPU_STACK_REGION_ALIGNMENT ?

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -160,6 +160,12 @@
...
+#define MPU_REGION_SIZE_ALIGNMENT 32		/* Min MPU region size is 32 bytes */
```
kishore-sn, line 166: In this define, we dont need to mention ALIGNMENT, because, this define indicates the size of MPU stack guard and not its alignment.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +255,14 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+		mpu_configure_app_regs(&tcb->mpu_regs[STACKGUARD_MPU_NUM * REG_MAX], STACKGUARD_MPU_NUM,
+			(uint32_t)tcb->stack_alloc_ptr + MPU_REGION_SIZE_ALIGNMENT, MPU_REGION_SIZE_ALIGNMENT, true, false);
```
kishore-sn, line 261: Here, the third arguement (region base address) must be tcb->stack_alloc_ptr only. (not tcb->stack_alloc_ptr + MPU_REGION_SIZE_ALIGNMENT)

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: changed

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: The address is aligned because of MPU address alignment issue, so I think we can keep same or even move outside of stack overflow feature.

os/arch/arm/src/common/up_createstack.c
sangam-swami: I have removed adding   MPU_REGION_SIZE_ALIGNMENT to the stack_alloc_ptr, as it is lower memory address. MPU region will set 32 bytes from this lower memory towards top of stack.

os/arch/arm/src/armv7-m/up_blocktask.c
```diff
@@ -169,6 +169,9 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 			/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
 			up_set_mpu_app_configuration(rtcb);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+			up_set_mpu_stack_guard(STACKGUARD_MPU_NUM, rtcb);
```
sunghan-chang, line 173: 1. This API is dedicated to set MPU region for stack. Do we need to give `STACKGUARD_MPU_NUM` even it is fixed in header?
2. In SAM (Software Architecture Maturity), it checks duplication of codes. This is almost the same as `up_set_mpu_app_configuration`. Do we need to split them?

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -160,6 +160,11 @@
 #define MPU_RASR_AP_RORO      (6 << MPU_RASR_AP_SHIFT)	/* P:RO   U:RO   */
 #define MPU_RASR_XN             (1 << 28)	/* Bit 28: Instruction access disable */
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+#define MPU_REGION_ADDRESS_ALIGNMENT 32	/* MPU regions must be 32 byte address aligned */
+#define MPU_REGION_MINIMUM_SIZE 32		/* Min MPU region size is 32 bytes */
```
sunghan-chang, line 165: Do you want to say `minimum size` or `stack guard size`? If it says minimum size, why is it in `CONFIG_MPU_STACK_OVERFLOW_PROTECTION`?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -180,12 +184,20 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		/* Use the kernel allocator if this is a kernel thread */
 
 		if (ttype == TCB_FLAG_TTYPE_KERNEL) {
-			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_size);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION	/* address must be aligned to 32 byte for MPU region */
+			tcb->stack_alloc_ptr = (uint32_t *)kmm_memalign(MPU_REGION_ADDRESS_ALIGNMENT, stack_size);
```
sunghan-chang, line 188: In description of MPU_STACK_OVERFLOW_PROTECTION, you say `This is applicable to user mode tasks created in protected build.`. Then, why is this necessary? This is for kernel thread.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -243,6 +255,14 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		heapinfo_exclude_stacksize(tcb->stack_alloc_ptr);
 #endif
 		board_led_on(LED_STACKCREATED);
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		/* The smallest size that can be programmed for an MPU region is 32 bytes */
+		mpu_configure_app_regs(&tcb->mpu_regs[STACKGUARD_MPU_NUM * REG_MAX], STACKGUARD_MPU_NUM,
+			(uint32_t)tcb->stack_alloc_ptr, MPU_REGION_MINIMUM_SIZE, true, false);
+		/* Adjust the adj_stack_size after guard_size calculationn */
+		tcb->adj_stack_size = tcb->adj_stack_size - MPU_REGION_MINIMUM_SIZE;
```
sunghan-chang, line 263: You allocate `stack_size` user requested. If we need stack guard, don't we need to add the guard size?
I mean
```
tcb->stack_alloc_ptr = kmm_....(stack_size + MPU_REGION_MINIMUM_SIZE);
tcb->adj_stack_size = tcb->adj_stack_size - MPU_REGION_MINIMUM_SIZE;
```

sunghan-chang: Whatever we support, the most important is we should guarantee user request. Even user want `stack_size`, we always return `stack_size - MPU_REGION_MINIMUM_SIZE`. I don't think this is good.

os/include/tinyara/sched.h
sunghan-chang: > in the case mentioned above it was protected build enabled, we use total 3 MPU regions including stack guard memory protection region (1.mpu_userflash, 2.mpu_userintsram, 3.stack guard mpu region).

There is no setting of `1.mpu_userflash, 2.mpu_userintsram` for protected build. And I don't know that is necessary. So to use the stack guard region, it spends much more memory. This must be modified when you change it as dynamic as you said.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -215,6 +220,18 @@ uint8_t mpu_log2regionfloor(uintptr_t base, size_t size);
...
+/****************************************************************************
+ * Name: mpu_configure_app_regs
+ *
+ * Description:
+ *   Configure the user application SRAM mpu settings into the tcb variables
+ *
+ ****************************************************************************/
+
+#if defined(CONFIG_APP_BINARY_SEPARATION) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+void mpu_configure_app_regs(uint32_t *regs, uint32_t region, uintptr_t base, size_t size, uint8_t readonly, uint8_t execute);
```
sunghan-chang, line 232: The name could be confused `mpu_configure_app_regs` and `up_set_mpu_xxxx`. How about renaming `mpu_configure_app_regs`?

mpu_get_register_value
mpu_make_register_value
mpu_value2register
...

os/arch/arm/src/armv8-m/mpu.h
jeongchanKim, line 165: If it is for stack guard, I think we can give this as config.

os/arch/arm/src/armv7-m/mpu.h
```diff
@@ -687,6 +704,35 @@ static inline void up_set_mpu_app_configuration(struct tcb_s *rtcb)
...
+
+#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+static inline void up_set_mpu_stack_guard(uint32_t region, struct tcb_s *rtcb)
+{
+	/* We update the MPU registers only if:
+	 * This is not a kernel thread AND
+	 * It has a non zero value of base address (This ensures valid MPU setting)
+	 */
+	int i = 3 * region;
+	if ((rtcb->flags & TCB_FLAG_TTYPE_MASK) == TCB_FLAG_TTYPE_KERNEL) {
```
jeongchanKim, line 723: Is there any special reason not to support kernel thread?

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,16 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config MPU_STACK_OVERFLOW_PROTECTION
+	bool "MPU Stack Overflow Protection"
+	depends on BUILD_PROTECTED
```
jeongchanKim, line 276: Can we support mpu stackoverflow protection with flat build?

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -160,6 +160,11 @@
 #define MPU_RASR_AP_RORO      (6 << MPU_RASR_AP_SHIFT)	/* P:RO   U:RO   */
 #define MPU_RASR_XN             (1 << 28)	/* Bit 28: Instruction access disable */
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+#define MPU_REGION_ADDRESS_ALIGNMENT 32	/* MPU regions must be 32 byte address aligned */
```
manoharahk, line 164: 1. Comment can be modified for better clarity . MPU Regions must be size aligned though minimum size possible is 32 bytes.. 
2. MPU_REGION_ADDRESS_ALIGNMENT is quiet generic. IMO, a specific one like MPU_REGION_STACKGAURD_ADDRESS_ALIGNMENT  would be good

os/include/tinyara/sched.h
```diff
@@ -86,18 +86,24 @@
...
 enum {
 	REG_RNR,
 	REG_RBAR,
-	REG_RASR
+	REG_RASR,
+	REG_MAX
 };
 
+#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+#define STACKGUARD_MPU_NUM		3	/* Region number 3 is reserved for stack protection */
```
manoharahk, line 98: STACKGUARD_MPU_REGION_NUM would be good.

os/arch/arm/src/armv7-m/up_blocktask.c
sangam-swami: 1. Yes removed passing region number to the function
2. up_set_mpu_app_configuration function configuring 3 regions , need to change this function to use as general. 
Once we remove hard coded MPU numbers, we can use common function

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: I have added stack guard size as configurable now

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes, removed the kernel thread stack creation changes

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes, modified and allocated user requested size

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: modified the name to mpu_get_register_value, to be stored to local tcb array.

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: yes added in config

os/arch/arm/src/armv7-m/mpu.h
sangam-swami: currently kernel is in privilege mode, so we need to disable it and MPU to kernel region.
There will be issues for accessing user space and other unknown issues as of now.

os/arch/arm/Kconfig
sangam-swami: in flat build kernel is in privilege mode, so we need to disable it and add MPU to kernel region.
I will explore the feasibility for this.

os/arch/arm/src/armv8-m/mpu.h
sangam-swami: 1. It is 32 byte address aligned along with 32 byte size aligned for MPU.
2. It is generic macro for MPU, so removed out of stack overflow define

os/include/tinyara/sched.h
sangam-swami: yes changed

os/arch/arm/src/common/up_createstack.c
```diff
@@ -180,12 +184,17 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		/* Use the kernel allocator if this is a kernel thread */
 
 		if (ttype == TCB_FLAG_TTYPE_KERNEL) {
-			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_size);
+			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_size);;
```
NitishAmbastha, line 187: please fix typo for double semicolon

os/include/tinyara/mpu.h
```diff
@@ -0,0 +1,53 @@
+/****************************************************************************
+ *
+ * Copyright 2016 Samsung Electronics All Rights Reserved.
```
NitishAmbastha, line 3: fix the year

os/include/tinyara/mpu.h
```diff
@@ -0,0 +1,53 @@
...
+/********************************************************************************
+ * Included Files
+ ********************************************************************************/
+
+#include <tinyara/config.h>
+
+/********************************************************************************
+ * Pre-processor Definitions
+ ********************************************************************************/
+#if defined(CONFIG_ARM_MPU) && (defined(CONFIG_APP_BINARY_SEPARATION) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION))
```
NitishAmbastha, line 33: please recheck the config checks
CONFIG_MPU_STACK_OVERFLOW_PROTECTION already depends on ARM_MPU and BUILD_PROTECTED

os/include/tinyara/sched.h
```diff
@@ -80,25 +80,10 @@
 #include <tinyara/mm/shm.h>
 #include <tinyara/fs/fs.h>
 #include <tinyara/net/net.h>
+#include <tinyara/mpu.h>
```
NitishAmbastha, line 83: I think header mpu.h should be included under CONFIG_MPU_STACK_OVERFLOW_PROTECTION check

os/arch/arm/src/common/up_createstack.c
NitishAmbastha, line 187: please fix typo for double semicolon

os/include/tinyara/mpu.h
sangam-swami: updated

os/include/tinyara/mpu.h
sangam-swami: yes modified

os/include/tinyara/sched.h
sangam-swami: I have created mpu.h to include generic definitions, it also include the usage for app_binary_separation, stack overflow.

os/arch/arm/src/common/up_createstack.c
sangam-swami: done

os/include/tinyara/sched.h
sangam-swami: I have created mpu.h to include generic definitions, it also include the usage for app_binary_separation, stack overflow.

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,28 @@ config ARCH_DABORTSTACK
...
+		This is applicable to user mode tasks created in protected build.
+
+if MPU_STACK_OVERFLOW_PROTECTION
+
+config MPU_STACK_GUARD_SIZE
+	int "Stack Guard size configuration range(32-1024)"
+	default 32
+	range 32 1024
+	---help---
+		Set size for the stack guard.
```
jeongchanKim, line 291: For stack guard, its size should be aligned. Please add description about this.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -185,7 +189,12 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 #endif
 		{
 			/* Use the user-space allocator if this is a task or pthread */
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION	/* address must be aligned to 32 byte for MPU region */
+			stack_size = stack_size + CONFIG_MPU_STACK_GUARD_SIZE;
```
jeongchanKim, line 193: `stack_size` is a param variable, We'd better declare one local variable.

os/arch/arm/src/common/up_createstack.c
```diff
@@ -185,7 +189,12 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+			tcb->stack_alloc_ptr = (uint32_t *)kumm_memalign(MPU_REGION_ADDRESS_ALIGNMENT, stack_size);
```
jeongchanKim, line 194: As my understanding, guard should be aligned to its size. But you made alignment with guard + stack size. Is it okay?

os/arch/arm/Kconfig
sangam-swami: changed

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes, used local variable for all the references

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes, it if fine, we are aligning the address assigned for the total size including guard memory.
Guard address will be the stack_alloc_ptr address which will be aligned to guard size/mpu region size.

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,28 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config MPU_STACK_OVERFLOW_PROTECTION
+	bool "MPU Stack Overflow Protection"
+	depends on BUILD_PROTECTED
+	depends on ARM_MPU
```
jeongchanKim, line 277: I think you can make dependency only with ARMv7m.
Because, as I know in ARMv8m, there is a stack limit register which can be used for stack overflow protection. So we don't need to use MPU.

os/include/tinyara/mpu.h
```diff
@@ -0,0 +1,54 @@
...
+};
+
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#define MPU_NUM_REGIONS		3
+#else
+#define MPU_NUM_REGIONS		1
+#endif
+#endif
+
+#define MPU_REGION_ADDRESS_ALIGNMENT 32	/* MPU regions must be 32 byte aligned address */
```
sunghan-chang, line 48: There is no code using this definition.

os/arch/arm/Kconfig
```diff
@@ -271,6 +271,28 @@ config ARCH_DABORTSTACK
...
+		This is applicable to user mode tasks created in protected build.
+
+if MPU_STACK_OVERFLOW_PROTECTION
+
+config MPU_STACK_GUARD_SIZE
+	int "Stack Guard size configuration range(32-1024)"
+	default 32
+	range 32 1024
+	---help---
+		Set size for the stack guard. It must be set with 32 byte aligned values.
```
sunghan-chang, line 291: To check `32 bytes aligned value`, we should add the condition in mkconfig.h

os/arch/arm/Kconfig
sangam-swami: we will enable this stack overflow feature for both the architectures. After development of the ARMv8m stack limit register feature, we can disable it/remove it

os/include/tinyara/mpu.h
sangam-swami: removed

os/tools/mkconfig.c
```diff
@@ -305,6 +305,16 @@ int main(int argc, char **argv, char **envp)
 	printf("# undef CONFIG_DEBUG_SPI\n");
 	printf("# undef CONFIG_DEBUG_HEAP\n");
 	printf("#endif\n\n");
+	printf("#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)\n");
+	printf("# if (CONFIG_MPU_STACK_GUARD_SIZE != 32) || \\\n");
+	printf("    (CONFIG_MPU_STACK_GUARD_SIZE != 64) || \\\n");
```
sunghan-chang, line 310: It should be &&

sangam-swami: modified , thanks

os/arch/arm/Kconfig
sangam-swami: added checking condition

PR #4257, opened 2020-04-22
Add #ifdef condition for avoiding build error.

jeongchanKim:
apps/example/testcase/fs_main : Add #ifdef for local variable in vfs unlink tc
apps/shell/tash_main : Add #ifdef for tash_clear_line function

### DIFF
--- a/apps/shell/tash_main.c
+++ b/apps/shell/tash_main.c
@@ -53,6 +53,7 @@ enum tash_input_state_e {
 
 static int tash_running = FALSE;
 
+#if TASH_MAX_STORE > 0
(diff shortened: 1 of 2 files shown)


sangam-swami:
Can you please mention the failed build configuration in the commits

apps/shell/tash_main.c
```diff
@@ -53,6 +53,7 @@ enum tash_input_state_e {
 
 static int tash_running = FALSE;
 
+#if TASH_MAX_STORE   > 0
```
jsdosa, line 56: could you remove white space?

PR #4258, opened 2020-04-24
arch/arm/src/armv7-m, armv8-m: Fix build error with unused variable

sangam-swami:
Unused tcb variable resulting into build failure with CONFIG_ARMV7M_MPU.
Fix build by removing tcb declaration.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_svcall.c
+++ b/os/arch/arm/src/armv7-m/up_svcall.c
@@ -256,11 +256,11 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		DEBUGASSERT(regs[REG_R1] != 0);
 		current_regs = (uint32_t *)regs[REG_R1];
 
-#if defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_TASK_MONITOR)
+#if (defined(CONFIG_ARMV7M_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)) || defined(CONFIG_TASK_MONITOR)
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/armv7-m/up_svcall.c
```diff
@@ -256,14 +256,12 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 		DEBUGASSERT(regs[REG_R1] != 0);
 		current_regs = (uint32_t *)regs[REG_R1];
 
-#if defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_TASK_MONITOR)
-		struct tcb_s *tcb = sched_self();
-#endif
 		/* Restore the MPU registers in case we are switching to an application task */
 #ifdef CONFIG_ARMV7M_MPU
-		up_set_mpu_app_configuration(tcb);
```
sunghan-chang, line 264: When both are enabled, it calls `sched_self` twice. I don't think that is better. How about adding the conditional for this?
```
#if (defined(CONFIG_ARMV7M_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)) || defined(CONFIG_TASK_MONITOR)	
		struct tcb_s *tcb = sched_self();	
#endif

#if defined(CONFIG_ARM7M_MPU) && defined(CONFIG_APP_BINARY_SEPARATION)
     up_set_mpu_app_configuration(tcb);
#endif 
```

sangam-swami: yes we can add extra define for this. Change is done

PR #4270, opened 2020-05-15
os/mm: Modify BASE_HEAP access method

kishore-sn:
Modify BASE_HEAP access method to make it same as Mediatek.
When this approach is used, the BASE_HEAP will have valid address
irrespective of running from flash or ram.


sunghan-chang:
@kishore-sn Could you remove definitions duplicated?

kishore-sn:
@sunghan-chang I removed the duplicate definitions.

os/include/tinyara/mm/mm.h
```diff
@@ -424,29 +424,36 @@ extern "C" {
...
+#include <tinyara/sched.h>
 #define USR_HEAP_TCB ((struct mm_heap_s *)((struct tcb_s*)sched_self())->uheap)
-#define USR_HEAP_CFG ((struct mm_heap_s *)(*(uint32_t *)((uint32_t)USERSPACE + sizeof(struct userspace_s))))
-#define BASE_HEAP (USR_HEAP_TCB == NULL ? USR_HEAP_CFG : USR_HEAP_TCB)
+#define USR_HEAP_CFG ((struct mm_heap_s *)(*(uint32_t *)(((struct userspace_s *)USERSPACE)->us_textstart)))
+#define USR_HEAP (USR_HEAP_TCB == NULL ? USR_HEAP_CFG : USR_HEAP_TCB)
+#else
+#define USR_HEAP ((struct mm_heap_s *)(*(uint32_t *)(((struct userspace_s *)USERSPACE)->us_textstart)))
+#endif
+
```
sangam-swami, line 441: Can we change like below to remove duplicate code
#define USR_HEAP_CFG ((struct mm_heap_s *)(*(uint32_t *)(((struct userspace_s *)USERSPACE)->us_textstart)))
#include <tinyara/sched.h>
#define USR_HEAP_TCB ((struct mm_heap_s *)((struct tcb_s*)sched_self())->uheap)	
#define USR_HEAP (USR_HEAP_TCB == NULL ? USR_HEAP_CFG : USR_HEAP_TCB)
#else
#define USR_HEAP USR_HEAP_CFG 
#endif

os/drivers/heapinfo/heapinfo_drv.c
```diff
@@ -88,14 +88,14 @@ static int heapinfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		switch (option->heap_type) {
 #ifndef CONFIG_BUILD_PROTECTED
 		case HEAPINFO_HEAP_TYPE_KERNEL:
-			heap = &BASE_HEAP[regionx_heap_idx[0]];
+			heap = &USR_HEAP[regionx_heap_idx[0]];
```
sunghan-chang, line 91: Because of naming, this looks strange. This means getting the kernel heap from user heap data.

os/include/tinyara/mm/mm.h
kishore-sn: I am going to raise a new PR for removing the USERSPACE completely when app separation is enabled. In that case, the above definitions will get much simpler. So, for the time being, let us keep the current code.

os/drivers/heapinfo/heapinfo_drv.c
kishore-sn: I understand that the name seems odd, but in flat build, all kernel memory operations get redirected to user mm and internally, USR_HEAP only is getting used. However, I could add a comment here to indicate the reason for using USR_HEAP for TYPE_KERNEL. Is that ok?

kishore-sn:
@jeongchanKim 
I am planning to raise a new PR to remove USERSPACE completely when app separation is enabled. In that case, the issues related to REGIONx will also be resolved. So, I will take care of your suggestion in the next PR.

os/drivers/heapinfo/heapinfo_drv.c
sunghan-chang: ok.

kishore-sn: Done

PR #4271, opened 2020-05-18
fix build warnings

sunghan-chang:
1. include/debug.h: fix typo in media debug definition 
2. tools/nxfuse: fix build warnings

### DIFF
--- a/os/binfmt/libelf/libelf_symbols.c
+++ b/os/binfmt/libelf/libelf_symbols.c
@@ -60,6 +60,9 @@
 #include <string.h>
 #include <errno.h>
 #include <debug.h>
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+#include <hashmap.h>
+#endif
 
 #include <tinyara/binfmt/elf.h>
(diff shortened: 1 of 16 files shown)


os/binfmt/libelf/libelf_symbols.c
```diff
@@ -221,7 +224,7 @@ void elf_readsymtab(FAR struct elf_loadinfo_s *loadinfo)
 		return;
 	}
 
-	if (elf_read(loadinfo, loadinfo->symtab, symtab->sh_size, symtab->sh_offset) < 0) {
+	if (elf_read(loadinfo, (FAR uint8_t *)loadinfo->symtab, symtab->sh_size, symtab->sh_offset) < 0) {
```
sangam-swami, line 227: do we need to add similar typecast at line 111 -> (FAR uint8_t *)loadinfo->strtab

sunghan-chang: Yeah it requires. I was modifying. It will be updated more. libfmt_loadbinary.c and binfmt_execmodule.c left.

PR #4277, opened 2020-05-20
docs/LoadableAppsConfigurationUserGuide : Add description about API u…

jeongchanKim:
…sage limitation

The usage of some APIs are distinguished from kernel and user.
So User should consider which API to use from each side.


docs/LoadableAppsConfigurationUserGuide.md
```diff
@@ -35,6 +35,14 @@ The document [Protected Build Guide](ProtectedBuildGuide.md) provides the instru
 2. Flash and ram memory needs to be available to accomodate the applications.
 3. The total ram requirement of an application needs to be fixed at build time.
 4. Currently TizenRT supports only elf binary format.
+5. The usage of some APIs are distinguished from kernel and user.
```
davidfather, line 38: Let's revise this sentence a bit.

PR #4278, opened 2020-05-20
binfmt : Add debug for saving each sections addr of bi…

jeongchanKim:
…nary

For finding/matching the symbols of user binary, saving each sections address is required.
So saving when loading, and releasing when unloading the address.

### DIFF
--- a/os/binfmt/binfmt_exit.c
+++ b/os/binfmt/binfmt_exit.c
@@ -61,6 +61,9 @@
 #include <debug.h>
 #ifdef CONFIG_APP_BINARY_SEPARATION
 #include <queue.h>
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+#include "libelf/libelf.h"
+#endif
 #endif
 
 #include <tinyara/kmalloc.h>
@@ -113,6 +116,10 @@ int binfmt_exit(FAR struct binary_s *bin)
 	if (ret < 0) {
 		berr("ERROR: unload_module() failed: %d\n", ret);
 	}
+
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+	elf_delete_bin_section_addr(bin);
+#endif
 
 #ifdef CONFIG_APP_BINARY_SEPARATION
 	uheap_start = (uint32_t)bin->uheap;
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -35,6 +35,10 @@
 #include <tinyara/binary_manager.h>
 #include <tinyara/mpu.h>
 
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+#include "libelf/libelf.h"
+#endif
+
 #include "binfmt.h"
 #include "binary_manager/binary_manager.h"
 
@@ -188,28 +192,29 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	if (bin->islibrary) {
 		g_umm_app_id = bin->alloc[ALLOC_DATA];
-	} else {
-		/* If we support common binary, then we need to place a pointer to the app's heap object
-		 * into the heap table which is present at the start of the common library data section
-		 */
-		uint32_t *heap_table = (uint32_t *)g_lib_binp->alloc[ALLOC_DATA] + 4;
-		heap_table[binary_idx] = bin->heapstart;
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+		elf_save_bin_section_addr(bin);
 #endif
-
-		/* Start the module */
-		pid = exec_module(bin);
-		if (pid < 0) {
-			errcode = pid;
-			berr("ERROR: Failed to execute program '%s': %d\n", filename, errcode);
-			goto errout_with_unload;
-		}
-
-		return pid;
-#ifdef CONFIG_SUPPORT_COMMON_BINARY
+		return OK;
 	}
-
-	return OK;
+	/* If we support common binary, then we need to place a pointer to the app's heap object
+	 * into the heap table which is present at the start of the common library data section
+	 */
+	uint32_t *heap_table = (uint32_t *)g_lib_binp->alloc[ALLOC_DATA] + 4;
+	heap_table[binary_idx] = bin->heapstart;
 #endif
+
+	/* Start the module */
+	pid = exec_module(bin);
+	if (pid < 0) {
+		errcode = pid;
+		berr("ERROR: Failed to execute program '%s': %d\n", filename, errcode);
+		goto errout_with_unload;
+	}
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+	elf_save_bin_section_addr(bin);
+#endif
+	return pid;
 
 errout_with_unload:
 	(void)unload_module(bin);
--- a/os/binfmt/libelf/libelf_sections.c
+++ b/os/binfmt/libelf/libelf_sections.c
@@ -65,6 +65,12 @@
 #include <tinyara/kmalloc.h>
 #include <tinyara/binfmt/elf.h>
 
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+#include <queue.h>
+#include <string.h>
+#include <tinyara/binfmt/binfmt.h>
+#endif
+
 #include "libelf.h"
 
 /****************************************************************************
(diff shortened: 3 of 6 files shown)


os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -192,6 +198,23 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
...
+		bin_info->rodata_addr = loadinfo->roalloc;
+		bin_info->data_addr = loadinfo->dataalloc;
+		bin_info->bss_addr = loadinfo->binp->bssstart;
+
+		sq_addlast((sq_entry_t *)bin_info, &g_bin_addr_list);
+	} else {
+		berr("ERROR : Failed to allocate for bin symbol.\n");
+	}
+#endif
+
```
kishore-sn, line 217: How about moving this code to  binfmt_loadbinary.c ?
If there is an error in loading binary after this point, then binary will not be loaded, but the bin_info will remain in the list. So, better to create the bin_info only upon successful loading of binary.

os/binfmt/binfmt_loadbinary.c
```diff
@@ -160,6 +164,24 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
...
+			bin_info->bss_addr = (uint32_t)bin->bssstart;
+
+			binfo("[%s] text_addr : %x, rodata_addr : %x, data_addr : %x, bss_addr : %x\n", bin_info->bin_name, bin_info->text_addr, bin_info->rodata_addr, bin_info->data_addr, bin_info->bss_addr);
+
+			sq_addlast((sq_entry_t *)bin_info, &g_bin_addr_list);
+		} else {
+			berr("ERROR : Failed to allocate for bin symbol.\n");
+		}
+#endif
+
```
kishore-sn, line 184: There is still a chance of error in line 203 and 226.
How about moving this after line 235?
We can move the return statements also after line 235.

os/binfmt/binfmt_loadbinary.c
```diff
@@ -204,12 +208,30 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 			goto errout_with_unload;
 		}
 
-		return pid;
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	}
 
-	return OK;
+	pid = OK;
```
kishore-sn, line 214: This will overwrite the pid value from line 204.
Better to initialize pid = 0 at line 191 or at the time of declaration.

os/binfmt/binfmt_exit.c
```diff
@@ -107,6 +111,19 @@ int binfmt_exit(FAR struct binary_s *bin)
...
+	info = (bin_addr_info_t *)sq_peek(&g_bin_addr_list);
+	while (info) {
+		if (strncmp(info->bin_name, bin->bin_name, BIN_NAME_MAX) == 0) {
+			sq_rem((sq_entry_t *)info, &g_bin_addr_list);
+			kmm_free(info);
+			break;
+		}
+		info = (FAR void *)sq_next((FAR sq_entry_t *)info);
+	}
+#endif
```
Taejun-Kwon, line 125: g_bin_addr_list is global, so set null its head and tail explicitly, i mean sq_init.

os/binfmt/binfmt_loadbinary.c
```diff
@@ -204,12 +208,28 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
...
+		bin_info->rodata_addr = (uint32_t)bin->alloc[ALLOC_DATA];
+		bin_info->data_addr = (uint32_t)bin->datastart;
+		bin_info->bss_addr = (uint32_t)bin->bssstart;
 
-	return OK;
+		binfo("[%s] text_addr : %x, rodata_addr : %x, data_addr : %x, bss_addr : %x\n", bin_info->bin_name, bin_info->text_addr, bin_info->rodata_addr, bin_info->data_addr, bin_info->bss_addr);
+
+		sq_addlast((sq_entry_t *)bin_info, &g_bin_addr_list);
+	} else {
+		berr("ERROR : Failed to allocate for bin symbol.\n");
```
Taejun-Kwon, line 229: it seems that something more needed, return pid is ok?
you didn't set errcode as -ENOMEM, also bin is still allocated.

os/binfmt/binfmt_loadbinary.c
```diff
@@ -204,12 +208,28 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
...
 
-		return pid;
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	}
+#endif
+
+#ifdef CONFIG_DEBUG_BIN_SYMBOL
+	bin_addr_info_t *bin_info;
+	bin_info = (bin_addr_info_t *)kmm_malloc(sizeof(bin_addr_info_t));
+	if (bin_info != NULL) {
```
Taejun-Kwon, line 218: If anything stopped because of return value, then you'd better follow below style to improve redability.
That is always easy to understand
```
if (a != b) {
    return or  goto
}
do_something;
```
Of course depth is just one here, you can ignore this comment.

jeongchanKim: This is just debug option, not for module operation. So even if not allocated properly, I don't want to stop the operation even if this debug functionality fails.

os/binfmt/binfmt_exit.c
jeongchanKim: This is not the whole list termination, only for exiting one binary(app).

os/binfmt/binfmt_loadbinary.c
jeongchanKim: I replied the comment in the above. Thanks.

os/binfmt/binfmt_exit.c
```diff
@@ -107,6 +111,19 @@ int binfmt_exit(FAR struct binary_s *bin)
 #endif
 	DEBUGASSERT(bin != NULL);
 
+#ifdef CONFIG_SAVE_BIN_SYMBOL
```
jeongarmy, line 114: I think it should be called when unload_module succeed. so you can move this to line 134.

jeongchanKim: Done. Thanks.

os/binfmt/binfmt_loadbinary.c
Taejun-Kwon: but it just return pid.

jeongchanKim: If no mem, there is no allocated resources from debug feature, and it just print the error msg.

os/binfmt/binfmt_exit.c
```diff
@@ -114,6 +118,19 @@ int binfmt_exit(FAR struct binary_s *bin)
 		berr("ERROR: unload_module() failed: %d\n", ret);
 	}
 
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
```
sunghan-chang, line 121: 1. same as save function, let's make this as a new function.
2. How about making a new file to get this and save?
  Because they are not matched with file name,

os/binfmt/binfmt_loadbinary.c
```diff
@@ -203,11 +228,16 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 			berr("ERROR: Failed to execute program '%s': %d\n", filename, errcode);
 			goto errout_with_unload;
 		}
-
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+		save_bin_section_addr(bin);
+#endif
 		return pid;
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
```
sunghan-chang, line 235: This looks unnecessary. We could move two lines under if statement.

os/binfmt/binfmt_exit.c
jeongchanKim: Modified. Thanks.

os/binfmt/libelf/libelf_sections.c
```diff
@@ -65,6 +65,12 @@
 #include <tinyara/kmalloc.h>
 #include <tinyara/binfmt/elf.h>
 
+#ifdef CONFIG_SAVE_BIN_SECTION_ADDR
+#include <queue.h>
+#include <string.h>
+#include <tinyara/mm/mm.h>
```
sunghan-chang, line 71: 1. Could you let me know why this is necessary?
2. binfmt.h is necessary

PR #4282, opened 2020-05-25
fs/mtd : Add common mtd driver which uses jedec interface

Taejun-Kwon:
1. Add mtd driver for jedec
2. Add routines for new mtd driver


os/fs/driver/mtd/mtd_jedec.c
```diff
@@ -0,0 +1,885 @@
+/****************************************************************************
+ *
+ * Copyright 2018 Samsung Electronics All Rights Reserved.
```
sunghan-chang, line 3: 2020

PR #4283, opened 2020-05-26
Disable USERSPACE binary

kishore-sn:
This PR contains following changes:
- When COMMON_BINARY is supported, then remove user binary
- Reduce size of user binary 
- User standard method for heap allocation


os/binfmt/binfmt_loadbinary.c
```diff
@@ -192,6 +194,7 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	if (bin->islibrary) {
 		g_umm_app_id = bin->alloc[ALLOC_DATA];
+		USERSPACE = bin->alloc[ALLOC_TEXT] + 1;
```
jeongchanKim, line 197: Please add type casting.

build/configs/imxrt1020-evk/loadable_elf_apps/defconfig
```diff
@@ -395,9 +395,9 @@ CONFIG_ARCH_USE_FLASH=y
 #
 CONFIG_FLASH_PARTITION=y
 CONFIG_FLASH_MINOR=0
-CONFIG_FLASH_PART_SIZE="2048,512,3072,128,128,"
-CONFIG_FLASH_PART_TYPE="kernel,none,smartfs,ftl,config,"
-CONFIG_FLASH_PART_NAME="kernel,app,userfs,ftl,config,"
```
jeongchanKim, line 400: We should consider there is no common binary. In this case, we should set the user binary partition.
How can we set this?

build/configs/imxrt1020-evk/loadable_elf_apps/defconfig
```diff
@@ -395,9 +395,9 @@ CONFIG_ARCH_USE_FLASH=y
...
+CONFIG_FLASH_PART_SIZE="2048,3072,128,128,"
+CONFIG_FLASH_PART_TYPE="kernel,smartfs,ftl,config,"
```
Taejun-Kwon, line 399: You'd better divide partition of smartfs, one is for userspace binary, another one is userfs.
Because of wearleveling, middle of user binary can be relocate anytime....
Binary affect to gc condition, performance also decreased significantly.

os/arch/arm/src/stm32l4/stm32l4_start.c
```diff
@@ -161,60 +161,60 @@ void __start(void) __attribute__ ((no_instrument_function));
...
+	putreg32(regval, NVIC_FPCCR);
 
-  /* Enable full access to CP10 and CP11 */
+	/* Enable full access to CP10 and CP11 */
 
-  regval = getreg32(NVIC_CPACR);
-  regval |= ((3 << (2*10)) | (3 << (2*11)));
-  putreg32(regval, NVIC_CPACR);
+	regval = getreg32(NVIC_CPACR);
+	regval |= ((3 << (2*10)) | (3 << (2*11)));
```
Taejun-Kwon, line 186: minor one, 2 * 10 & 2 * 11 (line 216 too)

build/configs/imxrt1020-evk/loadable_elf_apps/defconfig
kishore-sn: If there is no common binary, then we need to add the user binary partition manually. I dont think we can automate it.

build/configs/imxrt1020-evk/loadable_elf_apps/defconfig
kishore-sn: I understand. But, then we need to mount 2 copies of smartfs. Also, there will be changes in the way we generate and flash the smartfs image during the build. I am not very sure about how to implement it. Since, its a big change and not directly related to removal of user binary, we will take this up in a different PR.

os/arch/arm/src/stm32l4/stm32l4_start.c
kishore-sn: Done.

os/binfmt/binfmt_loadbinary.c
kishore-sn: Done

build/configs/imxrt1020-evk/loadable_elf_apps/defconfig
Taejun-Kwon: Ok nice man, but it is really necessary.
I don't mean multi-root dir,(CONFIG_SMARTFS_MULTI_ROOT_DIRS), Make 2 partition, mount each of partition. 
That is, they use different MTD partition.

os/include/tinyara/userspace.h
```diff
@@ -80,8 +80,13 @@ extern uint32_t __uflash_segment_start__[];
 /* This macro is used to access the struct userpace_s header that can be
  * found at the beginning of the user-space blob.
  */
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+struct userspace_s;
```
jeongchanKim, line 84: Is this needed?

os/binfmt/binfmt_loadbinary.c
```diff
@@ -200,7 +203,7 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	/* If we support common binary, then we need to place a pointer to the app's heap object
 	 * into the heap table which is present at the start of the common library data section
 	 */
-	uint32_t *heap_table = (uint32_t *)g_lib_binp->alloc[ALLOC_DATA] + 4;
+	uint32_t *heap_table = (uint32_t *)(g_lib_binp->alloc[ALLOC_DATA] + 1);
```
sunghan-chang, line 206: Could you explain why you modify this from 4 to 1?
And `usrheapptr` is located at start of text section. Don't you use that variable? I can't find the heap table at data section.
```
  .data :
  {
    _sdata = ABSOLUTE(.);
    *(.data .data.*)
    *(.gnu.linkonce.d.*)
    CONSTRUCTORS
    _edata = ABSOLUTE(.);
  } > usram AT > uflash
```

os/binfmt/binfmt_loadbinary.c
```diff
@@ -47,6 +47,8 @@
 #include <tinyara/mm/mm.h>
 
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
+#include <tinyara/userspace.h>
+struct userspace_s *USERSPACE;
```
sunghan-chang, line 51: It looks strange to define this here.
I think we should define this in upspace module and should take it as extern.

os/arch/arm/src/stm32l4/stm32l4_start.c
```diff
@@ -102,7 +102,7 @@ static inline void stm32l4_fpuconfig(void);
 #endif
 #ifdef CONFIG_STACK_COLORATION
 static void go_nx_start(void *pv, unsigned int nbytes)
-  __attribute__ ((naked, no_instrument_function, noreturn));
+	__attribute__ ((naked, no_instrument_function, noreturn));
```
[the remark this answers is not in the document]
sunghan-chang: Yeah, it is necessary to move another commit.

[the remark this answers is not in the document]
kishore-sn: Done

os/include/tinyara/userspace.h
sunghan-chang: How about switching them?
```
struct userspace_s {
...
};

extern struct userspace_s *USERSPACE;
#else
#define USERSPACE ((FAR struct userspace_s *)__uflash_segment_start__)
#endif 
```

sunghan-chang: And originally `#define USERSPACE ((FAR struct userspace_s *)__uflash_segment_start__)` is defined at previous of structure definition without prototype. It means that new definition should be ok. But anyway I think switching is better.

os/binfmt/binfmt_loadbinary.c
sunghan-chang: Even up_userspace funtion is not used, it will be compiled because you don't block it at Makefile. Then, we could use that file to define userspace I think.

kishore-sn: Ok. Done.

os/include/tinyara/userspace.h
kishore-sn: Ok. done.

os/arch/arm/src/imxrt/imxrt_userspace.c
```diff
@@ -70,6 +70,10 @@
  * Public Functions
  ****************************************************************************/
 
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
```
sunghan-chang, line 73: This file looks unnecessary. Let's use up_userspace.c instead of this.

os/arch/arm/src/stm32/stm32_userspace.c
```diff
@@ -74,6 +74,14 @@
  * Private Data
  ****************************************************************************/
 
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
```
sunghan-chang, line 81: Except calling mpu init from this file, others are the same as up_userspace.c. How about moving mpu init function at somewhere and removing this?

sunghan-chang: Managing several almost the same file is not good. That's why I suggest.

os/arch/arm/src/stm32l4/stm32l4_userspace.c
```diff
@@ -49,6 +49,14 @@
 
 #ifdef CONFIG_BUILD_PROTECTED
 
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
```
sunghan-chang, line 56: same as previous

os/arch/arm/src/tiva/tiva_userspace.c
```diff
@@ -74,6 +74,14 @@
  * Private Data
  ****************************************************************************/
 
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
```
sunghan-chang, line 81: Except boards we ported, other boards which come from Nuttx have the same format. We need to think which is better.

PR #4285, opened 2020-05-27
Remove all Bad Sector Management related code from Smartfs

amogh-samsung:
- Current BSM logic is incomplete and flawed,
  causes redundancy in smart_writesector.
- BSM logic not put to use as of now, might
  be improved upon and re-implemented later.


amogh-samsung:
Refer PR#4286 in Samsung/TizenRT -> fs_journal

PR #4291, opened 2020-06-01
binfmt/libelf : Add #ifdef for OPTIMIZE_RELOAD_TIME in libelf_section.c

jeongchanKim:
If CONFIG_OPTIMIZE_RELOAD_TIME is enabled, elf will be loaded with one block. So there is no each section's place.

### DIFF
--- a/os/binfmt/libelf/libelf.h
+++ b/os/binfmt/libelf/libelf.h
@@ -465,9 +465,11 @@ struct bin_addr_info_s {
 	struct bin_addr_info_s *flink;
 	char bin_name[BIN_NAME_MAX];
 	uint32_t text_addr;
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
 	uint32_t rodata_addr;
 	uint32_t data_addr;
 	uint32_t bss_addr;
+#endif
 };
 typedef struct bin_addr_info_s bin_addr_info_t;
 
--- a/os/binfmt/libelf/libelf_sections.c
+++ b/os/binfmt/libelf/libelf_sections.c
@@ -194,12 +194,17 @@ void elf_save_bin_section_addr(struct binary_s *bin)
 	if (bin_info != NULL) {
 		strncpy(bin_info->bin_name, bin->bin_name, BIN_NAME_MAX);
 		bin_info->text_addr = (uint32_t)bin->alloc[ALLOC_TEXT];
+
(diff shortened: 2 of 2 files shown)


os/binfmt/libelf/libelf.h
```diff
@@ -464,10 +464,12 @@ int elf_cache_read(int filfd, uint16_t binary_header_size, FAR uint8_t *buffer,
 struct bin_addr_info_s {
 	struct bin_addr_info_s *flink;
 	char bin_name[BIN_NAME_MAX];
-	uint32_t text_addr;
-	uint32_t rodata_addr;
-	uint32_t data_addr;
-	uint32_t bss_addr;
+	uint32_t text_paddr;
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
```
jsdosa, line 468: How about changing the config name to CONFIG_DYNAMIC_RELOAD?

jsdosa: or CONFIG_DYNAMIC_RELOAD_APP ?

jeongchanKim: This is the config for optimization when reloading. There is already another config which related to dynamic loading.

os/binfmt/libelf/libelf_sections.c
```diff
@@ -194,12 +194,15 @@ void elf_save_bin_section_addr(struct binary_s *bin)
 	if (bin_info != NULL) {
 		strncpy(bin_info->bin_name, bin->bin_name, BIN_NAME_MAX);
 		bin_info->text_addr = (uint32_t)bin->alloc[ALLOC_TEXT];
+
+		binfo("[%s] text_addr : %x", bin_info->bin_name, bin_info->text_addr);
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
```
Taejun-Kwon, line 199: You'd better add new line between line 205 & 206, otherwise this will mixed with next log if CONFIG_OPTIMIZE_APP_RELOAD_TIME is disabled.
But If I were you, I print out line per item
text addr : 0x12312313
rodata_adr : 0x15120481412
...

jeongchanKim: Done. Thanks.

os/binfmt/libelf/libelf.h
sijupunnnoose: Config is already in use, so no need to change it.

PR #4297, opened 2020-06-02
os/fs/driver/mtd: Fix CRC check function name and logicalsector check

amogh-samsung:
- Function name smart_validate_erase_block() is
 misleading. Rename to smart_validate_sector.
- Add logicalsector number validity check
 before adding to sector sequence log.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -1681,7 +1681,15 @@ static int smart_set_wear_level(FAR struct smart_struct_s *dev, uint16_t block, 
 }
 #endif
 
-static bool smart_validate_erase_block(FAR struct smart_struct_s *dev)
+/****************************************************************************
+ * Name: smart_validate_sector
+ *
+ * Description: Check CRC for the sector and determine if sector
+ *		contents are OK in case CRC does not match.
+ *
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -1862,22 +1870,24 @@ static int smart_scan(FAR struct smart_struct_s *dev)
 		if (logicalsector > 0 && header.status != CONFIG_SMARTFS_ERASEDSTATE && !status_released && status_committed) {
 
 			/* Map the sector and update the free sector counts. */
-			if (header.seq >= sector_seq_log[logicalsector]) {
+			if (logicalsector < dev->totalsectors) {
```
Taejun-Kwon, line 1873: you can add this condition beside of above state (logicalsecotr > 0 && logicalsector < dev->totalsectors)
But, most of all these code will be removed soon, you raised this PR to um..just clean current version?

amogh-samsung: I'll make the change as you suggested.
And yes, these changes only clean the current code a little bit, we have discussed the rest and I won't try to modify that as of now.

amogh-samsung:
Travis check getting stuck.

PR #4299, opened 2020-06-03
Reduce app loading memory requierment

kishore-sn:
This PR contains following commits:
- Modify memalign to reduce memory requirement for alginment
- Modify loader to avoid multiple memalign request for each elf section

### DIFF
--- a/os/binfmt/binfmt_unloadmodule.c
+++ b/os/binfmt/binfmt_unloadmodule.c
@@ -153,7 +153,6 @@ static inline int exec_dtors(FAR struct binary_s *binp)
 int unload_module(FAR struct binary_s *binp)
 {
 	int ret;
-	int i;
 
 	if (binp) {
 		/* Perform any format-specific unload operations */
@@ -189,19 +188,11 @@ int unload_module(FAR struct binary_s *binp)
 
 		/* Free allocated address spaces */
 
-#if defined(CONFIG_OPTIMIZE_APP_RELOAD_TIME) || !defined(CONFIG_APP_BINARY_SEPARATION)
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
-		if (!binp->reload) {
-#endif
-			for (i = 0; i < ALLOC_MAX; i++) {
-				if (binp->alloc[i]) {
-					binfo("Freeing alloc[%d]: %p\n", i, binp->alloc[i]);
-					kumm_free((FAR void *)binp->alloc[i]);
-				}
-			}
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#ifndef CONFIG_APP_BINARY_SEPARATION
+		if (binp->alloc[i]) {
+			binfo("Freeing alloc[%d]: %p\n", i, binp->alloc[i]);
+			kumm_free((FAR void *)binp->alloc[i]);
 		}
-#endif
 #endif
 
 		/* Notice that the address environment is not destroyed.  This should
--- a/os/binfmt/libelf/libelf_addrenv.c
+++ b/os/binfmt/libelf/libelf_addrenv.c
@@ -64,6 +64,9 @@
 #ifdef CONFIG_APP_BINARY_SEPARATION
 #include <tinyara/mm/mm.h>
 #endif
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#include <tinyara/mpu.h>
+#endif
 #include "libelf.h"
 
 /****************************************************************************
@@ -77,6 +80,36 @@
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
+{
+	size_t sizes[MPU_NUM_REGIONS] = {loadinfo->textsize, loadinfo->rosize, loadinfo->binp->ramsize};
+	uintptr_t *allocs[MPU_NUM_REGIONS] = {&loadinfo->textalloc, &loadinfo->roalloc, &loadinfo->dataalloc};
+	uint32_t totalsize = sizes[0] + sizes[1] + sizes[2];
+	size_t tmpsz;
+	uintptr_t *tmpalloc;
+	uint8_t i;
+	uint8_t j;
+
+	for (i = 0; i < MPU_NUM_REGIONS; i++) {
+		for (j = 0; j < MPU_NUM_REGIONS - (i + 1); j++) {
+			if (sizes[j] < sizes[j + 1]) {
+				tmpsz = sizes[j];
+				sizes[j] = sizes[j + 1];
+				sizes[j + 1] = tmpsz;
+
+				tmpalloc = allocs[j];
+				allocs[j] = allocs[j + 1];
+				allocs[j + 1] = tmpalloc;
+			}
+		}
+	}
+
+	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kumm_memalign(sizes[0], totalsize);
+	*allocs[1] = *allocs[0] + sizes[0];
+	*allocs[2] = *allocs[1] + sizes[1];
+}
+#endif
 
 /****************************************************************************
  * Public Functions
@@ -153,24 +186,20 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
 	uint32_t datamemsize = loadinfo->datasize + loadinfo->binp->ramsize;
 	uint32_t rosize = loadinfo->rosize;
+
 	/* loadinfo->datasize contains the size of both data and bss sections.
 	 * But we need to backup only the data section in the RO region. So,
 	 * we need to extend the rosize by just the data section size only. Hence,
 	 * we are subtracting bsssize from loadinfo->datasize.
 	 */
 	loadinfo->rosize += loadinfo->datasize - loadinfo->binp->bsssize;
-#ifdef CONFIG_ARM_MPU
+
 	loadinfo->textsize = 1 << mpu_log2regionceil(0, loadinfo->textsize);
 	loadinfo->rosize = 1 << mpu_log2regionceil(0, loadinfo->rosize);
 	datamemsize = 1 << mpu_log2regionceil(0, datamemsize);
 	loadinfo->binp->ramsize = datamemsize;
-#endif
-#endif
 
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
-	loadinfo->textalloc = (uintptr_t)kumm_memalign(loadinfo->textsize, loadinfo->textsize);
-	loadinfo->roalloc = (uintptr_t)kumm_memalign(loadinfo->rosize, loadinfo->rosize);
-	loadinfo->dataalloc = (uintptr_t)kumm_memalign(datamemsize, datamemsize);
+	allocateregions(loadinfo);
 
 	loadinfo->binp->data_backup = loadinfo->roalloc + rosize;
 	loadinfo->binp->uheap_size = datamemsize - loadinfo->datasize - sizeof(struct mm_heap_s);
--- a/os/mm/mm_heap/mm_memalign.c
+++ b/os/mm/mm_heap/mm_memalign.c
@@ -18,7 +18,7 @@
 /****************************************************************************
  * mm/mm_heap/mm_memalign.c
  *
- *   Copyright (C) 2007, 2009, 2011, 2013-2014 Gregory Nutt. All rights reserved.
+ *   Copyright (C) 2007, 2009, 2013-2014  Gregory Nutt. All rights reserved.
  *   Author: Gregory Nutt <gnutt@nuttx.org>
  *
  * Redistribution and use in source and binary forms, with or without
@@ -56,12 +56,29 @@
 
 #include <tinyara/config.h>
 
-#include <assert.h>
+#include <debug.h>
 
 #include <tinyara/mm/mm.h>
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+#include  <tinyara/sched.h>
+#endif
+#include "mm_node.h"
+
 /****************************************************************************
  * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Type Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Data
  ****************************************************************************/
 
 /****************************************************************************
@@ -71,7 +88,6 @@
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
-
 /****************************************************************************
  * Name: mm_memalign
  *
@@ -90,11 +106,13 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size,
 FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
 #endif
 {
-	FAR struct mm_allocnode_s *node;
-	size_t rawchunk;
-	size_t alignedchunk;
+	FAR struct mm_freenode_s *node;
+	void *ret = NULL;
+	int ndx;
+	size_t newsize;
+	FAR struct mm_allocnode_s *alignchunk = NULL;
+	bool found_align = false;
 	size_t mask = (size_t)(alignment - 1);
-	size_t allocsize;
 
 	/* If this requested alinement's less than or equal to the natural alignment
 	 * of malloc, then just let malloc do the work.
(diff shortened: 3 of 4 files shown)


os/mm/mm_heap/mm_memalign.c
```diff
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- * Copyright 2016 Samsung Electronics All Rights Reserved.
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
```
jeongchanKim, line 3: It is the first year that this file is generated. So we don't need to modify this.

os/mm/mm_heap/mm_memalign.c
```diff
@@ -56,22 +56,42 @@
...
+#include  <tinyara/sched.h>
+#endif
+#include "mm_node.h"
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
+#ifndef NULL
+#define NULL ((void*)0)
```
jeongchanKim, line 73: Please add space between `void` and `*`.

os/mm/mm_heap/mm_memalign.c
```diff
@@ -108,137 +130,146 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
+			 * if the remaining bytes before the alignment point is either zero or bigger than freenode.
+			 */
+			if (node->size >= alignsize && (remainsize == 0 || remainsize >= SIZEOF_MM_FREENODE)) {
+				found_align = true;
+				break;
+			}
+		}
 
-		DEBUGASSERT(alignedchunk >= rawchunk + 8);
+		if (found_align) {
```
jeongchanKim, line 196: If we cannot find the align one, `found_align` remains false. In this case, we will return NULL for failure.
So I think, if we modify this line with false checking, we don't need to check `found_align` in line 201.
```
if (found_align == false) {
    return NULL;
}
```

os/mm/mm_heap/mm_memalign.c
```diff
@@ -108,137 +130,146 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
+		/* Now, traverse the list in reverse direction, towards bigger size nodes */
+		for ( ; node; node = node->blink) {
+			alignchunk = (FAR struct mm_allocnode_s *)(((size_t)node + SIZEOF_MM_ALLOCNODE + mask) & ~mask);
+			size_t alignsize = (size_t)alignchunk - (size_t)node + size;
+			size_t remainsize = (size_t)alignchunk - SIZEOF_MM_ALLOCNODE - (size_t)node;
+
+			/* We found a suitable node if node size is more than required size after alignment and
+			 * if the remaining bytes before the alignment point is either zero or bigger than freenode.
+			 */
+			if (node->size >= alignsize && (remainsize == 0 || remainsize >= SIZEOF_MM_FREENODE)) {
```
jeongchanKim, line 190: I think, if `node->size` is greater than `alignsize` and `remainsize < SIZEOF_MM_FREENODE`, we can just take this node, and set the address to the aligned one. (We can ignore the remain, because it cannot be used anymore.)
Can we use `remainsize >= SIZEOF_MM_FREENODE` this case only?

os/mm/mm_heap/mm_memalign.c
```diff
@@ -108,137 +130,146 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
-		next->preceding = newnode->size | (next->preceding & MM_ALLOC_BIT);
-
-		/* Convert the newnode chunk size back into malloc-compatible size by
-		 * subtracting the header size SIZEOF_MM_ALLOCNODE.
-		 */
+			/* Add the original, newly freed node to the free nodelist */
+			mm_addfreechunk(heap, (FAR struct mm_freenode_s *)node);
 
-		allocsize = newnode->size - SIZEOF_MM_ALLOCNODE;
+			/* Replace the original node with the newly realloaced,
```
amandeepc, line 235: **reallocated

os/mm/mm_heap/mm_memalign.c
kishore-sn: ok

os/mm/mm_heap/mm_memalign.c
kishore-sn: This is not required. Removed it.

os/mm/mm_heap/mm_memalign.c
kishore-sn: We need to convert 
```ALLOCNODE---empty bytes---Allocated bytes```
into
```FREENODE---ALLOCNODE---Allocated bytes```

We cannot ignore the empty bytes, because every allocated memory must have below structure. 
```ALLOCNODE---Allocated bytes"

If there is empty bytes between them, then it is not possible to perform other operations like free on the allocated node. This is a basic requirement of tizenrt mm. If this requirement was not present, then the memalign would have been much simpler.

File: os/mm/mm_heap/mm_memalign.c (hunk 5 above)
Comment (kishore-sn) [reply os/mm/mm_heap/mm_memalign.c:235]: ok

**Comment by sunghan-chang** [comment]
You could set this as draft to prevent merging. After finishing implentation, you could change normal again.

File: os/binfmt/libelf/libelf_addrenv.c [hunk 6]
```diff
@@ -77,6 +77,51 @@
...
+	uint32_t totalsize;
+	size_t textsize, datasize, rosize;
+
+	textsize = loadinfo->textsize;
+	rosize = loadinfo->rosize;
+	datasize = loadinfo->binp->ramsize;
+
+	totalsize = textsize + rosize + datasize;
+
+	if (textsize >= rosize && textsize >= datasize) {
```
amogh-samsung, line 92: Could be done in short?
something like this: -

	`uint8_t arr[3] = {textsize, rosize, datasize};
	uint8_t *ptr1 = loadinfo->textalloc, *ptr2 = loadinfo->roalloc, *ptr3 = loadinfo->dataalloc;
	
		
	if (arr[0] < arr[1]) {
		<swap arr[0] and arr[1]>
		<swap ptr1 and ptr2>
	}
	
	if (arr[1] < arr[2]) {
		<swap arr[1] and arr[2]>
		<swap ptr2 and ptr3>
		if (arr[0] and arr[1]) {
			<swap arr[0] and arr[1]>
			<swap ptr1 and ptr2>
		}
	}
	
	loadinfo->binp->ramstart = ptr1 = (uintptr_t)kumm_memalign(arr[0], totalsize);

	ptr2 = ptr1 + arr[0];
	ptr3 = ptr2 + arr[1];
	
	<free pointers and array space>`

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +77,51 @@
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
+{
+	uint32_t totalsize;
+	size_t textsize, datasize, rosize;
+
+	textsize = loadinfo->textsize;
```
jeongchanKim, line 86: If there is a explanation why we try this, then we don't need to declare the variable with each section name.
We can just declare size_t sections[3], sort it by decreasing order, and allocate from start to end, we can reduce the code.
How about think of my suggestion?

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -153,24 +198,20 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
...
 	 * we need to extend the rosize by just the data section size only. Hence,
 	 * we are subtracting bsssize from loadinfo->datasize.
 	 */
 	loadinfo->rosize += loadinfo->datasize - loadinfo->binp->bsssize;
-#ifdef CONFIG_ARM_MPU
+
 	loadinfo->textsize = 1 << mpu_log2regionceil(0, loadinfo->textsize);
 	loadinfo->rosize = 1 << mpu_log2regionceil(0, loadinfo->rosize);
 	datamemsize = 1 << mpu_log2regionceil(0, datamemsize);
 	loadinfo->binp->ramsize = datamemsize;
```
jeongchanKim, line 212: I think we'd better change the variable `ramsize` name.. Because all other's are all for RAM.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: If we use size_t sections[3], then after sorting them in order, then we will not know which array element is for which type of section. So, it will be difficult to perform the allocation assignments.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Thanks @amogh-samsung . Done. 
@jeongchanKim I modified as you suggested to use array.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +77,51 @@
...
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
+{
+	size_t sizes[3] = {loadinfo->textsize, loadinfo->rosize, loadinfo->binp->ramsize};
+	uintptr_t *allocs[3] = {&loadinfo->textalloc, &loadinfo->roalloc, &loadinfo->dataalloc};
+	uint32_t totalsize = sizes[0] + sizes[1] + sizes[2];
+	size_t tmpsz;
+	uintptr_t *tmpalloc;
+
+
+	if (sizes[0] < sizes[1]) {
```
jeongchanKim, line 90: How about using bubble sort? We only have 3 members, so using bubble sort looks fine and simple.
```
for (i = 0; i < N; i++) {
    for (j = 0; j < N-(i+1); j++) {
        if (sizes[j] > sizes[j+1]) {
            temp = sizes[j+1];
            sizes[j+1] = sizes[j];
            sizes[j] = temp;

            temp = allocs[j+1];
            allocs[j+1] = allocs[j];
            allocs[j] = temp;
        }
    }
}
```

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Initially, i used ramstart and ramsize in case when RELOAD_OPTIMIZE is not defined. In that case, there is a single ram region for entire app. But, in this case, I understand that it does not provide the proper meaning. But, I still used this ramsize variable, because i did not want to introduce new variables in the structure. In future, if we decide to remove the RELOAD_OPTIMIZE config, then i will fix all the variable names properly.

os/binfmt/libelf/libelf_addrenv.c
amogh-samsung: Hi,
The current logic for comparison is the same as bubble sort but without the extra loop variables.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +77,51 @@
...
+			tmpalloc = allocs[0];
+			allocs[0] = allocs[1];
+			allocs[1] = tmpalloc;
+		}
+	}
+
+	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kumm_memalign(sizes[0], totalsize);
+	*allocs[1] = *allocs[0] + sizes[0];
+	*allocs[2] = *allocs[1] + sizes[1];
+}
```
amogh-samsung, line 123: Could we free the pointer array `allocs` and `tmpalloc`

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Done.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: We dont have to free them since they are static arrays and not allocated. And the values stored in alloc array are used by the loader in future.

jeongchanKim: `allocs` is local array. How can loader use this...?

kishore-sn: allocs is a pointer to pointer.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +77,35 @@
...
  ****************************************************************************/
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
+{
+	size_t sizes[3] = {loadinfo->textsize, loadinfo->rosize, loadinfo->binp->ramsize};
+	uintptr_t *allocs[3] = {&loadinfo->textalloc, &loadinfo->roalloc, &loadinfo->dataalloc};
+	uint32_t totalsize = sizes[0] + sizes[1] + sizes[2];
+	size_t tmpsz;
+	uintptr_t *tmpalloc;
+	uint8_t i, j, numsec = 3;
```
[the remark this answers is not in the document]
kishore-sn: Done

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +77,37 @@
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
+{
+	size_t sizes[3] = {loadinfo->textsize, loadinfo->rosize, loadinfo->binp->ramsize};
```
sijupunnnoose, line 83: Can you use a macro instead of using "3" everywhere?

kishore-sn: Done.

sijupunnnoose: I still see the "3" in code, change size_t sizes[3], uintptr_t *allocs[3] also

kishore-sn: Fixed.

os/mm/mm_heap/mm_memalign.c
```diff
@@ -108,137 +126,146 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
 
-	if (allocsize > size) {
-		/* Shrink the chunk by that much -- remember, mm_shrinkchunk wants
-		 * internal chunk sizes that include SIZEOF_MM_ALLOCNODE, and not the
-		 * malloc-compatible sizes that we have.
-		 */
-		mm_shrinkchunk(heap, node, size + SIZEOF_MM_ALLOCNODE);
-	}
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-	heapinfo_update_node(node, caller_retaddr);
```
jeongchanKim, line 237: I think we should update the heapinfo information also. Please check this.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +80,36 @@
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
```
sunghan-chang, line 84: How about adding `_` between two words? `allocate_regions`?
For better readability, could you add "what region" in function name?

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +80,36 @@
...
+{
+	size_t sizes[MPU_NUM_REGIONS] = {loadinfo->textsize, loadinfo->rosize, loadinfo->binp->ramsize};
+	uintptr_t *allocs[MPU_NUM_REGIONS] = {&loadinfo->textalloc, &loadinfo->roalloc, &loadinfo->dataalloc};
```
sunghan-chang, line 87: What's the difference between size and allocs?
It is difficult to know from the names.

os/mm/mm_heap/mm_memalign.c
```diff
@@ -108,137 +126,146 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
+	 * to the SYSLOG.
+	 */
 
-	heapinfo_add_size(heap, node->pid, node->size);
-	heapinfo_update_total_size(heap, node->size, node->pid);
+#ifdef CONFIG_DEBUG_MM
+	if (!ret) {
+		mdbg("Allocation failed, size %u\n", size);
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+		heapinfo_parse(heap, HEAPINFO_DETAIL_ALL, HEAPINFO_PID_ALL);
```
sunghan-chang, line 263: Do you intend to print this always when CONFIG_DEBUG_MM_HEAPINFO is enabled?

os/binfmt/binfmt_unloadmodule.c
```diff
@@ -189,19 +188,11 @@ int unload_module(FAR struct binary_s *binp)
...
-#endif
-			for (i = 0; i < ALLOC_MAX; i++) {
-				if (binp->alloc[i]) {
-					binfo("Freeing alloc[%d]: %p\n", i, binp->alloc[i]);
-					kumm_free((FAR void *)binp->alloc[i]);
-				}
-			}
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+#ifndef CONFIG_APP_BINARY_SEPARATION
+		if (binp->alloc[i]) {
```
sunghan-chang, line 192: There is no i definition and usages in this function.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -77,6 +80,36 @@
...
+				tmpalloc = allocs[j];
+				allocs[j] = allocs[j + 1];
+				allocs[j + 1] = tmpalloc;
+			}
+		}
+	}
+
+	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kumm_memalign(sizes[0], totalsize);
+	*allocs[1] = *allocs[0] + sizes[0];
+	*allocs[2] = *allocs[1] + sizes[1];
```
sunghan-chang, line 110: This is a local variable. Why do you calculate and save this?

sunghan-chang: Oh. it saves it in loadinfo structure as a member of alloc. Ok no problem

PR #4301, opened 2020-06-08
docs/ProtectedBuildGuide: add the Available APIs tab

sunghan-chang:
There are APIs which can be used on Protected build.
This commit adds the tab in the documentation of protected build guide
to say the API lists.

### DIFF
--- a/docs/ProtectedBuildGuide.md
+++ b/docs/ProtectedBuildGuide.md
@@ -3,7 +3,7 @@
 ## Contents
 - [Overview of Flat Build](#overview-of-flat-build)
 - [Overview of Protected Build](#overview-of-protected-build)
-- [Steps for suporting protected build in a new chipset](#steps-for-supporting-protected-build-in-a-new-chipset)
+- [Steps for supporting protected build in a new chipset](#steps-for-supporting-protected-build-in-a-new-chipset)
 	- [Make.defs](#makedefs)
 	- [Linker scripts](#linker-scripts)
 	- [Userspace Initialization](#userspace-initialization)
@@ -92,7 +92,7 @@ The following sections give detailed explanation regarding some of the steps / c
 
 ```
 
-- It is mandatory to define __uflash_segment_start__ linker variable. This will be used in the code to obtain the address of the uerspace object.
+- It is mandatory to define __uflash_segment_start__ linker variable. This will be used in the code to obtain the address of the userspace object.
 - The user can also enable the CONFIG_AUTOGEN_MEMORY_LDSCRIPT config, which will auto-generate the memory.ld file based on the Flash partition information in CONFIG_FLASH_PART_XXXX configs.
 
 #### user-space.ld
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@sangam-swami @kishore-sn @thapav @jeongarmy @jeongchanKim Could you review this? Is this enough to say the API lists for protected build?

docs/ProtectedBuildGuide.md
```diff
@@ -9,6 +9,7 @@
 	- [Userspace Initialization](#userspace-initialization)
 	- [MPU Initialization](#mpu-initialization)
 	- [Allocation of kernel and user heaps](#allocation-of-kernel-and-user-heaps)
+- [Available APIs](#available-apis)
```
sangam-swami, line 12: can we maintain same tab

sunghan-chang: with which one?

sunghan-chang: This is not a part of `Steps for suporting protected build in a new chipset`.

sangam-swami: oh It is added in contents, its fine I thought its sub section

docs/ProtectedBuildGuide.md
```diff
@@ -139,3 +140,14 @@ The following sections give detailed explanation regarding some of the steps / c
 - The user heap starts at the end of the user bss section and spans the entire remaining space in the user ram region.
 - However, if the chipset has a memory configuration that is different from the default one or if the vendor would like to initialize the kernel and user heaps at different locations or sizes, then the vendor can implement new versions of the APIs in a chipset specific file such as CHIPSET_allocateheap.c and build this new file instead of up_allocate.c file.
 - The vendor needs to take care that the prototypes of the APIs are not changed in their new implementation.
+
+## Available APIs
+Some APIs are not available on Protected build because kernel and user binaries are separated.
```
sangam-swami, line 145: Some APIs are not available => Some APIs can't be used ?

sunghan-chang: I modified as you said. But I've added more words as below:
```
Some APIs like scheduler APIs can't be used on user application binary with Protected build ....
```
Because it is not available in user side and I thought it was better to give the example of APIs.

docs/ProtectedBuildGuide.md
```diff
@@ -139,3 +140,14 @@ The following sections give detailed explanation regarding some of the steps / c
 - The user heap starts at the end of the user bss section and spans the entire remaining space in the user ram region.
 - However, if the chipset has a memory configuration that is different from the default one or if the vendor would like to initialize the kernel and user heaps at different locations or sizes, then the vendor can implement new versions of the APIs in a chipset specific file such as CHIPSET_allocateheap.c and build this new file instead of up_allocate.c file.
 - The vendor needs to take care that the prototypes of the APIs are not changed in their new implementation.
+
+## Available APIs
+Some APIs like scheduler APIs can't be used on user application binary with Protected build because kernel and user binaries are separated.
```
jeongchanKim, line 145: Some scheduler APIs like sched_getprioritymax, task_setcanceltype can be used from user... So I think we'd better change the example...

sunghan-chang: I've mentioned exact one example API (sched_gettcb API) for that.

docs/ProtectedBuildGuide.md
sangam-swami: yes looks good

PR #4312, opened 2020-06-12
examples/wifi_manager_sample: Modify WIFI Manager cmd on input checking

fengyunlai:
Modify Wifi manager tash command to check softap cmd argc, and sta connection arg copy.

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test.c
+++ b/apps/examples/wifi_manager_sample/wm_test.c
@@ -363,7 +363,7 @@ static void print_wifi_softap_profile(wifi_manager_softap_config_s *config, char
 static wifi_manager_ap_auth_type_e get_auth_type(const char *method)
 {
 	char data[20];
-	strcpy(data, method);
+	strncpy(data, method, 20);
 
 	char *result[3];
 	char *next_ptr;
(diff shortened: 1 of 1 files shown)


apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -962,6 +962,9 @@ static wm_test_e _wm_get_opt(int argc, char *argv[])
 
 static int _wm_test_softap(struct options *opt, int argc, char *argv[])
 {
+	if (argc < 5) {
+		return -1;
```
jsdosa, line 966: could you add log for error?

sunghan-chang:
@fengyunlai I agree this change but it is diffcult to know where and why you modify from commit title and descrption. We recommend to see our commit rules.

sunghan-chang:
```
examples/wifi_manager_sample: Modify WIFI Manager cmd on input checking

why...
```

apps/examples/wifi_manager_sample/wm_test.c
fengyunlai: Thank you for review. Since by right wm test will give USAGE info message if directly return, which means wrong input parameters. May I ask what log errors you want me to add?

sunghan-chang: Let's do that by wm engineer later if it requires.

sunghan-chang:
```sh
$ git commit --amend

examples/wifi_manager_sample: Modify WIFI Manager cmd on input checking

1. For wm_test softap, check cmd input parameter's argument count, if wrong such as missing password, return directly.
2. In get_auth_type(), change strcpy() to strncpy() to avoid exiting the data lenth when string copying.
```

PR #4313, opened 2020-06-15
mm/mm_heap/mm_memalign : Add re-search the aligned address inside the…

jeongchanKim:
… node

Even though there is next aligned address in the node, memalign will be fail only with checking the first-meet aligned address.
So add re-search logic inside the selected node.

### DIFF
--- a/os/mm/mm_heap/mm_memalign.c
+++ b/os/mm/mm_heap/mm_memalign.c
@@ -176,15 +176,24 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
 
 		/* Now, traverse the list in reverse direction, towards bigger size nodes */
 		for ( ; node; node = node->blink) {
-			alignchunk = (FAR struct mm_allocnode_s *)(((size_t)node + SIZEOF_MM_ALLOCNODE + mask) & ~mask);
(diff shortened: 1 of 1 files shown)


os/mm/mm_heap/mm_memalign.c
```diff
@@ -176,17 +176,25 @@ FAR void *mm_memalign(FAR struct mm_heap_s *heap, size_t alignment, size_t size)
...
-			size_t remainsize = (size_t)alignchunk - SIZEOF_MM_ALLOCNODE - (size_t)node;
-
-			/* We found a suitable node if node size is more than required size after alignment and
-			 * if the remaining bytes before the alignment point is either zero or bigger than freenode.
-			 */
-			if (node->size >= alignsize && (remainsize == 0 || remainsize >= SIZEOF_MM_FREENODE)) {
-				found_align = true;
-				break;
-			}
+		    for (alignchunk = (FAR struct mm_allocnode_s *)(((size_t)node + SIZEOF_MM_ALLOCNODE + mask) & ~mask);
```
sunghan-chang, line 179: Could you add the explanation of this for loop?

sangam-swami:
Even though there is next aligned address in the node, memalign will be fail only with checking the first-meet aligned address. => 
Even though there is next aligned address available in the current free node, memalign will fail only after checking the first-met aligned address.
can we change like this

PR #4315, opened 2020-06-15
os/binfmt: Add missing macros for elf_dumpbuffer

sangam-swami:
Add missing debug macros for CONFIG_ELF_DUMPBUFFER

### DIFF
--- a/os/binfmt/elf.c
+++ b/os/binfmt/elf.c
@@ -74,14 +74,6 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* CONFIG_DEBUG, CONFIG_DEBUG_INFO, and CONFIG_DEBUG_BINFMT have to be
- * defined or CONFIG_ELF_DUMPBUFFER does nothing.
- */
-
-#if !defined(CONFIG_DEBUG_INFO) || !defined(CONFIG_DEBUG_BINFMT)
-#undef CONFIG_ELF_DUMPBUFFER
-#endif
-
 #ifndef CONFIG_ELF_STACKSIZE
 #define CONFIG_ELF_STACKSIZE 2048
 #endif
--- a/os/binfmt/libelf/Kconfig
+++ b/os/binfmt/libelf/Kconfig
@@ -35,7 +35,7 @@ config ELF_BUFFERINCR
 config ELF_DUMPBUFFER
 	bool "Dump ELF buffers"
 	default n
-	depends on DEBUG_INFO
+	depends on DEBUG_BINFMT_INFO
 	---help---
 		Dump various ELF buffers for debug purposes
 
--- a/os/binfmt/libelf/libelf_bind.c
+++ b/os/binfmt/libelf/libelf_bind.c
@@ -73,14 +73,6 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* CONFIG_DEBUG_FEATURES, CONFIG_DEBUG_INFO, and CONFIG_DEBUG_BINFMT have to be
- * defined or CONFIG_ELF_DUMPBUFFER does nothing.
- */
-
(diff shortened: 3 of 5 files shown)


os/binfmt/elf.c
```diff
@@ -74,11 +74,11 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* CONFIG_DEBUG, CONFIG_DEBUG_INFO, and CONFIG_DEBUG_BINFMT have to be
+/* CONFIG_DEBUG, CONFIG_DEBUG_VERBOSE, and CONFIG_DEBUG_BINFMT have to be
  * defined or CONFIG_ELF_DUMPBUFFER does nothing.
  */
 
-#if !defined(CONFIG_DEBUG_INFO) || !defined(CONFIG_DEBUG_BINFMT)
+#if !defined(CONFIG_DEBUG_VERBOSE) || !defined(CONFIG_DEBUG_BINFMT)
```
Taejun-Kwon, line 81: Use CONFIG_DEBUG_BINFMT_INFO stand alone is not enough?

sangam-swami: done

os/binfmt/elf.c
```diff
@@ -74,11 +74,11 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
-/* CONFIG_DEBUG, CONFIG_DEBUG_INFO, and CONFIG_DEBUG_BINFMT have to be
+/* CONFIG_DEBUG, CONFIG_DEBUG_VERBOSE, and CONFIG_DEBUG_BINFMT have to be
  * defined or CONFIG_ELF_DUMPBUFFER does nothing.
  */
 
-#if !defined(CONFIG_DEBUG_INFO) || !defined(CONFIG_DEBUG_BINFMT)
+#if !defined(CONFIG_DEBUG_BINFMT_INFO)
```
sunghan-chang, line 81: Is this necessary to be two steps?
```c
#if !defined(CONFIG_DEBUG_BINFMT_INFO)
#undef CONFIG_ELF_DUMPBUFFER
#endif

#ifdef CONFIG_ELF_DUMPBUFFER
#define elf_dumpbuffer(m, b, n) binfodumpbuffer(m, b, n)
#else
#define elf_dumpbuffer(m, b, n)
#endif
```

How about checking it directly?
```c
#ifdef CONFIG_DEBUG_BINFMT_INFO
#define elf_dumpbuffer(m, b, n) binfodumpbuffer(m, b, n)
#else
#define elf_dumpbuffer(m, b, n)
#endif
```

sangam-swami: if CONFIG_DEBUG_BINFMT_INFO enabled and CONFIG_ELF_DUMPBUFFER is not, then it should not set elf_dumpbuffer, so I think the current approach is fine

sangam-swami: or we can replace with :
#ifdef CONFIG_DEBUG_BINFMT_INFO
#ifdef CONFIG_ELF_DUMPBUFFER
#define elf_dumpbuffer(m, b, n) binfodumpbuffer(m, b, n)
#else
#define elf_dumpbuffer(m, b, n)
#endif
#endif

sunghan-chang: os/binfmt/libelf/Kconfig
```kconfig
config ELF_DUMPBUFFER
	bool "Dump ELF buffers"
	default n
	depends on DEBUG_BINFMT_INFO  <---- here
	---help---
		Dump various ELF buffers for debug purposes
```

in here, os/binfmt/elf.c
```c
remove 1st conditional  <--- here

#ifdef CONFIG_ELF_DUMPBUFFER
#define elf_dumpbuffer(m, b, n) binfodumpbuffer(m, b, n)
#else
#define elf_dumpbuffer(m, b, n)
#endif
```

This is enough and is the same as current pr

sangam-swami: done

PR #4317, opened 2020-06-16
binfmt/libelf: Free memory if it is allocated memory

jeongarmy:
The memory should be freed when it is allocated.

### DIFF
--- a/os/mm/kmm_heap/kmm_free.c
+++ b/os/mm/kmm_heap/kmm_free.c
@@ -92,8 +92,12 @@
 
 void kmm_free(FAR void *mem)
 {
-	struct mm_heap_s *kheap = mm_get_heap(mem);
-	DEBUGASSERT(kmm_heapmember(mem));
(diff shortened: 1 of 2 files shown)


os/mm/kmm_heap/kmm_free.c
```diff
@@ -92,8 +92,12 @@
...
 void kmm_free(FAR void *mem)
 {
-	struct mm_heap_s *kheap = mm_get_heap(mem);
-	DEBUGASSERT(kmm_heapmember(mem));
-	mm_free(kheap, mem);
+	struct mm_heap_s *kheap;
+
+	if (mem) {
+		kheap = mm_get_heap(mem);
+		DEBUGASSERT(kmm_heapmember(mem));
```
jeongchanKim, line 99: How about changing the position between line 98 and 99?
We don't need to call mm_get_heap if mem is not in kernel heap.

PR #4328, opened 2020-06-25
[Realtek][RTL8721CSM]Add General Purpose Timer Peripheral

AdamZhang0124:
1) The timer peripheral is now supported.
2) It is tested by the timer example (timer_main.c).
3) Currently, the 'getstatus' and 'ioctl' functions are not supported.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
+++ b/os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
@@ -142,6 +142,8 @@ static struct amebad_gpt_lowerhalf_s g_gpt3_lowerhalf = {
 static void amebad_timout_handler(uint32_t data)
 {
 	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)data;
+	DEBUGASSERT(priv);
+
 	uint32_t next_interval_us = 0;
 
 	if (priv->callback(&next_interval_us, priv->arg)) {
@@ -179,7 +181,7 @@ static int amebad_gpt_start(struct timer_lowerhalf_s *lower)
 		}
 		gtimer_start_periodical(&priv->obj, priv->gpt_timeout, (void *)amebad_timout_handler, priv);
 		priv->started = true;
-		gtimer_reload(&priv->obj, 1000000);
+
 		tmrvdbg("Timer %d is started, callback %s\n", priv->obj.timer_id, priv->callback == NULL ? "none" : "set");
 		return OK;
 	}
@@ -206,6 +208,7 @@ static int amebad_gpt_start(struct timer_lowerhalf_s *lower)
 static int amebad_gpt_stop(struct timer_lowerhalf_s *lower)
 {
 	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
 
 	if (priv->started) {
 		gtimer_deinit(&priv->obj);
@@ -238,6 +241,7 @@ static int amebad_gpt_stop(struct timer_lowerhalf_s *lower)
 static int amebad_gpt_settimeout(struct timer_lowerhalf_s *lower, uint32_t timeout)
 {
 	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
 
 	if (priv->started) {
 		return -EPERM;
@@ -253,7 +257,7 @@ static int amebad_gpt_settimeout(struct timer_lowerhalf_s *lower, uint32_t timeo
  * Name: amebad_gpt_setcallback
  *
  * Description:
- *   Call this user provided timeout callback.
+ *   Set the user provided timeout callback.
  *
  * Input Parameters:
  *   lower    - A pointer the publicly visible representation of the
@@ -271,6 +275,8 @@ static int amebad_gpt_settimeout(struct timer_lowerhalf_s *lower, uint32_t timeo
 static void amebad_gpt_setcallback(struct timer_lowerhalf_s *lower, tccb_t callback, void *arg)
 {
 	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+
 	irqstate_t flags = irqsave();
 
 	/* Save the new callback */
@@ -307,15 +313,17 @@ int amebad_timer_initialize(const char *devpath, int timer)
 		break;
 
 	case TIMER2:
-		printf("Timer2\n");
 		priv = &g_gpt2_lowerhalf;
 		priv->obj.timer_id = TIMER2;
 		break;
 
 	case TIMER3:
-		printf("Timer3\n");
 		priv = &g_gpt3_lowerhalf;
 		priv->obj.timer_id = TIMER3;
+		break;
+
+	default: 
+		priv = NULL;
 		break;
 	}
 
--- a/os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
+++ b/os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
@@ -75,15 +75,15 @@
 /****************************************************************************
  * Private Types
  ****************************************************************************/
-
 struct amebad_gpt_lowerhalf_s {
 	const struct timer_ops_s *ops;	/* Lowerhalf operations */
 	gtimer_t obj;
-	uint8_t irq_id;
 	bool started;				/* True: Timer is started */
 	tccb_t callback;
 	void *arg;					/* Argument passed to upper half callback */
 	uint32_t gpt_timeout;
+	bool freerunmode;			/* True: Free Run Mode. False: Alarm Mode */
+	uint32_t freeruncount;
 };
 
 /****************************************************************************
@@ -102,10 +102,10 @@ static int amebad_gpt_ioctl(struct timer_lowerhalf_s *lower, int cmd, unsigned l
 static const struct timer_ops_s g_timer_ops = {
 	.start = amebad_gpt_start,
 	.stop = amebad_gpt_stop,
-	.getstatus = NULL,
+	.getstatus = amebad_gpt_getstatus,
 	.settimeout = amebad_gpt_settimeout,
 	.setcallback = amebad_gpt_setcallback,
-	.ioctl = NULL,
+	.ioctl = amebad_gpt_ioctl,
 };
 
 static struct amebad_gpt_lowerhalf_s g_gpt0_lowerhalf = {
@@ -138,21 +138,23 @@ static struct amebad_gpt_lowerhalf_s g_gpt3_lowerhalf = {
  * Returned Values:
  *
  ****************************************************************************/
-
-static void amebad_timout_handler(uint32_t data)
+static void amebad_gpt_handler(uint32_t data)
 {
 	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)data;
 	DEBUGASSERT(priv);
-
 	uint32_t next_interval_us = 0;
 
-	if (priv->callback(&next_interval_us, priv->arg)) {
-		if (next_interval_us > 0) {
-			priv->gpt_timeout = next_interval_us;
-			gtimer_reload(&priv->obj, priv->gpt_timeout);
+	if (priv->freerunmode) {	/* Free Run Mode */
+		priv->freeruncount++;
+	} else {					/* Alarm Mode */
+		if (priv->callback(&next_interval_us, priv->arg)) {
+			if (next_interval_us > 0) {
+				priv->gpt_timeout = next_interval_us;
+				gtimer_reload(&priv->obj, priv->gpt_timeout);
+			}
+		} else {
+			amebad_gpt_stop((struct timer_lowerhalf_s *)priv);
 		}
-	} else {
-		amebad_gpt_stop((struct timer_lowerhalf_s *)priv);
 	}
 }
 
@@ -176,10 +178,19 @@ static int amebad_gpt_start(struct timer_lowerhalf_s *lower)
 	DEBUGASSERT(priv);
 
 	if (!priv->started) {
-		if (priv->callback != NULL) {
-			gtimer_init(&priv->obj, priv->obj.timer_id);
+		if (priv->freerunmode) {	/* Free Run Mode */
+			priv->obj.handler = (void *)amebad_gpt_handler;
+			priv->obj.hid = priv;
+			gtimer_start_periodical(&priv->obj, 0xFFFFFFFF, (void *)amebad_gpt_handler, priv->obj.hid);
+		} else {				/* Alarm Mode */
+			if (priv->callback != NULL) {
+				priv->obj.handler = (void *)amebad_gpt_handler;
+				priv->obj.hid = priv;
+				gtimer_start_periodical(&priv->obj, priv->gpt_timeout, (void *)amebad_gpt_handler, priv->obj.hid);
+			} else {
+				gtimer_start_periodical(&priv->obj, priv->gpt_timeout, NULL, NULL);
+			}
 		}
-		gtimer_start_periodical(&priv->obj, priv->gpt_timeout, (void *)amebad_timout_handler, priv);
 		priv->started = true;
 
 		tmrvdbg("Timer %d is started, callback %s\n", priv->obj.timer_id, priv->callback == NULL ? "none" : "set");
@@ -221,6 +232,51 @@ static int amebad_gpt_stop(struct timer_lowerhalf_s *lower)
 	/* Return ENODEV to indicate that the timer was not running */
 	tmrdbg("Error!! Timer %d is not running\n", priv->obj.timer_id);
 	return -ENODEV;
+}
+
+/****************************************************************************
+ * Name: amebad_gpt_getstatus
+ *
+ * Description:
+ *   get GPT status
+ *
+ * Input Parameters:
+ *   lower   - A pointer the publicly visible representation of the
+ *             "lower-half" driver state structure.
+ *   status  - A pointer saved current GPT status
+ *
+ * Returned Values:
+ *   Zero on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+static int amebad_gpt_getstatus(struct timer_lowerhalf_s *lower, struct timer_status_s *status)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	uint32_t ticks;
+
+	/* Return the status bit */
+	status->flags = 0;
+	if (priv->started) {
+		status->flags |= TCFLAGS_ACTIVE;
+	}
+
+	if (priv->callback) {
+		status->flags |= TCFLAGS_HANDLER;
+	}
+
+	/* Return the actual timeout in microseconds */
+	status->timeout = priv->gpt_timeout;
+
+	/* Get the time remaining until the timer expires (in microseconds). */
+	ticks = gtimer_read_tick(&priv->obj);
+
+	if (priv->freerunmode) {	/* Free Run Mode */
+		status->timeleft = (uint64_t)((float)ticks * (1000000 / 32768) + (priv->freeruncount * 0xFFFFFFFF));
+	} else {					/* Alarm Mode */
+		status->timeleft = status->timeout - ticks * (1000000 / 32768);
+	}
+
+	return OK;
 }
 
 /****************************************************************************
@@ -285,13 +341,84 @@ static void amebad_gpt_setcallback(struct timer_lowerhalf_s *lower, tccb_t callb
 
 	/* If no callback is attached, the timer stop at the first interrupt */
 	if (callback != NULL && priv->started) {
-		priv->obj.handler = (void *)amebad_timout_handler;
-		gtimer_init(&priv->obj, priv->obj.timer_id);
+		priv->obj.handler = (void *)amebad_gpt_handler;
+		priv->obj.hid = priv;
 	} else {
-		gtimer_deinit(&priv->obj);
+		priv->obj.handler = NULL;
+		priv->obj.hid = NULL;
 	}
 
 	irqrestore(flags);
+}
+
+/****************************************************************************
+ * Name: amebad_gpt_ioctl
+ *
+ * Description:
+ *   Any ioctl commands that are not recognized by the "upper-half" driver
+ *   are forwarded to the lower half driver through this method.
+ *
+ * Input Parameters:
+ *   lower - A pointer the publicly visible representation of the "lower-half"
+ *           driver state structure.
+ *   cmd   - The ioctl command value
+ *   arg   - The optional argument that accompanies the 'cmd'.  The
+ *           interpretation of this argument depends on the particular
+ *           command.
+ *
+ * Returned Values:
+ *   Zero on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+static int amebad_gpt_ioctl(struct timer_lowerhalf_s *lower, int cmd, unsigned long arg)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+
+	int ret = -EINVAL;
+
+	tmrvdbg("GPT IOCTL cmd: %d, arg: %ld\n", cmd, arg);
+
+	switch (cmd) {
+	case TCIOC_SETDIV:
+		ret = OK;
+		break;
+	case TCIOC_GETDIV:
+		ret = OK;
+		break;
+	case TCIOC_SETMODE:
+		if ((timer_mode_t) arg == MODE_FREERUN) {
+			priv->freerunmode = true;
+			priv->freeruncount = 0;
+			ret = OK;
+		} else if ((timer_mode_t) arg == MODE_ALARM) {
+			priv->freerunmode = false;
+			priv->freeruncount = 0;
+			ret = OK;
+		} else {
+			ret = -EINVAL;
+		}
+		break;
+	case TCIOC_SETRESOLUTION:
+		if ((timer_resolution_t) arg == TIME_RESOLUTION_MS) {
+			// TO-DO develop
+			ret = OK;
+		} else {
+			ret = -EINVAL;
+		}
+		break;
+	case TCIOC_SETIRQPRIO:
+		ret = up_prioritize_irq(TIM_x[priv->obj.timer_id], arg);
+		break;
+	case TCIOC_SETCLKSRC:
+		ret = OK;
+		break;
+	default:
+		tmrdbg("Invalid cmd %d\n", cmd);
+		break;
+	}
+
+	return ret;
 }
 
 /****************************************************************************
@@ -322,7 +449,7 @@ int amebad_timer_initialize(const char *devpath, int timer)
 		priv->obj.timer_id = TIMER3;
 		break;
 
-	default: 
+	default:
 		priv = NULL;
 		break;
 	}
@@ -335,12 +462,14 @@ int amebad_timer_initialize(const char *devpath, int timer)
 	priv->started = false;
 	priv->callback = NULL;
 
+	/* Initializes the timer device, include timer registers and interrupt */
+	gtimer_init(&priv->obj, priv->obj.timer_id);
+
 	/*
 	 * Register the timer driver as /dev/timerX.  The returned value from
 	 * timer_register is a handle that could be used with timer_unregister().
 	 * REVISIT: The returned handle is discard here.
 	 */
-
 	if (!timer_register(devpath, (struct timer_lowerhalf_s *)priv)) {
 		/*
 		 * The actual cause of the failure may have been a failure to allocate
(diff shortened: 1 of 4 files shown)


os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+	priv->gpt_timeout = timeout;
+
+	return OK;
+}
+
+/****************************************************************************
+ * Name: amebad_gpt_setcallback
+ *
+ * Description:
+ *   Call this user provided timeout callback.
```
jeongchanKim, line 256: This function sets the user provided callback, not call that.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+ *  arg       - Argument that will be provided in the callback
+ *
+ * Returned Values:
+ *   The previous timer expiration function pointer or NULL is there was
+ *   no previous function pointer.
+ *
+ ****************************************************************************/
+static void amebad_gpt_setcallback(struct timer_lowerhalf_s *lower, tccb_t callback, void *arg)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
```
jeongchanKim, line 273: You'd better add `DEBUGASSERT(priv)` for checking.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+		break;
+
+	case TIMER3:
+		printf("Timer3\n");
+		priv = &g_gpt3_lowerhalf;
+		priv->obj.timer_id = TIMER3;
+		break;
+	}
+
+	if (!priv) {
```
Taejun-Kwon, line 322: I think this never comes..

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+static const struct timer_ops_s g_timer_ops = {
+	.start = amebad_gpt_start,
+	.stop = amebad_gpt_stop,
+	.getstatus = NULL,
+	.settimeout = amebad_gpt_settimeout,
+	.setcallback = amebad_gpt_setcallback,
+	.ioctl = NULL,
+};
+
+static struct amebad_gpt_lowerhalf_s g_gpt0_lowerhalf = {
```
Taejun-Kwon, line 111: What is diffrent each of timer here? for example resolution.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+ *
+ * Input Parameters:
+ *
+ * Returned Values:
+ *
+ ****************************************************************************/
+
+static void amebad_timout_handler(uint32_t data)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)data;
```
Taejun-Kwon, line 144: we need null check for priv here.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,348 @@
...
+/****************************************************************************
+ * Private Data
+ ****************************************************************************/
+static const struct timer_ops_s g_timer_ops = {
+	.start = amebad_gpt_start,
+	.stop = amebad_gpt_stop,
+	.getstatus = NULL,
+	.settimeout = amebad_gpt_settimeout,
+	.setcallback = amebad_gpt_setcallback,
+	.ioctl = NULL,
```
Taejun-Kwon, line 108: we need ioctl & get status to get some info & change some modes.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/timer_api.c
```diff
@@ -79,8 +79,14 @@ void gtimer_init (gtimer_t *obj, uint32_t tid)
 	TIM_InitStruct.TIM_UpdateEvent = ENABLE; /* UEV enable */
 	TIM_InitStruct.TIM_UpdateSource = TIM_UpdateSource_Overflow;
 	TIM_InitStruct.TIM_ARRProtection = ENABLE;
-	
+#ifndef CONFIG_PLATFORM_TIZENRT_OS
 	RTIM_TimeBaseInit(TIM_x[tid], &TIM_InitStruct, TIMx_irq[tid], (IRQ_FUN) gtimer_timeout_handler, (u32)obj);
+#else
+	RTIM_TimeBaseInit(TIM_x[tid], &TIM_InitStruct, TIMx_irq[tid], NULL, NULL);
```
Taejun-Kwon, line 85: Could you please add some comment regarding this?
If below is added to enable/disable interrupt, then amebad_timer_lowerhalf.c need to be fix.
(For example disable interrupt in setcallback if priv->callback is null)
I am right, it seems that start & stop also need to be fixed.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Noted, for that, I just followed the common description used by other platforms.  
I will change the description to 'Set the user provided timeout callback.' Thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Sure, I do have missed that, thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Sure, will add DEBUGASSERT(priv) for all the priv

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Understood. Since in 'timer.c', the getstatus and ioctl functions are marked as optional, only the essential functions are supported in this PR. The getstatus and ioctl will be supported in subsequent update.
Moreover, as so far timer_main.c is the only test case I can use to verify the functions, the ioctl commands used in timer ioctl function (eg. TCIOC_SETDIV) are not included in the test case. Therefore, I am unable to verify them.
Could you please suggest:
1) what ioctl commands need to be supported in this timer ioctl function.
2) the description of those commands and what parameters should I return.
Thanks

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/timer_api.c
AdamZhang0124: This change is made in order to init the handler and enable the interrupt via Tizenrt_service.c instead of via rom code.
For the "disable interrupt in setcallback if priv->callback is null", I have added gtimer_deinit to disable the interrupt.
For "start & stop also need to be fixed", may I ask for the detail about it?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Timer 0-3 are the same, they operate with clock source of 32kHz, and the resolution is 32 bits .

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: You are correct, I will add a default case for that.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/timer_api.c
AdamZhang0124: Yes, gtimer_init is enough. It sets ISR & enable interrupt and gtimer_deinit disables timer, closes and clears interrupt.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
jsdosa: i'm so sorry, Now I have confirmed this comment.
Could you check imxrt_timer_lowhalf.c ?
If you have any questions, ask me.

sunghan-chang: I think below is necessary.
    1. getstatus to show whether it is started, whether cb is registered, timer resolution (ms or us) and timeleft or current time (or ticks)
    2. set timer resolution (ms or us)
    3. set mode (alarm or freerun)
    4. and something else to control timer

AdamZhang0124: Hi @jsdosa, thanks for the reply.

I have the following questions to ask:
1) As we don’t have the option to change the clock divider, resolution, clock source, in this case, what should I add for those ioctl commands?
2) For the free run mode 'CONFIG_EXAMPLES_TIMER_FRT_MEASUREMENT', I would like to know what behaviour do you expect the timer to do?
3) How should I verify those functions after I finished the update. As the timer_main is the only test case I know and I think it does not cover all the tests.

Thanks!

sunghan-chang: 1. Does it support both alarm mode and freerun mode? If it is, mode change is necessary.
  Do you mean rtl gpt has fixed resolution? Could you let use know the resolution?
2. That config is usages example of FRT (FreeRun Timer). It just requests getting current time or tick after starting.
  You could find the common behavior by googling of "free run timer" or "free running timer".
3. The timer options are depended on the chipset so that we can't give and cover all of option testing. But the mandatory options can be tested by running timer example which is in apps/examples/timer.

AdamZhang0124: > I think below is necessary.
> 1. getstatus to show whether it is started, whether cb is registered, timer resolution (ms or us) and timeleft or current time (or ticks)
> 2. set timer resolution (ms or us)
> 3. set mode (alarm or freerun)
> 4. and something else to control timer

Hi @sunghan-chang ,

Sorry, just now I did not refresh the webpage, so I did not see your reply before I asked the questions.

For your reply,
1. I have done the getstatus function, and it will be committed with the rest of the changes.
2. The resolution of the timer is 32 bits and it shows in us.
3. For the timer mode, currently when I run the timer_main example, the timer is able to perform 10 times of notification before stop. Is it the free run mode behaviour you are excepting? 
4. If so, what behaviour of the alarm mode should it perform? Please advise. 

Many thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,356 @@
...
+
+static struct amebad_gpt_lowerhalf_s g_gpt3_lowerhalf = {
+	.ops = &g_timer_ops,
+};
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+/****************************************************************************
+ * Name: amebad_gpt_handler
```
jeongchanKim, line 131: This is not matched with below function name.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/timer_api.c
Taejun-Kwon: Got it but please add description to use it. If it is not useful, I mean gtimer_init & deinit is enough, then remove it. Build warning can be occured.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Yes, you are right, that will be updated with next major update.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/timer_api.c
AdamZhang0124: Actually, the original 'RTIM_TimeBaseInit' includes ISR set and interrupt enable. But it was registered in the rom code. I made the change to prevent the registration in rom code via RTIM_TimeBaseInit. So that the interrupt and ISR can be set via Tizenrt_service.c instead.
I will add description for that, thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
sunghan-chang: > 3. For the timer mode, currently when I run the timer_main example, the timer is able to perform 10 times of notification before stop. Is it the free run mode behaviour you are excepting?

It isn't the free-run mode behavior, it's the timer example operation. If you don't give the number of repeat with -n option, it has 10 iterate times by default.
```
In help message
		"  -n [REPEAT]  Number of samples to repeat test.\n" \
		"               DEFAULT 10\n" \

In source code
	if (repeat <= 0) {
		repeat = EXAMPLE_TIMER_NSAMPLES;
	}
```

> 4. If so, what behaviour of the alarm mode should it perform? Please advise.

That's the normal timer. set time and get notification after expired.

sunghan-chang: > 2. The resolution of the timer is 32 bits and it shows in us.

Could you confirm that RTL8721 supports GPT in us only, it does not support ms resolution? It's the same as not support clock source and divider.

AdamZhang0124: > Could you confirm that RTL8721 supports GPT in us only, it does not support ms resolution? It's the same as not support clock source and divider.

Rtl8721 uses only the fix clock source of 32kHz, it can provide a tick per 31us. So that the resolution, clock source and divider is fixed. 
Also, as the resolution is 31us, it should be classified as in ms level if that is the meaning of the resolution.

sunghan-chang: Ok. RTL8721 supports the timer in ms.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * Returned Values:
+ *   The previous timer expiration function pointer or NULL is there was
+ *   no previous function pointer.
+ *
+ ****************************************************************************/
+static void amebad_gpt_setcallback(struct timer_lowerhalf_s *lower, tccb_t callback, void *arg)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
```
sunghan-chang, line 278: Sorry but DEBUGASSERT is not enough.
DEBUGASSERT is working only with CONFIG_DEBUG enabled. When CONFIG_DEBUG is disabled and priv is null, the system could be crashed due to null dereference like `priv ->callback`. Please check that with if condition.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
sunghan-chang: Could you let us know what next major update is? Do you plan to support getstatus with next update?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+ ****************************************************************************/
+struct amebad_gpt_lowerhalf_s {
+	const struct timer_ops_s *ops;	/* Lowerhalf operations */
+	gtimer_t obj;
+	bool started;				/* True: Timer is started */
+	tccb_t callback;
+	void *arg;					/* Argument passed to upper half callback */
+	uint32_t gpt_timeout;
+	bool freerunmode;			/* True: Free Run Mode. False: Alarm Mode */
+	uint32_t freeruncount;
```
sunghan-chang, line 86: Could you let me know what this is, purpose is?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+ *
+ ****************************************************************************/
+static int amebad_gpt_start(struct timer_lowerhalf_s *lower)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+
+	if (!priv->started) {
+		if (priv->freerunmode) {	/* Free Run Mode */
+			priv->obj.handler = (void *)amebad_gpt_handler;
```
sunghan-chang, line 182: You set this twice, one is here, another is in amebad_gpt_setcallback. I think you could remove duplicated.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+
+	int ret = -EINVAL;
+
+	tmrvdbg("GPT IOCTL cmd: %d, arg: %ld\n", cmd, arg);
+
+	switch (cmd) {
+	case TCIOC_SETDIV:
+		ret = OK;
+		break;
+	case TCIOC_GETDIV:
```
sunghan-chang, line 386: Let's remove un-supported commands to return EINVAL.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+		ret = OK;
+		break;
+	case TCIOC_SETMODE:
+		if ((timer_mode_t) arg == MODE_FREERUN) {
+			priv->freerunmode = true;
+			priv->freeruncount = 0;
+			ret = OK;
+		} else if ((timer_mode_t) arg == MODE_ALARM) {
+			priv->freerunmode = false;
+			priv->freeruncount = 0;
```
sunghan-chang, line 396: Even timer is the alarm mode, is this necessary to be initialize?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+		break;
+	case TCIOC_GETDIV:
+		ret = OK;
+		break;
+	case TCIOC_SETMODE:
+		if ((timer_mode_t) arg == MODE_FREERUN) {
+			priv->freerunmode = true;
+			priv->freeruncount = 0;
+			ret = OK;
+		} else if ((timer_mode_t) arg == MODE_ALARM) {
```
sunghan-chang, line 394: Could you remove a space after type casting? There are many.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
sunghan-chang: Could you check previous comment?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Hi @sunghan-chang @jsdosa ,
As required, the getstatus , ioctl functions as well as the free run feature have been added in the newest update. Please help to review. Many thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+ *
+ ****************************************************************************/
+static void amebad_gpt_handler(uint32_t data)
+{
+	struct amebad_gpt_lowerhalf_s *priv = (struct amebad_gpt_lowerhalf_s *)data;
+	DEBUGASSERT(priv);
+	uint32_t next_interval_us = 0;
+
+	if (priv->freerunmode) {	/* Free Run Mode */
+		priv->freeruncount++;
```
sunghan-chang, line 148: Could you explain the meaning of this logic? Is there any case getting interrupt with free-run mode?

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+#include <tinyara/config.h>
+
+#include <sys/types.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <tinyara/kmalloc.h>
```
sunghan-chang, line 64: unnecessary.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: > Could you let me know what this is, purpose is?

It is used to adapt the application of free run mode for timer_main. The free run timer will continuously record the time elapse for the use by the getstatus function.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: I have checked it again, this duplicate is necessary. It happens that callback function is called before timer is started. So that 'callback is not NULL but priv->started is false'(Line343). In this case, the callback is assigned but the handler is not assigned. If the timer starts then, the handler need to be assigned again(Line186).

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Sure!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: No need, you are right! The 'freeruncount' is no need to initialize in alarm mode, unless there is a mode changing application for that timer.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Ok, they were auto-changed after using the cformatter. I will manually change them.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: As our timers do not have the free run mode setting, the free running feature has to be implemented in this logic. So that, maximum timeout is firstly set to that free run timer, when it timeout, the interrupt is triggered just to record the times of timeout. Then, the freeruncount is used to calculate the total time elapse.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: That will be removed, thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
Taejun-Kwon: without this state, you can return -ENODEV in default.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
sunghan-chang: Yeah it does, we try to fix it several times but we can't. Sorry for inconvenience but that's why we write below in readme of formatter.
```
Note: This sometimes makes incorrect changes.
Usually it adds a space between type casting and variable as shown below:
```

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
sunghan-chang: Let me summarize.
1. RTL8721 hw timer does not support free-run mode.
2. Even though 1st, you will support that in sw driver.
   When user sets it as free-run timer, amebad timer driver runs alam mode with maximum timeout value.
   When user calls get status function to get the running time, driver gets the value from alarm mode register and calculates how many times it is expired to get total time.

Is this right?

If it is, let's rename `freeruncount` like `num_of_xxxms_expired`. `xxx` could be calculated using your maximum value.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
Taejun-Kwon: > > Could you let me know what this is, purpose is?
> 
> It is used to adapt the application of free run mode for timer_main. The free run timer will continuously record the time elapse for the use by the getstatus function.

I think it is not necessary, if priv->started && priv->mode == freerunmode, it should record.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+ * Private Types
+ ****************************************************************************/
+struct amebad_gpt_lowerhalf_s {
+	const struct timer_ops_s *ops;	/* Lowerhalf operations */
+	gtimer_t obj;
+	bool started;				/* True: Timer is started */
+	tccb_t callback;
+	void *arg;					/* Argument passed to upper half callback */
+	uint32_t gpt_timeout;
+	bool freerunmode;			/* True: Free Run Mode. False: Alarm Mode */
```
Taejun-Kwon, line 85: Please use some integer type.
We defined timer_mode_e, pleas see os\include\tinyara\timer.h

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+
+	/* Save the new callback */
+	priv->callback = callback;
+	priv->arg = arg;
+
+	/* If no callback is attached, the timer stop at the first interrupt */
+	if (callback != NULL && priv->started) {
+		priv->obj.handler = (void *)amebad_gpt_handler;
+		priv->obj.hid = priv;
+	} else {
```
Taejun-Kwon, line 346: Could you please see below comments?
1. Mode is free-runcount, we can't register. In my opinion.
2. If mode is not alarm, we had better disable interrupt, right?
Also disable ISR.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+	case TCIOC_SETDIV:
+		ret = OK;
+		break;
+	case TCIOC_GETDIV:
+		ret = OK;
+		break;
+	case TCIOC_SETMODE:
+		if ((timer_mode_t) arg == MODE_FREERUN) {
+			priv->freerunmode = true;
+			priv->freeruncount = 0;
```
Taejun-Kwon, line 392: you already using timer_mode_e,
then priv->mode will be better than priv->freeruncount.
Of course type is integer.

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: > Could you check previous comment?

Sure, I was following the coding style that other in use platforms are using. I will also add 'if (priv)' along with DEBUGASSERT for all the functions. Thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Oh, I did not noticed the note. I will check that again after I use the formatter in future. Thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Yes, your understanding is right! I will change the name and add description for it. The max value can be set as 0xFFFFFFFF, so that the name as you suggest will be 'num_of_4294967ms_elapse'. Thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Sure, thanks!

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Please refer to this link to know the reason of using 'freerunmode', 'freeruncount' and the interrupt in Free Run mode, thanks!
https://github.com/Samsung/TizenRT/pull/4328#discussion_r450589393

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Please refer to this link to know the reason of using 'freerunmode', 'freeruncount' and the interrupt in Free Run mode, thanks!
https://github.com/Samsung/TizenRT/pull/4328#discussion_r450589393

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: The freeruncount is not a standalone new mode, it is used to count the total time elapse in freerunmode. I will make changes and add description on it.

Please refer to this link to know the reason of using 'freerunmode', 'freeruncount' and the interrupt in Free Run mode, thanks!
https://github.com/Samsung/TizenRT/pull/4328#discussion_r450589393

os/arch/arm/src/amebad/amebad_timer_lowerhalf.c
AdamZhang0124: Please refer to this link to know the reason of using 'freerunmode', 'freeruncount' and the interrupt in Free Run mode, thanks!
https://github.com/Samsung/TizenRT/pull/4328#discussion_r450589393

PR #4331, opened 2020-06-26
drivers/timer : Modify timer interrupt notification method

jeongchanKim:
For notifying the timer interrupt, FIN(Fast Interrupt Notification) can be used, even it is faster than Signal.

### DIFF
--- a/apps/examples/timer/timer_main.c
+++ b/apps/examples/timer/timer_main.c
@@ -60,13 +60,13 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <fcntl.h>
-#include <signal.h>
 #include <errno.h>
 #include <semaphore.h>
 #include <pthread.h>
 
 #include <tinyara/timer.h>
 #include <tinyara/clock.h>
+#include <tinyara/irq.h>
 
 /****************************************************************************
  * Pre-processor Definitions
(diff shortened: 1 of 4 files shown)


apps/examples/timer/timer_main.c
```diff
@@ -123,12 +124,6 @@ static pthread_addr_t timer_thread(pthread_addr_t arg)
 	uint32_t expected_time;
 #endif
 
-	sigset_t sig_set;
```
sunghan-chang, line 126: Is `#include <signal.h>` necessary without this?

os/drivers/timer.c
```diff
@@ -147,20 +147,10 @@ static const struct file_operations g_timerops = {
 static bool timer_notifier(FAR uint32_t *next_interval_us, FAR void *arg)
 {
 	FAR struct timer_upperhalf_s *upper = (FAR struct timer_upperhalf_s *)arg;
-#ifdef CONFIG_CAN_PASS_STRUCTS
```
sunghan-chang, line 150: same as previous

sunghan-chang:
@jeongchanKim You should remove `signo` member from timer structure, and should remove the code adding signo from example.

sunghan-chang:
@jeongchanKim Please check howtousertimer.md as well

PR #4337, opened 2020-06-26
network/library: fix getifaddrs for Iotivity

pillip8282:
getifaddrs is tempolary implementation to run Iotivity. it doesn't run
on TizenRT 3.0 loadable environment because it accesses to
g_netdevices which is defined in kernel layer to get a running NIC
name. So this commit adds new ioctl command(SIOCGIFNAME) to access interface name
from app layer. however this operation is not same to linux. so you'd
better to use it carefully.

### DIFF
--- a/lib/libc/netdb/lib_getifaddr.c
+++ b/lib/libc/netdb/lib_getifaddr.c
@@ -1,0 +1,101 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/ioctl.h>
+#include <errno.h>
+#include <netdb.h>
+#include <net/if.h>
+#include <ifaddrs.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getifaddrs
+ *
+ * Description:
+ *   Return active NIC information whereas linux implementation
+ *   returns all NIC information. it's temporary API for supporting Iotivity
+ *   It should be fixed later when requirements is changed.
+ *
+ * Input Parameters:
+ *   ifap - The list  consists of ifaddrs structures
+ *
+ * Returned Value:
+ *   0 on success, non-zero on failure
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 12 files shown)


lib/libc/netdb/lib_getifaddr.c
```diff
@@ -0,0 +1,101 @@
...
+	int sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd < 0) {
+		printf("socket() failed with errno: %d\n", errno);
+		return -1;
+	}
+
+	struct ifreq req;
+	memset(&req, 0, sizeof(req));
+
+	int ret = ioctl(sockfd, SIOCGIFNAME, (unsigned long)&req);
```
sangwon03, line 74: Is it correct to use request(req) without request type and options?

lib/libc/netdb/lib_getifaddr.c
```diff
@@ -0,0 +1,101 @@
...
+	ifa.ifa_next = NULL;
+	ifa.ifa_name = ifname;
+	ifa.ifa_flags = flags | IFF_RUNNING;
+	addr.sin_family = netmask.sin_family = AF_INET;
+	ifa.ifa_addr = (struct sockaddr *)&addr;
+	ifa.ifa_netmask = (struct sockaddr *)&netmask;
+
+	*ifap = &ifa;
+
+	return 0;
```
sangwon03, line 100: why don't you use 'ret' instead of '0' to know the result of ioctl

os/net/netdev/netdev_ifname.c
```diff
@@ -0,0 +1,83 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Function: netdev_getifaddrs
```
sangwon03, line 57: Please check function and parameters name.

lib/libc/netdb/lib_getifaddr.c
```diff
@@ -0,0 +1,101 @@
...
+#include <netdb.h>
+#include <net/if.h>
+#include <ifaddrs.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getifaddr
```
sangwon03, line 37: getifaddr --> getifaddrs

lib/libc/netdb/lib_getifaddr.c
```diff
@@ -0,0 +1,101 @@
...
+ * Input Parameters:
+ *   ifap - The list  consists of ifaddrs structures
+ *
+ * Returned Value:
+ *   0 on success, non-zero on failure
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: getifaddr
```
sangwon03, line 53: getifaddr --> getifaddrs

os/net/netmgr/netdev_mgr_internal.c
```diff
@@ -232,10 +233,12 @@ int nm_ifdown(struct netdev *dev)
...
-	if (ret < 0) {
-		ndbg("fail to down driver interface\n");
-		return -2;
+	if (dev->type == NM_WIFI) {
+		ret = dev->t_ops.wl->deinit(dev);
+		if (ret < 0) {
+			ndbg("fail to down driver interface\n");
+			return -2;
+		}
 	}
```
sangwon03, line 242: why don't you add logic about NM_ETHERNET deinit because nm_ifup() handle NM_ETHERNET initialization.

lib/libc/netdb/lib_getifaddr.c
pillip8282: it's correct.

PR #4341, opened 2020-06-30
libc/stdio: support hh, h, int, long, long long modifier in sscanf

sunghan-chang:
Currenlty it does not support. When the type of getting result
is less than int, sscanf removes outsid of result buffer.
for example,
char ret[10] = "efghi";
char buf[10] = "1:2:3";

sscanf(buf, "%hhx:%hhx:%hhx", &ret[0], &ret[1], &ret[2]);

Then, ret should be "123hi", but because of not support hh modifier
it becomes "123".

This is partially back-ported from nuttx commit 350295d.

### DIFF
--- a/lib/libc/stdio/lib_sscanf.c
+++ b/lib/libc/stdio/lib_sscanf.c
@@ -80,6 +80,12 @@
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 #endif
 
+#define HH_MOD -2
+#define H_MOD  -1
+#define NO_MOD  0
+#define L_MOD   1
+#define LL_MOD  2
+
 /****************************************************************************
  * Private Type Declarations
  ****************************************************************************/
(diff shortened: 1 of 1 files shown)


lib/libc/stdio/lib_sscanf.c
```diff
@@ -830,19 +909,38 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
...
+						pchar = va_arg(ap, unsigned char *);
+						*pchar = (unsigned char)nchars;
+						break;
+					case H_MOD:
+						pshort = va_arg(ap, unsigned short *);
+						*pshort = (unsigned short)nchars;
+						break;
+					case NO_MOD:
+						pint = va_arg(ap, unsigned int *);
+						*pint = (unsigned int)nchars;;
```
jeongchanKim, line 923: Please remove unnecessary `;`.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -340,13 +346,20 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
 	FAR const char *bufstart;
 	FAR char *tv;
 	FAR const char *tc;
-	bool lflag;
+	int modifier;
 	bool noassign;
 	int count;
 	int width;
 	int fwidth;
 	int base = 10;
```
jsdosa, line 354: in my opinion, It seems better to add a definition for 10.

sunghan-chang: This is not related with this PR. Let's do that later with new PR.

jsdosa: ok, i see.

PR #4343, opened 2020-06-30
os/fs/smartfs: Add function to handle Rename more efficiently

amogh-samsung:
- Add function smartfs_replaceentry which takes
  care of rename/open operations in smartfs.
- Case where source and taregt path for a file
  move/copy are the same treated separately.
- Use an invalidated entry to overwrite new entry
  before allocating new sector for entry.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -178,6 +178,9 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	uint16_t parentdirsector;
 	const char *filename;
 	struct smartfs_ofile_s *sf;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	struct smartfs_entry_s invalidentry;
+#endif
 
 	/* Sanity checks */
 
@@ -278,10 +281,28 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		if (parentdirsector != 0xFFFF) {
 			/* We can create in the given parent directory */
 
-			ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
-			if (ret != OK) {
-				goto errout_with_buffer;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			ret = smartfs_find_firstinvalidentry(fs, parentdirsector, &invalidentry);
+			if (ret == OK) {
+				invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+				ret = smartfs_replaceentry(fs, invalidentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, sf);
+				if (ret != OK) {
+					goto errout_with_buffer;
+				}
+			} else {
+				if (ret != -ENOENT) {
+					goto errout_with_buffer;
+				} else
+#endif
+				{
+					ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
+					if (ret != OK) {
+						goto errout_with_buffer;
+					}
+				}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 			}
+#endif
 		} else {
 			/* Trying to create in a directory that doesn't exist */
 
@@ -1656,6 +1677,9 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 	struct smartfs_entry_s entry;
 	uint16_t parentdirsector;
 	const char *filename;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	struct smartfs_entry_s invalidentry;
+#endif
 
 	/* Sanity checks */
 
@@ -1696,13 +1720,29 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 
 		/* Create the directory */
 
-		ret = smartfs_createentry(fs, parentdirsector, filename,
-								  SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
-		if (ret != OK) {
-			goto errout_with_semaphore;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		ret = smartfs_find_firstinvalidentry(fs, parentdirsector, &invalidentry);
+		if (ret == OK) {
+			invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+			ret = smartfs_replaceentry(fs, invalidentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else {
+			if (ret != -ENOENT) {
+				goto errout_with_semaphore;
+			} else
+#endif
+			{
+				ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
+				if (ret != OK) {
+					goto errout_with_semaphore;
+				}
+				ret = OK;
+			}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 		}
-
-		ret = OK;
+#endif
 	}
 
 errout_with_semaphore:
@@ -1819,7 +1859,9 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	uint16_t entrysize;
 	struct smartfs_entry_header_s *direntry;
 	struct smart_read_write_s readwrite;
-
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	struct smartfs_entry_s invalidentry;
+#endif
 	uint8_t *tmp_pntr = NULL;
 	uint16_t tmp_flag = 0;
 
@@ -1922,46 +1964,72 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		mode = oldentry.flags & SMARTFS_DIRENT_MODE;
 		type = oldentry.flags & SMARTFS_DIRENT_TYPE;
 
-		ret = smartfs_createentry(fs, newparentdirsector, newfilename, type, mode, &newentry, oldentry.firstsector, NULL);
-		if (ret != OK) {
-			goto errout_with_semaphore;
-		}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		if (oldparentdirsector == newparentdirsector) {
+			ret = smartfs_replaceentry(fs, oldentry, newfilename, type, mode, &newentry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else {
+			ret = smartfs_find_firstinvalidentry(fs, newparentdirsector, &invalidentry);
+			if (ret == OK) {
+				invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+				ret = smartfs_replaceentry(fs, invalidentry, newfilename, SMARTFS_DIRENT_TYPE_FILE, mode, &newentry, NULL);
+				if (ret != OK) {
+					goto errout_with_semaphore;
+				}
+			} else {
+				if (ret != -ENOENT) {
+					goto errout_with_semaphore;
+				} else
+#endif
+				{
+					ret = smartfs_createentry(fs, newparentdirsector, newfilename, type, mode, &newentry, oldentry.firstsector, NULL);
+					if (ret != OK) {
+						goto errout_with_semaphore;
+					}
+					/* Now mark the old entry as inactive */
+					readwrite.logsector = oldentry.dsector;
+					readwrite.offset = 0;
+					readwrite.count = fs->fs_llformat.availbytes;
+					readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
 
-		/* Now mark the old entry as inactive */
+					ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+					if (ret < 0) {
+						fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
+						goto errout_with_semaphore;
+					}
 
-		readwrite.logsector = oldentry.dsector;
-		readwrite.offset = 0;
-		readwrite.count = fs->fs_llformat.availbytes;
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-		if (ret < 0) {
-			fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
-			goto errout_with_semaphore;
-		}
-		// bug fix for rename
-		tmp_pntr = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
-		tmp_flag = tmp_pntr[0];
-		tmp_flag |= (tmp_pntr[1] << 8);
+					// bug fix for rename
+					tmp_pntr = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
+					tmp_flag = tmp_pntr[0];
+					tmp_flag |= (tmp_pntr[1] << 8);
 
 #if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-		tmp_flag &= ~SMARTFS_DIRENT_ACTIVE;
+					tmp_flag &= ~SMARTFS_DIRENT_ACTIVE;
 #else
-		tmp_flag |= SMARTFS_DIRENT_ACTIVE;
+					tmp_flag |= SMARTFS_DIRENT_ACTIVE;
 #endif
-		tmp_pntr[0] = (uint8_t)(tmp_flag & 0x00FF);
-		tmp_pntr[1] = (uint8_t)((tmp_flag >> 8) & 0x00FF);
+					tmp_pntr[0] = (uint8_t)(tmp_flag & 0x00FF);
+					tmp_pntr[1] = (uint8_t)((tmp_flag >> 8) & 0x00FF);
 
-		/* Now write the updated flags back to the device */
+					/* Now write the updated flags back to the device */
 
-		readwrite.offset = oldentry.doffset;
-		readwrite.count = sizeof(uint16_t);
-		readwrite.buffer = (uint8_t *)tmp_pntr;
+					readwrite.offset = oldentry.doffset;
+					readwrite.count = sizeof(uint16_t);
+					readwrite.buffer = (uint8_t *)tmp_pntr;
 
-		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-		if (ret < 0) {
-			fdbg("Error %d writing flag bytes for sector %d\n", ret, readwrite.logsector);
-			goto errout_with_semaphore;
+					ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
+					if (ret < 0) {
+						fdbg("Error %d writing flag bytes for sector %d\n", ret, readwrite.logsector);
+						goto errout_with_semaphore;
+					}
+				}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			}
 		}
+#endif
+
 	} else {
 		/* Trying to create in a directory that doesn't exist */
 
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -86,6 +86,7 @@
 #define SET7BITSFROMMSB 254		//Binary: 0b11111110
 #define SET8BITSFROMMSB 255		//Binary: 0b11111111
 #define NEG8BIT(A) ((uint8_t)(~A & 0x000000FF))
+
 #endif
 /****************************************************************************
  * Private Types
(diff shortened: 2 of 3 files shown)


os/fs/smartfs/smartfs_smart.c
```diff
@@ -1795,10 +1795,64 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
...
+		if (newrelpath[j] == '\\') {
+			return 0;
+		}
+		newname[j - lastsubdir - 1] = newrelpath[j];
+		j++;
+	}
+	newname[j - lastsubdir - 1] = newrelpath[j];
+
+	//Check that old and new file names are not same in case of rename request - invalid request
+	if (strcmp(oldname, newname) == 0) {
```
jeongchanKim, line 1845: If you can, please use `strncmp` instead of `strcmp`.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1922,45 +1977,55 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		mode = oldentry.flags & SMARTFS_DIRENT_MODE;
 		type = oldentry.flags & SMARTFS_DIRENT_TYPE;
 
-		ret = smartfs_createentry(fs, newparentdirsector, newfilename, type, mode, &newentry, oldentry.firstsector, NULL);
-		if (ret != OK) {
+		rename_chk = smartfs_check_rename_only(oldrelpath, newrelpath);
+		if (rename_chk == -EINVAL) {
 			goto errout_with_semaphore;
-		}
+		} else if (rename_chk == 1) {
```
jeongchanKim, line 1983: What is the meaning of `1`? Could you use pre-defined value for readability?

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1795,10 +1795,64 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
...
+ *              directory path.
+ *
+ ***************************************************************************/
+
+int smartfs_check_rename_only(const char *oldrelpath, const char *newrelpath)
+{
+	int i = 0;
+	int j;
+	int lastsubdir = -1;
+	char oldname[33];
```
jsdosa, line 1812: in my opinion, It seems better to add a definition for 33.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1795,10 +1795,64 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
...
+	char newname[33];
```
jsdosa, line 1813: same as above.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,127 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+		if (ret < 0) {
+			goto errout;
+		}
+
+		nextsector = (uint16_t)ret;
+
+		/* Set the newly allocated sector's type (file or dir) */
+
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		if (sf) {
```
Taejun-Kwon, line 1318: Actually this is problem.
For open case we hold SF buffer.
This will update when it call write.
so we write entry in parent sector, so entry in parent sector points to invalidate sector(physically nothing written)
so before write, if power off happened, entry points empty sector. This will lead to read error.
so I added code to remove entry if child is not valid.. see smartfs_examine_sector, you can see code from below annotation.
```
/* If entry item is not valid, then remove from entry */
```
But actually this step bring some write operation, so if we also can hold this new entry sector in buffer, then everything is fine.
Please think about this carefully.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,127 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	readwrite.offset = oldentry.doffset;
+	readwrite.count = entrysize;
+	readwrite.buffer = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
+	ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long) &readwrite);
+	if (ret < 0) {
+		fdbg("failed to write new entry to parent directory psector : %d\n", oldentry.dsector);
+		goto errout;
+	}
+
+	/* Now fill in the entry */
```
Taejun-Kwon, line 1363: please check this very carefully.
It assume that only rename case.
If it is open, direntry->dsector should be newly allocated sector in above .

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1795,10 +1795,64 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
...
+/***************************************************************************
+ * Name: smartfs_check_rename_only
+ *
+ * Description: Check if the operation requested from smartfs_rename only
+ *              involves renaming an existing file with a new name at the
+ *              directory path.
+ *
+ ***************************************************************************/
+
+int smartfs_check_rename_only(const char *oldrelpath, const char *newrelpath)
```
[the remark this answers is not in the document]
amogh-samsung: Ohkayy, yes, agreed.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1795,10 +1795,64 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
...
+	}
+
+	return 1;
+}
+
 /****************************************************************************
  * Name: smartfs_rename
  *
- * Description: Rename a file or directory
+ * Description: Rename/Move/Copy a file or directory
```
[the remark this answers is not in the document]
amogh-samsung: Ohkay, done

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,127 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	readwrite.count = fs->fs_llformat.availbytes;
+	readwrite.offset = 0;
+	readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
+	ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+	if (ret < 0) {
+		fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
+		goto errout;
+	}
+
+	if (oldentry.firstsector == 0xFFFF) {
```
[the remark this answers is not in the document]
amogh-samsung: Changes for smartfs_open to use new logic are yet to be implemented.
PR still in progress.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -178,6 +178,9 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	uint16_t parentdirsector;
 	const char *filename;
 	struct smartfs_ofile_s *sf;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	struct smartfs_entry_s invalidentry;
```
Taejun-Kwon, line 182: Actually target number of sector where invalid entry saved, and offset is enough to use.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -277,10 +280,34 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 
 		if (parentdirsector != 0xFFFF) {
 			/* We can create in the given parent directory */
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			if (smartfs_findinvalidentry(fs, parentdirsector, &invalidentry)) {
 
-			ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
-			if (ret != OK) {
-				goto errout_with_buffer;
+				invalidentry.firstsector = 0xFFFF;
```
Taejun-Kwon, line 286: It always has invalid firstsector, because entry is already invalidated.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -277,10 +280,34 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
```
Taejun-Kwon, line 287: We will memset offset of invalid entry and put some info to there, am I right?
Then below is not necessary, IMOP, eh?

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Using `smartfs_entry_s invalidsector` in smartfs_open() allows using smartfs_replaceentry() function along the same lines as for smartfs_rename() without the need for adding two additional parameters for sector no. and offset.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: The flags for the old invalidated entry will be different from what we need for the new entry.

So I think new flags need to be prepared so that they can be written to the entry.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Got it.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1922,45 +1950,60 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+			}
+		} else if (smartfs_findinvalidentry(fs, newparentdirsector, &invalidentry)) {
+			ret = smartfs_replaceentry(fs, invalidentry, newfilename, type, &newentry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else
+#endif
+		{
+			ret = smartfs_createentry(fs, newparentdirsector, newfilename, type, mode, &newentry, oldentry.firstsector, NULL);
```
Taejun-Kwon, line 1967: inside of smartfs_createentry, there's code under CONFIG_SMARTFS_USE_SECTOR_BUFFER.
It is useless after we decide use replaceentry instead of createentry.
Please clean it. (At least put TODO comment there)

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: My comment was we'd better do this in replaceentry, but you want here?
Then Please add some comment. 
For example.
/* Now we found target invalidated entry. so update it before replaced */

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,206 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+/****************************************************************************
+ * Name: smartfs_findvalidentry
+ * 
+ * Description: Finds an invalid entry in the parent directory that can be
+ *              used to create a new file entry without allocating a new
+ *              sector.
+ *
+ ****************************************************************************/
+
+bool smartfs_findinvalidentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, struct smartfs_entry_s *invalidentry)
```
Taejun-Kwon, line 1281: Please change bool to int and please see comment below.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,206 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	while (psector != SMARTFS_ERASEDSTATE_16BIT) {
+		/* Read the next sector */
+
+		readwrite.logsector = psector;
+		readwrite.count = fs->fs_llformat.availbytes;
+		readwrite.offset = 0;
+		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
+		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+		if (ret < 0) {
+			return FALSE;
```
Taejun-Kwon, line 1304: This is file operation,  VFS will save errno. please see function open in fs_open.
You can change it easily by refer to smartfs_finddirentry.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,206 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+		offset = sizeof(struct smartfs_chain_header_s);
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
+
+		while ((offset + entrysize) < readwrite.count) {
+			/* Check if this entry is invalid */
+
+			if (!(ENTRY_VALID(entry))) {
+				invalidentry->dsector = psector;
+				invalidentry->doffset = offset;
+				return TRUE;
```
Taejun-Kwon, line 1321: So, return OK here.

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
Taejun-Kwon: hmm it is actually for mkdir & open, right? please add some comment here.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Understood, thanks

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Agreed

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Agreed

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Agreed

os/fs/smartfs/smartfs_smart.c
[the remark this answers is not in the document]
amogh-samsung: Let's use smartfs_find_firstinvalidentry()

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1268,206 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
 	return ret;
 }
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+/****************************************************************************
+ * Name: smartfs_findvalidentry
+ * 
+ * Description: Finds an invalid entry in the parent directory that can be
```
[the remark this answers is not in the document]
amogh-samsung: Ohkay, let's use smartfs_find_firstinvalidentry()

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Yes, agreed, thanks.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1696,13 +1716,25 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 
 		/* Create the directory */
 
-		ret = smartfs_createentry(fs, parentdirsector, filename,
-								  SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
-		if (ret != OK) {
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		ret = smartfs_find_firstinvalidentry(fs, parentdirsector, &invalidentry);
+		if (ret == OK) {
```
Taejun-Kwon, line 1721: Let's clean here.
First we'd better clear specific error first such like a this situation.
```
if (ret != OK)  {
    if (ret != -ENOENT) {
          goto error_with_semaphore;
    } else {
         
    }
}

File: os/fs/smartfs/smartfs_smart.c [hunk 19]
```diff
@@ -1696,13 +1716,25 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
+			invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+			ret = smartfs_replaceentry(fs, invalidentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else if (ret < 0 && ret != -ENOENT) {
```
Taejun-Kwon, line 1727: -ENOENT should not come to as a result.
1. Find first invalid entry.
2. If there's only valid entry, return first empty entry. (we can call createentry there?)
So I don't think it returns -ENOENT ...

Taejun-Kwon, line 1727: fs always return OK or errno, so below is correct.
```
if (ret == ok) {
   blah_blah;
} else {
   if (ret != -ENOENT) {
      bah bah!
   } else {
      lol lol!
   }
}

File: os/fs/smartfs/smartfs_smart.c [hunk 20]
```diff
@@ -1922,46 +1956,70 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
-			goto errout_with_semaphore;
-		}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		if (oldparentdirsector == newparentdirsector) {
+			ret = smartfs_replaceentry(fs, oldentry, newfilename, type, mode, &newentry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else {
+			ret = smartfs_find_firstinvalidentry(fs, newparentdirsector, &invalidentry);
```
Taejun-Kwon, line 1966: Actually I don't like this name, you are good at english, could you please suggest nice name something means that 'overwrite' or .something? (Not this PR, consider it when you have some time)

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1922,46 +1956,70 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+			if (ret == OK) {
+				invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+				ret = smartfs_replaceentry(fs, invalidentry, newfilename, SMARTFS_DIRENT_TYPE_FILE, mode, &newentry, NULL);
+				if (ret != OK) {
+					goto errout_with_semaphore;
+				}
+			} else if (ret < 0 && ret != -ENOENT) {
```
Taejun-Kwon, line 1973: please see above comment regarding same logic.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1283,224 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+		nextsector = SMARTFS_NEXTSECTOR(chainheader);
+
+		/* Search for an invalid entry in this sector */
+		offset = sizeof(struct smartfs_chain_header_s);
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
+
+		while ((offset + entrysize) < readwrite.count) {
+			/* Check if this entry is invalid */
+
+			if (entry->flags != SMARTFS_ERASEDSTATE_16BIT && !(ENTRY_VALID(entry))) {
```
Taejun-Kwon, line 1337: If entry is invalidated, we don't have to care about another flag in entry->flags, I think.
Because If it is invalid entry, we don't use it as a entry anymore.

Taejun-Kwon:
Main point of our new system is that create new entry first -> and then chain or update in parent sector.
So that we can recovery it if chaining or updating was not happened.
please change state to ready.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkkay, I'll give this a thoight.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Done

amogh-samsung: Function find_firstinvalidentry() will return -ENOENT when there is no invalidated entry in any of the chained sectors.

However, now the logic has been improved. This function also checks for free entries along the way so if no invalidated entry is found, it can return the first free entry found.
Still if no free or invalidated entry is found it will return -ENOENT.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I added `entry->flags != SMARTFS_ERASEDSTATE_16BIT` to counter the case when find_firstinvalidentry() would come across an erased sector and consider an entry with flags = ERASEDSTATE is invalid and return that for replaceentry.
This works, but it is not logically correct.

However, now with updated logic, find_fistinvalidentry will save the first entry found with flags = ERASEDSTATE and return that as a free entry.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1696,13 +1716,29 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		ret = smartfs_find_firstinvalidentry(fs, parentdirsector, &invalidentry);
+		if (ret == OK) {
+			invalidentry.firstsector = SMARTFS_ERASEDSTATE_16BIT;
+			ret = smartfs_replaceentry(fs, invalidentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, NULL);
+			if (ret != OK) {
+				goto errout_with_semaphore;
+			}
+		} else {
+			if (ret != -ENOENT) {
```
Taejun-Kwon, line 1728: line 292 also can be applied this logic.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1295,241 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+				invalidentry->dsector = psector;
+				invalidentry->doffset = offset;
+				ret = OK;
+				goto errout;
+			}
+
+			/* If we find an empty entry in the process and no invalidated entry is found
+			   in the end, we can return the empty entry we had found.
+			 */
+			if (saved_doffset == -1 && IS_EMPTY_ENTRY(entry)) {
```
Taejun-Kwon, line 1361: where saved_doffset defined? I couldn't find. is it saved_offset?

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: Then,
```
if (!ENTRY_VALID(entry)) {
    if (IS_EMPTY_ENTRY(entry)) {
        saved_psector = psector;
	saved_doffset = offset;
    } else {
        invalidentry->dsector = psector;
	invalidentry->doffset = offset;
	ret = OK;
	goto errout;
    }
}
```
This doesn't work? can you please test?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1153,19 +1193,7 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
-	entry->flags = (uint16_t)(SMARTFS_DIRENT_ACTIVE | SMARTFS_DIRENT_DELETING | SMARTFS_DIRENT_RESERVED | type | (mode & SMARTFS_DIRENT_MODE));
-#endif
-#else							/* CONFIG_SMARTFS_ERASEDSTATE == 0xFF */
-#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
-	smartfs_wrle16(&entry->flags, (uint16_t)(SMARTFS_DIRENT_EMPTY | type | (mode & SMARTFS_DIRENT_MODE)));
-#else
-	entry->flags = (uint16_t)(SMARTFS_DIRENT_EMPTY | type | (mode & SMARTFS_DIRENT_MODE));
-#endif
-#endif							/* CONFIG_SMARTFS_ERASEDSTATE == 0xFF */
+	WRITE_ENTRY_FLAGS(entry->flags, mode, type);
```
Taejun-Kwon, line 1196: How do you think change WRITE -> SET ?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1295,241 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
+	psector = oldentry.dsector;
+	offset = oldentry.doffset;
+
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+	if (smartfs_rdle16(&oldentry.flags) == SMARTFS_ERASEDSTATE_16BIT) {
+#else
+	if (oldentry.flags == SMARTFS_ERASEDSTATE_16BIT) {
+#endif
+		WRITE_ENTRY_FLAGS(oldentry.flags, mode, type);
```
Taejun-Kwon, line 1417: please remove one tab

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1268,6 +1295,241 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+
+	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
+	psector = oldentry.dsector;
+	offset = oldentry.doffset;
+
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+	if (smartfs_rdle16(&oldentry.flags) == SMARTFS_ERASEDSTATE_16BIT) {
+#else
+	if (oldentry.flags == SMARTFS_ERASEDSTATE_16BIT) {
+#endif
```
Taejun-Kwon, line 1416: Why you don't you make MACRO for this? like a wrle16?

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Corrected

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Ohkkay, no major difference.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Indentation is alright I think.
"WRITE_ENTRY_FLAGS" is inside an if construct.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I dont think that is necessary.
Also this kind of check for flags in ERASEDSTATE is performed only once.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -278,10 +281,28 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+					goto errout_with_buffer;
+				} else
+#endif
+				{
+					ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
+					if (ret != OK) {
+						goto errout_with_buffer;
+					}
+				}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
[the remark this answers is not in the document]
Taejun-Kwon: anyway please refer to pseudo code in confluence. this need to be refactored.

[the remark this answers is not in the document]
amogh-samsung: Corrected for now.
Refactoring pending

amogh-samsung:
Please refer to PR #4376 for changes to smartfs file entry operations.

PR #4347, opened 2020-07-02
kernel/semaphore : Remove the holder from the holder list in sem_releaseholder

jeongarmy:
The holder should be removed from the holder list and freed.

### DIFF
--- a/os/kernel/semaphore/sem_holder.c
+++ b/os/kernel/semaphore/sem_holder.c
@@ -908,11 +908,8 @@ void sem_releaseholder(FAR sem_t *sem, FAR struct tcb_s *rtcb)
 		 */
 		pholder->counts--;
 
-#ifndef CONFIG_PRIORITY_INHERITANCE
-		if (pholder->counts <= 0) {
-			sem_freeholder(sem, pholder);
-		}
-#endif
+		/* Remove the holder from the holder list */
+		sem_freeholder(sem, pholder);
 	}
 }
 


os/kernel/semaphore/sem_holder.c
```diff
@@ -908,11 +908,8 @@ void sem_releaseholder(FAR sem_t *sem, FAR struct tcb_s *rtcb)
 		 */
 		pholder->counts--;
 
-#ifndef CONFIG_PRIORITY_INHERITANCE
-		if (pholder->counts <= 0) {
-			sem_freeholder(sem, pholder);
-		}
-#endif
+		/* Remove the holder from the holder list */
```
sangam-swami, line 911: Can we check if the holder thread is active or not with :  sched_verifytcb(htcb) 
If the thread is not active then we can sem_freeholder in else case?

os/kernel/semaphore/sem_holder.c
```diff
@@ -908,11 +908,8 @@ void sem_releaseholder(FAR sem_t *sem, FAR struct tcb_s *rtcb)
...
+		sem_freeholder(sem, pholder);
```
jeongchanKim, line 912: I think we need
`#if !defined(CONFIG_PRIORITY_INHERITANCE) && defined(CONFIG_BINMGR_RECOVERY)`
and we should check the pholder counts.
Because pholder counts can be more than 0 when same task tries to wait the sem in normal case.

jeongarmy:
@jeongchanKim will raise new commit about this change.

PR #4349, opened 2020-07-02
os/fs/smartfs : Apply checkin / checkout logic

Taejun-Kwon:
1. Add checkin logic
2. Add checkout logic
3. Add block write logic that is use both of checkin / checkout logic
4. Add dump for journal, it will be displayed after journal_init

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -406,26 +406,43 @@ typedef uint32_t crc_t;
 #define AREA_OLD_UNERASED 0xFF
 #endif
 
-/* State of Journal Log Low 2 bits */
-#define SMART_JOURNAL_STATE_CHECKIN  0x01
-#define SMART_JOURNAL_STATE_CHECKOUT 0x02
+/* State of Journal Log Low 4 bits */
+#define SMART_JOURNAL_STATE_CHECKIN  (1 << 1)
+#define SMART_JOURNAL_STATE_CHECKOUT (1 << 2)
 
-/* Type of Journal Data next 6 bits Low 2bits for journal area */
-#define SMART_JOURNAL_TYPE_USED      0x04
-#define SMART_JOURNAL_TYPE_UNERASED  0x08
-#define SMART_JOURNAL_TYPE_ALLOC     0x10
-#define SMART_JOURNAL_TYPE_COMMIT    0x20
-#define SMART_JOURNAL_TYPE_RELEASE   0x40
-#define SMART_JOURNAL_TYPE_ERASE     0x80
+/* Type of Journal Data high 4 bits */
+#define SMART_JOURNAL_TYPE_AREA      1		/* Write state bits to toggle to Next Area */
+#define SMART_JOURNAL_TYPE_ALLOC     2		/* If we need to MTD header first which we allocated */
+#define SMART_JOURNAL_TYPE_COMMIT    3		/* Common Type, When write data on entire of sector */
+#define SMART_JOURNAL_TYPE_RELEASE   4		/* Release Sector */
+#define SMART_JOURNAL_TYPE_ERASE     5		/* MTD Erase (c.f smart_erase_block_if_empty..) */
+
+/* MACRO For Journal Type */
+#define SET_JOURNAL_TYPE(t, v) ((t) = ((t) & 0x0f) | ((v) << 4))
+#define GET_JOURNAL_TYPE(t) ((t) >> 4)
+
+/* MACRO For Journal Status, Consider erase state, it should be called before set type & state */
+#define JOURNAL_STATUS_RESET(t) ((t) = ((t) | (CONFIG_SMARTFS_ERASEDSTATE >> 4)))
+
+#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
+#define SET_JOURNAL_STATE(t, s) ((t) = ((t) & ~(s)))
+#else
+#define SET_JOURNAL_STATE(t, s) ((t) = ((t) | (s)))
+#endif
+
+#define CHECK_JOURNAL_CHECKIN(t)   (((t) & SMART_JOURNAL_STATE_CHECKIN) != (SMART_JOURNAL_STATE_CHECKIN & CONFIG_SMARTFS_ERASEDSTATE))
+#define CHECK_JOURNAL_CHECKOUT(t)  (((t) & SMART_JOURNAL_STATE_CHECKOUT) != (SMART_JOURNAL_STATE_CHECKOUT & CONFIG_SMARTFS_ERASEDSTATE))
+
 
 struct smart_journal_logging_data_s {
 	struct smart_sect_header_s mtd_header;	/* MTD Header to checkpoint */
 	uint8_t crc16[2];						/* CRC-16 for current journal data */
-	uint8_t psector[2];						/* Target Physical Sector */
+	uint8_t psector[2];						/* Target Physical Sector, For MTD_ERASE, Target Physical Block */
 	uint8_t seq[2];							/* Sequence number of Journal for validation check */
 	uint8_t status;							/* Journal Staus low 4bits for state, high 4bits for type */
 };
 
+typedef struct smart_journal_logging_data_s journal_log_t;
 #endif
 
 /****************************************************************************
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+		dev->journal_seq++;
+	}
+	return OK;
+}
+
+/****************************************************************************
+ * Name: smart_journal_checkout
+ *
+ * Description:
+ *   Check in journal data
```
jeongchanKim, line 5241: Could you check the description? The function name is about checkout.

os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+		return ret;
+	}
+
+	if (UINT8TOUINT16(buf.crc16) != smart_calc_journal_crc(&buf)) {
+		fdbg("Checkin Error, verify crc failed buf crc16 : %d calc crc : %d\n", UINT8TOUINT16(buf.crc16), smart_calc_journal_crc(&buf));
+		return ret;
+	}
+
+	/* Increase sequence here, journal_seq is always less than total number of journal data */
+	if (dev->journal_seq == dev->njournaldata -1) {
```
jeongchanKim, line 5229: Please add space after `-`.

os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+	/* Update MTD Header on target physical sector First, But Erase doesn't do that */
+	if (type != SMART_JOURNAL_TYPE_ERASE) {
+		ret = MTD_WRITE(dev->mtd, psector * dev->sectorsize, mtd_size, (FAR uint8_t *)&log->mtd_header);
+		if (ret != mtd_size) {
+			fdbg("Checkout error, write mtd header.. ret : %d\n", ret);
+			return ret;
+		}
+	}
+	
+	switch(type) {
```
jeongchanKim, line 5263: Please add space after `switch`.

os/fs/driver/mtd/smart.c
```diff
@@ -5166,6 +5408,92 @@ static int smart_journal_erase(FAR struct smart_struct_s *dev, uint16_t physsect
...
+				fdbg("read failed! ret : %d addr : %u\n", addr);
+				break;
+			}
+			
+			if (smart_journal_check_log_empty(&log)) {
+				fdbg("end of journal\n");
+				break;
+			}
+
+			fdbg("Journal data #%d\n", count);
```
jeongchanKim, line 5475: Did you use fdbg instead of fvdbg intentionally?

os/fs/driver/mtd/smart.c
```diff
@@ -5166,6 +5408,92 @@ static int smart_journal_erase(FAR struct smart_struct_s *dev, uint16_t physsect
...
+
+	/* TODO All dump log will be hide Later, Now enable it to develop journaling */
+	addr = area0;
+
+	while (addr < last) {
+		if (addr == area0 || addr == area1) {
+			uint8_t id_bits;
+			ret = MTD_READ(dev->mtd, addr, 1, &id_bits);
+			if (ret < 0) {
+				fdbg("read failed! ret : %d addr : %u\n", addr);
```
jeongchanKim, line 5458: There are two identifier, but there is only one `addr`. Maybe you missed `ret`.

os/fs/driver/mtd/smart.c
```diff
@@ -5166,6 +5408,92 @@ static int smart_journal_erase(FAR struct smart_struct_s *dev, uint16_t physsect
...
+				break;
+			}
+			fdbg("area #%d status : 0x%x\n", (count == 0)? 0 : 1, id_bits);
+			addr++;
+		} else {
+			ret = MTD_READ(dev->mtd, addr, sizeof(journal_log_t), (FAR uint8_t *)&log);
+			if (ret < 0) {
+				fdbg("read failed! ret : %d addr : %u\n", addr);
```
jeongchanKim, line 5466: Same as above.

os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+{
+	uint16_t psector;
+	size_t seq;
+	uint32_t *offset = address;
+
+	psector = dev->neraseblocks * dev->sectorsPerBlk;
+
+	/* Get Physical sector & sequence of each area First */
+	if (dev->area == 0) {
+		if (dev->journal_seq >= dev->njournaldata / 2) {
```
davidfather, line 5175: It might be better to use dev->njournaldata >> 1 instead of divide operation.

os/fs/driver/mtd/smart.c
sangam-swami, line 5475: can we use different debug log for error and info....fdbg for error and fvdbg for info?

Taejun-Kwon: See line 5450. I will keep this and will change or disable it when product level is almost ended.

os/fs/driver/mtd/smart.c
Taejun-Kwon: Thanks. I really need this kind of review because of my personality..

os/fs/driver/mtd/smart.c
Taejun-Kwon: Good point, I will apply it everywhere, thanks.

os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+	/* Verify written data */
+	ret = MTD_READ(dev->mtd, address, log_size, (FAR uint8_t *)&buf);
+	if (ret != log_size) {
+		fdbg("Checkin Error, read log data... physical address %u ret : %d\n", address, ret);
+		return ret;
+	}
+
+	if (UINT8TOUINT16(buf.crc16) != smart_calc_journal_crc(&buf)) {
+		fdbg("Checkin Error, verify crc failed buf crc16 : %d calc crc : %d\n", UINT8TOUINT16(buf.crc16), smart_calc_journal_crc(&buf));
+		return ret;
```
davidfather, line 5225: This error type should be discriminated from the error, which might happen at MTD_READ().

os/fs/driver/mtd/smart.c
```diff
@@ -5109,13 +5147,169 @@ static int smart_joutnal_set_area_state(FAR struct smart_struct_s *dev, int area
...
+		if (ret != mtd_size) {
+			fdbg("Checkout error, write mtd header.. ret : %d\n", ret);
+			return ret;
+		}
+	}
+	
+	switch(type) {
+		/* If type is commit, we will check validation of entire sector */
+		case SMART_JOURNAL_TYPE_COMMIT: 
+			ret = MTD_BREAD(dev->mtd, psector * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)dev->rwbuffer);
```
davidfather, line 5266: What does MTD_BREAD mean?

os/fs/driver/mtd/smart.c
Taejun-Kwon: Thanks. I will apply it

os/fs/driver/mtd/smart.c
Taejun-Kwon: It is macro to execute mtd_dev_s->bread it read entire of logical sector (physically n * page).
To check crc, we have to read all data in current sector.

os/fs/driver/mtd/smart.c
Taejun-Kwon: see above comment. This should be kept to debug journaling.

Taejun-Kwon:
> Overall, I feel like there are somewhat inconsistent return type usages through this file. For example, sometimes return type seems boolean, but sometimes it seems the byte size of read, write operations.

I agree with you it need some more clean but please understand it is newly added code.
But most case follows fs style, I think you feel like that because you are not used do it.
FS should return size of operation or errno. So If any function uses MTD operation it should return integer.
But checkin / checkout is util function for journal type , result of mtd operation is different.
That's why I use OK
Only smart_journal_check_log_empty returns boolean.
I always try to keep this rule.

Taejun-Kwon:
Target board is little bit slow when erase, so change logic regading journal block.
Hence, close it.

PR #4350, opened 2020-07-03
kernel/binary_manager_load : Modify sem restore logic for fault case

jeongchanKim:
If a task which occupied the kernel sem in fault binary, we should remove the sem holder from holder list.
So before calling the releaseholder, set the holder counts as 1.
And after releasing the kernel sem, there can be a task which waits that sem.
So call the sem_post instead of just calling sem_releaseholder.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -556,7 +556,9 @@ void binary_manager_release_binary_sem(int bin_idx)
 			if (holder && holder->htcb && holder->htcb->group && holder->htcb->group->tg_binidx == bin_idx) {
 				/* Increase semcount and release itself from holder */
 				sem->semcount++;
-				sem_releaseholder(sem, holder->htcb);
+				sem_freeholder(sem, holder);
+				/* And after releasing the kernel sem, there can be a task which waits that sem. So unblock the waiting task. */
+				sem_unblock_task(sem, holder->htcb);
 			}
 		}
 		sem = sq_next(sem);
--- a/os/kernel/semaphore/sem_holder.c
+++ b/os/kernel/semaphore/sem_holder.c
@@ -190,7 +190,7 @@ static inline FAR struct semholder_s *sem_findorallocateholder(sem_t *sem, FAR s
  * Name: sem_freeholder
  ****************************************************************************/
 
-static inline void sem_freeholder(sem_t *sem, FAR struct semholder_s *pholder)
+void sem_freeholder(sem_t *sem, FAR struct semholder_s *pholder)
 {
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
 	FAR struct semholder_s *curr;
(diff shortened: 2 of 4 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -555,8 +555,10 @@ void binary_manager_release_binary_sem(int bin_idx)
 		{
 			if (holder && holder->htcb && holder->htcb->group && holder->htcb->group->tg_binidx == bin_idx) {
 				/* Increase semcount and release itself from holder */
+				holder->counts = 1;
 				sem->semcount++;
 				sem_releaseholder(sem, holder->htcb);
```
jeongarmy, line 560: It is duplicated because sem_releaseholder finds tcb of holders by searching a holder list and then free holder.
So how about calling sem_freeholder(need to change it to public function) without line558?

os/kernel/semaphore/sem_post.c
```diff
@@ -93,6 +93,61 @@
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
+void sem_unblock_to_waiting(sem_t *sem, struct tcb_s *htcb)
```
jeongarmy, line 96: how about sem_unblock_waiting_task?

os/kernel/semaphore/semaphore.h
```diff
@@ -89,6 +89,12 @@ extern "C" {
 #define EXTERN extern
 #endif
 
+struct sem_restore_s {
```
jsdosa, line 92: could you add comments?

os/kernel/semaphore/sem_holder.c
```diff
@@ -563,9 +564,10 @@ static int sem_restoreholderprioA(FAR struct semholder_s *pholder, FAR sem_t *se
 
 static int sem_restoreholderprioB(FAR struct semholder_s *pholder, FAR sem_t *sem, FAR void *arg)
 {
-	FAR struct tcb_s *rtcb = this_task();
-	if (pholder->htcb == rtcb) {
-		(void)sem_restoreholderprio(pholder, sem, arg);
+	sem_restore_t *restore_tcbs = (sem_restore_t *)arg;
+	FAR struct tcb_s *htcb = restore_tcbs->htcb;
+	if (pholder->htcb == htcb) {
+		(void)sem_restoreholderprio(pholder, sem, restore_tcbs->stcb);
```
sangam-swami, line 570: any reason to make it void here and returning 1 here...

os/kernel/semaphore/semaphore.h
jeongchanKim: Added. Thanks.

os/kernel/semaphore/sem_holder.c
jeongchanKim: sem_restoreholderprio always return 0, so we don't use the return of it.
And 1 also has not special reason I think.
But it is not related to this PR. I will check again and remove/change if needed with different commit. Thanks :)

os/kernel/semaphore/sem_holder.c
```diff
@@ -662,10 +664,12 @@ static inline void sem_restorebaseprio_irq(FAR struct tcb_s *stcb, FAR sem_t *se
  *
  ****************************************************************************/
 
-static inline void sem_restorebaseprio_task(FAR struct tcb_s *stcb, FAR sem_t *sem)
+static inline void sem_restorebaseprio_task(FAR struct tcb_s *stcb, FAR struct tcb_s *htcb, FAR sem_t *sem)
 {
-	FAR struct tcb_s *rtcb = this_task();
 	FAR struct semholder_s *pholder;
+	sem_restore_t restore_tcbs;
```
sijupunnnoose, line 670: How about using sem_restore_t restore_tcbs = {stcb, htcb}; ??

PR #4352, opened 2020-07-06
apps/examples/tc_pthread : Modify wrong return check in pthread join tc

jeongchanKim:
1. Pthread_join in line 641 tries to wait/join the already terminated pthread.
In this case, pthread_join returns ESRCH.
2. Add EINVAL case TC. EINVAL will be returned when passing the task's pid in pthread_join.

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_pthread.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_pthread.c
@@ -611,6 +611,16 @@ static void tc_pthread_pthread_barrier_init_destroy_wait(void)
 	TC_SUCCESS_RESULT();
 }
 
+static int infinite_sleep(int argc, char *argv[])
+{
+	while (1) {
+		/* This task will be terminated by task_delete from main tc task. */
+		sleep(10);
+	}
+
+	return 0;
+}
+
 /**
 * @fn                   :tc_pthread_pthread_create_exit_join
 * @brief                :creates a new thread with a specified attributes and \
@@ -626,6 +636,7 @@ static void tc_pthread_pthread_create_exit_join(void)
 {
 	int ret_chk;
 	pthread_t pthread;
+	int task_pid;
 	void *p_value = 0;
 
 	ret_chk = pthread_create(&pthread, NULL, self_pthread_join_n_exit, NULL);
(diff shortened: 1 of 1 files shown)


apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -635,12 +647,22 @@ static void tc_pthread_pthread_create_exit_join(void)
...
 
 	ret_chk = pthread_join(INVALID_PID, NULL);
 	TC_ASSERT_EQ("pthread_join", ret_chk, ESRCH);
 
+	task_pid = task_create("inval_task", 100, 1024, inval_task, NULL);
+	TC_ASSERT_GT("task_create", task_pid, 0);
+
+	ret_chk = pthread_join(task_pid, NULL);
+	TC_ASSERT_EQ("pthread_join", ret_chk, EINVAL);
+	inval_task_alive = false;
```
sunghan-chang, line 664: How about using task_delete instead of controlling the task lifecycle using static variable? I think it is better to understand.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -611,6 +612,16 @@ static void tc_pthread_pthread_barrier_init_destroy_wait(void)
 	TC_SUCCESS_RESULT();
 }
 
+static int inval_task(int argc, char *argv[])
```
sunghan-chang, line 615: Let's name it to represent the operation.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -626,6 +636,7 @@ static void tc_pthread_pthread_create_exit_join(void)
 {
 	int ret_chk;
 	pthread_t pthread;
+	int task_pid;
```
pillip8282, line 639: how about to initialize the task_pid?

jeongchanKim: Thanks for review. But I think we don't need to initialize this variable.
Because before using this variable, it always set the value in line 658.

PR #4354, opened 2020-07-06
Remove User heap and protected build

kishore-sn:
This PR contains following changes:
- Remove user ram and user heap from TizenRT
- Fix issues in kernel after uheap removal
- Remove protected build feature from TizenRT
- Modify all config files to remove protected build and uheap dependencies
- Modify kernel and arch specific code to remove protected build and uheap dependencies


os/arch/arm/src/amebad/amebad_allocateheap.c
```diff
@@ -148,36 +109,38 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
  *
  ****************************************************************************/
 
-#if defined(CONFIG_BUILD_PROTECTED) && defined(CONFIG_MM_KERNEL_HEAP)
 void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
 {
+	*heap_start = (FAR void *)(g_idle_topstack & ~(0x7));
+
+	/* There may be a special scenraio where we might configure a different region
```
jeongchanKim, line 116: Please change `scenraio` to `scenario`.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -105,7 +105,7 @@ static void allocateregions(FAR struct elf_loadinfo_s *loadinfo)
 		}
 	}
 
-	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kumm_memalign(sizes[0], totalsize);
+	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
```
jeongchanKim, line 108: It means we will allocate the user binary's section into the kernel side. Is it okay?

os/arch/arm/src/stm32l4/stm32l4_allocateheap.c
```diff
@@ -335,90 +98,39 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
...
-		usize = (1 << log2);
-		ubase = SRAM1_END - usize;
-
-    /* Return the kernel heap settings (i.e., the part of the heap region
-     * that was not dedicated to the user heap).
-     */
-
-		*heap_start = (FAR void *)USERSPACE->us_bssend;
-		*heap_size  = ubase - (uintptr_t)USERSPACE->us_bssend;
+	*heap_start = (FAR void *)(g_idle_topstack & ~(0x7));
```
jeongchanKim, line 103: Originally, why we remain this is we want to use each vendor's heap allocation code.
If we want to use our generic heap allocation code, we can use up_allocateheap instead of this.
Is it possible now also?

os/tools/mkconfig.c
```diff
@@ -205,21 +205,15 @@ int main(int argc, char **argv, char **envp)
...
-	printf("#else\n");
-	printf("#define CONFIG_MM_REGION_NUM CONFIG_MM_REGIONS\n");
-	printf("#endif\n");
+	printf("#endif\n\n");
+	printf("#ifdef __KERNEL__\n");
+	printf("# define CONFIG_MM_REGIONS CONFIG_KMM_REGIONS\n");
 	printf("#else\n");
-	printf("#define CONFIG_MM_REGION_NUM CONFIG_MM_REGIONS\n");
-	printf("#endif\n");
+	printf("# define CONFIG_MM_REGIONS 1\n");
```
jeongchanKim, line 214: You removed the `CONFIG_MM_REGIONS` and `CONFIG_MM_NHEAPS`. Are these needed?

sunghan-chang:
@kishore-sn Could you check build breaks?

os/arch/arm/src/amebad/amebad_allocateheap.c
kishore-sn: Done

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -197,17 +121,26 @@ void up_allocate_heap(FAR void **heap_start, size_t *heap_size)
...
 void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
 {
 	/* Return the kernel heap settings (i.e., the part of the heap region
 	 * that was not dedicated to the user heap).
 	 */
 
 	*heap_start = (FAR void *)(g_idle_topstack & ~(0x7));
-	*heap_size = (uint32_t)((uintptr_t)__usram_segment_start__) - (uint32_t)(*heap_start);
+
+	/* There may be a special scenraio where we might configure a different region
```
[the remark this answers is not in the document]
kishore-sn: Done

os/arch/arm/src/stm32l4/stm32l4_allocateheap.c
kishore-sn: These changes were make to simplify the heap alloaction.
Any vendor specific setting for ram usage can be pre-calculated and set in the RAM_KREGIONx config.
In this case, we can use the common up_alocateheap instead of using the chip specific code. 
In a similar way, it is possible to move mpuinit.c, blocktask.c, unblocktask.c and other files to common folder. 
If we agree upon this, I will make the required changes in a new PR.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Yes. We will allocate user sections from kernel heap, and then we will use the mpu to provide permissions as required for each section.

jeongchanKim: In this case, kernel developer should know how big the user binary is. I think it is not easy.

kishore-sn: In any case there needs to be a balance between the kernel and user areas, and it always depends on the size of the user apps. If we allocate app sections from user heap, then we need to know before hand, the size of user heap. Even in this case, the developer needs to split the ram between user and kernel based on the app size. 
I think this method is better because, in this case we dont have to predetermine the size of the user heap. If the app size is small enough, then there is no problem from kernel side. On the other hand, if the app size becomes too big and takes up too much kernel heap, then having the app allocation from a separate user heap would also cause the same problem.

os/tools/mkconfig.c
kishore-sn: Fixed

PR #4357, opened 2020-07-06
kernel/semaphore: Register user semaphores to release all holders of them

jeongarmy:
In fault recovery, it is necessary to release all holders of it despite user sempahore.
That's because all holder data is in kernel region.

### DIFF
--- a/os/include/sys/syscall.h
+++ b/os/include/sys/syscall.h
@@ -106,12 +106,13 @@
 #define SYS_sem_timedwait              (CONFIG_SYS_RESERVED + 16)
 #define SYS_sem_trywait                (CONFIG_SYS_RESERVED + 17)
 #define SYS_sem_wait                   (CONFIG_SYS_RESERVED + 18)
+#define SYS_sem_init                   (CONFIG_SYS_RESERVED + 19)
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
-#define SYS_sem_setprotocol            (CONFIG_SYS_RESERVED + 19)
-#define __SYS_named_sem                (CONFIG_SYS_RESERVED + 20)
+#define SYS_sem_setprotocol            (CONFIG_SYS_RESERVED + 20)
+#define __SYS_named_sem                (CONFIG_SYS_RESERVED + 21)
 #else
-#define __SYS_named_sem                (CONFIG_SYS_RESERVED + 19)
+#define __SYS_named_sem                (CONFIG_SYS_RESERVED + 20)
 #endif
 
 /* Named semaphores */
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -534,7 +534,7 @@ static int loading_thread(int argc, char *argv[])
  * Name: binary_manager_release_binary_sem
  *
  * Description:
- *	 This function releases all kernel semaphores held by the threads in binary.
+ *	 This function releases all semaphores held by the threads in binary.
  *
  ****************************************************************************/
 void binary_manager_release_binary_sem(int bin_idx)
@@ -545,6 +545,7 @@ void binary_manager_release_binary_sem(int bin_idx)
 
 	flags = irqsave();
 
+	/* Release kernel semaphores held by the threads and all holders of them */
 	sem = (sem_t *)sq_peek(&g_sem_list);
 	while (sem) {
 #if CONFIG_SEM_PREALLOCHOLDERS > 0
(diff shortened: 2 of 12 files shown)


os/syscall/syscall_lookup.h
```diff
@@ -87,6 +87,10 @@ SYSCALL_LOOKUP(sem_wait,                  1, STUB_sem_wait)
 SYSCALL_LOOKUP(sem_setprotocol,           2, STUB_sem_setprotocol)
 #endif
 
+#ifdef CONFIG_BINMGR_RECOVERY
+SYSCALL_LOOKUP(sem_register,              1, STUB_sem_register)
```
sunghan-chang, line 91: 1. How about moving sem_init to kernel inside and registering it as syscall instead of sem_register?
  Because sem_init is the posix api so that it looks ok to register.
2. Could you split this as two commits? One is to move sem api to kernel api or to register syscall api, another is getting sem list in binary manager.

os/include/sys/syscall.h
```diff
@@ -109,9 +109,16 @@
 
 #ifdef CONFIG_PRIORITY_INHERITANCE
 #define SYS_sem_setprotocol            (CONFIG_SYS_RESERVED + 19)
-#define __SYS_named_sem                (CONFIG_SYS_RESERVED + 20)
+#define __SYS_sem_register             (CONFIG_SYS_RESERVED + 20)
```
jeongchanKim, line 112: I think we'd better register the syscall not `sem_register` but `sem_init`.

lib/libc/semaphore/sem_init.c
```diff
@@ -135,10 +132,10 @@ int sem_init(FAR sem_t *sem, int pshared, unsigned int value)
 		}
 #endif
 
-#if defined(CONFIG_BINMGR_RECOVERY) && defined(__KERNEL__)
+#if defined(CONFIG_BINMGR_RECOVERY)
 		/* Register semaphore in kernel region for kernel resource management */
```
jeongchanKim, line 136: Please modify the comment.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -563,6 +564,25 @@ void binary_manager_release_binary_sem(int bin_idx)
...
+#endif
+		{
+			if (holder) {
+				/* Free a holder of semaphore */
+				sem_freeholder(sem, holder);
+			}
+		}
+		sem = sq_next(sem);
+	}
+	sq_init(&BIN_SEMLIST(bin_idx));
```
jeongchanKim, line 584: You only initialize the sem list in here. Is it not needed to initialize before using?

os/kernel/semaphore/sem_list.c
```diff
@@ -43,11 +47,30 @@ sq_queue_t g_sem_list;
...
  ****************************************************************************/
+static void sem_addlist(FAR sem_t *sem, sq_queue_t *list)
+{
+	sem_t *sem_ptr;
+	irqstate_t flags;
+
+	flags = irqsave();
+
+	sem_ptr = (sem_t *)sq_peek(list);
+	while (sem_ptr) {
```
jeongchanKim, line 58: I'm worried about this code. Because sometimes this list can be very long, then it takes so much time.
How about using another way to check that sem is registered or not, like hash?

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -563,6 +564,25 @@ void binary_manager_release_binary_sem(int bin_idx)
 		}
 		sem = sq_next(sem);
 	}
+
+	/* Release all holders for user sempahores */
+	sem = (sem_t *)sq_peek(&BIN_SEMLIST(bin_idx));
```
jeongchanKim, line 569: We should consider the shared library case also. If sem is in shared library, we should check we need context switching or not.

os/syscall/syscall_stublookup.c
```diff
@@ -105,6 +105,7 @@ uintptr_t STUB_sem_timedwait(int nbr, uintptr_t parm1, uintptr_t parm2);
 uintptr_t STUB_sem_trywait(int nbr, uintptr_t parm1);
 uintptr_t STUB_sem_unlink(int nbr, uintptr_t parm1);
 uintptr_t STUB_sem_wait(int nbr, uintptr_t parm1);
+uintptr_t STUB_sem_register(int nbr, uintptr_t parm1);
```
jeongchanKim, line 108: Is it okay not to enclose the ifdef?

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: It is ok because of global variable but I will update.

jeongarmy:
I will raise another PR later.

PR #4359, opened 2020-07-07
kernel/binary_manager : Change binary load-failure log to lldbg

jeongchanKim:
It is good to show always about binary loading failure log.
So change bmdbg to lldbg.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -84,7 +84,7 @@ static int binary_manager_load_binary(int bin_idx, char *path, load_attr_t *load
 			usleep(1000);
 		}
 		retry_count++;
-		bmdbg("Load '%s' %dth fail, errno %d\n", BIN_NAME(bin_idx), retry_count, errno);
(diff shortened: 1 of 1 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -84,7 +84,7 @@ static int binary_manager_load_binary(int bin_idx, char *path, load_attr_t *load
 			usleep(1000);
 		}
 		retry_count++;
-		bmdbg("Load '%s' %dth fail, errno %d\n", BIN_NAME(bin_idx), retry_count, errno);
+		lldbg("Load '%s' %dth fail, errno %d\n", BIN_NAME(bin_idx), retry_count, errno);
```
sunghan-chang, line 87: lldbg does not print the message always. It has dependencies with CONFIG_DEBUG_ERROR and CONFIG_ARCH_LOWPUTC. Is this what you intend?

jeongchanKim: Yes. I think it is enough. If we change this to printf, it can be a problem when making real product.

PR #4361, opened 2020-07-07
os/fs/smartfs : Add smart_validate_journal_crc to verify crc of journal

Taejun-Kwon:
1. add smart_validate_journal_crc
2. remove useless definitions

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -395,17 +395,6 @@ typedef uint32_t crc_t;
 #error "CRC-16 is necessary for journaling"
 #endif
 
-/* State of first byte of each journal Area */
-#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-#define AREA_UNUSED 0xFF
-#define AREA_USED   0xF0
-#define AREA_OLD_UNERASED 0x00
-#else
-#define AREA_UNUSED 0x00
-#define AREA_USED   0x0F
-#define AREA_OLD_UNERASED 0xFF
-#endif
-
 /* State of Journal Log Low 4 bits */
 #define SMART_JOURNAL_STATE_CHECKIN  (1 << 1)
 #define SMART_JOURNAL_STATE_CHECKOUT (1 << 2)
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5408,16 +5425,19 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
...
 	fvdbg("Printf recovery data journal_seq : %d!!\n", dev->journal_seq);
 	smart_journal_print_log(dev, log);
 
-	crc = smart_calc_journal_crc(log);
-
-	if (crc != UINT8TOUINT16(log->crc16)) {
-		fdbg("Invalid CRC %d calculated crc : %d written crc : %d\n", crc, log->crc16);
+	/* Recovery step is Check crc -> Check something more based on type -> checkout -> verify based on type */
+	if (smart_validate_journal_crc(log) != OK) {
+		fdbg("Invalid CRC %d calculated crc : %d written crc : %d\n", smart_calc_journal_crc(log), log->crc16);
```
jsdosa, line 5435: could you check arguments?

jsdosa: i think, it need the three arguments.

Taejun-Kwon: thanks. fixed. something more.

os/fs/driver/mtd/smart.c
```diff
@@ -5408,16 +5425,19 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
...
+		/* Invalid one, so make it checkout */
+		ret = smart_journal_checkout(dev, log, address);
+		if (ret != OK) {
+			return ret;
```
jsdosa, line 5439: This will never return OK.

Taejun-Kwon: No if checkout finished properly, ret is OK but crc is not valid so finally it return -EINVAL

jsdosa: If the ret value is not OK, the ret value is always -EINVAL.

os/fs/driver/mtd/smart.c
```diff
@@ -2709,6 +2700,32 @@ static crc_t smart_calc_sector_crc(FAR struct smart_struct_s *dev)
 #endif
 
 #ifdef CONFIG_MTD_SMART_JOURNALING
+/****************************************************************************
+ * Name: smart_validate_crc
```
amandeepc, line 2704: *smart_validate_journal_crc

os/fs/driver/mtd/smart.c
Taejun-Kwon: It should return -EINVAL because it is invalid journal, it means that power failed during writing journal.
So it should checkout to disable target journal, but inside of smart_journal_checkout I mean old one, there were crc check logic so it alwasys return -EIO.
So I divided smart_journal_checkout.

os/fs/driver/mtd/smart.c
Taejun-Kwon: Thanks. fixed.

PR #4365, opened 2020-07-08
os/fs/smartfs: Change flow of smartfs_deleteentry

Taejun-Kwon:
1. To guarantee integrity of entry or chaining, we have to reorder flow.
2. For delete, update entry or chaining should be done at the first time.
3. If power off after change entry or chaining, there should be isolated sector
4. We will reuse fs_recover in master branch to handle this.

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -1290,48 +1290,15 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 	struct smartfs_entry_header_s *direntry;
 	struct smartfs_chain_header_s *header;
 	struct smart_read_write_s readwrite;
+	bool inactive_entry = TRUE;
+	/** Our policy is that integrity entry & chaining.
+	  * If Isolated sector created but entry or chain is safe, we can handle it through the fs_recover.
+	  * So We will always process regarding entry & chain first when delete entry.
+      */
 
-	/* Okay, delete the file.  Loop through each sector and release them
-
-	 * TODO:  We really should walk the list backward to avoid lost
-	 *        sectors in the event we lose power. However this requires
-	 *        allocating a buffer to build the sector list since we don't
-	 *        store a doubly-linked list of sectors on the device.  We
-	 *        could test if the sector data buffer is big enough and
-	 *        just use that, and only allocate a new buffer if the
-	 *        sector buffer isn't big enough.  Do do this, however, we
-	 *        need to change the code below as it is using the a few
-	 *        bytes of the buffer to read in header info.
-	 */
-
-	nextsector = entry->firstsector;
+	/* First Find current directory has only one item which is target entry */
+	ret = OK;
 	header = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
-	readwrite.offset = 0;
-	readwrite.count = sizeof(struct smartfs_chain_header_s);
-	readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-	while (nextsector != SMARTFS_ERASEDSTATE_16BIT) {
-		/* Read the next sector into our buffer */
-
-		sector = nextsector;
-		readwrite.logsector = sector;
-		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-		if (ret < 0) {
-			fdbg("Error reading sector %d\n", nextsector);
-			break;
-		}
-
-		/* Release this sector */
-
-		nextsector = SMARTFS_NEXTSECTOR(header);
-		ret = FS_IOCTL(fs, BIOC_FREESECT, sector);
-		if (ret < 0) {
-			fdbg("Error freeing sector %d\n", nextsector);
-			break;
-		}
-	}
-
-	/* Remove the entry from the directory tree */
-
 	readwrite.logsector = entry->dsector;
 	readwrite.offset = 0;
 	readwrite.count = fs->fs_llformat.availbytes;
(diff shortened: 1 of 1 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1452,11 +1388,78 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
...
+	while (nextsector != SMARTFS_ERASEDSTATE_16BIT) {
+		/* Read the next sector into our buffer */
 
-	ret = OK;
+		sector = nextsector;
+		readwrite.logsector = sector;
+		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+		if (ret < 0) {
+			fdbg("Error reading sector %d\n", nextsector);
+			break;
```
sangwon03, line 1450: why don't you use 'goto errorout' instead of 'break' to maintain consistency.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1452,11 +1388,78 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
...
+		ret = FS_IOCTL(fs, BIOC_FREESECT, sector);
+		if (ret < 0) {
+			fdbg("Error freeing sector %d\n", nextsector);
+			break;
+		}
+	}
+	
 errout:
 	return ret;
 }
```
sangwon03, line 1465: All functions are so long. Is it possible to separate the code for easy maintenance?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1434,9 +1371,8 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 						fdbg("Error unchaining sector (%d)\n", nextsector);
 						goto errout;
 					}
-
+
```
gcjjyy, line 1374: Remove tabs in the blank line.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1342,61 +1309,31 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
...
-
-		if (count == 0) {
+		/* Test if the count is one.  If it is, then we will release the sector */
+		if (count == 1) {
+			/* We don't have to inactive entry, we will release entire of sector */
+			inactive_entry = FALSE;
+			
 			/* Okay, to release the sector, we must find the sector that we
-			 * are chained to and remove ourselves from the chain.  First
+			 * are chained to and remove ourselves from the chain.	First
```
gcjjyy, line 1336: Change tab to space before the 'First' word.

Taejun-Kwon: Fixed, thanks.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: I already know, not only this file long functions are spread. 
There are 2 reason why I keep current state,
One is if there are something big size duplicated code, I will make it function. But nothing here. Then we'd better reduce jump time, we are rtos.
Another one is sometimes we brought bug fix or something useful from our Mommy's repository, so as possible as I can, I prefer to keep originality of code.
Enough?

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: Ok I agree with you.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: Yup!

os/fs/smartfs/smartfs_utils.c
sangwon03: ok. I agree

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1290,48 +1290,15 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 	struct smartfs_entry_header_s *direntry;
 	struct smartfs_chain_header_s *header;
 	struct smart_read_write_s readwrite;
+	bool inactive_entry = TRUE;
+	/** Our policy is that integrity entry & chaining.
+	  * If Isolated sector created but entry or chain is safe, we can handle it through the fs_recover.
+	  * So We will always process regarding entry & chain first when delete entry.
```
sangwon03, line 1296: Isolated -> isolated,   
So We -> So we

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1342,62 +1309,32 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
...
 			/* Advance to next entry */
-
 			offset += entrysize;
 		}
 
-		/* Test if the count it zero.  If it is, then we will release the sector */
-
-		if (count == 0) {
+		/* Test if the count is one.  If it is, then we will release the sector */
+		if (count == 1) {
```
amogh-samsung, line 1331: What if we don't unchain a sector if this is the last active entry that we are deleting?
Even if it has no valid entries, it can be used for allocation of future entries with smartfs_find_availableentry().

Taejun-Kwon: We're not sure next operation, if we remain it as active sector but consist of inactive entries, it will consume sectors.
I think it is not good approach, because our target is small nor flash.
If we handle big enough storage & we spend much of time for experiment regarding behavior of file operation, then your suggestion will be nice.

PR #4366, opened 2020-07-08
tidyup of sched_yield

sunghan-chang:
1. arch/arm: move up_schedyield to each architecture
2. arch/arm: add missing codes in schedyield.c 
3. kernel/sched: optimize sched_yield with CONFIG_SCHED_YIELD_OPTIMIZATION 
4. arch/arm, kernel/sched: move yield functionality into arch for optimization

### DIFF
--- a/os/arch/arm/src/armv7-m/up_schedyield.c
+++ b/os/arch/arm/src/armv7-m/up_schedyield.c
@@ -20,6 +20,22 @@
  * Included Files
  ****************************************************************************/
 
+#include <tinyara/config.h>
+
+#ifdef CONFIG_TASK_MONITOR
+#include <sys/types.h>
+#endif
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+#include <stdint.h>
+#endif
+#include <tinyara/sched.h>
+#ifdef CONFIG_ARMV7M_MPU
+#include <tinyara/mpu.h>
+#endif
+#ifdef CONFIG_TASK_SCHED_HISTORY
+#include <tinyara/debug/sysdbg.h>
+#endif
+
 #include "sched/sched.h"
 #include "up_internal.h"
 
@@ -30,6 +46,10 @@
 /****************************************************************************
  * Private Data
  ****************************************************************************/
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+extern uint32_t g_umm_app_id;
+#endif
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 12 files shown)


os/arch/arm/src/armv8-m/up_schedyield.c
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Description:
+ *   Called when current task is willing to
+ *   yield cpu resource to other ready-to-run task.
+ *   context switch would occur in below case:
+ *
+ *   1) The priority of the currently running task is either equal or more than other ready to run task
+ *
+ * Inputs:
+ *   tcb: The TCB of the task that has been reprioritized
```
jeongchanKim, line 73: Please check the description.

os/arch/arm/src/armv8-m/up_schedyield.c
```diff
@@ -0,0 +1,197 @@
+/****************************************************************************
+ *
+ * Copyright 2016 Samsung Electronics All Rights Reserved.
```
jeongchanKim, line 3: This file is also new. Please check the license year.

os/arch/arm/src/armv7-r/arm_schedyield.c
```diff
@@ -0,0 +1,174 @@
...
+ *   context switch would occur in below case:
+ *
+ *   1) The priority of the currently running task is either equal or more than other ready to run task
+ *
+ * Inputs:
+ *   tcb: The TCB of the task that has been reprioritized
+ *
+ ****************************************************************************/
+
+void up_schedyield(struct tcb_s *rtcb)
```
jeongchanKim, line 67: I'm not sure the param `rtcb` is needed.

sunghan-chang: Yeah you're right. That is unnecessary so that I removed somewhere but I missed here and header.

os/arch/arm/src/armv8-m/up_schedyield.c
sunghan-chang: This file was located in common and was moved here so that 2016 is right. The armv7-m and armv7-r is newly created.

os/kernel/sched/sched_yield.c
```diff
@@ -102,63 +104,13 @@ int sched_yield(void)
...
-			}
-
-			up_schedyield(rtcb);
-		}
-
-		irqrestore(saved_state);
-	}
-	return OK;
+	saved_state = irqsave();
+	ret = up_schedyield();
```
jeongchanKim, line 111: `up_schedyield()` doesn't have return type. But I think, for checking validation of tcb, you'd better make `up_schedyield()` to return int.

os/arch/arm/src/armv7-m/up_schedyield.c
```diff
@@ -0,0 +1,196 @@
...
+ ****************************************************************************/
+
+void up_schedyield(void)
+{
+	struct tcb_s *rtcb = this_task();
+	struct tcb_s *ntcb = rtcb->flink;
+
+	if (ntcb == NULL) {
+		sdbg("Fail to yield, there is no next tcb. Current task : %d\n", rtcb->pid);
+		return ERROR;
```
jeongchanKim, line 85: I remained the comment in the below, but anyway please check the return type of this function.

os/arch/arm/src/armv7-m/up_schedyield.c
```diff
@@ -0,0 +1,196 @@
...
+	}
```
jeongchanKim, line 86: And I'm not sure we need ERROR return. Based on your implementation, if there is no tcb next to the current, yield will be failed. It means current tcb still has the context. 
But if we return OK, even though we will not make context switching, current tcb still has the context.
For context, it is same situation regardless of the return. So I think it is good to return OK for all case with dbg msg.

sunghan-chang: agree. removed and return OK always.

PR #4367, opened 2020-07-09
os/fs/smartfs : Adjust formula to calculate number of journal block

Taejun-Kwon:
When wrtie 1 sector & erase by file operation, fs uses only 1sector
But journal need 3 journal data.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -1069,6 +1069,11 @@ static int smart_setsectorsize(FAR struct smart_struct_s *dev, uint16_t size)
 		dev->njournaleraseblocks++;
 	}
 
+	/** We can change specific sector's state to commit, release, erase. In this case we need 3 journals,
+	  * but it uses only 1 data sector. Hence Multiply it by 3, Ideally.
+	  */
+	dev->njournaleraseblocks *= 3;
+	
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -1069,6 +1069,9 @@ static int smart_setsectorsize(FAR struct smart_struct_s *dev, uint16_t size)
 		dev->njournaleraseblocks++;
 	}
 
+	/* Data Sector share 3 type of operation(Commit, Release, Erase) But Journal doesn't share, more storage needed */
```
jsdosa, line 1072: 3 types

os/fs/driver/mtd/smart.c
```diff
@@ -1069,6 +1069,11 @@ static int smart_setsectorsize(FAR struct smart_struct_s *dev, uint16_t size)
 		dev->njournaleraseblocks++;
 	}
 
+	/** We can change specific sector's state to commit, release, erase. In this case we need 3 journals,
+	  * but it uses only 1 data sector. Hence Multiply it by 3, Ideally.
+	  */
+	dev->njournaleraseblocks *= 3;
```
amogh-samsung, line 1075: We should definitely optimize this during testing.
Even the entire formula for calculating journal sectors, based on minimum, average and maximum journal entries.

PR #4369, opened 2020-07-09
kernel/binary_manager_load : Add Debug message for checking g_sem_list

jeongchanKim:
If g_sem_list is empty, it means there is no kernel semaphore which used by threads.
For checking this, added the debug message.


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -547,6 +547,9 @@ void binary_manager_release_binary_sem(int bin_idx)
 	flags = irqsave();
 
 	sem = (sem_t *)sq_peek(&g_sem_list);
+	if (sem == NULL) {
+		bmdbg("g_sem_list is empty.\n");
+	}
 	while (sem) {
```
sunghan-chang, line 553: How about adding `else` condition to prevent duplicated checking when it is null?

PR #4371, opened 2020-07-09
tc,os/drivers/watchdog: fix tc for invalid result

jsdosa:
1. The result of the test case for the wrong input value is incorrect.
2. The watchdog does not handle an exception.

### DIFF
--- a/os/drivers/watchdog.c
+++ b/os/drivers/watchdog.c
@@ -273,10 +273,14 @@ static int wdog_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	wdvdbg("cmd: %d arg: %ld\n", cmd, arg);
 	DEBUGASSERT(upper && lower);
 
-	/* Get exclusive access to the device structures */
+	if (cmd < 0) {
+		return -EINVAL;
+	}
 
+	/* Get exclusive access to the device structures */
 	ret = sem_wait(&upper->exclsem);
 	if (ret < 0) {
+		ret = -get_errno();
 		return ret;
 	}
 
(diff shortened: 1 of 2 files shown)


os/drivers/watchdog.c
```diff
@@ -274,6 +274,9 @@ static int wdog_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	DEBUGASSERT(upper && lower);
 
 	/* Get exclusive access to the device structures */
+	if (cmd < 0) {
```
sunghan-chang, line 277: This is not matched with comment. Please move the comment.

os/drivers/watchdog.c
```diff
@@ -274,6 +274,9 @@ static int wdog_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		return -EINVAL;
+	}
 
 	ret = sem_wait(&upper->exclsem);
 	if (ret < 0) {
```
jeongchanKim, line 282: I think we'd better return this failure case with errno, not ERROR.

os/drivers/watchdog.c
jsdosa: yes

os/drivers/watchdog.c
jsdosa: you right, i modify that.

os/drivers/watchdog.c
jsdosa: i did it

jsdosa:
i will recommit

PR #4373, opened 2020-07-09
[Realtek][RTL8721CSM]Add Watchdog Timer Peripheral

AdamZhang0124:
1. The watchdog peripheral is now supported.
2. It is tested by the watchdog testcase (tc_watchdog.c).
3. Systimer is used to get timeleft.
4. Both reset mode and interrupt mode are supported.
5. Both WDIOC_PAUSE and WDIOC_RESUME ioctl commands are supported.
6. Fixed the logic of changing timeout mode when watchdog is paused.
7. Fixed the timeout and timeleft calculation issue after watchdog is resumed.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
+++ b/os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
@@ -64,7 +64,7 @@
 
 #define WDT_STOP	0
 #define WDT_START	1
-#define ON_PAUSE	2
+#define WDT_PAUSE	2
 
 /****************************************************************************
  * Private Types
@@ -78,7 +78,7 @@ struct amebad_wdg_lowerhalf_s {
 	xcpt_t handler;
 	uint32_t timeout_ms;
 	uint32_t timeout_left;		/* Record remaining timeout for PAUSE and RESUME */
-	uint32_t wdt_status;		/* 0: WDT_STOP; 1: WDT_START; 2: ON_PAUSE */
+	uint32_t wdt_status;		/* 0: WDT_STOP; 1: WDT_START; 2: WDT_PAUSE */
 	uint32_t start_tick;		/* Use to calculate timeleft before timeout */
 	uint32_t on_pause_duration;	/* Record the duration of watchdog timer on pause */
 	bool int_mode;				/* True for timeout interrupt mode on */
@@ -138,9 +138,10 @@ static int amebad_wdg_start(FAR struct watchdog_lowerhalf_s *lower)
 		priv->wdt_status = WDT_START;
 		priv->start_tick = SYSTIMER_TickGet();	// Record time right after wdt start
 		priv->on_pause_duration = 0;	// Reset on_pause_duration
+		return OK;
 	}
 
-	return OK;
+	return -EBUSY;
 }
 
 /****************************************************************************
@@ -202,9 +203,10 @@ static int amebad_wdg_keepalive(FAR struct watchdog_lowerhalf_s *lower)
 		/* Update start time and reset on_pause_duration */
 		priv->start_tick = SYSTIMER_TickGet();
 		priv->on_pause_duration = 0;
+		return OK;
 	}
 
-	return OK;
+	return -EPERM;
 }
 
 /****************************************************************************
@@ -233,32 +235,31 @@ static int amebad_wdg_getstatus(FAR struct watchdog_lowerhalf_s *lower, FAR stru
 	/* Reture watchdog flag */
 	status->flags = WDFLAGS_RESET;
 
-	/* Return ACTIVE if watchdog timer is started */
-	if (priv->wdt_status == WDT_START) {
-		status->flags |= WDFLAGS_ACTIVE;
-	}
-
 	/* Return CAPTURE if interrupt mode is enabled */
 	if (priv->int_mode) {
 		status->flags |= WDFLAGS_CAPTURE;
 	}
 
-	/* Return WAIT if watchdog timer is on pause */
-	if (priv->wdt_status == ON_PAUSE) {
-		status->flags |= WDFLAGS_WAIT;
-	}
-
 	/* Reture timeout in ms */
 	status->timeout = priv->timeout_ms;
 
-	/* Reture the timeleft before watchdog timeout in ms */
-	if (priv->wdt_status == ON_PAUSE) {
+	/* Return WAIT if watchdog timer is on pause */
+	if (priv->wdt_status == WDT_PAUSE) {
+		status->flags |= WDFLAGS_WAIT;
 		status->timeleft = priv->timeout_left;
 	} else {
 		uint32_t time_elapse;
 		time_elapse = ((SYSTIMER_TickGet() - priv->start_tick - priv->on_pause_duration) * 31) / 1000;	// Get time elapse in ms
 		status->timeleft = status->timeout - time_elapse;
+
+		/* Return ACTIVE if watchdog timer is started */
+		if (priv->wdt_status == WDT_START) {
+			status->flags |= WDFLAGS_ACTIVE;
+		}
 	}
+
+
+
 
 	return OK;
 }
@@ -338,7 +339,7 @@ static xcpt_t amebad_wdg_capture(FAR struct watchdog_lowerhalf_s *lower, xcpt_t 
 	priv->handler = handler;
 
 	/* Set to default RESET_MODE first */
-	if (priv->wdt_status == WDT_START) {	// Skip the timeout_left calculation if the status is ON_PAUSE
+	if (priv->wdt_status == WDT_START) {	// Skip the timeout_left calculation if the status is WDT_PAUSE
 		priv->timeout_left = priv->timeout_ms - ((SYSTIMER_TickGet() - priv->start_tick - priv->on_pause_duration) * 31) / 1000;
 	}
 	watchdog_init(priv->timeout_left);	// Re-init watchdog with remaining timeout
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+
+	/* Change to INT_MODE if handler exists */
+	if (handler) {
+		watchdog_irq_init((void *)priv->handler, (uint32_t)priv);	// INT_MODE enabled
+		priv->int_mode = true;
+	}
+
+	/* Re-start watchdog only if the status is WDT_START */
+	if (priv->wdt_status == WDT_START) {
+		watchdog_start();
```
jeongchanKim, line 355: You mentioned `Re-start watchdog`. Is it okay to call `watchdog_start`, not `watchdog_refresh`?

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+static int amebad_wdg_getstatus(FAR struct watchdog_lowerhalf_s *lower, FAR struct watchdog_status_s *status)
+{
+	struct amebad_wdg_lowerhalf_s *priv = (struct amebad_wdg_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+	if (!priv) {
+		return -ENODEV;
+	}
+
+	/* Reture watchdog flag */
+	status->flags = WDFLAGS_RESET;
```
jeongchanKim, line 234: We'd better check that status is NULL or not. Sometimes it can be dereferencing.

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+	}
+	if (priv->wdt_status == WDT_STOP) {
+		/* Start watchdog timer */
+		watchdog_start();
+		priv->wdt_status = WDT_START;
+		priv->start_tick = SYSTIMER_TickGet();	// Record time right after wdt start
+		priv->on_pause_duration = 0;	// Reset on_pause_duration
+	}
+
+	return OK;
```
sunghan-chang, line 143: Even if this timer is running, return value will be OK. Let's do like below
```
if (... == WDT_STOP) {
   start();
   return OK;
}

return -EBUSY;
```

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+	if (priv->wdt_status == WDT_START) {
+		/* Petting the dog */
+		watchdog_refresh();
+
+		/* Update start time and reset on_pause_duration */
+		priv->start_tick = SYSTIMER_TickGet();
+		priv->on_pause_duration = 0;
+	}
+
+	return OK;
```
sunghan-chang, line 207: same as previous
```
if (... == START) {
 ...
 return OK;
}

return -EPERM;
```

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+
+	/*  Reset the watchdog timer with new timeout value */
+	priv->timeout_ms = timeout;
+	watchdog_init(priv->timeout_ms);
+	if (priv->int_mode) {
+		watchdog_irq_init((void *)priv->handler, (uint32_t)priv);
+	}
+
+	/* Re-start watchdog only if the status is WDT_START */
+	if (priv->wdt_status == WDT_START) {
```
sunghan-chang, line 297: Is this ok to re-set above without stopping in this case?

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+static int amebad_wdg_stop(FAR struct watchdog_lowerhalf_s *lower)
+{
+	struct amebad_wdg_lowerhalf_s *priv = (struct amebad_wdg_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+	if (!priv) {
+		return -ENODEV;
+	}
+
+	/* Stop watchdog timer */
+	watchdog_stop();
```
jsdosa, line 169: could you check started?
```
if (isStarted) {
   watchdog_stop()
   return OK;
}

return other errocode;

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 1 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:355]: When using '**watchdog_init** then **watchdog_start**', the watchdog will be reset with new timeout value (here using timeout_left). It is the same function as refresh.

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 2 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:234]: Sure, do you mean if (!status), skip everything? In this case, what return value should I provide? Thanks!

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 3 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:143]: Sure, thanks!

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 4 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:207]: Sure, thanks!

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 5 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:297]: Yes, every time using **watchdog_init**, the watchdog register will be reset, including the watchdog enable bit that is used by watchdog_stop. 
Also, if '**init**' without a '**start**' (no matter if watchdog is already started before), the watchdog is not working.  

In this case, if watchdog is started, '**init**' + '**start**'; If watchdog is on pause, '**init**' only, and '**start**' is only done by WDIOC_RESUME; If watchdog is stopped, '**init**' only, and '**start**' is only done by **amebad_wdg_start**.

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c (hunk 6 above)
Comment (AdamZhang0124) [reply os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c:169]: My idea about this stop function is that it can be used at any moment.
Your suggested changes will ignore the wdt status of **ON_PAUSE**.

Although, it can be changed to `if(priv->wdt_status != WDT_STOP)` if it is necessary to feedback errorcode when multiple stop functions are called from upper layer.

File: os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c [hunk 7]
```diff
@@ -0,0 +1,482 @@
...
+ *
+ * Returned Values:
+ *   Zero on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+static int amebad_wdg_start(FAR struct watchdog_lowerhalf_s *lower)
+{
+	struct amebad_wdg_lowerhalf_s *priv = (struct amebad_wdg_lowerhalf_s *)lower;
+	DEBUGASSERT(priv);
+	if (!priv) {
```
Taejun-Kwon, line 132: 1. ASSERT is already there.
2. if parameter is not valid, we return -EINVAL

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+#include <tinyara/kmalloc.h>
+#include <tinyara/watchdog.h>
+#include <tinyara/irq.h>
+#include "wdt_api.h"
+
+#ifdef CONFIG_WATCHDOG
+
+#define WDT_STOP	0
+#define WDT_START	1
+#define ON_PAUSE	2
```
Taejun-Kwon, line 67: WDT_PAUSE is better..

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,482 @@
...
+	DEBUGASSERT(priv);
+	if (!priv) {
+		return -ENODEV;
+	}
+
+	/* Reture watchdog flag */
+	status->flags = WDFLAGS_RESET;
+
+	/* Return ACTIVE if watchdog timer is started */
+	if (priv->wdt_status == WDT_START) {
```
Taejun-Kwon, line 237:
```
if (priv->wdt_status == ON_PAUSE) {
     status->timeleft = priv->timeout_left;
     status->flag |= WDFLAG_WAIT;
} else {
     ...
     status->timeout = priv->timeout_ms;
     status->timeleft = status->timeout - timeelapse;
     if (priv->wdt_status == WDT_START) {
         status->flag |= WDFLAG_ACTIVE;
     }
}

if (priv->intmode). ..

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
AdamZhang0124: 1. This is to prevent system crash when CONFIG_DEBUG is disabled and priv is null.
2. For returning -ENODEV, I actually refereed to your previous comment on https://github.com/Samsung/TizenRT/pull/4328#discussion_r450198673.
Please let me know if they need to be changed, thanks!

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
AdamZhang0124: Sure, that will be improved. Thanks!

os/arch/arm/src/amebad/amebad_watchdog_lowerhalf.c
AdamZhang0124: The status->timeout also need to assign value in ON_PAUSE status actually.
I get what you mean, I will re-arrange that part.

PR #4375, opened 2020-07-10
Adjust init time of journal

Taejun-Kwon:
After journal we should update some counts in smartfs_dev, for example dev->releasecount
To calculate this, we have to finish smart_scan. So modify to initialize journal
before smart_scan execute.
So we don't need extra Map structure for loser anymore.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -307,7 +307,6 @@ struct smart_struct_s {
 	uint16_t njournalPerBlk;		/* Total Number of Journal data per Erase block */
 	uint16_t njournaleraseblocks;	/* Total Number of Journal Erase block */
 	uint32_t njournaldata;			/* Total Number of Journal Data */
-	FAR uint16_t *sSeqLogMap;		/* Keep duplicated sector until journaling determine winner */
 #endif
 };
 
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5484,14 +5448,17 @@ static int smart_journal_recovery(FAR struct smart_struct_s *dev, journal_log_t
 
 	/* For Release & Erase, Nothing for now, Retry Checkout */
 	case SMART_JOURNAL_TYPE_RELEASE:
+		return smart_journal_checkout_process(dev, log, address);
 	case SMART_JOURNAL_TYPE_ERASE:
-		break;
+		ret = smart_journal_checkout_process(dev, log, address);
+		if (ret = OK) {
```
jeongchanKim, line 5454: Please modify to `ret == OK`.

os/fs/driver/mtd/smart.c
```diff
@@ -2232,24 +2210,10 @@ static int smart_scan(FAR struct smart_struct_s *dev)
...
-			goto err_out;
-		}
-	}
-#endif
 	ret = OK;
 
 err_out:
-#ifdef CONFIG_MTD_SMART_JOURNALING
-	if (dev->sSeqLogMap != NULL) {
-		kmm_free(dev->sSeqLogMap);
```
jeongchanKim, line 2250: There is a memory allocation for `rootdirdev`. But there is no free for failure case. Is it okay?

Taejun-Kwon: yes it should be kept. it is block driver's private data, all the time it should be kept until smartfs unmounted.

os/fs/driver/mtd/smart.c
Taejun-Kwon: fixed.

PR #4376, opened 2020-07-10
os/fs/smartfs: Refactoring of code for Smartfs file operations

amogh-samsung:
- Separate functions to find available entries,
  create new entries and write new entries.
- Handle renaming of files at the same location
  separately for optimization.
- Pre-allocation of firstsector for new file/dir
  entries.
- Separate function to invalidate old entry
  wherever required.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -3782,7 +3782,6 @@ static int smart_write_alloc_sector(FAR struct smart_struct_s *dev, uint16_t log
 	header = (FAR struct smart_sect_header_s *)dev->rwbuffer;
 	header->logicalsector[0] = (uint8_t)(logical & 0x00FF);
 	header->logicalsector[1] = (uint8_t)(logical >> 8);
-	//*((FAR uint16_t *)header->logicalsector) = logical;
 
 	header->seq = 0;
 
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -178,6 +178,9 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	uint16_t parentdirsector;
 	const char *filename;
 	struct smartfs_ofile_s *sf;
+	struct smartfs_entry_s newentry;
+	uint16_t data_sector;
+	bool new_chain = FALSE;
 
 	/* Sanity checks */
 
@@ -265,7 +268,6 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		}
 	} else if (ret == -ENOENT) {
 		/* The file does not exist.  Were we asked to create it? */
-
 		if ((oflags & O_CREAT) == 0) {
 			/* No.. then we fail with -ENOENT */
 
@@ -274,12 +276,44 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		}
 
 		/* Yes... test if the parent directory is valid */
-
 		if (parentdirsector != 0xFFFF) {
 			/* We can create in the given parent directory */
-
-			ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
+			/* First we allocate a new data sector for the file */
+			ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_FILE, sf);
 			if (ret != OK) {
+				fdbg("Failed to allocate data scetor for entry\n");
+				goto errout_with_buffer;
+			}
+
+			/* First try to find an invalid or empty entry available in one of the chained parent sectors */
+			ret = smartfs_find_availableentry(fs, parentdirsector, &newentry);
+			if (ret != OK) {
+				if (ret == -ENOENT) {
+
+					/* If no available entry was found, we have to chain anew sector and allocate an entry.
+					   In this case, new entry will return with the details of the last chained sector,
+					   so, we will just chain a new sector after this.
+					 */
+					parentdirsector = newentry.dsector;
+					new_chain = TRUE;
+					ret = smartfs_createentry(fs, parentdirsector, &newentry);
+					if (ret != OK) {
+						fdbg("smartfs_createentry failed\n");
+						goto errout_with_buffer;
+					}
+				} else {
+					/* find_availableentry encountered a problem with lower layer operations, return error */
+					fdbg("smartfs_find_availableentry() encountered a problem, unable to find entry for writing\n");
+					goto errout_with_buffer;
+				}
+			}
+
+			/* At this point, either an available entry was found or a new one has been created */
+			newentry.firstsector = data_sector;
+
+			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, new_chain);
+			if (ret != OK) {
+				fdbg("Write entry failed\n");
 				goto errout_with_buffer;
 			}
 		} else {
@@ -487,10 +521,7 @@ static ssize_t smartfs_read(FAR struct file *filep, char *buffer, size_t buflen)
 
 		/* Read the curent sector into our buffer */
 
-		readwrite.logsector = sf->currsector;
-		readwrite.offset = 0;
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-		readwrite.count = fs->fs_llformat.availbytes;
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d reading sector %d data\n", ret, sf->currsector);
@@ -600,11 +631,7 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 
 		/* Write the entire sector to FLASH */
 
-		readwrite.logsector = sf->currsector;
-		readwrite.offset = 0;
-		readwrite.count = fs->fs_llformat.availbytes;
-		readwrite.buffer = sf->buffer;
-
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
 		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d writing used bytes for sector %d\n", ret, sf->currsector);
@@ -624,11 +651,8 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 
 		/* Read the existing sector used bytes value */
 
-		readwrite.logsector = sf->currsector;
-		readwrite.offset = 0;
 		header = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-		readwrite.count = sizeof(struct smartfs_chain_header_s);
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
 		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d reading sector %d data\n", ret, sf->currsector);
@@ -652,10 +676,7 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 			header->used[1] = (uint8_t)(tmp >> 8);
 		}
 #endif
-		readwrite.offset = offsetof(struct smartfs_chain_header_s, used);
-		readwrite.count = sizeof(uint16_t);
-		readwrite.buffer = (uint8_t *)&fs->fs_rwbuffer[readwrite.offset];
-
+		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
 		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d writing used bytes for sector %d\n", ret, sf->currsector);
@@ -733,11 +754,7 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 		 * loop below to write the additional data to the end of the file.
 		 */
 
-		readwrite.offset = sf->curroffset;
-		readwrite.logsector = sf->currsector;
-		readwrite.buffer = (uint8_t *)&buffer[byteswritten];
-		readwrite.count = fs->fs_llformat.availbytes - sf->curroffset;
-
+		smartfs_setbuffer(&readwrite, sf->currsector, sf->curroffset, fs->fs_llformat.availbytes - sf->curroffset, (uint8_t *)&buffer[byteswritten]);
 		/* Limit the write based on available data to write */
 
 		if (readwrite.count > buflen) {
@@ -777,9 +794,7 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 			 * header to get the sector chain info.
 			 */
 
-			readwrite.offset = 0;
-			readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-			readwrite.count = sizeof(struct smartfs_chain_header_s);
+			smartfs_setbuffer(&readwrite, sf->currsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
 			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 			if (ret < 0) {
 				fdbg("Error %d reading sector %d header\n", ret, sf->currsector);
@@ -810,10 +825,7 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 		sf->bflags |= SMARTFS_BFLAG_DIRTY;
 
 #else							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
-		readwrite.offset = sf->curroffset;
-		readwrite.logsector = sf->currsector;
-		readwrite.buffer = (uint8_t *)&buffer[byteswritten];
-		readwrite.count = fs->fs_llformat.availbytes - sf->curroffset;
+		smartfs_setbuffer(&readwrite, sf->currsector, sf->curroffset, fs->fs_llformat.availbytes - sf->curroffset, (uint8_t *)&buffer[byteswritten]);
 		if (readwrite.count > buflen) {
 			/* Limit the write base on remaining bytes to write */
 
@@ -907,10 +919,7 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 				header->nextsector[0] = (uint8_t)(ret & 0x00FF);
 				header->nextsector[1] = (uint8_t)((ret >> 8) & 0x00FF);
 
-				readwrite.offset = offsetof(struct smartfs_chain_header_s, nextsector);
-				readwrite.buffer = (uint8_t *)header->nextsector;
-				readwrite.count = sizeof(uint16_t);
-
+				smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)header->nextsector);
 				ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 				if (ret < 0) {
 					fdbg("Error %d writing next sector\n", ret);
@@ -1033,10 +1042,7 @@ static off_t smartfs_seek_internal(struct smartfs_mountpt_s *fs, struct smartfs_
 	while ((sf->currsector != SMARTFS_ERASEDSTATE_16BIT) && (sf->filepos + fs->fs_llformat.availbytes - sizeof(struct smartfs_chain_header_s) < newpos)) {
 		/* Read the sector's header */
 
-		readwrite.logsector = sf->currsector;
-		readwrite.offset = 0;
-		readwrite.count = sizeof(struct smartfs_chain_header_s);
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
 		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d reading sector %d header\n", ret, sf->currsector);
@@ -1062,10 +1068,7 @@ static off_t smartfs_seek_internal(struct smartfs_mountpt_s *fs, struct smartfs_
 	 */
 
 	if (sf->currsector != SMARTFS_ERASEDSTATE_16BIT) {
-		readwrite.logsector = sf->currsector;
-		readwrite.offset = 0;
-		readwrite.count = fs->fs_llformat.availbytes;
-		readwrite.buffer = (uint8_t *)sf->buffer;
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
 		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d reading sector %d header\n", ret, sf->currsector);
@@ -1326,10 +1329,7 @@ static int smartfs_readdir(struct inode *mountpt, struct fs_dirent_s *dir)
 	while (dir->u.smartfs.fs_currsector != SMARTFS_ERASEDSTATE_16BIT) {
 		/* Read the logical sector */
 
-		readwrite.logsector = dir->u.smartfs.fs_currsector;
-		readwrite.count = fs->fs_llformat.availbytes;
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-		readwrite.offset = 0;
+		smartfs_setbuffer(&readwrite, dir->u.smartfs.fs_currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			goto errout_with_semaphore;
@@ -1658,9 +1658,12 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 {
 	struct smartfs_mountpt_s *fs;
 	int ret;
-	struct smartfs_entry_s entry;
+	struct smartfs_entry_s entry; //Details of newly written entry are returned with this variable
 	uint16_t parentdirsector;
 	const char *filename;
+	struct smartfs_entry_s newentry;
+	uint16_t data_sector;
+	bool new_chain = FALSE;
 
 	/* Sanity checks */
 
@@ -1676,7 +1679,6 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 
 	entry.name = NULL;
 	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
-
 	/* Three possibililities: (1) a node exists for the relpath and
 	 * dirinfo describes the directory entry of the entity, (2) the
 	 * node does not exist, or (3) some error occurred.
@@ -1688,26 +1690,45 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 		ret = -EEXIST;
 		goto errout_with_semaphore;
 	} else if (ret == -ENOENT) {
+
 		/* It doesn't exist ... we can create it, but only if we have
 		 * the right permissions and if the parentdirsector is valid. */
 
 		if (parentdirsector == 0xFFFF) {
+			fdbg("Invalid parent dir sector\n");
 			/* Invalid entry in the path (non-existant dir segment) */
-
 			goto errout_with_semaphore;
 		}
-
-		/* Check mode */
-
 		/* Create the directory */
-
-		ret = smartfs_createentry(fs, parentdirsector, filename,
-								  SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
+		/* Allocate new data sector for entry */
+		ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_DIR, NULL);
 		if (ret != OK) {
+			fdbg("Failed to allocate data sector for entry\n");
 			goto errout_with_semaphore;
 		}
-
-		ret = OK;
+		/* Try to find empty/invalid entry available in one of he chained parent sectors */
+		ret = smartfs_find_availableentry(fs, parentdirsector, &newentry);
+		if (ret == -ENOENT) {
+			parentdirsector = newentry.dsector;
+			new_chain = TRUE;
+			/* No available entry, add new sector to parent chain and allocate entry */
+			ret = smartfs_createentry(fs, parentdirsector, &newentry);
+			if (ret != OK) {
+				fdbg("Createentry failed, cannot create new entry for writing\n");
+				goto errout_with_semaphore;
+			}
+		} else if (ret != OK) {
+			fdbg("find_availableentry failed, cannot find entry for writing\n");
+			goto errout_with_semaphore;
+		}
+		/* Now we have an entry allocated for writing */
+		newentry.firstsector = data_sector;
+		/* Write new entry to sector */
+		ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, new_chain);
+		if (ret != OK) {
+			fdbg("Failed to write new entry to sector\n");
+			goto errout_with_semaphore;
+		}
 	}
 
 errout_with_semaphore:
@@ -1809,8 +1830,8 @@ errout_with_semaphore:
 
 int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *newrelpath)
 {
-	struct smartfs_mountpt_s *fs;
 	int ret;
+	struct smartfs_mountpt_s *fs;
 	struct smartfs_entry_s oldentry;
 	uint16_t oldparentdirsector;
 	const char *oldfilename;
@@ -1824,9 +1845,8 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	uint16_t entrysize;
 	struct smartfs_entry_header_s *direntry;
 	struct smart_read_write_s readwrite;
-
-	uint8_t *tmp_pntr = NULL;
-	uint16_t tmp_flag = 0;
+	struct smartfs_entry_s f_newentry;
+	bool new_chain = FALSE;
 
 	/* Sanity checks */
 
@@ -1844,6 +1864,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	newentry.name = NULL;
 	ret = smartfs_finddirentry(fs, &oldentry, oldrelpath, &oldparentdirsector, &oldfilename);
 	if (ret < 0) {
+		fdbg("Old entry doesn't exist\n");
 		goto errout_with_semaphore;
 	}
 
@@ -1867,14 +1888,12 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 
 		/* Nope, it's a directory.  Now search the directory for oldfilename */
 
+		newfilename = oldfilename;
 		sector = newentry.firstsector;
 		while (sector != SMARTFS_ERASEDSTATE_16BIT) {
 			/* Read the next sector of diretory entries */
 
-			readwrite.logsector = sector;
-			readwrite.offset = 0;
-			readwrite.count = fs->fs_llformat.availbytes;
-			readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
+			smartfs_setbuffer(&readwrite, sector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 			if (ret < 0) {
 				fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
@@ -1889,7 +1908,8 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 				/* Test the next entry */
 
 				direntry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-				if (((direntry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((direntry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
+				/* Check if entry flags are erased OR empty OR inactive and skip the entry */
+				if ((direntry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((direntry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((direntry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
 					/* This entry isn't valid, skip it */
 
 					offset += entrysize;
@@ -1914,57 +1934,55 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		}
 
 		/* Okay, we will create newfilename in the newentry directory */
-
 		newparentdirsector = newentry.firstsector;
-		newfilename = oldfilename;
 	}
 
 	/* Test if the new parent directory is valid */
-
 	if (newparentdirsector != 0xFFFF) {
 		/* We can move to the given parent directory */
 
 		mode = oldentry.flags & SMARTFS_DIRENT_MODE;
 		type = oldentry.flags & SMARTFS_DIRENT_TYPE;
 
-		ret = smartfs_createentry(fs, newparentdirsector, newfilename, type, mode, &newentry, oldentry.firstsector, NULL);
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+		if (oldparentdirsector == newparentdirsector) {
+			/* We will not use any new entry found, we will overwrite the existing entry */
+			ret = smartfs_writeentry(fs, oldentry, newfilename, type, mode, &newentry, new_chain);
+			if (ret != OK) {
+				fdbg("Error writing new entry\n");
+			}
+			/* Old entry doesn't have to be invalidated, directly go to end */
+			goto errout_with_semaphore;
+		}
+#endif
+		/* Find an available invalid/empty entry to write the new entry */
+		ret = smartfs_find_availableentry(fs, newparentdirsector, &f_newentry);
 		if (ret != OK) {
+			if (ret == -ENOENT) {
+				newparentdirsector = f_newentry.dsector;
+				new_chain = TRUE;
+				ret = smartfs_createentry(fs, newparentdirsector, &f_newentry);
+				if (ret != OK) {
+					fdbg("create entry failed, couldn't allocate new entry\n");
+					goto errout_with_semaphore;
+				}
+			} else {
+				fdbg("find_availableentry encountered a problem, cannot find entry for writing\n");
+				goto errout_with_semaphore;
+			}
+		}
+
+		f_newentry.firstsector = oldentry.firstsector;
+		ret = smartfs_writeentry(fs, f_newentry, newfilename, type, mode, &newentry, new_chain);
+		if (ret != OK) {
+			fdbg("Error writing new entry\n");
 			goto errout_with_semaphore;
 		}
 
 		/* Now mark the old entry as inactive */
-
-		readwrite.logsector = oldentry.dsector;
-		readwrite.offset = 0;
-		readwrite.count = fs->fs_llformat.availbytes;
-		readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-		ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-		if (ret < 0) {
-			fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
-			goto errout_with_semaphore;
-		}
-		// bug fix for rename
-		tmp_pntr = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
-		tmp_flag = tmp_pntr[0];
-		tmp_flag |= (tmp_pntr[1] << 8);
-
-#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-		tmp_flag &= ~SMARTFS_DIRENT_ACTIVE;
-#else
-		tmp_flag |= SMARTFS_DIRENT_ACTIVE;
-#endif
-		tmp_pntr[0] = (uint8_t)(tmp_flag & 0x00FF);
-		tmp_pntr[1] = (uint8_t)((tmp_flag >> 8) & 0x00FF);
-
-		/* Now write the updated flags back to the device */
-
-		readwrite.offset = oldentry.doffset;
-		readwrite.count = sizeof(uint16_t);
-		readwrite.buffer = (uint8_t *)tmp_pntr;
-
-		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-		if (ret < 0) {
-			fdbg("Error %d writing flag bytes for sector %d\n", ret, readwrite.logsector);
+		ret = smartfs_invalidateentry(fs, oldentry.dsector, oldentry.doffset);
+		if (ret != OK) {
+			fdbg("Unable to invalidate entry\n");
 			goto errout_with_semaphore;
 		}
 	} else {
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -116,6 +116,43 @@ extern size_t smart_sect_header_size;
 						(SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))
 #endif
 
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+#define RESET_ENTRY_FLAGS(e) (smartfs_wrle16(&(e)->flags, SMARTFS_ERASEDSTATE_16BIT))
+#else
+#define RESET_ENTRY_FLAGS(e) ((e)->flags = SMARTFS_ERASEDSTATE_16BIT)
+#endif
+
+#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
+
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+#define SET_ENTRY_FLAGS(f, m, t) (smartfs_wrle16(&f, (uint16_t)(SMARTFS_DIRENT_ACTIVE | SMARTFS_DIRENT_DELETING | SMARTFS_DIRENT_RESERVED | t | (m & SMARTFS_DIRENT_MODE))));
+#else
+#define SET_ENTRY_FLAGS(f, m, t) (f = (uint16_t)(SMARTFS_DIRENT_ACTIVE | SMARTFS_DIRENT_DELETING | SMARTFS_DIRENT_RESERVED | t | (m & SMARTFS_DIRENT_MODE)));
+#endif
+
+#else                                                   /* CONFIG_SMARTFS_ERASEDSTATE == 0xFF */
+
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+#define SET_ENTRY_FLAGS(f, m, t) (smartfs_wrle16(&f, (uint16_t)(SMARTFS_DIRENT_EMPTY | t | (m & SMARTFS_DIRENT_MODE))));
+#else
+#define SET_ENTRY_FLAGS(f, m, t) (f = (uint16_t)(SMARTFS_DIRENT_EMPTY | t | (m & SMARTFS_DIRENT_MODE)));
+#endif
+
+#endif
+
+/* This macro determines whether an entry is invalid/empty or not,
+   i.e. whether it is available for writing a new entry.
+ */
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+#define IS_AVAILABLE_ENTRY(e) ((smartfs_rdle16(&(e)->flags) == SMARTFS_ERASEDSTATE_16BIT) ||\
+				((smartfs_rdle16(&(e)->flags) & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) ==\
+					(~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE))))
+#else
+#define IS_AVAILABLE_ENTRY(e) (((e)->flags == SMARTFS_ERASEDSTATE_16BIT) ||\
+				(((e)->flags & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) ==\
+					(~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE))))
+#endif
+
 #define SET_TO_REMAIN(v, n)		v[n / 8] |= (1 << (7 - (n % 8)))
 #define SET_TO_USED(v, n)		v[n / 8] &= ~(1 << (7 - (n % 8)))
 #define IS_SECTOR_REMAIN(v, n)	v[n / 8] & (1 << (7 - (n % 8)))
@@ -267,6 +304,34 @@ void smartfs_wrle32(uint8_t *dest, uint32_t val)
 
 	smartfs_wrle16(dest, (uint16_t)(val & 0xffff));
 	smartfs_wrle16(dest + 2, (uint16_t)(val >> 16));
+}
+
+/****************************************************************************
+ * Name: smartfs_setbuffer
+ *
+ * Description:
+ *   Set the attributes of the smart_read_write_s structure objects for
+ *   BIOC_READSECT and BIOC_WRITESECT requests.
+ *
+ * Input Parameters:
+ *   rw - pointer to read_write object to be set.
+ *   logsector - logicalsector to read from/write to.
+ *   offset - offset to read from/write to.
+ *   count - number of btes to be read/written.
+ *   buffer - data to buffer to read to/write from.
+ *
+ * Returned Value:
+ *   None
+ *
+ ***************************************************************************/
+
+void smartfs_setbuffer(struct smart_read_write_s *rw, uint16_t logsector, uint16_t offset, uint16_t count, uint8_t *buffer)
+{
+	rw->logsector = logsector;
+	rw->offset = offset;
+	rw->count = count;
+	rw->buffer = buffer;
+	return;
 }
 
 #ifdef CONFIG_SMARTFS_DYNAMIC_HEADER
@@ -834,10 +899,7 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 			{
 				/* Read the next directory in the chain */
 
-				readwrite.logsector = dirsector;
-				readwrite.count = fs->fs_llformat.availbytes;
-				readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-				readwrite.offset = 0;
+				smartfs_setbuffer(&readwrite, dirsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 				ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 				if (ret < 0) {
 					goto errout;
@@ -914,14 +976,10 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 							if ((entry->flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE) {
 								dirsector = entry->firstsector;
 #endif
-								readwrite.count = sizeof(struct smartfs_chain_header_s);
-								readwrite.buffer = (uint8_t *)fs->fs_rwbuffer;
-								readwrite.offset = 0;
-
 								while (dirsector != SMARTFS_ERASEDSTATE_16BIT) {
 									/* Read the next sector of the file */
 
-									readwrite.logsector = dirsector;
+									smartfs_setbuffer(&readwrite, dirsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
 									ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 									if (ret < 0) {
 										fdbg("Error in sector chain at %d!\n", dirsector);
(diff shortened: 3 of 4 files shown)


os/fs/smartfs/smartfs_smart.c
```diff
@@ -1683,26 +1731,53 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
 		}
 
-		/* Check mode */
-
 		/* Create the directory */
 
-		ret = smartfs_createentry(fs, parentdirsector, filename,
-								  SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
+		/* Allocate new data sector for entry */
+		ret = smartfs_alloc_firstsector(&data_sector);
```
Taejun-Kwon, line 1749: Is it for first sector ? anyway please think about how we distinguish directory or file

os/fs/smartfs/smartfs_smart.c
```diff
@@ -277,9 +280,51 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+			/* First we allocate a new data sector for the file */
+			ret = smartfs_alloc_firstsector(&data_sector); //More attributes to be added
+			if (ret != OK) {
+				fdbg("\nError allocating new data sector for file\n");
+				goto errout_with_buffer;
+			}
+			fdbg("\nAllocated firstsector = %d\n", data_sector);
+
+			/* First try to find an invalid or empty entry available in one of the chained parent sectors */
+			ret = smartfs_findentry(fs, parentdirsector, &newentry);
```
Taejun-Kwon, line 292: it is very similar with finddirentry..
I really don't like invalidentry,,,but it seems that no way..
because of ENTRY_VALID MARO

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1683,26 +1731,53 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
+				fdbg("\ncreate entry fialed, cannot create new entry for writing\n");
+				goto errout_with_semaphore;
+			}
+		} else if (ret != OK) {
+			fdbg("\nfind entry failed, cannot find entry for writing\n");
+			goto erorut_with_semaphore;
+		}
+
+		/* Now we have an entry allocated for writing */
+		invalidentry.firstsector = data_sector;
```
Taejun-Kwon, line 1772: newentry right?

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Yes, to allocate a first sector for a new entry.
I will have to pass entry type.
Function yet to be written, I'll take care of that.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: I was also thinking that the name can be confusing.
Lets call it smartfs_find_availableentry or smartfs_find_unusedentry?

Also, VALID_ENTRY macro may be split up into two parts
1. IS_VALID_ENTRY - checks if flags are invalidated or not
2. A second check for if flags are in erased state

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Oh yes, missed out on this one.
Thanks.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	uint16_t psector;
+	uint16_t offset;
+	uint16_t entrysize;
+	struct smartfs_entry_header_s *entry;
 
-		/* Now update to the next sector */
+	/* Validate the name isn't too long */
 
-		psector = nextsector;
+	if (strlen(filename) > fs->fs_llformat.namesize) {
```
Taejun-Kwon, line 1237: Flow is create or find entry and then write entry right?
Then this should be checked very first time. 
So, start of create or find entry is correct position.
Then we don't have to waste many of procedure until it reaches here with long file name.
If we consider expandability, modulization is most important.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+		return -ENAMETOOLONG;
 	}
 
-	/* We found an insertion point.  Create the entry at sector,offset */
+	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
```
Taejun-Kwon, line 1241: Align with above. Below must be determined when it create new_entry.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
-			}
+errout:
+	fdbg("\nret = %d, invalidentry->dsector = %d, invalidentry->doffset = %d\n", ret, ret_entry->dsector, ret_entry->doffset);
+	if (ret != OK) {
+		ret = -ENOENT;
+	}
 
-			nextsector = (uint16_t)ret;
+	if (ret == OK) {
+		RESET_ENTRY_FLAGS(ret_entry);
```
Taejun-Kwon, line 1207: hmm can we update all entry flag inside of find or create entry?
I thin both of them should complete operation for new entry....
Please define role of each function first.
1. createentry : Create new entry if....
2. findentry: find...
3. writeentry : write entry data on flash.

So, if you modify some entry data in writeentry, actually it means writeentry does another function's role.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
-			}
+/****************************************************************************
+ * Name: smartfs_writeentry
+ *
+ * Description: Takes care of the operation of renaming an existing file
+ *              in a particular directory to a new name in the same dircetory
+ *              while minimizing the need to allocate new sectors for
+ *              writing the changed entry to MTD flash.
+ *
+ ****************************************************************************/
```
Taejun-Kwon, line 1223: Like a this function, it has many parameters and same type of params also there.(new_entry, ret_entry).
In this case you add comment regarding them.
* Input Parameters :
* fs
* new_entry
* blah
* lol
* rofl

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
-			  * It is temp FIX, If we use sector buffer to support CRC, currently created new parent sector which is chained 
-			  * to previous parent sector is not committed yet. So Now we break loop to prevent read error 
-			  */
-			offset = sizeof(struct smartfs_chain_header_s);
-			entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-			psector = nextsector;
-			break;
-#endif
-		}
+int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *ret_entry)
```
Taejun-Kwon, line 1225: Please align callee & caller.
you always use new_entry for last parameter, but here it is ret_entry.
It is needed when some kind of type casting, but we do not cast or nothing related to casting. (I like newentry, actually =D)

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+		if (ret_entry->name == NULL) {
 			ret = ERROR;
 			goto errout;
 		}
 	}
 
-	memset(direntry->name, 0, fs->fs_llformat.namesize + 1);
-	strncpy(direntry->name, filename, fs->fs_llformat.namesize);
+	memset(ret_entry->name, 0, fs->fs_llformat.namesize + 1);
+	strncpy(ret_entry->name, filename, fs->fs_llformat.namesize);
```
Taejun-Kwon, line 1314: yeah.. this also need to be checked.
I think this should be done when it created.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	struct smart_read_write_s readwrite;
+	uint16_t alloc_sector;
+	uint8_t chainheader_type;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	struct smartfs_chain_header_s *chainheader;
+#endif
+
+	ret = FS_IOCTL(fs, BIOC_ALLOCSECT, 0xFFFF);
+	if (ret < 0) {
+		return ret;;
```
Taejun-Kwon, line 1342: please remove one.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1922,44 +2007,65 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+			{
+				/* No emtpy entry was found available in one of the chained parent sectors */
+				/* We will not use any new entry found, we will overwrite the existing entry */
+				ret = smartfs_writeentry(fs, oldentry, newfilename, type, mode, &newentry);
+				if (ret != OK) {
+					fdbg("\nError writing new entry\n");
+					goto errout_with_semaphore;
+				}
+				/* Old entry doesn't have to be invalidated */
+				goto errout_with_semaphore;
```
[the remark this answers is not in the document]
amogh-samsung: After (oldparentdirsector == newparentdirsector) and the entry has been written,
Further checks are not needed and old entry invalidation is also not needed.
So we can go directly to the end of the function.

Even though the label is errout_with_semaphore, the value of ret = OK.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -277,9 +280,51 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+				 */
+				parentdirsector = newentry.dsector;
+
+				/* Not passing filename, entry type, mode since entry will not be written by createentry,
+				   Not passing sectorno/firstsector = 0xFFFF since createentry will not allocate new firstsector.
+				   sf->entry returns with the details of the new entry that is written, so it will be 
+				   passed to the entry writing function
+				 */
+				ret = smartfs_createentry(fs, parentdirsector, &newentry);
+				fdbg("\nCreateentry returns ret = %d, logicalsector = %d\n", ret, newentry.dsector);
```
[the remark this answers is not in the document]
amogh-samsung: Unnecessary logs will be removed.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1028,35 +1063,98 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
 
-	/* Validate the name isn't too long */
+	chainheader = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
+	nextsector = SMARTFS_NEXTSECTOR(chainheader);
 
-	if (strlen(filename) > fs->fs_llformat.namesize) {
-		return -ENAMETOOLONG;
+	if (nextsector != SMARTFS_ERASEDSTATE_16BIT) {
+		fdbg("\nOops, something did not go as planned\n");
 	}
```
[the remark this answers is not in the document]
amogh-samsung: Unnecessary logs will be removed.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -125,6 +126,40 @@ extern size_t smart_sect_header_size;
...
+#endif
+
+#endif
+
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+#define IS_EMPTY_ENTRY(e) ((smartfs_rdle16(&(e)->flags) == SMARTFS_ERASEDSTATE_16BIT) ||\
+				((smartfs_rdle16(&(e)->flags) & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) ==\
+					 (~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE))))
+#else
+#define IS_EMPTY_ENTRY(e) ((e->flags == SMARTFS_ERASEDSTATE_16BIT) ||\
```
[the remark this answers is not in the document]
amogh-samsung: The name of that Macro should be changed.
Earlier, createentry would check for both invalidated and emtpy entries with the same check.
So now, I have used a similar logic, but I'll rename this to `IS_AVAILABLE_ENTRY(e)`, what do you say?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1065,209 +1163,276 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
-#endif
-					(SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) == (~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)))) {
-				/* We found an empty entry.  Use it. */
+		while ((offset + entrysize) < readwrite.count) {
+			/* Check if this entry is available */
+			fdbg("\nEntry flags = %x\n", entry->flags);
+			if (IS_EMPTY_ENTRY(entry)) {
+				if (is_empty && entry->flags == SMARTFS_ERASEDSTATE_16BIT) {
+					fdbg("\nEmpty entry is foud\n");
+					*is_empty = TRUE;
```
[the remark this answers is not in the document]
amogh-samsung: I'll keep a single check which includes checking for both invalid and empty entries.
We can work with a common check, don't need to check separately.

We need to know if it is an empty entry only in one specific case, so for that I'll use this if construct.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1922,44 +2007,65 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+			if (ret != OK) {
+				fdbg("\ncreate entry failed, couldn't allocate new entry\n");
+				goto errout_with_semaphore;
+			}
+		}
 
-		readwrite.offset = oldentry.doffset;
-		readwrite.count = sizeof(uint16_t);
-		readwrite.buffer = (uint8_t *)tmp_pntr;
+		f_newentry.firstsector = oldentry.firstsector;
```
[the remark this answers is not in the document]
amogh-samsung: Yes agreed, Done.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes, agreed. I will move this check to smartfs_smart.c.
Every function hat involves creation of an entry with a new name will go thorugh this check first.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Done,
Now writeentry takes care of entry->flags

If the entry is being reused, writeentry will first clear its flags and then set new flags before writing the entry to MTD.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes, done.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: `newentry` is always a pointer to the entry that will hold the details of the newly found (in case of find_availableentry) or created (in case of createentry) entry returned for writing.

`ret_entry` is always a pointer to the entry that will hold details of the entry that was written to MTD by writeentry

So it becomes kind of self-explanatory
`newentry `- newly allocated entry for writing
`ret_entry `- returning final entry that was written to MTD

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I didn't get this comment sorry?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I didn't get this comment sorry?

amogh-samsung:
1. Extra debug logs have not been removed yet. They will be taken care of after testing is complete. Till then, this PR will stay a draft only.
2. More changes are pending.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1912,54 +1961,83 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+		ret = smartfs_writeentry(fs, f_newentry, newfilename, type, mode, &newentry);
```
Taejun-Kwon, line 2028: This is corner case.
If power failed after this line, we have 2 entry which points same first sector...
How we realize which latest?
We need discuss it.
Actually my idea is use utc in entry header,
if CONFIG_RTC is enabled, we can use RTC, otherwise sequence number.
For example. we want rename /mnt/a to /mnt/test/a and it's first sector is logical 20.
/mnt/a entry has low entry->utc then /mnt/test/a's entry->utc
During the fs_recover after journal recover, if any logical sector chained to more than 1 time,
Then invalidate old one.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Ohkay, I can keep the entry returned by writeentry after writing to MTD as direntry.
Let's do that.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1026,248 +1087,407 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+	return ret;
+}
 
-			offset += entrysize;
-			entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-		}
+/****************************************************************************
+ * Name: smartfs_find_availableentry
+ *
+ * Description: Finds an invalid entry in the parent directory that can be
```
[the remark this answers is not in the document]
amogh-samsung: Yes, finds both.
Returns whichever is found first.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -273,13 +276,56 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+		}
+
 		/* Yes... test if the parent directory is valid */
 
 		if (parentdirsector != 0xFFFF) {
 			/* We can create in the given parent directory */
+			/* First we allocate a new data sector for the file */
+			ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_FILE, sf);
+			if (ret != OK) {
+				fdbg("\nError allocating new data sector for file\n");
```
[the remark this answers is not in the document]
amogh-samsung: Sorry, debug logs will all be reconsidered.
Currently in place for my convenience of testing.
Will be checked according to our coding practices before this PR is marked ready for Review.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1026,248 +1087,407 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+		/* Search for an available entry in this sector */
+		offset = sizeof(struct smartfs_chain_header_s);
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
 
-			/* Chain the next sector into this sector sector */
+		while ((offset + entrysize) < readwrite.count) {
+			/* Check if this entry is available */
+			fdbg("\nEntry flags = %x\n", entry->flags);
+			if (IS_AVAILABLE_ENTRY(entry)) {
+				if (is_empty && entry->flags == SMARTFS_ERASEDSTATE_16BIT) {
```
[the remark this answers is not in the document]
amogh-samsung: This is a tricky one.
I will put this as a comment in the general discussion board for this PR.
This has been done for maximum optimization of smartfs_rename.

amogh-samsung:
`if (is_empty && entry->flags == SMARTFS_ERASEDSTATE_16BIT)`
Taejun Kwon commented: - "I still think is_empty is useless, because (entry->flags == SMARTFS_ERASESTATE_16BIT) is enough..."

*is_empty is a pointer to a boolean used to return if the entry found for allocation in function smartfs_find_availableentry is empty or not (invalidated one otherwise).
*is_empty is only passed as not NULL in the case of smartfs_rename. mkdir and open will pass *is_empty as NULL.

This is only for the case of Simple RENAME i.e. changing the file name in the same path ONLY.
1. If an empty entry was found for writing, then going by the current (old) method works better as there are 2 MTD writes but 0 
    new sectors allocated (with CRC disabled).
2. In ALL other cases, CRC enabled and CRC disabled with invalid entry found/no entry found, it is better to use new method i.e. 
    use the existing entry to overwrite the new entry.

So, is_empty when passed != NULL returns with the type of entry allocated and we use the new or old method for rename depending on it.

os/fs/smartfs/smartfs_smart.c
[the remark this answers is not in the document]
amogh-samsung: Ah yes, thankyou

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1026,248 +1087,407 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+	fdbg("\nret = %d, invalidentry->dsector = %d, invalidentry->doffset = %d\n", ret, ret_entry->dsector, ret_entry->doffset);
+	fdbg("\nfIND_FIRSTINVALIDENTRY returns invalidentry->flags = %x\n", ret_entry->flags);
+	return ret;
+}
 
-		nextsector = (uint16_t)ret;
+/****************************************************************************
+ * Name: smartfs_writeentry
+ *
+ * Description: Takes care of the operation of renaming an existing file
```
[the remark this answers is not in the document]
amogh-samsung: Corrected

os/fs/smartfs/smartfs_utils.c
```diff
@@ -247,6 +285,34 @@ void smartfs_wrle32(uint8_t *dest, uint32_t val)
...
+ *
+ ***************************************************************************/
+
+void smartfs_set_readwrite(struct smart_read_write_s *rw, uint16_t logsector, uint16_t offset, uint16_t count, uint8_t *buffer)
+{
+	(*rw).logsector = logsector;
+	(*rw).offset = offset;
+	(*rw).count = count;
+	(*rw).buffer = buffer;
+	return;
```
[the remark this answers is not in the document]
amogh-samsung: Yes, done

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1683,26 +1698,57 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
+		} else if (ret != OK) {
+			fdbg("\nfind entry failed, cannot find entry for writing\n");
+			goto errout_with_semaphore;
+		}
+
+		/* Now we have an entry allocated for writing */
+		newentry.firstsector = data_sector;
+
+		/* Write new entry to sector */
+		ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry);
```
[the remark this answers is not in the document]
amogh-samsung: Got it.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Yes, as discussed,
Seq number + fs_recovery will do the job.

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
amogh-samsung: With new logic, all flag ops moved to smartfs_writeentry.
The same logic can be worked without the boolean pointer `is_empty`.

The entry flags are checked after find_availableentry in smartfs_rename for the specific rename case.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -273,13 +276,56 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+					fdbg("\nsmartfs_find_availableentry() encountered a problem, unable to find entry for writing\n");
+					goto errout_with_buffer;
+				}
+			}
 
-			ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
+			/* At this point, either an available entry was found or a new one has been created */
+			newentry.firstsector = data_sector;
+
+			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry);
```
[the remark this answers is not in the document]
amogh-samsung: This has some confusion/discrepancies
Current code writes new entry to sector via  smartfs_createentry.
Makes call via IOCTL with BIOC_WRITESECT.

So I implemented the same control flow with new refactored code.

Maybe some logic changes need to be implemented, perhaps separately.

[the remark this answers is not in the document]
Taejun-Kwon: OK but, first you have to test it.
As I know entry will be written in sync_internal.
So if power down happend right after this code, entry has wrong information of next sector. you have to verify it.
If I am right, then Please add 'TODO' comment like a below
/* TODO New entry should be chained with parent sector after new entry commited in sync_internal */

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1028,35 +1063,98 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+	/* Allocate a new sector and chain it to the last one */
+
+	ret = FS_IOCTL(fs, BIOC_ALLOCSECT, 0xFFFF);
+	if (ret < 0) {
+		goto errout;
+	}
+
+	nextsector = (uint16_t)ret;
+
+	/* Chain the next sector into this sector sector */
```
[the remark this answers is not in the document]
Taejun-Kwon: Ignore above comment, I think we'd better handle it in smartfs_open & sync_internal as I commented in smartfs_open, line 326

Taejun-Kwon:
Please clean fdbg before change state to 'ready to review'

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
Taejun-Kwon: Please change description, that's what I asking you.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1912,54 +1960,87 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
 		newfilename = oldfilename;
+		/* Check this, does this function always use the old file name for the new file?
+		   Or does it use the old name only when a new name is not provided for a move request?
+		 */
+		fdbg("\nNew file name changed to %s\n", newfilename);
 	}
 
-	/* Test if the new parent directory is valid */
+	/* Ensure that the new file name isn't too long */
+	if (strlen(newfilename) > fs->fs_llformat.namesize) {
```
Taejun-Kwon, line 1970: I think this is not good option.
All mountpoint operation which access entry data checks validation of entry with target path through the smartfs_finddirentry..
I think we  can check this at the start of there.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1021,226 +1082,252 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+ *     written.
+ *   filename - name of new entry.
+ *   type - type of new entry (file/dir).
+ *   mode - mode of creation of new entry.
+ *   direntry - pointer to smartfs entry object to return details of newly
+ *     written entry.
+ *
+ * Returned Values:
+ *   OK - on successfully writing new entry to MTD.
+ *   ERROR codes - appropriaely returned otherwise.
```
jeongchanKim, line 1265: What is the meaning of ERROR codes? And please check the spell `appropriaely `

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1021,226 +1082,252 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+	direntry->doffset = sizeof(struct smartfs_chain_header_s);
+	direntry->dsector = nextsector;
 
-			/* Not available.  Skip to next entry */
+	ret = OK;
 
-			offset += entrysize;
-			entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-		}
+errout:
```
jeongchanKim, line 1159: If there is nothing to do before return, it is not good to use `goto` syntax.

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
amogh-samsung: Yes, I was a bit confused with this.
Hence I hadn't made any changes regarding this yet.

We can do as discussed for smartfs_open and smartfs_sync_internal.

We hold the new entry from being written till the sf buffer is full with data for the firstsector of the file and it is written to MTD.
So we write new entry and firstsector together to avoid chaining errors.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkkay,
Can be moved to function smartfs_finddirentry().

However, on an unrelated note
one key change is required in smartfs_rename while checking whether new filename exists.

amogh-samsung:
TODO: - separate bug resolutions for future PRs perhaps
1. Write MTD header to newly allocated sector when CRC is enabled.
2. Smart_allocsector returns sectors which are not empty with smartfs_open when sf is used. 
    Flow of sector allocation and entry writing needs to be re checked.
3. New entry with smartfs_open can be held until sf buffer is full with data and can be written to MTD.
    New entry should also be written at the same time.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -247,6 +285,34 @@ void smartfs_wrle32(uint8_t *dest, uint32_t val)
 	smartfs_wrle16(dest + 2, (uint16_t)(val >> 16));
 }
 
+/****************************************************************************
+ * Name: smartfs_setbuffer
+ *
+ * Description:
+ *   Set the attributes of the smart_read_write_s sturcture objects for
```
amandeepc, line 292: * structure

os/fs/smartfs/smartfs_utils.c
```diff
@@ -247,6 +285,34 @@ void smartfs_wrle32(uint8_t *dest, uint32_t val)
...
+ *   BIOC_READSECT and BIOC_WRITESECT requests.
+ *
+ * Input Parameters:
+ *   rw - pointer to read_write object to be set.
+ *   logscetor - logicalsector to read from/write to.
```
amandeepc, line 297: *** logsector

os/fs/smartfs/smartfs_utils.c

os/fs/smartfs/smartfs_utils.c

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
 	/* Now write the new entry to the parent directory sector */
-
-	readwrite.logsector = psector;
-	/* If it chained right before, we should update chainheader */
-	if (update_chain) {
-		memset(fs->fs_rwbuffer, CONFIG_SMARTFS_ERASEDSTATE, sizeof(struct smartfs_chain_header_s));
-		chainheader = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
-		chainheader->type = SMARTFS_SECTOR_TYPE_DIR;
-		readwrite.count = entrysize + sizeof(struct smartfs_chain_header_s);
+	if (newchain) {
```
Taejun-Kwon, line 1318: yes it also we don't need it. My old PR just changed sequence of chaing...not affect to this.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1867,14 +1893,12 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 
 		/* Nope, it's a directory.  Now search the directory for oldfilename */
 
+		newfilename = oldfilename;
```
Taejun-Kwon, line 1896: below code doesn't use newfile name, so original position is correct. also align with comment good too.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1914,57 +1938,79 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
-		// bug fix for rename
-		tmp_pntr = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
-		tmp_flag = tmp_pntr[0];
-		tmp_flag |= (tmp_pntr[1] << 8);
-
-#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-		tmp_flag &= ~SMARTFS_DIRENT_ACTIVE;
+		/* Check if it is a simple rename operation, i.e. in the same directory */
+		if (oldparentdirsector == newparentdirsector) {
+#ifndef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
Taejun-Kwon, line 1960: In my opinion, this is wrong, smartfs_find_availableentry should find entry based on CONFIG_SMARTFS_USE_SECTOR_BUFFER. (Empty or Overwrite able)

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+		memset(fs->fs_rwbuffer, CONFIG_SMARTFS_ERASEDSTATE, entrysize + sizeof(struct smartfs_chain_header_s));
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
+		chainheader = (struct smartfs_chain_header_s *)&fs->fs_rwbuffer[0];
+		chainheader->type = SMARTFS_SECTOR_TYPE_DIR;
 	} else {
-		/* Use the provided sector number */
-
-		nextsector = sectorno;
+		memset(fs->fs_rwbuffer, CONFIG_SMARTFS_ERASEDSTATE, entrysize);
```
Taejun-Kwon, line 1300: I guess this is wrong...
In this time fs_rwbuffer may include information of chain header this never be cleared.
So you should use below, without memset.
entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1287,6 +1358,125 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+
+		smartfs_setbuffer(&readwrite, alloc_sector, offsetof(struct smartfs_chain_header_s, type), 1, (uint8_t *)&chainheader_type);
+		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
+		if (ret < 0) {
+			fdbg("Error %d setting new sector type for sector %d\n", ret, alloc_sector);
+			return ret;
+		}
+	}
+
+	(*sector) = alloc_sector;
```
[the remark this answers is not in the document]
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-				found = TRUE;
-				break;
+		while ((offset + entrysize) < readwrite.count) {
+			/* Check if this entry is available */
+			if (IS_AVAILABLE_ENTRY(entry)) {
+				direntry->dsector = psector;
+				direntry->doffset = offset;
+				direntry->flags = entry->flags;
+				ret = OK;
+				goto errout;
```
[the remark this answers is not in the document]
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-			update_readwrite.count = sizeof(uint16_t);
-			update_readwrite.buffer = update_header.nextsector;
-
-			fvdbg("chain with next sector.. current : %d next : %d\n", psector, nextsector);
-			update_chain = 1;
-
+errout:
+	if (ret != OK) {
+		ret = -ENOENT;
+	}
```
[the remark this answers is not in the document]
amogh-samsung: Done

os/fs/smartfs/smartfs_smart.c
```diff
@@ -274,12 +276,48 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+			/* First we allocate a new data sector for the file */
+			ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_FILE, sf);
+			if (ret != OK) {
+				fdbg("Error allocating new data sector for file\n");
+				goto errout_with_buffer;
+			}
+
+			/* First try to find an invalid or empty entry available in one of the chained parent sectors */
+			ret = smartfs_find_availableentry(fs, parentdirsector, &newentry);
+			if (ret == -ENOENT) {
```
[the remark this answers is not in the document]
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-			if ((smartfs_rdle16(&entry->flags) == SMARTFS_ERASEDSTATE_16BIT) || ((smartfs_rdle16(&entry->flags) &
-#else
-			if ((entry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((entry->flags &
-#endif
-					(SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) == (~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)))) {
-				/* We found an empty entry.  Use it. */
 
-				found = TRUE;
-				break;
+		while ((offset + entrysize) < readwrite.count) {
```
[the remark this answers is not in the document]
amogh-samsung: The sector has been read till "offset".
In the current loop iteration we are reading the entry after the offset,
so we have to check if the entire entry is available in the readwrite buffer.

For Eg - after reading the 12th entry
Offset  = 12*40 + 5 = 485
Now if we check  (offset < readwrite.count) then  we will try to read next entry which is not there.
But if we check (offset + entrysize< readwrite.count) then we will not check for the next entry.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,251 +1076,261 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-			readwrite.offset = offsetof(struct smartfs_chain_header_s, type);
-			readwrite.buffer = (uint8_t *)&chainheader->type;
-			readwrite.logsector = nextsector;
-			ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-			if (ret < 0) {
-				fdbg("Error %d setting new sector type for sector %d\n", ret, nextsector);
-				goto errout;
-			}
-		}
+	if (newchain) {
```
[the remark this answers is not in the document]
amogh-samsung: Createentry allocates the new chained sector for the parent, psector->nextsector is updated with the allocated logicalsector number.
But chainheader for the new sector is not written in createentry, it is written along with the 1st entry in writeentry.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: With refactoring, your PR changes will also become outdated.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: We will NOT use the  new method for rename only when CRC is not enabled and an empty entry is found for allocation.
In all other cases it is better to use the new method i.e. overwrite the old entry.

I shall post the numbers backing this decision on confluence.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,232 +1076,259 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+ *
+ * Returned Values:
+ *   OK - on successfully writing new entry to MTD.
+ *   ERROR - appropriate error code returned otherwise.
+ *
+ ****************************************************************************/
 
-		psector = nextsector;
-	}
+int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry, bool newchain)
```
Taejun-Kwon, line 1266: Your code flows below. (For example we create /mnt/a)
1. First it create firstsector by smartfs_alloc_firstsector. 
    It request allocsect to MTD layer, and keep returned target alloc sect.
    So now. newly written sector : 0 newly allocated sector : 1 (firstsector)

2. Now it try to find availableentry to get offset where target entry can be inserted.
    If there is space, fs->rw_buffer hold data of entire sector.
    So now. newly written sector : 0 newly allocated sector : 1 (firstsector)

3. If no space in parent sector, it alloc new parent sector and update previous sector's chain.
   Actually it should be done last time
    So now. newly written sector : 0 newly allocated sector : 2 (firstsector, 2nd parent sector - 1st sector has next sector value)

4. Finally it comes here.
    But if there was no space in 1st parent sector, newchain set as TRUE.
    In this case, this funtion never write first sector... so link error will be happened.
    so first you should write firstsector, and then if chained, then write 2nd parent sector.
    (and then write chain value of first sector is correct sequence)
   
 So this code can create only 12 files per directory.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,232 +1076,259 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-			readwrite.logsector = nextsector;
-			ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-			if (ret < 0) {
-				fdbg("Error %d setting new sector type for sector %d\n", ret, nextsector);
-				goto errout;
-			}
-		}
+	if (newchain) {
+		memset(fs->fs_rwbuffer, CONFIG_SMARTFS_ERASEDSTATE, entrysize + sizeof(struct smartfs_chain_header_s));
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
```
Taejun-Kwon, line 1295: Below is better, it means first offset of entry explicitly, right after chainheader.
```
entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[sizeof(struct smartfs_entry_header_s)]; 
```

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1015,232 +1076,259 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-	readwrite.logsector = psector;
-	readwrite.offset = offset;
-	readwrite.count = entrysize;
-	readwrite.buffer = (uint8_t *)&fs->fs_rwbuffer[offset];
+	if (newchain) {
+		/* If this is a newly chained sector, write new entry and chain header both */
+		smartfs_setbuffer(&readwrite, psector, 0, entrysize + sizeof(struct smartfs_chain_header_s), (uint8_t *)&fs->fs_rwbuffer[0]);
+	} else {
+		/* Otherwise, we only have to write the new entry */
+		smartfs_setbuffer(&readwrite, psector, offset, entrysize, (uint8_t *)&fs->fs_rwbuffer[0]);
```
Taejun-Kwon, line 1322: Yes same as line 1299, below is correct.
```
smartfs_setbuffer(&readwrite, psector, offset, entrysize, (uint8_t *)&fs->fs_rwbuffer[offset]);
```

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: below is pseudo code.
```
		if (find_current_path_exist) {
			return -EEXIST;
		}
		
		smartfs_alloc_firstsector(); // Here we alloc first sector.
		
		Entry entry; //Entry which target to be inserted
		Entry parententry; //Information of parent entry;
		
		ret = find_availableentry(&entry, &parententry);
		if (ret == ENOENT) {
			/* Now we got new entry, also next parent sector's number if current parent is full */
			create_entry(&entry, &parententry); 
		} else {
			/* No next parent, now we found target entry & parententry's info */
			if (ret == OK) {
				fvdbg("found entry\n");
			} else {
				fdbg("error returned ret : %d errno : %d\n", ret, errno);
			}
		}

		/* Now we write first sector */
		if (sf) {
			sf->currsector = entry->firstsector;
			sf->parentsector = parentsector; //expand or reuse structure to represent it.
		} else {
			/* Below flow should be done by write_entry */
			write_sector(entry); // write first sector
		}
			
		/* Update chaining */
		if (chained) {
			if (sf) {
				sf->x = previous_sector_of_parent_sector;
			} else {
				write_next_parent(pareantentry);
				chain_parent(parententry);
			}
		}
```
We should expand sf for sync_internal...
```	
		sync_internal()
		
		write_sector(sf->currsector);
		write_sector(sf->parentsector);
		chain_parent(sf->parent, sf->x);
```

os/fs/smartfs/smartfs_smart.c
amogh-samsung: I think we should keep it this way only so that find_availableentry reamains a generic function.
Otherwise find_availableentry will have to behave specially and check specific case when smartfs_rename is the calling function.
I don't wish for that.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: My apologies, I thought you meant you agree with this change...

amogh-samsung: The comment that  you are referring to: -
`/* Okay, we will create newfilename in the newentry directory */`
I think this comment is fine.
It is implying that in the loop, no conflicting entry was found to exist,
So we'll create a new entry in the new parent directory now.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: You mean
entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[sizeof(struct smartfs_**chain**_header_s)];  ?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Made changes

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Made changes

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1914,57 +1938,79 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
 		}
-		// bug fix for rename
-		tmp_pntr = (uint8_t *)&fs->fs_rwbuffer[oldentry.doffset];
-		tmp_flag = tmp_pntr[0];
-		tmp_flag |= (tmp_pntr[1] << 8);
 
-#if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-		tmp_flag &= ~SMARTFS_DIRENT_ACTIVE;
+		/* Check if it is a simple rename operation, i.e. in the same directory */
+		if (oldparentdirsector == newparentdirsector) {
```
Taejun-Kwon, line 1959: if they are same sector, we will overwrite old one, right? so below logic is correct for now, In my opinion.
```
ret = smartfs_finddirentry(fs, &oldentry, oldrelpath, &oldparentdirsector, &oldfilename);
...
ret = smartfs_finddirentry(fs, &newentry, newrelpath, &newparentdirsector, &newfilename);

#if CONFIG_SMARTFS_USE_SECTOR_BUFFER
if (oldparentdirsector == newparentdirsector) {
	overwrite_old_sector;
} else 
#endif
	{
		smartfs_find_availableentry();
		if (ret == -ENOENT) {
			smartfs_createentry();
		} else {
			if (ret != OK) {
				fdbg("error.....");
				goto error;
			}
			fvdbg("Entry found in parent sector : %d, offset : %d\n");
		}
		smartfs_writeentry();
		invalidate_entry(oldentry);
	}
}
```
How do you think?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1040,259 +1101,284 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-				fdbg("Error %d setting new sector type for sector %d\n", ret, nextsector);
-				goto errout;
-			}
+	if (newchain) {
+		/* We cannot read the new sector into fs->fs_rwbuffer as it is totally empty.
+		   MTD header will be written when MTD_WRITE is called for the sector 1st time.
+		   Chainheader is written along with the 1st entry.
+		   So currently, smart_readsector will fail to invalidate CRC on empty sector
+		 */
+		tmp_buf = (char *)kmm_malloc(entrysize + sizeof(struct smartfs_entry_header_s));
```
Taejun-Kwon, line 1323: yes. good approach.
But you should consider sf & our flow line 1366..
so TODO is needed.
/* TODO if sf is available, do not write for now....*/

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1040,259 +1101,284 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
+		}
+		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
 	}
 
-	/* Create the directory entry to be written in the parent's sector */
+	/* Populate the directory entry to be written in the parent's sector */
 
 #ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
-	smartfs_wrle16(&entry->firstsector, nextsector);
+	smartfs_wrle16(&entry->firstsector, new_entry.firstsector);
```
Taejun-Kwon, line 1345: it's node duplicated with line 1381~1401?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1317,6 +1403,130 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	struct smart_read_write_s readwrite;
+
+	smartfs_setbuffer(&readwrite, oldentry.dsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
+	ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+	if (ret < 0) {
+		fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
+		return ret;
+	}
+
+	// bug fix for rename
```
Taejun-Kwon, line 1509: Actually another guy fixed this about 5-6 years ago but now I can't remember it.
but now I think this is useless, let me hear your opinion.

os/fs/driver/mtd/smart.c
```diff
@@ -4374,6 +4373,11 @@ static int smart_allocsector(FAR struct smart_struct_s *dev, unsigned long reque
 	int ret;
 #endif
 
+	/*TODO: smart_allocsector may return sectors with a temporary
```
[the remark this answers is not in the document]
amogh-samsung: Actually I forgot to remove this TODO comment.
Sector allocation works fine now.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1688,26 +1694,46 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 		ret = -EEXIST;
 		goto errout_with_semaphore;
 	} else if (ret == -ENOENT) {
+
```
sangam-swami, line 1697: remove extra newline

os/fs/smartfs/smartfs_smart.c
```diff
@@ -274,12 +276,48 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+			ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_FILE, sf);
+			if (ret != OK) {
+				fdbg("Error allocating new data sector for file\n");
+				goto errout_with_buffer;
+			}
+
+			/* First try to find an invalid or empty entry available in one of the chained parent sectors */
+			ret = smartfs_find_availableentry(fs, parentdirsector, &newentry);
+			if (ret != OK) {
+				if (ret == -ENOENT) {
```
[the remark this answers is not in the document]
amogh-samsung: I think you had given a comment earlier to handle all error cases under one if structure.
if (ret != OK i.e. ERROR) {
    then if (ret == -ENOENT)
    {}
    else
    {}
}

os/fs/smartfs/smartfs_smart.c
```diff
@@ -274,12 +276,48 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+					fdbg("smartfs_find_availableentry() encountered a problem, unable to find entry for writing\n");
+					goto errout_with_buffer;
+				}
+			}
+
+			/* At this point, either an available entry was found or a new one has been created */
+			newentry.firstsector = data_sector;
 
-			ret = smartfs_createentry(fs, parentdirsector, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, 0xFFFF, sf);
+			/*TODO: Do we need to write the entry here? The entry can be held
```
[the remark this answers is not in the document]
amogh-samsung: Ohkkay, agreed

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkay

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1688,26 +1694,46 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
...
-		/* Check mode */
-
 		/* Create the directory */
-
-		ret = smartfs_createentry(fs, parentdirsector, filename,
-								  SMARTFS_DIRENT_TYPE_DIR, mode, &entry, 0xFFFF, NULL);
+		/* Allocate new data sector for entry */
+		ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_DIR, NULL);
 		if (ret != OK) {
+			fdbg("Failed to allocate data sector for entry\n");
```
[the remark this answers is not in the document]
amogh-samsung: Comments aligned

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1889,7 +1913,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 				/* Test the next entry */
 
 				direntry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-				if (((direntry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((direntry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
+				if ((direntry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((direntry->flags & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)) == (~SMARTFS_ERASEDSTATE_16BIT & (SMARTFS_DIRENT_EMPTY | SMARTFS_DIRENT_ACTIVE)))) {
```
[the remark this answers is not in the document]
amogh-samsung: Corrected the checking condition,
but still kept all three conditions combined in one.
I don't think we should have 3 separate if constructs, lots of extra lines.
However, I added a comment to support the checks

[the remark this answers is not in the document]
amogh-samsung: This loop is required for now.
Assume a request like
`mv /mnt/test/file1.txt /mnt/test_2`

finddirentry will return the parent directory as test_2 as the relative path ends here.
However, we still have to check if the file file1.txt already exists in test_2.
This is what this while loop is doing.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: The logic is slightly different.
```
if (oldparentdirsector == newparentdirsector)
    if (CRC is not enabled && we can find an erased entry)
         Do not use old entry to overwrite
    else
         Overwrite old entry
else
     find_availableentry()
     if (not found)
          createentry()
     writeentry()
     invalidateentry()
```

os/fs/smartfs/smartfs_utils.c
amogh-samsung: **sf** buffer changes will be an improvement as discussed.
Will be done in a separate PR however.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Added TODO comment

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes `direntry` object is also populated with same attributes for returning the entry that is written.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Let's align this code with all other functions.
We'll use a `smartfs_entry_header_s` structure for making changes to the entry flags.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: ah ok I confused for a while

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: No, I don't agree. We can overwrite only below condition
```
if ((old == new) && (USE_SECTOR_BUFFER == TRUE))
```
otherwise we should find entry .
That's why I commented many time that find_availableentry should return first invalid entry if CONFIG_SMARTFS_USE_SECTOR_BUFFER is enabled, of course return emptry entry if CONFIG_SMARTFS_USE_SECTOR_BUFFER is not enabled.

os/fs/smartfs/smartfs_smart.c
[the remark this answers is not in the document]
Taejun-Kwon: > Corrected the checking condition,
> but still kept all three conditions combined in one.
> I don't think we should have 3 separate if constructs, lots of extra lines.
> However, I added a comment to support the checks

Suggested it point of enhanced readability. it doesn't affect to binary size much.
Anyway, if it is active and not empty, will be returned. why you changed condition? (adding ERASESTATE is acceptable.)
> This loop is required for now.
> Assume a request like
> `mv /mnt/test/file1.txt /mnt/test_2`
> 
> finddirentry will return the parent directory as test_2 as the relative path ends here.
> However, we still have to check if the file file1.txt already exists in test_2.
> This is what this while loop is doing.

ok I see.

os/fs/smartfs/smartfs_smart.c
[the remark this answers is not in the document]
Taejun-Kwon: > I think you had given a comment earlier to handle all error cases under one if structure.
> 
> ```
> if (ret != OK i.e. ERROR) {
>     then if (ret == -ENOENT)
>     {}
>     else
>     {}
> }
> ```

yes I remember but refactored code changed my point. 
But I still thinking smartfs_find_availableentry should find entry anyway,

os/fs/smartfs/smartfs_smart.c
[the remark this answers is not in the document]
Taejun-Kwon: I mean my comment regarding if-else in smartfs_open can be applied here.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: So we agree on using the same entry for overwriting when CRC is enabled.

When CRC is not enabled: -
Case 1 - We can find an erased entry in the sector (not invalidated)
My logic - use new entry
Your logic - use new entry
So we are at the same logic.

Case 2 - We can find an invalidated entry
My logic - use the same entry for overwriting
`Read the sector into rwbuffer
Make changes to the buffer
Write from the buffer back to the MTD`
So we used 1 new sector while writing to MTD because of changes in bits from 0 to 1
And only 1 MTD WRITE operation.
Old sector is released.

Your logic - Use the invalidated entry that is found
`Find_availableentry
read the parent sector into rwbuffer
Make changes to buffer contents
Write back buffer to MTD
read old entry into buffer
Mark flags invalid and write back to MTD`

So 1 new sector is used while writing to the invalidated entry because bits are changed from 0 to 1 (same as my logic)
But MTD_WRITE is called 2 times (one for writing entry, one for invalidating old entry)
So one extra MTD_WRITE that has to be journaled.

Case 3 - no available entry
Similar to case 2
My logic uses 1 new sector and 1 MTD_WRITE
Your logic will use 1 new sector but 2 MTD_WRITES

Taejun-Kwon:
```
ret = smartfs_finddirentry(fs, &oldentry, oldrelpath, &oldparentdirsector, &oldfilename);
...
ret = smartfs_finddirentry(fs, &newentry, newrelpath, &newparentdirsector, &newfilename);

#if CONFIG_SMARTFS_USE_SECTOR_BUFFER
if (oldparentdirsector == newparentdirsector) {
	overwrite_old_sector; // overwrite entry of source path.
} else 
#endif
	{
		smartfs_find_availableentry(); // This had better return entry. If crc enabled, first inactive entry, other wise first empty entry.
		if (ret == -ENOENT) {
			smartfs_createentry(); // this had better called inside of smartfs_find_availableentry. This will create  new sector for entry and return sector & offset.
		} else {
			if (ret != OK) {
				fdbg("error.....");
				goto error;
			}
			fvdbg("Entry found in parent sector : %d, offset : %d\n");
		}
		smartfs_writeentry();
		invalidate_entry(oldentry);
	}
}

File: os/fs/smartfs/smartfs_smart.c (hunk 50 above)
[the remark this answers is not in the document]
Comment (amogh-samsung) [reply os/fs/smartfs/smartfs_smart.c:291]: I'll add call to createentry from within find_availableentry as discussed in a separate PR
So this comment will become obsolete.

File: os/fs/smartfs/smartfs_smart.c (hunk 52 above)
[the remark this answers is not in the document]
Comment (amogh-samsung) [reply os/fs/smartfs/smartfs_smart.c:1701]: You mean the comment for the next if-else construct after find_availableentry()
Line 1706.

As I replied to the previous comment, your suggestion of handling all error cases together under
`if (ret != OK)` makes more sense

Anyways, this comment and if-else construct will become obsolete if we include call to createentry inside find_availableentry as per your suggestion in future PRs.

File: os/fs/smartfs/smartfs_smart.c (hunk 53 above)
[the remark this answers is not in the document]
Comment (amogh-samsung) [reply os/fs/smartfs/smartfs_smart.c]: > Suggested it point of enhanced readability. it doesn't affect to binary size much.
Anyway, if it is active and not empty, will be returned. why you changed condition? (adding ERASESTATE is acceptable.)

I added all the three conditions that you mentioned but in one statement.
I don't see the point in adding 3 if statements

> if (direntry->flag == SMARTFS_ERASEDSTATE_16BIT) {
     offset += entrysize; //if we change while to for loop, we can remove it.
     continue; 
}
if ((direntry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) {
   offset += entrysize;
   continue;
}
if (((direntry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE)) {
    offset += entrysize;
    continue;
}

I think it is better to have all 3 conditions in one if statement with a comment to describe it.
This is what I have done^.

File: os/fs/smartfs/smartfs_smart.c (hunk 24 above)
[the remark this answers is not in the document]
Comment (amogh-samsung) [reply os/fs/smartfs/smartfs_smart.c]: TODO has been added in smartfs_writeentry to take care of sf buffer operations.
Will be done ion upcoming PRs.

**Comment by Taejun-Kwon** [comment]
Line 979, you can change to below as you changed before.

File: os/fs/smartfs/smartfs_smart.c [hunk 54]
```diff
@@ -652,10 +676,7 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 			header->used[1] = (uint8_t)(tmp >> 8);
 		}
 #endif
-		readwrite.offset = offsetof(struct smartfs_chain_header_s, used);
-		readwrite.count = sizeof(uint16_t);
-		readwrite.buffer = (uint8_t *)&fs->fs_rwbuffer[readwrite.offset];
-
+		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
```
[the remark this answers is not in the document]
amogh-samsung: Previous code sets readwrite.offset and then uses it's value for readwrite.buffer
It is easy to understand as it is in two different steps.

Now, if we are setting readwrite.offset in smartfs_setbuffer and passing it as a parameter to the same function simultaneously, it becomes a bit confusing to understand.

So I think this implementation is fine.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1317,6 +1405,132 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+ *  offset - offset of entry in parent directory sector
+ *
+ * Returned Values:
+ *  OK - on sucessfully invalidating the entry.
+ *  ERROR - appropriate error code otherwise
+ *
+ ****************************************************************************/
+
+/* TODO IF entry is empty after invalid, we'd better release it. To do this entry needed as a parameter...*/
+int smartfs_invalidateentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, uint16_t offset)
```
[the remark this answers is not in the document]
amogh-samsung: Yes, thanks

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1699,10 +1895,7 @@ int smartfs_truncatefile(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *e
 	}
 #endif
 
-	ret = OK;
-
-errout:
-	return ret;
+	return OK;
```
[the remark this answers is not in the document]
amogh-samsung: Yes, thanks.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1317,6 +1405,132 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+
+/* TODO IF entry is empty after invalid, we'd better release it. To do this entry needed as a parameter...*/
+int smartfs_invalidateentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, uint16_t offset)
+{
+	int ret;
+	struct smart_read_write_s readwrite;
+	struct smartfs_entry_header_s *entry;
+
+	smartfs_setbuffer(&readwrite, parentdirsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
+	ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
```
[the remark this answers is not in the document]
amogh-samsung: MTD_READ in most of the functions serves as a preliminary check to verify that the contents of the sector are fine.
While reading CRC is validated on the sector content.

Let's read just 40 bytes instead = 1 entry only?
What would you say?

amogh-samsung:
> Line 979, you can change to below as you changed before.
Line 1607
smartfs_setbuffer(&readwrite, entry->dsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
Line 1622
smartfs_setbuffer(&readwrite, sector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), header->nextsector);

The above mentioned lines have not been converted to smartfs_setbuffer format.
All of these are inside loops where `readwrite.offset, count and buffer` are assigned one time outside the loop.
`readwrite.logsector` is changed in every loop iteration.
So we would have to call smartfs_setbuffer for all 4 attributes inside the loop even though we have to change only 1.
I think that would be a waste.

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
amogh-samsung: Minimized to read 2 bytes only - for current entry flags.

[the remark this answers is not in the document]
Taejun-Kwon: Before call smartfs_invalidateentry, as I know it already check validation of entry.
So if it is 0xFFFF, we will not invalidate it. this is always used to delete exist entry.
So we don't have to read anything.
If we call this function without any checking, then we'd better read entire of sector, actually.
Please check this.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1317,6 +1401,132 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+#else
+#if CONFIG_SMARTFS_ALIGNED_ACCESS
+	smartfs_wrle16(entry_flags, smartfs_rdle16(entry_flags) | SMARTFS_DIRENT_ACTIVE);
+#else
+	*entry_flags |= SMARTFS_DIRENT_ACTIVE;
+#endif
+#endif
+	/* Now write the updated flags back to the device */
+
+	smartfs_setbuffer(&readwrite, parentdirsector, offset, sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[0]);
```
Taejun-Kwon, line 1523: If you want to use fs_rwbuffer instead of simple variable, &flag, I think you'd better use offset instead of 0.
Because we need to test this more, actually every case is not clear now, and another place using like that, we'd better align it.

PR #4377, opened 2020-07-13
kernel/semaphore: Initialize semaphore with sem_init instead of SEM_INITIALIZER

jeongarmy:
All kernel semaphores are registered to a list of kernel semaphores, g_sem_list for fault recovery.
So they should call sem_register by calling sem_init instead of SEM_INITIALIZER.

### DIFF
--- a/os/arch/xtensa/src/esp32/spi_flash/include/esp_partition.h
+++ b/os/arch/xtensa/src/esp32/spi_flash/include/esp_partition.h
@@ -175,6 +175,11 @@ const esp_partition_t *esp_partition_get(esp_partition_iterator_t iterator);
 esp_partition_iterator_t esp_partition_next(esp_partition_iterator_t iterator);
 
 /**
+ * @brief Initialize a mutex for partition access
+ */
+void esp_partition_initialize(void);
+
+/**
  * @brief Release partition iterator
  *
  * @param iterator Iterator obtained using esp_partition_find. Must be non-NULL.
--- a/os/arch/xtensa/src/esp32/spi_flash/include/esp_partition.h
+++ b/os/arch/xtensa/src/esp32/spi_flash/include/esp_partition.h
@@ -175,7 +175,7 @@ const esp_partition_t *esp_partition_get(esp_partition_iterator_t iterator);
 esp_partition_iterator_t esp_partition_next(esp_partition_iterator_t iterator);
 
 /**
- * @brief Initialize a mutex for partition access
+ * @brief Initialize a semaphore for partition access
  */
 void esp_partition_initialize(void);
 
--- a/os/arch/xtensa/src/esp32/spi_flash/partition.c
+++ b/os/arch/xtensa/src/esp32/spi_flash/partition.c
@@ -75,6 +75,8 @@
 #include <stdio.h>
 #include <pthread.h>
 
+#include <tinyara/sched.h>
+
 #include "esp_flash_partitions.h"
 #include "esp_attr.h"
 #include "esp_flash_data_types.h"
@@ -105,7 +107,7 @@ static esp_partition_iterator_opaque_t *iterator_create(esp_partition_type_t typ
 static esp_err_t load_partitions(void);
 
 static SLIST_HEAD(partition_list_head_, partition_list_item_) s_partition_list = SLIST_HEAD_INITIALIZER(s_partition_list);
-static pthread_mutex_t s_partition_list_lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_mutex_t s_partition_list_lock;
 
 esp_partition_iterator_t esp_partition_find(esp_partition_type_t type, esp_partition_subtype_t subtype, const char *label)
 {
@@ -232,6 +234,13 @@ static esp_err_t load_partitions(void)
 	}
 	spi_flash_munmap(handle);
 	return ESP_OK;
+}
+
+void esp_partition_initialize(void)
+{
+	/* Initialize a mutex for partition access */
+	sem_init(&s_partition_list_lock.sem, 0, 1);
+	s_partition_list_lock.pid = INVALID_PROCESS_ID;
 }
 
 void esp_partition_iterator_release(esp_partition_iterator_t iterator)
--- a/os/arch/xtensa/src/esp32/spi_flash/partition.c
+++ b/os/arch/xtensa/src/esp32/spi_flash/partition.c
@@ -73,9 +73,7 @@
 #include <assert.h>
 #include <string.h>
 #include <stdio.h>
-#include <pthread.h>
-
-#include <tinyara/sched.h>
+#include <semaphore.h>
 
 #include "esp_flash_partitions.h"
 #include "esp_attr.h"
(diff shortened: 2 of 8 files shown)


os/net/netmgr/netdev_mgr_internal.c
```diff
@@ -46,7 +46,7 @@ static int g_eth_idx = 0;
  * and application doesn't require write operation to netdev
  * so if there is no write operation then protection wouldn't be required
  */
-sem_t g_netdev_lock = SEM_INITIALIZER(1);
+sem_t g_netdev_lock;
```
sijupunnnoose, line 49: Is it required only in these 2 places? I observed in some stm files as well. How about PTHREAD_MUTEX_INITIALIZER?

jeongchanKim, line 49: I think this variable is used only in this file. If I'm right, please add static.

jeongarmy: Yes, I checked PTHREAD_MUTEX_INITIALIZER also.
There are 2 places more in stm and esp files. I will modify them.

os/arch/xtensa/src/esp32/spi_flash/partition.c
```diff
@@ -234,6 +234,13 @@ static esp_err_t load_partitions(void)
 	return ESP_OK;
 }
 
+void esp_partition_initialize(void)
+{
+	/* Initialize a mutex for partition access */
+	sem_init(&s_partition_list_lock.sem, 0, 1);
+	s_partition_list_lock.pid = -1;
```
jeongchanKim, line 241: How about using `INVALID_PROCESS_ID` instead of `-1`?

os/arch/xtensa/src/esp32/spi_flash/include/esp_partition.h
```diff
@@ -174,6 +174,11 @@ const esp_partition_t *esp_partition_get(esp_partition_iterator_t iterator);
  */
 esp_partition_iterator_t esp_partition_next(esp_partition_iterator_t iterator);
 
+/**
+ * @brief Initialize a mutex for partition access
+ */
+void esp_partition_initialize(void);
```
jeongchanKim, line 180: This is not related to this commit. Please separate the commit.

os/arch/xtensa/src/esp32/spi_flash/esp32_flash.c
```diff
@@ -36,6 +36,8 @@
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/fs/mtd.h>
 
+#include "esp_partition.h"
```
jeongchanKim, line 39: Why did you include this header?

os/arch/xtensa/src/esp32/spi_flash/partition.c
```diff
@@ -234,6 +236,13 @@ static esp_err_t load_partitions(void)
 	return ESP_OK;
 }
 
+void esp_partition_initialize(void)
+{
+	/* Initialize a mutex for partition access */
+	sem_init(&s_partition_list_lock.sem, 0, 1);
```
jeongchanKim, line 242: We have `pthread_mutex_init()`. How about using this?

os/board/stm32l4r9ai-disco/src/stm32_boot.c
```diff
@@ -73,6 +73,9 @@ void board_initialize(void)
     board_led_initialize();
 #endif
 
+  /* Initialize for flash access */
+  stm32l4_flash_initialize();
```
Taejun-Kwon, line 77: Mostly, flash should be initialized after SPI, because flash uses spi as a bus.
It already has stm32_exflash_initialize, please check it and use it if possible.

os/board/stm32l4r9ai-disco/include/stm32l4r9i_discovery.h
```diff
@@ -359,6 +359,8 @@ extern void stm32_touch_printf_coord(void);
 
 extern void stm32_dsi_refresh(void);
 extern void up_hal_timer_initialize(void);
+
+extern void stm32l4_flash_initialize(void);
```
Taejun-Kwon, line 363: May I know reason why you add this?

os/board/esp32_devkit/src/esp32_boot.c
```diff
@@ -234,6 +237,9 @@ void board_initialize(void)
 {
 	/* Perform board-specific initialization */
 	(void)esp32_bringup();
+#ifdef CONFIG_ESP32_SPIFLASH
+	esp_partition_initialize();
```
Taejun-Kwon, line 241: Same as above, it should be done after spi initialize but I think it has only 1 spi port, so it doesn't initialize spi explicitly.
so instead of this, you can use up_flashinitialize in esp32_flash.c, I think.
This assume that never access flash directly before MTD driver initialize.

os/board/stm32l4r9ai-disco/src/stm32_boot.c
Taejun-Kwon: Otherwise, up_fliashinitialize in stm32_flash.c

jeongarmy: A semaphore initialized by stm32l4_flash_initiailize is used in up_progmem functions.
But in mtd_initialize,
#ifdef CONFIG_MTD_PROGMEM
  progmem_initialize -> common code
#else
  up_flashinitialize() -> board_specific
We cannot put stm32l4_flash_initiailize() in up_flashinitialize so I called this function in board_initialize before calling mtd_partition.

os/board/stm32l4r9ai-disco/include/stm32l4r9i_discovery.h
jeongarmy: I added it to call this function in stm32_boot.c

os/arch/xtensa/src/esp32/spi_flash/partition.c
```diff
@@ -75,6 +75,8 @@
 #include <stdio.h>
 #include <pthread.h>
```
sunghan-chang, line 76: This may not be necessary because of removing pthread_mutex functions.

os/arch/xtensa/src/esp32/spi_flash/partition.c
```diff
@@ -75,6 +75,8 @@
...
 
+#include <tinyara/sched.h>
```
sunghan-chang, line 78: Is this necessary? not <semaphore.h>?

sunghan-chang:
@jeongarmy Please find build warnings as below
```
CC:  chip/esp32_spi.c
chip/esp32_spi.c: In function 'up_spiinitialize':
chip/esp32_spi.c:1590:12: warning: passing argument 1 of 'sem_init' discards 'volatile' qualifier from pointer target type [-Wdiscarded-qualifiers]
   sem_init(&g_spi_dma.dma_lock, 0, 1);
            ^
In file included from /root/tizenrt/os/include/sys/types.h:66:0,
                 from /root/tizenrt/os/include/stdio.h:76,
                 from chip/esp32_spi.c:24:
/root/tizenrt/os/include/semaphore.h:198:5: note: expected 'sem_t * {aka struct sem_s *}' but argument is of type 'volatile sem_t * {aka volatile struct sem_s *}'
 int sem_init(FAR sem_t *sem, int pshared, unsigned int value);
     ^
```

PR #4379, opened 2020-07-13
os/fs/smartfs : Add logic to get blockdriver's physical sector from b…

Taejun-Kwon:
…lock device.

To recover isolated sector due to power failure, FS should know which logical sector is
mapped to active physical sector

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -1490,6 +1490,11 @@ static int smartfs_bind(FAR struct inode *blkdriver, const void *data, void **ha
 	}
 
 	*handle = (void *)fs;
+	
+	ret = smartfs_sector_recovery(fs);
+	if (ret != 0) {
+		goto error_with_semaphore;
+	}
 
 	smartfs_semgive(fs);
 	return ret;
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -1665,4 +1665,58 @@ struct smartfs_mountpt_s *smartfs_get_first_mount(void)
 {
 	return g_mounthead;
 }
+
+/****************************************************************************
+ * Name: smartfs_sector_recovery
+ *
+ * Description: Recover Isolated sector that is not able to access in smartfs.
+ *              Because of power off, Isolated sector can be exist.
+ *
+ ****************************************************************************/
+
+int smartfs_sector_recovery(struct smartfs_mountpt_s *fs)
+{
+	int ret;
+	long sector;
+	/* Alloc Logical Map */
+	uint8_t *map = (uint8_t*)kmm_malloc(sizeof(uint8_t*) * fs->fs_llformat.nsectors);
+
+	if (map == NULL) {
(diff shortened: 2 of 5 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1665,4 +1665,58 @@ struct smartfs_mountpt_s *smartfs_get_first_mount(void)
...
+	/* If any of logical sector is mapped to physical block it means active block, Mark it */
+	for (sector = SMARTFS_ROOT_DIR_SECTOR; sector < fs->fs_llformat.nsectors; sector++) {
+		ret = FS_IOCTL(fs, BIOC_FIBMAP, (unsigned long)sector);
+		if (ret < 0) {
+			fdbg("Error %d\n", ret);
+			return ret;
+		}
+
+		if (ret != 0xFFFF) {
+			map[sector / 8] |= (1 << (7 - (sector % 8)));
```
jsdosa, line 1698: How about writing a comment?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1665,4 +1665,58 @@ struct smartfs_mountpt_s *smartfs_get_first_mount(void)
...
+		return -ENOMEM;
+	}
+	memset(map, 0, fs->fs_llformat.nsectors);
+
+	/* If any of logical sector is mapped to physical block it means active block, Mark it */
+	for (sector = SMARTFS_ROOT_DIR_SECTOR; sector < fs->fs_llformat.nsectors; sector++) {
+		ret = FS_IOCTL(fs, BIOC_FIBMAP, (unsigned long)sector);
+		if (ret < 0) {
+			fdbg("Error %d\n", ret);
+			return ret;
```
jsdosa, line 1694: Is there no need to release memory for map?

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: Comments in line 1681 & 1689 are enough, I think

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: Thanks. fixed.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1490,6 +1490,11 @@ static int smartfs_bind(FAR struct inode *blkdriver, const void *data, void **ha
 	}
 
 	*handle = (void *)fs;
+	
+	ret = smartfs_sector_recovery(fs);
+	if (ret != 0) {
```
jeongchanKim, line 1495: `smartfs_sector_recovery` returns `OK` or errno. How about comparing with `OK` instead of `0`?

Taejun-Kwon: Yes you're right. I already know that, some of in this file use '0' instead of OK, I just aligned with above line's code style(line 1488).
it will be cleaned totally, after https://github.com/Samsung/TizenRT/pull/4376 merged, may in someday =D

PR #4381, opened 2020-07-14
se/ameba: Resolve compile error, and index problem for SE non-secure

ziyik:
1. Resolve compile error by adding include semaphore.h
2. Fix key index convert, array index start from 0, key index start from 1
3. Added rtl_write_factory_cert_wrapper() API for future reference

### DIFF
--- a/os/se/ameba/rtl_security_api_wrapper.c
+++ b/os/se/ameba/rtl_security_api_wrapper.c
@@ -16,10 +16,10 @@
 *
 ***************************************************************************************************/
 
-#include <stdint.h>
-#include <stdio.h>
+#include <sys/types.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdint.h>
 #include <tinyara/kmalloc.h>
 
 #include "flash_api.h"
@@ -79,7 +79,7 @@ static uint8_t crypto_inited = 0;
 /**
  * Read secure efuse
  */
-static void rtl_ReadEfuse(u32 addr, u8 size, u8* key)
+static void rtl_ReadEfuse(uint32_t addr, uint8_t size, uint8_t *key)
 {
 	uint32_t Idx = 0;
 	uint32_t CtrlSetting = HAL_READ32(SYSTEM_CTRL_BASE, REG_HS_EFUSE_CTRL1_S);
@@ -119,7 +119,7 @@ uint32_t rtl_cryptoEngine_init_wrapper(void)
 /**
  * AES Crpyto Wrapper
  */
-uint32_t rtl_cryptoAES_ecb_wrapper(uint8_t* key, uint32_t keylen, unsigned char* message, uint32_t msglen, unsigned char* pResult, uint8_t mode)
+uint32_t rtl_cryptoAES_ecb_wrapper(uint8_t *key, uint32_t keylen, unsigned char *message, uint32_t msglen, unsigned char *pResult, uint8_t mode)
 {
 	uint32_t ret = RTL_HAL_SUCCESS;
 
@@ -145,7 +145,7 @@ uint32_t rtl_cryptoAES_ecb_wrapper(uint8_t* key, uint32_t keylen, unsigned char*
 	return ret;
 }
 
-uint32_t rtl_cryptoAES_cbc_wrapper(uint8_t* key, uint32_t keylen, unsigned char* message, uint32_t msglen, unsigned char* pResult, uint8_t mode)
+uint32_t rtl_cryptoAES_cbc_wrapper(uint8_t *key, uint32_t keylen, unsigned char *message, uint32_t msglen, unsigned char *pResult, uint8_t mode)
 {
 	uint32_t ret = RTL_HAL_SUCCESS;
 	unsigned char iv[16];
@@ -172,7 +172,7 @@ uint32_t rtl_cryptoAES_cbc_wrapper(uint8_t* key, uint32_t keylen, unsigned char*
 	return ret;
 }
 
-uint32_t rtl_cryptoAES_ctr_wrapper(uint8_t* key, uint32_t keylen, unsigned char* message, uint32_t msglen, unsigned char* pResult, uint8_t mode)
+uint32_t rtl_cryptoAES_ctr_wrapper(uint8_t *key, uint32_t keylen, unsigned char *message, uint32_t msglen, unsigned char *pResult, uint8_t mode)
 {
 	uint32_t ret = RTL_HAL_SUCCESS;
 	unsigned char iv[16];
--- a/os/se/ameba/rtl_security_api_wrapper.c
+++ b/os/se/ameba/rtl_security_api_wrapper.c
@@ -239,6 +239,37 @@ uint32_t rtl_read_factory_cert_wrapper(hal_data *data)
 	return ret;
 }
 
+uint32_t rtl_write_factory_cert_wrapper(hal_data *data)
+{
+	uint32_t ret;
+	uint8_t buf_length[4];
+	flash_t flash;
+	uint32_t data_len = data->data_len;
+
+	uint8_t	*buf_flash = (uint8_t *) malloc(data_len + 4);
+
+	if (buf_flash) {
+		itoa(data_len, buf_length, 10);
+		memcpy(buf_flash, buf_length, 4);
+		memcpy(buf_flash + 4, data->data, data->data_len);
+
+		device_mutex_lock(RT_DEV_LOCK_FLASH);
+		flash_erase_sector(&flash, FACTORY_CERT_ADDR);
+		flash_stream_write(&flash, FACTORY_CERT_ADDR, data_len + 4, buf_flash);
+		device_mutex_unlock(RT_DEV_LOCK_FLASH);
+
+		ret = RTL_HAL_SUCCESS;
+	} else {
+		ret = RTL_HAL_FAIL;
+	}
+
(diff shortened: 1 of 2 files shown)


os/se/ameba/rtl_security_api_wrapper.c
```diff
@@ -239,6 +239,37 @@ uint32_t rtl_read_factory_cert_wrapper(hal_data *data)
...
 }
 
+uint32_t rtl_write_factory_cert_wrapper(hal_data *data)
+{
+	uint32_t ret;
+	uint8_t buf_length[4];
+	flash_t flash;
+	uint32_t data_len = data->data_len;
+
+	uint8_t	*buf_flash = (uint8_t *) malloc(data_len + 4);
```
jeongchanKim, line 249: Please remove space after type casting. 
And please use kmm_malloc instead of malloc, because this module is in kernel side.

os/se/ameba/security_ameba_wrapper.c
```diff
@@ -270,6 +272,18 @@ static void* my_calloc(size_t nelements, size_t elementSize)
 	return ptr;
 }
 
+static uint32_t rtl_index_convt(uint32_t key_idx)
+{
+	uint32_t ret_key_idx = KEY_STORAGE_INDEX_MAX;
+
+	/* Because Ameba SE storing Key in array, array start from Zero */
+	if (key_idx != 0) {	/* Index not Zero */
+		/* Valid  Inded are 0 - 31, Total 32 Index slot */
```
jeongchanKim, line 281: Please fix `Inded` to `Index`.

os/se/ameba/rtl_security_api_wrapper.c
```diff
@@ -239,6 +239,37 @@ uint32_t rtl_read_factory_cert_wrapper(hal_data *data)
...
+		itoa(data_len, buf_length, 10);
+		memcpy(buf_flash, buf_length, 4);
+		memcpy(buf_flash + 4, data->data, data->data_len);
+
+		device_mutex_lock(RT_DEV_LOCK_FLASH);
+		flash_erase_sector(&flash, FACTORY_CERT_ADDR);
+		flash_stream_write(&flash, FACTORY_CERT_ADDR, data_len + 4, buf_flash);
+		device_mutex_unlock(RT_DEV_LOCK_FLASH);
+
+		ret = RTL_HAL_SUCCESS;
```
jeongchanKim, line 261: I think we can move line 267(free) at the end of this if condition.
```
if (buf_flash) {
   ...
   ret = RTL_HAL_SUCCESS;
   free(buf_flash);
}
```

os/se/ameba/rtl_security_api_wrapper.c
sangam-swami, line 249: if (buf_flash == NULL) {
		RTL_HAL_FAIL;
}
then no need to check if or else for below code

os/se/ameba/security_ameba_wrapper.c
```diff
@@ -399,7 +419,8 @@ int ameba_hal_set_key(hal_key_type mode, uint32_t key_idx, hal_data *key, hal_da
 	if (prikey != NULL) {	/* If private key is pass in */
 		return HAL_NOT_SUPPORTED;
 	}
-	if (key_idx > KEY_STORAGE_INDEX_MAX) {	/* Index Out of Range */
+	key_idx = rtl_index_convt(key_idx);
```
sangam-swami, line 422: is key_idx returned to calling function?
It is passed as argument and calculating value here

os/se/ameba/security_ameba_wrapper.c
```diff
@@ -735,6 +774,11 @@ int ameba_hal_get_hmac(hal_hmac_type mode, hal_data *input, uint32_t key_idx, ha
 		return ret;
 	}
 
+	key_idx = rtl_index_convt(key_idx);
+	if (key_idx >= KEY_STORAGE_INDEX_MAX) {	/* Index Out of Range */
```
sangam-swami, line 778: you can add debug error logs here?

os/se/ameba/security_ameba_wrapper.c
sangam-swami: shall we use macro for this function?

os/se/ameba/security_ameba_wrapper.c
ziyik: Okay, i will add it like other API.

os/se/ameba/security_ameba_wrapper.c
ziyik: Yes, it is return to calling function after convert the index.

os/se/ameba/rtl_security_api_wrapper.c
ziyik: Modified.

os/se/ameba/security_ameba_wrapper.c
ziyik: So I prefer to use as static function, will it cause any problem?

os/se/ameba/rtl_security_api_wrapper.c
ziyik: Modified.

ziyik: Modified.

os/se/ameba/rtl_security_api_wrapper.c
```diff
@@ -239,6 +240,34 @@ uint32_t rtl_read_factory_cert_wrapper(hal_data *data)
...
+		free(buf_flash);
```
sunghan-chang, line 263: free should be changed to kmm_free for the same reason of kmm_malloc.

os/se/ameba/security_ameba_wrapper.c
```diff
@@ -714,7 +753,7 @@ int ameba_hal_get_hash(hal_hash_type mode, hal_data *input, hal_data *hash)
 		return HAL_NOT_SUPPORTED;
 	}
 
-	len = strlen((char*) output);
+	len = strlen((char*)output);
```
jeongchanKim, line 756: Please add space between `char` and `*`.

ziyik: Modified.

os/se/ameba/rtl_security_api_wrapper.c
ziyik: Sorry, din think about it on last commit, Modified.

os/se/ameba/rtl_security_api_wrapper.c
```diff
@@ -283,13 +312,70 @@ uint32_t rtl_read_factory_key_wrapper(hal_data *data)
...
+		kmm_free(buf_plain);
+	}
+	if (buf_crypt != NULL) {
+		kmm_free(buf_crypt);
+	}
+
+	return ret;
+}
+
+uint32_t rtl_write_factory_key_wrapper(hal_data *data)
```
sunghan-chang, line 327: There is no code to call these new functions, rtl_write_factory_key_wrapper and rtl_write_factory_cert_wrapper. Could you let me know the purpose (or plan) of them?

ziyik: rtl_write_factory_cert_wrapper() and rtl_write_factory_key_wrapper() is for future reference. When want to write factory key or cert can directly use these API.

ziyik:
Any other comment or suggestion on this PR?

os/se/ameba/rtl_security_api_wrapper.c
```diff
@@ -16,10 +16,7 @@
 *
 ***************************************************************************************************/
 
-#include <stdint.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
+#include <tinyara/kmalloc.h>
```
sunghan-chang, line 19: memcpy, memset requires <string.h>
NULL does <sys/types.h>
uint32_t does <stdint.h>
itoa does <stdlib.h>

sorry for late

ziyik: I have a question about this.
The reason i remove these header is because I received comment "stdint.h" it is not need on PR#4385.
So I removed them and it have no compile error and it working fine, should I add them back?

sunghan-chang: He left the comment to remove including <string.h>. At that PR, memcpy, memset and other APIs which are defined in string.h are not used. That's why Jeongchan left the comment. But in this PR, you use memcpy and memset so that you should include string.h.

sunghan-chang: You should include the headers for what you use in a file. That's not TizenRT specific rule. As you said, it could work because one of header includes that inside. But it makes strong dependency between them. <tinyara/kmalloc.h> is made to make a dependency with this file. It makes sometimes it works, but sometimes it doesn't. It's not right usages.

ziyik: Thank you for the explanation, I get the point now. Sorry for the misunderstanding.
I will include the necessary header base on usage.

ziyik: I have corrected the included header, please let me know if there it is any other problem.
Thank for your help.

sunghan-chang:
@pillip8282 Could you review this?

sunghan-chang (approved):

PR #4383, opened 2020-07-14
[DO NOT REVIEW] os/fs/smartfs: Add recovery logic

Taejun-Kwon:
1. Add validation check of logical sector
2. Add routine which checks 2 sectors point same firstsector

### DIFF
--- a/os/fs/smartfs/smartfs.h
+++ b/os/fs/smartfs/smartfs.h
@@ -347,6 +347,9 @@ struct smartfs_mountpt_s {
 	uint8_t *fs_chunk_buffer;
 #endif
 	uint8_t fs_rootsector;		/* Root directory sector num */
+#ifndef CONFIG_UTC
+	uint32_t entry_seq;
+#endif
 };
 
 /****************************************************************************
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -87,9 +87,58 @@
 #define SET8BITSFROMMSB 255		//Binary: 0b11111111
 #define NEG8BIT(A) ((uint8_t)(~A & 0x000000FF))
 #endif
+
+/****************************************************************************
+ * Private Definitions
+ ****************************************************************************/
+ 
+#if defined(CONFIG_SMARTFS_MULTI_ROOT_DIRS) || \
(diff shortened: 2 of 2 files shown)


os/fs/smartfs/smartfs.h
```diff
@@ -347,6 +347,9 @@ struct smartfs_mountpt_s {
 	uint8_t *fs_chunk_buffer;
 #endif
 	uint8_t fs_rootsector;		/* Root directory sector num */
+#ifndef CONFIG_UTC
+	uint32_t entry_seq;
```
amogh-samsung, line 351: I though you had suggested that smartfs_entry_s structure will hold sequence number.
The sequence number for that particular entry.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -549,6 +572,9 @@ int smartfs_mount(struct smartfs_mountpt_s *fs, bool writeable)
 	/* We did it! */
 
 	fs->fs_mounted = TRUE;
+#ifndef CONFIG_UTC
+	fs->entry_seq = 0;
```
amogh-samsung, line 576: So global sequence number initialized to 0 when we mount smartfs...

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1217,11 +1243,19 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector,
...
+	smartfs_wrle32(&entry->utc, time(NULL));
+#else
+	smartfs_wrle32(&entry->utc, fs->entry_seq++);
+#endif
 #else
 	entry->firstsector = nextsector;
+#ifdef CONFIG_UTC
 	entry->utc = time(NULL);
+#else
+	entry->utc = fs->entry_seq++;
```
amogh-samsung, line 1257: Suppose smartfs is unmounted and remounted (device resets after power off).
fs->entry_seq is reset to 0.

Smartfs partition already has active entries with sequence numbers say 45, 57, 89

Now, these entries are to be replaced (due to file operations or other requests).
The new entries will have entry->utc = 0, 1, 2, 3....... which are smaller than old sequence numbers.

How will we recover base don this? We expect a higher sequence number to be the latest entry.

Taejun-Kwon: It will updated during fs_recover.

Taejun-Kwon:
Ok I will close this.

PR #4386, opened 2020-07-16
kernel/binary_manager: Free a semaphore holder directly when priority inheritance is disabled

jeongarmy:
All semphoare holders should be freed by binary_manager_release_binary_sem.
If priority inheritance is disabled, a holder is freed directly.
If not, it is freed after restoring priorities of threads holding semaphore in sem_restorebaseprio.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -560,7 +560,7 @@ void binary_manager_release_binary_sem(int bin_idx)
 				if (holder && holder->htcb && holder->htcb->group && holder->htcb->group->tg_binidx == bin_idx) {
 					/* Increase semcount and release itself from holder */
 					sem->semcount++;
-					sem_freeholder(sem, holder);
+
 					/* And after releasing the kernel sem, there can be a task which waits that sem. So unblock the waiting task. */
 					sem_unblock_task(sem, holder->htcb);
 				}
--- a/os/kernel/semaphore/sem_post.c
+++ b/os/kernel/semaphore/sem_post.c
@@ -96,6 +96,10 @@
 void sem_unblock_task(sem_t *sem, struct tcb_s *htcb)
 {
 	struct tcb_s *stcb = NULL;
+#ifdef SAVE_SEM_HOLDER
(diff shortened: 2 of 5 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -560,7 +560,12 @@ void binary_manager_release_binary_sem(int bin_idx)
 				if (holder && holder->htcb && holder->htcb->group && holder->htcb->group->tg_binidx == bin_idx) {
 					/* Increase semcount and release itself from holder */
 					sem->semcount++;
-					sem_freeholder(sem, holder);
+#ifdef CONFIG_PRIORITY_INHERITANCE
+					if ((sem->flags & PRIOINHERIT_FLAGS_DISABLE) != 0)
```
jeongchanKim, line 564: I think we should move this code into the sem_unblock_task.
Because, in normal case, there is no free in sem_post. In this case, holder will be remained.
And you'd better remove the destroyholder in sem_setprotocol, because we should trace that sem also.

os/kernel/semaphore/sem_setprotocol.c
```diff
@@ -116,9 +116,11 @@ int sem_setprotocol(FAR sem_t *sem, int protocol)
 
 			sem->flags |= PRIOINHERIT_FLAGS_DISABLE;
 
+#ifndef CONFIG_BINMGR_RECOVERY
```
sunghan-chang, line 119: Please add the comment to say why this should be excluded.

os/kernel/semaphore/sem_post.c
```diff
@@ -147,7 +149,16 @@ void sem_unblock_task(sem_t *sem, struct tcb_s *htcb)
 #ifdef CONFIG_PRIORITY_INHERITANCE
 	if ((sem->flags & PRIOINHERIT_FLAGS_DISABLE) == 0) {
 		sem_restorebaseprio(stcb, htcb, sem);
+	} else
+#endif
+	{
```
sunghan-chang, line 154: This block should be in CONFIG_BINMGR_RECOVERY conditional.

PR #4387, opened 2020-07-16
network: Add zerocopy feature in Network Manager.

pillip8282:
To increase the performance, zerocopy feature is implemented.
it'll pass pbuf from lwIP to wi-Fi driver while it's sending data. so
the number of memory copy is reduced. On the contrary, the network manager
pass pbuf to lwIP directly when it receive data from Wi-Fi
driver. this feature is dependent to Wi-Fi driver. So it have to be
handled carefully

### DIFF
--- a/os/net/netmgr/netdev_lwip.c
+++ b/os/net/netmgr/netdev_lwip.c
@@ -49,10 +49,15 @@
 #define AF_INETX AF_INET
 #endif							/* CONFIG_NET_IPv6 */
 
+#define LW_GETND(nic) (struct netdev *)(*(struct netdev **)(&((char *)nic)[sizeof(struct netif)]))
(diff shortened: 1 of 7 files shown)


os/net/netmgr/netdev_lwip.c
```diff
@@ -70,9 +75,58 @@ static inline int _netif_down(struct netif *ni)
...
+		return ERR_IF;
+	}
+
+	return ERR_OK;
+}
+
+
+static int lwip_input(struct netdev *dev, void *frame_ptr, uint16_t len)
+{
+	(void)len;
```
jeongchanKim, line 95: I think this is not needed.

pillip8282: I did it to show that it is unused variable when ZEROCOPY is enabled

PR #4388, opened 2020-07-16
os/fs/smartfs : Do not GC after recover journal

Taejun-Kwon:
After recover journal, some of date(c.f freecount) is not updated,
so next GC will be happned during write data

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -5446,24 +5446,24 @@ static int smart_journal_recovery(FAR struct smart_struct_s *dev, journal_log_t 
 			}
 			
 			/* Data sector is released, checkout journal data */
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5446,24 +5446,24 @@ static int smart_journal_recovery(FAR struct smart_struct_s *dev, journal_log_t
 			}
 			
 			/* Data sector is released, checkout journal data */
-			return smart_journal_checkout(dev, log, address);
+			ret = smart_journal_checkout(dev, log, address);
```
jsdosa, line 5449: I have one question.
If crc fails, do we have to delete the sector?

Taejun-Kwon: crc fails mean that written data is invalid so we have to drop it.

PR #4389, opened 2020-07-16
os/fs/smartfs: apply sector recovery

Taejun-Kwon:
1. Recover unchained(isolated) sector due to sudden power
2. This can be used stand alone but with journal is much better.
3. Use timestamp instead of UTC if config is not set

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -87,16 +87,9 @@
 #define SET8BITSFROMMSB 255		//Binary: 0b11111111
 #define NEG8BIT(A) ((uint8_t)(~A & 0x000000FF))
 #endif
-/****************************************************************************
- * Private Types
- ****************************************************************************/
 
 /****************************************************************************
- * Private Function Prototypes
- ****************************************************************************/
-
-/****************************************************************************
- * Private Variables
+ * Private Definitions
  ****************************************************************************/
 
 #if defined(CONFIG_SMARTFS_MULTI_ROOT_DIRS) || \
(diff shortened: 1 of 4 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1685,25 +1718,334 @@ struct smartfs_mountpt_s *smartfs_get_first_mount(void)
...
 {
 	int ret;
 	long sector;
+	struct sector_recover_info_s info = {0,};
+	size_t size = (fs->fs_llformat.nsectors / 8) + 1;
+	
 	/* Alloc Logical Map */
-	uint8_t *map = (uint8_t*)kmm_malloc(sizeof(uint8_t*) * fs->fs_llformat.nsectors);
-
+	char *map = (char*)kmm_malloc(sizeof(uint8_t*) * size);
```
jsdosa, line 2043: could you add white space?
i mean 
```
(char *)
```

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1685,25 +1718,334 @@ struct smartfs_mountpt_s *smartfs_get_first_mount(void)
...
+		node = (struct sector_entry_queue_s *)sq_next(node);
+
+	}
+	ret = OK;
+errout :
+	node = (struct sector_entry_queue_s *)sq_peek(&entry_queue);
+	while (node != NULL) {
+		sq_rem((FAR sq_entry_t *)node, &entry_queue);
+		kmm_free(node);
+		node = (struct sector_entry_queue_s *)sq_next(node);
```
sunghan-chang, line 2021: After freeing node, and then you use it to get some data. It could cause system crash.

PR #4391, opened 2020-07-17
kernel/binary_manager: Skip crc checking of binary files when system booting

jeongarmy:
Binary manager validates header values of binary file without checking crc when system boot.
That's because it takes a long time to read whole file for checking crc value.
And then it validates header values and crc value when loading binary.


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -612,26 +612,27 @@ int binary_manager_read_header(char *path, binary_header_t *header_data)
...
-	/* Calculate checksum and Verify it */
-	check_crc = crc32part((uint8_t *)header_data + CHECKSUM_SIZE, header_data->header_size, check_crc);
-	file_size = header_data->bin_size;
-	while (file_size > 0) {
-		read_size = file_size < CRC_BUFFER_SIZE ? file_size : CRC_BUFFER_SIZE;
-		ret = read(fd, (FAR uint8_t *)crc_buffer, read_size);
-		if (ret < 0 || ret != read_size) {
-			bmdbg("Failed to read : %d, errno %d\n", ret, errno);
-			goto errout_with_fd;
+	if (crc_check) {
```
jsdosa, line 615: How about changing one of the two?
i mean, it's about check_crc and crc_check.
just in my opinion.

PR #4394, opened 2020-07-20
kernel/binary_manager: Load real-time binaries when initial loading

jeongarmy:
Binary manager checks RT types of binaries by scanning all binary files when initial loading.
And then it loads real-time binaries first.


os/kernel/binary_manager/binary_manager.h
```diff
@@ -156,6 +157,7 @@ typedef struct statecb_node_s statecb_node_t;
 binmgr_uinfo_t *binary_manager_get_udata(uint32_t bin_idx);
 #define BIN_ID(bin_idx)                                 binary_manager_get_udata(bin_idx)->bin_id
 #define BIN_STATE(bin_idx)                              binary_manager_get_udata(bin_idx)->state
+#define BIN_RTTYPE(bin_idx)                             binary_manager_get_udata(bin_idx)->rttype
```
jeongchanKim, line 160: How about renaming as `BIN_TYPE`? I think RT and NRT are type.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -255,10 +255,23 @@ static int binary_manager_load_all(void)
 	load_cnt = 0;
 	bin_count = binary_manager_get_ucount();
 
+	/* First, Load Real-time binaries */
```
jsdosa, line 258: The following is the same operation twice.
How about separating the functions or following below?
'''

jsdosa: Would you please double-check that the intention is correct?

jeongchanKim: We want to load REALTIME first and NONREALTIME later.
Because of this, she made separated loop.
I think we can make function with type param, but we cannot avoid the duplication.

jsdosa: i understood.
If the logic is duplicated, it is better to separate it into functions.

os/tools/mkbinheader.py
```diff
@@ -164,6 +164,12 @@ def get_static_ram_size(bin_type):
 COMP_MINIZ = 2
 COMP_MAX = COMP_MINIZ
 
+SCHED_PRIORITY_MIN = 1
+SCHED_PRIORITY_MAX = 255
+
+BM_PRIORITY_MIN = 200
```
jeongchanKim, line 170: In this case, we'd better make config for binary manager.
If not, we should modify both this and binary_manager.h.
That can be some mistakes sometime.

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -178,7 +179,7 @@ int binary_manager_create_entry(int requester_pid, char *bin_name, int version)
 		}
 	} else {
 		/* If it it not registered, register it */
-		ret = binary_manager_register_ubin(bin_name);
+		ret = binary_manager_register_ubin(bin_name, SCHED_PRIORITY_MIN);
```
jeongchanKim, line 182: Why do you try to register with lowest priority?
Isn't it  possible to register with real priority?

jeongchanKim, line 182: And we already checked that `bin_name` is already exist or not from `binary_manager_get_index_with_name`.
But in `binary_manager_register_ubin`, we checked again.
I think these are duplicated.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -116,6 +116,10 @@ int binary_manager_register_ubin(char *name)
 	for (bin_idx = 1; bin_idx <= g_bin_count; bin_idx++) {
 		/* Already Registered */
 		if (!strncmp(BIN_NAME(bin_idx), name, strlen(name) + 1)) {
+			if (priority > BIN_PRIORITY(bin_idx)) {
+				/* Update higher priority for early loading */
```
jeongchanKim, line 120: This is not always used for early loading.
If priority is higher than binary manager, it will be. But even if the priority is higher than before, but lower than binary manager, it will not be.
Isn't it?

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -104,7 +104,7 @@ binmgr_kinfo_t *binary_manager_get_kdata(void)
  *	 This function registers user binaries.
  *
  ****************************************************************************/
-int binary_manager_register_ubin(char *name)
+int binary_manager_register_ubin(char *name, uint8_t priority)
```
jeongchanKim, line 107: Please include stdint.h for uint8_t.

os/kernel/binary_manager/binary_manager_binfile.c
jeongarmy: Yes. we don't know a priority of new binary to be downloaded.

os/tools/mkbinheader.py
jeongarmy: I agree with your opinion. but these values are changed by user. I will think about it more.

jeongchanKim: Yes, these can be changed by user, so that can make problem.

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -116,8 +118,13 @@ void binary_manager_scan_ubin(void)
...
 					continue;
 				}
-				/* If binary is not registered, register it */
-				(void)binary_manager_register_ubin(header_data.bin_name);
+				bin_idx = binary_manager_get_index_with_name(header_data.bin_name);
+				if (bin_idx < 0) {
+					/* If binary is not registered, register it */
+					(void)binary_manager_register_ubin(header_data.bin_name, header_data.bin_priority);
+				} else if (header_data.bin_priority > BM_PRIORITY_MAX) {
+					BIN_PRIORITY(bin_idx) = header_data.bin_priority;
```
sunghan-chang, line 126: 1. Is there any chance to be reach here even this function is called at start of binary manager once?
2. Is it enough to save priority only? When do you write other information?

sunghan-chang: For 2nd, we could reduce the duplicated operation.
solution1.
1. scan binary at start of binary manager and at binary update and update the binary management table
2. load binary using above

solution2.
1. no update binary management table
2. update all of information when load binary

reason
1. Binary header which includes binary information is in flash so that it will not be changed
  => no problem with updating the table at early time
2. no need to read binary header twice for performance

os/kernel/binary_manager/binary_manager.h
```diff
@@ -156,12 +159,15 @@ typedef struct statecb_node_s statecb_node_t;
...
+#define BIN_USEIDX(bin_idx)                             binary_manager_get_udata(bin_idx)->useidx
 #define BIN_RTLIST(bin_idx)                             binary_manager_get_udata(bin_idx)->rt_list
 #define BIN_NRTLIST(bin_idx)                            binary_manager_get_udata(bin_idx)->nrt_list
 
-#define BIN_VER(bin_idx)                                binary_manager_get_udata(bin_idx)->bin_ver
+#define BIN_VER(bin_idx)                                binary_manager_get_udata(bin_idx)->load_attr.bin_ver
 #define BIN_KERNEL_VER(bin_idx)                         binary_manager_get_udata(bin_idx)->kernel_ver
 #define BIN_CBLIST(bin_idx)                             binary_manager_get_udata(bin_idx)->cb_list
+#define BIN_FILECNT(bin_idx)                            binary_manager_get_udata(bin_idx)->file_cnt
+#define BIN_FILEVER(bin_idx, file_idx)                  binary_manager_get_udata(bin_idx)->bin_ver[file_idx]
```
jeongchanKim, line 170: I think this is not file version. This is binary version. And we don't have any rule for file version.
We'd better rename like `BIN_VER` as `BIN_LATEST_VER`, and `BIN_FILEVER` as `BIN_VER`. Because active(running) binary will be always the latest one.

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -92,10 +93,11 @@ static int binary_manager_clear_binfile(int bin_idx)
  *	 This function scans user binary files.
  *
  ****************************************************************************/
-void binary_manager_scan_ubin(void)
+void binary_manager_scan_ubinfs(void)
```
jeongchanKim, line 96: How about changing the function name as `binary_manager_update_ubin_table`?
It can be confused because we have similar function, `binary_manager_scan_ubin`. It is not easy to distinguish.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -104,29 +105,24 @@ binmgr_kinfo_t *binary_manager_get_kdata(void)
...
-
 	/* If partition is not registered, Register it as a new user partition */
 	g_bin_count++;
 	BIN_ID(g_bin_count) = -1;
+	BIN_FILECNT(g_bin_count) = 1;
 	BIN_RTLIST(g_bin_count) = NULL;
 	BIN_NRTLIST(g_bin_count) = NULL;
+	BIN_PRIORITY(g_bin_count) = priority;
 	BIN_STATE(g_bin_count) = BINARY_INACTIVE;
+	if (version != NULL) {
```
jeongchanKim, line 123: Most of the other params are checked in the first of the function. I think we can check version also in the first, if there can be a problem without version. (I think if there is no version, it is not a valid one.)

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -126,6 +139,80 @@ void binary_manager_scan_ubin(void)
...
+		/* Read and Verify a binary file */
+		if (DIRENT_ISFILE(entryp->d_type) \
+			&& !strncmp(entryp->d_name, bin_name, name_len) && entryp->d_name[name_len] == '_') {
+			snprintf(filepath, CONFIG_PATH_MAX, "%s/%s", BINARY_DIR_PATH, entryp->d_name);
+			ret = binary_manager_read_header(filepath, &header_data, false);
+			if (ret == OK) {
+				BIN_FILECNT(bin_idx)++;
+				strncpy(BIN_FILEVER(bin_idx, file_idx), header_data.bin_ver, BIN_VER_MAX);
+				version = (int)atoi(header_data.bin_ver);
+				bmvdbg("Found valid header in version %d\n", version);
```
jeongchanKim, line 195: I think `Found valid header with version` is better. Because header is not in version.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -138,93 +131,59 @@ static int binary_manager_load(int bin_idx)
...
-			load_attr.stack_size = header_data[latest_idx].bin_stacksize;
-			load_attr.priority = header_data[latest_idx].bin_priority;
-			load_attr.offset = CHECKSUM_SIZE + header_data[latest_idx].header_size;
-			strncpy(load_attr.bin_ver, header_data[latest_idx].bin_ver, BIN_VER_MAX);
+			/* Read header data and Check crc */
+			snprintf(filepath, CONFIG_PATH_MAX, "%s/%s_%s", BINARY_DIR_PATH, BIN_NAME(bin_idx), BIN_FILEVER(bin_idx, BIN_USEIDX(bin_idx)));
+			ret = binary_manager_read_header(filepath, &header_data, true);
+			if (ret != OK) {
+				if (--bin_count > 0) {
+					bmdbg("Failed to read head %s, try to load another file\n", filepath);
```
jeongchanKim, line 152: When failure, we will try to read header again, not try to load. So I think `Failed to read head %s, try to read another file` is better.

os/kernel/binary_manager/binary_manager_data.c
jeongarmy: It is for new binary to be downloaded. We cannot know the version at this time.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -104,29 +105,24 @@ binmgr_kinfo_t *binary_manager_get_kdata(void)
...
 
-	for (bin_idx = 1; bin_idx <= g_bin_count; bin_idx++) {
-		/* Already Registered */
-		if (!strncmp(BIN_NAME(bin_idx), name, strlen(name) + 1)) {
-			bmdbg("Already registered for binary %s\n", BIN_NAME(bin_idx));
-			return ERROR;
-		}
-	}
-
 	/* If partition is not registered, Register it as a new user partition */
```
jeongchanKim, line 115: I think this comment is not matched with our current implementation.

os/kernel/binary_manager/binary_manager_data.c
jeongchanKim: You can use `version` even for new binary.

os/tools/mkbinheader.py
```diff
@@ -103,7 +114,7 @@ def get_static_ram_size(bin_type):
 # +--------------------------------------------------------------------------
 # ----------------------------------------------------------------------
 # | Binary name | Binary version | Binary ram size | Binary stack size |
-# |  (16bytes)  |   (16bytes)    |    (4bytes)     |     (4bytes)      |
+# |  (16bytes)  |    (4bytes)    |    (4bytes)     |     (4bytes)      |
```
jeongchanKim, line 117: Please update the total header size also.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -37,7 +37,7 @@
 #define APP_NAME                   "micom"
 
 #define NEW_APP_NAME               "newapp"
-#define NEW_APP_VERSION            "20200421"
+#define NEW_APP_VERSION            20200421
```
jeongchanKim, line 40: Please check this code which uses NEW_APP_VERSION as int.

os/include/tinyara/binfmt/binfmt.h
```diff
@@ -177,7 +177,7 @@ struct binary_s {
 	uint8_t compression_type;		/* Binary Compression type */
 #ifdef CONFIG_BINARY_MANAGER
 	uint8_t binary_idx;             /* Index of binary in binary table */
-	char bin_ver[BIN_VER_MAX];      /* version of binary */
+	uint8_t bin_ver;                /* version of binary */
```
jeongchanKim, line 180: I think this should be uint32_t

build/configs/imxrt1050-evk/loadable_elf_apps/defconfig
```diff
@@ -1015,6 +1015,8 @@ CONFIG_MPU_TEST_APP_ADDR=0x80200100
 # Binary manager
 #
 CONFIG_BINARY_MANAGER=y
+CONFIG_BM_PRIORITY_MAX=205
+CONFIG_BM_PRIORITY_MIN=200
```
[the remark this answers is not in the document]
jeongarmy: 1. These values are used in mkbinheader.py and binary manager codes to validate priority so I made it to avoid mismatching between them.
2. I will add it

os/tools/mkbinheader.py
```diff
@@ -173,6 +187,10 @@ def get_static_ram_size(bin_type):
     print("Dynamic ram size : %d, Main stack size : %d" %(int(dynamic_ram_size), int(main_stack_size)))
     sys.exit(1)
 
+config_path = os.getenv('TOPDIR') + '/.config'
+BM_PRIORITY_MAX = get_config_value(config_path, "CONFIG_BM_PRIORITY_MAX=")
```
jeongchanKim, line 191: How about setting the default BM_PRIORITY_MAX and MIN if `get_config_value` returns fail?

os/tools/mkbinheader.py
```diff
@@ -96,14 +107,14 @@ def get_static_ram_size(bin_type):
 #
 # header information :
 #
-# total header size is 61bytes.
+# total header size is 49bytes.
```
jeongchanKim, line 110: I think you didn't update the header information. Please update this.

os/tools/mkbinheader.py
```diff
@@ -37,6 +37,17 @@ def roundup_power_two(size):
 
     return size
 
+def get_config_value(file_name, config):
+    with open(file_name, 'r+') as f:
```
jeongchanKim, line 41: Please use tab instead of space.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -633,7 +643,7 @@ int binary_manager_read_header(char *path, binary_header_t *header_data, bool cr
 			goto errout_with_fd;
 		}
 	}
-	bmvdbg("Binary header : %d %d %d %s %s %d %s %d\n", header_data->header_size, header_data->bin_type, header_data->bin_size, header_data->bin_name, header_data->bin_ver, header_data->bin_ramsize, header_data->kernel_ver, header_data->jump_addr);
+	bmvdbg("Binary header : %d %d %d %s %d %d %s %d\n", header_data->header_size, header_data->bin_type, header_data->bin_size, header_data->bin_name, header_data->bin_ver, header_data->bin_ramsize, header_data->kernel_ver, header_data->jump_addr);
```
jeongchanKim, line 646: How about printing loading priority also?

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -138,93 +167,60 @@ static int binary_manager_load(int bin_idx)
...
-			strncpy(load_attr.bin_name, header_data[latest_idx].bin_name, BIN_NAME_MAX);
-			load_attr.bin_size = header_data[latest_idx].bin_size;
-			load_attr.compression_type = header_data[latest_idx].compression_type;
-			load_attr.ram_size = header_data[latest_idx].bin_ramsize;
-			load_attr.stack_size = header_data[latest_idx].bin_stacksize;
-			load_attr.priority = header_data[latest_idx].bin_priority;
-			load_attr.offset = CHECKSUM_SIZE + header_data[latest_idx].header_size;
-			strncpy(load_attr.bin_ver, header_data[latest_idx].bin_ver, BIN_VER_MAX);
+			/* Set a priority of loader according to loading priority */
+			(void)binary_manager_set_loader_priority(BIN_LOAD_PRIORITY(bin_idx, BIN_USEIDX(bin_idx)));
```
jeongchanKim, line 184: `BIN_LOAD_PRIORITY` is uint8_t. So you'd better make type casting.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -59,6 +59,42 @@ extern sq_queue_t g_sem_list;
...
+/****************************************************************************
+ * Name: binary_manager_set_loader_priority
+ *
+ * Description:
+ *   This function sets priority of loader according to loading priority of binary.
+ *
+ ****************************************************************************/
+static void binary_manager_set_loader_priority(int priority)
+{
+	int ret;
```
jeongchanKim, line 71: This is not needed.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -104,29 +105,22 @@ binmgr_kinfo_t *binary_manager_get_kdata(void)
  *	 This function registers user binaries.
  *
  ****************************************************************************/
-int binary_manager_register_ubin(char *name)
+int binary_manager_register_ubin(char *name, uint32_t version, uint8_t priority)
```
jeongchanKim, line 108: How about changing `priority` to `load_prio`?

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -59,6 +59,42 @@ extern sq_queue_t g_sem_list;
...
+		loadthd_priority = LOADINGTHD_PRIORITY_LOW;
+		break;
+	case BINARY_LOADING_MID:
+		loadthd_priority = LOADINGTHD_PRIORITY_MID;
+		break;
+	case BINARY_LOADING_HIGH:
+		loadthd_priority = LOADINGTHD_PRIORITY_HIGH;
+		break;
+	default:
+		bmdbg("Invalid priority\n", priority);
```
jeongchanKim, line 86: You missed `%d`.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -255,10 +251,15 @@ static int binary_manager_load_all(void)
 	load_cnt = 0;
 	bin_count = binary_manager_get_ucount();
 
-	for (bin_idx = 1; bin_idx <= bin_count; bin_idx++) {
-		ret = binary_manager_load(bin_idx);
-		if (ret == OK) {
-			load_cnt++;
+	/* Load the binaries from high priority */
```
jeongchanKim, line 254: I think `with high priority` is better.

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: I changed parameter type of the function.

os/kernel/binary_manager/binary_manager.h
```diff
@@ -74,7 +86,7 @@
 #define COMMLIB_IDX                0
 
 /* The number of arguments for loading thread */
-#define LOADTHD_ARGC               2
+#define LOADTHD_ARGC               1
```
jeongchanKim, line 89: You'd better rename this also as `LOADER_ARGC`.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -405,127 +355,175 @@ static int binary_manager_reload(int bin_idx)
...
- *   bin_name   -   The name of binary to be reload
+ ****************************************************************************/
+static int loading_thread(int argc, char *argv[])
+{
+	if (argc <= 1) {
+		bmdbg("Invalid arguments for loading, argc %d\n", argc);
+		return ERROR;
+	}
+
+	/* [1] binary index for loading */
```
jeongchanKim, line 371: I think `argv[1]` is better.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -405,127 +355,175 @@ static int binary_manager_reload(int bin_idx)
...
+	return binary_manager_load((int)atoi(argv[1]));
+}
+
+/****************************************************************************
+ * Name: loadingall_thread
  *
- * Returned Value:
- *   None
+ * Description:
+ *   This function loads all user binaries in binary table.
```
jeongchanKim, line 379: You mean `This function loads all user binaries WHICH ARE in binary table.` ? or `This function loads all user binaries FROM binary table.`?

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -405,127 +355,175 @@ static int binary_manager_reload(int bin_idx)
...
+		return ERROR;
+	}
+
+	if (BIN_STATE(bin_idx) < BINARY_INACTIVE || BIN_STATE(bin_idx) > BINARY_RUNNING) {
+		bmdbg("Invalid binary state %d\n", BIN_STATE(bin_idx));
+		return BINMGR_OPERATION_FAIL;
+	}
+
+	/* Terminate binary if binary is already loaded */
+	if (BIN_STATE(bin_idx) == BINARY_LOADING_DONE || BIN_STATE(bin_idx) == BINARY_RUNNING) {
```
jeongchanKim, line 504: If there is no new binary, so we don't need to update, then we don't need to terminate the previous one.
We can just ignore the update request.
Please check this logic.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -59,6 +59,38 @@ extern sq_queue_t g_sem_list;
...
+/****************************************************************************
+ * Name: binary_manager_get_loader_priority
+ *
+ * Description:
+ *   This function gets a priority of loader according to loading priority of binary.
+ *
+ ****************************************************************************/
+static int binary_manager_get_loader_priority(uint8_t load_priority)
+{
+	int loader_priority;
```
jeongchanKim, line 71: You'd better declare as unsigned.

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: yes. I agree with your opinion. I will modify later.

os/tools/mkbinheader.py
```diff
@@ -199,10 +234,20 @@ def get_static_ram_size(bin_type):
...
-        main_priority = int(main_priority)
-    else :
-        print("Error : This binary priority is not valid")
+    main_priority = int(main_priority)
+    if (main_priority < SCHED_PRIORITY_MIN) or (main_priority > SCHED_PRIORITY_MAX) or (BM_PRIORITY_MIN <= main_priority and main_priority <= BM_PRIORITY_MAX) :
+        print("Error : This binary priority ", main_priority, " is not valid")
+        sys.exit(1)
+
+    # Loading priority
+    if loading_priority == 'H' or loading_priority == 'HIGH' :
```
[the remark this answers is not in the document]
jeongarmy: loading_priority is string type,
but I assign int value and call pack function after checking value.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -59,6 +59,38 @@ extern sq_queue_t g_sem_list;
...
  * Private Functions
  ****************************************************************************/
+/****************************************************************************
+ * Name: binary_manager_get_loader_priority
+ *
+ * Description:
+ *   This function gets a priority of loader according to loading priority of binary.
+ *
+ ****************************************************************************/
+static int binary_manager_get_loader_priority(uint8_t load_priority)
```
jeongchanKim, line 69: Please change the return type as uint8_t.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -59,6 +59,38 @@ extern sq_queue_t g_sem_list;
...
+		loader_priority = LOADER_PRIORITY_LOW;
+		break;
+	case BINARY_LOADPRIO_MID:
+		loader_priority = LOADER_PRIORITY_MID;
+		break;
+	case BINARY_LOADPRIO_HIGH:
+		loader_priority = LOADER_PRIORITY_HIGH;
+		break;
+	default:
+		bmdbg("Invalid loading priority : %d\n", load_priority);
```
jeongchanKim, line 86: Please print the log with `%u`.

PR #4404, opened 2020-07-25
os/fs/smartfs: Optimize number of loop iterations in smartfs_deleteentry()

amogh-samsung:
- Exit the loop for counting entries when count>1.
- If count>1, we know that sector can't be released.

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -1579,6 +1579,10 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 			if (ENTRY_VALID(direntry)) {
 				/* Count this entry */
 				count++;
+				if (count == 2) {
+					/* If valid entry count exceeds 1, we know we cannot release the sector */
+					break;
+				}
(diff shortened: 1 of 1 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1579,6 +1579,10 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 			if (ENTRY_VALID(direntry)) {
 				/* Count this entry */
 				count++;
+				if (count == 2) {
```
Taejun-Kwon, line 1582: good! =D I like it. but below is better.
```
if (ENTRY_VALID(direntry)) {
     /* If there's already valid entry, break it */
     if (count == 1) {
           break;
     }
     count++;
}
```

amogh-samsung: This way, when the loop exits with count = 1
We don't know if the loop ended naturally
OR it was due to the break statement.

Taejun-Kwon: ah yes. you're right.

PR #4406, opened 2020-07-26
os/fs/smartfs: Write nextsector to parent after writing new entry

amogh-samsung:
- write entry to new sector first, then write chainheader->nestsector
  for old parent sector.
- If a power off occurs before the last step, smartfs recovery will
  handle the new isolated sector.
- remove unnecessary sector read from smartfs_createentry().

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -1127,53 +1127,20 @@ errout:
 
 int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, struct smartfs_entry_s *direntry)
 {
-	struct smart_read_write_s readwrite;
 	int ret;
 	uint16_t nextsector;
-	struct smartfs_chain_header_s *chainheader;
 
-	/* Start at the 1st sector in the parent directory */
 	/* If we are in the context of smartfs_createentry, then smartfs_find_availableentry did not find an available entry.
 	   Createentry has to chain a new nextsector to parentdirsector, and allocate the first entry.
 	 */
 
-	smartfs_setbuffer(&readwrite, parentdirsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
-	ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-	if (ret < 0) {
-		fdbg("Unable to read parent sector\n");
-		return ret;
-	}
-
-	chainheader = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
-	nextsector = SMARTFS_NEXTSECTOR(chainheader);
-
-	if (nextsector != SMARTFS_ERASEDSTATE_16BIT) {
-		fdbg("Cannot chain new sector\n");
-		return -EIO;
-	}
 	/* Allocate a new sector and chain it to the last one */
-
 	ret = FS_IOCTL(fs, BIOC_ALLOCSECT, 0xFFFF);
 	if (ret < 0) {
 		return ret;
 	}
 
 	nextsector = (uint16_t)ret;
-
-	/* Chain the next sector into this sector sector */
-	chainheader->nextsector[0] = (uint8_t)(nextsector & 0x00FF);
-	chainheader->nextsector[1] = (uint8_t)(nextsector >> 8);
-
-	/* Here we are writing to the current sector's chainheader which needed a new sector in the chain.
-	   Not writing to the newly chained sector, so it is completely empty as of now.
-	 */
-
-	smartfs_setbuffer(&readwrite, parentdirsector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), chainheader->nextsector);
-	ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-	if (ret < 0) {
-		fdbg("Error chaining sector %d\n", nextsector);
-		return ret;
-	}
 
 	/* Automatically selecting first entry of new chained sector */
 	direntry->doffset = sizeof(struct smartfs_chain_header_s);
@@ -1192,13 +1159,12 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, 
  *
  * Input Parameters:
  *   fs - pointer to smartfs mountpoint structure.
- *   parentdirsetor - sector in which an available entry is to be found
+ *   parentdirsetor - pointer to integer to keep track of new sector
+ *     allocation and chaining.
  *   direntry - pointer to smartfs entry structure for returning details
  *     of the available entry found.
  *     If no available entry is found, it returns with the logical sector
  *     number of the last sector chained to parentdirsector.
- *   new_chain - pointer to boolean object to return whether a new sector
- *     was chained to the parent for the new entry or not.
  *
  * Returned Value:
  *   OK - on finding an available entry.
@@ -1207,7 +1173,7 @@ int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, 
  *
  ****************************************************************************/
 
-int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, struct smartfs_entry_s *direntry, bool *new_chain)
+int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdirsector, struct smartfs_entry_s *direntry)
 {
 	int ret = -ENOENT;
 	struct smart_read_write_s readwrite;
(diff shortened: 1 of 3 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1296,9 +1266,11 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 	struct smartfs_entry_header_s *entry;
 	struct smartfs_chain_header_s *chainheader;
 	char *tmp_buf = NULL;
+	uint16_t nextsector;
 
 	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
 	psector = new_entry.dsector;
+	/* In case newchain == TRUE, offset will hold the logical sector numer to which the new sector is to be chained */
```
deepaksrma, line 1273: is it sector number?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1375,10 +1349,25 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
...
+		   The new sector has to be chained to this parent sector after the entry is written.
+		 */
+
+		/* Chain the next sector into this sector sector */
+		nextsector = psector;
+
+		smartfs_setbuffer(&readwrite, offset, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)&nextsector);
+		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
+		if (ret < 0) {
+			fdbg("Error chaining sector %d\n", psector);
```
jeongchanKim, line 1363: Is it okay not to free the `tmp_buf`?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1127,56 +1127,26 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-		fdbg("Error chaining sector %d\n", nextsector);
-		return ret;
-	}
-
 	/* Automatically selecting first entry of new chained sector */
-	direntry->doffset = sizeof(struct smartfs_chain_header_s);
+	/* With createentry, offset is known to be sizeof(chainheader) automatically,
+	   Let's use direntry->doffset for saving the parent logical sector to write chainheader->nextsector later.
+	 */
+	direntry->doffset = parentdirsector;
```
Taejun-Kwon, line 1149: I think this is not good option.
1. Actually this only called from find_availableentry, so it already know parentdirsector.
2. There's chance that asign sector number to 'offset' will bring great confusion later. 
    This can make hard to debugging later.
3. If we use offset as a entire of given flash for file system, then it is acceptable but many things should be changed.
   for example,
```
direntry->doffset = parentdirsector * size_of_sector + target_offset
```

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1375,10 +1349,25 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
...
 
+	if (newchain) {
+		/* In case we are writing to a new sector, new_entry.doffset holds the parent logical sector number.
+		   The new sector has to be chained to this parent sector after the entry is written.
+		 */
+
+		/* Chain the next sector into this sector sector */
+		nextsector = psector;
+
+		smartfs_setbuffer(&readwrite, offset, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)&nextsector);
```
Taejun-Kwon, line 1360: I think we'd better send parentsector instead of newchain as a parameter.
```
int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry, uint16_t parentsector)
{
   ...
   bool newchain = false;
   ...
   if (parentsector != newentry->dsector) {
         newchain = true;
   }
   ....
}

Comment (amogh-samsung) [reply os/fs/smartfs/smartfs_utils.c:1360]: Previous implementation did not have new_chain as extra parameter, right?
You prefer to keep it that way if I'm right?

File: os/fs/smartfs/smartfs_utils.c [hunk 5]
```diff
@@ -1375,10 +1349,25 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 		goto errout;
 	}
 
+	if (newchain) {
```
Taejun-Kwon, line 1352: I think this will work correct for mkdir.
But we should consider more for open->write->sync_internal.
for example we create /mnt/a as a 13th entry in /mnt folder, and write 300bytes and close, we have to follow below.
1. alloc sector for /mnt/a save sf regarding this. (but we do not asign alloc sector to sf->currsector, is it neede?)
2. fill the buffer when write being called.
3. Now we have to sync if close has called.
4. First write /mnt/a's firstsector (sf->currsector)
5. create 2nd entry sector of /mnt
6. Chain first~2nd entry of /mnt

Will you handle this next PR? otherwise, we need TODO

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: This function added in last PR, I can't remember whole of history.
It was acceptable, but now we need information of parentsector, that's all.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes I need to make sf buffer changes and buffer sync modifications.
That is pending for smartfs_open

os/fs/smartfs/smartfs_utils.c
amogh-samsung: How about assigning
direntry.dfirst = parentdirsector

Or you want that no attribute of direntry should be assigned?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Changes done

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Resolved

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Corrected,  thankyou

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1303,9 +1271,11 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 	struct smartfs_entry_header_s *entry;
 	struct smartfs_chain_header_s *chainheader;
 	char *tmp_buf = NULL;
+	uint16_t nextsector;
 
 	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
 	psector = new_entry.dsector;
+	/* In case newchain == TRUE, offset will hold the logical sector numer to which the new sector is to be chained */
```
Taejun-Kwon, line 1278: Old description, please remove this line or change

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1321,7 +1291,8 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 		SET_ENTRY_FLAGS(new_entry.flags, mode, type);
 	}
 
-	if (newchain) {
+	/* If passed parameter parentdirsector != new_entry.dsector, it means it is a new chain sector that we will write to */
+	if (parentdirsector != psector) {
```
Taejun-Kwon, line 1295: line 1309, below is correct, am I right?
chainheader->type = type;

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1382,6 +1353,22 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
...
 
+	if (parentdirsector != psector) {
+		/* In case we are writing to a new sector, new_entry.doffset holds the parent logical sector number.
+		   The new sector has to be chained to this parent sector after the entry is written.
+		 */
+
+		/* Chain the next sector into this sector sector */
+		nextsector = psector;
+
+		smartfs_setbuffer(&readwrite, parentdirsector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)&nextsector);
```
Taejun-Kwon, line 1364: yeah.. this is much better.

os/fs/smartfs/smartfs_utils.c

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Type parameter passed is for the type of the new entry that is to be written.

Chainheader is for writing the chainheader of the new sector that was allocated for chaining to the parent, so it will always have type = SMARTFS_DIRENT_TYPE_DIR

Taejun-Kwon: yeah correct. =)

PR #4408, opened 2020-07-27
os/driver/usb Fix ioctl of cdcacm

Taejun-Kwon:
both of host and dev of cdcacm use priv structure in ioctl.
Fix to use priv properly


os/drivers/usbdev/cdcacm.c
```diff
@@ -1682,7 +1682,9 @@ static void cdcuart_detach(FAR struct uart_dev_s *dev)
 
 static int cdcuart_ioctl(FAR struct uart_dev_s *dev, int cmd, unsigned long arg)
 {
-	FAR uart_dev_t *serdev = &priv->serdev;
+	struct cdcacm_dev_s *priv = (FAR struct cdcacm_dev_s *)dev->priv;
+	serdev = &priv->serdev;
```
sunghan-chang, line 1686: There is no serdev definition.

PR #4411, opened 2020-07-27
configs/rtl8721csm: Add 'download' function support and Image Download Tool

AdamZhang0124:
1. Added AmebaD image download tool.
2. Added 'make download' function for all configurations.
3. Parameter $(2) is set as serial port number. ttyUSB1 is the default value.
4. Tested build and download functions in both tz enabled and disabled modes.


build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ -n "$2" ];then
+	PORT="$2"
+else
+	PORT="ttyUSB1"
+fi
+
+WARNING="\n Port $PORT is selected\n\n
+	############################################\n
+	WARNINGS:\n
+	1. Make sure the board is in DOWNLOAD MODE.\n
```
jeongchanKim, line 30: Is there any way to check the board is DOWNLOAD MODE or not?

build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -0,0 +1,56 @@
...
+
+cd ${IMG_TOOL_PATH}
+
+echo "Start Download"
+$SUDO ./amebad_image_tool /dev/$PORT
+
+$SUDO [ -e km0_boot_all.bin ] && rm km0_boot_all.bin
+$SUDO [ -e km4_boot_all.bin ] && rm km4_boot_all.bin
+$SUDO [ -e km0_km4_image2.bin ] && rm km0_km4_image2.bin
+
```
amogh-samsung, line 56: Could the extra line be removed from the end please?

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Hi @jeongchanKim ,
There are two ways to check if AmebaD is in download mode.
1) You can open SSH clients (like Putty, Tera Term) to check. If AmebaD is in download mode, you will see the message. (Reminder to disconnect it after checking to release the occupation of the serial port.)
2) More obviously, if AmebaD is in download mode, one of the led will be bilking.

Thanks!

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Hi @amogh-samsung ,
Sure, it will be removed with other amendment requirements, if any.
Thanks!

build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
sunghan-chang, line 1: To make it running on MAC, let's use below.
```bash
#!/usr/bin/env bash
```

build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -0,0 +1,56 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+# rtl8721csm_download.sh
+
+if [ -n "$2" ];then
```
sunghan-chang, line 21: Is there any usages of $1? Why do you use $2 without $1?
Could you add the message to say how to use this script?

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Hi @sunghan-chang ,

Sure, that will be fixed, thanks!

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Hi @sunghan-chang ,

The $1 is currently represents the selected download option. 
image

The $2 is an optional input. Without providing it, the serial port is set as ttyUSB1 by default.
This parameter is added only for leaving an configurable option for users. 

To download image, you can use following command:
1)  ./dbuild menu  (will provide default serial port setting)
or
2)  ./dbuild.sh download ALL ttyUSB1

Thanks!

sunghan-chang: In this script, there is no `$1` handling. It means `$1` is redundant.
If I give the command as `./dbuild.sh download os ttyUSB1`. Does it download os binary only with bootloader?
It looks no.

AdamZhang0124: Yes, you are right!
This download function only support 'download ALL' function for now.
$1 is currently reserved and left for further use of specialized download options.

AdamZhang0124: To prevent confusion, I will add some lines to restrict the input that only 'ALL' option can be selected for now. For the not provided inputs, an reminder message will be displayed. Thanks!

sunghan-chang: That's good.

AdamZhang0124:
Dear @sunghan-chang , @jeongchanKim , @amogh-samsung ,
The aforementioned suggestions have been all included in the latest commit. Thanks!

As for the unsupported download option inputed, the warning message will be displayed as follow for your information:    
image

PR #4413, opened 2020-07-28
fix syntax errors on Kconfig

sunghan-chang:
1. examples/rssi_report: fix syntax error of Kconfig
2. os/Kconfig: fix syntax error in binary separation configs


sunghan-chang:
@kishore-sn Could you review 3rd commit? https://github.com/Samsung/TizenRT/pull/4413/commits/d72966f28df89fed0abeeb616b7216187aafc5b8

build/configs/artik053/rssi_report/defconfig
```diff
@@ -217,9 +216,9 @@ CONFIG_BOOT_RUNFROMFLASH=y
 #
 # Boot Memory Configuration
 #
-CONFIG_RAM_REGIONx_START="0x02023800"
-CONFIG_RAM_REGIONx_SIZE="968704"
-CONFIG_RAM_KREGIONx_HEAP_INDEX="0,"
+CONFIG_RAM_KREGIONx_START="0x02023800"
+CONFIG_RAM_KREGIONx_SIZE="968704"
+CONFIG_RAM_MALLOC_PRIOR_INDEX=0
```
jeongchanKim, line 221: In this defconfig, we only use the single heap. In this case, we don't need this config.

os/mm/Kconfig
```diff
@@ -4,22 +4,12 @@
...
-	depends on BUILD_PROTECTED || BUILD_KERNEL
 	---help---
-		Partition heap memory into two parts:  (1) a protected, kernel-mode
-		heap accessible only by the TinyAra kernel, and (2) an unprotected
-		user-mode heap for use by applications.  If you are only interested
-		in protected the kernel from read access, then this option is not
-		necessary.  If you wish to secure the kernel data as well, then
-		this option should be selected.
-
 		The kernel heap size that is used is provided a a platform-specific
```
jeongchanKim, line 10: Pleaes check this comment also. There are some typos and I think this is not proper explanation.

sunghan-chang: Removed.

build/configs/artik053/rssi_report/defconfig
sunghan-chang: This is not a part of this PR. Even though I entered menuconfig, this is not disappeared. To remove this, we should check Kconfig first. Let's do that with another PR.

PR #4416, opened 2020-07-30
os/fs/driver/mtd: Do away with unnecessary ERROR condition labels

amogh-samsung:


Taejun-Kwon:
good.

PR #4417, opened 2020-07-30
os/fs/smartfs : Modify number of cleaned entry properly

Taejun-Kwon:
If next entry or sector is not exist, it clean entry but it doesn't increase count


amogh-samsung:
Looks good

PR #4418, opened 2020-07-31
fs/smartfs/nxfuse : Fix error nxfuse with new journaling

Taejun-Kwon:
New journaling initialized before smart_scan, so if write journal data failed, fusing also failed
1. Fix unnecessary reset of journal state when checkout
2. Set 0xFF when fuse make image
3. Improve logic of filemtd_write

### DIFF
--- a/tools/nxfuse/mksmartfsimg.sh
+++ b/tools/nxfuse/mksmartfsimg.sh
@@ -67,22 +67,32 @@ mkdir $CONTENTSDIR/$BOARDNAME/mnt/bins
 fi
 fi
 
+echo "make a dummy"
 #Make a dummy .bin file
 touch $BINDIR/$BINNAME
 
+echo "make image file"
 # Making image file
-dd if=/dev/zero of=$BINDIR/$BINNAME bs=$blksize count=$blkcount
+if [ "${CONFIG_SMARTFS_ERASEDSTATE}" == 0xff ]; then
+	dd if=/dev/zero bs=$blksize count=$blkcount | tr "\000" "\377" > $BINDIR/$BINNAME
+else
+	dd if=/dev/zero of=$BINDIR/$BINNAME bs=$blksize count=$blkcount
+fi
 
+echo "copy the nxfuse"
 #Copying the nxfuse
 cp $NXFUSEDIR/nxfuse .
 
 
+echo "Now format image"
 # Formatting
 ./nxfuse -p $pagesize -e $erasesize -l $blksize -t smartfs -m $BINDIR/$BINNAME || exit 1
 
+echo "mount binary"
 # Mounting mnt
 ./nxfuse -o nonempty -p $pagesize -e $erasesize -l $blksize -t smartfs $CONTENTSDIR/$BOARDNAME/mnt $BINDIR/$BINNAME || exit 1
 
+echo "Copy files"
 # Copying files to smartfs file system
 cp -rf $CONTENTSDIR/$BOARDNAME/base-files/* $CONTENTSDIR/$BOARDNAME/mnt/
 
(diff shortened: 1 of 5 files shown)


tools/nxfuse/mksmartfsimg.sh
```diff
@@ -67,22 +67,32 @@ mkdir $CONTENTSDIR/$BOARDNAME/mnt/bins
...
 fi
 
+echo "make a dummy"
 #Make a dummy .bin file
 touch $BINDIR/$BINNAME
 
+echo "make image file"
 # Making image file
-dd if=/dev/zero of=$BINDIR/$BINNAME bs=$blksize count=$blkcount
+if [ "${CONFIG_SMARTFS_ERASEDSTATE}" == 0xff ]; then
```
amogh-samsung, line 76: Why now is there a need to translate "\000" to "\377" while creating unformatted image?
Currently it works fine too I believe...

Or have you added this to create correct unformatted image in the 1st time itself? (With erasedstate = 0xFF)

tools/nxfuse/src/filemtd.c
```diff
@@ -161,17 +161,25 @@ static ssize_t filemtd_write(FAR struct file_dev_s *priv, size_t offset, FAR con
 	uint8_t srcvalue;
 	uint8_t newvalue;
 	size_t seekpos;
-
+	size_t bufsize = sizeof(buf);
```
amogh-samsung, line 164: Can we use a macro for this?

tools/nxfuse/src/filemtd.c
```diff
@@ -161,17 +161,25 @@ static ssize_t filemtd_write(FAR struct file_dev_s *priv, size_t offset, FAR con
...
-			buflen = read(priv->fd, buf, sizeof(buf));
+			if (len < bufsize) {
+				buflen = len;
+			} else {
+				buflen = bufsize;
+			}
+			readsize = read(priv->fd, buf, buflen);
+			if (readsize != buflen) {
+				return ERROR;
+			}
```
amogh-samsung, line 182: Is len -= buflen required here?

tools/nxfuse/mksmartfsimg.sh
Taejun-Kwon: That means set 8bit to 0xff, 377 is 0xff for octal.

tools/nxfuse/src/filemtd.c
Taejun-Kwon: sizeof is not function it is kind of operator, so I think there's no benefit from macro.

tools/nxfuse/src/filemtd.c
Taejun-Kwon: please understand role of this function, it is read specific size of data, and check bit byte by byte. len is decreased at the end of loop.

tools/nxfuse/src/filemtd.c
amogh-samsung: I mean use MACRO instead of 128
`char buf[128]`

tools/nxfuse/mksmartfsimg.sh
amogh-samsung: Adding the condition to check erasedstate = 0xff
It will create unformatted image with all bytes set to 0xff initially instead of 0x00, right?

That is the intention with this change

amogh-samsung: So with this, we can actually save one extra step of first erasing the entire partition to 0xff
If it is initially formatted to 0x00

Taejun-Kwon: yes to prevent an error when write journal data at the first time, as I told you flow of nxfuse is not aligned with smartfs's flow.
So it try to record first journal data when it llformat.

tools/nxfuse/src/filemtd.c
Taejun-Kwon: yes, using constant value is always not good. I will update.

PR #4421, opened 2020-08-05
os/binfmt : Fix wrong RO section address

jeongchanKim:
RO section should be from ALLOC_RO, not from ALLOC_DATA.
There are two wrong settings, so fix them.

### DIFF
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -226,7 +226,7 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	/* Print Binary section address & size details */
 
 	binfo("[%s] text    start addr =  0x%x  size = %u\n", bin->bin_name, (uint32_t)bin->alloc[ALLOC_TEXT], bin->textsize);
-	binfo("[%s] rodata  start addr =  0x%x  size = %u\n", bin->bin_name, (uint32_t)bin->alloc[ALLOC_DATA], bin->rosize);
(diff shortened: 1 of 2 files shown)


os/binfmt/binfmt_loadbinary.c
```diff
@@ -226,7 +226,7 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	/* Print Binary section address & size details */
 
 	binfo("[%s] text    start addr =  0x%x  size = %u\n", bin->bin_name, (uint32_t)bin->alloc[ALLOC_TEXT], bin->textsize);
-	binfo("[%s] rodata  start addr =  0x%x  size = %u\n", bin->bin_name, (uint32_t)bin->alloc[ALLOC_DATA], bin->rosize);
+	binfo("[%s] rodata  start addr =  0x%x  size = %u\n", bin->bin_name, (uint32_t)bin->alloc[ALLOC_RO], bin->rosize);
```
Taejun-Kwon, line 229: 1 Question, RO & DATA are defined as enum, also enabled by CONFIG_APP_BINARY_SEPARATION, they are safe here without CONFIG_APP_BINARY_SEPARATION?

jeongchanKim: Yes. It is okay. Because `load_binary()` is enclosed with CONFIG_BINARY_MANAGER, and CONFIG_BINARY_MANAGER has dependency with CONFIG_APP_BINARY_SEPARATION.

Taejun-Kwon: ok I thought so but I was just curious.

PR #4423, opened 2020-08-05
os/fs/smartfs: Modify sequence for writing new file entry/data with CRC enabled

amogh-samsung:
- When CRC is enabled
- Sync file data with MTD first.
- Then write the new file entry ot MTD.
- If the entry was written to a new sector on the parent directory,
  Chain the new sector to the parent directory now.

### DIFF
--- a/os/fs/smartfs/smartfs.h
+++ b/os/fs/smartfs/smartfs.h
@@ -175,6 +175,10 @@
 #define INODE_STATE_FILE          (CONFIG_NXFFS_ERASEDSTATE ^ 0x22)
 #define INODE_STATE_DELETED       (CONFIG_NXFFS_ERASEDSTATE ^ 0xaa)
 
+/* Smartfs worbuffer maxuimum length */
+
+#define SMARTFS_MAX_WORKBUFFER_LEN 256
+
 /* Directory entry flag definitions */
 
 #define SMARTFS_DIRENT_EMPTY      0x8000	/* Set to non-erase state when entry used */
@@ -262,6 +266,7 @@ struct smartfs_entry_s {
 	FAR char *name;				/* inode name */
 	uint32_t utc;				/* Time stamp */
 	uint32_t datlen;			/* Length of inode data */
+	uint16_t prev_parent;		/* To track the processed directory sector to find parent for entry. Holds last sctive sector in parent chain if new sector has tobe chained */
 };
 
 /* This is an on-device representation of the SMART inode which exists on
@@ -376,13 +381,13 @@ int smartfs_mount(struct smartfs_mountpt_s *fs, bool writeable);
 
 int smartfs_unmount(struct smartfs_mountpt_s *fs);
 
-int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry, const char *relpath, uint16_t *parentdirsector, const char **filename);
+int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry, const char *relpath, const char **filename);
 
-int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, struct smartfs_entry_s *direntry);
+int smartfs_createdirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry);
 
-int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdirsector, struct smartfs_entry_s *direntry);
+int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry);
 
-int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry, uint16_t parentdirsector);
+int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry);
 
 int smartfs_alloc_firstsector(struct smartfs_mountpt_s *fs, uint16_t *sector, uint16_t type, FAR struct smartfs_ofile_s *sf);
 
--- a/os/fs/smartfs/smartfs.h
+++ b/os/fs/smartfs/smartfs.h
@@ -225,8 +225,10 @@
 
 /* Buffer flags (when CRC enabled) */
 
+#define SMARTFS_BFLAG_UNMOD       0x00	/* Set if there are no unsynced changes in sf->buffer */
 #define SMARTFS_BFLAG_DIRTY       0x01	/* Set if data changed in the sector */
 #define SMARTFS_BFLAG_NEWALLOC    0x02	/* Set if sector not written since alloc */
+#define SMARTFS_BFLAG_NEW_ENTRY   0x04	/* Set if the open file structure object corresponds to a new file entry yet to be written to MTD */
 
 #define SMARTFS_ERASEDSTATE_16BIT (uint16_t)((CONFIG_SMARTFS_ERASEDSTATE << 8) | \
 								  CONFIG_SMARTFS_ERASEDSTATE)
@@ -398,9 +400,14 @@ int smartfs_deleteentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *en
 int smartfs_countdirentries(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *entry);
 
 int smartfs_truncatefile(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *entry, FAR struct smartfs_ofile_s *sf);
+
 int smartfs_shrinkfile(FAR struct smartfs_mountpt_s *fs, FAR struct smartfs_ofile_s *sf, off_t length);
+
 int smartfs_extendfile(FAR struct smartfs_mountpt_s *fs, FAR struct smartfs_ofile_s *sf, off_t length);
+
 void smartfs_setbuffer(struct smart_read_write_s *rw, uint16_t logsector, uint16_t offset, uint16_t count, uint8_t *buffer);
+
+void smartfs_set_entry_flags(struct smartfs_entry_s *new_entry, mode_t mode, uint16_t type);
 
 uint16_t smartfs_rdle16(FAR const void *val);
 
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -176,14 +176,11 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	struct inode *inode;
 	struct smartfs_mountpt_s *fs;
 	int ret;
-	uint16_t parentdirsector;
 	const char *filename;
 	struct smartfs_ofile_s *sf;
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 	struct smart_read_write_s readwrite;
 #endif
-	struct smartfs_entry_s newentry;
-	uint16_t data_sector;
 
 	/* Sanity checks */
 
@@ -226,8 +223,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
 	sf->entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &sf->entry, relpath, &parentdirsector, &filename);
-
+	ret = smartfs_finddirentry(fs, &sf->entry, relpath, &filename);
 	/* Three possibilities: (1) a node exists for the relpath and
 	 * dirinfo describes the directory entry of the entity, (2) the
 	 * node does not exist, or (3) some error occurred.
@@ -279,17 +275,17 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		}
 
 		/* Yes... test if the parent directory is valid */
-		if (parentdirsector != 0xFFFF) {
+		if (sf->entry.dsector != 0xFFFF) {
 			/* We can create in the given parent directory */
 			/* First we allocate a new data sector for the file */
-			ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_FILE, sf);
+			ret = smartfs_alloc_firstsector(fs, &sf->entry.firstsector, SMARTFS_DIRENT_TYPE_FILE, sf);
 			if (ret != OK) {
 				fdbg("Failed to allocate data scetor for entry\n");
 				goto errout_with_buffer;
 			}
 
 			/* First try to find an invalid or empty entry available in one of the chained parent sectors */
-			ret = smartfs_find_availableentry(fs, &parentdirsector, &newentry);
+			ret = smartfs_find_availableentry(fs, &sf->entry);
 			if (ret != OK) {
 				/* find_availableentry encountered a problem with lower layer operations, return error */
 				fdbg("smartfs_find_availableentry() encountered a problem, unable to find entry for writing\n");
@@ -297,9 +293,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 			}
 
 			/* At this point, either an available entry was found or a new one has been created */
-			newentry.firstsector = data_sector;
-
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			ret = smartfs_writeentry(fs, sf->entry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry);
 			if (ret != OK) {
 				fdbg("Write entry failed\n");
 				goto errout_with_buffer;
@@ -1324,7 +1318,6 @@ static int smartfs_opendir(struct inode *mountpt, const char *relpath, struct fs
 	struct smartfs_mountpt_s *fs;
 	int ret;
 	struct smartfs_entry_s entry;
-	uint16_t parentdirsector;
 	const char *filename;
 
 	/* Sanity checks */
@@ -1342,7 +1335,7 @@ static int smartfs_opendir(struct inode *mountpt, const char *relpath, struct fs
 	/* Search for the path on the volume */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
+	ret = smartfs_finddirentry(fs, &entry, relpath, &filename);
 	if (ret < 0) {
 		goto errout_with_semaphore;
 	}
@@ -1673,7 +1666,6 @@ static int smartfs_unlink(struct inode *mountpt, const char *relpath)
 	int ret;
 	struct smartfs_entry_s entry;
 	const char *filename;
-	uint16_t parentdirsector;
 
 	/* Sanity checks */
 
@@ -1688,8 +1680,7 @@ static int smartfs_unlink(struct inode *mountpt, const char *relpath)
 	/* Locate the directory entry for this path */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
-
+	ret = smartfs_finddirentry(fs, &entry, relpath, &filename);
 	if (ret == OK) {
 		/* The name exists -- validate it is a file, not a dir */
 
@@ -1731,11 +1722,9 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 {
 	struct smartfs_mountpt_s *fs;
 	int ret;
-	struct smartfs_entry_s entry; //Details of newly written entry are returned with this variable
-	uint16_t parentdirsector;
+	struct smartfs_entry_s entry;
+	struct smartfs_entry_s newentry;	//Details of newly written entry are returned with this variable
 	const char *filename;
-	struct smartfs_entry_s newentry;
-	uint16_t data_sector;
 
 	/* Sanity checks */
 
@@ -1750,7 +1739,8 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 	/* Locate the directory entry for this path */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
+	newentry.name = NULL;
+	ret = smartfs_finddirentry(fs, &entry, relpath, &filename);
 	/* Three possibililities: (1) a node exists for the relpath and
 	 * dirinfo describes the directory entry of the entity, (2) the
 	 * node does not exist, or (3) some error occurred.
@@ -1764,30 +1754,28 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 	} else if (ret == -ENOENT) {
 
 		/* It doesn't exist ... we can create it, but only if we have
-		 * the right permissions and if the parentdirsector is valid. */
+		 * the right permissions and if the previous parent is valid. */
 
-		if (parentdirsector == 0xFFFF) {
+		if (entry.dsector == 0xFFFF) {
 			fdbg("Invalid parent dir sector\n");
 			/* Invalid entry in the path (non-existant dir segment) */
 			goto errout_with_semaphore;
 		}
 		/* Create the directory */
 		/* Allocate new data sector for entry */
-		ret = smartfs_alloc_firstsector(fs, &data_sector, SMARTFS_DIRENT_TYPE_DIR, NULL);
+		ret = smartfs_alloc_firstsector(fs, &entry.firstsector, SMARTFS_DIRENT_TYPE_DIR, NULL);
 		if (ret != OK) {
 			fdbg("Failed to allocate data sector for entry\n");
 			goto errout_with_semaphore;
 		}
 		/* Try to find empty/invalid entry available in one of he chained parent sectors */
-		ret = smartfs_find_availableentry(fs, &parentdirsector, &newentry);
+		ret = smartfs_find_availableentry(fs, &entry);
 		if (ret != OK) {
 			fdbg("find_availableentry failed, cannot find entry for writing\n");
 			goto errout_with_semaphore;
 		}
-		/* Now we have an entry allocated for writing */
-		newentry.firstsector = data_sector;
-		/* Write new entry to sector */
-		ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &entry, parentdirsector);
+		/* Now we have an entry allocated for writing, write new entry to sector */
+		ret = smartfs_writeentry(fs, entry, filename, SMARTFS_DIRENT_TYPE_DIR, mode, &newentry);
 		if (ret != OK) {
 			fdbg("Failed to write new entry to sector\n");
 			goto errout_with_semaphore;
@@ -1795,11 +1783,11 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 	}
 
 errout_with_semaphore:
-	if (entry.name != NULL) {
+	if (newentry.name != NULL) {
 		/* Free the filename space allocation */
 
-		kmm_free(entry.name);
-		entry.name = NULL;
+		kmm_free(newentry.name);
+		newentry.name = NULL;
 	}
 
 	smartfs_semgive(fs);
@@ -1819,7 +1807,6 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
 	int ret;
 	struct smartfs_entry_s entry;
 	const char *filename;
-	uint16_t parentdirsector;
 
 	/* Sanity checks */
 
@@ -1836,8 +1823,7 @@ int smartfs_rmdir(struct inode *mountpt, const char *relpath)
 	/* Locate the directory entry for this path */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
-
+	ret = smartfs_finddirentry(fs, &entry, relpath, &filename);
 	if (ret == OK) {
 		/* The name exists -- validate it is a dir, not a file */
 
@@ -1896,19 +1882,16 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	int ret;
 	struct smartfs_mountpt_s *fs;
 	struct smartfs_entry_s oldentry;
-	uint16_t oldparentdirsector;
 	const char *oldfilename;
 	struct smartfs_entry_s newentry;
-	uint16_t newparentdirsector;
 	const char *newfilename;
 	mode_t mode;
 	uint16_t type;
 	uint16_t sector;
 	uint16_t offset;
 	uint16_t entrysize;
-	struct smartfs_entry_header_s *direntry;
+	struct smartfs_entry_header_s *entry;
 	struct smart_read_write_s readwrite;
-	struct smartfs_entry_s f_newentry;
 
 	/* Sanity checks */
 
@@ -1924,7 +1907,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 
 	oldentry.name = NULL;
 	newentry.name = NULL;
-	ret = smartfs_finddirentry(fs, &oldentry, oldrelpath, &oldparentdirsector, &oldfilename);
+	ret = smartfs_finddirentry(fs, &oldentry, oldrelpath, &oldfilename);
 	if (ret < 0) {
 		fdbg("Old entry doesn't exist\n");
 		goto errout_with_semaphore;
@@ -1939,7 +1922,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	 * doens't exsit.
 	 */
 
-	ret = smartfs_finddirentry(fs, &newentry, newrelpath, &newparentdirsector, &newfilename);
+	ret = smartfs_finddirentry(fs, &newentry, newrelpath, &newfilename);
 	if (ret == OK) {
 		/* Test if it's a file.  If it is, then it's an error */
 
@@ -1969,9 +1952,9 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 			while (offset + entrysize < fs->fs_llformat.availbytes) {
 				/* Test the next entry */
 
-				direntry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
+				entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
 				/* Check if entry flags are erased OR empty OR inactive and skip the entry */
-				if ((direntry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((direntry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((direntry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
+				if ((entry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((entry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((entry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
 					/* This entry isn't valid, skip it */
 
 					offset += entrysize;
@@ -1980,7 +1963,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 
 				/* Test if this entry matches newfilename */
 
-				if (strcmp(newfilename, direntry->name) == 0) {
+				if (strcmp(newfilename, entry->name) == 0) {
 					/* Uh-oh, looks like the entry already exists */
 
 					ret = -EEXIST;
@@ -1996,20 +1979,20 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		}
 
 		/* Okay, we will create newfilename in the newentry directory */
-		newparentdirsector = newentry.firstsector;
+		newentry.dsector = newentry.firstsector;
 	}
 
 	/* Test if the new parent directory is valid */
-	if (newparentdirsector != 0xFFFF) {
+	if (newentry.dsector != SMARTFS_ERASEDSTATE_16BIT) {
 		/* We can move to the given parent directory */
 
 		mode = oldentry.flags & SMARTFS_DIRENT_MODE;
 		type = oldentry.flags & SMARTFS_DIRENT_TYPE;
 
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
-		if (oldparentdirsector == newparentdirsector) {
+		if (oldentry.dsector == newentry.dsector) {
 			/* We will not use any new entry found, we will overwrite the existing entry */
-			ret = smartfs_writeentry(fs, oldentry, newfilename, type, mode, &newentry, oldparentdirsector);
+			ret = smartfs_writeentry(fs, oldentry, newfilename, type, mode, &newentry);
 			if (ret != OK) {
 				fdbg("Error writing new entry\n");
 			}
@@ -2018,14 +2001,14 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		}
 #endif
 		/* Find an available invalid/empty entry to write the new entry */
-		ret = smartfs_find_availableentry(fs, &newparentdirsector, &f_newentry);
+		ret = smartfs_find_availableentry(fs, &newentry);
 		if (ret != OK) {
 			fdbg("find_availableentry encountered a problem, cannot find entry for writing\n");
 			goto errout_with_semaphore;
 		}
 
-		f_newentry.firstsector = oldentry.firstsector;
-		ret = smartfs_writeentry(fs, f_newentry, newfilename, type, mode, &newentry, newparentdirsector);
+		newentry.firstsector = oldentry.firstsector;
+		ret = smartfs_writeentry(fs, newentry, newfilename, type, mode, &newentry);
 		if (ret != OK) {
 			fdbg("Error writing new entry\n");
 			goto errout_with_semaphore;
@@ -2106,7 +2089,6 @@ static int smartfs_stat(struct inode *mountpt, const char *relpath, struct stat 
 	struct smartfs_mountpt_s *fs;
 	struct smartfs_entry_s entry;
 	int ret;
-	uint16_t parentdirsector;
 	const char *filename;
 
 	/* Sanity checks */
@@ -2123,7 +2105,7 @@ static int smartfs_stat(struct inode *mountpt, const char *relpath, struct stat 
 	/* Find the directory entry corresponding to relpath */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
+	ret = smartfs_finddirentry(fs, &entry, relpath, &filename);
 	if (ret < 0) {
 		goto errout_with_semaphore;
 	}
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -201,7 +201,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 
 	/* Locate the directory entry for this path */
 
-	sf = (struct smartfs_ofile_s *)kmm_malloc(sizeof * sf);
+	sf = (struct smartfs_ofile_s *)kmm_zalloc(sizeof(struct smartfs_ofile_s));
 	if (sf == NULL) {
 		ret = -ENOMEM;
 		goto errout_with_semaphore;
@@ -219,7 +219,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		goto errout_with_semaphore;
 	}
 
-	sf->bflags = 0;
+	sf->bflags = SMARTFS_BFLAG_UNMOD;
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
 	sf->entry.name = NULL;
@@ -292,12 +292,37 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 				goto errout_with_buffer;
 			}
 
+#if defined(CONFIG_SMARTFS_USE_SECTOR_BUFFER) && !defined(NXFUSE_HOST_BUILD)
+			/* If CRC is enabled, hold the entry for writing later */
+			/* Mark flags to indicate that the entry will be written to a new sector in the parent directory.
+			 * This sector will then be chained in the end.
+			 */
+			sf->bflags |= SMARTFS_BFLAG_NEW_ENTRY;
+			/* Save entry mode to write to MTD later */
+			sf->entry.flags = SMARTFS_ERASEDSTATE_16BIT;
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
+			smartfs_wrle16(&sf->entry.flags, (uint16_t)(mode & SMARTFS_DIRENT_MODE));
+#else
+			sf->entry.flags = (uint16_t)(mode & SMARTFS_DIRENT_MODE);
+#endif
+			if (sf->entry.name == NULL) {
+				sf->entry.name = (FAR char *)kmm_malloc(fs->fs_llformat.namesize + 1);
+				if (sf->entry.name == NULL) {
+					ret = -ENOMEM;
+					goto errout_with_buffer;
+				}
+			}
+			memset(sf->entry.name, 0, fs->fs_llformat.namesize + 1);
+			strncpy(sf->entry.name, filename, fs->fs_llformat.namesize);
+#else
+			/* If CRC is disabled, write the new file entry */
 			/* At this point, either an available entry was found or a new one has been created */
 			ret = smartfs_writeentry(fs, sf->entry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry);
 			if (ret != OK) {
 				fdbg("Write entry failed\n");
 				goto errout_with_buffer;
 			}
+#endif
 		} else {
 			/* Trying to create in a directory that doesn't exist */
 
@@ -639,7 +664,16 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 		}
 
 		sf->byteswritten = 0;
-		sf->bflags = 0;
+		/* File's data sector has been synced with MTD, now check if this is a new file entry and write the entry */
+		if (sf->bflags & SMARTFS_BFLAG_NEW_ENTRY) {
+			/* Flags for this entry have already been set and stored, so mode will not be used */
+			ret = smartfs_writeentry(fs, sf->entry, sf->entry.name, SMARTFS_DIRENT_TYPE_FILE, (sf->entry.flags & SMARTFS_DIRENT_MODE), &sf->entry);
+			if (ret < 0) {
+				fdbg("Failed to write new file entry ot MTD\n");
+				goto errout;
+			}
+		}
+		sf->bflags = SMARTFS_BFLAG_UNMOD;
 	}
 #else							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -1284,12 +1284,12 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 		   Chainheader is written along with the 1st entry.
 		   So currently, smart_readsector will fail to invalidate CRC on empty sector
 		 */
-		tmp_buf = (char *)kmm_malloc(entrysize + sizeof(struct smartfs_entry_header_s));
+		tmp_buf = (char *)kmm_malloc(entrysize + sizeof(struct smartfs_chain_header_s));
 		if (tmp_buf == NULL) {
 			fdbg("Unable to allocate memory\n");
 			return -ENOMEM;
 		}
-		memset(tmp_buf, CONFIG_SMARTFS_ERASEDSTATE, entrysize + sizeof(struct smartfs_entry_header_s));
+		memset(tmp_buf, CONFIG_SMARTFS_ERASEDSTATE, entrysize + sizeof(struct smartfs_chain_header_s));
 		entry = (struct smartfs_entry_header_s *)&tmp_buf[sizeof(struct smartfs_chain_header_s)];
 		chainheader = (struct smartfs_chain_header_s *)&tmp_buf[0];
 		chainheader->type = SMARTFS_SECTOR_TYPE_DIR;
@@ -1483,7 +1483,7 @@ int smartfs_invalidateentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsect
 
 	entry_flags = (uint8_t *)&fs->fs_rwbuffer[0];
 #if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-#if CONFIG_SMARTFS_ALIGNED_ACCESS
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
 	smartfs_wrle16(entry_flags, smartfs_rdle16(entry_flags) & ~SMARTFS_DIRENT_ACTIVE);
 #else
 	*entry_flags &= ~SMARTFS_DIRENT_ACTIVE;
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -791,14 +791,14 @@ int smartfs_unmount(struct smartfs_mountpt_s *fs)
  *
  *              If the final directory segment of relpath just before the
  *              last segment (the target file/dir) is valid, then the
- *              parentdirsector will indicate the logical sector number of
- *              the parent directory where a new entry should be created,
+ *              direntry->prev_parent will indicate the logical sector number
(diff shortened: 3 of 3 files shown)


os/fs/smartfs/smartfs_smart.c
```diff
@@ -747,6 +750,17 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 		goto errout_with_semaphore;
 	}
 
+	/* Allocate memory for buffer space to write entry to MTD later and return if space can't be allocated */
+	entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
+	tmp_buf = (char *)kmm_malloc(entrysize);
+	if (tmp_buf == NULL) {
+		fdbg("Unable to allocate memory\n");
+		ret -ENOMEM;
```
amandeepc, line 758: Assignment (=) is missing i think

amogh-samsung: Yes Done

os/fs/smartfs/smartfs_smart.c
```diff
@@ -298,8 +298,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 
 			/* At this point, either an available entry was found or a new one has been created */
 			newentry.firstsector = data_sector;
-
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector, sf);
```
Taejun-Kwon, line 301: We'll write entry data in parent sector after first sector written.
So here we don't have to call writeentry.
We can call writeentry after sync_internal.
Please check below.
```
ret = smartfs_find_availableentry(fs, &parentdirsector, &sf->entry);
...
smartfs_wrle32(sf->entry->utc, fs->entry_seq++);
/* add below if we need it */
memset(direntry->name, 0, fs->fs_llformat.namesize + 1); 
strncpy(direntry->name, filename, fs->fs_llformat.namesize);
```

os/fs/smartfs/smartfs_smart.c
```diff
@@ -747,6 +751,19 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 		goto errout_with_semaphore;
 	}
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
Taejun-Kwon, line 754: I think added code in this function is not necessary.
in smartfs_sync_internal, we can add below of line 648, I think.
```
smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
if (ret < 0) {
	fdbg("Error %d writing used bytes for sector %d\n", ret, sf->currsector);
	goto errout;
}

sf->byteswritten = 0;
sf->bflags = 0;

/* If it is firstsector, update entry */
if (sf->entry->datlen == fs->fs_llformat.availbytes + sizeof(struct smartfs_chain_header_s)) {
      mode = entry->flag | .... // expand entry struct????.
       smartfs_writeentry(fs, sf->entry, sf->entry->name, SMARTFS_DIRENT_TYPE_FILE, mode, entry->dsector);
}
```

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1258,17 +1258,14 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
  *
  ****************************************************************************/
 
-int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry, uint16_t parentdirsector)
+int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_entry, const char *filename, uint16_t type, mode_t mode, struct smartfs_entry_s *direntry, uint16_t parentdirsector, struct smartfs_ofile_s *sf)
```
Taejun-Kwon, line 1261: so we don't have to expand parameter....please check above comment.

Taejun-Kwon:
I think there's no way to handle chainheader of file, I mean if we append 1000bytes to exist file(502bytes),
There's no way avoid write nextsector in header of 1st, 2nd sector eariler than entire of 2nd 3rd dat asector.
You have any idea>?

os/fs/smartfs/smartfs.h
```diff
@@ -221,8 +221,9 @@
 
 /* Buffer flags (when CRC enabled) */
 
-#define SMARTFS_BFLAG_DIRTY       0x01	/* Set if data changed in the sector */
-#define SMARTFS_BFLAG_NEWALLOC    0x02	/* Set if sector not written since alloc */
+#define SMARTFS_BFLAG_DIRTY		0x01	/* Set if data changed in the sector */
+#define SMARTFS_BFLAG_NEWALLOC		0x02	/* Set if sector not written since alloc */
```
Taejun-Kwon, line 225: 1. Code align is needed here.
2. To use this values with bit operation, 0x04 is correct for new one.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -298,12 +298,51 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+				   This sector will then be chained in the end.
+				 */
+				sf->bflags |= SMARTFS_BFLAG_NEW_DIRSECTOR;
+				sf->last_psector = parentdirsector;
+			}
 
+			sf->entry.firstsector = newentry.firstsector;
+			sf->entry.dsector = newentry.dsector;
+			sf->entry.doffset = newentry.doffset;
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
```
Taejun-Kwon, line 314: utc number will be increased when it written by writeentry..
Or you can this in find_availableentry, I think.
name also can be set there.
Then I estimate that we can use sf->entry...

os/fs/smartfs/smartfs_smart.c
```diff
@@ -949,9 +1006,55 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 	}
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
Taejun-Kwon, line 1009: Do not write entry here, entry should be written after data sector's sync finished.
This will write entry for every write call..
Below is very simple copy operation, in this time this will
```
uint8_t buf = mallioc(st.st_blksize);
for (int i = 0; i < 16; i++) {
    read(fd1, buf, sizeof(buf);
    write(fd2, buf, sizeof(buf); //append data
}
close(fd1);
close(fd2);
```
In above case it will write entry 16times.

amogh-samsung: This has not been updated yet.
Pending changes

os/fs/smartfs/smartfs.h
amogh-samsung: Updated

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Modified

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Modified

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Modified

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Updated

os/fs/smartfs/smartfs.h
```diff
@@ -311,6 +313,7 @@ struct smartfs_ofile_s {
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 	uint8_t *buffer;			/* Sector buffer to reduce writes */
 	uint8_t bflags;				/* Buffer flags */
+	uint16_t last_psector;			/* Last sector in old parent directory chain, will be used to update chain after writing entry */
```
Taejun-Kwon, line 316: it is useful for only entry, I think you'd better expand entry.
```
struct smartfs_entry_s {
	uint16_t firstsector;		/* Sector number of the name */
	uint16_t dsector;			/* Sector number of the directory entry */
	uint16_t doffset;			/* Offset of the directory entry */
	uint16_t dfirst;			/* 1st sector number of the directory entry */
        uint16_t dprev;                /* Previous entry sector that is need to be chained with our parent */
	uint16_t flags;				/* Flags, including mode */
	FAR char *name;				/* inode name */
	uint32_t utc;				/* Time stamp */
	uint32_t datlen;			/* Length of inode data */
};
```
or keep current and below(we optimized read(memcpy in progmem) performance of target board 3times faster)
```
if (sf->entry->dsector != sf->entry->dfirst) {
     smartfs_setbuffer(&readwrite, entry->dsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
     FS_IOCTL(READ);
     while (nextsector_is_not_0xff) {
        if (SMARTFS_NEXTSECTOR(header) == 0xff) {
            write_next_header;
        }
      }
}
```

os/fs/smartfs/smartfs_smart.c
```diff
@@ -693,7 +753,66 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 	}
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	/* File's data sector has been synced with MTD, now check if this is a new file entry and write the entry */
+	if (sf->bflags && SMARTFS_BFLAG_NEW_ENTRY) {
```
Taejun-Kwon, line 758: please check how sf->bflags is changed.
I think only first time it include SMARTFS_BFLAG_NEWALLOC after second time, it doesn't have it.
Because after sync sf->bflags is asigned as SMARTFS_BFLAG_DIRTY.
so below will be work. please check it.
```
if (sf->bflags & SMARTFS_BFLAG_NEWALLOC) {
     write_first_data_sector;
      write_entry;
}

File: os/fs/smartfs/smartfs_smart.c [hunk 10]
```diff
@@ -693,7 +753,66 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
...
+		smartfs_wrle16(&entry->firstsector, sf->entry.firstsector);
+#ifdef CONFIG_SMARTFS_ENTRY_TIMESTAMP
+		smartfs_wrle32((uint8_t*)&entry->utc, fs->entry_seq++);
+#else
+		smartfs_wrle32((uint8_t*)&entry->utc, time(NULL));
+#endif
+#else
+		entry->firstsector = sf->entry.firstsector;
+#ifdef CONFIG_SMARTFS_ENTRY_TIMESTAMP
+		entry->utc = fs->entry_seq++;
```
Taejun-Kwon, line 782: so, open -> sync will inrease sequence number 2 times.
can't we do this kind of 'set attribute value of entry' in find_available_entry ?

os/fs/smartfs/smartfs_smart.c
```diff
@@ -693,7 +753,66 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
...
+#else
+		entry->utc = time(NULL);
+#endif
+#endif
+		memset(entry->name, 0, fs->fs_llformat.namesize);
+		strncpy(entry->name, sf->entry.name, fs->fs_llformat.namesize);
+		entry->flags = sf->entry.flags;
+
+		smartfs_setbuffer(&readwrite, sf->entry.dsector, sf->entry.doffset, entrysize, (uint8_t *)&tmp_buf[0]);
```
Taejun-Kwon, line 791: very similar with writeentry, Please think more how we can optimize writeentry.

os/fs/smartfs/smartfs.h
amogh-samsung: Adding last_psector to smartfs_entry_s seems meaningless.
last_psector has meaning only till the new sector is not chained to the parent directory.

Regarding the loop that you suggested, I made some more changes to align code with writeentry.
So last_psector is needed to store parendirsector because it is passed as a parameter to writeentry.
Could you please take a look once again?

os/fs/smartfs/smartfs_smart.c
amogh-samsung: bflags |= NEW_ENTRY will be done only when a new entry is created for the file.
This will be done in smartfs_open

During sync, this entry will be written to MTD,
After that, checking for new entry is no longer required.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Modified

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Modified to use writeentry code.

os/fs/smartfs/smartfs.h
```diff
@@ -221,8 +221,9 @@
 
 /* Buffer flags (when CRC enabled) */
 
-#define SMARTFS_BFLAG_DIRTY       0x01	/* Set if data changed in the sector */
-#define SMARTFS_BFLAG_NEWALLOC    0x02	/* Set if sector not written since alloc */
+#define SMARTFS_BFLAG_DIRTY			0x01	/* Set if data changed in the sector */
+#define SMARTFS_BFLAG_NEWALLOC			0x02	/* Set if sector not written since alloc */
```
sangam-swami, line 225: maintain same tab

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,40 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
 			}
 
 			/* At this point, either an available entry was found or a new one has been created */
-			newentry.firstsector = data_sector;
-
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			sf->entry.firstsector = data_sector;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			/* If CRC is enabled, hold the entry for writing later */
+			/* Mark flags to indicate that the entry will be written to anew sector in the parent directory.
```
thapav, line 302: a new*

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1369,13 +1364,13 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 	direntry->firstsector = new_entry.firstsector;
 	direntry->dsector = psector;
 	direntry->doffset = offset;
+
```
thapav, line 1367: is this needed ?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1496,7 +1492,7 @@ int smartfs_invalidateentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsect
 
 	entry_flags = (uint8_t *)&fs->fs_rwbuffer[0];
 #if CONFIG_SMARTFS_ERASEDSTATE == 0xFF
-#if CONFIG_SMARTFS_ALIGNED_ACCESS
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
```
thapav, line 1495: you can add this change separately under a different commit (refactoring)

os/fs/smartfs/smartfs.h
Taejun-Kwon: Add to entry is correct, in my opinion, because  parent sector(previous) sector's is necessary only for part of entry process.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -693,6 +715,18 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
...
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	/* File's data sector has been synced with MTD, now check if this is a new file entry and write the entry */
+	if (sf->bflags & SMARTFS_BFLAG_NEW_ENTRY) {
+		ret = smartfs_writeentry(fs, sf->entry, sf->entry.name, SMARTFS_DIRENT_TYPE_FILE, sf->mode, &sf->entry, sf->last_psector);
+		if (ret < 0) {
+			fdbg("Failed to write new file entry ot MTD\n");
+			goto errout;
+		}
+		sf->bflags = 0;
```
Taejun-Kwon, line 726: I think you can make nice name for default one.
#define SMARTFS_BFLAG_XXX  0

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,40 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
-
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			sf->entry.firstsector = data_sector;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			/* If CRC is enabled, hold the entry for writing later */
+			/* Mark flags to indicate that the entry will be written to anew sector in the parent directory.
+			   This sector will then be chained in the end.
+			 */
+			sf->last_psector = parentdirsector;
+			sf->bflags |= SMARTFS_BFLAG_NEW_ENTRY;
```
Taejun-Kwon, line 306: Good. this is correct approach.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1297,12 +1294,12 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 		   Chainheader is written along with the 1st entry.
 		   So currently, smart_readsector will fail to invalidate CRC on empty sector
 		 */
-		tmp_buf = (char *)kmm_malloc(entrysize + sizeof(struct smartfs_entry_header_s));
+		tmp_buf = (char *)kmm_malloc(entrysize + sizeof(struct smartfs_chain_header_s));
```
Taejun-Kwon, line 1297: There was bug! very good..

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,40 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+			sf->mode = mode;
+			if (sf->entry.name == NULL) {
```
Taejun-Kwon, line 308: hmm ok, in next pr please think how we can reduce duplicated code here =)
(here & writeentry)
Then regardless of crc, we can use it.

amogh-samsung: Yes, I want to avoid so many code lines for allocating space for name.
Let's try for a macro/function in the future.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes but probably should be done in a separate commit as suggested by @thapav in previous comments.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkay, let's do that.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Updated

os/fs/smartfs/smartfs.h
amogh-samsung: Done

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Done

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Can be done away with.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Ohkay

os/fs/smartfs/smartfs.h
```diff
@@ -221,8 +221,10 @@
 
 /* Buffer flags (when CRC enabled) */
 
-#define SMARTFS_BFLAG_DIRTY       0x01	/* Set if data changed in the sector */
-#define SMARTFS_BFLAG_NEWALLOC    0x02	/* Set if sector not written since alloc */
+#define SMARTFS_BFLAG_UNMOD		0x00	/* Indicates that the buffer in unmodified, no changes to be recorded in MTD */
```
Taejun-Kwon, line 224: Usually use 'ideal' 'clean' 'empty' or something has similar meaning with this, but It is first time that I saw kind of this name.
Is this better than all of these?

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: I mean I think this will be set 2times, here and writeentry.
Instead of using the MACRO, we'd better remove it from somewhere.

os/fs/smartfs/smartfs.h
amogh-samsung: UNMOD -> Un Modified
Since we are recording changes in the sf buffer to MTD
Buffer flags set as Un Modified means no changes in buffer to record to MTD.
I think this name is very apt.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Here, allocating space for entry name is  necessary to store the entry name for writing later.
If the space is allocated here, it will not be allocated in writeentry for direntry.

However, before line 1334 of smartfs_utils.c
I think space needs to be allocated for entry->name.
We might have missed out on that...

Taejun-Kwon: It is needed only to create new entry, so If we do this in smartfs_find_availableentry, we can reduce it. please consider it.
Also smartfs_createentry is needed to create new entry sector when it couldn't find entry parent directory sector , so rename is needed smartfs_createdirentry (I think I already reviewed it).

os/fs/smartfs/smartfs.h
Taejun-Kwon: got it. but description need to be changed, 'Set if ....(To represent initial state & can be set if we don't want write this entry....blahblah)

os/fs/smartfs/smartfs.h
```diff
@@ -221,8 +221,10 @@
...
+#define SMARTFS_BFLAG_DIRTY		0x01	/* Set if data changed in the sector */
+#define SMARTFS_BFLAG_NEWALLOC		0x02	/* Set if sector not written since alloc */
```
Taejun-Kwon, line 226: please check space.

os/fs/smartfs/smartfs.h
amogh-samsung: I think the description is alright too.
Same as the above comment...

os/fs/smartfs/smartfs_smart.c
amogh-samsung: smartfs_find_availableentry only finds the space for writing new entry.
Allocating space for new entry->name is done while writing only.

So this should not be done in find_availableentry.

`filename `is copied to RAM buffer only when we are populating attributes of entry to be written.
Before that let's not allocate entry->name separately.

os/fs/smartfs/smartfs.h
amogh-samsung: Ohkkay I can align it with other descriptions: -
"Set if ......"

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: I don't think so, 'creating entry' means that find space & make entry struct.
Entry can be hold both of flash or RAM. Even if we store it in RAM, it already completed entry.
So you can do something more here =)
Please do not restrict to use current code only.
You can rename current function & expand it, or add more function.
Current writeentry doing 2 roles, update entry info and write entry data on flash.
How to optimize below?
```
main()
{
    func_a();
    func_b();
    if (condition == true) {
        func_c(); // function_c set data only
        func_d();
    } else {
        func_e();
    }
    ...
    if (condition == true) {
        func_e();
    }
}

func_e()
{
    func_c();
    func_f();
}
```
Expand or rename function a or b, otherwise add new function g, which can call func_a, func_b, func_c will be one of answer.

os/fs/smartfs/smartfs.h
Taejun-Kwon: please check above comment to.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -693,6 +714,18 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 	}
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
Taejun-Kwon, line 717: I think these codes can be optimized with codes in line 665

os/fs/smartfs/smartfs.h
```diff
@@ -258,6 +260,9 @@ struct smartfs_entry_s {
 	uint16_t dsector;			/* Sector number of the directory entry */
 	uint16_t doffset;			/* Offset of the directory entry */
 	uint16_t dfirst;			/* 1st sector number of the directory entry */
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	uint16_t last_psector;		/* Last sector in old parent directory chain, will be used to update chain after writing entry */
```
Taejun-Kwon, line 264: How about below?
```
uint16_t prev_'d'sector;     /* Directory entry which will be chained to current dsector at the last time */
```

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1387,6 +1380,7 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 
 	memset(direntry->name, 0, fs->fs_llformat.namesize + 1);
 	strncpy(direntry->name, filename, fs->fs_llformat.namesize);
+	direntry->flags = new_entry.flags;
```
Taejun-Kwon, line 1383: regarding direntry->name here, you gonna update in next PR, right?
at that time please consider we have to reveal direntry here.. I think we can do this in previous function as per I told you.
Then some of parameter also need to modified.

os/fs/smartfs/smartfs.h
```diff
@@ -311,6 +316,7 @@ struct smartfs_ofile_s {
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 	uint8_t *buffer;			/* Sector buffer to reduce writes */
 	uint8_t bflags;				/* Buffer flags */
+	mode_t mode;
```
Taejun-Kwon, line 319: good, I forgot regarding mode, but this also part of entry.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes, changes for allocation or direntry->name in the next PR.

os/fs/smartfs/smartfs.h
amogh-samsung: Changing this now

os/fs/smartfs/smartfs.h
amogh-samsung: Minor change, can do this before merging once logic is complete.

os/fs/smartfs/smartfs.h
amogh-samsung: Can take care of this before merging.
Some more critical logical changes first.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Sorry could you explain please?

amogh-samsung: Ohkay you mean adding this to the part with
`#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER`
Because otherwise this part of code is not used.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1125,13 +1151,13 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-int smartfs_createentry(struct smartfs_mountpt_s *fs, uint16_t parentdirsector, struct smartfs_entry_s *direntry)
+int smartfs_createdirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry)
 {
 	int ret;
 	uint16_t nextsector;
 
-	/* If we are in the context of smartfs_createentry, then smartfs_find_availableentry did not find an available entry.
-	   Createentry has to chain a new nextsector to parentdirsector, and allocate the first entry.
+	/* If we are in the context of smartfs_createdirentry, then smartfs_find_availableentry did not find an available entry.
+	   Createdirentry has to allocate a new nextsector for chaining to parentdirsector, and allocate the first entry.
```
Taejun-Kwon, line 1160: So here we can set 
```
direntry->last_dsector = direntry->dsector;
direntry->dsector = nextsector;
```
and when it sync or finish entry process,
```
if (direntry->last_dsector != direntry->dsector) {
    update_chain(direntry->last_dsector, direntry->dsector); 
}
```

os/fs/smartfs/smartfs.h
```diff
@@ -258,6 +260,12 @@ struct smartfs_entry_s {
 	uint16_t dsector;			/* Sector number of the directory entry */
 	uint16_t doffset;			/* Offset of the directory entry */
 	uint16_t dfirst;			/* 1st sector number of the directory entry */
+	uint16_t last_scanned_dsector;	/* Last directory sector in old parent chain that was checked for new entry.
```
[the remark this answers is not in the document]
amogh-samsung: Won't `last_dsector` and `dsector` become confusing?
I want the name to reflect that this is the last sector in the path of the new entry that has been traversed for writing new entry.
While, dsector will be the final parent directory sector where the entry will be written.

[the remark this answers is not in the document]
Taejun-Kwon: I think it is ok. basically use dsector, but if dsector changed to number of new entry sector, at that time last_dsector hold previous value.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -334,6 +334,32 @@ void smartfs_setbuffer(struct smart_read_write_s *rw, uint16_t logsector, uint16
...
+		if (smartfs_rdle16(&new_entry.flags) != SMARTFS_ERASEDSTATE_16BIT) {
+#else
+		if (new_entry.flags != SMARTFS_ERASEDSTATE_16BIT) {
+#endif
+			RESET_ENTRY_FLAGS(&new_entry);
+		}
+		SET_ENTRY_FLAGS(new_entry.flags, mode, type);
+	}
+
+	return;
```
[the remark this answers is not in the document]
amogh-samsung: 2 would actually be a problem, I was analyzing that before I made this change.

Let's go with the first one then...

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I think this is related to the first comment above?

> last_dsector is enough I think.
Actually dsector is parent entry, so instead of this, you'd better use dsector.

Or you  are referring to entry finding loop in find_availableentry?

Taejun-Kwon: both !

os/fs/smartfs/smartfs.h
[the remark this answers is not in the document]
amogh-samsung: Can we keep it this way?
-> Use last_dsector to track sector numbers
-> update dsector when entry is allocated
-> Compare last_dsector and dsector to check for new chains

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Ahh, ohkkay :P

os/fs/smartfs/smartfs.h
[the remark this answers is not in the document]
Taejun-Kwon: In my opinion, use dsector is better, for both of chain & none-chain case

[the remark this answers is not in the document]
thapav: Generally what is the terminology followed in cases where one variable is used to track a previous value ?
We can go ahead with a standard one to keep it same throughout Smartfs/TizenRT implementations.

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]
amogh-samsung: Ohkkay, for type 1 that you metioned above,
Macro ENTRY_VALID() is not available in smartfs_smart.c

Now I remember why I didn't do it earlier...

What do you think, we go with type 2?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1211,22 +1230,21 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
...
+				direntry->last_dsector = direntry->dsector;
 				return OK;
 			}
 
 			/* Not an available entry, check the next one */
 			offset += entrysize;
 			entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
 		}
-		direntry->dsector = psector;
+		direntry->last_dsector = psector;
```
Taejun-Kwon, line 1241: dsector here, right?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1276,33 +1291,24 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
...
-			RESET_ENTRY_FLAGS(&new_entry);
-		}
-		SET_ENTRY_FLAGS(new_entry.flags, mode, type);
-	}
+	smartfs_set_entry_flags(new_entry, type, mode);
 
-	/* If passed parameter parentdirsector != new_entry.dsector, it means it is a new chain sector that we will write to */
-	if (parentdirsector != psector) {
+	/* If passed parameter new_entry.last_dsector != new_entry.dsector, it means it is a new chain sector that we will write to */
+	if (last_parentdirsector != psector) {
```
Taejun-Kwon, line 1300: if (dsector != last_dsector) is correct approach, am I right?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1211,22 +1230,21 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
 				direntry->dsector = psector;
 				direntry->doffset = offset;
 				direntry->flags = entry->flags;
-				*parentdirsector = direntry->dsector;
+				direntry->last_dsector = direntry->dsector;
```
Taejun-Kwon, line 1233: please move this to between line 1230~1231 and below
```
/* initialize dsector & last_dsector as psector, dsector will be changed to new entry sector if current entry sector is full */
direntry->dsector = psector;
direntry->last_dsector = psector;

File: os/fs/smartfs/smartfs_utils.c [hunk 32]
```diff
@@ -1339,7 +1345,7 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 	entry->flags = new_entry.flags;
 
 	/* Now write the new entry to the parent directory sector */
-	if (parentdirsector != psector) {
+	if (last_parentdirsector != psector) {
```
Taejun-Kwon, line 1348: same comment in line 1300

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1351,13 +1357,11 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 		fdbg("failed to write new entry to parent directory psector : %d\n", new_entry.dsector);
 		goto errout;
 	}
-
-	if (parentdirsector != psector) {
-
+	if (last_parentdirsector != psector) {
```
Taejun-Kwon, line 1360: same comment in line 1300 & 1348

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Using aliases for both as is the general tradition in all functions

os/fs/smartfs/smartfs_utils.c
amogh-samsung: It is the same check.
Using alias variables for constant comparison as done in other fs layer functions.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Alias variables

os/fs/smartfs/smartfs_utils.c
amogh-samsung: No this is last_dsector only
If code comes here, it means no entry found.

So createdirentry below will update dsector.

It is the same as using dsector here then
Last_dsector = dsector outside the loop

os/fs/smartfs/smartfs_utils.c
amogh-samsung: If this part of code is reached
It means entry has been found. 
So this comment is not right.
After this line, function will return and not look further.

amogh-samsung: Dsector will not need to be changed because: -
Line 1229 `if (IS_AVAILABLE_ENTRY(entry)) {`

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: yes I understand what you mean, but please see below
```
while (psector != SMARTFS_ERASEDSTATE_16BIT) {
	while ((offset + entrysize) < readwrite.count) {
		/* Check if this entry is available */
		if (IS_AVAILABLE_ENTRY(entry)) {
			direntry->doffset = offset;
			direntry->flags = entry->flags;
			return OK;
		}

		/* Not an available entry, check the next one */
		offset += entrysize;
		entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
	}

	direntry->dsector = psector;
	psector = nextsector;
}

/* We couldn't find entry, so create new directory entry */
direntry->last_dsector = direntry->dsector; // can do this in smartfs_createentry
ret = smartfs_createentry(fs, direntry);
if (ret != OK) {
	fdbg();
}

return ret;
```

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: As I understand, some variable like a entry(mostly sf->entry) are added to prevent corruption by error which happened in the middle of function.
But It seems that dsector(psector) & last_dsector(last_parentdirsector) is not changed here....

os/fs/smartfs/smartfs_utils.c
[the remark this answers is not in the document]

os/fs/smartfs/smartfs_utils.c
amogh-samsung: This is what I mentioned
Same as current implementation, if you want I can do this hut it is only extra few lines.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: I'd like to take this up in a separate set of changes then.
Depending on how all the functions use aliases and how they can be optimized.

amogh-samsung: Let's let the flow follow the same general tradition currently.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: In my opinion, dsector represents current direntry, last_dsector is used for sector which is chain to current sector(newly added).
So use dsector and set last_dsector if it couldn't find entry is correct.
Also please note that ret is always ENOENT in line 1245, if is not necessary.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1339,43 +1343,40 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
...
 		/* Chain the next sector into this sector sector */
-		nextsector = psector;
+		nextsector = new_entry.dsector;
 
-		smartfs_setbuffer(&readwrite, parentdirsector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)&nextsector);
+		smartfs_setbuffer(&readwrite, new_entry.last_dsector, offsetof(struct smartfs_chain_header_s, nextsector), sizeof(uint16_t), (uint8_t *)&nextsector);
 		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 		if (ret < 0) {
-			fdbg("Error chaining new sector %d\n", psector);
+			fdbg("Error chaining new sector %d\n", new_entry.dsector);
```
Taejun-Kwon, line 1365: new_entry.last_dsector

os/fs/smartfs/smartfs_smart.c
```diff
@@ -205,6 +203,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	/* Locate the directory entry for this path */
 
 	sf = (struct smartfs_ofile_s *)kmm_malloc(sizeof * sf);
+	memset(sf, 0, sizeof(struct smartfs_ofile_s));
```
Taejun-Kwon, line 206: it's not necessary, I think

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,43 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
+			   This sector will then be chained in the end.
+			 */
+			sf->bflags |= SMARTFS_BFLAG_NEW_ENTRY;
+			/* Set sf->entry flags (to be writtne later) and store them */
+
+			sf->entry.flags = SMARTFS_ERASEDSTATE_16BIT;
+			smartfs_set_entry_flags(&sf->entry, SMARTFS_DIRENT_TYPE_FILE, mode);
 
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			if (sf->entry.name == NULL) {
```
Taejun-Kwon, line 311: this must be null so we can remove if here.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1996,20 +2019,20 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		}
 
 		/* Okay, we will create newfilename in the newentry directory */
-		newparentdirsector = newentry.firstsector;
+		newentry.dsector = newentry.firstsector;
```
Taejun-Kwon, line 2022: this is incorrect, I think..

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1211,7 +1230,7 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
 				direntry->dsector = psector;
 				direntry->doffset = offset;
 				direntry->flags = entry->flags;
-				*parentdirsector = direntry->dsector;
+				direntry->last_dsector = direntry->dsector;
```
Taejun-Kwon, line 1233: we don't have to update dsector & last_dsector here, it must be out side of while.
assume that sector is chained #3 -> #12  and they are filled with entries(total 24 if sector size is 512)
if sector 13 is empty, anyway it will find from sector 3
At the start of loop, both of dsector & last_dsector are probably 3.
After first loop,  dsector become 12, but last_dsector is still 3..
and second loop, it will break loop. dsector become maybe 0xffff
and dsector changed to new alloc sector by smartfs_createdirentry, it may alloc sector 13.
So it will chain sector 3 to 13 but sector 3 has next sector(12) already,
so some of bits will be corrupted at that time.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1223,10 +1242,11 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
 		psector = nextsector;
 	}
 
-	*parentdirsector = direntry->dsector;
+	/* Retain last active sector in current parent directory to comapre while writing new entry */
+	direntry->last_dsector = direntry->dsector;
 	if (ret != OK) {
 		/* No available entry was found, so we create one */
-		ret = smartfs_createentry(fs, *parentdirsector, direntry);
+		ret = smartfs_createdirentry(fs, direntry);
```
Taejun-Kwon, line 1249: actually line 1247, ret is always -ENOENT, we can remove 'if'

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Actually I added this to resolve a bug.

sf->entry.datlen was not cleared and has old values.
So in smartfs_write, it would cause malfunctioning.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: This part is pending for change
Allocation of entry name space.
Will do in upcoming PR

os/fs/smartfs/smartfs_smart.c
amogh-samsung: 1. finddirentry returns newentry as the entry of the parent directory.
2. So newentry.firstsector is actually the 1st data sector of new parent directory.
3. Now, all sectors in this directory are scanned to find if the entry already exists.
4. If does not exist,
     dsector = firstsector and then find available entry in one of the directory sectors to write new entry
5. dsector and last_dsector get updated ahead. firstsector too.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Will change, thanks.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: > dsector become maybe 0xffff

This cannot happen because
`direntry->dsector = psector;
		psector = nextsector;`
So, psector can become 0xffff, but dsector can not

So at the end of checking parent chain, dsector = 12, psector = 0xffff, last_dsector = 3

Now, Before going for createentry
`direntry->last_dsector = direntry->dsector;`
So, last_dsector = 12 now
Dsector allocated 13 by createentry
13 will be chained to 12 during sync.
And not 
> So it will chain sector 3 to 13

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Actually it is not always -ENOENT beacuse of line
`ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);`
Which may change the value of ret

But yes, if we had found an entry, function would already have returned,
So here we can avoid the check.

os/fs/smartfs/smartfs.h
```diff
@@ -254,10 +256,16 @@
...
 struct smartfs_entry_s {
-	uint16_t firstsector;		/* Sector number of the name */
-	uint16_t dsector;			/* Sector number of the directory entry */
+	uint16_t firstsector;			/* First data sector number of the file */
+	uint16_t dsector;			/* Sector number of the directory entry
+						   We will always scan starting from this sector to find space for the new entry.
+						   After finddirentry(), this holds the last directory in the path for the new entry, i.e. immediate parent.
+						   After find_availableentry(), this holds the new entry sector or the last active sector
+						   to which new sector will be chained after writing entry.
+						 */
```
kishore-sn, line 265: For multiline comments, we generally put a star (*) at the start of each line.

os/fs/smartfs/smartfs_smart.c
kishore-sn, line 206: How about using zalloc instead?

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,43 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
 				goto errout_with_buffer;
 			}
 
 			/* At this point, either an available entry was found or a new one has been created */
-			newentry.firstsector = data_sector;
+			sf->entry.firstsector = data_sector;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			/* If CRC is enabled, hold the entry for writing later */
+			/* Mark flags to indicate that the entry will be written to a new sector in the parent directory.
+			   This sector will then be chained in the end.
```
kishore-sn, line 303: Multiline comments begin with *

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,43 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
 
-			ret = smartfs_writeentry(fs, newentry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry, parentdirsector);
+			if (sf->entry.name == NULL) {
+				sf->entry.name = (FAR char *)kmm_malloc(fs->fs_llformat.namesize + 1);
+				if (sf->entry.name == NULL) {
+					ret = -ENOMEM;
+					goto errout_with_buffer;
+				}
+			}
+			memset(sf->entry.name, 0, fs->fs_llformat.namesize + 1);
```
kishore-sn, line 318: Please use zalloc in all places instead of memset zero.

os/fs/smartfs/smartfs.h
amogh-samsung: Will make the change, thanks.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: I'll check this, thankyou

os/fs/smartfs/smartfs_smart.c
```diff
@@ -289,21 +288,43 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
...
 			/* At this point, either an available entry was found or a new one has been created */
-			newentry.firstsector = data_sector;
+			sf->entry.firstsector = data_sector;
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			/* If CRC is enabled, hold the entry for writing later */
+			/* Mark flags to indicate that the entry will be written to a new sector in the parent directory.
+			   This sector will then be chained in the end.
+			 */
+			sf->bflags |= SMARTFS_BFLAG_NEW_ENTRY;
+			/* Set sf->entry flags (to be writtne later) and store them */
```
amandeepc, line 306: **written

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Currently in line with all other entry->name space allocations.
All of which are to be changed an upcoming next commit.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: New file must be not exist, if there is new file exist, it should return -EEXIST. but now it overwrite.
Current this will move oldpath to subdirectory of newpath, because dsector will be changed to firstsector.
For example,
```
mkdir("/mnt/old", 0777);
mkdir("/mnt/old/a", 0777);
mkdir("/mnt/old/b", 0777);
mkdir("/mnt/new", 0777);
mkdir("/mnt/new/c", 0777);
mkdir("/mnt/new/d", 0777);
rename("/mnt/old", "/mnt/new");
```
Result of this test will shows /mnt/new has 3 directories,  'a', 'b', and 'old'

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: Got it.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -645,7 +662,16 @@ static int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_of
 		}
 
 		sf->byteswritten = 0;
-		sf->bflags = 0;
+		/* File's data sector has been synced with MTD, now check if this is a new file entry and write the entry */
+		if (sf->bflags & SMARTFS_BFLAG_NEW_ENTRY) {
+			/* Flags for this entry have already been set and stored, so mode will not be used */
+			ret = smartfs_writeentry(fs, sf->entry, sf->entry.name, SMARTFS_DIRENT_TYPE_FILE, 0xFFFF, &sf->entry);
```
Taejun-Kwon, line 668: Actually sf->entry.flags keeps mode too.
And smartfs_writeentry will reset flag & set mode again with this 0xFFFF...

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1731,11 +1755,9 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 {
 	struct smartfs_mountpt_s *fs;
 	int ret;
-	struct smartfs_entry_s entry; //Details of newly written entry are returned with this variable
-	uint16_t parentdirsector;
+	struct smartfs_entry_s entry;
```
Taejun-Kwon, line 1758: Below is my style, (grouping)
```
int ret;
int i;
struct smart_entry_s entry;
struct smart_entry_s newentry;

File: os/fs/smartfs/smartfs_utils.c [hunk 46]
```diff
@@ -849,7 +858,11 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 			seglen++;
 			ptr++;
 		}
-
+		/* Perform a check to avoid bufer overflow */
+		if (seglen >= 256) {
```
Taejun-Kwon, line 862: actually workerbuffer is allocated at mount, it also uses constant value. i think we can make definition in smartfs.h
#define SMARTFS_WORKERBUFFER_LEN or something.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1210,8 +1214,12 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
 			if (IS_AVAILABLE_ENTRY(entry)) {
 				direntry->dsector = psector;
 				direntry->doffset = offset;
+#ifdef CONFIG_SMARTFS_ALIGNED_ACCESS
```
Taejun-Kwon, line 1217: Actually I added this when I send mail to you but...we will reset flag, so this is not necessary right?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1223,13 +1231,11 @@ int smartfs_find_availableentry(struct smartfs_mountpt_s *fs, uint16_t *parentdi
...
+	direntry->prev_parent = direntry->dsector;
+	/* No available entry was found, so we create one */
+	ret = smartfs_createdirentry(fs, direntry);
 	if (ret != OK) {
-		/* No available entry was found, so we create one */
-		ret = smartfs_createentry(fs, *parentdirsector, direntry);
-		if (ret != OK) {
-			fdbg("Createentry unable to create space for writing\n");
-		}
+		fdbg("Createentry unable to create space for writing\n");
```
Taejun-Kwon, line 1238: it is very minor one, let's remove 'if' in this commit.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkkay yes I missed this.
The check to see if flags are already valid was removed in your mail.
So some modifications required. Ill do the same.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkay. Minor change, can do.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes, always better to use macro.
Will update when we implement solution to stat fail issue.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: You mean Line 1237 to check if createdirentry returns OK or not?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Yes, shouldn't be necessary.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkkay, now I got your concern regarding rename.
Actually I made an assumption earlier, I thought rename is only intended for files and it was made that way only.
From the 1st time I looked at smartfs_rename() function, I knew this would only work for renaming files.

This function WILL NOT work for renaming directories. I think it was not written that way from the very beginning.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1333,47 +1316,47 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 	entry->utc = time(NULL);
 #endif
 #endif
+	smartfs_set_entry_flags(&new_entry, mode, type);
```
Taejun-Kwon, line 1319: if this is processed before writeentry, then we can prevent duplicated set.
Also mode & type can be removed from parameter.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1750,7 +1769,8 @@ static int smartfs_mkdir(struct inode *mountpt, const char *relpath, mode_t mode
 	/* Locate the directory entry for this path */
 
 	entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &entry, relpath, &parentdirsector, &filename);
+	newentry.name = NULL;
```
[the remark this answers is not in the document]
amogh-samsung: Ohkkay,
Pending for discussion

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Updated

os/fs/smartfs/smartfs.h
```diff
@@ -175,6 +175,10 @@
 #define INODE_STATE_FILE          (CONFIG_NXFFS_ERASEDSTATE ^ 0x22)
 #define INODE_STATE_DELETED       (CONFIG_NXFFS_ERASEDSTATE ^ 0xaa)
 
+/* Smartfs worbuffer maxuimum length */
```
jsdosa, line 178: could you check spelling?
worbuffer -> workbuffer
maxuimum -> maximum

Taejun-Kwon (approved):
Please remember your todo list,
Especially, rename & direntry in writeentry..

os/fs/smartfs/smartfs.h
amogh-samsung: Noted, thankyou.
Pending though for upcoming PRs.

PR #4426, opened 2020-08-10
apps/system: Fix vi exit logic to return from exit()

gcjjyy:
In the tash system, exit() function halts all system. So, didn't back to the TASH.
Changed the exit logic to the method of return instead of calling exit()


apps/system/vi/Makefile
```diff
@@ -64,7 +64,7 @@ APPNAME = vi
 FUNCNAME = vi_main
 PRIORITY = SCHED_PRIORITY_DEFAULT
 STACKSIZE = 16384
```
sunghan-chang, line 66: Because you change the type to sync, vi application will be executed on TASH task. TASH has 4KB stack but because it uses stack internally, vi could use few stack size. We need to check TASH stack size.

jeongchanKim: How about reverting the THREADEXEC type as async and setting the priority as higher than TASH? In this case, we can vi with its own stack.

gcjjyy: I'll try to test both.

gcjjyy: @sunghan-chang @jeongchanKim
Tested with ASYNC mode and higher priority. But the input is ignored by every 2 steps.
So I removed the vi's own stack size to be part of the Tash task.

PR #4427, opened 2020-08-10
apps/shell: Add clear command on TASH

gcjjyy:
there was no clear command. Added.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -94,6 +94,7 @@ struct tash_cmd_info_s {
  ********************************************************************************/
 
 static int tash_help(int argc, char **args);
+static int tash_clear(int argc, char **args);
 static int tash_exit(int argc, char **args);
 #if defined(CONFIG_BOARDCTL_RESET)
 static int tash_reboot(int argc, char **argv);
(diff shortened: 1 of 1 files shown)


apps/shell/tash_command.c
```diff
@@ -191,6 +193,15 @@ static int tash_help(int argc, char **args)
 	return 0;
 }
 
+/** @brief Clear the terminal screen and move cursor to top
+ *  @ingroup tash
+ */
+static int tash_clear(int argc, char **args)
+{
+       printf("\x1b[1;1H\x1b[2J\n");
+       return 0;
```
sunghan-chang, line 202: Could you check violations of coding rules?
Please use a tab instead of spaces.

gcjjyy: Thank you. Fixed.

apps/shell/tash_command.c
```diff
@@ -191,6 +193,15 @@ static int tash_help(int argc, char **args)
 	return 0;
 }
 
+/** @brief Clear the terminal screen and move cursor to top
+ *  @ingroup tash
+ */
+static int tash_clear(int argc, char **args)
+{
+	printf("\x1b[1;1H\x1b[2J\n");
```
Taejun-Kwon, line 201: As I Remember, this is kind of ANSI escape code, right?
If you have some time later, define these of values to somewhere(header file will be enough).

PR #4431, opened 2020-08-11
kernel/semaphore : Save semaphore holders when binary manager is enabled

jeongarmy:
Semaphore holders should be saved to be released when binary recovery or binary update.

### DIFF
--- a/os/include/semaphore.h
+++ b/os/include/semaphore.h
@@ -73,8 +73,8 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-/* Save semaphore holder data when priority inheritance or binary manager recovery is enabled. */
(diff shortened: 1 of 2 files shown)


sunghan-chang:
@jeongarmy Please re-write the title and description. It looks it's new feature. But actually it isn't. This PR changes the conditional.

os/include/semaphore.h
```diff
@@ -119,7 +119,7 @@ struct semholder_s {
  * @brief Structure of generic semaphore
  */
 struct sem_s {
-#ifdef CONFIG_BINMGR_RECOVERY
+#if CONFIG_BINARY_MANAGER
```
sunghan-chang, line 122: ifdef
Please check below as well

PR #4436, opened 2020-08-12
fs/smartfs : Fix to avoid set next sector as empty

Taejun-Kwon:
If file modified but appending is needed due to remained buffer,
smartfs will change current sector to next sector of current sector
But if current sector was last sector, next sector in header is empty, so chain breaks.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -778,10 +778,22 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 		/* Now perform the write. */
 
 		if (readwrite.count > 0) {
-			ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
-			if (ret < 0) {
-				fdbg("Error %d writing sector %d data\n", ret, sf->currsector);
-				goto errout_with_semaphore;
+
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+			/* If sector buffer enabled and it is not last sector, then overwrite it.
+			 * Otherwise, it will be written by sync_internal
+			 */
+			if (readwrite.count == sf->entry.datlen - sf->filepos) {
+				memcpy(&sf->buffer[sf->curroffset], &buffer[byteswritten], readwrite.count);
+				sf->bflags |= SMARTFS_BFLAG_DIRTY;
+			} else
+#endif				
+			{
+				ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
+				if (ret < 0) {
+					fdbg("Error %d writing sector %d data\n", ret, fs->fs_llformat.availbytes - sf->curroffset);
+					goto errout_with_semaphore;
+				}
 			}
 
 			/* Update our control variables */
@@ -807,10 +819,16 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 				goto errout_with_semaphore;
 			}
 
-			/* Now get the chained sector info and reset the offset */
-
-			sf->curroffset = sizeof(struct smartfs_chain_header_s);
-			sf->currsector = SMARTFS_NEXTSECTOR(header);
+			/* If file is modified and more data remains to be appended to file, but no next sector is available,
+			 * do not update sf->currsector and curroffset,
+			 * Will be handled when buffer data is synced.
+			 */
+			if (SMARTFS_NEXTSECTOR(header) != 0xFFFF) {
+				sf->currsector = SMARTFS_NEXTSECTOR(header);
+				
+				/* Now get the chained sector info and reset the offset */
+				sf->curroffset = sizeof(struct smartfs_chain_header_s);
+			}
 		}
 	}
 


os/fs/smartfs/smartfs_smart.c
```diff
@@ -810,7 +810,14 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 			/* Now get the chained sector info and reset the offset */
 
 			sf->curroffset = sizeof(struct smartfs_chain_header_s);
-			sf->currsector = SMARTFS_NEXTSECTOR(header);
+
+			/* If file modified in this loop but buf if buf data is remain, it will try to append data.
```
thapav, line 814: Do you mean >
 If file modified in this loop but if buf is not empty , it will try to append data ?

Taejun-Kwon: yes it will get nextsector and go to outside of loop.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -783,7 +783,11 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 				fdbg("Error %d writing sector %d data\n", ret, sf->currsector);
 				goto errout_with_semaphore;
 			}
-
```
amogh-samsung, line 786: Line 772 `if (readwrite.count > sf->entry.datlen - sf->filepos) {`

If Line 763 sets readwrite.offset as sf->curroffset, why does 772 check for `readwrite.count > sf->entry.datlen - sf->filepos` ?

If offset to start writing is `sf->curroffset`,
then `readwrite.count = sf->entry.datlen - sf->filepos might` not match the space remaining in the sector for writing.

I think code needs to be aligned in these two places.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: I understand what you are trying to convey with the comment.

Could you please change the comment as follows: -
> If file is modified and more data remains to be appended to file, but no next sector is available,
do not update sf->currsector and curroffset,
Will be handled when buffer data is synced.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -807,10 +811,16 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
+			/* If file modified in this loop but buf if buf data is remain, it will try to append data.
+			 * In this case, next sector of last sector is empty. So do not set header data Here, 
+			 * It will handled when it sync
+			 */
+			if (SMARTFS_NEXTSECTOR(header) != 0xFFFF) {
+				sf->currsector = SMARTFS_NEXTSECTOR(header);
+				
+				/* Now get the chained sector info and reset the offset */
+				sf->curroffset = sizeof(struct smartfs_chain_header_s);
+			}
```
amogh-samsung, line 823: Do we need a `break;` statement here?
In case `SMARTFS_NEXTSECTOR(header) != 0xFFFF` or otherwise?

os/fs/smartfs/smartfs_smart.c
```diff
@@ -783,7 +783,11 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 				fdbg("Error %d writing sector %d data\n", ret, sf->currsector);
 				goto errout_with_semaphore;
 			}
-
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
```
amogh-samsung, line 786: Adding this code after BIOC_WRITESECT has been called means
bytes are copied from buffer[byteswritten] to sf->buffer for synchronizing later.

But, with WRITESECT, bytes from position buffer[byteswritten] have already been written to MTD.

So either they should be written or else copied for writing later if CRC is enabled, not both.

a. > #ifndef CONFIG_SMARTFS_USE_SECTOR_BUFFER
    > BIOC_WRITESECT (Line 781 - 785)
    > #else 
    > memcpy to sf->buffer

OR

b. > BIOC_WRITESECT
    > byteswritten += readwrite.count
    > memcpy to sf->buffer (copy next set of bytes)

Should be done in one of the above ways.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: in, line 763 we set write count as max size refer to fs->fs_llformat.availbytes.
But st->entry.datlen represents current file's size, so if readwrite.count + sf->filepos is greater than sf->entry.datlen, then it will overflow end of file.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: You had pass my quiz, I wanted answer 'a' =D
Yes. current code will bring 2times of write.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: oh. ok thanks!

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: If nextsector is 0xffff it means that it reaches lastsector.
Because loop checks sf->filepos < sf->entry.datlen, datlen calculated refer to used of chainheader of each sector when file opened.
So nextsector is 0xffff means that filepos reaches lastsector, it will break by condition of loop =)

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: @amogh-samsung  I thought same as you but I faced another situation, will be modifed more.

Taejun-Kwon: updated.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Yes exactly,
datlen-filepos can be much greater than availbytes-curroffset, so mismatch

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Ohkkay now both copy to sf buffer and write to MTD will not happen as I had mentioned.

But why for the last sector, smartfs_sync_internal is chosen for writing?

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: yes so it goes to next loop for appending, and count of readwrite become to 0.
but even if we change logic in next loop, I think performance is not changed significantly, so I will keep it.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: yes because it has nextsector, working through the loop one more time.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Understood, handled in next loop

PR #4440, opened 2020-08-13
os/fs/smartfs : Fix truncate issues

Taejun-Kwon:
1. If user open with flag O_TRUNCATE but do not close, then it never update flash
   if sector buffer is enabled.
2. If O_TRUNCATE is set when open, do not read whole of sector because it already
   read by smartfs_truncatefile

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -331,11 +331,14 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		 */
 
 		if (sf->currsector != SMARTFS_ERASEDSTATE_16BIT) {
-			smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
-			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long) &readwrite);
-			if (ret < 0) {
-				fdbg("ERROR: Error %d reading sector %d header\n", ret, sf->currsector);
-				goto errout_with_buffer;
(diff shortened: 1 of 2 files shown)


os/fs/smartfs/smartfs_smart.c
```diff
@@ -331,11 +331,14 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 		 */
 
 		if (sf->currsector != SMARTFS_ERASEDSTATE_16BIT) {
-			smartfs_setbuffer(&readwrite, sf->currsector, 0, fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
-			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long) &readwrite);
-			if (ret < 0) {
-				fdbg("ERROR: Error %d reading sector %d header\n", ret, sf->currsector);
-				goto errout_with_buffer;
+			/* For truncate, we already read header and set data in sf->buffer */
+			if (((oflags & (O_CREAT | O_TRUNC)) == 0) || ((oflags & O_APPEND) == 0)) {
```
amogh-samsung, line 335: Truncate file called when
File open mode is (O_CREAT | O_TRUNC) BUT NOT (O_APPEND)
if otherwise, chain header will not be read.

So shouldn't this condition be
`if (((oflags & (O_CREAT | O_TRUNC)) == 0) || ((oflags & O_APPEND) != 0)) {`

Taejun-Kwon: ah yes. you're right. thanks.

amogh-samsung:
Seems all good now

jeongchanKim (approved):
LGTM

PR #4441, opened 2020-08-14
apps/system/utils/utils_proc.h : Remove unnecessary #ifdef condition for proc utils

jeongchanKim:
We don't need to enclose with ENABLE_PROC_UTILS for using proc utils.
Even it is not good to add every configs.
So remove unnecessary condition.

### DIFF
--- a/apps/system/utils/utils_proc.h
+++ b/apps/system/utils/utils_proc.h
@@ -22,11 +22,6 @@
 #include <tinyara/config.h>
 #include <dirent.h>
 
-#if defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_PS) || defined(CONFIG_ENABLE_STACKMONITOR) || defined(CONFIG_ENABLE_HEAPINFO)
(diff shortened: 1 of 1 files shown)


apps/system/utils/utils_proc.h
```diff
@@ -22,7 +22,7 @@
 #include <tinyara/config.h>
 #include <dirent.h>
 
-#if defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_PS) || defined(CONFIG_ENABLE_STACKMONITOR) || defined(CONFIG_ENABLE_HEAPINFO)
+#if defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_IRQINFO) || defined(CONFIG_ENABLE_PS) || defined(CONFIG_ENABLE_STACKMONITOR) || defined(CONFIG_ENABLE_HEAPINFO) || defined(CONFIG_ENABLE_CPULOAD)
```
sunghan-chang, line 25: I don't know why this is necessary. Could you check the operation without this conditional?

PR #4442, opened 2020-08-18
board: Support partition configuration of external flash

jeongarmy:
User can configure partitions of external flash.

### DIFF
--- a/os/board/common/Kconfig
+++ b/os/board/common/Kconfig
@@ -54,3 +54,54 @@ config FLASH_PART_NAME
 
 endif # FLASH_PARTITION
 endif # ARCH_USE_FLASH && (NFILE_DESCRIPTORS != 0)
+
+config ARCH_USE_SECOND_FLASH
+	bool
+	default n
+	---help---
+		This enables Second FLASH to use.
+		It is selected by chip config.
+
+if ARCH_USE_SECOND_FLASH && (NFILE_DESCRIPTORS != 0)
+comment "Second Flash Partition Options"
+
+config SECOND_FLASH_PARTITION
+	bool "Enable partition support on Second FLASH"
+	default n
+	select MTD
+	select MTD_PARTITION
+	---help---
+		Enables creation of partitions on the Second FLASH
+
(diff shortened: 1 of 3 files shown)


os/board/common/Kconfig
```diff
@@ -54,3 +54,53 @@ config FLASH_PART_NAME
...
+
+config EXTERNAL_FLASH_PART_SIZE
+	string "Flash partition size list (in KBytes)"
+	default "512,2048,"
+	---help---
+		Comma separated list of partition sizes in KB.
+
+config EXTERNAL_FLASH_PART_TYPE
+	string "Flash partition type list"
+	default "kernel,smartfs," if !APP_BINARY_SEPARATION
```
jeongchanKim, line 92: I think kernel and smartfs are for APP_BINARY_SEPARATION. You'd better change with below.

os/board/common/Kconfig
```diff
@@ -54,3 +54,53 @@ config FLASH_PART_NAME
 
 endif # FLASH_PARTITION
 endif # ARCH_USE_FLASH && (NFILE_DESCRIPTORS != 0)
+
+config ARCH_USE_EXTERNAL_FLASH
+	bool
+	default n
+	---help---
+		This shows that user enables External FLASH to use.
```
jeongchanKim, line 62: How can we enable this config? Could you add description?

sunghan-chang: Because you remove the prompt, user can't control this config.
This config should be selected by chip config. That's why I request removing the prompt.

os/board/common/Kconfig
```diff
@@ -54,3 +54,53 @@ config FLASH_PART_NAME
...
+
+if ARCH_USE_EXTERNAL_FLASH && (NFILE_DESCRIPTORS != 0)
+comment "External Flash Partition Options"
+
+config EXTERNAL_FLASH_PARTITION
```
sunghan-chang, line 67: How about using `2nd` instead of `external`? Because almost flash is an external module.

PR #4444, opened 2020-08-19
fs/smartfs : Fix sector recovery logic

Taejun-Kwon:
If error happened during scan entries, it free all remained sector.
Modify code to avoid this behavior

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -2029,9 +2029,10 @@ int smartfs_scan_entry(struct smartfs_mountpt_s *fs, char *map, struct sector_re
 			/* Read all of entries in current sector */
 			smartfs_setbuffer(&readwrite, logsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+			/* If current sector is not valid, move to next sector */
 			if (ret < 0) {
 				fdbg("Error %d reading sector %d data\n", ret, logsector);
-				goto errout;
+				break;
 			}
 			
 			/* Check it is already marked */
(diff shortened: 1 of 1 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -2029,9 +2029,10 @@ int smartfs_scan_entry(struct smartfs_mountpt_s *fs, char *map, struct sector_re
 			/* Read all of entries in current sector */
 			smartfs_setbuffer(&readwrite, logsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
 			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+			/* If current sector is not valid, move to next sector */
 			if (ret < 0) {
 				fdbg("Error %d reading sector %d data\n", ret, logsector);
-				goto errout;
+				break;
```
amogh-samsung, line 2035: Line 2161: - `ret = OK;`
won't this line reset ret = OK before the function exits?

Taejun-Kwon: yes, it will pop next node and will test this again in main loop, while (node != NILL)

amogh-samsung: I mean earlier it could return with ret<0 by going to label "errout"
But now ret will always be assigned = OK before it reaches label errout?

Taejun-Kwon: For stable recovery, If read is failed, it will continue to recovery because read doesn't change contents on flash.
But another ioctl include BIOC_WRITESECT, it can affect to contents on flash, so it will go to errout.

amogh-samsung: Ohkkay.
Then it seems alright to continue with recovery instead of exiting in case of read failure.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -2217,6 +2218,10 @@ int smartfs_sector_recovery(struct smartfs_mountpt_s *fs)
```
davidfather, line 2218: At line number 156, 157, 156, 2187, the division operation (i.e.,  X / 8) should be replaced with ( X >> 3) for computation efficiency.

Taejun-Kwon: Will do it in another commit

Taejun-Kwon:
> At line number 2211, what happens if (sector != SMARTFS_ROOT_DIR_SECTOR)?
> Is it an error case or not?

It is same as Like a linux's FIBMAP, it will check current logical sector physically, so it means that we do not use that logical sector.
So it will not be added to active logical map.
But SMARTFS_ROOT_DIR_SECTOR is root sector(/mnt), it should be exist all the time, if it is not there, it is broken file system...

PR #4445, opened 2020-08-19
fs/smartfs : Update journal recovery logic

Taejun-Kwon:
Journal follows checkin -> data write -> header write -> checkout.
If power cut happened after data write, it should checkout process but it didn't
Also if transaction failed in the middle of journal process, release working sector & checkout current journal

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -465,7 +465,7 @@ static crc_t smart_calc_sector_crc(FAR struct smart_struct_s *dev);
 static int smart_journal_init(FAR struct smart_struct_s *dev);
 static int smart_journal_checkin(FAR struct smart_struct_s *dev, journal_log_t *log, uint32_t address, uint16_t psector, int type);
 static int smart_journal_checkout(FAR struct smart_struct_s *dev, journal_log_t *log, uint32_t address);
-static int smart_journal_checkout_process(FAR struct smart_struct_s *dev, journal_log_t *log, uint32_t address);
+static int smart_journal_process_transaction(FAR struct smart_struct_s *dev, journal_log_t *log);
 static int smart_journal_bwrite(FAR struct smart_struct_s *dev, uint16_t physsector);
 static ssize_t smart_journal_bytewrite(FAR struct smart_struct_s *dev, size_t offset, int nbytes, FAR const uint8_t *buffer);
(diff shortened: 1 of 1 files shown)


amogh-samsung:
LGTM

jsdosa (approved):
Looks nice to me.

amogh-samsung:
Changed the function flow for checkout during recovery with new functions?

Taejun-Kwon:
> Changed the function flow for checkout during recovery with new functions?

I just implemented TODO list regarding recovery, and it didn't release & checkout if it failed during any of transaction processing.

os/fs/driver/mtd/smart.c
```diff
@@ -4943,22 +4973,29 @@ static int smart_journal_checkin(FAR struct smart_struct_s *dev, journal_log_t *
...
-		return -EIO;
+		ret = -EIO;
+		goto errout;
 	}
 
 	if (smart_validate_journal_crc(&buf) != OK) {
 		fdbg("Checkin Error, verify crc failed buf crc16 : %d calc crc : %d\n", UINT8TOUINT16(buf.crc16), smart_calc_journal_crc(&buf));
-		return -EIO;
+		ret = -EIO;
+		goto errout;
```
jeongchanKim, line 4991: When going to errout, `ret` will be overwritten by `smart_journal_checkout()`. So I think we don't need to assing `ret` as `-EIO`. How about setting that to errno?

Taejun-Kwon: you're right, it will overwrite, I will fix it.
Anyway set_errno is not good choice, file system always set_errno in vfs level, not sub file system. Maybe kernel also same.

os/fs/driver/mtd/smart.c
```diff
@@ -5068,51 +5117,55 @@ static int smart_journal_checkout_process(FAR struct smart_struct_s *dev, journa
...
 	if (ret != OK) {
-		/* If checkout failed, we release data & change state journal to checkout to remove all data */
-
-		/* TODO Should we make journal for journal to handle This?? */
-		goto error_with_commit;
+		result = ret;
 	}
-
+	
+	if (result != OK) {
```
davidfather, line 5142: This if statement can be merged with the if statement right above as follows:
if (ret != OK) {
  result = ret;
  psector = ...
  ret = smart_journal_release_sector();
  ....
}

Taejun-Kwon: Yes right. I fixed it as you suggested, thanks.

PR #4447, opened 2020-08-20
os: Fix build warnings

jeongarmy:
Fix build warnings because of wrong type definition or type casting.

### DIFF
--- a/os/binfmt/libelf/libelf_addrenv.c
+++ b/os/binfmt/libelf/libelf_addrenv.c
@@ -105,10 +105,11 @@ static int allocateregions(FAR struct elf_loadinfo_s *loadinfo)
 		}
 	}
 
-	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
-	if (loadinfo->binp->ramstart == NULL) {
+	*allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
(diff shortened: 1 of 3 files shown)


os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -106,7 +106,7 @@ static int allocateregions(FAR struct elf_loadinfo_s *loadinfo)
 	}
 
 	loadinfo->binp->ramstart = *allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
-	if (loadinfo->binp->ramstart == NULL) {
+	if (!loadinfo->binp->ramstart) {
```
jeongchanKim, line 109: `kmm_memalign` return the pointer or NULL, and the meaning of ramstart is a address even its type is uint.
So I think we'd better check with `(uintptr_t)NULL`.

Taejun-Kwon, line 109: we can reduce 'lea' operation in very lower level.
```
*allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
if (*allocs[0] == NULL) {
    return -ENOMEM;
}
loadinfo->binp->ramstart = *allocs[0];
```

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4450, opened 2020-08-24
fs/smartfs : Fix power cut issue

Taejun-Kwon:
If sector data & header written properly but checkout wasan't, checkout only should be done.
But it checked data & journal data again, and do something more, so fix logic to avoid
unnecessary process.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -4915,8 +4915,8 @@ static int smart_journal_release_sector(FAR struct smart_struct_s *dev, uint16_t
 	offset = psector * dev->sectorsize;
 	ret = MTD_READ(dev->mtd, offset, sizeof(struct smart_sect_header_s), (FAR uint8_t *)&header);
 	if (ret != sizeof(struct smart_sect_header_s)) {
-		fdbg("read failed, so we will make header forcely..ret : %d\n", ret);
-		header.status = CONFIG_SMARTFS_ERASEDSTATE;
+		fdbg("Read header failed psector : %d offset : %d\n", psector, offset);
+		return -EIO;
 	}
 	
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5476,9 +5476,9 @@ static int smart_journal_recovery(FAR struct smart_struct_s *dev, journal_log_t
 			fvdbg("Entire of sector is valid!!!!\n");
 			ret = smart_journal_checkout(dev, log, address);
 			if (ret != OK) {
-				/* If checkout failed, keep last journal */
-				return ret;
-			}
+				fdbg("Sector was valid but checkout failed\n");
+			}
```
amandeepc, line 5480: Please remove whitespace after closing bracket

os/fs/driver/mtd/smart.c
```diff
@@ -5476,9 +5476,9 @@ static int smart_journal_recovery(FAR struct smart_struct_s *dev, journal_log_t
 			fvdbg("Entire of sector is valid!!!!\n");
 			ret = smart_journal_checkout(dev, log, address);
 			if (ret != OK) {
-				/* If checkout failed, keep last journal */
-				return ret;
+				fdbg("Sector was valid but checkout failed\n");
```
jsdosa, line 5479: @Taejun-Kwon how about this?
```
if (ret == OK) {
			fvdbg("Entire of sector is valid!!!!\n");
			ret = smart_journal_checkout(dev, log, address);
			if (ret != OK) {
				fdbg("Sector was valid but checkout failed\n");
			}
			return ret;
		}
		
		fvdbg("Entire of sector is invalid\n");
```

Taejun-Kwon: yeah more good. I will modify it. thanks.

sunghan-chang (approved):

PR #4453, opened 2020-08-24
add RTC driver initialization in arch for artik05x, sidk_s5jt200, stm…

sunghan-chang:
…32l4r9ai-disco

RTC driver in board gets lowerhalf structure from arch and initializes
using driver function. But other drivers like timer, gpio and etc do it in arch
and board just calls it.
We don't need to deliver lowerhalf to board so that we had better do it as others.

### DIFF
--- a/os/arch/arm/src/s5j/s5j_rtc_lowerhalf.c
+++ b/os/arch/arm/src/s5j/s5j_rtc_lowerhalf.c
@@ -59,6 +59,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <errno.h>
+#include <debug.h>
 
 #include <tinyara/arch.h>
 #include <tinyara/rtc.h>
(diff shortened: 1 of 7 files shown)


os/arch/arm/src/stm32l4/stm32l4_rtc_lowerhalf.c
```diff
@@ -697,32 +698,28 @@ static int stm32l4_cancelperiodic(FAR struct rtc_lowerhalf_s *lower, int id)
...
 
-FAR struct rtc_lowerhalf_s *stm32l4_rtc_lowerhalf(void)
+void stm32l4_rtc_initialize(void)
 {
-  nxsem_init(&g_rtc_lowerhalf.devsem, 0, 1);
+	struct rtc_lowerhalf_s *lower = (struct rtc_lowerhalf_s *)&g_rtc_lowerhalf;
+	int ret;
 
-  return (FAR struct rtc_lowerhalf_s *)&g_rtc_lowerhalf;
+	ret = rtc_initialize(0, lower);
```
thapav, line 719: May be we can keep the original comment:

/* Bind the lower half driver and register the combined RTC driver	
 * as /dev/rtc0	
 */

sunghan-chang: Added.

os/arch/arm/src/s5j/s5j_rtc_lowerhalf.c
```diff
@@ -290,29 +291,29 @@ static struct s5j_lowerhalf_s g_rtc_lowerhalf = {
...
- *   Instantiate the RTC lower half driver for the S5J.
- *   General usage:
- *
- *     #include <tinyara/rtc.h>
- *     #include "s5j_rtc.h"
- *
- *     struct rtc_lowerhalf_s *lower;
- *     lower = s5j_rtc_lowerhalf();
- *     rtc_initialize(0, lower);
+ *   Register the RTC driver to provide RTC functionality
```
Taejun-Kwon, line 297: In my opinion, lower half driver of RTC is correct, instead of RTC driver, function name too.

os/arch/arm/src/s5j/s5j_rtc_lowerhalf.c
```diff
@@ -290,29 +291,29 @@ static struct s5j_lowerhalf_s g_rtc_lowerhalf = {
...
  *
  * Returned Value:
- *   On success, a non-NULL RTC lower interface is returned. NULL is
- *   returned on any failure.
+ *   None
  *
  ****************************************************************************/
-FAR struct rtc_lowerhalf_s *s5j_rtc_lowerhalf(void)
+
+void s5j_rtc_initialize(void)
```
Taejun-Kwon, line 307: Currently, rtc_lowerhalf_s type is not useful but generally it had better keep lowerhalf type, then it is easy to add multiple driver in future. I don't think multiple RTC is possible for us, but it is better option..

os/board/sidk_s5jt200/src/s5jt200_tash.c
```diff
@@ -404,20 +403,9 @@ int board_app_initialize(void)
...
-		if (rtclower) {
-			ret = rtc_initialize(0, rtclower);
-			if (ret < 0) {
-				lldbg("Failed to register the RTC driver: %d\n",
-						ret);
-			}
-		}
-	}
-#endif /* CONFIG_RTC_DRIVER */
+#ifdef CONFIG_RTC_DRIVER
```
Taejun-Kwon, line 406: yes actually previous is more correct way.

sunghan-chang: Agree. Let me revert.

PR #4454, opened 2020-08-25
os/fs/smartfs: Optimize parameters for finddirentry()

amogh-samsung:
- Can work smartfs_finddirentry() without extra filename parameter.
- fix:smartfs_finddirentry returns filename NULL.
- Allocate entry name space minimum times.
- Use kmm_zalloc instead of kmm_malloc + memset(0).

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -176,7 +176,6 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 	struct inode *inode;
 	struct smartfs_mountpt_s *fs;
 	int ret;
-	const char *filename;
 	struct smartfs_ofile_s *sf;
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 	struct smart_read_write_s readwrite;
@@ -223,7 +222,7 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 #endif							/* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 
 	sf->entry.name = NULL;
-	ret = smartfs_finddirentry(fs, &sf->entry, relpath, &filename);
+	ret = smartfs_finddirentry(fs, &sf->entry, relpath);
 	/* Three possibilities: (1) a node exists for the relpath and
 	 * dirinfo describes the directory entry of the entity, (2) the
 	 * node does not exist, or (3) some error occurred.
@@ -305,19 +304,10 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 #else
 			sf->entry.flags = (uint16_t)(mode & SMARTFS_DIRENT_MODE);
 #endif
-			if (sf->entry.name == NULL) {
-				sf->entry.name = (FAR char *)kmm_malloc(fs->fs_llformat.namesize + 1);
-				if (sf->entry.name == NULL) {
-					ret = -ENOMEM;
-					goto errout_with_buffer;
-				}
-			}
-			memset(sf->entry.name, 0, fs->fs_llformat.namesize + 1);
-			strncpy(sf->entry.name, filename, fs->fs_llformat.namesize);
 #else
 			/* If CRC is disabled, write the new file entry */
 			/* At this point, either an available entry was found or a new one has been created */
-			ret = smartfs_writeentry(fs, sf->entry, filename, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry);
+			ret = smartfs_writeentry(fs, sf->entry, SMARTFS_DIRENT_TYPE_FILE, mode, &sf->entry);
 			if (ret != OK) {
 				fdbg("Write entry failed\n");
 				goto errout_with_buffer;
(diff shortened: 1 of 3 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1095,10 +1090,17 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 
 			if (*ptr == '\0') {
 				direntry->dsector = dirstack[depth];
-				*filename = segment;
+
+				if (direntry->name == NULL) {
+					direntry->name = kmm_zalloc(fs->fs_llformat.namesize + 1);
+					if (!direntry->name) {
+						return -ENOMEM;
```
Taejun-Kwon, line 1097: It is wrong here. return & goto

os/fs/smartfs/smartfs_utils.c
```diff
@@ -966,15 +963,14 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 							direntry->dsector = readwrite.logsector;
 							direntry->doffset = offset;
 							direntry->dfirst = dirstack[depth];
+
 							if (direntry->name == NULL) {
-								direntry->name = (char *)kmm_malloc(fs->fs_llformat.namesize + 1);
+								direntry->name = (char *)kmm_zalloc(fs->fs_llformat.namesize + 1);
```
Taejun-Kwon, line 968: anyway we have to use direntry->name, instead of filename, then we can allocate it at the start of function.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1361,15 +1358,13 @@ int smartfs_writeentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s new_
 #endif
 	direntry->datlen = 0;
 	if (direntry->name == NULL) {
-		direntry->name = (FAR char *)kmm_malloc(fs->fs_llformat.namesize + 1);
+		direntry->name = (FAR char *)kmm_zalloc(fs->fs_llformat.namesize + 1);
```
Taejun-Kwon, line 1361: this direntry is useless as I told you many time,

os/fs/smartfs/smartfs_utils.c
amogh-samsung: direntry->name will need space to be allocated but not always.
Line - 1103, direntry->name is left as NULL

os/fs/smartfs/smartfs_utils.c
amogh-samsung: My bad, will correct.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: You mean there's no need for allocating name space for diretnry here?
Or no need to return updated entry in direntry at all?

Taejun-Kwon: 2nd one.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: I mean normally we should use it, it is rare case, it can be free in errout.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -2047,7 +2019,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
 		if (oldentry.dsector == newentry.dsector) {
 			/* We will not use any new entry found, we will overwrite the existing entry */
-			ret = smartfs_writeentry(fs, oldentry, newfilename, type, mode, &newentry);
+			ret = smartfs_writeentry(fs, oldentry, type, mode);
```
Taejun-Kwon, line 2022: maybe old entry has old path, this will not change anything. How do you think?

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1986,9 +1961,8 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 			goto errout_with_semaphore;
 		}
 
-		/* Nope, it's a directory.  Now search the directory for oldfilename */
+		/* Nope, it's a directory.  Now search the directory for old file name */
 
-		newfilename = oldfilename;
```
Taejun-Kwon, line 1991: oldentry.dsector is correct?
```
fdbg("Error %d reading sector %d data\n", ret, oldentry.dsector);
```

os/fs/smartfs/smartfs_smart.c
amogh-samsung: We are using the old entry only, so path is OK
I should check the name that will be written however.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Hmm, I think we lost consistency here with multiple changes.
I'll take care of this.

Taejun-Kwon (approved):

PR #4455, opened 2020-08-25
fs/smartfs : update wearlevel count after journal recovery if checkin…

Taejun-Kwon:
… journal is ERASE

If wear level is enabled, we should update level count after each of MTD_ERASE.


amogh-samsung (commented):
Looks good.
However, what is the reason for journal_process_transaction in case of JOURNAL_TYPE_RELEASE?

Taejun-Kwon:
> Looks good.
> However, what is the reason for journal_process_transaction in case of JOURNAL_TYPE_RELEASE?

Redo release once again!

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4457, opened 2020-08-26
os/board: tidyup CONFIG_ARCH_USE_FLASH, CONFIG_ARCH_USE_SECOND_FLASH

sunghan-chang:
1. remove the prompt to hide from user
  They show the board has the flash, they are not the way to enable flash.
  CONFIG_FLASH_PARTITION and CONFIG_SECOND_FLASH_PARTITION are for that.
  So, user should not control them.

2. make a dependency with board config, not with chip config
  Ususally flash has a dependency with board, not with chip.
  Let's move the selection them under board config in Kconfig.

3. rename it to CONFIG_ARCH_BOARD_HAVE_FLASH
  By 2nd modification, it had better have BOARD prefix.
  And by 1st, it had better have HAVE prefix instead of USE.

### DIFF
--- a/os/arch/arm/src/s5j/Kconfig
+++ b/os/arch/arm/src/s5j/Kconfig
@@ -351,7 +351,6 @@ config S5J_WDT_DEFTIMEOUT
 config S5J_SFLASH
 	bool "SFLASH"
 	default y
-	select ARCH_USE_FLASH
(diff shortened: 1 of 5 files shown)


jsdosa (approved):
looks nice to me.

Taejun-Kwon (approved):

PR #4462, opened 2020-08-26
libc/misc: add size limitation of sendfile

sunghan-chang:
The sendfile has return value with ssize_t. But the count argument which
gives the size user requests is size_t type. So, the ntransferred variable
which shows transferred size could be overflowed.
Before sending the read data, let's check the transferred size and new
requested sending data size and adjust the size not to overflow.

### DIFF
--- a/lib/libc/misc/lib_sendfile.c
+++ b/lib/libc/misc/lib_sendfile.c
@@ -61,6 +61,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <errno.h>
+#include <limits.h>
 
 #include "lib_internal.h"
 
@@ -136,6 +137,7 @@ ssize_t sendfile(int outfd, int infd, off_t *offset, size_t count)
 	FAR uint8_t *wrbuffer;
 	off_t startpos = 0;
 	ssize_t nbytesread;
+	size_t read_buf_size;
 	ssize_t nbyteswritten;
 	ssize_t ntransferred;
 	bool endxfr;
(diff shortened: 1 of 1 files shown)


lib/libc/misc/lib_sendfile.c
```diff
@@ -216,6 +216,17 @@ ssize_t sendfile(int outfd, int infd, off_t *offset, size_t count)
...
 			wrbuffer = iobuffer;
+
+			/* Check the limitation of ntransferred (SSIZE_MAX) */
+
+			if ((size_t)ntransferred + nbytesread > SSIZE_MAX) {
+
+				/* Re-calculate the size which will be sent not to overflow maximum value of ssize_t */
+
+				nbytesread = SSIZE_MAX - ntransferred;
+				endxfr = true;
```
Taejun-Kwon, line 227: In my opinion, This should not be set here.
```
if ((size_t)ntransferred + nbytesread > SSIZE_MAX) {
    nbytesread = SSIZE_MAX - ntransferred;
}
do {
    ...
    nbyteswritten = write(outfd, wrbuffer, nbytesread);
    if (ntransffered == SSIZE_MAX) {
          break;
    }
}
```

sunghan-chang: `break` is not enough. It will escape the do loop but it will continue the for loop even it sent the maximum size already.

sunghan-chang: And before meet `if (ntransffered == SSIZE_MAX)` statement, `nbytesread` becomes zero and do loop will be end by while conditional.
I mean 
1. The sendfile function already sent the maximum size so that it should be end. But `break` can't stop the loop.
2. `break` statement is duplicated with while conditional and it is unreachable statement.

Taejun-Kwon: I didn't see do-while sorry about it =)
But still it is not correct here, it should restrict size of read first.

sunghan-chang: Fixed as you said.

sangwon03 (approved):
LGTM

PR #4467, opened 2020-08-27
fs/fscmd : Prevent overflow of fscmd_buffer

Taejun-Kwon:
Check buffer size when memcpy to fscmd_buffer


sunghan-chang (approved):

jsdosa (approved):
looks nice to me.

sunghan-chang (commented):
In buffer overflow point of view, I think equal should be added at line 221. Could you check that?

sunghan-chang (approved):

PR #4469, opened 2020-08-27
os/board/rtl8721csm: Add RAM data copy API on app_start

ziyik:
Add RAM data copy API to app_start for code reference.
RAM data copy have been done on boot loader, add it on app_start for code reference.

### DIFF
--- a/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
+++ b/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
@@ -1257,6 +1257,22 @@ static void app_driver_call_os_func_init(void)
 	driver_call_os_func_map.driver_exit_critical = vPortExitCritical;
 }
 
+IMAGE2_RAM_TEXT_SECTION
+void app_RAM_data_copy(void)
+{
+	IMAGE_HEADER *img2_hdr = NULL;
+	IMAGE_HEADER *FlashImage2DataHdr = NULL;
+
+	img2_hdr = (IMAGE_HEADER *)((__flash_text_start__) - IMAGE_HEADER_LEN);
+
+	FlashImage2DataHdr = (IMAGE_HEADER *)(__flash_text_start__ + img2_hdr->image_size);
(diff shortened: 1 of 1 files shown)


os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1257,6 +1257,22 @@ static void app_driver_call_os_func_init(void)
 	driver_call_os_func_map.driver_exit_critical = vPortExitCritical;
 }
 
+IMAGE2_RAM_TEXT_SECTION
+void app_RAM_data_copy(void)
```
sangam-swami, line 1261: Is it verified? The data copy is verified at run time?

ziyik: ya, i have verify it.
It can copy RAM data during app_start. Data in RAM section will be reset.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1266,6 +1282,7 @@ void app_start(void)
 	VectorTableOverride();
 
 	app_section_init();
+	app_RAM_data_copy();
```
sangam-swami, line 1285: I have verified on latest code base, the global data from flash (example : kregionx_start , kregionx_size) is copied to RAM properly. So the current change of app_RAM_data_copy is not required as you mentioned its done in BL.

Can you remove os_heap_init() function call below and its definition, as kregionx_start is set at up_allocate_kheap() in os_start common codes. It is duplicate otherwise to call below.

ziyik: Just to confirm, Realtek os_heap_init() used the free stack info on .ld. So it will allocate all used stack as free heap.
Are up_allocate_kheap() does the same, or it is allocated base on user config file?

ziyik:
This PR will be close, because it is not needed any more.
RAM data copy have been done on boot loader, no need to do it again on app_start.

os_heap_init functionality in Realtek is better and safer, we agree to remain it.

Thanks

PR #4470, opened 2020-08-27
apps/system/utils: tidyup

sunghan-chang:
1. apps/system/utils: fix typo in utils_ps.c 
2. apps/system/utils/fscmd: tidyup tash_cat
    1. remove duplicated fscmd_free for src/dest_fullpath at end of function
    2. In redirection case, written size could be less than requested size.
       add while loop to write entire read data to destination

### DIFF
--- a/apps/system/utils/fscmd.c
+++ b/apps/system/utils/fscmd.c
@@ -273,7 +273,9 @@ static int tash_cat(int argc, char **args)
 	int destfd;
 	int i;
 	int flags;
-	ssize_t ret = 0;
+	ssize_t read_size;
+	ssize_t written_size = 0;
+	int ret;
 
 	direction.mode = FSCMD_NONE;
 	direction.index = -1;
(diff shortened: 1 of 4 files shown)


sunghan-chang:
@sijupunnnoose I pushed 2nd commit for what you suggested. Could you review this?

apps/system/utils/fscmd.c
```diff
@@ -273,7 +273,9 @@ static int tash_cat(int argc, char **args)
 	int destfd;
 	int i;
 	int flags;
-	ssize_t ret = 0;
+	ssize_t read_size;
+	ssize_t written_size = 0;
+	int ret;
```
jeongchanKim, line 278: How about initializing the `ret` as `ERROR` and removing line 317, 356, 363 and 370?

sunghan-chang: Thank you for suggestion. But current way is easier to know the failure and does not get affected from code change.  It will make us check all of code to check whether it is set well.

Taejun-Kwon (approved):
@sunghan-chang please update cat with sendfile, I love you.

PR #4471, opened 2020-08-28
os/fs/smartfs: Support renaming of directories in smartfs

amogh-samsung:
- Current code supports only rename/move operations on files.
- We should support directory renaming also.
- Moving directories will not be supported.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -1917,11 +1917,6 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	struct smartfs_entry_s newentry;
 	mode_t mode;
 	uint16_t type;
-	uint16_t sector;
-	uint16_t offset;
-	uint16_t entrysize;
-	struct smartfs_entry_header_s *entry;
-	struct smart_read_write_s readwrite;
 
 	/* Sanity checks */
 
@@ -1938,75 +1933,21 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	oldentry.name = NULL;
 	newentry.name = NULL;
 	ret = smartfs_finddirentry(fs, &oldentry, oldrelpath);
-	if (ret < 0) {
+	if (ret != OK) {
 		fdbg("Old entry doesn't exist\n");
 		goto errout_with_semaphore;
 	}
 
-	/* Search for the new entry and validate it DOESN'T exist, unless we
-	 * are copying to a directory and keeping the same file name, such as:
-	 *
-	 *    mv /mntpoint/somefile.txt /mntpoint/somedir
-	 *
-	 * in which case, we need to validate the /mntpoint/somedir/somefile.txt
-	 * doens't exsit.
-	 */
-
+	/* Search for the new entry and validate it DOESN'T exist */
 	ret = smartfs_finddirentry(fs, &newentry, newrelpath);
 	if (ret == OK) {
-		/* Test if it's a file.  If it is, then it's an error */
-
-		if ((newentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE) {
-			ret = -EEXIST;
-			goto errout_with_semaphore;
-		}
-
-		/* Nope, it's a directory.  Now search the directory for old file name */
-
-		sector = newentry.firstsector;
-		while (sector != SMARTFS_ERASEDSTATE_16BIT) {
-			/* Read the next sector of diretory entries */
-
-			smartfs_setbuffer(&readwrite, sector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
-			ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-			if (ret < 0) {
-				fdbg("Error %d reading sector %d data\n", ret, sector);
-				goto errout_with_semaphore;
-			}
-
-			offset = sizeof(struct smartfs_chain_header_s);
-			entrysize = sizeof(struct smartfs_entry_header_s) + fs->fs_llformat.namesize;
-			while (offset + entrysize < fs->fs_llformat.availbytes) {
-				/* Test the next entry */
-
-				entry = (struct smartfs_entry_header_s *)&fs->fs_rwbuffer[offset];
-				/* Check if entry flags are erased OR empty OR inactive and skip the entry */
-				if ((entry->flags == SMARTFS_ERASEDSTATE_16BIT) || ((entry->flags & SMARTFS_DIRENT_EMPTY) == (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_EMPTY)) || ((entry->flags & SMARTFS_DIRENT_ACTIVE) != (SMARTFS_ERASEDSTATE_16BIT & SMARTFS_DIRENT_ACTIVE))) {
-					/* This entry isn't valid, skip it */
-
-					offset += entrysize;
-					continue;
-				}
-
-				/* Test if this entry matches the filename */
-
-				if (strcmp(newentry.name, entry->name) == 0) {
-					/* Uh-oh, looks like the new entry name already exists */
-
-					ret = -EEXIST;
-					goto errout_with_semaphore;
-				}
-
-				offset += entrysize;
-			}
-
-			/* Chain to next sector */
-
-			sector = SMARTFS_NEXTSECTOR(((struct smartfs_chain_header_s *)fs->fs_rwbuffer));
-		}
-
-		/* Okay, we will create new file in the newentry directory */
-		newentry.dsector = newentry.firstsector;
+		/* The entry to be created after rename already exists, cannot complete rename request */
+		ret = -EEXIST;
+		goto errout_with_semaphore;
+	}
+	if (ret != -ENOENT) {
+		fdbg("Cannot rename entry, error = %d\n", ret);
+		goto errout_with_semaphore;
 	}
 
 	/* Test if the new parent directory is valid */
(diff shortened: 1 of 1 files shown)


amogh-samsung:
Current state of commits will support: -
1. Renaming of files -> rename("/mnt/tmp/file1", "/mnt/tmp/file2");
2. Moving files to a new directory/path with the same name -> rename("/mnt/tmp/file1", "/mnt/tmp_2");
3. Renaming directories -> rename("/mnt/tmp/dir1", "/mnt/tmp/dir2");

Moving directories will NOT be supported -> rename("/mnt/tmp/dir1", "/mnt/tmp_2");

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1954,15 +1954,22 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 
 	ret = smartfs_finddirentry(fs, &newentry, newrelpath);
 	if (ret == OK) {
-		/* Test if it's a file.  If it is, then it's an error */
-
-		if ((newentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE) {
+		/* If the old entry and the new entry found are of the same type, then rename isn't possible as the entry already exists */
+		if ((oldentry.flags & SMARTFS_DIRENT_TYPE) == (newentry.flags & SMARTFS_DIRENT_TYPE)) {
```
Taejun-Kwon, line 1958: We don't have to check type, if path is same it shoud return error.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1954,15 +1954,22 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+			fdbg("Entry of the same name and type already exists\n");
 			ret = -EEXIST;
 			goto errout_with_semaphore;
 		}
 
-		/* Nope, it's a directory.  Now search the directory for old file name */
+		/* Can't complete the request if we are being asked to move a directory to under a file */
+		if (((oldentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_DIR) && ((newentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE)) {
```
Taejun-Kwon, line 1965: same as above.
so logic in while from line 1974, we can remove it.

Taejun-Kwon (commented):
Also I think line 2027, (oldentry.dsector == newentry.dsector), finddirentry always return first parent directory sector? If itnot, dfirstsector is correct ..

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Will change if we let go of move functionality

amogh-samsung: Will update line 2027 accordingly too

Taejun-Kwon:
> 2. Moving files to a new directory/path with the same name -> rename("/mnt/tmp/file1", "/mnt/tmp_2");
In my opinion, IT's not rename's role esentially.. if /mnt/tmp_2 is there, it should return error.
mv must be handled above layer, for example fscmd can implmenet it with sendfile()

amogh-samsung:
> Moving files to a new directory/path with the same name -> rename("/mnt/tmp/file1", "/mnt/tmp_2");
In my opinion, IT's not rename's role esentially.. if /mnt/tmp_2 is there, it should return error.
mv must be handled above layer, for example fscmd can implmenet it with sendfile()

Yes this is actually a Move file operation,
And this is what is causing the confusion.

Two separate functions for move and rename will work much better.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1938,75 +1933,22 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
-
+	/* Search for the new entry and validate it DOESN'T exist */
 	ret = smartfs_finddirentry(fs, &newentry, newrelpath);
 	if (ret == OK) {
-		/* Test if it's a file.  If it is, then it's an error */
-
-		if ((newentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE) {
-			ret = -EEXIST;
+		ret = -EEXIST;
+		fdbg("Entry of the same name already exists, cannot rename\n");
```
thapav, line 1945: May be we can modify as:
Duplicate entry, cannot rename

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1938,75 +1933,22 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
+		goto errout_with_semaphore;
+	} else {
+		if (ret != -ENOENT) {
+			fdbg("Cannot rename entry\n");
```
thapav, line 1949: Here we could update error number in log

os/fs/smartfs/smartfs_smart.c
amogh-samsung: "Duplicate entry" would perhaps mean two identical are found, which would't technically be true.
That is the situation we are actually avoiding.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Smartfs does not always print the error code on the console, keeping the same for now.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1938,75 +1933,22 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
 	ret = smartfs_finddirentry(fs, &newentry, newrelpath);
 	if (ret == OK) {
-		/* Test if it's a file.  If it is, then it's an error */
-
-		if ((newentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE) {
-			ret = -EEXIST;
+		ret = -EEXIST;
+		fdbg("Entry of the same name already exists, cannot rename\n");
+		goto errout_with_semaphore;
+	} else {
```
Taejun-Kwon, line 1947: else is unnecessary, please see below
```
	if (ret == OK) {
		ret = -EEXIST; // it is enough information that file is already exist.
		goto errout_with_semaphore;
	}
	
	if (ret != -ENOENT) {
		goto errout_with_semaphore; // same, with errno user can realize type of error easily ..
	}
```

os/fs/smartfs/smartfs_smart.c
```diff
@@ -2020,6 +1962,7 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		if (oldentry.dsector == newentry.dsector) {
 			/* We will not use any new entry found, we will overwrite the existing entry but with a new name */
 			strncpy(oldentry.name, newentry.name, fs->fs_llformat.namesize);
+			oldentry.prev_parent = oldentry.dsector;
```
Taejun-Kwon, line 1965: For newentry, finddirentry returns start sector of parent folder's sector number because there's no entry there.
But oldentry, entry is already there, it will return correct dsector.
So to compare parent directory properly, below is correct.
```
 if (oldentry.dfirst == newentry.dsector) 
```
You can test it easily put some logs..

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Mistake in pushing version heads, sorry

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Mistake in pushing version heads, sorry

Taejun-Kwon (approved):

Taejun-Kwon (approved):

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1938,75 +1933,21 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 	oldentry.name = NULL;
 	newentry.name = NULL;
 	ret = smartfs_finddirentry(fs, &oldentry, oldrelpath);
-	if (ret < 0) {
+	if (ret != OK) {
```
sangam-swami, line 1936: can we check if (ret)

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1938,75 +1933,21 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
...
-			sector = SMARTFS_NEXTSECTOR(((struct smartfs_chain_header_s *)fs->fs_rwbuffer));
-		}
-
-		/* Okay, we will create new file in the newentry directory */
-		newentry.dsector = newentry.firstsector;
+		/* The entry to be created after rename already exists, cannot complete rename request */
+		ret = -EEXIST;
+		goto errout_with_semaphore;
+	}
+	if (ret != -ENOENT) {
```
sangam-swami, line 1948: we can replace with else if? it will avoid else condition check in case if is success

amogh-samsung: In case of success (ret == OK), the operations can not be completed, it is checked above.

We do not want it to be successful.
However, failure should also only be due to inability to find the entry.
Hence in all cases other than -ENOENT, we have to exit the operation.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: writing (ret != OK) makes it understandable as checking if the function execution is not OK, then exit the operation.

Done in the same way for all smartfs fs operations.

sangam-swami: It might reduce instructions used

amogh-samsung: @Taejun-Kwon 
This might be a good point? Extra instructions might be required to resolve the value of OK macro ?

However, since this is done in the same way at all instances in smartfs, changes will be done altogether only if required in future commits.

os/fs/smartfs/smartfs_smart.c
sangam-swami: as in 1st if condition you are exiting, so if else don't make any difference, so its fine

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: @sangam-swami  Your suggestion will work fine here, but smartfs uses mtd operation many times, these mtd operation returns written bytes, or read bytes or number of pages in sector...
In this case it will return positive value, so I requested to him to use OK explicitly to get correct result.

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: @sangam-swami Anyway if first is success, it will not check second if because goto is there. 
Please see old comment which I guided to him.
If any of branch always return value, it had better divide code to improve readability.
Of course not for everyone, but at least I feel more easier.

PR #4472, opened 2020-08-28
fs/fscmd : Add mv

Taejun-Kwon:
there's no mv, so we can't test rename in tash, so add it


apps/system/utils/fscmd.c
```diff
@@ -1308,6 +1358,9 @@ const static tash_cmdlist_t fs_utilcmds[] = {
 #ifndef CONFIG_DISABLE_ENVIRON
 	{"rmdir",     tash_rmdir,     TASH_EXECMD_SYNC},
 #endif
+#ifndef CONFIG_DISABLE_ENVIRON
+	{"mv",        tash_mv,        TASH_EXECMD_SYNC},
```
sunghan-chang, line 1362: Could you add this by Arphabet order?

Taejun-Kwon: D.O.N.E

sunghan-chang (approved):

PR #4479, opened 2020-09-01
debug/fs : fix build error when nxfusing run

Taejun-Kwon:
can't find dbg_noarg when nxfusing, because config is different


sangwon03 (approved):
LGTM

PR #4480, opened 2020-09-01
[DO NOT MERGE] os/fs/smartfs: Modify all return value checks at fs layer

amogh-samsung:
- Lower level returns OK on success and error code otherwise.
- Check (ret != OK) instead of (ret < 0) for better understanding.


Taejun-Kwon:
This should not be changed such a simply, before change this, smart_ioctl should be checked very carefully.

Taejun-Kwon:
Need to check more and more, we will open if we need it in future.

PR #4486, opened 2020-09-02
kernel/binary_manager: Get path to download binary from binary_manager_open_new_entry

jeongarmy:
User can get a path download binary instead of opened fd from binary_manager_open_new_entry.
And then user can perform file operations like open, read, write.

### DIFF
--- a/framework/include/binary_manager/binary_manager.h
+++ b/framework/include/binary_manager/binary_manager.h
@@ -111,7 +111,7 @@ int binary_manager_unregister_state_changed_callback(void);
  *  Otherwise, it creates empty file with version and returns a path of the file.
  * @param[in] binary_name The binary name to update
  * @param[in] version The binary version to update
- * @param[out] download_path The path to the download binary. The maximum length is CONFIG_PATH_MAX.
+ * @param[out] download_path The path to the download binary. The maximum length is BINARY_PATH_LEN.
  * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) on success. On failure, negative value is returned.
  * @since TizenRT v3.1 PRE
--- a/framework/include/binary_manager/binary_manager.h
+++ b/framework/include/binary_manager/binary_manager.h
@@ -43,11 +43,11 @@
  *  This function loads a new binary after downloading the binary.\n
  * It requests the binary manager to load the input binary name.
  * @param[in] binary_name The name of a new binary to be loaded
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_update_binary(char *binary_name);
+binmgr_result_type_e binary_manager_update_binary(char *binary_name);
 
 /**
  * @brief Get the binary information with name
@@ -55,53 +55,53 @@ int binary_manager_update_binary(char *binary_name);
  *  It requests the binary manager to get the binary information through input binary name.
  * @param[in] binary_name The binary name which to get
  * @param[out] binary_info The address value to receive the binary information
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_get_update_info(char *binary_name, binary_update_info_t *binary_info);
+binmgr_result_type_e binary_manager_get_update_info(char *binary_name, binary_update_info_t *binary_info);
 
 /**
  * @brief Get the all binaries information
  * @details @b #include <binary_manager/binary_manager.h>\n
  *  It requests the binary manager to get the all of binaries' information.
  * @param[out] binary_info_list The address value to receive the binary information list
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_get_update_info_all(binary_update_info_list_t *binary_info_list);
+binmgr_result_type_e binary_manager_get_update_info_all(binary_update_info_list_t *binary_info_list);
 #endif
 
 /**
  * @brief Notify that loaded binary is started to binary manager
  * @details @b #include <binary_manager/binary_manager.h>\n
  *  It sends a message the binary manager to notify that loaded binary is started well.
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_notify_binary_started(void);
+binmgr_result_type_e binary_manager_notify_binary_started(void);
 
 /**
  * @brief Set the callback which will be called when the the states of binaries are changed
  * @details @b #include <binary_manager/binary_manager.h>\n
  *  It sets callback for the changes of binary state.
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_register_state_changed_callback(binmgr_statecb_t handler, void *cb_data);
+binmgr_result_type_e binary_manager_register_state_changed_callback(binmgr_statecb_t handler, void *cb_data);
 
 /**
  * @brief Clear the registered callback for the changes of binary state.
  * @details @b #include <binary_manager/binary_manager.h>\n
  *  It clears registered callback for the changes of binary state.
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0
  */
-int binary_manager_unregister_state_changed_callback(void);
+binmgr_result_type_e binary_manager_unregister_state_changed_callback(void);
 
 /**
  * @brief Get the path to download binary with version.
@@ -112,11 +112,11 @@ int binary_manager_unregister_state_changed_callback(void);
  * @param[in] binary_name The binary name to update
  * @param[in] version The binary version to update
  * @param[out] download_path The path to the download binary. The maximum length is BINARY_PATH_LEN.
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) on success. On failure, negative value is returned.
  * @since TizenRT v3.1 PRE
  */
-int binary_manager_get_download_path(char *binary_name, uint32_t version, char *download_path);
+binmgr_result_type_e binary_manager_get_download_path(char *binary_name, uint32_t version, char *download_path);
 
 /**
  * @brief Get a state of binary
@@ -124,11 +124,11 @@ int binary_manager_get_download_path(char *binary_name, uint32_t version, char *
  *  It sends a message the binary manager to get a state of binary with binary name.
  * @param[in] binary_name The binary name to get state
  * @param[out] state The state of binary with binary_name
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) on success. On failure, negative value is returned.
  * @since TizenRT v3.1 PRE
  */
-int binary_manager_get_state(char *binary_name, int *state);
+binmgr_result_type_e binary_manager_get_state(char *binary_name, int *state);
 
 #endif
 /**
--- a/framework/include/binary_manager/binary_manager.h
+++ b/framework/include/binary_manager/binary_manager.h
@@ -104,15 +104,19 @@ int binary_manager_register_state_changed_callback(binmgr_statecb_t handler, voi
 int binary_manager_unregister_state_changed_callback(void);
 
 /**
- * @brief Create a empty file where to be downloaded with version.
+ * @brief Get the path to download binary with version.
  * @details @b #include <binary_manager/binary_manager.h>\n
- *  It creates and opens a empty file where to be downloaded binary with version.
- *  User can perform file operations with returned fd : read, write, lseek, close.
+ *  User can perform file operations with returned filepath : open, read, write, lseek, close.
+ *  In case of kernel binary, it returns a path of inactive kernel partition.
+ *  Otherwise, it creates empty file with version and returns a path of the file.
+ * @param[in] binary_name The binary name to update
+ * @param[in] version The binary version to update
+ * @param[out] download_path The path to the download binary. The maximum length is CONFIG_PATH_MAX.
  * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
- *         fd of a created file on success. On failure, negative value is returned.
+ *         0 (BINMGR_OK) on success. On failure, negative value is returned.
  * @since TizenRT v3.1 PRE
  */
-int binary_manager_open_new_entry(char *binary_name, uint32_t version);
+int binary_manager_get_download_path(char *binary_name, uint32_t version, char *download_path);
 
 #endif
 /**
--- a/framework/include/binary_manager/binary_manager.h
+++ b/framework/include/binary_manager/binary_manager.h
@@ -118,6 +118,18 @@ int binary_manager_unregister_state_changed_callback(void);
  */
 int binary_manager_get_download_path(char *binary_name, uint32_t version, char *download_path);
 
+/**
+ * @brief Get a state of binary
+ * @details @b #include <binary_manager/binary_manager.h>\n
+ *  It sends a message the binary manager to get a state of binary with binary name.
+ * @param[in] binary_name The binary name to get state
+ * @param[out] state The state of binary with binary_name
+ * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ *         0 (BINMGR_OK) on success. On failure, negative value is returned.
+ * @since TizenRT v3.1 PRE
+ */
+int binary_manager_get_state(char *binary_name, int *state);
+
 #endif
 /**
  * @}
--- a/framework/src/binary_manager/binary_manager_get_state.c
+++ b/framework/src/binary_manager/binary_manager_get_state.c
@@ -1,0 +1,64 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/***************************************************************************
+ * Included Files
+ ***************************************************************************/
+
+#include <debug.h>
+#include <string.h>
+#include <tinyara/binary_manager.h>
+#include <binary_manager/binary_manager.h>
+#include "binary_manager_internal.h"
+
+int binary_manager_get_state(char *binary_name, int *state)
+{
+	int ret;
+	binmgr_request_t request_msg;
+	binmgr_getstate_response_t response_msg;
+
+	if (binary_name == NULL || strlen(binary_name) > BIN_NAME_MAX - 1) {
+		bmdbg("get_binary_info failed : invalid param.\n");
+		return BINMGR_INVALID_PARAM;
+	}
+
+	ret = binary_manager_set_request(&request_msg, BINMGR_GET_STATE, binary_name);
+	if (ret < 0) {
+		return BINMGR_COMMUNICATION_FAIL;
+	}
+
+	ret = binary_manager_send_request(&request_msg);
+	if (ret < 0) {
+		bmdbg("Failed to send request msg %d\n", ret);
+		return BINMGR_COMMUNICATION_FAIL;
+	}
+
+	ret = binary_manager_receive_response(&response_msg, sizeof(binmgr_getstate_response_t));
+	if (ret < 0) {
+		bmdbg("Failed to receive response msg %d\n", ret);
+		return BINMGR_COMMUNICATION_FAIL;
+	}
+
+	if (response_msg.result == BINMGR_OK) {
+		*state = response_msg.state;
+	} else {
+		bmdbg("Binary manager getstate FAIL %d\n", response_msg.result);
+	}
+
+	return response_msg.result;
+}
--- a/framework/src/binary_manager/binary_manager_get_state.c
+++ b/framework/src/binary_manager/binary_manager_get_state.c
@@ -26,9 +26,9 @@
 #include <binary_manager/binary_manager.h>
 #include "binary_manager_internal.h"
 
-int binary_manager_get_state(char *binary_name, int *state)
+binmgr_result_type_e binary_manager_get_state(char *binary_name, int *state)
 {
-	int ret;
+	binmgr_result_type_e ret;
 	binmgr_request_t request_msg;
 	binmgr_getstate_response_t response_msg;
 
@@ -38,20 +38,20 @@ int binary_manager_get_state(char *binary_name, int *state)
 	}
 
 	ret = binary_manager_set_request(&request_msg, BINMGR_GET_STATE, binary_name);
-	if (ret < 0) {
-		return BINMGR_COMMUNICATION_FAIL;
+	if (ret != BINMGR_OK) {
+		return ret;
 	}
 
 	ret = binary_manager_send_request(&request_msg);
-	if (ret < 0) {
+	if (ret != BINMGR_OK) {
 		bmdbg("Failed to send request msg %d\n", ret);
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
 	}
 
 	ret = binary_manager_receive_response(&response_msg, sizeof(binmgr_getstate_response_t));
-	if (ret < 0) {
+	if (ret != BINMGR_OK) {
 		bmdbg("Failed to receive response msg %d\n", ret);
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
 	}
 
 	if (response_msg.result == BINMGR_OK) {
--- a/framework/src/binary_manager/binary_manager_notify.c
+++ b/framework/src/binary_manager/binary_manager_notify.c
@@ -27,20 +27,20 @@
 #include <binary_manager/binary_manager.h>
 #include "binary_manager_internal.h"
 
-int binary_manager_notify_binary_started(void)
+binmgr_result_type_e binary_manager_notify_binary_started(void)
 {
-	int ret;
+	binmgr_result_type_e ret;
 	binmgr_request_t request_msg;
 
 	ret = binary_manager_set_request(&request_msg, BINMGR_NOTIFY_STARTED, BINMGR_REQUEST_MQ);
-	if (ret < 0) {
-		return BINMGR_COMMUNICATION_FAIL;
+	if (ret != BINMGR_OK) {
+		return ret;
 	}
 
 	ret = binary_manager_send_request(&request_msg);
-	if (ret < 0) {
+	if (ret != BINMGR_OK) {
 		bmdbg("Failed to send request msg %d\n", ret);
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
 	}
 
 	return BINMGR_OK;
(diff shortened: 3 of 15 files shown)


framework/src/binary_manager/binary_manager_update.c
```diff
@@ -161,12 +161,8 @@ int binary_manager_open_new_entry(char *binary_name, uint32_t version)
...
 
 	if (response_msg.result == BINMGR_OK) {
-		fd = open(response_msg.binpath, O_RDWR | O_CREAT, 0666);
-		if (fd < 0) {
-			bmdbg("Failed to open file %s : %d\n", response_msg.binpath, errno);
-			return BINMGR_OPERATION_FAIL;
-		}
-		return fd;
+		bmvdbg("Create file path : %s\n", response_msg.binpath);
+		strncpy(download_path, response_msg.binpath, CONFIG_PATH_MAX);
```
Taejun-Kwon, line 165: I think strlen(response_msg.binpath) is correct, instead of CONFIG_PATH_MAX

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -72,6 +72,7 @@ static void binary_update_download_binary(binary_update_info_t *binary_info, int
 	uint8_t buffer[BUFFER_SIZE];
 	binary_header_t header_data;
 	char filepath[CONFIG_PATH_MAX];
+	char new_filepath[CONFIG_PATH_MAX];
 
 	snprintf(filepath, CONFIG_PATH_MAX, "%s/%s_%u", BINARY_DIR_PATH, APP_NAME, (uint32_t)binary_info->version);
```
Taejun-Kwon, line 77: length of BINARY_DIR_PATH can be smaller than CONFIG_PATH_MAX, then you should calculate size, and use it instead of CONFIG_PATH_MAX also if length is smaller than CONFIG_PATH_MAX, add null character at the last of string.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -127,7 +127,7 @@ int binary_manager_get_update_info_all(binary_update_info_list_t *binary_info_li
 	return response_msg.result;
 }
 
-int binary_manager_open_new_entry(char *binary_name, uint32_t version)
+int binary_manager_open_new_entry(char *binary_name, uint32_t version, char *download_path)
```
jeongchanKim, line 130: I think get is better than open.

Taejun-Kwon: @jeongchanKim Me Too.

jeongchanKim (commented):
Please change the commit msg : open -> get

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -92,11 +93,17 @@ static void binary_update_download_binary(binary_update_info_t *binary_info, int
 
 	new_version = header_data.bin_ver + 1;
 
-	write_fd = binary_manager_open_new_entry(APP_NAME, new_version);
+	ret = binary_manager_get_download_path(APP_NAME, new_version, new_filepath);
+	if (ret < 0) {
+		fail_cnt++;
+		printf("binary_manager_get_download_path FAIL %d,\n", ret);
+		return;
```
Taejun-Kwon, line 100: hmm...I think you had better return type of this function, because after call this function, binary_update_invalid_binary_test  try reload... You think is it safe? how do you think?

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -225,6 +246,7 @@ static void binary_update_download_new_binary(void)
 	/* Write the binary header. */
 	ret = write(write_fd, (FAR uint8_t *)&header_data, sizeof(binary_header_t));
 	if (ret != sizeof(binary_header_t)) {
+		ret = ERROR;
 		printf("Failed to write header: %d\n", ret);
```
jeongchanKim, line 250: I think you'd better print the `ret` before assigning to `ERROR`.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -423,67 +465,105 @@ static void binary_update_same_version_test(void)
...
 	int ret;
-	char filepath[CONFIG_PATH_MAX];
+	char filepath[BINARY_PATH_LEN];
 	binary_update_info_t pre_bin_info;
 	binary_update_info_t cur_bin_info;
 
-	binary_update_getinfo(APP_NAME, &pre_bin_info);
+	ret = binary_update_getinfo(APP_NAME, &pre_bin_info);
+	if (ret != OK) {
+		return;
```
jeongchanKim, line 474: How about printing the log about fail case?

os/include/tinyara/binary_manager.h
```diff
@@ -70,6 +70,14 @@
 /* Mount point for User binaries */
 #define BINARY_DIR_PATH                  CONFIG_MOUNT_POINT"bins"
 
+/* The length of binary file or kernel partition path.
+ * A path is same like below:
+ *  - User filepath : "/<mount_path>/<binary_name>_<binary_version>"
+ *         length : CONFIG_NAME_MAX + BIN_NAME_MAX + binary_version (32) + /,_(3) + '\0'(1)
+ *  - Kernel partpath : "/dev/mtdblock#" defined as BINMGR_DEVNAME_FMT
+ */
+#define BINARY_PATH_LEN                  (CONFIG_NAME_MAX + BIN_NAME_MAX + 36)
```
jeongchanKim, line 79: I think you'd better define 36 with strlen or some macro.

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -274,11 +274,11 @@ int binary_manager_create_entry(int requester_pid, char *bin_name, int version)
 	response_msg.result = BINMGR_OPERATION_FAIL;
 
 	/* Create a new file */
-	snprintf(filepath, CONFIG_PATH_MAX, "%s/%s_%d", BINARY_DIR_PATH, bin_name, version);
+	snprintf(filepath, BINARY_PATH_LEN, "%s/%s_%d", BINARY_DIR_PATH, bin_name, version);
 	fd = open(filepath, O_RDWR | O_CREAT, 0666);
 	if (fd > 0) {
 		bmvdbg("Created file '%s' for binary %s\n", filepath, bin_name);
-		strncpy(response_msg.binpath, filepath, CONFIG_PATH_MAX);
+		strncpy(response_msg.binpath, filepath, BINARY_PATH_LEN);
```
jeongchanKim, line 281: You'd better copy with `strlen(filepath) + 1` for security.

os/include/tinyara/binary_manager.h
jeongarmy: It can't define value with function because it is used in array like line 221.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: binary_update_getinfo prints error message.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -87,17 +88,23 @@ static void binary_update_download_binary(binary_update_info_t *binary_info, int
...
 	}
 
 	new_version = header_data.bin_ver + 1;
 
-	write_fd = binary_manager_open_new_entry(APP_NAME, new_version);
+	ret = binary_manager_get_download_path(APP_NAME, new_version, new_filepath);
+	if (ret < 0) {
+		fail_cnt++;
+		printf("binary_manager_get_download_path FAIL %d,\n", ret);
+		return ERROR;
```
Taejun-Kwon, line 100: Now readfd is till open.
you'd better make error_out_fd1, error_out_fd2, instead of fds.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -172,14 +186,15 @@ static void binary_update_download_new_binary(void)
...
+	char filepath[BINARY_PATH_LEN];
+	char new_filepath[BINARY_PATH_LEN];
 	int filesize;
 	FILE *fp;
 
 	/* Get 'micom' binary info */
 	binary_manager_get_update_info(APP_NAME, &bin_info);
 
-	snprintf(filepath, CONFIG_PATH_MAX, "%s/%s_%u", BINARY_DIR_PATH, APP_NAME, (uint32_t)bin_info.version);
+	snprintf(filepath, BINARY_PATH_LEN, "%s/%s_%u", BINARY_DIR_PATH, APP_NAME, (uint32_t)bin_info.version);
```
Taejun-Kwon, line 197: Line 200~26
```
off_t filesize;
read_fd = open(filepath, O_RDONLY);
if (read_fd < 0) {
...
}
fseek = lseek(fd, 0, SEEK_END); //lseek returns changed pos of file.
if (filesize <= 0) ..... 
```

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -207,15 +222,21 @@ static void binary_update_download_new_binary(void)
...
-		return;
+		return ERROR;
 	}
 
-	write_fd = binary_manager_open_new_entry(NEW_APP_NAME, NEW_APP_VERSION);
+	ret = binary_manager_get_download_path(NEW_APP_NAME, NEW_APP_VERSION, new_filepath);
+	if (ret < 0) {
+		fail_cnt++;
+		printf("binary_manager_get_download_path FAIL %d,\n", ret);
+		return ERROR;
```
Taejun-Kwon, line 232: here and line 239, read_fd still alive.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -402,17 +438,23 @@ static void binary_update_unregister_state_changed_callback(void)
...
 		printf("Unregister state changed callback FAIL, ret %d\n", ret);
 	}
+
+	return ret;
 }
 
 static void binary_update_same_version_test(void)
 {
-	int ret;
+	int ret;
```
Taejun-Kwon, line 447: This function is tightly related to result of other function, in this case you had better change function type to int to notify to user clearly. but it just sample app you can ignore this comment.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -423,67 +465,105 @@ static void binary_update_same_version_test(void)
...
+	if (ret != OK) {
+		return;
+	}
 
 	sleep(2);
 
-	binary_update_getinfo(APP_NAME, &cur_bin_info);
+	ret = binary_update_getinfo(APP_NAME, &cur_bin_info);
+	if (ret != OK) {
+		return;
```
Taejun-Kwon, line 492: yes... like this funcion. Keep void is very dangerous but same as above. you can ignore because it just sample app.

os/include/tinyara/binary_manager.h
```diff
@@ -70,18 +70,37 @@
...
```
Taejun-Kwon, line 79: We have CONFIG_PATH_MAX. so you have to do some extra.

os/include/tinyara/binary_manager.h
```diff
@@ -70,18 +70,37 @@
...
 	BINARY_UNLOADED = 1,          /* Binary is unloaded */
 	BINARY_READYTOUNLOAD = 2,     /* Binary will be unloaded */
 };
 
+enum binary_state_e {
+	BINARY_UNREGISTERED = 0,     /* File is not existing or Partition is not registered */
+	BINARY_INACTIVE = 1,         /* File is not existing or Partition is registered, but binary is not loaded yet */
+	BINARY_LOADING_DONE = 2,     /* Loading binary is done */
+	BINARY_RUNNING = 3,          /* Loaded binary gets scheduling */
+	BINARY_WAITUNLOAD = 4,       /* Loaded binary would be unloaded */
```
Taejun-Kwon, line 93: According to description, BINARY_LOADED & BINARY_UNLODING is correct, I think?

os/kernel/binary_manager/binary_manager_getinfo.c
```diff
@@ -135,6 +135,41 @@ int binary_manager_get_available_size(int bin_idx)
...
 }
 
+/****************************************************************************
+ * Name: binary_manager_get_state
+ *
+ * Description:
+ *	 This function gets binary state with binary name.
+ *
+ ****************************************************************************/
+void binary_manager_get_state(int requester_pid, char *bin_name)
```
jeongchanKim, line 145: There is a same function name. You'd better rename one of them.

framework/include/binary_manager/binary_manager.h
```diff
@@ -73,6 +73,16 @@ int binary_manager_get_update_info(char *binary_name, binary_update_info_t *bina
 int binary_manager_get_update_info_all(binary_update_info_list_t *binary_info_list);
 #endif
 
+/**
+ * @brief Get a state of binary
+ * @details @b #include <binary_manager/binary_manager.h>\n
+ *  It sends a message the binary manager to get a state of binary with binary name.
+ * @return A state value of binary_state_e in <tinyara/binary_manager.h> on success.
+ *         On failure, A negative value of binmgr_response_result_type in <tinyara/binary_manager.h> is returned.
+ * @since TizenRT v3.0 PRE
```
jeongchanKim, line 82: Some APIs are v3.0 PRE and some are v3.1 PRE. Which is correct?

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: done

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: done

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: Each functions which have a name binary_update_XXX_test are independent. So it doesn't have to return result I think.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: Each functions which have a name binary_update_XXX_test are independent so it doesn't have to return result I think.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: modified

framework/src/binary_manager/binary_manager_get_state.c
```diff
@@ -0,0 +1,65 @@
...
+ * Included Files
+ ***************************************************************************/
+
+#include <debug.h>
+#include <string.h>
+#include <tinyara/binary_manager.h>
+#include <binary_manager/binary_manager.h>
+#include "binary_manager_internal.h"
+
+int binary_manager_get_state(char *binary_name)
```
Taejun-Kwon, line 29: It is framework, instead of int you should use name of enum which include errors in line 37, 42, 48...

os/kernel/binary_manager/binary_manager_getinfo.c
```diff
@@ -135,6 +135,41 @@ int binary_manager_get_available_size(int bin_idx)
 	return size;
 }
 
+/****************************************************************************
+ * Name: binary_manager_get_state
```
jeongchanKim, line 139: Please check the function name.

os/kernel/binary_manager/binary_manager_getinfo.c
```diff
@@ -135,6 +135,41 @@ int binary_manager_get_available_size(int bin_idx)
...
+		return;
+	}
+	snprintf(q_name, BIN_PRIVMQ_LEN, "%s%d", BINMGR_RESPONSE_MQ_PREFIX, requester_pid);
+
+	memset((void *)&response_msg, 0, sizeof(binmgr_getstate_response_t));
+	response_msg.result = BINMGR_NOT_FOUND;
+
+	bin_count = binary_manager_get_ucount();
+	for (bin_idx = 1; bin_idx <= bin_count; bin_idx++) {
+		if (!strncmp(BIN_NAME(bin_idx), bin_name, BIN_NAME_MAX)) {
```
jeongchanKim, line 163: I think you'd better compare with `strlen(bin_name) + 1` instead of `BIN_NAME_MAX`.

framework/src/binary_manager/binary_manager_get_state.c
```diff
@@ -0,0 +1,65 @@
...
+	int ret;
+	binmgr_request_t request_msg;
+	binmgr_getstate_response_t response_msg;
+
+	if (binary_name == NULL || strlen(binary_name) > BIN_NAME_MAX - 1) {
+		bmdbg("get_binary_info failed : invalid param.\n");
+		return BINMGR_INVALID_PARAM;
+	}
+
+	ret = binary_manager_set_request(&request_msg, BINMGR_GET_STATE, binary_name);
```
Taejun-Kwon, line 40: binary_manager_set_request should return BINMGR_COMMUNICATION_FAIL or something, instead of ERROR to reduce unnecessary code here.
Actually line 45 in binary_manager_set_request should return BINMGR_INVALID_PARAM or something. of course line 35 in this function checking it but it is necessary if there's chance that some other place also call binary_manager_set_request.

os/include/tinyara/binary_manager.h
```diff
@@ -70,18 +70,38 @@
...
 #define BINARY_DIR_PATH                  CONFIG_MOUNT_POINT"bins"
 
+/* The length of binary file or kernel partition path.
+ * A path is same like below:
+ *  - User filepath : "/<mount_path>/<binary_name>_<binary_version>"
+ *         length : CONFIG_NAME_MAX + BIN_NAME_MAX + binary_version (32) + /,_(3) + '\0'(1)
+ *  - Kernel partpath : "/dev/mtdblock#" defined as BINMGR_DEVNAME_FMT
+ * So define binary path length as (CONFIG_NAME_MAX + BIN_NAME_MAX + 40) with some extra.
+ */
+#define BINARY_PATH_LEN                  (CONFIG_NAME_MAX + BIN_NAME_MAX + 40)
```
jeongchanKim, line 80: What is the meaning of 40?

framework/include/binary_manager/binary_manager.h
```diff
@@ -104,15 +104,29 @@ int binary_manager_register_state_changed_callback(binmgr_statecb_t handler, voi
...
+
+/**
+ * @brief Get a state of binary
+ * @details @b #include <binary_manager/binary_manager.h>\n
+ *  It sends a message the binary manager to get a state of binary with binary name.
+ * @return A state value of binary_state_e in <tinyara/binary_manager.h> on success.
+ *         On failure, A negative value of binmgr_response_result_type in <tinyara/binary_manager.h> is returned.
+ * @since TizenRT v3.1 PRE
+ */
+int binary_manager_get_state(char *binary_name);
```
jeongchanKim, line 129: How about defining `binmgr_response_result_type` and using this type as return?

framework/src/binary_manager/binary_manager_get_state.c
```diff
@@ -0,0 +1,65 @@
...
+	if (ret < 0) {
+		bmdbg("Failed to receive response msg %d\n", ret);
+		return BINMGR_COMMUNICATION_FAIL;
+	}
+
+	if (response_msg.result == BINMGR_OK) {
+		ret = response_msg.state;
+	} else {
+		bmdbg("Binary manager getstate FAIL %d\n", response_msg.result);
+		ret = response_msg.result;
```
Taejun-Kwon, line 61: If defined values for state(guess binary_state_e) include error case, you don't have to use structure, enum will be enough.
Please make sure that name of this function is 'get_state'
So, here you got message through the mq from lower, you should parse it and return newly defined values in this framework to requested layer.
To sum up,
1. Define state values in framework.
2. Here we need to parse message from lower layer, and return newly defined values to user.
This will help upper layer to reduce reference code, that's why we make framework, actually.

At least, here you should return error value of framework defined, not result itself.
Currently this returns values in framework or msg from lower layer...it's not good.

os/include/tinyara/binary_manager.h
```diff
@@ -70,18 +70,38 @@
...
 /* Binary states used in state callback */
 enum binary_statecb_state_e {
 	BINARY_STARTED = 0,           /* Binary is started */
 	BINARY_UNLOADED = 1,          /* Binary is unloaded */
 	BINARY_READYTOUNLOAD = 2,     /* Binary will be unloaded */
 };
 
+enum binary_state_e {
+	BINARY_UNREGISTERED = 0,     /* File is not existing or Partition is not registered */
+	BINARY_INACTIVE = 1,         /* File is not existing or Partition is registered, but binary is not loaded yet */
```
jeongchanKim, line 91: There is one situation (`File is not existing`), but two states(`BINARY_UNREGISTERED`, `BINARY_INACTIVE`).
I think this should be `File is existing`.

os/include/tinyara/binary_manager.h
```diff
@@ -202,10 +219,16 @@ typedef struct binmgr_request_s binmgr_request_t;
 
 struct binmgr_createbin_response_s {
 	int result;
-	char binpath[CONFIG_PATH_MAX];
+	char binpath[BINARY_PATH_LEN];
 };
 typedef struct binmgr_createbin_response_s binmgr_createbin_response_t;
 
+struct binmgr_getstate_response_s {
+	int result;
```
jeongchanKim, line 227: If you use int, not enum type, you can use half of the int for each. Because `result` and `state` cannot be over than half of the int-max.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: done

os/include/tinyara/binary_manager.h
jeongarmy: There was a typo. I modified

framework/src/binary_manager/binary_manager_get_state.c
jeongarmy: Yes. I agree with your opinion. Currently all values are defined in <tinyara/binary_manager.h> and user and kernel use them. so user should include that file for binary manager usage.
But what I worried about is that there are duplicated definitions in framework and kernel if they are defined in framework also.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -79,16 +79,15 @@ static void binary_update_download_binary(binary_update_info_t *binary_info, int
...
 	}
 
 	/* Read the binary header. */
 	ret = read(read_fd, (FAR uint8_t *)&header_data, sizeof(binary_header_t));
 	if (ret != sizeof(binary_header_t)) {
 		fail_cnt++;
 		printf("Failed to read header %s: %d\n", filepath, ret);
-		close(read_fd);
-		return;
+		goto errout_with_close_fd1;
```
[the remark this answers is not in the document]
jeongarmy: 1 and 2 means one and two, not first, second.
errout_with_fd2 includes closing both readfd and writefd.

framework/include/binary_manager/binary_manager.h
```diff
@@ -43,76 +43,92 @@
...
+binmgr_result_type_e binary_manager_notify_binary_started(void);
 
 /**
  * @brief Set the callback which will be called when the the states of binaries are changed
  * @details @b #include <binary_manager/binary_manager.h>\n
  *  It sets callback for the changes of binary state.
- * @return A defined value of binmgr_response_result_type in <tinyara/binary_manager.h>
+ * @return A defined value of binmgr_result_type_e in <tinyara/binary_manager.h>
  *         0 (BINMGR_OK) On success. On failure, negative value is returned.
  * @since TizenRT v3.0 PRE
```
jeongchanKim, line 92: This is already changed by PR#4548.
Please rebase this.

framework/src/binary_manager/binary_manager_notify.c
```diff
@@ -27,7 +27,7 @@
 #include <binary_manager/binary_manager.h>
 #include "binary_manager_internal.h"
 
-int binary_manager_notify_binary_started(void)
+binmgr_result_type_e binary_manager_notify_binary_started(void)
```
jeongchanKim, line 30: In this function, `binary_manager_xxx` APIs return `binmgr_result_type_e`, not int. But you declared `int ret` for getting return of them. Please check this.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -106,14 +103,12 @@ int binary_manager_get_update_info_all(binary_update_info_list_t *binary_info_li
 
 	ret = binary_manager_send_request(&request_msg);
 	if (ret < 0) {
-		bmdbg("Failed to send request msg %d\n", ret);
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
```
jeongchanKim, line 106: `ret` is not `binmgr_result_type_e` type.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -145,28 +139,22 @@ int binary_manager_open_new_entry(char *binary_name, uint32_t version)
 
 	ret = binary_manager_set_request(&request_msg, BINMGR_CREATE_BIN, &data);
 	if (ret < 0) {
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
```
jeongchanKim, line 142: Same as above.

os/include/tinyara/binary_manager.h
jeongarmy: The maximum length of user file path is 36 and I added some extra.

framework/src/binary_manager/binary_manager_get_state.c
```diff
@@ -0,0 +1,64 @@
...
+	binmgr_getstate_response_t response_msg;
+
+	if (binary_name == NULL || strlen(binary_name) > BIN_NAME_MAX - 1) {
+		bmdbg("get_binary_info failed : invalid param.\n");
+		return BINMGR_INVALID_PARAM;
+	}
+
+	ret = binary_manager_set_request(&request_msg, BINMGR_GET_STATE, binary_name);
+	if (ret < 0) {
+		return BINMGR_COMMUNICATION_FAIL;
```
Taejun-Kwon, line 42: Below is correct for line 40~51, I think
```
if (ret != BINMGR_OK) {
  return ret;
}
```

framework/src/binary_manager/binary_manager_get_state.c
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if (response_msg.result == BINMGR_OK) {
+		*state = response_msg.state;
+	} else {
+		bmdbg("Binary manager getstate FAIL %d\n", response_msg.result);
+	}
+
+	return response_msg.result;
```
Taejun-Kwon, line 63: Return type should be one of binmgr_result_type_e
```
	if (response_msg.result == BINMGR_OK) {
		*state = response_msg.state;
	} else {
		bmdbg("Binary manager getstate FAIL %d\n", response_msg.result);
                return BINMGR_OPERATION_FAIL; // or some other errors.
	}

	return BINMGR_OK;
}
```
so caller can use below code. (it just example)
```
if (binary_manager_get_state(path, &state) == BINMGR_OK) {
    switch (state) {
        case A:
        ...
    }
}
```

framework/src/binary_manager/binary_manager_notify.c
Taejun-Kwon:
```
binmgr_result_type_e ret;
ret = binary_manager_set_request(&request_msg, BINMGR_NOTIFY_STARTED, BINMGR_REQUEST_MQ);
if (ret != BINMGR_OK) {
   return ret;
}
```

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -30,31 +30,30 @@
...
-int binary_manager_update_binary(char *binary_name)
+binmgr_result_type_e binary_manager_update_binary(char *binary_name)
 {
 	int ret;
 	binmgr_request_t request_msg;
 
 	if (binary_name == NULL || strlen(binary_name) > BIN_NAME_MAX - 1) {
 		bmdbg("load_new_binary failed : invalid binary name\n");
-		return BINMGR_INVALID_PARAM;
+		return ret;
```
Taejun-Kwon, line 40: ret is not initialized here.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -30,31 +30,30 @@
...
 	}
 
 	ret = binary_manager_set_request(&request_msg, BINMGR_UPDATE, binary_name);
 	if (ret < 0) {
-		return BINMGR_COMMUNICATION_FAIL;
+		return ret;
```
Taejun-Kwon, line 45: This work actually, but to make it more explicitly, 'binmgr_result_type_e ret' is better option.

framework/src/binary_manager/binary_manager_get_state.c
Taejun-Kwon: You'd better divide values. do not put all values to <tinyara/binary_manager.h>.
Framework is necessary to provide small set of reference code to above layer to make their application more easy & efficiently.
So put some values for user to some headers in framework/include, application doesn't have to include <tinyara/binary_manager.h> because they do not need all values in there.
Below is sample of re-definition of lower values. (lower has 12 of PCM_FORMAT but app doesn't need it)
```
typedef enum audio_format_type_e {
	/* Signed 8 bit */
	AUDIO_FORMAT_TYPE_S8 = PCM_FORMAT_S8,
	/* Signed 16 bit use this as a default */
	AUDIO_FORMAT_TYPE_S16_LE = PCM_FORMAT_S16_LE,
	/* Signed 32 bit */
	AUDIO_FORMAT_TYPE_S32_LE = PCM_FORMAT_S32_LE
} audio_format_type_t;
```

framework/src/binary_manager/binary_manager_get_state.c
jeongarmy: Framework just sends commands and receives result from binary manager.
So it just returns result received from binary manager here because binary manager has all actual data and does operations on them.

Taejun-Kwon: Let's think about why framework is necessary.
Framework is kind of abstracted and mixed layer of lower code to make application more easier.
So, usually it doesn't use whole of code in lower layer, and take some of lower codes.
(For example, camera framework takes most of camera code, and some of sound code, some of graphic code)
So with this framework, application doesn't have to care about **ANY OF** lower layer's code.
But with this framework, application should refer to both of framework and bm driver.

That's why I commented above, (https://github.com/Samsung/TizenRT/pull/4486#discussion_r500689499) refining is necessary.

Anyway your function's type is binmgr_response_result_type, it's not response_msg.result, right?

jeongarmy: A response_msg.result has binmgr_response_result_type type also so it returns  response_msg.result here.
Anyway I agree with your opinion. I will modify them with another commit.

sunghan-chang (approved):

PR #4487, opened 2020-09-02
os/binfmt : Fix elf section alignment for ARMv8M mpu

kishore-sn:
MPU region size and alignment requirements for armv8m are
different from those of armv7m. This commit modfiles the
elf section allocation accordingly. Also, some sanity checks
for region size and alignment is added for armv8m mpu code.

The following logs show the mpu base and size during a memfault. 
It can be seen that lower 5 bits of base and size are zeros, indicating alignment to 32 bytes.

`mpu_show_regioninfo: *****************************************************************************`
`mpu_show_regioninfo: *REGION_NO.        BASE_ADDRESS        SIZE           STATUS        ACCESS         NO_EXECUTE*`
`mpu_show_regioninfo: *****************************************************************************`
`mpu_show_regioninfo:        0           10055C60           11860               1               3               0`
`mpu_show_regioninfo:        1           100674E0            86A0               1               3               1`
`mpu_show_regioninfo:        2            2000020           7EAE0               1               1               1`


sunghan-chang:
1. Could you update the operation with real data? Is that working as what you expected?
2. Could you fix the violation of coding rule as shown above?

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -105,7 +105,21 @@ static int allocateregions(FAR struct elf_loadinfo_s *loadinfo)
...
+	/* ARMV7M requires addresses to be aligned to the size of the region.
+	 * In this case, we align the first region to the size of first region.
+	 * Since the regions are arranged in descending order of sizes and each
+	 * region size is a power of two, the remaining regions will also be
+	 * aligned to their size.
+	 */
 	*allocs[0] = (uintptr_t)kmm_memalign(sizes[0], totalsize);
+#elif CONFIG_ARMV8M_MPU
+	/* ARMV8M requires 32 bytes alignment of addresses */
+	*allocs[0] = (uintptr_t)kmm_memalign(MPU_ALIGNMENT_BYTES, totalsize);
```
jeongchanKim, line 118: With v8m, I think we don't need to concatenate the regions for memory usage flexibility.
How about your opinion?

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -200,9 +214,20 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
...
 	loadinfo->rosize += loadinfo->datasize - loadinfo->binp->bsssize;
 
+#ifdef CONFIG_ARMV7M_MPU
+	/* ARMV7M requires MPU region size to be a power of two */
 	loadinfo->textsize = 1 << mpu_log2regionceil(0, loadinfo->textsize);
 	loadinfo->rosize = 1 << mpu_log2regionceil(0, loadinfo->rosize);
 	datamemsize = 1 << mpu_log2regionceil(0, datamemsize);
+#elif CONFIG_ARMV8M_MPU
+	/* ARMV8M requires MPU region size to be aligned to 32 bytes */
+	loadinfo->textsize = MPU_ALIGN_UP(loadinfo->textsize);
```
jeongchanKim, line 224: As above my comment, if we don't need to concatenate the regions, then we can just call memalign with 32bytes in here.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -200,9 +214,20 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo, size_t textsize, size
...
+	loadinfo->rosize = MPU_ALIGN_UP(loadinfo->rosize);
+	datamemsize = MPU_ALIGN_UP(datamemsize);
+#else
+#error "Unknown MPU version. Expected either ARMV7M or ARMV8M"
```
jeongchanKim, line 228: We can use binary loading regardless of mpu, even it is not a good way. So I think this can be a problem.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: True. I modified it.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: Done.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: This is true. But there are many places where MPU is tightly coupled to APP separation. 
If we need to make them independent, then we better refractor the code properly. 
Please let me know if this is required. We will do it in a separate PR.

sunghan-chang (commented):
Let me suggest code cleanup.
1. We could add this PR as current. (Need to check below comment)
2. Even if we merge this and it works well, I request you to tidy up.
    1. Let's remove MPU config in loader. With new config A, it allocates merged block. Without A, it allocates three regions seperately.
    2. Board config or ARM config has a dependency with A by selection.
I want to keep loader common. I don't want to include two different algorithm wuth two different config in loader. That's why I suggest above.
Could you let me know your opinion? Kishore, Jeongchan, Ahreum and Sangam?

sunghan-chang:
@kishore-sn Could you let me know your opinion about my suggestion?

os/arch/arm/src/armv8-m/mpu.h
```diff
@@ -209,14 +209,14 @@
...
-#define MPU_MEM_ATTR6                   (DEVICE_NG_NR_NE)
-#define MPU_MEM_ATTR7                   (DEVICE_NG_NR_NE)
+#define MPU_MEM_ATTR0			((NORMAL_NC << 4) | NORMAL_NC)
+#define MPU_MEM_ATTR1			((NORMAL_WT_T_RA << 4)  | NORMAL_WT_T_RA)
+#define MPU_MEM_ATTR2			((NORMAL_WB_T_RWA << 4)  | NORMAL_WB_T_RWA)
+#define MPU_MEM_ATTR3			(DEVICE_NG_NR_NE)
+#define MPU_MEM_ATTR4			(DEVICE_NG_NR_NE)
+#define MPU_MEM_ATTR5			(DEVICE_NG_NR_NE)
+#define MPU_MEM_ATTR6			(DEVICE_NG_NR_NE)
+#define MPU_MEM_ATTR7			(DEVICE_NG_NR_NE)
```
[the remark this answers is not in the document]
kishore-sn: Yes. You are right. I had changed that due to some mis-understanding. I reverted the change.

kishore-sn:
@sunghan-chang 
I agree with your suggestion to not have mpu related configs in loader. 
We will do that in a new PR, since it requires some planning. 
Also, there is an issue when app relaoad optimization is disabled. But it is not related to this PR. 
So, you can merge this PR. I will handle the issues separately.

sunghan-chang (approved):

PR #4491, opened 2020-09-03
Dbuild update

jeongchanKim:


jeongchanKim:
> How about 1050 for the 2nd commit?

There is no sudo permission in 1050 case.

an4967 (commented):
LGTM

sunghan-chang (approved):

PR #4492, opened 2020-09-04
fs/smartfs : Update log infor of journal & mtd operation

Taejun-Kwon:
Apply log level for smart_journal_print_log, also add more info if write_sector failed

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -4195,7 +4195,7 @@ static int smart_writesector(FAR struct smart_struct_s *dev, unsigned long arg)
 		ret = MTD_BWRITE(dev->mtd, physsector * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)dev->rwbuffer);
 #endif
 		if (ret != dev->mtdBlksPerSector) {
-			fdbg("Error writing to physical sector %d\n", physsector);
+			fdbg("Error writing to physical sector %d ret : %d\n", physsector, ret);
 			return -EIO;
 		}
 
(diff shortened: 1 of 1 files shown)


pillip8282 (approved):
lgtm

sunghan-chang (approved):

os/fs/driver/mtd/smart.c
```diff
@@ -4852,8 +4852,11 @@ static int smart_journal_init(FAR struct smart_struct_s *dev)
 {
 	int ret = OK;
 	dev->journal_seq = 0;
-
+#ifdef CONFIG_DEBUG_FS_INFO
+	ret = smart_journal_scan(dev, true);
```
amogh-samsung, line 4856: Does this have to be done this way? Seems very unconventional.
Instead of passing extra parameter to smart_journal_scan, can we not check
`#ifdef CONFIG_DEBUG_FS_INFO` or `#ifndef CONFIG_DEBUG_FS_INFO`
instead of
`if (print_dump)` or `if (!print_dump)`
Inside function smart_journal_scan.

Taejun-Kwon: Please see inside of smart_journal_scan, if print_dump is false, it shows only active journal only.

amogh-samsung: Since print_dump depends on CONFIG_DEBUG_FS_INFO
Instead of passing print_dump and checking True or False
Can we check if CONFIG_DEBUF_FS_INFO is defined in smart_journal_scan or not?

Taejun-Kwon: I prefer remove CONFIG as possible as I can, if smart_journal_scan has many parameters, I would consider, but now it doesnt.
actually smart_journal_scan is kind of utility, this can be added anywhere when we debug code.(c.f after recovery to check changed state.)
So I added last params.

amogh-samsung: **Instead of this**
```
smart_journal_init()
{
....
smart_journal_scan(True);
....
}

smart_journal_scan(bool print_dump)
{
if(print_dump)
smart_print_journal_log();
....
}
```

**How about this?**
```
smart_journal_init()
{
.....
smart_journal_scan(); -> No boolean passed
....
}

smart_journal_scan()
{
bool print_dump
#ifdef CONFIG_DEBUG_FS_INFO
print_dump = True;
#else 
print_dump = False;
#endif 

....
if (print_dump)
smart_print_journal_log();
....
}
```

PR #4495, opened 2020-09-04
fs/smartfs : Modify debug message

Taejun-Kwon:
Display sector scan log if DEBUG_INFO is enabled, and clean code regarding fsdbg in debug.h

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -1856,7 +1856,7 @@ static int smart_scan(FAR struct smart_struct_s *dev)
 		status_released = SECTOR_IS_RELEASED(header);
 		status_committed = SECTOR_IS_COMMITTED(header);
 #ifdef CONFIG_MTD_SMART_JOURNALING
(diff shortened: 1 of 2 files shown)


os/include/debug.h
```diff
@@ -1159,6 +1159,7 @@ int get_errno(void);
 #else
 #define fdbg        (void)
 #define flldbg      (void)
+#define fsdbg       (void)
```
sunghan-chang, line 1162: You need to check line 381, 397 and 401 as well.

os/fs/driver/mtd/smart.c
```diff
@@ -1855,7 +1855,7 @@ static int smart_scan(FAR struct smart_struct_s *dev)
 
 		status_released = SECTOR_IS_RELEASED(header);
 		status_committed = SECTOR_IS_COMMITTED(header);
-#ifdef CONFIG_MTD_SMART_JOURNALING
+#if defined(CONFIG_MTD_SMART_JOURNALING) && defined(CONFIG_DEBUG_FS_INFO)
```
sunghan-chang, line 1858: That's `fdbg`, not `fvdbg so that you should cover that with `CONFIG_DEBUG_FS_ERROR`.

sunghan-chang (approved):

PR #4497, opened 2020-09-04
os/arch/arm: Stack protection code cleanup.

sangam-swami:
Cleanup stack overflow protection by adding guard region to kernel
tasks as well

### DIFF
--- a/os/arch/arm/src/common/up_createstack.c
+++ b/os/arch/arm/src/common/up_createstack.c
@@ -164,6 +164,15 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 {
 	size_t stack_alloc_size = stack_size;
 
+	/* new thread does not have the uheap value yet, It's added in the task_
+	 * schedsetup function.
+	 * Parent and child threads must have the same uheap value, so we are
+	 * using the parent thread uheap
+	 */
+#if defined(HAVE_KERNEL_HEAP) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+	uint32_t uheap = sched_self()->uheap;	/* User heap pointer */
+#endif
+
 	/* Is there already a stack allocated of a different size?  Because of
 	 * alignment issues, stack_size might erroneously appear to be of a
 	 * different size.  Fortunately, this is not a critical operation.
@@ -188,7 +197,7 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		 * object in the tcb of current running task. uheap is always non-null
 		 * in user threads and null for kernel threads.
 		 */
-		if (!sched_self()->uheap) {
+		if (!uheap) {
 			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_alloc_size);
 		} else
 #endif
@@ -242,8 +251,10 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		size_of_stack = top_of_stack - (uint32_t)tcb->stack_alloc_ptr + 4;
 
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-		/* Adjust stack size after guard_size calculationn */
-		size_of_stack = size_of_stack - CONFIG_MPU_STACK_GUARD_SIZE;
+		if (uheap) {
+			/* Adjust stack size after guard_size calculation */
+			size_of_stack = size_of_stack - CONFIG_MPU_STACK_GUARD_SIZE;
+		}
 #endif
 		/* Save the adjusted stack values in the struct tcb_s */
 
@@ -263,9 +274,11 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 #endif
 		board_led_on(LED_STACKCREATED);
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-		/* The smallest size that can be programmed for an MPU region is 32 bytes */
-		mpu_get_register_config_value(&tcb->stack_mpu_regs[0], MPU_REG_NUM_STK,
-			(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
+		if (uheap) {
+			/* The smallest size that can be programmed for an MPU region is 32 bytes */
+			mpu_get_register_config_value(&tcb->stack_mpu_regs[0], MPU_REG_NUM_STK,
+				(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
+		}
 #endif
 
 		return OK;


os/arch/arm/src/common/up_createstack.c
```diff
@@ -263,9 +263,11 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 #endif
 		board_led_on(LED_STACKCREATED);
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-		/* The smallest size that can be programmed for an MPU region is 32 bytes */
-		mpu_get_register_config_value(&tcb->stack_mpu_regs[0], MPU_REG_NUM_STK,
-			(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
+		if (sched_self()->uheap) {
```
sunghan-chang, line 266: You should check line 244 as well.

jeongchanKim (commented):
Is there any reason you make restriction only for user task? I know kernel doesn't have mpu region, but I think it is okay to have stack overflow checking in kernel also.
How do you think about this?

jeongchanKim:
> Is there any reason you make restriction only for user task? I know kernel doesn't have mpu region, but I think it is okay to have stack overflow checking in kernel also.
> How do you think about this?

Sorry, I asked the same question before. I understood why. You can ignore this.

sunghan-chang:
@sangam-swami As @jeongchanKim said, kernel thread could have mpu regions excluding the stack guard of kernel thread. Could you check that?

os/arch/arm/src/common/up_createstack.c
sangam-swami: yes added kernel guard as well and verified on nxp and rtk boards

sangam-swami:
> Is there any reason you make restriction only for user task? I know kernel doesn't have mpu region, but I think it is okay to have stack overflow checking in kernel also.
> How do you think about this?

Yes, I have added for kernel threads as well now. Verified on nxp and rtk, overflow result into hard fault in privileged mode

jeongchanKim (dismissed):
LGTM

os/arch/arm/src/common/up_createstack.c
```diff
@@ -189,11 +189,17 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
...
+			stack_alloc_size = stack_alloc_size + CONFIG_MPU_STACK_GUARD_SIZE;
+			tcb->stack_alloc_ptr = (uint32_t *)kmm_memalign(CONFIG_MPU_STACK_GUARD_SIZE, stack_alloc_size);
+#else
 			tcb->stack_alloc_ptr = (uint32_t *)kmm_malloc(stack_alloc_size);
+#endif
 		} else
 #endif
 		{
-			/* Use the user-space allocator if this is a task or pthread */
+		/* Use the user-space allocator if this is a task or pthread */
```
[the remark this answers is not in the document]
sangam-swami: updated

os/arch/arm/src/common/up_createstack.c
```diff
@@ -189,11 +189,17 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		 * in user threads and null for kernel threads.
 		 */
 		if (!sched_self()->uheap) {
+		/* Use the kernel-space allocator if this is a task or pthread */
```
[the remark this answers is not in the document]
sangam-swami: Updated

sunghan-chang:
@sangam-swami To work on kernel thread, is this enough?
1. Could you confirm whether kernel thread runs on privilege mode or not?
2. If it is on privilege mode, you should use new API instead of the mpu_get_register_config_value function. Because it sets RW for privilege even it gets readonly attribute..

sunghan-chang:
@sangam-swami Could you reply what I asked?

sangam-swami:
> @sangam-swami To work on kernel thread, is this enough?
> 
> 1. Could you confirm whether kernel thread runs on privilege mode or not?
> 2. If it is on privilege mode, you should use new API instead of the mpu_get_register_config_value function. Because it sets RW for privilege even it gets readonly attribute..

Hi @sunghan-chang 
1. We are running kernel thread in privilege mode.
2. I tried to change the mpu setting at readonly to RORO similar to ARMv8, but there was crash at binary loader.
So we have 2 options, 
- To maintain user thread stack protection as of now, and create separate PR for kernel thread support after fixing the crash issue.
OR
- Fix the crash issue at binary loader and change this PR only.

Please suggest your views

sunghan-chang:
1st one.

To maintain user thread stack protection as of now, and create separate PR for kernel thread support after fixing the crash issue

sangam-swami:
> 1st one.
> 
> To maintain user thread stack protection as of now, and create separate PR for kernel thread support after fixing the crash issue

yes modified accordingly

os/arch/arm/src/common/up_createstack.c
```diff
@@ -242,8 +242,10 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		size_of_stack = top_of_stack - (uint32_t)tcb->stack_alloc_ptr + 4;
 
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-		/* Adjust stack size after guard_size calculationn */
-		size_of_stack = size_of_stack - CONFIG_MPU_STACK_GUARD_SIZE;
+		if (sched_self()->uheap) {
+			/* Adjust stack size after guard_size calculationn */
```
jeongchanKim, line 246: Please fix calculationn to calculation.

sangam-swami: done

os/arch/arm/src/common/up_createstack.c
```diff
@@ -263,9 +265,11 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 #endif
 		board_led_on(LED_STACKCREATED);
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-		/* The smallest size that can be programmed for an MPU region is 32 bytes */
-		mpu_get_register_config_value(&tcb->stack_mpu_regs[0], MPU_REG_NUM_STK,
-			(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
+		if (sched_self()->uheap) {
```
sunghan-chang, line 268: You call `sched_self()->uheap` several times. It looks redundant. How about checking it once and using?

sunghan-chang: And could you leave the comment to say why sched_self is used instead of tcb given?
This function creates the stack of new thread so that it should be checked with new thread's heap. But currently it checks parent's heap value.
1. new thread does not have the uheap value yet. (It's added in the task_schedsetup function)
2. parent and child threads must have the same uheap value.

sangam-swami: modified, thanks

os/arch/arm/src/common/up_createstack.c
```diff
@@ -164,6 +164,15 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 {
 	size_t stack_alloc_size = stack_size;
 
+	/* new thread does not have the uheap value yet, It's added in the task_
+	 * schedsetup function.
+	 * Parent and child threads must have the same uheap value, so we are
+	 * using the parent thread uheap
+	 */
+#if defined(HAVE_KERNEL_HEAP) || defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
+	uint32_t uheap = sched_self()->uheap;	/* User heap pointer */
```
sunghan-chang, line 173: How about this for readability?
```c
bool is_user_thread = (sched_self()->uheap != NULL);
```
I don't know above is right syntax or not.

or
```c
bool is_user_thread = (sched_self()->uheap != NULL) ? true : false;
```

sunghan-chang (approved):

PR #4503, opened 2020-09-10
fs/smartfs : Add journal format to signature sector, logical 0

Taejun-Kwon:
If level of crc is same but journal state is different, guide it to lower level format

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -93,6 +93,15 @@
 #define SMART_STATUS_SIZEBITS     0x1C
 #define SMART_STATUS_VERBITS      0x03
 
+#define SMART_JOURNAL_DISABLE     0x01
+#define SMART_JOURNAL_ENABLE      0x02
+
+#ifdef CONFIG_MTD_SMART_JOURNALING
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -124,8 +133,9 @@
 #define SMART_FMT_SIG4            'T'
 
 #define SMART_FMT_VERSION_POS     (SMART_FMT_POS1 + 4)
-#define SMART_FMT_NAMESIZE_POS    (SMART_FMT_POS1 + 5)
-#define SMART_FMT_ROOTDIRS_POS    (SMART_FMT_POS1 + 6)
+#define SMART_FMT_JOURNAL_POS     (SMART_FMT_POS1 + 5)
```
amogh-samsung, line 136: Any particular reason for adding this before NAMESIZE and ROOTDIR signature bytes?
Order of significance?

Taejun-Kwon: Journaling is tightly coupled with crc, so I added here.

amogh-samsung: Fair enough

sangwon03 (approved):
LGTM

PR #4509, opened 2020-09-14
app/smartfs_powercut : Modify powercut test app

Taejun-Kwon:
Added ftruncate function, so modify app to use it.
Now we can test below.
1. Create entry(mkdir, open)
2. Move Entry(rename)
3. Update Contents
4. Append Contents
5. ftruncate(shrink, extend)
6. Delete entry(rmdir, unlink)

### DIFF
--- a/apps/examples/smartfs_powercut/smartfs_powercut_main.c
+++ b/apps/examples/smartfs_powercut/smartfs_powercut_main.c
@@ -69,6 +69,8 @@
 #include <tinyara/config.h>
 
 /* Below paths are for test file which use small size of buffer(same as st_blksize) of stat */
+
+/* TODO More names for text based file */
 #define TEST_SMALL              "/mnt/test_small"
 #define TEST_SMALL_BACKUP       "/mnt/backup_small/test_small"
 #define TEST_SMALL_BACKUP_DIR   "/mnt/backup_small"
@@ -82,12 +84,21 @@
 int TEST_SMALL_BUFSIZE;
 
 /* Write Time for each file */
-#define TEST_WRITECOUNT         30
+#define TEST_WRITECOUNT         10
+
+/* File is not exist, so no verify & recovery needed, we can rollback backup file */
+#define FILE_CHUNK_NO_INDEX     -1
 
 struct buffer_data_s {
 	uint16_t crc;
 	char data[TEST_LARGE_BUFSIZE];
 };
+
+enum file_type_e {
+	FILE_TYPE_RANDOM = 0,
+	FILE_TYPE_TEXT = 1
+};
+typedef enum file_type_e file_type_t;
 
 static struct buffer_data_s g_buf;
 
@@ -99,7 +110,7 @@ void make_random_chars(char *buf, size_t len)
 	buf[len - 1] = '\0';
 }
 
-static int write_file(int fd, size_t bufsize)
+static int write_random_data_file(int fd, size_t bufsize)
 {
 	int ret;
 	int data_len = bufsize - sizeof(g_buf.crc);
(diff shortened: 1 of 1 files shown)


apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -222,6 +249,26 @@ int recovery_file(char *src, char* backup, char* backup_dir, int bufsize, int id
 			return ERROR;
 		}
 
+		/* Extend last file same as buckup file, it's not common usuage but add it to test */
```
amandeepc, line 252: **backup, **usage

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -222,6 +249,26 @@ int recovery_file(char *src, char* backup, char* backup_dir, int bufsize, int id
...
+	
+		ret = ftruncate(fd1, filesize);
+		if (ret != OK) {
+			printf("ftruncate(extend)error filename : %s target size : %d errno : %d\n", src, filesize, errno);
+			goto error_with_fd1;
+		}
+		if (stat(src, &st) == OK) {
+			if (st.st_size != filesize) {
+				printf("size is wrong. filename : %s target size : %d file size : %d\n", src, filesize, st.st_size);
+				return ERROR;
```
jeongchanKim, line 265: Is it okay not to close the fd?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -222,6 +250,28 @@ int recovery_file(char *src, char* backup, char* backup_dir, int bufsize, int id
 			return ERROR;
 		}
 
+		/* Extend last file same as backup file, it's not common usage but add it to test */
```
amogh-samsung, line 253: Maybe better to test extendfile() separately only...

amogh-samsung (commented):
We will add tests for shrink and extend.
With these changes, we only test them in specific cases when a file cannot be recovered completely after a power down.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: Mixing is more complicated case, I modified it to test append too.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -182,6 +183,32 @@ int verify_file(char *path, int bufsize)
...
+		if (ret != OK) {
+			printf("ftruncate(shrink)error filename : %s target size : %d errno : %d\n", path, filesize, errno);
+			close(fd);
+			return ERROR;
+		}
+
+		if (stat(path, &st) == OK) {
+			if (st.st_size != filesize) {
+				printf("size is wrong. filename : %s target size : %d file size : %d\n", path, filesize, st.st_size);
+				return ERROR;
```
jeongchanKim, line 203: I think line 203 and line 208 need close(fd).

Taejun-Kwon: yes. I will fix it thanks.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -193,8 +222,10 @@ int recovery_file(char *src, char* backup, char* backup_dir, int bufsize, int id
 	int fd1;
 	int fd2;
 	int ret;
+	struct stat st;
 	off_t offset;
 	off_t start_offset;
+	int filesize;
 	
 	/* If idx is ERROR, it means it didn't create src file so bring backup file */
 	printf("recovery_file src : %s idx : %d\n", src, idx);
```
amogh-samsung, line 231: Line 232 `if (idx == ERROR || idx == 0 || idx == TEST_WRITECOUNT)`

if (idx == 0) -> File has no valid data, AND file has been truncated to length 0.

I get that we want to test shrink function, but
instead of unlinking the source file and renaming backup file,
We can also append to this shrinked file right?

amogh-samsung: Also,
is recovery_file ever called with `idx == TEST_WRITECOUNT`

Taejun-Kwon: Updated totally. hence resolve ths

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -405,18 +511,18 @@ int smartfs_powercut_main(int argc, char *argv[])
...
+		ret = do_test(TEST_LARGE, TEST_LARGE_BACKUP, TEST_LARGE_BACKUP_DIR, TEST_LARGE_BUFSIZE, FILE_TYPE_RANDOM);
 		if (ret != OK) {
 			printf("test large file failed\n");
-			break;
+			return ret;
 		}
 	}
 	
-	return ret;
+	return OK;
```
amogh-samsung, line 527: Line accessible?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -323,65 +437,57 @@ int init_test_file(void)
...
+	if (ret != OK) {
+		printf("recovery_file failed file : %s\n", src);
+		return ERROR;
+	} 
+
 	printf("do_test create backupdir : %s\n", backupdir);
 	ret = mkdir(backupdir, 0777);
 	if (ret != OK) {
-		printf("Test loop, create backup dir for file %s failed errno : %d\n", src, errno);
+		printf("Test loop, create backup dir for file failed dir : %s errno : %d\n", src, errno);
```
amogh-samsung, line 473: I think we need to print name "backup" instead of "src" for dir name.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -111,94 +122,162 @@ static int write_file(int fd, size_t bufsize)
...
+	int filesize;
 	
-	/* If idx is ERROR, it means it didn't create src file so bring backup file */
 	printf("recovery_file src : %s idx : %d\n", src, idx);
-	if (idx == ERROR || idx == 0 || idx == TEST_WRITECOUNT) {
+	/* If index is FILE_CHUNK_NO_INDEX, There is no tested file, so just rollback backup file
+	 * If index is 0, we should recovery whole of data from back up file, so just rollback it.
+	 * If idnex is same as _TEST_WRITECOUNT, totally written, so no recovery necessary.
+	 */
+	if (idx == FILE_CHUNK_NO_INDEX || idx == 0 || idx == TEST_WRITECOUNT) {
```
amogh-samsung, line 280: in case `idx == TEST_WRITECOUNT`
We unlink existing file which is complete?
Is that necessary??

Why do we call recovery if index = TEST_WRITECOUNT?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -111,94 +122,162 @@ static int write_file(int fd, size_t bufsize)
...
 	if (stat(path, &st) == OK) {
-		printf("path %s already exist size : %d\n", path, st.st_size);
+		printf("File already exist path : %s size : %d\n", path, st.st_size);
 		return OK;
+
 	}
-			
-	fd = open(path, O_CREAT | O_EXCL | O_WROK);
+	
+	fd = open(path, O_CREAT | O_WROK);
```
amogh-samsung, line 146: Exclusivity while writing to src file not needed?

amogh-samsung (commented):
Similar to current version for random char files.
Yet to add separate functions for verifying and recovering text based files right?

amogh-samsung:
What attributes will be different for text based files? Proper text lines instead of random chars?

What extra checks and tests to be performed for text based files?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: Can't understand what you mean exactly, but if all files and directories created properly, or already exist, it returns ok.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: Please make sure that is application's role, it is testing. As possible as it can, it should test many thing.
if idx is same as TEST_WRITECOUNT, it means that power has cut off after write is finished, and CRCs are fine.
In this case, we don't have to recovery.
But make sure that we have to TEST anything , in this case It RENAME backup to src, and filanny src to backup, so we can test 2 time of rename.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: You're right, I will fix it.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: see above line, it checked by stat.

Taejun-Kwon (commented):
> Similar to current version for random char files.
> Yet to add separate functions for verifying and recovering text based files right?

It is pretty present for you.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: I meant is this line of code reachable?
Since the while loop now only exits when one of the tests returns an error

Taejun-Kwon: function type is int, it should return something or warning or build error will happen.

sunghan-chang (approved):

PR #4510, opened 2020-09-14
os/binfmt: Fix crash issue when reload optimization is disabled

kishore-sn:
Commit 1:

    When reload optimization is disabled, a single chunk of memory is
    allocated for holding all elf sections in RAM. In this case, the
    alloc[] array in binary_s structure will contain a single entry for
    the whole binary and any access to bin->alloc[DATA] will be invalid.
    So, we change all such access to bin->datastart which will hold the
    start address of data section.

Commit 2:

    arch/armv7-m: Set MPU permission to RO-RO for read only region
    
    Armv7m allows setting of different permissions for previleged and
    user modes. So, previously we were setting RW(P)-RO(U) permission for
    read only region. But armv8-m does not allow setting different permissions
    based on the mode. So, we are forced to set RO-RO permission for all
    read only regions. This commit changes the permission for armv7m read only
    region to RO-RO, so that the behavior is uniform between armv7m and armv8m.
    This commit also fixes a memfault issue which occurs when RO-RO permission
    is set and Common binary is enabled.

### DIFF
--- a/os/binfmt/binfmt_execmodule.c
+++ b/os/binfmt/binfmt_execmodule.c
@@ -181,7 +181,7 @@ int exec_module(FAR struct binary_s *binp)
 	/* The first 4 bytes of the text section of the application must contain a
 	pointer to the application's mm_heap object. Here we will store the mm_heap
 	pointer to the start of the text section */
-	*(uint32_t *)(binp->alloc[ALLOC_DATA]) = (uint32_t)binp->uheap;
+	*(uint32_t *)(binp->datastart) = (uint32_t)binp->uheap;
 	rtcb = (struct tcb_s *)sched_self();
 	rtcb->uheap = (uint32_t)binp->uheap;
 
@@ -189,11 +189,11 @@ int exec_module(FAR struct binary_s *binp)
 #ifdef CONFIG_ARM_MPU
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
 	/* Configure text section as RO and executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_TXT, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
+	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP_TXT, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
 	/* Configure ro section as RO and non-executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
 #else
 	/* Complete RAM partition will be configured as RW region */
 	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP, (uintptr_t)binp->ramstart, binp->ramsize, false, true);
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -167,25 +167,6 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 			goto errout_with_bin;
 		}
 
-#ifdef CONFIG_SUPPORT_COMMON_BINARY
-		if (bin->islibrary) {
-#if (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU))
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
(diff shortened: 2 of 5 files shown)


os/binfmt/binfmt_execmodule.c
```diff
@@ -181,19 +181,19 @@ int exec_module(FAR struct binary_s *binp)
...
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
 	/* Configure text section as RO and executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_TXT, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
+	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP_TXT, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
 	/* Configure ro section as RO and non-executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
```
sunghan-chang, line 196: Could you let me know why this number change is necessary?

kishore-sn: This change is not related to the current issue. 
Actually, the ordering of the numbers does not matter. But the 0 -> 3 -> 6 order seems neat.

sunghan-chang: We reorder the region by size for size optimization and usually data section is biggest because of heap. Didn't you change that to reduce the time of reordering as current??

os/include/tinyara/binfmt/binfmt.h
```diff
@@ -111,6 +111,7 @@ struct binary_s {
 	uint32_t ramstart;		/* Start address of ram partition */
 	uint32_t ramsize;		/* Size of the RAM paritition */
 	uint32_t heapstart;		/* Start address of app heap area */
+	uint32_t datastart;		/* Start address of data section */
```
jeongchanKim, line 114: In this case, do we still need `alloc[ALLOC_DATA]`? Because we always keep the datastart address in binary_s.

os/binfmt/binfmt_execmodule.c
kishore-sn: You are talking about the order in which the sections are allocated. That is different from the order in which the mpu regions are configured. The mpu regions can be configured in any order. It does not matter. It only matters if there are overlapping regions. Since, we dont have overlaps, its ok for us. In fact, the above change has nothing to do with any of the ordering. It just looks neat and makes more sense to set the values in numerical order, like 0 - 3- 6.. This is how I had implemented initially. But its got modified during some issue fix. So, I just corrected the order now

os/include/tinyara/binfmt/binfmt.h
kishore-sn: The alloc array is used to hold pointers to allocated sections. At the time of unload / exit, all the pointers kept in alloc array are freed. When reload opt is disabled, we allocate only a single region. So, we can keep only the ALLOC_TEXT in alloc array. But still we need a pointer to the start of data section. So, we use the datastart variable. 

But, we still need ALLOC_DATA when reload opt is enabled. Because, in that case, we actually allocate a separate data section.

os/binfmt/binfmt_execmodule.c
sunghan-chang: ok

os/include/tinyara/binfmt/binfmt.h
jeongchanKim: What is the difference between `datastart` and `ALLOC_DATA`?

kishore-sn: All the allocated memory regions for the elf are stored in alloc array. And ALLOC_DATA is a an index for holding the pointer for data section in this array. But ALLOC_DATA is valid only if reload opt is enabled and we allocate the data section separately.
But, datastart variable always points to start of data section. Irrespective of whether data section is separately allocated or not.

kishore-sn:
Added a new commit related to mpu settings as follows:

    arch/armv7-m: Set MPU permission to RO-RO for read only region
    
    Armv7m allows setting of different permissions for previleged and
    user modes. So, previously we were setting RW(P)-RO(U) permission for
    read only region. But armv8-m does not allow setting different permissions
    based on the mode. So, we are forced to set RO-RO permission for all
    read only regions. This commit changes the permission for armv7m read only
    region to RO-RO, so that the behavior is uniform between armv7m and armv8m.
    This commit also fixes a memfault issue which occurs when RO-RO permission
    is set and Common binary is enabled.

os/binfmt/binfmt_loadbinary.c
```diff
@@ -203,6 +184,22 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 #ifdef CONFIG_SAVE_BIN_SECTION_ADDR
 		elf_save_bin_section_addr(bin);
 #endif
+
+#if (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU))
```
sunghan-chang, line 188: Let's split the commit. Could you add why this should be moved in description?

sunghan-chang (commented):
Could you let me know this PR is verified on the board?

os/binfmt/binfmt_loadbinary.c
kishore-sn: Done.

kishore-sn:
> Could you let me know this PR is verified on the board?

Yes. It has been tested on the board.

sunghan-chang (approved):

PR #4511, opened 2020-09-14
Modify stack coloration bug

jeongchanKim:


os/arch/arm/src/common/up_createstack.c
```diff
@@ -256,7 +256,7 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		 */
 
 #ifdef CONFIG_STACK_COLORATION
-		up_stack_color(tcb->stack_alloc_ptr, tcb->adj_stack_size);
+		up_stack_color(tcb->stack_alloc_ptr, tcb->adj_stack_size + CONFIG_MPU_STACK_GUARD_SIZE);
```
sunghan-chang, line 263: Could you add the comment to say description of this code?

os/arch/arm/src/common/up_checkstack.c
```diff
@@ -178,12 +178,12 @@ static size_t do_stackcheck(uintptr_t alloc, size_t size)
 
 size_t up_check_tcbstack(FAR struct tcb_s *tcb)
 {
-	return do_stackcheck((uintptr_t)tcb->stack_alloc_ptr, tcb->adj_stack_size);
+	return do_stackcheck((uintptr_t)tcb->stack_alloc_ptr, tcb->adj_stack_size + CONFIG_MPU_STACK_GUARD_SIZE);
```
sunghan-chang: Because the purpose of this function is returning the size of remain stack, this should be as below
```
return do_stackcheck((uintptr_t)tcb->stack_alloc_ptr + CONFIG_MPU_STACK_GUARD_SIZE, tcb->adj_stack_size);
```

os/binfmt/binfmt_execmodule.c
```diff
@@ -217,15 +217,21 @@ int exec_module(FAR struct binary_s *binp)
...
 	ARCH_GET_RET_ADDRESS
-	stack = (FAR uint32_t *)mm_malloc(binp->uheap, binp->stacksize, retaddr);
+	stack = (FAR uint32_t *)mm_malloc(binp->uheap, stack_alloc_size, retaddr);
 #else
-	stack = (FAR uint32_t *)mm_malloc(binp->uheap, binp->stacksize);
+	stack = (FAR uint32_t *)mm_malloc(binp->uheap, stack_alloc_size);
 #endif
 #else
-	stack = (FAR uint32_t *)kumm_malloc(binp->stacksize);
+	stack = (FAR uint32_t *)kumm_malloc(stack_alloc_size);
```
sunghan-chang, line 234: When CONFIG_MPU_STACK_OVERFLOW_PROTECTION is enabled, we should use memalign function.
How about use up_createstack instead of using this separate allocation and calling up_use_stack?

os/arch/arm/src/common/up_createstack.c
```diff
@@ -256,7 +256,11 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 		 */
 
 #ifdef CONFIG_STACK_COLORATION
-		up_stack_color(tcb->stack_alloc_ptr, tcb->adj_stack_size);
+		/* If CONFIG_MPU_STACK_OVERFLOW_PROTECTION is enabled, stack is created with requested size + CONFIG_MPU_STACK_GUARD_SIZE.
+		 * In this case, we should make coloration from stack_alloc_ptr to adj_stack_ptr.
+		 * So we should re-calculate the coloration size with CONFIG_MPU_STACK_GUARD_SIZE.
+		 */
+		up_stack_color(tcb->stack_alloc_ptr, tcb->adj_stack_size + CONFIG_MPU_STACK_GUARD_SIZE);
```
sangam-swami, line 263: but the stack_alloc_ptr is allocated with CONFIG_MPU_STACK_GUARD_SIZE, why we need to add additional here?

sunghan-chang: To colorize only stack region, start addr should be `stack alloc ptr + stack guard size`, not `stack alloc ptr`. But we have a plan to verify guard region value at context switching to check stack corruption from heap overflow. That's why we color it additionally.

os/binfmt/binfmt_execmodule.c
```diff
@@ -217,24 +217,16 @@ int exec_module(FAR struct binary_s *binp)
...
-	stack = (FAR uint32_t *)mm_malloc(binp->uheap, binp->stacksize, retaddr);
-#else
-	stack = (FAR uint32_t *)mm_malloc(binp->uheap, binp->stacksize);
-#endif
-#else
-	stack = (FAR uint32_t *)kumm_malloc(binp->stacksize);
-#endif
-	if (!stack) {
-		ret = -ENOMEM;
+	ret = up_create_stack(tcb, binp->stacksize, TCB_FLAG_TTYPE_TASK);
```
sunghan-chang, line 220: Please resolve compilation warning.
```
CC:  binfmt_execmodule.c
binfmt_execmodule.c: In function 'exec_module':
binfmt_execmodule.c:220:24: warning: passing argument 1 of 'up_create_stack' from incompatible pointer type [-Wincompatible-pointer-types]
  ret = up_create_stack(tcb, binp->stacksize, TCB_FLAG_TTYPE_TASK);
                        ^~~
In file included from binfmt_execmodule.c:66:0:
/root/tizenrt/os/include/tinyara/arch.h:292:5: note: expected 'struct tcb_s *' but argument is of type 'struct task_tcb_s *'
 int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype);
     ^~~~~~~~~~~~~~~
```

sunghan-chang (approved):

PR #4000, opened 2019-11-04
external/libsodium: add sodium lib

SRCN-LiQin:
Add libsodium for st-device-sdk on tizenrt

PR #4512, opened 2020-09-15
[Armv8] Stack protection with limit register

sangam-swami:


apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -54,13 +54,15 @@ int stack_prot_main(int argc, char *argv[])
 			ptr = &array_normal[STACK_NORMAL_SIZE];
 			for (i = 0; i < STACK_NORMAL_SIZE; i++) {
 				*ptr = 'a';
+				printf(" sangam ptr -----0x%x\n", ptr);
```
kishore-sn, line 57: I think this is not the right way to test stack protection with stack limit register. 
The idea behind stack limit register is that an exception will be triggered if the value of SP register goes below the value of limit register. The sp value will change only based on the variables / array we declare in the api.. or due to the function recursion. In our case, we are defining a array of normal size (10 x 4 bytes) which is well within the stack limit. So, no exception will be triggered in any case when we run this program. 
On the other hand, if you define a array of OVERFLOW size, then the stack limit exception will be triggered as soon as this function is called, even without user selecting any option. This is because, if we create an array of 1000 integers, then compiler generates instructions like below:

00000000 <main>:
   0:	e92d4800 	push	{fp, lr}
   4:	e28db004 	add	fp, sp, #4
   8:	e24ddefa 	sub	sp, sp, #4000	; 0xfa0

At this line stack limit exception is triggered.

sangam-swami: Yes, I am able to get usage fault after the limit value. Modified the test case for it

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,38 @@
...
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 
+void stack_prot_overflow(int *val, char ch)
+{
+	char *ptr;
+	int i = 0;
+	char array_normal[STACK_NORMAL_SIZE];
+
```
kishore-sn, line 40: It is difficult to understand what is being done in below code. Can you please add some comments for each operation?

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,38 @@
...
+	if (ch == 'N' || ch == 'n') {
+		if (*val == ((STACK_OVERFLOW_SIZE / STACK_NORMAL_SIZE) / 2))
+			return;
+	}
+	else {
+		if (*val == (STACK_OVERFLOW_SIZE / STACK_NORMAL_SIZE))
+			return;
+	}
+
+	stack_prot_overflow(val, ch);
```
kishore-sn, line 58: Stack limit checking happens when value of SP goes beyond SPLIM. 
This can happen in 2 cases - recursion and definition of large arrays. 
I think it would be good to test both these scenarios.

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,38 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define	STACK_NORMAL_SIZE		10
+#define	STACK_NORMAL_SIZE		300
```
kishore-sn, line 29: What is the basis for deciding these values? Is it related to some default stack size? Can we comment about it here?

os/arch/arm/src/armv8-m/up_blocktask.c
```diff
@@ -189,6 +189,12 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 #endif
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
 				up_mpu_set_register(rtcb->stack_mpu_regs);
+#endif
+#ifdef CONFIG_REG_STACK_OVERFLOW_PROTECTION
+				static int32_t result;
```
kishore-sn, line 194: A better way to use the limit register is to intialize the limit register in the up_initialstate() API and store / restore the register in the up_exception.S file. Then, we will not need to add this code in multiple files.

os/arch/arm/Kconfig
```diff
@@ -288,6 +288,16 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config REG_STACK_OVERFLOW_PROTECTION
+	bool "REG Stack Overflow Protection"
+	depends on BUILD_PROTECTED
```
kishore-sn, line 293: Why does this option depend on protected build? I think this stack limit check applies to flat build also.

jeongchanKim: The below config `MPU_STACK_OVERFLOW_PROTECTION` also need not have dependency with protected build. Please check this also.

os/arch/arm/Kconfig
```diff
@@ -288,6 +288,16 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+config REG_STACK_OVERFLOW_PROTECTION
```
jeongchanKim, line 291: I think we can choose the stack protection way just one of below configs, REG or MPU.
How about making option for selecting just one?

sunghan-chang:
No update today?

os/arch/arm/Kconfig
sangam-swami: Added condition

os/arch/arm/Kconfig
sangam-swami: removed dependency of reg_stack_overflow_protection on protected build..
for mpu_stack will create other PR

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: Its just predefined size used for test

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: added both situation for test case for overflow condition check

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: added comments

sangam-swami:
> No update today?

updated, please review

os/arch/arm/Kconfig
```diff
@@ -288,6 +288,16 @@ config ARCH_DABORTSTACK
...
+	bool "REG Stack Overflow Protection"
+	depends on !MPU_STACK_OVERFLOW_PROTECTION
+	depends on ARCH_ARMV8M_FAMILY
+	default n
+	---help---
+		Enables stack overflow protection with limit registers.
+		It protects each task's stack overflow outside its max allocated stack memory.
+		This is applicable to user mode tasks created in protected build.
```
sunghan-chang, line 299: Please check this line.

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,51 @@
...
 #define	STACK_OVERFLOW_SIZE		4096
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: stack_prot_memory_usage
+ ****************************************************************************/
+
+void stack_overflow_array(void)
```
sunghan-chang, line 39: Is this valid test in v7m as well?

os/arch/arm/Kconfig
sunghan-chang: Even this is selected, MPU based function is in the list of menuconfig. How about using choice?

os/arch/arm/src/armv8-m/up_blocktask.c
```diff
@@ -189,6 +189,12 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 #endif
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
 				up_mpu_set_register(rtcb->stack_mpu_regs);
+#endif
+#ifdef CONFIG_REG_STACK_OVERFLOW_PROTECTION
```
sunghan-chang, line 193: minor one.
To say these are the same functionality, how about using below?
```
#if defined(CONFIG_MPU_xxxx)

#elif defined(CONFIG_REG_xxx)

#endif 
```

sangam-swami: changed

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: no, there are not used in v7m

os/arch/arm/Kconfig
sangam-swami: modified

os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
```diff
@@ -166,6 +166,12 @@ void up_unblock_task_without_savereg(struct tcb_s *tcb)
 #endif
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
 			up_mpu_set_register(rtcb->stack_mpu_regs);
+#endif
+#ifdef CONFIG_REG_STACK_OVERFLOW_PROTECTION
+			static int32_t result;
+			result = get_PSPLIM();
+			set_PSPLIM((uint32_t) rtcb->stack_alloc_ptr);
+			result = get_PSPLIM();
```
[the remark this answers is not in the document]
sangam-swami: these added during testing, have removed now.
stack_alloc_ptr base only we assign the limit

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,51 @@
...
+	array_normal[0] = 'a';
+	return;
+}
+
+void stack_prot_inlimit(void)
+{
+	char array_normal[STACK_NORMAL_SIZE];
+
+	array_normal[0] = 'a';
+	return;
```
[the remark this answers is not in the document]
sangam-swami: ok

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,51 @@
...
+/****************************************************************************
+ * Name: stack_prot_memory_usage
+ ****************************************************************************/
+
+void stack_overflow_array(void)
+{
+	volatile char array_normal[STACK_OVERFLOW_SIZE];
+
+	array_normal[0] = 'a';
+	return;
```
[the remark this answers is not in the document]
sangam-swami: ok

os/arch/arm/Kconfig
sangam-swami: done

os/arch/arm/Kconfig
```diff
@@ -288,6 +288,17 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+choice
+	prompt "Stack Protection feature"
+
+config REG_STACK_OVERFLOW_PROTECTION
+	bool "REG Stack Overflow Protection"
+	depends on ARCH_ARMV8M_FAMILY
+	default n
```
sunghan-chang, line 297: This is wrong syntax of choice.
Each config under choice should not have default.
Choice should have the default.

sunghan-chang: How to disable stack protection? You miss the way to disable.

sangam-swami: changed to meet the requirement

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -46,29 +83,62 @@ int stack_prot_main(int argc, char *argv[])
 	printf("Stack Overflow Protection testcase!!\n");
 
 	while (1) {
-		printf("\nPress N - Test Stack access within range\n");
-		printf("Press O - Test Stack access in overflow\n");
+		printf("\nPress M - Test Stack protection with MPU\n");
+		printf("Press R - Test Stack protection with CPU register\n");
```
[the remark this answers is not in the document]
sangam-swami: Added condition to display the options to user

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: yes changed

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -46,29 +83,66 @@ int stack_prot_main(int argc, char *argv[])
 	printf("Stack Overflow Protection testcase!!\n");
 
 	while (1) {
-		printf("\nPress N - Test Stack access within range\n");
-		printf("Press O - Test Stack access in overflow\n");
+		printf("\nPress M - Test Stack protection with MPU\n");
+#if defined(CONFIG_ARCH_ARMV8M_FAMILY)
```
sunghan-chang, line 87: I think CONFIG_REG_STACK_OVERFLOW_PROTECTION is right.

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -46,29 +83,66 @@ int stack_prot_main(int argc, char *argv[])
 	printf("Stack Overflow Protection testcase!!\n");
 
 	while (1) {
-		printf("\nPress N - Test Stack access within range\n");
-		printf("Press O - Test Stack access in overflow\n");
+		printf("\nPress M - Test Stack protection with MPU\n");
```
sunghan-chang, line 86: You should cover this with CONFIG_MPU_STACK_OVERFLOW_PROTECTION as well.
This is the same as register. When you enable REG config, this still is in the list. But we can't test MPU way at that time.

apps/examples/stack_protection/stack_protection_main.c
```diff
@@ -26,12 +26,49 @@
...
 #define	STACK_OVERFLOW_SIZE		4096
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: stack_prot_memory_usage
+ ****************************************************************************/
+
+static void stack_overflow_array(void)
```
sunghan-chang, line 39: You should cover those functions with some config. If you don't, compiler will give warnings about defined but not used.

os/arch/arm/Kconfig
```diff
@@ -288,15 +288,34 @@ config ARCH_DABORTSTACK
 		will be used during data abort handling. Recommended data abort stack
 		size is 1K.
 
+
+menu "Stack Protection Feature Support"
+
+config STACK_PROTECTION_SUPPORT
```
sunghan-chang, line 294: I think we had better add new config in choice.
```
choice

config REG

config MPU

config NO

endchoice
```
If we do that like above, all of option can be controlled by one step.
But if we keep current change, we should do 2 steps to turn on.

os/0003-build-configs-rtl8721cms-Enable-stack-protection-wit.patch
```diff
@@ -0,0 +1,93 @@
...
+ #
+ CONFIG_BINARY_MANAGER=y
++CONFIG_BM_PRIORITY_MAX=205
++CONFIG_BM_PRIORITY_MIN=200
+ CONFIG_BINMGR_RECOVERY=y
+ CONFIG_BINMGR_UPDATE=y
+ 
+-- 
+2.7.4
+
```
sunghan-chang, line 93: Please check these patch files.

sangam-swami: updated

os/arch/arm/Kconfig
sangam-swami: done

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: done

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: done

apps/examples/stack_protection/stack_protection_main.c
sangam-swami: changed

sunghan-chang (approved):

jeongchanKim (approved):
Looks good to me.

PR #4514, opened 2020-09-15
build/configs/rtl8721csm: Bug fix for download_tool

AdamZhang0124:
This update is to fix the "Specified filename ttyUSB1 does not exist" bug.

### DIFF
--- a/build/configs/rtl8721csm/rtl8721csm_download.sh
+++ b/build/configs/rtl8721csm/rtl8721csm_download.sh
@@ -19,17 +19,17 @@
 # rtl8721csm_download.sh
 
 if [ -n "$2" ]; then
-	TTYDEV="$2"
+	PORT="$2"
 else
-	TTYDEV="ttyUSB1"
+	PORT="ttyUSB1"
 fi
 
-WARNING="\n Port $TTYDEV is selected\n\n
+WARNING="\n Port $PORT is selected\n\n
 	############################################\n
 	WARNINGS:\n
 	1. Make sure the board is in DOWNLOAD MODE.\n
 	2. Make sure NO other application like putty,\n
-	is occupying $TTYDEV.\n
+	is occupying $PORT.\n
 	############################################\n"
 
 echo -e $WARNING
@@ -44,6 +44,7 @@ OS_PATH=${TOP_PATH}/os
 SMARTFS_BIN_PATH=${BIN_PATH}/rtl8721csm_smartfs.bin
 FLASH_START_ADDR=0x08000000
 
+TTYDEV="/dev/${PORT}"
 CONFIG=${OS_PATH}/.config
 source ${CONFIG}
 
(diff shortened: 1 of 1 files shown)


build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -60,11 +61,6 @@ function rtl8721csm_sanity_check()
 		exit 1
 	fi
 
-	if [ ! -f ${TINYARA_BIN} ]; then
```
sunghan-chang, line 63: Could you let me know why you remove this?

AdamZhang0124: This is because in our platform, the tinyara.bin will not be generated. Checking that file will report error.

sunghan-chang: Oh. TINYARA_BIN definition is missing. The purpose is checking existence of binary which should be downloaded. How about adding the definition as below instead of removing?
```sh
TINYARA_BIN=${BIN_PATH}/km0_km4_image2.bin
```

jeongchanKim (approved):
LGTM

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Sure, but it actually has already checked if the km0_km4_image2.bin is missing in line 261
`cp -p ${BIN_PATH}/km0_km4_image2.bin ${IMG_TOOL_PATH}/km0_km4_image2.bin`
If it is missing, it will report error message:
`cp: cannot stat '/root/tizenrt/build/configs/rtl8721csm/../../output/bin/km0_km4_image2.bin': No such file or directory`

I will still add the change you suggested to keep the consistency with other platforms, thanks!

sunghan-chang: Then it's ok. No need to add. Thank you for reply.

sunghan-chang (approved):

sunghan-chang:
One request.

Please use the commit title as the PR title you used.
Currently commit title is `Bug fix for download_tool`, but pr title is `build/configs/rtl8721csm: Bug fix for download_tool`.
Because the format you used in pr is the rule of commit, not pr and user checks commit history, not pr history.

build/configs/rtl8721csm/rtl8721csm_download.sh
AdamZhang0124: Noted with thanks!

sunghan-chang:
Until now, I modified commit title manually at every PR merge. But it's too redundant.

AdamZhang0124:
> One request.
> 
> Currently commit title is `Bug fix for download_tool`, but pr title is `build/configs/rtl8721csm: Bug fix for download_tool`.
> Please use the commit title as the PR title you used.
> Because the format you used in pr is the rule of commit, not pr and user checks commit history, not pr history.



> One request.
> 
> Currently commit title is `Bug fix for download_tool`, but pr title is `build/configs/rtl8721csm: Bug fix for download_tool`.
> Please use the commit title as the PR title you used.
> Because the format you used in pr is the rule of commit, not pr and user checks commit history, not pr history.

Sure, I will keep them consistent for the future commits. Sorry for the trouble on the previous PRs.

sunghan-chang:
Thank you for understanding.

PR #4515, opened 2020-09-16
fs/ftruncate : set errno as EINVAL if length is negative

Taejun-Kwon:
It is Same as posix, change to set EINVAL as a errno

### DIFF
--- a/os/fs/vfs/fs_truncate.c
+++ b/os/fs/vfs/fs_truncate.c
@@ -180,8 +180,12 @@ int ftruncate(int fd, off_t length)
 	FAR struct file *filep;
 	int ret;
 
-	/* Get the file structure corresponding to the file descriptor. */
+	if (length < 0) {
+		ret = -EINVAL;
+		goto errout;
+	}
 
+	/* Get the file structure corresponding to the file descriptor. */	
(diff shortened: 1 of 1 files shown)


os/fs/vfs/fs_truncate.c
```diff
@@ -180,8 +180,12 @@ int ftruncate(int fd, off_t length)
 	FAR struct file *filep;
 	int ret;
 
-	/* Get the file structure corresponding to the file descriptor. */
+	if (length < 0) {
+		ret = -EINVAL;
```
jeongchanKim, line 184: How about assigning the `ret` with positive errno and calling just `set_errno(ret)` in line 204?

Taejun-Kwon: This is correct, lower level of this will return OK or negative.

jeongchanKim: But this function will return the ERROR, not -ERRORNO. It is just used for setting errno.

Taejun-Kwon: he understood after I explained, hence resolve it.

sunghan-chang (approved):

PR #4517, opened 2020-09-16
os/fs/smartfs: Fix improper start address of buffer passed to append_data

amogh-samsung:
- Pass buffer with 0 offset to smartfs_append_data().
- 'byteswritten' parameter takes care of offset internally.


Taejun-Kwon (commented):
could you please change description? It's not tightly related to pwrite, it changed pointer of buffer when smartfs_write need append..

Taejun-Kwon (approved):

PR #4518, opened 2020-09-16
apps/smartfs_powercut: Modify to add test functions for text files

amogh-samsung:
- Add functions to create, verify and recover text files.

### DIFF
--- a/apps/examples/smartfs_powercut/smartfs_powercut_main.c
+++ b/apps/examples/smartfs_powercut/smartfs_powercut_main.c
@@ -68,39 +68,49 @@
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/config.h>
 
-/* Below paths are for test file which use small size of buffer(same as st_blksize) of stat */
+/* File and Directory paths for testing */
+/* Proper text files */
+#define TEST_SMALL_T            "/mnt/small_text.txt"
+#define TEST_SMALL_T_BACKUP     "/mnt/backup_small_text/small_text.txt"
+#define TEST_LARGE_T            "/mnt/large_text.txt"
+#define TEST_LARGE_T_BACKUP     "/mnt/backup_large_text/large_text.txt"
 
-/* TODO More names for text based file */
-#define TEST_SMALL              "/mnt/test_small"
-#define TEST_SMALL_BACKUP       "/mnt/backup_small/test_small"
-#define TEST_SMALL_BACKUP_DIR   "/mnt/backup_small"
+/* Random generated character files */
+#define TEST_SMALL_R            "/mnt/small_random"
+#define TEST_SMALL_R_BACKUP     "/mnt/backup_small_random/small_random"
+#define TEST_LARGE_R            "/mnt/large_random"
+#define TEST_LARGE_R_BACKUP     "/mnt/backup_large_random/large_random"
 
-#define TEST_LARGE              "/mnt/test_large"
-#define TEST_LARGE_BACKUP       "/mnt/backup_large/test_large"
-#define TEST_LARGE_BACKUP_DIR   "/mnt/backup_large"
+/* Sample Text file for testing, to be flashed to the board by the user */
+#define SAMPLE_TXT_FILE         "/mnt/sample_file.txt"
 
+/* Backup directoires for test files */
+#define TEST_SMALL_R_BACKUP_DIR   "/mnt/backup_small_random"
+#define TEST_LARGE_R_BACKUP_DIR   "/mnt/backup_large_random"
+#define TEST_SMALL_T_BACKUP_DIR   "/mnt/backup_small_text"
+#define TEST_LARGE_T_BACKUP_DIR   "/mnt/backup_large_text"
 
+/* 2 buffer sizes for testing */
 #define TEST_LARGE_BUFSIZE      8192
 int TEST_SMALL_BUFSIZE;
 
-/* Write Time for each file */
+/* No. of times file is written */
 #define TEST_WRITECOUNT         10
 
-/* File is not exist, so no verify & recovery needed, we can rollback backup file */
+/* File does not exist, verification & recovery not needed, we can rollback to backup file */
 #define FILE_CHUNK_NO_INDEX     -1
 
 struct buffer_data_s {
 	uint16_t crc;
 	char data[TEST_LARGE_BUFSIZE];
 };
+static struct buffer_data_s g_buf;
 
 enum file_type_e {
 	FILE_TYPE_RANDOM = 0,
 	FILE_TYPE_TEXT = 1
 };
 typedef enum file_type_e file_type_t;
-
-static struct buffer_data_s g_buf;
 
 void make_random_chars(char *buf, size_t len)
 {
@@ -115,21 +125,56 @@ static int write_random_data_file(int fd, size_t bufsize)
 	int ret;
 	int data_len = bufsize - sizeof(g_buf.crc);
 
-	memset(g_buf.data, 0xff, data_len);
-	make_random_chars(g_buf.data, data_len);
+	for (int i = 0; i < TEST_WRITECOUNT; i++) {
+		memset(g_buf.data, 0xff, data_len);
+		make_random_chars(g_buf.data, data_len);
 
-	g_buf.crc = crc16((const uint8_t *)g_buf.data, data_len);
+		g_buf.crc = crc16((const uint8_t *)g_buf.data, data_len);
 
-	ret = write(fd, (void *)&g_buf, bufsize);
-	if (ret != bufsize) {
-		printf("write_random_data_file error ret : %d errno : %d\n", ret, errno);
-		return ERROR;
+		ret = write(fd, (void *)&g_buf, bufsize);
+		if (ret != bufsize) {
+			printf("write_random_data_file error ret : %d errno : %d\n", ret, errno);
+			return ERROR;
+		}
 	}
 
 	return OK;
 }
 
-/* small file uses buffer size same as stat.st_blksize */
+static int write_text_file(int fd, size_t bufsize)
+{
+	int ret;
+	int fd_sample;
+
+	fd_sample = open(SAMPLE_TXT_FILE, O_RDOK);
+	if (fd_sample < 0) {
+		printf("Unable to open file: %s, errno: %d\n", SAMPLE_TXT_FILE, errno);
+		return ERROR;
+	}
+
+	/* Read text from sample text file and write to text file for testing */
+	for (int i = 0; i < TEST_WRITECOUNT; i++) {
+		ret = read(fd_sample, (void *)&g_buf.data, bufsize);
+		if (ret != bufsize) {
+			printf("Unable to read file %s, ret : %d, errno : %d\n", SAMPLE_TXT_FILE, ret, errno);
+			ret = ERROR;
+			goto error_with_fd_sample;
+		}
+
+		ret = write(fd, (void *)&g_buf.data, bufsize);
+		if (ret != bufsize) {
+			printf("Unable to write to text file for testing, ret : %d, errno : %d\n", ret, errno);
+			ret = ERROR;
+			goto error_with_fd_sample;
+		}
+	}
+
+	ret = OK;
+error_with_fd_sample:
+	close(fd_sample);
+	return ret;
+}
+
 static int create_file(char *path, ssize_t bufsize, file_type_t type)
 {
 	int fd;
@@ -140,29 +185,25 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
 	if (stat(path, &st) == OK) {
 		printf("File already exist path : %s size : %d\n", path, st.st_size);
 		return OK;
-
 	}
 	
 	fd = open(path, O_CREAT | O_WROK);
 	if (fd < 0) {
-		printf("create file %s error errno : %d\n", path, errno);
+		printf("create_file %s error errno : %d\n", path, errno);
 		return ERROR;
 	}
-	
-	for (int i = 0; i < TEST_WRITECOUNT; i++) {
-		/* write data based on type */
-		if (type == FILE_TYPE_RANDOM) {
-			ret = write_random_data_file(fd, bufsize);
-		} else {
-			/* TODO proper text will be added here */
-		}
-		
-		if (ret != OK) {
-			printf("write failed path : %s ret : %d bufsize : %d\n", path, ret, bufsize);
-			goto error_with_fd;
-		}
-	}
 
+	/* write data based on type */
+	if (type == FILE_TYPE_RANDOM) {
+		ret = write_random_data_file(fd, bufsize);
+	} else {
+		ret = write_text_file(fd, bufsize);
+	}
+		
+	if (ret != OK) {
+		printf("write failed path : %s ret : %d bufsize : %d\n", path, ret, bufsize);
+		goto error_with_fd;
+	}
 	ret = OK;
 	
 error_with_fd:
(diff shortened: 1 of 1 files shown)


apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,76 +212,148 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
+			return ERROR;
+		}
+	}
+	return OK;
+}
+
+static int verify_file(char *path, int bufsize, int *index, file_type_t type)
 {
 	int fd;
+	int fd_sample;
```
Taejun-Kwon, line 261: Simply, fd1, fd2 is enough. but choice is yours.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,76 +212,148 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
+		crc = crc16((const uint8_t *)g_buf.data, data_len);
+		if (crc != g_buf.crc) {
+			printf("verify_random_file : crc mismatch, calulcated crc : %d, written crc : %d\n", crc, g_buf.crc);
+			return ERROR;
+		}
+	}
+	return OK;
+}
+
+static int verify_text_file_data(int fd, char *sample_buf, uint16_t sample_crc, int bufsize)
```
Taejun-Kwon, line 237: I don't think these 2 functions are tightly related to its type... very similar.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,76 +212,148 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
 
+	if (type == FILE_TYPE_TEXT) {
+		fd_sample = open(SAMPLE_TXT_FILE, O_RDOK);
+		if (fd_sample < 0) {
+			printf("Unable to open Sample Text File : %s, errno : %d\n", SAMPLE_TXT_FILE, errno);
+			ret = ERROR;
+			goto error_with_fd;
+		}
+
+		tmp_buf = (char *)kmm_malloc(bufsize);
```
Taejun-Kwon, line 289: you can use g_buf again, you do not handle something with 2 buffers...

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,76 +212,148 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
 			*index = FILE_CHUNK_NO_INDEX;
 			return OK;
 		}
 		
-		printf("verify_file %s open error... errno : %d\n", path, errno);
+		printf("Unable to open file : %s for verifying, errno : %d\n", path, errno);
 		return ERROR;
 	}
 
+	if (type == FILE_TYPE_TEXT) {
```
Taejun-Kwon, line 281: When you make function, First you should think about role.
verify file's role is verify each of data chunk and reveal last valid index of chunk.
```
crc_t crc1, crc2;

fd = open(path, O_RDONLY);
for (int i = 0; i < TEST_COUNT; i++) {
	read(fd, g_buf, bufsize);

	if (type == FILE_TYPE_TEXT) {
		crc1 = crc16(buf, bufsize);
		read(fd2, g_buf, bufszie);
		crc2 = crc16(buf, bufsize);
	} else {
		crc1 = crc16(g_buf.data, bufsize - sizeof(g_buf.crc);
		crc2 = g_buf.crc
	}

	if (crc1 != crc2) {
		*index = i;
                 break;
	} else {
		do_something2;
	}
}
```

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -68,40 +68,50 @@
...
+/* 2 buffer sizes for testing */
 #define TEST_LARGE_BUFSIZE      8192
 int TEST_SMALL_BUFSIZE;
 
-/* Write Time for each file */
+/* No. of times file is written */
 #define TEST_WRITECOUNT         10
 
-/* File is not exist, so no verify & recovery needed, we can rollback backup file */
+/* File does not exist, verification & recovery not needed, we can rollback to backup file */
```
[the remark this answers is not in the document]
amogh-samsung: :)

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -311,57 +415,57 @@ static int recovery_random_file(char *src, char* backup, char* backup_dir, int b
 		/* Extend last file, it's not common usage but add it to test ftruncate(extend) */
 		ret = ftruncate(fd1, filesize);
 		if (ret != OK) {
-			printf("ftruncate(extend)error filename : %s target size : %d errno : %d\n", src, filesize, errno);
+			printf("Ftruncate(extend) error, filename : %s, target size : %d, errno : %d\n", src, filesize, errno);
```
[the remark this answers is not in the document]
amogh-samsung: I'll keep this in mind, thankyou

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -394,100 +498,153 @@ static int recovery_random_file(char *src, char* backup, char* backup_dir, int b
...
-	
+
+	/* Small test files use buffer size same as stat.st_blksize */
 	TEST_SMALL_BUFSIZE = st.st_blksize;
-	
-	ret = mkdir(TEST_SMALL_BACKUP_DIR, 0777);
+
+	/* If sample text file of proper size is present, initialize text files for tesing */
+	ret = stat(SAMPLE_TXT_FILE, &st);
+	if (ret == OK && st.st_size >= TEST_LARGE_BUFSIZE) {
```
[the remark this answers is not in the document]
amogh-samsung: The same part of sample file will be copied 'TEST_WRITECOUNT' number of times to text file
so we need only (LARGE_BUFSIZE * 1)

Then I thought even if user already has a file > (LARGE_BUFSIZE*1), they can use it as a sample file,
no need to cut it to exact length before flashing?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -394,100 +498,153 @@ static int recovery_random_file(char *src, char* backup, char* backup_dir, int b
...
-	}
-	/* TODO recovery proper text type..maybe it will memcpy from text. */
-	return OK;
-}
-
 int init_test_file(void)
 {
 	struct stat st;
 	int ret;
+	bool test_txt = false;
```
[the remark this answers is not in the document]
amogh-samsung: Both sizes are being tested, this flag is for testing type of file
Random files are always tested
This acts as a flag to check if text files have to be tested.
because it depends on if user provides sample_file.txt

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -394,100 +498,153 @@ static int recovery_random_file(char *src, char* backup, char* backup_dir, int b
...
 	if (ret != OK) {
-		printf("recovery_file failed file : %s\n", src);
+		printf("Recovery failed for file : %s\n", src);
 		return ERROR;
-	} 
+	}
+	printf("Recovery of file %s completed successfully\n", src);
 
-	printf("do_test create backupdir : %s\n", backupdir);
+	printf("Creating create backupdir : %s\n", backupdir);
```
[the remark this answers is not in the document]
amogh-samsung: Yes please, could be very useful.
Can we keep it for the next commit? -> Add debug function for smartfs_powercut

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -394,100 +498,153 @@ static int recovery_random_file(char *src, char* backup, char* backup_dir, int b
...
 	}
 
-	stat(src, &st);
-	printf("make backup.. origin file : %s size : %d\n", src, st.st_size);
-	ret = rename(src, backup);
-	if (ret != OK) {
-		printf("make back up file failed src : %s backup : %s errno : %d\n", src, backup, errno);
+	if (stat(src, &st) == OK) {
+		printf("Rename original file : %s size : %d to backup file : %s\n", src, st.st_size, backup);
+		ret = rename(src, backup);
```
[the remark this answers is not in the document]
amogh-samsung: You mean
if (stat !- OK)
    return;

ret = rename()
if (ret != OK)
    return

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -129,10 +139,23 @@ static int write_random_data_file(int fd, size_t bufsize)
 	return OK;
 }
 
-/* small file uses buffer size same as stat.st_blksize */
+static int write_text_file(int fd, size_t bufsize)
```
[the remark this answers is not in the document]
amogh-samsung: The caller function <`write_file`> reads data from the source file sample_file.txt into the global buffer `g_buf`

The same <`bufsize`> number of bytes are written <`TEST_WRITECOUNT`> number of times

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -68,40 +68,50 @@
...
-
-
+/* File and Directory paths for testing */
+/* Proper text files */
+#define TEST_SMALL_T            "/mnt/small_text.txt"
+#define TEST_SMALL_T_BACKUP     "/mnt/backup_small_text/small_text.txt"
+#define TEST_LARGE_T            "/mnt/large_text.txt"
+#define TEST_LARGE_T_BACKUP     "/mnt/backup_large_text/large_text.txt"
+
+/* Random generated character files */
```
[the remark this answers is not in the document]
amogh-samsung: Ohkkay,
So no file type suffix, right?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: We'll keep it same if it's not causing problems?
fd1 and fd2 are kept for main file and backup file respectively.

So fd_sample specifically says this is the descriptor for sample text file.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: One calculates CRC on data part of buffer -> Random file function
Other uses CRC calculated on entire buffer of sample_file.txt and entire buffer read from file -> Text file function

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -148,21 +171,39 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
 		printf("create file %s error errno : %d\n", path, errno);
 		return ERROR;
 	}
+
+	if (type == FILE_TYPE_TEXT) {
```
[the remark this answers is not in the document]
amogh-samsung: Ahh yes,
Can do with one descriptor also.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Yes I tried that, but faced an issue

Line 305 `sample_crc = crc16((const uint8_t *)tmp_buf, bufsize);`
I can't calculate crc16 on g_buf it seems because g_buf is a struct with 2 members of different types

So pointer typecasting gave me an error when I used the following: -
`sample_crc = crc16((const uint8_t *)&g_buf, bufsize);`

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -498,30 +655,56 @@ int smartfs_powercut_main(int argc, char *argv[])
...
-		ret = do_test(TEST_LARGE, TEST_LARGE_BACKUP, TEST_LARGE_BACKUP_DIR, TEST_LARGE_BUFSIZE, FILE_TYPE_RANDOM);
+		printf("Do test on %s\n", TEST_LARGE_R);
+		ret = do_test(TEST_LARGE_R, TEST_LARGE_R_BACKUP, TEST_LARGE_R_BACKUP_DIR, TEST_LARGE_BUFSIZE, FILE_TYPE_RANDOM);
 		if (ret != OK) {
-			printf("test large file failed\n");
+			printf("Powercut test on large random file failed\n");
 			return ret;
 		}
+
+		if (test_txt) {
```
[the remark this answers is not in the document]
amogh-samsung: This flag indicates if sample_file.txt was given by user
Which means we will now test text files also.

similar to reply on the other related comment

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -129,10 +139,23 @@ static int write_random_data_file(int fd, size_t bufsize)
...
+		return ERROR;
+	}
+
+	return OK;
+}
+
 static int create_file(char *path, ssize_t bufsize, file_type_t type)
 {
 	int fd;
+	int fd_sample;
```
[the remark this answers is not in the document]
amogh-samsung: Ahh yes,
Can do with one descriptor also.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: Of course I know it but I think there's no reason for divided functions.
please consider comment what I commented in line 281

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: that is very common usage style of fd, vfs also using like that.
But anyway, I don't care about this. you can keep yours.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: yes

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: My comment means that 'file is exist && has completed data'.
Otherwise, it should be created or append data again.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: you can create sample_file.txt in init_test_file. we don't need pre-load file by nxfuse or romfs.
If you want to test only one of this for short test, KConfig is correct. not local variable.
Because of restricted flash, there are many chances that we have to reduce size of contents for another model.
So we'd better create contents file after we realize bufsize & testcount.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
Taejun-Kwon: That is pointer, you can use g_buf.data but length is bufsize.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: sure.

Taejun-Kwon (commented):
If you want divide most of function totally, you can consider below style(like a interface)
```
struct powercut_test_ops_s {
	int (*verify_file)(char *path, int bufsize, int *idx);
	int (*recovery_file)(char *src, char* backup, char* backup_dir, int bufsize, int idx);
	int (*write_file)(int fd, size_t bufsize);
	...
};

static const struct powercut_test_ops_s g_testentry[] = {
	{verify_text, recovery_text ...},
	{verify_random, recovery_random ...}
}

int smartfs_powercut_main(int argc, char *argv[])
{
   ...
   do_test(TEST_SMALL, TEST_SMALL_BACKUP, TEST_SMALL_BACKUP_DIR, TEST_SMALL_BUFSIZE, FILE_TYPE_RANDOM, 0); //for random
   ...
   do_test(...., 1); //for text
   ...
}
   
static int do_test(char *src, char *backup, char *backupdir, int bufsize, uint8_t entry_type)
{
	int idx;
	ret = g_testentry[entry_type].verify_file(src, bufsize, &idx);
	...
	ret = g_testentry[entry_type].recovery_file(src, backup, backupdir, bufsize, idx);
	...
}
```

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
amogh-samsung: The size of sample text file does not depend on TEST_WRITECOUNT
We only need one full buffer worth of data
The same part of file will be copied again and again TEST_WRITECOUNT number of times to create main testing_ file and backup file.

so we only need sample_file.txt of size 8192 bytes

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: I thought we created separate APIs for verify text and random because separation was needed.
However, in accordance with your previous comment related to this, I will make these common now.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: No I just saying concept of test.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
Taejun-Kwon: Please refer to comment in line 517, buffer size has more chance that reduce its size than TEST_WRITECOUNT.
Because our target board always has very small ram size.
(Mediatek board, we have around hundreds Kilo bytes, without any service app now.)
So large size can be defined in KConfig, I think it will be better. But we don't have to do in this time.
Anyway, where is your source data? what is your plan?

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: updated

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
amogh-samsung: Updated with a function to create text data and store it in a file for verification purposes.
Please let me know what you think about it now.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
amogh-samsung: Referring to the same
Creating sample_file.txt during init()

apps/examples/smartfs_powercut/smartfs_powercut_main.c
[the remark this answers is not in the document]
amogh-samsung: Modified, obsolete now.
Inclusion  of Kconfig to chose file type and buffer size may affect this.

amogh-samsung:
Chose test file type and buffer size on the basis of Kconfig options - Pending

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -115,21 +125,56 @@ static int write_random_data_file(int fd, size_t bufsize)
...
+	if (fd_sample < 0) {
+		printf("Unable to open file: %s, errno: %d\n", SAMPLE_TXT_FILE, errno);
+		return ERROR;
+	}
+
+	/* Read text from sample text file and write to text file for testing */
+	for (int i = 0; i < TEST_WRITECOUNT; i++) {
+		ret = read(fd_sample, (void *)&g_buf.data, bufsize);
+		if (ret != bufsize) {
+			printf("Unable to read file %s, ret : %d\n", SAMPLE_TXT_FILE, ret);
```
Taejun-Kwon, line 159: please add errno for here and line 166

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,98 +212,151 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
 	return ret;
 }
 
-static int verify_random_file(char *path, int bufsize, int *index)
+static int create_sample_text_file(void)
+{
+	int ret;
+	int fd_sample;
```
Taejun-Kwon, line 218: If it created already, we can skip it right? or...you want create it every time?
For both of 2 cases, we need extra process.
If we keep created file, then we should check that file is exist and size of file with 'stat' & 'stat.st_size' .
or if we create newly every time, at least O_TRUNC is needed.. or unlink.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,98 +212,151 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
+			crc2 = g_buf.crc;
+		}
+
+		if (crc1 != crc2) {
+			printf("Verify file : %s failed at index %d, crc1 : %d, crc2 : %d\n", path, i, crc1, crc2);
+			break;
 		}
 	}
 
+	close(fd_sample);
```
Taejun-Kwon, line 312: should check type.

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,98 +212,151 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
-			break;
-		} else {
-			/* Maybe this won't be happened but check crc for double check */
-			crc = crc16((const uint8_t *)g_buf.data, data_len);
-			if (crc != g_buf.crc) {
-				printf("verify_file crc mismatch!! path : %s calulcated crc : %d written crc : %d index : %d\n", path, crc, g_buf.crc, i);
+		if (type == FILE_TYPE_TEXT) {
+			ret = read(fd, (void *)g_buf.data, bufsize);
+			if (ret != bufsize) {
+				printf("Unable to read file: %s, ret: %d\n", path, ret);
```
Taejun-Kwon, line 284: Please add information regarding 'i'

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,98 +212,151 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
-	/* shrink to last valid index here, it is necessary to bury invalid data chunk */
+	/* Shrink file to last valid index here, it is necessary to bury invalid data chunk */
 	if ((i > 0) && (i < TEST_WRITECOUNT)) {
 		/* truncate current file to include valid data only */
 		size = i * bufsize;
 		
 		ret = lseek(fd, 0, SEEK_SET);
 		if (ret != OK) {
-			printf("verify_file, seek to 0 failed\n");
+			printf("Verify_file : %s, seek to 0 failed, ret : %d, errno : %d\n", path, ret, errno);
```
Taejun-Kwon, line 321: oops my mistake. lseek returns newpos so if (ret != 0) is correct. (it works now because OK is 0 but it is wrong)

apps/examples/smartfs_powercut/smartfs_powercut_main.c
```diff
@@ -171,98 +212,151 @@ static int create_file(char *path, ssize_t bufsize, file_type_t type)
...
 			*index = FILE_CHUNK_NO_INDEX;
 			return OK;
 		}
 		
-		printf("verify_file %s open error... errno : %d\n", path, errno);
+		printf("Unable to open file : %s for verifying, errno : %d\n", path, errno);
 		return ERROR;
 	}
 
+	data_len = bufsize - sizeof(g_buf.crc);
```
Taejun-Kwon, line 269: now this can be removed, only used in line 302..

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Done

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Updated

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Done

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Done

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: My bad, updated

apps/examples/smartfs_powercut/smartfs_powercut_main.c
amogh-samsung: Updated

Taejun-Kwon (approved):

PR #4519, opened 2020-09-17
mm/mm_sem : Add NULL check of heap for avoiding dereferencing

jeongchanKim:
In mm APIs, heap pointer can be dereferencing sometimes.
To avoid dereferencing, add null checking.

### DIFF
--- a/os/mm/mm_heap/mm_sem.c
+++ b/os/mm/mm_heap/mm_sem.c
@@ -55,7 +55,7 @@
  ****************************************************************************/
 
 #include <tinyara/config.h>
-
+#include <debug.h>
 #include <unistd.h>
 #include <errno.h>
 #include <assert.h>
@@ -65,24 +65,6 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-
-/* Define the following to enable semaphore state monitoring */
-//#define MONITOR_MM_SEMAPHORE 1
-
-#ifdef MONITOR_MM_SEMAPHORE
-#ifdef CONFIG_DEBUG
-#include <debug.h>
-#define msemdbg dbg
-#else
-#define msemdbg printf
-#endif
-#else
-#ifdef CONFIG_CPP_HAVE_VARARGS
-#define msemdbg(...)
-#else
-#define msemdbg (void)
-#endif
-#endif
 
 /****************************************************************************
  * Private Data
@@ -171,7 +153,7 @@ void mm_takesemaphore(FAR struct mm_heap_s *heap)
 	} else {
 		/* Take the semaphore (perhaps waiting) */
 
-		msemdbg("PID=%d taking\n", my_pid);
+		mvdbg("PID=%d taking\n", my_pid);
 		while (sem_wait(&heap->mm_semaphore) != 0) {
 			/* The only case that an error should occur here is if
 			 * the wait was awakened by a signal.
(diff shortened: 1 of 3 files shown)


os/mm/mm_heap/mm_sem.c
```diff
@@ -106,6 +106,11 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
 	 * private data sets.
 	 */
 
+	if (heap == NULL) {
+		msemdbg("Fail to get heap.\n");
```
sunghan-chang, line 110: We should say invalid input. This message could give wrong information.
"Invalid Heap address given, Fail to initialize sem"

os/mm/mm_heap/mm_sem.c
```diff
@@ -125,7 +130,14 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
 
 int mm_trysemaphore(FAR struct mm_heap_s *heap)
 {
-	pid_t my_pid = getpid();
+	pid_t my_pid;
+
+	if (heap == NULL) {
+		msemdbg("Fail to get heap.\n");
+		return;
```
sunghan-chang, line 137: This function has the return type as int. 
```c
return ERROR;
```

os/mm/mm_heap/mm_sem.c
```diff
@@ -106,6 +106,11 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
 	 * private data sets.
 	 */
 
+	if (heap == NULL) {
```
sunghan-chang, line 109: This function is used only from mm_initialize and that also gets heap addr as argu. I think we should check it in mm_initialize, not here.

os/mm/mm_heap/mm_sem.c
```diff
@@ -125,7 +130,14 @@ void mm_seminitialize(FAR struct mm_heap_s *heap)
 
 int mm_trysemaphore(FAR struct mm_heap_s *heap)
 {
-	pid_t my_pid = getpid();
+	pid_t my_pid;
+
+	if (heap == NULL) {
```
sunghan-chang, line 135: This function is called from umm_trysemaphore, kmm_trysemaphore.
The umm function verifies heap first and then call this function so that it does not have any problem.
But the kmm doesn't and it may causes null dereference.
We should keep the same policy for checking null.
If you want to check null here, you should remove heap checking from umm. It's duplicated.

sunghan-chang: Please check others in this point of view.

os/mm/mm_heap/mm_sem.c
```diff
@@ -171,7 +153,7 @@ void mm_takesemaphore(FAR struct mm_heap_s *heap)
 	} else {
 		/* Take the semaphore (perhaps waiting) */
 
-		msemdbg("PID=%d taking\n", my_pid);
+		mdbg("PID=%d taking\n", my_pid);
```
sunghan-chang, line 156: Let's use mvdbg because this is not an error log.

os/mm/mm_heap/mm_sem.c
```diff
@@ -186,7 +168,7 @@ void mm_takesemaphore(FAR struct mm_heap_s *heap)
 		heap->mm_counts_held = 1;
 	}
 
-	msemdbg("Holder=%d count=%d\n", heap->mm_holder, heap->mm_counts_held);
+	mdbg("Holder=%d count=%d\n", heap->mm_holder, heap->mm_counts_held);
```
sunghan-chang, line 171: same as previous and please check others as well.

os/mm/umm_heap/umm_sem.c
```diff
@@ -101,8 +102,8 @@ int umm_trysemaphore(void *address)
 	if (heap) {
 		return mm_trysemaphore(heap);
```
sunghan-chang, line 103: The mm_trysemaphore returns OK or ERROR, not negated errno. Let's check the result of it and if it fails, let's get errno and return negated errno as description says.

sunghan-chang (approved):

PR #4520, opened 2020-09-17
os/mm: Fix memory leak issue due to mm_get_heap

kishore-sn:
We maintain a list of app heap objects in app_heap_list. When an
application is loaded for the first time, a new node is created and
added to this list. However, since the is_active variable in the node
is not set, it leads to a failure when searching the app_heap_list
for a heap corresponding to an address. In such case, the mm_free
operation fails to free the address, resulting in memory leak. The
issue is fixed by setting is_active to true when the app heap is being
added to the list.

### DIFF
--- a/os/binfmt/binfmt_execmodule.c
+++ b/os/binfmt/binfmt_execmodule.c
@@ -205,7 +205,8 @@ int exec_module(FAR struct binary_s *binp)
 
 	tcb = (FAR struct task_tcb_s *)kmm_zalloc(sizeof(struct task_tcb_s));
 	if (!tcb) {
-		return -ENOMEM;
+		ret = -ENOMEM;
+		goto errout_with_appheap;
 	}
 #if defined(CONFIG_ARCH_ADDRENV) && defined(CONFIG_BUILD_KERNEL)
 	/* Instantiate the address environment containing the user heap */
@@ -387,6 +388,10 @@ int exec_module(FAR struct binary_s *binp)
 
 	return (int)pid;
 
+errout_with_appheap:
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	mm_remove_app_heap_list(binp->uheap);
+#endif
 errout_with_tcbinit:
 #ifdef CONFIG_BINARY_MANAGER
 	BIN_ID(binary_idx) = -1;
--- a/os/binfmt/binfmt_exit.c
+++ b/os/binfmt/binfmt_exit.c
@@ -137,13 +137,14 @@ int binfmt_exit(FAR struct binary_s *bin)
 		}
 		address = (FAR void *)sq_next((FAR sq_entry_t *)address);
 	}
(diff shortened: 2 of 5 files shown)


os/mm/mm_heap/mm_getheap.c
```diff
@@ -76,12 +76,13 @@ void mm_add_app_heap_list(struct mm_heap_s *heap, char *app_name)
 		node = dq_next(node);
 	}
 
-	node = (app_heap_s *)kmm_malloc(sizeof(app_heap_s));
+	node = (app_heap_s *)kmm_zalloc(sizeof(app_heap_s));
```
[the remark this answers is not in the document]
kishore-sn: True. This is not required now since I am setting is_active variable. I had modified it initially to check the memory leak issue. I reverted this change now.

sunghan-chang (dismissed):

kishore-sn:
Added new commit. 

    os/mm: Fix memory leak in mm_get_heap during app update
    
    In case of reload optimization being disabled, or in case of
    app update, a new app heap node is created and added to the
    app_heap_list in mm_get_heap.c. At this time, the previous
    app_heap node is not freed. This results in memory leak when
    reload or update app. This commit will fix this issue by
    removing the old app_heap_node from the list when exit from
    application (without reload).

sunghan-chang (dismissed):

os/binfmt/binfmt_exit.c
```diff
@@ -137,13 +137,14 @@ int binfmt_exit(FAR struct binary_s *bin)
 		}
 		address = (FAR void *)sq_next((FAR sq_entry_t *)address);
 	}
-	mm_remove_app_heap_list(bin->uheap);
+	mm_disable_app_heap_list(bin->uheap);
```
jeongarmy, line 140: please add mm_disable/remove_app_heap_list in exec_module when it is failed also.

kishore-sn: Done.

os/binfmt/binfmt_execmodule.c
```diff
@@ -205,7 +205,7 @@ int exec_module(FAR struct binary_s *binp)
 
 	tcb = (FAR struct task_tcb_s *)kmm_zalloc(sizeof(struct task_tcb_s));
 	if (!tcb) {
-		return -ENOMEM;
+		goto errout_with_appheap;
```
jeongchanKim, line 208: In `errout_with_appheap`, it returns `ret` but there is no assigning of `ret`.
How about initializing or assigning the `ret` from the first or here?

kishore-sn: Done.

jeongchanKim (approved):
Looks good to me.

jeongarmy (approved):
It looks good to me

sunghan-chang (approved):

PR #4521, opened 2020-09-18
fs/smartfs : Add boundary check logic FIBMAP

Taejun-Kwon:
Currently we can request FIBMAP with EMTPY SECTOR(0xFFFF) in this case it is greater than
sMap, so error happened. Max size should be smaller than number of sectors in dev


sangam-swami (commented):
looks good

sunghan-chang (approved):

PR #4522, opened 2020-09-18
apps/shell: refactoring the tash_get_cmd_from_history

sunghan-chang:
1. Return the address of command according to history number given instead of giving back to 2nd argu.
2. Rename the variables for readability
3. Add code comments to explain the operation

### DIFF
--- a/apps/shell/tash_main.c
+++ b/apps/shell/tash_main.c
@@ -305,7 +305,7 @@ int tash_execute_cmdline(char *buff)
 #if TASH_MAX_STORE > 0
 	if (buff[0] == '!') {
 		int cmd_number = atoi(buff + 1);
-		tash_get_cmd_from_history(cmd_number, buff);
+		buff = tash_get_cmd_from_history(cmd_number);
 	}
 
 	tash_store_cmd(buff);
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -233,28 +233,36 @@ static int tash_history(int argc, char **args)
 	return 0;
 }
 
-void tash_get_cmd_from_history(int num, char *cmd)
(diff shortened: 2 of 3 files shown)


sunghan-chang:
This is on update.

apps/shell/tash_main.c
```diff
@@ -305,7 +305,7 @@ int tash_execute_cmdline(char *buff)
 #if TASH_MAX_STORE > 0
 	if (buff[0] == '!') {
 		int cmd_number = atoi(buff + 1);
-		tash_get_cmd_from_history(cmd_number, buff);
```
kishore-sn, line 308: This change makes use of the pointer to command in command store, instead of creating a copy of the command. But, in following code, there are several places where contents pointed by "buff" gets modified. So, this would cause modification to the value stored in the command store. Is this ok?

sunghan-chang: The usages of buff is the same as previous. It's used as outgoing pointer. The tash_get_cmd_from_history function returns the buff pointer with parsed command characters. (Only the cmd_number is ingoing)
This commit just changes the way how to get the buff. Previous is using argument, this commit is using return value.
(I don't know I understand what you say properly)

kishore-sn: Previous code is making a copy of the cmd_store. 

	while (cmd_store[head_idx][pos] != 0) {
		cmd[pos] = cmd_store[head_idx][pos];
		pos++;
         }

But, current implementation returns a pointer to the cmd_store. 
	return cmd_store[cmd_idx];

This return pointer can cause modification to the contents of cmd_store.

sunghan-chang: Yes. You're right. Let me check.

sunghan-chang:
This will be re-opened after entire tidyup.

PR #4524, opened 2020-09-18
fs/smartfs : Fix recovery if power failed before write new path's entry

Taejun-Kwon:
1. Because of free node in queue logic, index of queue has changed, so some of node
skipped from checking. It will free at the last time of recovery.
2. If any of entry already checked, it's chained sectors should be skipped next time

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -2284,15 +2284,11 @@ int smartfs_invalidate_old_entry(struct smartfs_mountpt_s *fs, uint16_t logsecto
 		if (firstparent->time > secondparent->time) {
 			ret = smartfs_invalidateentry(fs, secondparent->parentsector, secondparent->parentoffset);
 			if (ret == OK) {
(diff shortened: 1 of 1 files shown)


os/fs/smartfs/smartfs_utils.c
```diff
@@ -2284,15 +2284,11 @@ int smartfs_invalidate_old_entry(struct smartfs_mountpt_s *fs, uint16_t logsecto
 		if (firstparent->time > secondparent->time) {
 			ret = smartfs_invalidateentry(fs, secondparent->parentsector, secondparent->parentoffset);
 			if (ret == OK) {
-				sq_rem((FAR sq_entry_t *)secondparent, entry_queue);
```
amogh-samsung, line 2287: Not adding the last extra node to the queue now?

Taejun-Kwon: It came from smartfs_scan_entry, nothing need to be added here.

sunghan-chang (approved):

PR #4525, opened 2020-09-18
os/fs/smartfs: Fix improper type value for subsequent file sectors

amogh-samsung:
CRC Enabled - Currently, chainheader type is written as SMARTFS_DIRENT_TYPE_FILE
  It should be SMARTFS_SECTOR_TYPE_FILE instead.

CRC Disabled - Chainheader type not written to MTD, write it along with data for new sectors.

### DIFF
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -2180,7 +2180,7 @@ ssize_t smartfs_append_data(FAR struct smartfs_mountpt_s *fs, FAR struct smartfs
 			sf->currsector = SMARTFS_NEXTSECTOR(chainheader);
 			sf->curroffset = sizeof(struct smartfs_chain_header_s);
 			memset(sf->buffer, CONFIG_SMARTFS_ERASEDSTATE, fs->fs_llformat.availbytes);
-			chainheader->type = SMARTFS_DIRENT_TYPE_FILE;
+			chainheader->type = SMARTFS_SECTOR_TYPE_FILE;
 		}
 #else                                                   /* CONFIG_SMARTFS_USE_SECTOR_BUFFER */
 		if (sf->curroffset == fs->fs_llformat.availbytes) {
--- a/os/fs/smartfs/smartfs_utils.c
+++ b/os/fs/smartfs/smartfs_utils.c
@@ -739,7 +739,6 @@ int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_ofile_s *
 		if (sf->entry.datalen > 0 && used_value == 0) {
 			used_value = fs->fs_llformat.availbytes - sizeof(struct smartfs_chain_header_s);
 		}
-
 		set_used_byte_count((uint8_t *)header->used, used_value);
 #else
 		if (SMARTFS_USED(header) == SMARTFS_ERASEDSTATE_16BIT) {
@@ -751,8 +750,12 @@ int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_ofile_s *
 			header->used[0] = (uint8_t)(tmp & 0x00FF);
 			header->used[1] = (uint8_t)(tmp >> 8);
 		}
+
+		if (header->type == CONFIG_SMARTFS_ERASEDSTATE) {
+			header->type = SMARTFS_SECTOR_TYPE_FILE;
+		}
 #endif
-		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
+		smartfs_setbuffer(&readwrite, sf->currsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
 		ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 		if (ret < 0) {
 			fdbg("Error %d writing used bytes for sector %d\n", ret, sf->currsector);


Taejun-Kwon (approved):

os/fs/smartfs/smartfs_utils.c
```diff
@@ -2221,6 +2230,7 @@ ssize_t smartfs_append_data(FAR struct smartfs_mountpt_s *fs, FAR struct smartfs
 
 				sf->currsector = SMARTFS_NEXTSECTOR(chainheader);
 				sf->curroffset = sizeof(struct smartfs_chain_header_s);
+				chainheader->type = SMARTFS_SECTOR_TYPE_FILE;
```
Taejun-Kwon, line 2233: It should be moved to line 2215.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -2136,6 +2136,15 @@ ssize_t smartfs_append_data(FAR struct smartfs_mountpt_s *fs, FAR struct smartfs
 				fdbg("Error writing sector %d data, ret : %d\n", readwrite.logsector, ret);
 				return ret;
 			}
+			/* If this is a newly chained data sector, write the chainheader->type for the sector */
+			if (chainheader != NULL) {
```
Taejun-Kwon, line 2140: Append_data is called only when it extend file, by ftruncate or O_APPEND or something.
Anyway it is only for write.
So type is always same,  this had better moved to above, before call FS_IOCTL(WRITESECT)
If it is already set type, it will skipped by physically, so it is ok.

amogh-samsung: Yes, this is only needed when we allocate a new sector for file write.

But for first sector, type is passed from calling function (ftruncate or smartfs_write), hence I checked if (chainheader != NULL)

I think you mean that chainheader and data should both be written in one FS_IOCTL call,
Did I get that right?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: If I move it to 2215, before FS_IOCTL,
The new type will get written to previous MTD sector, which is not intended.

The initial sector passed from calling function has chaiheader->type set already.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: I think apply type only in sync_internal will be fine.
There we also update number of written bytes.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: please refer to above comment.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: For that, we will have to pass one extra parameter to sync_internal...
Is that necessary?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -752,7 +752,12 @@ int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_ofile_s *
 			header->used[1] = (uint8_t)(tmp >> 8);
 		}
 #endif
-		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
+		if (header->type == CONFIG_SMARTFS_ERASEDSTATE) {
```
Taejun-Kwon, line 755: anyway dynamic header also need to be removed also if it enabled, we can' use new code.
so your code can be merged to #else part.

amogh-samsung: Resolved, thankyou

os/fs/smartfs/smartfs_utils.c
```diff
@@ -739,20 +739,25 @@ int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_ofile_s *
...
 		}
-
 		set_used_byte_count((uint8_t *)header->used, used_value);
+
+		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
 #else
 		if (SMARTFS_USED(header) == SMARTFS_ERASEDSTATE_16BIT) {
-			header->used[0] = (uint8_t)(sf->byteswritten & 0x00FF);
-			header->used[1] = (uint8_t)(sf->byteswritten >> 8);
+			*((uint16_t *)header->used) = sf->byteswritten;
```
Taejun-Kwon, line 747: actually previous is correct, if we consider MIPS....that is modified when our target board is mips, not arm based...so please roll back.. previous can support both of mips and arm based chip

os/fs/smartfs/smartfs_utils.c
```diff
@@ -739,20 +739,25 @@ int smartfs_sync_internal(struct smartfs_mountpt_s *fs, struct smartfs_ofile_s *
 		if (sf->entry.datalen > 0 && used_value == 0) {
 			used_value = fs->fs_llformat.availbytes - sizeof(struct smartfs_chain_header_s);
 		}
-
 		set_used_byte_count((uint8_t *)header->used, used_value);
+
+		smartfs_setbuffer(&readwrite, sf->currsector, offsetof(struct smartfs_chain_header_s, used), sizeof(uint16_t), (uint8_t *)&fs->fs_rwbuffer[offsetof(struct smartfs_chain_header_s, used)]);
```
Taejun-Kwon, line 744: Here also we should check something more...
It is too complicated code now,  actually writing entire of chain header doesn't takes such a long time than write used only, Let's change to write entire of chain header.....simply...

Taejun-Kwon (approved):

PR #4527, opened 2020-09-18
os/arch/common: Add error check for kheap allocate api

kishore-sn:
Kheap is created based on ram regions defined in the configuration.
The configuration provides the physical start address and size of each
region. It is possible that some ram region might contain data and stack
of the kernel. In such case, the kheap is created after the stack. But,
if the ram region is defined such that the data and stack regions extend
outside of the ram region, then it is an error. This commit checks for
such error condition and causes the device to assert on errors.

### DIFF
--- a/os/arch/arm/src/common/up_allocateheap.c
+++ b/os/arch/arm/src/common/up_allocateheap.c
@@ -129,14 +129,16 @@ void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
 	/* Return the kernel heap settings (i.e., the part of the heap region
 	 * that was not dedicated to the user heap).
 	 */
+	void *stack_end = (void *)(g_idle_topstack & ~(0x7));
 
-	*heap_start = (FAR void *)(g_idle_topstack & ~(0x7));
-
-	/* There may be a special scenario where we might configure a different region
-	 * for heap. In such case, if end of bss falls outside of the region address range,
-	 * then we use the whole region for heap.
-	 */
-	if (*heap_start < (void *)KREGION_START || *heap_start > (void *)KREGION_END) {
+	if ((void *)&_sdata <= (void *)KREGION_END && stack_end >= (void *)KREGION_END) {
+		lldbg("ERROR: Failed to allocate kheap for ram region configuration\n");
+		lldbg("Region start = 0x%x region end = 0x%x\n_sdata = 0x%x end of stack = 0x%x\n",
+				KREGION_START, KREGION_END, &_sdata, stack_end);
+		PANIC();
+	} else if (stack_end >= (void *)KREGION_START && stack_end < (void *)KREGION_END) {
+		*heap_start = stack_end;
+	} else {
 		*heap_start = (void *)KREGION_START;
 	}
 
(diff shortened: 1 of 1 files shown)


jeongchanKim (dismissed):
LGTM

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -129,14 +129,16 @@ void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
...
-	*heap_start = (FAR void *)(g_idle_topstack & ~(0x7));
-
-	/* There may be a special scenario where we might configure a different region
-	 * for heap. In such case, if end of bss falls outside of the region address range,
-	 * then we use the whole region for heap.
-	 */
-	if (*heap_start < (void *)KREGION_START || *heap_start > (void *)KREGION_END) {
+	void *stack_end = (void *)(g_idle_topstack & ~(0x7));
+
+	if ((void *)&_sdata >= (void *)KREGION_START && (void *)&_sdata <= (void *)KREGION_END && stack_end >= (void *)KREGION_END) {
```
sunghan-chang, line 134:
```c
if ((void *)KREGION_START < stack_end && (void *)KREGION_END > (void *)&_stext) {
```
I am sorry but I don't know why you check it with _sdata. 
Usually _stext is the start of flash layout and stack_end is the last so that I think above covers all of cases.

kishore-sn: As you mentioned, stext is for flash region. We need to check for ram region. So, I am checking with sdata. 
I am checking if data section starts within current ram region and also the stack ends within the current ram region. If this condition is not satisfied, then it is an error.

sunghan-chang: Oh ok. You're right. We should check it with sdata, not stext.
Then
```
(void *)&_sdata <= (void *)KREGION_END && stack_end >= (void *)KREGION_END
```
Isn't this enough?
```
(void *)&_sdata >= (void *)KREGION_START
```
Is this necessary?
If we add this conditional, we can't check the case entire kregion is used (overlapped).
Layout like below:
`_sdata - KREGION_START - KREGION_END - stack_end`

kishore-sn: Yes. You are right. I modified it.

sunghan-chang (approved):

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -129,14 +129,16 @@ void up_allocate_kheap(FAR void **heap_start, size_t *heap_size)
...
-	 * for heap. In such case, if end of bss falls outside of the region address range,
-	 * then we use the whole region for heap.
-	 */
-	if (*heap_start < (void *)KREGION_START || *heap_start > (void *)KREGION_END) {
+	void *stack_end = (void *)(g_idle_topstack & ~(0x7));
+
+	if ((void *)&_sdata <= (void *)KREGION_END && stack_end >= (void *)KREGION_END) {
+		lldbg("ERROR: Failed to allocate kheap for ram region configuration\n");
+		lldbg("Region start = 0x%x region end = 0x%x\n_sdata = 0x%x end of stack = 0x%x\n",
+				KREGION_START, KREGION_END, &_sdata, stack_end);
```
sunghan-chang, line 137: One minor question, 
If we call the PANIC after lldbg, Is that working well?
I mean above is printed well?

sunghan-chang (approved):

os/arch/arm/src/common/up_allocateheap.c
kishore-sn: I am not sure. Because, when there is panic or fault, we have observed that sometimes, the previous logs are missing or getting corrupted. But, some other times, it is printed well. For example, in rtk, the logs are always printed well during crash. But in other boards, we have seen problems sometimes. 
But, what is the alternate method to get the logs?

sunghan-chang: Yeah, that's not the issue in this PR. I just want to know that lldbg is necessary. If it is corrupted always, we don't need to add that.

PR #4530, opened 2020-09-22
shell: tidyup semicolon and exclamation commands

sunghan-chang:
1. refactoring the tash_get_cmd_from_history 
    1. Return the address of command according to history index given on success, return null on failure
    2. Rename the variables for readability
    3. Add code comments to explain the operation
    4. Add user feedback messages on failure case
    5. Fix the bug on checking index validation
2. tidyup treatement of semicolon 
    1. Add error feedback message on single semicolon
        When only semicolon is input without any command, Linux returns
        syntax error. Let's do the same.
    2. Rename the local variable, has_nextcmd
        It shows there is a next continouse command. When it is set,
        Main loop should check continousely. For this usages, 'has' is
        better to understand the meaning. Let's rename 'is_nextcmd' to 'has_nextcmd'.
3. add combination command with exclamation command support
    Currently TASH supports single an exclamation command like "!2".
    But it does not support combination commands like "!2;!3", "!1test", "test!3" and so on.
    This commit parses the exclamation mark and replaces it to real command.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -240,7 +240,7 @@ static int tash_history(int argc, char **args)
  * @return     The address of command buffer selected on success, null on failure
  */
 
-char *tash_get_cmd_from_history(int hist_idx)
+static char *tash_get_cmd_from_history(int hist_idx)
 {
 	int cmd_idx;
 
@@ -346,6 +346,50 @@ bool tash_search_cmd(char *cmd, int *pos, char status)
 		}
 	}
 	return true;
+}
+
+/**
+ * @brief     treat exclamation command
+ * @details   Find the exclamation command and replace it to real command into buffer
+ * @param[in] buff The pointer of user input string
+ * @return    OK on success, ERROR on failure
+ */
+
+int check_exclam_cmd(char *buff)
+{
+	int hist_idx;
+	int histcmd_size;
+	char *histcmd_ptr;
+	char *exclam_ptr;
+	int ret = OK;
+
+	/* Find the '!' in the input character */
+
+	while ((exclam_ptr = strchr(buff, ASCII_EXCLAM)) != NULL) {
+		/* Get the command from history according to the given index */
+
+		hist_idx = strtol(exclam_ptr + 1, &buff, 10);
+		histcmd_ptr = tash_get_cmd_from_history(hist_idx);
+		if (!histcmd_ptr) {
+			/* No command, Let's finish */
+
+			ret = ERROR;
+			break;
+		} else {
+			histcmd_size = strlen(histcmd_ptr);
+			if (histcmd_size != (int)(buff - exclam_ptr)) {
+				/* "!number" does not have the same size as the command from the history list.
+				 * Let's adjust the location of next string to replace "!number" to real command string.
+				 */
+
+				memmove(exclam_ptr + histcmd_size, buff, strlen(buff));
+			}
+			/* Replace "!number" to real command string in buff */
+
+			strncpy(exclam_ptr, histcmd_ptr, histcmd_size);
+		}
+	}
+	return ret;
 }
 #endif
 
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -276,8 +276,12 @@ static char *tash_get_cmd_from_history(int hist_idx)
 
 void tash_store_cmd(char *cmd)
 {
+	/* Clear temporary buffer */
+
+	cmd_line[0] = ASCII_NUL;
+
 	/* If there is no command, it is not saved. */
-	if (cmd == NULL || cmd[0] == '\0') {
+	if (cmd == NULL || cmd[0] == ASCII_NUL) {
 		return;
 	}
 
@@ -287,6 +291,7 @@ void tash_store_cmd(char *cmd)
 		CMD_INDEX_DOWN(prev);
 
 		if (strncmp(cmd_store[prev], cmd, TASH_LINEBUFLEN) == 0) {
+			cmd_pos = cmd_tail;
 			return;
 		}
 	}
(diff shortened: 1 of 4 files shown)


apps/shell/tash_command.c
```diff
@@ -233,28 +233,49 @@ static int tash_history(int argc, char **args)
...
+ */
+
+char *tash_get_cmd_from_history(int hist_idx, char *buf)
 {
-	if (num < 1 || num >= TASH_MAX_STORE) {
-		return;
+	int cmd_idx;
+
+	if (hist_idx < 1 || hist_idx >= TASH_MAX_STORE) {
+		printf("!%d: invalid event\n", hist_idx);
```
Taejun-Kwon, line 249: hmm because it is shell command, so printf will annoy us.
Could you please change printf to some kind of verbose? it will really annoy us....=(

sunghan-chang: This is the message Linux feedbacks to user with error case.

sunghan-chang: You can find this message with invalid number.

apps/shell/tash_main.c
```diff
@@ -338,11 +415,21 @@ int tash_execute_cmdline(char *buff)
...
 					break;
 
-				case ASCII_LF:
 				case ASCII_SEMICOLON:
+					if (argc) {
+						/* Set the has_nextcmd true to check contiuouse command and execute current as ASCII_LF */
+
+						has_nextcmd = true;
+					} else {
+						/* Fail to treate semicolon without any command */
```
jeongchanKim, line 424: Please change treate to treat

apps/shell/tash_main.c
```diff
@@ -338,11 +415,21 @@ int tash_execute_cmdline(char *buff)
 					argv[argc++] = buff + 1;
 					break;
 
-				case ASCII_LF:
 				case ASCII_SEMICOLON:
+					if (argc) {
+						/* Set the has_nextcmd true to check contiuouse command and execute current as ASCII_LF */
```
jeongchanKim, line 420: Please change contiuouse  to continuous

apps/shell/tash_command.c
```diff
@@ -233,103 +233,233 @@ static int tash_history(int argc, char **args)
...
+
+	while ((exclam_ptr = strchr(buff, ASCII_EXCLAM)) != NULL) {
+		/* Get the command from history according to the given index */
+
+		hist_idx = strtol(exclam_ptr + 1, &buff, 10);
+		histcmd_ptr = tash_get_cmd_from_history(hist_idx, NULL);
+		if (!histcmd_ptr) {
+			/* No command, Let's finish */
+
+			ret = ERROR;
```
Taejun-Kwon, line 445: If histcmd_ptr was valid in previous loop, it still return ERROR. is it acceptable?
If not initialize ret as ERROR, then set as OK in else part will be better, If I am right.

apps/shell/tash_main.c
```diff
@@ -106,11 +106,24 @@ static void tash_remove_char(char *char_pos)
...
+
 	int bufsize = TASH_LINEBUFLEN;
 	int pos = 0;
-	int nbytes = 0;
-	int char_idx = 0;
+	int nbytes;
+	int char_idx;
 	bool is_tab_pressed = false;
+#if TASH_MAX_STORE > 0
+	bool is_esc_pressed = false;
```
Taejun-Kwon, line 121: Why don't you use bit operation for these? if I were you I would do that.
But you can ignore this comment because of different style.

apps/shell/tash_command.c
```diff
@@ -233,34 +233,59 @@ static int tash_history(int argc, char **args)
...
+		printf("!%d: event not found\n", hist_idx);
+		return NULL;
 	}
 
-	while (cmd_store[head_idx][pos] != 0) {
-		cmd[pos] = cmd_store[head_idx][pos];
-		pos++;
+	/* Copy the data found into the buf pointer */
+
+	if (buf) {
```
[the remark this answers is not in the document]
sunghan-chang: This function returns the buffer address and copies the data into 2nd argument if it is given. They have little bit different purpose.

apps/shell/tash_command.c
```diff
@@ -233,34 +233,59 @@ static int tash_history(int argc, char **args)
...
+	/* Return the address of command buffer indexed */
+
+	return cmd_store[cmd_idx];
 }
 
 void tash_store_cmd(char *cmd)
 {
+	/* Clear temporary buffer */
+
+	cmd_line[0] = ASCII_NUL;
```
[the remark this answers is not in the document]
sunghan-chang: In this state, temporary buffer must be cleared. This is intended.

apps/shell/tash_command.c
```diff
@@ -233,103 +233,233 @@ static int tash_history(int argc, char **args)
...
 
-	if (status == ASCII_A) { // up
+	if (UP_KEY_PRESSED) {
+		/* UP key Pressed */
+
+		if (IS_HIST_AT_TOP) {
+			/* Already reached the top of history list (no more command at up direction), let's ignore. */
+
+			return false;
+		} else if (IS_HIST_AT_BOT && !HAS_USER_INPUT) {
```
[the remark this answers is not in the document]
sunghan-chang: This means that before getting the command from history, store the user command which stays in a command line.
For example, 
1. type some command without enter (keep it in a line)
2. push a direction key

After typing some command in a line and then push up key and down key, then you can find the command you typed.

apps/shell/tash_command.c
sunghan-chang: It should return ERROR when it meets a failure regardless of previous loop result.
You can find the same result on Linux.
When history has upto 100 index, let's type `!99;!101`. Then it does not execute `!99` even it is valid.

apps/shell/tash_main.c
sunghan-chang: Because they are stack variables and they are not deliverable, I don't know bit operation is better or not.

apps/shell/tash_command.c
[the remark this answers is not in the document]
Taejun-Kwon: then you can reduce 2nd parameter, as you saying sometimes buf is null, it had better remove it.
- 'a function should do one thing' - It is always good principle to clean the code.

apps/shell/tash_command.c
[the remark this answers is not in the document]
Taejun-Kwon: okie

apps/shell/tash_command.c
Taejun-Kwon: okie understood.

apps/shell/tash_command.c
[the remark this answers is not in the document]
sunghan-chang: Done

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4531, opened 2020-09-23
fs/smartfs : Add verify code for release type

Taejun-Kwon:
Actually it is very less chance that happen it failed, but add it to for powercut test.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -5036,10 +5036,25 @@ static int smart_journal_release_sector(FAR struct smart_struct_s *dev, uint16_t
 	offset += offsetof(struct smart_sect_header_s, status);
 	ret = MTD_WRITE(dev->mtd, offset, 1, &header.status);
 	if (ret != 1) {
-		fdbg("Release Invalid data sector Error ret : %d target sector : %d\n", ret, psector);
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -5036,10 +5036,25 @@ static int smart_journal_release_sector(FAR struct smart_struct_s *dev, uint16_t
 	offset += offsetof(struct smart_sect_header_s, status);
 	ret = MTD_WRITE(dev->mtd, offset, 1, &header.status);
 	if (ret != 1) {
-		fdbg("Release Invalid data sector Error ret : %d target sector : %d\n", ret, psector);
+		fdbg("Release sector Error ret : %d target sector : %d\n", ret, psector);
 		return -EIO;
 	}
 	
+	/* TODO Now verify release here but it just 1 byte, I guess fail never happened as per my experience...
```
amogh-samsung, line 5043: Why the need for re-verification of status bytes after marking it as released?

Taejun-Kwon: Powercut test environment takes 4~5 second until it power down totally, during this time it may process file operation in very low working condition.
So we should check this one byte has written properly or not.

sunghan-chang (approved):

PR #4534, opened 2020-09-24
fs/smartfs : Add logic to check journal contents before recovery

Taejun-Kwon:
Before calculate journal crc, check validation of its contents first.


Taejun-Kwon:
Under power-cut testing.

sunghan-chang (approved):

PR #4535, opened 2020-09-24
arch/amebad: Flash performance improvement

AdamZhang0124:
In amebad_flash.c:
1) Remove time consuming instructions like malloc and memcpy
2) Remove irqsave and alignment check as they are done in flash_api

In flash_api.c:
1) Using 256B programming unit instead of 12B to write large size data

### DIFF
--- a/os/arch/arm/src/amebad/amebad_flash.c
+++ b/os/arch/arm/src/amebad/amebad_flash.c
@@ -117,17 +117,22 @@ static ssize_t amebad_write(FAR struct mtd_dev_s *dev, off_t offset, size_t nbyt
  ************************************************************************************/
 static ssize_t amebad_erase_page(size_t page)
 {
-    flash_t flash;
     uint32_t address;
+    irqstate_t irqs;
+
     if (page > (AMEBAD_START_SECOTR + AMEBAD_NSECTORS)) {
         printf("Invalid page number\n");
         return -EFAULT;
     }
+    /* Disable IRQs while erasing sector */
+    irqs = irqsave();
 
     /* do erase */
     address = page * CONFIG_AMEBAD_FLASH_BLOCK_SIZE;
-    flash_erase_sector(&flash, address); 
+    flash_erase_sector(NULL, address);
 
+    /* Restore IRQs */
+    irqrestore(irqs);
     return OK;
 }
 
@@ -149,30 +154,61 @@ static int amebad_erase(FAR struct mtd_dev_s *dev, off_t startblock, size_t nblo
 
 static ssize_t amebad_flash_write(size_t addr, const void *buf, size_t length)
 {
-    flash_t flash;
     int32_t result = 0;
-    ssize_t ret = 0;
-    result = flash_stream_write(&flash, addr, length, buf);
+    irqstate_t irqs;
+
+    /* Disable IRQs while erasing sector */
+    irqs = irqsave();
+
+    result = flash_stream_write(NULL, addr, length, buf);
+
+    /* Restore IRQs */
+    irqrestore(irqs);
+
     if (result < 0) {
-        ret = -EIO;
+        return -EIO;
     } else {
-        ret = length;
+        return result;
     }
-    return ret;
 }
 
 ssize_t amebad_flash_read(size_t addr, void *buf, size_t length)
 {
-    flash_t flash;
     int32_t result = 0;
     ssize_t ret = 0;
+    irqstate_t irqs;
 
-    result = flash_stream_read(&flash, addr, length, buf);
-    if (result < 0) {
-        ret = -EIO;
-    } else {
-        ret = length;
-    }
+    /* Disable IRQs while erasing sector */
+    irqs = irqsave();
+    if ((addr & 0x3) == 0) {
+            //! if addr is 4 bytes aligned
+            result = flash_stream_read(NULL, addr, length, buf);
+            if (result < 0) {
+                ret = -EIO;
+            } else {
+                ret = result;
+            }
+        } else {
+            //! if addr is not 4 bytes aligned
+            uint32_t offset = addr & 0x3;
+            int8_t *aligned_read_buf = (int8_t *)kmm_malloc(length + offset);
+            if (aligned_read_buf == NULL) {
+                ret = -EPERM;
+            } else {
+                result = flash_stream_read(NULL, (addr & 0xfffffffc), length + offset, (char *)aligned_read_buf);
+                if (result < 0) {
+                    ret = -EIO;
+                } else {
+                    ret = result;
+                }
+                memcpy(buf, aligned_read_buf + offset, length);
+                kmm_free(aligned_read_buf);
+                aligned_read_buf = NULL;
+            }
+        }
+    /* Restore IRQs */
+    irqrestore(irqs);
+
     return ret;
 }
 
@@ -182,6 +218,7 @@ ssize_t amebad_flash_read(size_t addr, void *buf, size_t length)
 static ssize_t amebad_bread(FAR struct mtd_dev_s *dev, off_t startblock, size_t nblocks, FAR uint8_t *buffer)
 {
     ssize_t result;
+
     result = amebad_flash_read(CONFIG_AMEBAD_FLASH_BASE + (startblock << PAGE_SHIFT), buffer, nblocks << PAGE_SHIFT);
     return result < 0 ? result : nblocks;
 }
@@ -194,6 +231,7 @@ static ssize_t amebad_bwrite(FAR struct mtd_dev_s *dev, off_t startblock, size_t
     ssize_t result;
     result = amebad_flash_write( CONFIG_AMEBAD_FLASH_BASE + (startblock << PAGE_SHIFT), buffer, nblocks << PAGE_SHIFT);
     return result < 0 ? result : nblocks;
+
 }
 
 /************************************************************************************
--- a/os/arch/arm/src/amebad/amebad_flash.c
+++ b/os/arch/arm/src/amebad/amebad_flash.c
@@ -119,22 +119,15 @@ static ssize_t amebad_erase_page(size_t page)
 {
     flash_t flash;
     uint32_t address;
-    irqstate_t irqs;
-
     if (page > (AMEBAD_START_SECOTR + AMEBAD_NSECTORS)) {
         printf("Invalid page number\n");
         return -EFAULT;
     }
-    /* Disable IRQs while erasing sector */
-    irqs = irqsave();
 
     /* do erase */
-
     address = page * CONFIG_AMEBAD_FLASH_BLOCK_SIZE;
     flash_erase_sector(&flash, address); 
 
-    /* Restore IRQs */
-    irqrestore(irqs);
     return OK;
 }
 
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -159,35 +152,12 @@ static ssize_t amebad_flash_write(size_t addr, const void *buf, size_t length)
     flash_t flash;
```
Taejun-Kwon, line 152: It seems that removable.

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -159,35 +152,12 @@ static ssize_t amebad_flash_write(size_t addr, const void *buf, size_t length)
...
-    /* Disable IRQs while erasing sector */
-    irqs = irqsave();
-    if ((addr & 0xff) != 0) {
-        //! value should be 256 byte alignment
-        ret = -EFAULT;
-    } else if ((length & 0x3) != 0) {
-        ret = -EINVAL;
+    result = flash_stream_write(&flash, addr, length, buf);
+    if (result < 0) {
+        ret = -EIO;
```
Taejun-Kwon, line 157: ret is not necessary actually, you can just return -EIO or length.
But flash_stream_write should check how many byte being written, and  each of error type(ex. if length is smaller than 0, return -EINVAL)

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -196,41 +166,13 @@ ssize_t amebad_flash_read(size_t addr, void *buf, size_t length)
     flash_t flash;
     int32_t result = 0;
     ssize_t ret = 0;
-    irqstate_t irqs;
```
Taejun-Kwon, line 199: So, regardless aligning, it works for all cases?

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -196,41 +166,13 @@ ssize_t amebad_flash_read(size_t addr, void *buf, size_t length)
...
-                }
-                memcpy(buf, aligned_read_buf + offset, length);
-                kmm_free(aligned_read_buf);
-                aligned_read_buf = NULL;
-            }
-        }
-    /* Restore IRQs */
-    irqrestore(irqs);
 
+    result = flash_stream_read(&flash, addr, length, buf);
```
Taejun-Kwon, line 170: same as above.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
```diff
@@ -358,7 +358,7 @@ int  flash_stream_write(flash_t *obj, u32 address, u32 len, u8 * data)
 	u32 addr_end = (page_cnt == 1) ? (address + len) : (page_begin + 0x100);
 	u32 size = addr_end - addr_begin;
 	u8 *buffer = data;
-	u8 write_data[12];
+	u8 write_data[256];
```
Taejun-Kwon, line 361: hmm here we already know pointer of buffer from above layer, I mean data, and length too.
Can't we use data directly without this extra buf? also it calls each of function based on remain data as below right?
```
size_t written_bytes = 0;
while (written_bytes < len) {
	if (len > 256) {
		FLASH_TxData256B(addr_begin, length, data);
		written_bytes += 256;
	} else if (len > 12) {
		FLASH_TxData12B(addr_begin, length, data);
		written_bytes += 12;
	} else if (len < 4) {
		FLASH_TxData4B(addr_begin, length, data);
		written_bytes += 4;
	} else {
		...
	}
	
	...
}
...
return written_bytes;
```

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -119,22 +119,15 @@ static ssize_t amebad_erase_page(size_t page)
...
     flash_t flash;
     uint32_t address;
-    irqstate_t irqs;
-
     if (page > (AMEBAD_START_SECOTR + AMEBAD_NSECTORS)) {
         printf("Invalid page number\n");
         return -EFAULT;
     }
-    /* Disable IRQs while erasing sector */
-    irqs = irqsave();
```
jeongchanKim, line 129: Is it okay there will be interrupt when erasing or writing to flash?
And I think this does not have big impact on improving performance..

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
```diff
@@ -384,6 +384,16 @@ int  flash_stream_write(flash_t *obj, u32 address, u32 len, u8 * data)
 		}
 
 		addr_begin = (((addr_begin-1) >> 2) + 1) << 2;
+		for(;size >= 256 ;size -= 256){
+			_memcpy(write_data, buffer, 256);
+			FLASH_TxData256B(addr_begin, 256, write_data);
+#ifdef MICRON_N25Q00AA
+			FLASH_ReadFlagStatusReg();
+#endif
+			buffer += 256;
```
sangam-swami, line 393: why do we use only 256 bytes here? can we use macro and replace these hardcoding

os/arch/arm/src/amebad/amebad_flash.c
AdamZhang0124: The **flash_write_lock** function is used to lock CPU when write or erase flash in flash_api.c.
I have test it again, seems adding irq_save does not have significant effect on the speed.
In this case, I will add irq_save back to be safe. Thanks!

os/arch/arm/src/amebad/amebad_flash.c
AdamZhang0124: Sure, that will be modified. Thanks!

os/arch/arm/src/amebad/amebad_flash.c
AdamZhang0124: Yes, I have run the fs_main.c test case and all test cases are passed.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
AdamZhang0124: The 256 byte used here is only to be consistent with **FLASH_TxData256B**.

os/arch/arm/src/amebad/amebad_flash.c
jeongchanKim: Okay. If there is no big difference, I think adding irq_save is better :)

sunghan-chang:
@AdamZhang0124 Could you update this for comments and your final modification?

AdamZhang0124:
> @AdamZhang0124 Could you update this for comments and your final modification?

Hi @sunghan-chang, 
Sorry for the delay, since the flash erase issue affects the writing speed in the recent tests, the update for this pr was not finalized. I will test the recent fixes and update this as soon as possible.
Thanks!

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
```diff
@@ -384,6 +378,16 @@ int  flash_stream_write(flash_t *obj, u32 address, u32 len, u8 * data)
 		}
 
 		addr_begin = (((addr_begin-1) >> 2) + 1) << 2;
+		for(;size >= 256 ;size -= 256){
+			_memcpy(write_data, buffer, 256);
```
Taejun-Kwon, line 382: I still can't understand why memcpy is needed here, can't we use data(3rd parameter) directly as I   commented above?

Taejun-Kwon (commented):
First parameter, I mean fash *obj of flash_stream_write & flash_stream_read, and so on is not necessary.. please check it.

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -156,59 +154,39 @@ static int amebad_erase(FAR struct mtd_dev_s *dev, off_t startblock, size_t nblo
...
-            adpt_buf = NULL;
-        }
-    }
+    result = flash_stream_write(NULL, addr, length, buf);
 
     /* Restore IRQs */
     irqrestore(irqs);
-    return ret;
+
+    if (result < 0) {
```
Taejun-Kwon, line 168:
```
if (result < 0) {
   return -EIO;
}
return result;
```
But flash_stream_write should return error value. for example
```
int  flash_stream_write(flash_t *obj, u32 address, u32 len, u8 * data)
{
    if (data == NULL) {
        return -EINVAL;
    }
    if (addr is_out_of_valid_address) {
        return -EFAULT;
     }
    ...and something more
}
```

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
```diff
@@ -214,7 +211,6 @@ int  flash_read_word(flash_t *obj, u32 address, u32 * data)
 	}else{
```
Taejun-Kwon, line 211: please add blank each side of else

sunghan-chang: Please ignore this comment, you don't need to apply our coding rules for chip vendor codes.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
AdamZhang0124: Hi @Taejun-Kwon , thank you for the advice. I have tried your suggested change for this part using the while loop instead of for loop. It seems that this change does not significantly improve the writing speed. Therefore, we decided to keep this code as it was as much as possible.

Taejun-Kwon: My question is why extra 256bytes, write_data & memcpy are necessary?

AdamZhang0124: Hi @Taejun-Kwon ,
Based on my opinion and some tests I did, there is no difference in using memcpy instead of directly using the data. But since you are keen to know, I have already forwarded the question to its original author to investigate the initial considerations and make sure it can be safely replaced. 
So far I have not received the response, I will update to you once I know the result, thanks!

os/arch/arm/src/amebad/amebad_flash.c
AdamZhang0124: This error return you suggested will be done in amebad_flash.c instead of flash_api.c .
And it will be updated with other changes, thanks!

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/flash_api.c
AdamZhang0124: I have received the response that it can be safely replaced with your suggested way. I have updated this change in the newest commit, thanks!

sunghan-chang (approved):

PR #4545, opened 2020-10-05
kernel/sched/sched_releasetcb : Skip releasing some of task's resourc…

jeongchanKim:
…es if pid is 0

If there is a failure before assinging pid when creating a task/thread, we don't need to release some of task's resources.
Skipped resources : terminated stackinfo, heapinfo tcbinfo, heapinfo groupinfo, timer, pid
If not, IDLE task's resources are spoiled.

### DIFF
--- a/os/kernel/sched/sched_releasetcb.c
+++ b/os/kernel/sched/sched_releasetcb.c
@@ -150,40 +150,39 @@ int sched_releasetcb(FAR struct tcb_s *tcb, uint8_t ttype)
 	int ret = OK;
 
 	if (tcb) {
+		/* Release the some of task's resources if PID was assigned.
+		 * PID zero is reserved for the IDLE task.  The TCB of the IDLE
+		 * task is never release so a value of zero simply means that
+		 * the process ID was never allocated to this TCB.
+		 */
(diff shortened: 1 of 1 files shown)


os/kernel/sched/sched_releasetcb.c
```diff
@@ -150,40 +150,38 @@ int sched_releasetcb(FAR struct tcb_s *tcb, uint8_t ttype)
 	int ret = OK;
 
 	if (tcb) {
+		/* Release the some of task's resources if PID was assigned.
+		 * PID zero is reserved for the IDLE task.  The TCB of the IDLE
+		 * task is never release so a value of zero simply means that
+		 * the process ID was never allocated to this TCB.
+		 */
+		if (tcb->pid) {
```
Taejun-Kwon, line 158: 'tcb->pid != 0' is same as below code but Consider above comment, it is better to make it explicitly.

jeongchanKim: Yes, but I will leave it as it is for code unity. Thanks.

sunghan-chang (approved):

PR #4550, opened 2020-10-07
board/rtl8721csm: Remove dumplicate assignment for kregion values.

sangam-swami:
Remove the dulicate assignment of kregion values.
kregin start and size are assigned and heap allocated at up_allocateheap function.


os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -41,22 +41,7 @@ extern void vPortExitCritical(void);
...
-#define HEAP_BASE  ((uintptr_t)&_ebss+CONFIG_IDLETHREAD_STACKSIZE)
-#define HEAP_LIMIT ((uintptr_t)&__StackLimit)
-#define PSRAM_HEAP_BASE ((uintptr_t)&__psram_bss_end__[0])
-#define PSRAM_HEAP_LIMIT ((uintptr_t)&__PsramStackLimit)
-void os_heap_init(void){
-	kregionx_start[0] = (void *)HEAP_BASE;
-	kregionx_size[0] = (size_t)(HEAP_LIMIT - HEAP_BASE);
-	if(TRUE == psram_dev_config.psram_dev_enable) {
-		kregionx_start[1] = (void *)PSRAM_HEAP_BASE;
-		kregionx_size[1] = (size_t)(PSRAM_HEAP_LIMIT - PSRAM_HEAP_BASE);
```
sunghan-chang, line 54: Could you confirm that `PSRAM_HEAP_BASE` and size are the same as our configurations?
It means there is no data in psram text, data and bss sections.

sangam-swami: Yes, I have verified PARAM_HEAP_BASE and size are same as the one calculated at up_allocateheap.c.
There is no text, data, bss in psram.

build/configs/rtl8721csm/tc/defconfig
```diff
@@ -188,8 +188,8 @@ CONFIG_BOOT_RUNFROMFLASH=y
 #
 # Boot Memory Configuration
 #
-CONFIG_RAM_KREGIONx_START="0x0,0x0,"
-CONFIG_RAM_KREGIONx_SIZE="0x0,0x0,"
+CONFIG_RAM_KREGIONx_START="0x10005000,0x02000000,"
+CONFIG_RAM_KREGIONx_SIZE="466944,3145728,"
```
sunghan-chang, line 192: Actually this is not what I want. 
This change just sets PSRAM as 3MB regardless of TZ usages. When user disables TZ, user should modify here as well.
But as Realtek said, user could not know that.
The concept of kregion is RAM configuration with physical start address and size. Then, internally TizenRT calculates available region.
The `__ebss` is the mark of the end of system use for first kregion. How about adding something like that to get available region? If we don't, we had better use Realtek way.

sunghan-chang (commented):
Please add the reason why you add kregion[1] with 0x0200 0000, 3,145,728 as you added for kregion[0].
```
The BD RAM memory layout in trustzone enable case is shown below
BD_RAM_NS (rwx) : ORIGIN = 0x10005000, LENGTH = 0x10077000 - 0x10005000   /* MAIN RAM NS: 456K */
BD_RAM_S (rwx)  : ORIGIN = 0x10077000, LENGTH = 0x1007B000 - 0x10077000   /* MAIN RAM S: 16K */
BD_RAM_NSC (rwx): ORIGIN = 0x1007B000, LENGTH = 0x1007C000 - 0x1007B000   /* MAIN RAM NSC: 4K */
```

sunghan-chang:
@sangam-swami Any update?

sunghan-chang:
@sangam-swami This is unncessary now. Let's close.

PR #4551, opened 2020-10-07
libc/time: Change malloc to lib_malloc for memory allocation

jeongarmy:
A libc is used by kernel and user spaces.
So it should use lib_malloc for memory allocation instead of malloc.


jeongchanKim (approved):
Looks good to me.

Taejun-Kwon (approved):

PR #4555, opened 2020-10-08
drivers/wireless, se: use kmm APIs

sunghan-chang:
Because se and drivers/wireless folders are parts of kernel binary, they should use kmm APIs like
kmm_malloc, kmm_free, kmm_calloc and etc instead of malloc, free and calloc.
This commits fix them.


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4559, opened 2020-10-12
loadable_apps/binary_update: Close opened file descriptors

jeongarmy:
Close file descriptors opened by function 'open' but not closed.

### DIFF
--- a/loadable_apps/loadable_sample/wifiapp/binary_update.c
+++ b/loadable_apps/loadable_sample/wifiapp/binary_update.c
@@ -157,7 +157,7 @@ static int binary_update_download_binary(binary_update_info_t *binary_info, int 
 	}
 	printf("Download binary %s version %d Done!\n", APP_NAME, new_version);
 
-	return OK;
+	ret = OK;
 errout_with_close_fd2:
 	close(write_fd);
 errout_with_close_fd1:
@@ -166,7 +166,7 @@ errout_with_close_fd1:
 		fail_cnt++;
 	}
 
-	return ERROR;
+	return ret;
 }
 
 static int binary_update_download_new_binary(void)
@@ -202,7 +202,7 @@ static int binary_update_download_new_binary(void)
 	if (filesize <= 0 || bin_info.available_size <= 0 || filesize >= bin_info.available_size) {
 		fail_cnt++;
 		printf("Can't copy file, size %d, available size %d in fs\n", filesize, bin_info.available_size);
-		return ERROR;
+		goto errout_with_close_fd1;
 	}
 
 	/* Set file offset to the start of file */
@@ -210,7 +210,7 @@ static int binary_update_download_new_binary(void)
 	if (ret < 0) {
 		fail_cnt++;
 		printf("Failed to set fileoffset %s: %d\n", filepath, get_errno());
-		return ERROR;
+		goto errout_with_close_fd1;
 	}
 
 	/* Read the binary header. */
@@ -243,7 +243,7 @@ static int binary_update_download_new_binary(void)
 	ret = write(write_fd, (FAR uint8_t *)&header_data, sizeof(binary_header_t));
 	if (ret != sizeof(binary_header_t)) {
 		printf("Failed to write header: %d\n", ret);
-		goto errout_with_close_fd1;
+		goto errout_with_close_fd2;
 	}
 
 	crc_hash = crc32part((uint8_t *)&header_data + CHECKSUM_SIZE, header_data.header_size, crc_hash);
@@ -285,7 +285,7 @@ static int binary_update_download_new_binary(void)
 	}
 	printf("Download binary %s Done!\n", NEW_APP_NAME);
 
-	return OK;
+	ret = OK;
 errout_with_close_fd2:
 	close(write_fd);
 errout_with_close_fd1:
@@ -294,7 +294,7 @@ errout_with_close_fd1:
 		fail_cnt++;
 	}
 
-	return ERROR;
+	return ret;
 }
 
 static void print_binary_info(binary_update_info_t *binary_info)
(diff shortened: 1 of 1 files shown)


loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -166,7 +166,7 @@ static int binary_update_download_binary(binary_update_info_t *binary_info, int
 		fail_cnt++;
 	}
 
-	return ERROR;
+	return ret;
```
sunghan-chang, line 169: This causes mis-operation for failure case. When write_fd is failed to open, ret has a positive value. Then that will not increase the fail_cnt. Please check.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -243,7 +243,7 @@ static int binary_update_download_new_binary(void)
 	ret = write(write_fd, (FAR uint8_t *)&header_data, sizeof(binary_header_t));
 	if (ret != sizeof(binary_header_t)) {
 		printf("Failed to write header: %d\n", ret);
-		goto errout_with_close_fd1;
+		goto errout_with_close_fd2;
```
jeongchanKim, line 246: In this case, `ret` can be greater than 0, but fail. But you return the ret. Is it okay?

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -200,38 +205,38 @@ static int binary_update_download_new_binary(void)
 	/* Get 'micom' binary file size and check it with available size */
 	filesize = lseek(read_fd, 0, SEEK_END);
 	if (filesize <= 0 || bin_info.available_size <= 0 || filesize >= bin_info.available_size) {
-		fail_cnt++;
 		printf("Can't copy file, size %d, available size %d in fs\n", filesize, bin_info.available_size);
-		return ERROR;
+		ret = ERROR;
```
jeongchanKim, line 209: I think we'd better initialize the `ret` as `ERROR` and assign `OK` when success case. Then we can reduce much code.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -243,7 +248,8 @@ static int binary_update_download_new_binary(void)
 	ret = write(write_fd, (FAR uint8_t *)&header_data, sizeof(binary_header_t));
 	if (ret != sizeof(binary_header_t)) {
```
jeongchanKim, line 249: If it tries to write some data, but not sizeof(binary_header_t), some data will be written, but not erased.
Is it okay?

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -148,16 +155,20 @@ static int binary_update_download_binary(binary_update_info_t *binary_info, int
...
 	if (ret != 0) {
 		printf("Failed to lseek %d, errno %d\n", ret, errno);
+		unlink(new_filepath);
+		ret = ERROR;
 		goto errout_with_close_fd2;
 	}
 	ret = write(write_fd, (FAR uint8_t *)&crc_hash, CHECKSUM_SIZE);
 	if (ret != CHECKSUM_SIZE) {
 		printf("Failed to write %d\n", ret);
+		unlink(new_filepath);
```
sunghan-chang, line 165: 1. Don't you need this for ok case?
2. Is it ok to call this before closing the fd?

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -148,25 +156,30 @@ static int binary_update_download_binary(binary_update_info_t *binary_info, int
...
 
-	return OK;
+	ret = OK;
 errout_with_close_fd2:
 	close(write_fd);
 errout_with_close_fd1:
 	close(read_fd);
 	if (ret < 0) {
 		fail_cnt++;
+		if (is_newfile_created) {
```
Taejun-Kwon, line 177:
```
if (ret != OK) {
   fail_cnt++;
   struct stat st; 
   if (stat(new_filepath, &st) == OK) {
       unlink(new_filepath);
   }
}
```

Taejun-Kwon: But, with journaling, exist file is always valid.
If it stopped in the middle of downloading, written data in file is always valid.
so I think extra logic is needed to continue downloading.

jeongarmy: 1. A file creation is verified when OK is returned by binary_manager_get_download_path. so it is enough to set and check a flag without calling stat to check file existing I think.
2. There is no logic for retrying write operation or something like what you said.
That's because retrying would be failed if the first try is failed.

sunghan-chang (dismissed):

loadable_apps/loadable_sample/wifiapp/binary_update.c
Taejun-Kwon: about #2, I can accept but you should use fs very restrictively, because we don't have a file permission, so every application can access that file.
So you need to unlink after stat, not flag. please consider delay of I/O and context switch.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: A 'ret' variable is used for return value of functions many times so I set ERROR when each failures.

loadable_apps/loadable_sample/wifiapp/binary_update.c
Taejun-Kwon: And If file is not exist, unlink will return -ENOENT, so if you want to try unlink for all cases, you can just call unlink without any checking.

sunghan-chang (approved):

Taejun-Kwon (approved):
ok for me.

jeongchanKim (approved):
LGTM

PR #4560, opened 2020-10-12
os: use kmm APIs

sunghan-chang:
Kernel modules should allocate the memory in kernel space for internal usages.
Let's use kmm APIs instead of memory allocation APIs for user.
This commit includes the change in arch, board, crypto, drivers and logm.

### DIFF
--- a/os/arch/xtensa/src/esp32/spi_flash/flash_mmap.c
+++ b/os/arch/xtensa/src/esp32/spi_flash/flash_mmap.c
@@ -145,7 +145,7 @@ esp_err_t IRAM_ATTR spi_flash_mmap(size_t src_addr, size_t size, spi_flash_mmap_
 #ifdef CONFIG_SPIRAM_SUPPORT
 	int *pages = heap_caps_malloc(sizeof(int) * page_count, MALLOC_CAP_INTERNAL);
 #else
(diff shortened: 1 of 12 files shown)


sunghan-chang:
NEED TO VERIFY

os/arch/xtensa/src/esp32/spi_flash/flash_mmap.c
```diff
@@ -145,7 +145,7 @@ esp_err_t IRAM_ATTR spi_flash_mmap(size_t src_addr, size_t size, spi_flash_mmap_
 #ifdef CONFIG_SPIRAM_SUPPORT
 	int *pages = heap_caps_malloc(sizeof(int) * page_count, MALLOC_CAP_INTERNAL);
```
jeongchanKim, line 146: In `heap_caps_malloc`, it still allocates the memory through `malloc_at`. We'd better change this also.

sunghan-chang: Done

jeongchanKim (approved):
Looks good to me.

jeongarmy (approved):
LGTM

Taejun-Kwon (approved):

PR #4565, opened 2020-10-19
os/arch/arm/src/armv8-m : New method for handing TZ Secure context

kishore-sn:
A secure context is created by TZ for each task by calling the context
allocate api. It returns a pointer to the secure context, which needs to be
stored somewhere. During every context switch from one task to next, we need
to store the context of previous task and load context of new task. This is
done by calling the corresponding load and store apis in TZ and passing the
pointer to the secure context of the task.

Previously, we were storing the secure context pointer in register R8 and also
in a global variable called tz_memory. And we had modified the exception handler
for svc exception to perform the store and load of secure context. However, this
implementation is causing several problems when protected build is used, because
in this case svc is also used to switch between user and kernel modes.

In order to overcome these problems, this commit modifies the method of storing the
secure context pointer and calling the load / store secure context API's.
- Secure context pointer will be stored in tcb of each task
- Exception handler of svc will not be modified. It will be same as exception_common
- TZ load / store api for secure context will be called in the arch specified
  code which handles the actual context switch, such as blocktask, unblocktask, etc..
- Fix the EXC_RETURN value for TZ enabled scenarios


jeongchanKim:
Please separate the PR based on platform/os and apps.

os/arch/arm/src/armv8-m/up_initialstate.c
```diff
@@ -114,12 +114,6 @@ void up_initial_state(struct tcb_s *tcb)
 	/* Set the stack limit value */
 
 	xcp->regs[REG_R10] = (uint32_t)tcb->stack_alloc_ptr + 64;
-#endif
-#ifdef CONFIG_ARMV8M_TRUSTZONE
-	/* By default, a task is created without a secure context */
-	xcp->regs[REG_R8] = 0;
-	/* for stack limit register */
-	xcp->regs[REG_R9] = (uint32_t)tcb->stack_alloc_ptr;
```
jeongchanKim, line 122: This is stack limit register setting. Is it okay to remove?

sunghan-chang (commented):
Regardless of this PR, we set MPU and TZ setting in up_xxx files for the case `if (current_regs)`, but we do them in up_svcall for others. For the same setting, do we need to do that at different location?

And how about making a new function for the duplicated?
```
#ifdef CONFIG_TASK_SCHED_HISTORY
			/* Save the task name which will be scheduled */
			save_task_scheduling_status(rtcb);
#endif

#ifdef CONFIG_ARMV8M_MPU
			/* Condition check : Update MPU registers only if this is not a kernel thread. */
			if ((rtcb->flags & TCB_FLAG_TTYPE_MASK) != TCB_FLAG_TTYPE_KERNEL) {
#if defined(CONFIG_APP_BINARY_SEPARATION)
				for (int i = 0; i < 3 * MPU_NUM_REGIONS; i += 3) {
					up_mpu_set_register(&rtcb->mpu_regs[i]);
				}
#endif
#if defined(CONFIG_MPU_STACK_OVERFLOW_PROTECTION)
				up_mpu_set_register(rtcb->stack_mpu_regs);
#elif defined(CONFIG_REG_STACK_OVERFLOW_PROTECTION)
				set_PSPLIM((uint32_t) rtcb->stack_alloc_ptr);
#endif
			}
#endif
			if (g_umm_app_id) {
				*g_umm_app_id = rtcb->app_id;
			}
#endif

#ifdef CONFIG_TASK_MONITOR
			/* Update rtcb active flag for monitoring. */
			rtcb->is_active = true;
#endif

			if (ntcb->tz_context) {
				TZ_LoadContext_S(ntcb->tz_context);
			}
#endif
```

os/arch/arm/src/armv8-m/exc_return.h
```diff
@@ -78,7 +78,11 @@
 
 /* EXC_RETURN_BASE: Bits that are always set in an EXC_RETURN value. */
 
+#ifdef CONFIG_ARMV8M_TRUSTZONE
+#define EXC_RETURN_BASE          0xffffffa0
```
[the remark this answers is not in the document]
kishore-sn: Done

os/arch/arm/src/armv8-m/up_initialstate.c
kishore-sn: ARMv8 has a different stack limit register and we dont have to store this value in R9.

sunghan-chang (approved):

kishore-sn:
> Regardless of this PR, we set MPU and TZ setting in up_xxx files for the case `if (current_regs)`, but we do them in up_svcall for others. For the same setting, do we need to do that at different location?
> 
> And how about making a new function for the duplicated?
> 


I have considered the duplication of code and make a new PR to fix that. As my understanding goes, we are making these settings when the final context switch happens. In case of `if (current_regs)`, the context switch is done in the up_xxx files, but in the "else" case, the context switch actually happens in up_svcall.

jeongchanKim (approved):
Looks good to me.

sunghan-chang:
> I have considered the duplication of code and make a new PR to fix that. As my understanding goes, we are making these settings when the final context switch happens. In case of if (current_regs), the context switch is done in the up_xxx files, but in the "else" case, the context switch actually happens in up_svcall.

You're right. But I think it's the same between setting them at up_xxx before context switching function call and setting them at svcall inside. No?

kishore-sn:
> > I have considered the duplication of code and make a new PR to fix that. As my understanding goes, we are making these settings when the final context switch happens. In case of if (current_regs), the context switch is done in the up_xxx files, but in the "else" case, the context switch actually happens in up_svcall.
> 
> You're right. But I think it's the same between setting them at up_xxx before context switching function call and setting them at svcall inside. No?

Yes. Even I think so. I will consider it in more detail and modify it if required in next PR.

sunghan-chang:
If we don't do that, we had better leave some comment at up_xxx. I just want to prevent user confused.

sunghan-chang:
@kishore-sn So, we can enable TZ config in loadable defconfig with this PR merged. Right?

kishore-sn:
> If we don't do that, we had better leave some comment at up_xxx. I just want to prevent user confused. That's not mandatory.

Sure. I will make a new PR for the code duplication. At the same time, I will take care of this also.

kishore-sn:
> @kishore-sn So, we can enable TZ config in loadable defconfig with this PR merged. Right?

Yes. We can enable TZ.  But we cannot enable or test RDP demo since it is designed for flat build only.

sunghan-chang:
> Yes. We can enable TZ. But we cannot enable or test RDP demo since it is designed for flat build only.

It's ok. Thank you for confirmation.

sunghan-chang:
> Sure. I will make a new PR for the code duplication. At the same time, I will take care of this also

Ok. Thank you.

sunghan-chang (approved):

PR #4569, opened 2020-10-20
arch/arm/src/armv7-m: Add support for clean & invalidate api for a specific region of dcache

sijupunnnoose:
-IMXRT USB Host driver requires to flush and invalid the dcache specific region, added
two api to handle this requirement as currently this is not supported.

This requirement is from NXP to support USB Host driver, and is part of the release from NXP


jeongchanKim (commented):
Please modify the license year to 2020.

sijupunnnoose:
> Please modify the license year to 2020.
Done

sunghan-chang (approved):

PR #4572, opened 2020-10-20
userspace: move including <tinyara/arch.h> to up_usersapce.c from use…

sunghan-chang:
…rspace.h

The up_userspace.c file needs <tinyara/arch.h> for up_signal_handler,
the userspace.h does not.
Let's move including the header from userspace.h to up_userspace.c.
It could also resolves recursive including.


jeongchanKim (approved):
Looks good to me.

sangwon03 (approved):
LGTM

PR #4581, opened 2020-10-27
kernel/preference: Support private preference by task group

jeongarmy:
The values of private preference are managed by task group.
Each 'value' is saved in a file named 'key' and they are located in a directory /mnt/private/<tg_name>.

### DIFF
--- a/os/include/tinyara/sched.h
+++ b/os/include/tinyara/sched.h
@@ -490,6 +490,14 @@ struct task_group_s {
 
 	struct group_shm_s tg_shm;	/* Task shared memory logic                 */
 #endif
+
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+	/* Preference **************************************************************** */
+	/* The values of private preference are managed by task group.
+	 * Each 'value' is saved in a file named 'key' and they are located in a directory /mnt/private/<tg_name>.
+	 */
+	char tg_name[CONFIG_TASK_NAME_SIZE + 1]; /* Group name (with NULL terminator)  */
+#endif
 };
 #endif
 
--- a/os/kernel/group/group_create.c
+++ b/os/kernel/group/group_create.c
@@ -60,6 +60,9 @@
 #include <assert.h>
 #include <errno.h>
 #include <debug.h>
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+#include <string.h>
+#endif
 
 #include <tinyara/kmalloc.h>
 
@@ -321,7 +324,13 @@ int group_initialize(FAR struct task_tcb_s *tcb)
 	irqrestore(flags);
 
 #endif
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+	/* The values of private preference are managed by task group.
+	 * Set a name of task group, tg_name for private preference. */
 
+	strncpy(group->tg_name, tcb->cmn.name, strlen(tcb->cmn.name) + 1);
+
+#endif
 	/* Save the ID of the main task within the group of threads.  This needed
 	 * for things like SIGCHILD.  It ID is also saved in the TCB of the main
 	 * task but is also retained in the group which may persist after the main
--- a/os/kernel/preference/preference_common.c
+++ b/os/kernel/preference/preference_common.c
@@ -20,39 +20,31 @@
  ****************************************************************************/
 #include <tinyara/config.h>
 #include <debug.h>
+#if CONFIG_TASK_NAME_SIZE > 0
 #include <sys/types.h>
 #include <tinyara/sched.h>
 #include <tinyara/preference.h>
 
 #include "sched/sched.h"
+#endif
 
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 int preference_get_private_keypath(const char *key, char **path)
 {
+#if CONFIG_TASK_NAME_SIZE > 0
 	int ret;
-#ifdef CONFIG_APP_BINARY_SEPARATION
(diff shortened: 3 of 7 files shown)


os/include/tinyara/sched.h
```diff
@@ -490,6 +490,14 @@ struct task_group_s {
 
 	struct group_shm_s tg_shm;	/* Task shared memory logic                 */
 #endif
+
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+	/* Preference **************************************************************** */
```
deepaksrma, line 495: can you change it like below:

/**********Preference******************************
* comment
* comment
*/

os/kernel/group/group_create.c
```diff
@@ -321,7 +324,13 @@ int group_initialize(FAR struct task_tcb_s *tcb)
 	irqrestore(flags);
 
 #endif
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+	/* The values of private preference are managed by task group.
+	 * Set a name of task group, tg_name for private preference. */
+
+	strncpy(group->tg_name, tcb->cmn.name, CONFIG_TASK_NAME_SIZE);
```
jeongchanKim, line 331: You'd better use the size with `strlen(tcb->cmn.name) + 1`.

os/kernel/preference/preference_common.c
```diff
@@ -20,44 +20,41 @@
...
-#ifdef CONFIG_APP_BINARY_SEPARATION
 	tcb = this_task();
-	pid = tcb->group->tg_binid;
-	if (pid > 0) {
-		tcb = sched_gettcb(pid);
-		if (tcb == NULL) {
-			prefdbg("Failed to get main task %d\n", pid);
-			return PREFERENCE_OPERATION_FAIL;
-		}
+	if (!tcb || !tcb->group) {
```
jeongchanKim, line 42: I think `tcb` cannot be NULL, so we don't need to check this.

os/kernel/preference/preference_remove.c
```diff
@@ -111,22 +110,19 @@ int preference_remove_all_key(int type, const char *path)
...
+#if CONFIG_TASK_NAME_SIZE > 0
 		tcb = this_task();
-		pid = tcb->group->tg_binid;
-		if (pid > 0) {
-			tcb = sched_gettcb(pid);
-			if (tcb == NULL) {
-				prefdbg("Failed to get main task %d\n", pid);
-				return PREFERENCE_OPERATION_FAIL;
-			}
+		if (!tcb || !tcb->group) {
```
jeongchanKim, line 115: Same as above.

os/kernel/group/group_create.c
jeongchanKim, line 331: Is it okay to use thread name as a preference identifier? What I worried is that thread name can be duplicated.

jeongchanKim, line 331: Is it okay to use thread name as preference identifier? What I worried is thread name can be duplicated.

jeongchanKim (commented):
You'd better change `PREFERENCE_FREE` and `PREFERENCE_ALLOC` to user space API.
Currently, they are trying to allocate to kernel space, if binary separated.

os/kernel/group/group_create.c
jeongarmy: I agree with your concerns. But a name is only identifier to identify task group I think.. Do you have any ideas for this?

os/include/tinyara/sched.h
jeongarmy: Your opinion is right but I did to follow other comments in this file.
/* modules *****************/
/* descriptions
 *
 */

os/kernel/group/group_create.c
jeongchanKim: How about appending the group id or task pid for identifying?

os/kernel/group/group_create.c
```diff
@@ -321,7 +324,13 @@ int group_initialize(FAR struct task_tcb_s *tcb)
 	irqrestore(flags);
 
 #endif
+#if defined(CONFIG_PREFERENCE) && CONFIG_TASK_NAME_SIZE > 0
+	/* The values of private preference are managed by task group.
+	 * Set a name of task group, tg_name for private preference. */
+
+	strncpy(group->tg_name, tcb->cmn.name, CONFIG_TASK_NAME_SIZE + 1);
```
jeongchanKim, line 331: I didn't mean the `CONFIG_TASK_NAME_SIZE + 1`.
I think `strlen(tcb->cmn.name) + 1` is correct, because of security attack.

os/kernel/group/group_create.c
jeongarmy: A group id or pid would be different after reloading or reboot.

os/kernel/preference/preference_write.c
```diff
@@ -29,32 +29,28 @@
 #include <sys/stat.h>
 #include <tinyara/preference.h>
 
-#ifdef CONFIG_APP_BINARY_SEPARATION
+#if CONFIG_TASK_NAME_SIZE > 0
 #include "sched/sched.h"
```
jeongchanKim, line 33: Please include `tinyara/sched.h` for `struct tcb_s`.

jeongarmy: "sched/sched.h" is needed for this_task(). I added <tinyara/sched.h> also.

jeongchanKim: Yes. This is what I mean. Thanks.

jeongchanKim (approved):
Looks good to me.

sunghan-chang (approved):

PR #4583, opened 2020-10-28
libc/network: remove assert in freeaddrinfo()

pillip8282:
It's dangerous to call assert(). furthermore linux doens't make
crash when freeaddrinfo() receive null. so libc prints error log
instead of calling assert()


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4590, opened 2020-11-03
Add TZ RDP test for loadable apps

kishore-sn:
This PR covers following items:
- Fix issues in trustzone context handling
- Add RDP demo in kernel testcase driver for loadable apps scenario
- Fix memfault issue due to improper PC value when returning from svcall after running rdp demo


os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,76 @@
...
+#include <arch/chip/amebad_nsc.h>
+
+/****************************************************************************
+ * Private Function
+ ****************************************************************************/
+static uint32_t no_protection_func(uint32_t data)
+{
+	uint32_t result;
+
+	result = data / 5 + 3;
```
deepaksrma, line 35: although it's good but result = (data / 5) + 3 looks better.

os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,76 @@
...
+	result *= 2;
+	result += 8;
+
+	return result;
+}
+
+
+
```
deepaksrma, line 43: can you remove 2 extra lines from here?

os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,76 @@
...
+	uint32_t rdp_result;
+	uint32_t no_rdp_result;
+
+	/*
+	 * Tasks are not created with a secure context.
+	 * Any task that is going to call secure functions must call up_allocate_secure_context()
+	 * to allocate itself a secure context before it calls any secure function
+	 */
+	up_allocate_secure_context(1024);
+	for (i = 0; i < 1; i++) {
```
deepaksrma, line 59: what's the use of for loop here for a single value of i?

sunghan-chang:
@kishore-sn Please revert the file permission.
```
build/configs/rtl8721csm/loadable_apps/defconfig  100644 → 100755
```

build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -81,8 +81,9 @@ CONFIG_ARCH_ARMV8M_FAMILY=y
 CONFIG_ARCH_FAMILY="armv8-m"
 CONFIG_ARCH_HAVE_CMNVECTOR=y
 CONFIG_ARM_CMNVECTOR=y
-# CONFIG_ARCH_HAVE_FPU is not set
+CONFIG_ARCH_HAVE_FPU=y
```
jeongchanKim, line 84: Why do we need FPU for this test? Is there any specific reason?

sunghan-chang:
@kishore-sn I don't think that we need two duplicate sample for the same. How about modifying existed rdp sample instead of adding in loadable apps?

os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	up_allocate_secure_context(1024);
+	for (i = 0; i < 32; i++) {
+		rdp_result = rdp_protection_entry(i);
+		no_rdp_result = rdp_no_protection_call(no_protection_func, i);
+
+		if (rdp_result != no_rdp_result) {
+			dbg("rdp call fail!\n");
+			dbg("rdp_result = 0x%x, no_rdp_result=0x%x\n", rdp_result, no_rdp_result);
+			goto end;
```
jeongchanKim, line 64: How about using `break` instead of `goto`? Of course, then we should modify the print msg in line 68 also.

os/drivers/testcase/testcase_proto.h
```diff
@@ -28,4 +28,7 @@ int test_sched(int cmd, unsigned long arg);
 int test_signal(int cmd, unsigned long arg);
 #endif
 int test_compress_decompress(int cmd, unsigned long arg);
+#ifdef CONFIG_AMEBAD_TRUSTZONE
```
jeongchanKim, line 31: I think using `CONFIG_ARMV8M_TRUSTZONE` is much better. Because, in the future we can support more boards which have trustzone, not only amebad.

loadable_apps/loadable_sample/wifiapp/wifiapp.c
```diff
@@ -103,6 +109,16 @@ int wifiapp_main(int argc, char **argv)
 		case 'u':
 			binary_update_test();
 			break;
+#endif
+#ifdef CONFIG_AMEBAD_TRUSTZONE
+		case 'Z':
+		case 'z':
+			{
+				int tc_fd = open(TESTCASE_DRVPATH, O_WRONLY);
+				ioctl(tc_fd, TESTIOC_RDP, 0);
```
jeongarmy, line 118: Need to check returned values  from open and ioctl and add logs about result

kishore-sn:
Since there are several issues with Trustzone and RDP, I am closing this PR. 
I will raise a separate PR to fix TZ issues first and later add RDP in a different PR.

os/drivers/testcase/test_rdp.c
kishore-sn: It would become bit difficult to print the success message. Let us keep it like this.

os/drivers/testcase/testcase_proto.h
kishore-sn: rdp api's are defined only for amebad. So I have kept it under this flag. In future, if we add new vendor implementation or a generic TZ verification code, then we can modify this.

PR #4591, opened 2020-11-04
kernel/binary_manager: Set OK value on success in binary_manager_execute_loader

jeongarmy:
Set OK value on success in binary_manager_execute_loader and fix condition for checking result returned from it.


sijupunnnoose (dismissed):
LGTM

sunghan-chang (commented):
Is there any better point compared to previous? I don't know why this is necessary.

thapav (approved):
LGTM

sunghan-chang (approved):

PR #4592, opened 2020-11-05
Fix TZ context and memory fail issues

kishore-sn:
1. Fix tz context free issue
2. Fix tz memory fail issue
Malloc secure memory failed [free heap size: 6040]


sunghan-chang (approved):

jeongarmy (approved):
Looks good to me

PR #4593, opened 2020-11-05
Reserve MPU regions

thapav:
This PR adds support to reserve MPU regions by the vendor. (CONFIG_MPU_RESERVED)
In such a scenario, the TizenRT MPU configurations will be done past the reserved MPU regions.
It adds the following API's to mpuinit.c
mpu_get_region_info(), mpu_set_region_info(), mpu_regs_initialize()

### DIFF
--- a/os/arch/arm/src/common/up_mpuinit.c
+++ b/os/arch/arm/src/common/up_mpuinit.c
@@ -38,6 +38,7 @@
  ****************************************************************************/
 
 #include <tinyara/config.h>
+#include <tinyara/mpu.h>
 
 #include <assert.h>
 
@@ -50,10 +51,95 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+static struct mpu_usages_s g_mpu_region_usages;
+
+#ifdef CONFIG_ARMV7M_MPU
+#define MPU_MAX_NREGION		CONFIG_ARMV7M_MPU_NREGIONS
+#elif defined(CONFIG_ARMV8M_MPU)
+#define MPU_MAX_NREGION		CONFIG_ARMV8M_MPU_NREGIONS
+#endif
 
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
+
+/****************************************************************************
+ * Name: mpu_region_initialize
+ *
+ * Description:
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_region_initialize(struct mpu_usages_s *mpu)
+{
+	uint8_t offset = (struct mpu_usages_s *)mpu->nregion_board_specific;
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+	offset += MPU_NUM_REGIONS;
+	mpu->nregion_common_bin = offset;
+#endif
+
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	offset += MPU_NUM_REGIONS;
+	mpu->nregion_app_bin = offset;
+#endif
+
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+	offset += 1;
+	mpu->nregion_stackovf = offset;
+#endif
+	mpu->max_nregion = offset;
+}
+
+/****************************************************************************
+ * Name: mpu_get_nregion_info
+ *
+ * Description:
+ *   Get the MPU region numbers.
+ *
+ ****************************************************************************/
+
+uint8_t mpu_get_nregion_info(enum mpu_region_usages_e usages)
+{
+	uint8_t nregion;
+	switch (usages) {
+	case MPU_REGION_BOARD_SPECIFIC:
+		nregion = g_mpu_region_usages.nregion_board_specific;
+		break;
+	case MPU_REGION_COMMON_BIN:
+		nregion = g_mpu_region_usages.nregion_common_bin;
+		break;
+	case MPU_REGION_APP_BIN:
+		nregion = g_mpu_region_usages.nregion_app_bin;
+		break;
+	case MPU_REGION_STACKOVF:
+		nregion = g_mpu_region_usages.nregion_stackovf;
+		break;
+	case MPU_REGION_MAX:
+		nregion = g_mpu_region_usages.max_nregion;
+		break;
+	default:
+		nregion = 0;
+		break;
+	}
+
+	return nregion;
+}
+
+/****************************************************************************
+ * Name: mpu_set_nregion_board_specific
+ *
+ * Description:
+ *   Initialize the number of MPU regions for board specific purpose
+ *
+ ****************************************************************************/
+
+void mpu_set_nregion_board_specific(uint8_t num)
+{
+	/* Initialize number of mpu regions for board-specific purpose */
+	g_mpu_region_usages.nregion_board_specific = num;
+}
 
 /****************************************************************************
  * Name: up_mpuinitialize
@@ -66,6 +152,12 @@
 
 void up_mpuinitialize(void)
 {
+	/* Initialize MPU register numbers information */
+	mpu_region_initialize(&g_mpu_region_usages);
+
+	/* Maximum MPU region number sanity check */
+	DEBUGASSERT(g_mpu_region_usages.max_nregion < MPU_MAX_NREGION);
+
 	/* Show MPU information */
 	mpu_showtype();
 
--- a/os/binfmt/binfmt_execmodule.c
+++ b/os/binfmt/binfmt_execmodule.c
@@ -188,15 +188,17 @@ int exec_module(FAR struct binary_s *binp)
 	/* Initialize the MPU registers in tcb with suitable protection values */
 #ifdef CONFIG_ARM_MPU
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	uint8_t nregion = mpu_get_nregion_info(MPU_REGION_APP_BIN);
+
 	/* Configure text section as RO and executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP_TXT, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
+	mpu_get_register_config_value(&rtcb->mpu_regs[0], nregion - 3, (uintptr_t)binp->alloc[ALLOC_TEXT], binp->textsize, true, true);
 	/* Configure ro section as RO and non-executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[3], nregion - 2, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[6], nregion - 1, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
 #else
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP, (uintptr_t)binp->ramstart, binp->ramsize, false, true);
+	mpu_get_register_config_value(&rtcb->mpu_regs[0], nregion - 1, (uintptr_t)binp->ramstart, binp->ramsize, false, true);
 #endif
 #endif
 #endif /* CONFIG_APP_BINARY_SEPARATION */
--- a/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
+++ b/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
@@ -1338,6 +1338,9 @@ extern void __libc_init_array(void);
 #ifdef CONFIG_PLATFORM_TIZENRT_OS
 	mpu_init();
 	app_mpu_nocache_init();
+
+	/* Initialize number of mpu regions for board specific purpose */
+	mpu_set_nregion_board_specific(1);
 #endif
 	app_vdd1833_detect();
 	memcpy_gdma_init();
--- a/os/include/tinyara/mpu.h
+++ b/os/include/tinyara/mpu.h
@@ -34,11 +34,30 @@
  * Pre-processor Definitions
  ********************************************************************************/
 
+/* Structure for MPU region numbers */
+
+struct mpu_usages_s {
+	uint8_t nregion_board_specific;
+	uint8_t nregion_common_bin;
+	uint8_t nregion_app_bin;
+	uint8_t nregion_stackovf;
+	uint8_t max_nregion;
+};
+
 enum {
 	REG_RNR,
 	REG_RBAR,
 	REG_RASR,
 	REG_MAX
+};
+
+/* Enum to to get MPU region number info for MPU usages */
+enum mpu_region_usages_e {
+	MPU_REGION_BOARD_SPECIFIC,
+	MPU_REGION_COMMON_BIN,
+	MPU_REGION_APP_BIN,
+	MPU_REGION_STACKOVF,
+	MPU_REGION_MAX
 };
 
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
@@ -47,47 +66,10 @@ enum {
 #define MPU_NUM_REGIONS		1
 #endif
 
-/* Enum for MPU region numbers */
-
-enum MPU_REG_NUM {
-#ifdef CONFIG_APP_BINARY_SEPARATION
-#ifdef CONFIG_SUPPORT_COMMON_BINARY
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
-	MPU_REG_NUM_COM_LIB_TXT,	/* Common Binary */
-	MPU_REG_NUM_COM_LIB_RO,
-	MPU_REG_NUM_COM_LIB_DATA,
-#else
-	MPU_REG_NUM_COM_LIB,
+#ifdef CONFIG_ARMV8M_MPU
+#define MPU_ALIGNMENT_BYTES    32
+#define MPU_ALIGN_UP(a)                (((a) + MPU_ALIGNMENT_BYTES - 1) & ~(MPU_ALIGNMENT_BYTES - 1))
 #endif
-#endif
-#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
-	MPU_REG_NUM_APP_TXT,		/* Apps */
-	MPU_REG_NUM_APP_RO,
-	MPU_REG_NUM_APP_DATA,
-#else
-	MPU_REG_NUM_APP,
-#endif
-#endif
-#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
-	MPU_REG_NUM_STK,		/* Stack */
-#endif
-	MPU_REG_NUM_MAX
-};
-
-#ifdef CONFIG_ARMV7M_MPU
-#define CUR_MPU_REG_NUM		CONFIG_ARMV7M_MPU_NREGIONS
-#elif defined(CONFIG_ARMV8M_MPU)
-#define CUR_MPU_REG_NUM		CONFIG_ARMV8M_MPU_NREGIONS
-#define MPU_ALIGNMENT_BYTES	32
-#define MPU_ALIGN_UP(a)		(((a) + MPU_ALIGNMENT_BYTES - 1) & ~(MPU_ALIGNMENT_BYTES - 1))
-#endif
-
-#if defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU)
-#if (MPU_REG_NUM_MAX >= CUR_MPU_REG_NUM)
-#error "MPU region numbers exceeded !!"
-#endif
-#endif
-
 
 /********************************************************************************
  * Public Function Prototypes
@@ -100,5 +82,11 @@ void up_mpu_set_register(uint32_t *mpu_regs);
 bool up_mpu_check_active(uint32_t *mpu_regs);
 
 void up_mpu_disable_region(uint32_t *mpu_regs);
+
+void mpu_region_initialize(struct mpu_usages_s *mpu);
+
+void mpu_set_nregion_board_specific(uint8_t num);
+
+uint8_t mpu_get_nregion_info(enum mpu_region_usages_e usages);
 
 #endif
--- a/os/arch/arm/src/common/up_createstack.c
+++ b/os/arch/arm/src/common/up_createstack.c
@@ -299,8 +299,9 @@ int up_create_stack(FAR struct tcb_s *tcb, size_t stack_size, uint8_t ttype)
 #endif
 		board_led_on(LED_STACKCREATED);
 #ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+		uint8_t nregion = mpu_get_nregion_info(MPU_REGION_STACKOVF);
 		/* The smallest size that can be programmed for an MPU region is 32 bytes */
-		mpu_get_register_config_value(&tcb->stack_mpu_regs[0], MPU_REG_NUM_STK,
+		mpu_get_register_config_value(&tcb->stack_mpu_regs[0], nregion - 1,
 			(uint32_t)tcb->stack_alloc_ptr, CONFIG_MPU_STACK_GUARD_SIZE, true, false);
 #endif
 
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -187,12 +187,13 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 
 #if (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU))
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+		uint8_t nregion = mpu_get_nregion_info(MPU_REGION_COMMON_BIN);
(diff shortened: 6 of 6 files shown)


os/arch/Kconfig
```diff
@@ -107,13 +107,27 @@ config ARCH_USE_MMU
 config ARCH_USE_MPU
 	bool "Enable MPU"
 	default n
+	select MPU_RESERVED
```
sunghan-chang, line 110: When A is mandatory for B, then we use "select" syntax. But I don't think this case is fit.
How about removing this select and adding "depends on ARCH_USE_MPU" in MPU_RESERVED?
Because MPU_RESERVED could be unused with ARCH_USE_MPU, but ARCH_USE_MPU must be enabled to use MPU_RESERVED.

sunghan-chang (commented):
My original request is how to add board-specific MPU region.
We list the MPU regions in os/include/tinyara/mpu.h. But it's the common header, not a board-specific header.
Adding board-specific MPU regions in the common header is not good. That's why I requested.

This pr only includes the way how to give the offset. No way how to add board-specific value.
Could you update me about this?

sunghan-chang:
In first commit description, teh -> the

sunghan-chang:
By verification? confirmation?, could you add the conditional to check the max region number for each architecture?
```
for ARMv8-M
if (MPU_REG_NUM_MAX >= CONFIG_ARMV8M_MPU_NREGIONS) {
  error
}
for ARMv7-M
...
```
I think we could add this in mkconfig.c to verify in compilation time.

os/arch/Kconfig
thapav: Yes, I agree that would be more apt. I will modify

os/arch/Kconfig
```diff
@@ -107,13 +107,27 @@ config ARCH_USE_MMU
...
 
+if ARCH_USE_MPU
+
+config MPU_RESERVED
+	int "Number of reserved MPU regions"
+	default -1
+	---help---
+		The number of MPU regions reserved by the vendor. Some boards use
+		MPU regions for board specific functionalities. In such a scenario,
+		the current MPU regions need to be updated so that vendor specfic
```
jeongchanKim, line 126: specfic -> specific

os/include/tinyara/mpu.h
```diff
@@ -47,9 +47,18 @@ enum {
...
+ */
+
+#ifndef CONFIG_MPU_RESERVED
+#define CONFIG_MPU_RESERVED	(-1)
+#endif
+
 /* Enum for MPU region numbers */
 
 enum MPU_REG_NUM {
+	MPU_REG_NUM_RESERVE = CONFIG_MPU_RESERVED,
```
kishore-sn, line 61: I think we can use CONFIG_MPU_RESERVED directly in the first place of the enum. 
There is no need to introduce a new variable called MPU_REG_NUM_RESERVE

os/include/tinyara/mpu.h
```diff
@@ -47,9 +47,18 @@ enum {
 #define MPU_NUM_REGIONS		1
 #endif
 
+/* Number of reserved MPU regions for a board. This is defined by the vendor.
+ * If not, default value is taken to be 0.
```
kishore-sn, line 51: This statement seems wrong. Default value of reserved is taken as -1, so that our own region starts from zero.

os/arch/Kconfig
```diff
@@ -107,13 +107,27 @@ config ARCH_USE_MMU
...
 	---help---
 		The architecture supports MPU. Enable this option in order to
 		enable use of the MPU. For most architectures, this option is
 		enabled by other, platform-specific logic. In those cases, this
 		selection will always be forced.
 
+if ARCH_USE_MPU
+
+config MPU_RESERVED
+	int "Number of reserved MPU regions"
```
kishore-sn, line 121: Its better to use depends instead of "if ARCH_USE_MPU".

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,56 @@
...
+ *
+ * Description:
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_regs_initialize(FAR struct mpu_regs *mpu)
+{
+
+#ifdef CONFIG_APP_BINARY_SEPARATION
```
sunghan-chang, line 70: We should use another way to get the value for each.
Currently 2nd block concerns the config of 1st block. This makes code complex.
In 2nd block, only the config of 2nd block should be concerned.
How about below?
```c
uint8_t offset;

offset = mpu->board...;

#ifdef config_common
#ifdef config_reload_time
offse += 3;
#else
offset += 1;
#endif
mpu->common.. = offset;
#endif

#ifdef config_app_separation
#ifdef config_reload_time
offse += 3;
#else
offset += 1;
#endif
mpu->app_regs = offset;
#endif
```

os/include/tinyara/mpu.h
```diff
@@ -34,6 +34,16 @@
  * Pre-processor Definitions
  ********************************************************************************/
 
+struct mpu_regs {
+	uint8_t board_mpu_regs;
```
sunghan-chang, line 38: What does the member mean? The number of mpu region for each? The highest number for each? The start number?

sunghan-chang:
Please add the code to check overflow mpu region like
```c
if (mpu->max >= CONFIG_ARMV7M_MPU_NUM)
    #error alkfjalfjal;f;alfj
```

os/include/tinyara/mpu.h
thapav: this is the number of MPU reserved regions for the vendor

thapav:
Updated the PR, Kindly check.

> Please add the code to check overflow mpu region like
> 
> ```c
> if (mpu->max >= CONFIG_ARMV7M_MPU_NUM)
>     #error alkfjalfjal;f;alfj
> ```

Since we cannot add compile time check using variables, I have added DEBUGASSERT for the same. 
But, we have a similar check in mpu_get_register_config_value() where the board asserts as well. I think this check is not needed. Let me know your opinion.

os/arch/arm/src/common/up_mpuinit.c
thapav: Done

os/include/tinyara/mpu.h
```diff
@@ -34,6 +34,18 @@
...
 
+/* Structure for MPU region numbers */
+
+struct mpu_regs {
+	uint8_t nmpu_regions;
+	uint8_t common_regions;
+	uint8_t app_regions;
+	uint8_t stackovf_regions;
+	uint8_t maxmpu_regions;
+};
```
sunghan-chang, line 45: I suggest below
```c
struct mpu_usages_s {
    uint8_t nregion_board_specific;
    uint8_t nregion_common_bin;
    uint8_t nregion_app_bin;
    uint8_t nregion_stackovf;
    uint8_t max_nregion;
};
```

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -66,6 +106,20 @@
...
+	/* Maximum MPU region number sanity check */
+#ifdef CONFIG_ARMV7M_MPU
+	if (g_mpuregs.maxmpu_regions > CONFIG_ARMV7M_MPU_NREGIONS) {
+		DEBUGASSERT(g_mpuregs.maxmpu_regions < CONFIG_ARMV7M_MPU_NREGIONS);
+	}
+#elif defined(CONFIG_ARMV8M_MPU)
+	if (g_mpuregs.maxmpu_regions > CONFIG_ARMV8M_MPU_NREGIONS) {
+		DEBUGASSERT(g_mpuregs.maxmpu_regions < CONFIG_ARMV8M_MPU_NREGIONS);
+	}
+#endif
```
sunghan-chang, line 121: Let's reduce duplicated code.
```c
#ifdef CONFIG_V7
#define MPU_MAX_NREGION    CONFIG_...V7M...
#elif defined (CONFIG_V8)
#define 
#endif 

void up_mpuinitialize(void)
{
    ....

    if (g_mpu_... >= MPU_MAX_NREGION) {
        debugassert
    }
}
```
1. debugassert is enough? not assert? What happen when we try to set invalid mpu region?
2. In the conditional, don't we need to '='? I think it needs.

os/include/tinyara/mpu.h
```diff
@@ -34,6 +34,18 @@
...
+
+extern struct mpu_regs g_mpuregs;
```
sunghan-chang, line 47: The name 'g_mpuregs' looks like "gloabl_mpu_registers". We already use `reg` in tcb_s for register. Let's rename it as `g_mpu_usages`? `g_mpu_region_info`? or something else..

sunghan-chang: SAM recommends not to use global variable. Let's make a new function and call it from outside to get mpu region number information.
```
in mpu.h
enum mpu_region_usages {
   MPU_REGION_BOARD_SPECIFIC,
   ...
}

in up_mpuinit.c (common file)
static struct mpu_usages_s g_mpu_usages;

uint8_t mpu_get_region_info(enum mpu_region_usages usages)
{
     uint8_t nregion;

     switch (usages) {
     case MPU_REGION_BOARD_SPECIFIC:
          nregion = g_mpu_xxxx.nregion_board_specific;
           break;
      case ...
      }
   return nregion;
}
```

We need to make a set function for board specific value.

thapav: modified

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+ * Name: mpu_regs_initialize
+ *
+ * Description:
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_regs_initialize(FAR struct mpu_usages_s *mpu)
+{
+
```
sunghan-chang, line 75: Could you remove this empty line?

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
 
+/****************************************************************************
+ * Name: mpu_regs_initialize
+ *
+ * Description:
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_regs_initialize(FAR struct mpu_usages_s *mpu)
```
sunghan-chang, line 73: Let's remove `FAR`

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+		break;
+	case MPU_REGION_APP_BIN:
+		g_mpu_region_usages.nregion_app_bin = val;
+		break;
+	case MPU_REGION_STACKOVF:
+		g_mpu_region_usages.nregion_stackovf = val;
+		break;
+	case MPU_REGION_MAX:
+		g_mpu_region_usages.max_nregion = val;
+		break;
```
sunghan-chang, line 162: We set almost region_usages in this file except board specific. I think no need to support others.

os/include/tinyara/mpu.h
```diff
@@ -34,61 +34,43 @@
...
+
 enum {
 	REG_RNR,
 	REG_RBAR,
 	REG_RASR,
 	REG_MAX
 };
 
+/* Enum to to get MPU region number info for MPU usages */
+enum mpu_region_usages {
```
sunghan-chang, line 55: Could you add `_e` postfix to say the type?
`enum mpu_region_usages_e`

os/binfmt/binfmt_execmodule.c
```diff
@@ -188,15 +188,17 @@ int exec_module(FAR struct binary_s *binp)
...
 	/* Configure ro section as RO and non-executable region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[3], MPU_REG_NUM_APP_RO, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[3], nregion - 2, (uintptr_t)binp->alloc[ALLOC_RO], binp->rosize, true, false);
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[6], MPU_REG_NUM_APP_DATA, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
+	mpu_get_register_config_value(&rtcb->mpu_regs[6], nregion - 1, (uintptr_t)binp->alloc[ALLOC_DATA], binp->ramsize, false, false);
 #else
 	/* Complete RAM partition will be configured as RW region */
-	mpu_get_register_config_value(&rtcb->mpu_regs[0], MPU_REG_NUM_APP, (uintptr_t)binp->ramstart, binp->ramsize, false, true);
+	mpu_get_register_config_value(&rtcb->mpu_regs[0], nregion - 1, (uintptr_t)binp->ramstart, binp->ramsize, false, true);
```
sunghan-chang, line 201: Here is `nregion - 1` but in stack protection, you use `nregion`. Something wrong either here or stack protection.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_regs_initialize(FAR struct mpu_usages_s *mpu)
+{
+
+	uint8_t offset = mpu->nregion_board_specific;
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
```
jeongchanKim, line 78: `CONFIG_SUPPORT_COMMON_BINARY` and `CONFIG_OPTIMIZE_APP_RELOAD_TIME` can be enabled only when `CONFIG_APP_BINARY_SEPARATION`.
So I think you can move line 87 in the above.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	offset += 3;
```
jeongchanKim, line 80: Could you add comment why you add 3 or 1 in case of each config?

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+ *
+ * Description:
+ *   Initialize the MPU region numbers.
+ *
+ ****************************************************************************/
+
+void mpu_regs_initialize(FAR struct mpu_usages_s *mpu)
+{
+
+	uint8_t offset = mpu->nregion_board_specific;
```
jeongchanKim, line 76: `mpu` can be dereferencing.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+static struct mpu_usages_s g_mpu_region_usages;
```
jeongchanKim, line 53: For `mpu_usages_s`, please include `tinyara/mpu.h`.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +50,119 @@
...
+		break;
+	case MPU_REGION_STACKOVF:
+		nregion = g_mpu_region_usages.nregion_stackovf;
+		break;
+	case MPU_REGION_MAX:
+		nregion = g_mpu_region_usages.max_nregion;
+		break;
+	}
+
+	return nregion;
```
jeongchanKim, line 134: If there is not matched case, it returns trash value. Please add default or initialize the `nregion`.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -66,6 +174,14 @@
 
 void up_mpuinitialize(void)
 {
+	/* Initialize MPU register numbers' information */
+	mpu_regs_initialize(&g_mpu_region_usages);
+
+	/* Maximum MPU region number sanity check */
+	if (g_mpu_region_usages.max_nregion >= MPU_MAX_NREGION) {
+		DEBUGASSERT(MPU_MAX_NREGION < CONFIG_ARMV8M_MPU_NREGIONS);
```
kishore-sn, line 182: Please add a check for ARMV7 also. Else it would cause compile error.

os/binfmt/binfmt_execmodule.c
thapav: Stack protection needs to be updated

os/arch/arm/src/common/up_mpuinit.c
thapav: But since we need to distinguish between the scenarios of configuring complete RAM partition vs that of (text, RO & data) separately, we need CONFIG_OPTIMIZE_APP_RELOAD_TIME
And to have a common binary usage , we need CONFIG_SUPPORT_COMMON_BINARY

os/arch/arm/src/common/up_mpuinit.c
thapav: This has been supported while configuring MPU regions. So i think not needed here again.

thapav:
Addressed the review comments

os/binfmt/binfmt_execmodule.c
```diff
@@ -188,15 +188,17 @@ int exec_module(FAR struct binary_s *binp)
 	/* Initialize the MPU registers in tcb with suitable protection values */
 #ifdef CONFIG_ARM_MPU
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	uint8_t nregion = mpu_get_region_info(MPU_REGION_APP_BIN);
```
sunghan-chang, line 191: It looks we need to check the return value for 0.

os/arch/arm/src/common/up_mpuinit.c
sunghan-chang: I found the `MPU_NUM_REGIONS` in <tinyara/mpu.h>. It looks the same purpose of this.
Could you check that below is right?
```c
     offset += MPU_NUM_REGIONS;
     mpu->nregion_common_bin = offset;
#endif 

     offset += MPU_NUM_REGIONS;
     mpu->nregion_app_bin = offset;
#endif 
```

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1338,6 +1338,9 @@ extern void __libc_init_array(void);
 #ifdef CONFIG_PLATFORM_TIZENRT_OS
 	mpu_init();
 	app_mpu_nocache_init();
+
+	/* Initialize number of board mpu regions */
+	mpu_set_region_info(MPU_REGION_BOARD_SPECIFIC, 1);
```
sunghan-chang, line 1343: Please split the commit for this change.

os/include/tinyara/mpu.h
```diff
@@ -101,4 +83,10 @@ bool up_mpu_check_active(uint32_t *mpu_regs);
 
 void up_mpu_disable_region(uint32_t *mpu_regs);
 
+void mpu_regs_initialize(FAR struct mpu_usages_s *mpu);
```
sunghan-chang, line 86: As I said before, regs could be "registers", not "regions" because of tcb_s. Let's write full word for region as you did for other function and variable.
```
void mpu_region_initialize....
```

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -66,6 +162,14 @@
 
 void up_mpuinitialize(void)
 {
+	/* Initialize MPU register numbers' information */
+	mpu_regs_initialize(&g_mpu_region_usages);
+
+	/* Maximum MPU region number sanity check */
+	if (g_mpu_region_usages.max_nregion >= MPU_MAX_NREGION) {
+		DEBUGASSERT(g_mpu_region_usages.max_nregion < MPU_MAX_NREGION);
```
sunghan-chang, line 170: if and debugassert are duplicated.
You can remove if conditional.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,106 @@
...
+ ****************************************************************************/
+
+void mpu_set_region_info(enum mpu_region_usages_e usages, uint8_t val)
+{
+	switch (usages) {
+	case MPU_REGION_BOARD_SPECIFIC:
+		g_mpu_region_usages.nregion_board_specific = val;
+		break;
+	}
+}
```
sunghan-chang, line 152: This function only supports board-specific. 
1. The usages argument looks redundant.
2. There is no way to get any result for invalid request.

How about changing the function to support only board-specific?
```
/* Set a number of mpu region for board-specific purpose */
void mpu_set_nregion_board_specific(uint8_t num)
{
    g_mpu_region_usages.nregion_board_specific = num;
}
```

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,106 @@
...
+#endif
+	mpu->nregion_app_bin = offset;
+#endif
+
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+	mpu->nregion_stackovf = mpu->nregion_app_bin + 1;
+	mpu->max_nregion = mpu->nregion_stackovf;
+#else
+	mpu->max_nregion = mpu->nregion_app_bin;
+#endif
```
sunghan-chang, line 101: How about below?
```
#ifdef CONFIG_STACK...
   offset += 1;
   mpu->.... = offset;
#endif 

   mpu->max_nregion = offset;
```

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
thapav: done

thapav:
Addressed the review comments, kindly check

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1338,6 +1338,9 @@ extern void __libc_init_array(void);
 #ifdef CONFIG_PLATFORM_TIZENRT_OS
 	mpu_init();
 	app_mpu_nocache_init();
+
+	/* Initialize number of mpu regions for board specific purpose */
+	mpu_set_nregion_board_specific(1);
```
sijupunnnoose, line 1343: How about having no of MPU regions as a config instead of hardcoding in code? Just a thought!

thapav: As discussed over knox, I will mark this comment as resolved

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+#elif defined(CONFIG_ARMV8M_MPU)
+#define MPU_MAX_NREGION		CONFIG_ARMV8M_MPU_NREGIONS
+#endif
 
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: mpu_regs_initialize
```
sunghan-chang, line 67: Could you check the name?

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+void mpu_region_initialize(struct mpu_usages_s *mpu)
+{
+	uint8_t offset = (struct mpu_usages_s *)mpu->nregion_board_specific;
+
+#ifdef CONFIG_SUPPORT_COMMON_BINARY
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	offset += MPU_NUM_REGIONS;
+#else
+	offset += MPU_NUM_REGIONS;
+#endif
```
sunghan-chang, line 83: #ifdef and #else are the same. Let's remove the conditional and duplicated line.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+	mpu->nregion_common_bin = offset;
+#endif
+
+#ifdef CONFIG_APP_BINARY_SEPARATION
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	offset += MPU_NUM_REGIONS;
+#else
+	offset += MPU_NUM_REGIONS;
+#endif
```
sunghan-chang, line 92: same as previous

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+	offset += 1;
+	mpu->nregion_stackovf = offset;
+#endif
+	mpu->max_nregion = offset;
+}
+
+/****************************************************************************
+ * Name: mpu_get_region_info
```
sunghan-chang, line 104: check the name

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+ *
+ * Description:
+ *   Get the MPU region numbers.
+ *
+ ****************************************************************************/
+
+uint8_t mpu_get_nregion_info(enum mpu_region_usages_e usages)
+{
+	uint8_t nregion = 0;
```
sunghan-chang, line 113: Let's add `default` instead of zero initialization. Because it can reduce unnecessary operation in the cases of valid usages.

os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -50,11 +51,102 @@
...
+	case MPU_REGION_MAX:
+		nregion = g_mpu_region_usages.max_nregion;
+		break;
+	}
+
+	return nregion;
+}
+
+/****************************************************************************
+ * Name: mpu_set_region_info
```
sunghan-chang, line 137: name

os/include/tinyara/mpu.h
```diff
@@ -101,4 +83,10 @@ bool up_mpu_check_active(uint32_t *mpu_regs);
 
 void up_mpu_disable_region(uint32_t *mpu_regs);
 
+void mpu_region_initialize(FAR struct mpu_usages_s *mpu);
```
sunghan-chang, line 86: could you remove `FAR`?

os/arch/arm/src/common/up_mpuinit.c
thapav: done

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #4598, opened 2020-11-06
apps/examples/context_switching_performance : Add context switching p…

jeongchanKim:
…erformance measure app

This application makes two tasks which have highest priority.
They call the sched_yield for 10,000th iteration.
After that, calculate the average context switching time.

### DIFF
--- a/apps/examples/context_switching_performance/ctx_switch_main.c
+++ b/apps/examples/context_switching_performance/ctx_switch_main.c
@@ -1,0 +1,115 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * examples/context_switching_performance/ctx_switch_main.c
+ *
+ *   Copyright (C) 2008, 2011-2012 Gregory Nutt. All rights reserved.
+ *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name NuttX nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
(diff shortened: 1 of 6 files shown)


apps/examples/context_switching_performance/ctx_switch_main.c
```diff
@@ -0,0 +1,117 @@
...
+	clock_gettime(CLOCK_MONOTONIC, &end);
+
+	diffTime = ((double)end.tv_sec + 1.0e-9 * end.tv_nsec) - ((double)start.tv_sec + 1.0e-9 * start.tv_nsec);
+
+	printf("%d-th Average Context Switching Time is %.10f seconds\n", SWITCHING_ITERATIONS, (double)diffTime / (2 * SWITCHING_ITERATIONS));
+
+	return 0;
+}
+
+int yield_task_2(int a, char *b[])
```
sunghan-chang, line 88: Let's add static.

apps/examples/context_switching_performance/ctx_switch_main.c
```diff
@@ -0,0 +1,117 @@
...
+#else
+int ctx_switch_main(int argc, char *argv[])
+#endif
+{
+	printf("Context Switching Performance Measurement\n");
+
+	task_create("A_Task", SCHED_PRIORITY_MAX, 1024, yield_task_1, NULL);
+	task_create("B_Task", SCHED_PRIORITY_MAX, 1024, yield_task_2, NULL);
+
+	printf("Started!\n");
```
sunghan-chang, line 111: Do we need this and belows?

sunghan-chang (commented):
We could some more performance example. How about having the performance test folder and then put this there?

jeongchanKim:
> We could some more performance example. How about having the performance test folder and then put this there?

I will make folder with another PR.

apps/examples/context_switching_performance/ctx_switch_main.c
```diff
@@ -0,0 +1,117 @@
...
+
+#ifdef CONFIG_BUILD_KERNEL
+int main(int argc, FAR char *argv[])
+#else
+int ctx_switch_main(int argc, char *argv[])
+#endif
+{
+	printf("Context Switching Performance Measurement\n");
+
+	task_create("A_Task", SCHED_PRIORITY_MAX, 1024, yield_task_1, NULL);
```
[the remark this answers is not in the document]
jeongchanKim: Done.

apps/examples/context_switching_performance/ctx_switch_main.c
```diff
@@ -0,0 +1,118 @@
...
+	printf("%d-th Average Context Switching Time is %.10f seconds\n", SWITCHING_ITERATIONS, (double)diff_time / (2 * SWITCHING_ITERATIONS));
+
+	return 0;
+}
+
+static int yield_task_2(int a, char *b[])
+{
+	int cnt = SWITCHING_ITERATIONS;
+
+	usleep(1);
```
sunghan-chang, line 93: You should remove usleep because you use sched_lock and unlock.

apps/examples/context_switching_performance/ctx_switch_main.c
```diff
@@ -0,0 +1,118 @@
...
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <sched.h>
+#include <time.h>
+#include <sys/types.h>
+
+#define SWITCHING_ITERATIONS 10000
```
sunghan-chang, line 63: Could you increase the number of iteration? Because the time of context switching is several micro seconds, after 10000 iterations, it takes several mili seconds. It could be almost the same as a tick. Let's do that at least 1,000,000 times.

sunghan-chang (approved):

sunghan-chang (approved):

PR #4601, opened 2020-11-09
Add rdp demo for loadable apps

kishore-sn:
- Add rdp demo for loadable apps using kernel test driver ioctl
- Enable fpu to fix memfault issue when returning from rdp demp ioctl


build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -81,8 +81,9 @@ CONFIG_ARCH_ARMV8M_FAMILY=y
 CONFIG_ARCH_FAMILY="armv8-m"
 CONFIG_ARCH_HAVE_CMNVECTOR=y
 CONFIG_ARM_CMNVECTOR=y
-# CONFIG_ARCH_HAVE_FPU is not set
+CONFIG_ARCH_HAVE_FPU=y
```
jeongchanKim, line 84: Is there any specific reason to use FPU?

kishore-sn: The amebad chip has a fpu in the hardware. Due to this, when an exception happens, the hardware saves the fpu registers on the stack by default. But if we do not set the FPU config in software, then this somehow affects the default stacking mechanism during exception handling. This is resulting in a corruption of PC register value when returning from the rdp demo test.  I think whenever, a FPU hardware is present, it is mandatory to set this config for proper operation.

sunghan-chang: Could you enable this in all of defconfigs for realtek board?

kishore-sn: Done.

jeongchanKim (commented):
As you said in PR#4590, if rdp is only for AMEBAD, then how about renaming the example and file name with amebad_rdp_main.c and test_amebad_rdp.c?

os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	up_allocate_secure_context(1024);
+	for (i = 0; i < 32; i++) {
+		rdp_result = rdp_protection_entry(i);
+		no_rdp_result = rdp_no_protection_call(no_protection_func, i);
+
+		if (rdp_result != no_rdp_result) {
+			dbg("rdp call fail!\n");
+			dbg("rdp_result = 0x%x, no_rdp_result=0x%x\n", rdp_result, no_rdp_result);
+			goto end;
```
jeongchanKim, line 64: `test_rdp` always returns 0, so we cannot check this test runs well or not.
But you checked the return of this function in rdp_main.c:143.
I think we should declare `ret` and set to ERROR or something for failure case.

os/drivers/testcase/test_rdp.c
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+
+	dbg("rdp demo call succeed!\n");
+end:
```
Taejun-Kwon, line 69: You'd better use goto to handle exception like a try-catch. there's no reason for 'end' here... please see below.
```
if (rdp_result != no_rdp_result) {
     dbg();
     break;
}
...

up_free_secure_context();
if (i == 32) {
    dbg("success\n");
    return 0;
} 

dbg("Failed");
return -1;

```

os/drivers/testcase/testcase_drv.c
```diff
@@ -168,6 +169,12 @@ static int testcase_drv_ioctl(FAR struct file *filep, int cmd, unsigned long arg
 		break;
 #endif
 
+#ifdef CONFIG_AMEBAD_TRUSTZONE
+	case TESTIOC_RDP:
+		ret = test_rdp();
+		break;
```
Taejun-Kwon, line 175: It is driver, it should return one of error in errno.h if ret is not OK.

kishore-sn:
> As you said in PR#4590, if rdp is only for AMEBAD, then how about renaming the example and file name with amebad_rdp_main.c and test_amebad_rdp.c?

How about I do something like below:
- I will rename test_rdp to test_trustzone
- I will enable the test_trustzone.c based on CONFIG_ARMV8M_TRUSTZONE
- Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions. 
- However, we will still call the ioctl for IOC_TZ in the amebad rdp_demo.c app. 

Does this sound ok? In kernel side, it will be common code based on CONFIG_ARMV8M_TRUSTZONE. In user side, we will call it from rdp_demo.c.

os/drivers/testcase/test_rdp.c
kishore-sn: Done.

os/drivers/testcase/test_rdp.c
kishore-sn: Done

os/drivers/testcase/testcase_drv.c
kishore-sn: Done

apps/examples/rdp/rdp_main.c
```diff
@@ -119,16 +125,26 @@ int main(int argc, FAR char *argv[])
...
+	int pid = task_create("RDP DEMO", SCHED_PRIORITY_DEFAULT, CONFIG_USERMAIN_STACKSIZE, (main_t)rdp_demo, (FAR char *const *)NULL);
 	if (pid < 0) {
 		printf("Failed to create rdp demo thread\n");
 	}
+#else
+	int ret = 0;
+	int tc_fd = open(TESTCASE_DRVPATH, O_WRONLY);
+	if (tc_fd < 0) {
+		printf("FAILED to open testcase driver %s\n", TESTCASE_DRVPATH);
+		return;
```
Taejun-Kwon, line 140: it should return integer.

apps/examples/rdp/rdp_main.c
```diff
@@ -119,16 +125,26 @@ int main(int argc, FAR char *argv[])
...
+	}
+
+	ret = ioctl(tc_fd, TESTIOC_RDP, 0);
+	if (ret	< 0) {
+		printf("ERROR executing RDP test ioctl ret = %d\n", ret);
```
Taejun-Kwon, line 145: ret is always ERROR if ioctl failed. so we should check errno actually.

sunghan-chang:
> I will rename test_rdp to test_trustzone
I will enable the test_trustzone.c based on CONFIG_ARMV8M_TRUSTZONE
Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions.
However, we will still call the ioctl for IOC_TZ in the amebad rdp_demo.c app.
Does this sound ok? In kernel side, it will be common code based on CONFIG_ARMV8M_TRUSTZONE. In user side, we will call it from rdp_demo.c.

Ok. Let's do that.

apps/examples/rdp/rdp_main.c
kishore-sn: Fixed

apps/examples/rdp/rdp_main.c
kishore-sn: Fixed

sunghan-chang:
@kishore-sn Because of configuration modification, we should modify the Kconfig.
```
config EXAMPLES_RDP
	bool "Read Protect example"
	default n
	depends on ARCH_CHIP_AMEBAD if BUILD_FLAT
	depends on AMEBAD_TRUSTZONE if BUILD_FLAT
	depends on ARMV8M_TRUSTZONE if APP_BINARY_SEPARATION
	---help---
		Enable the Read Protect example
```

kishore-sn:
@sunghan-chang 
Is this Kconfig change necessary? 
Even in case of app separation, the rdp example finally depends on AMEBAD_TRUSTZONE only. And if AMEBAD_TRUSTZONE is enabled, then ARMV8M_TRUSTZONE is selected by default.

Taejun-Kwon (dismissed):
good.

os/drivers/testcase/test_tz.c
```diff
@@ -0,0 +1,86 @@
...
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include "testcase_proto.h"
+#include <tinyara/arch.h>
+#ifdef CONFIG_AMEBAD_TRUSTZONE
+#include <arch/chip/amebad_nsc.h>
+#endif
+
+#include <errno.h>
```
thapav, line 30: Shall we maintain headers as below ?
#include <tinyara/config.h>
##standard headers <errno.h>
##TizenRT specific public headers like <tinyara/xxx.h>
##chip specific headers
##internal headers with " "

os/drivers/testcase/test_tz.c
```diff
@@ -0,0 +1,86 @@
...
+	result += 8;
+
+	return result;
+}
+#endif
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+int test_tz(void)
```
thapav, line 50: Can we add function description please ?

sunghan-chang:
> @sunghan-chang
Is this Kconfig change necessary?
Even in case of app separation, the rdp example finally depends on AMEBAD_TRUSTZONE only. And if AMEBAD_TRUSTZONE is enabled, then ARMV8M_TRUSTZONE is selected by default.

@kishore-sn `Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions.` This is what you said. If you don't modify the Kconfig, it does not work like above.

os/drivers/testcase/test_tz.c
kishore-sn: done

os/drivers/testcase/test_tz.c
kishore-sn: done

jeongchanKim:
> * I will rename test_rdp to test_trustzone
> * I will enable the test_trustzone.c based on CONFIG_ARMV8M_TRUSTZONE
> * Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions.
> * However, we will still call the ioctl for IOC_TZ in the amebad rdp_demo.c app.

I agree with your opinion. Thanks.

kishore-sn:
> @kishore-sn `Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions.` This is what you said. If you don't modify the Kconfig, it does not work like above.

@sunghan-chang 
My question is whether we need to differentiate between flat and app separation build in this case? In either case, rdp demo will work only if AMEBAD_TRUSTZONE is enabled. I understand your concern that in kernel size test_tz() depends on ARMV8_TRUSTZONE, but in user side rdp_demo() depends on AMEBAD_TRUSTZONE. I had mentioned this point before when I made this suggestion.

> However, we will still call the ioctl for IOC_TZ in the amebad rdp_demo.c app.
> Does this sound ok? In kernel side, it will be common code based on CONFIG_ARMV8M_TRUSTZONE. In user side, we will call it from rdp_demo.c.

However, ARMV8_TRUSTZONE will get automatically enabled when AMEBAD_TRUSTZONE is enabled, if menuconfig utility is run. So, I think in this case, everything will work well. But in a future project, if we enable only ARMV8_TRUSTZONE, then the test_tz will return error. If we add a new vendor specific mechanism for test_tz, then it will work based on the new mechanism. This is what I meant when i said this.

> Inside test_trustzone() api, I will run rdp test if CONFIG_AMEBAD_TRUSTZONE is enabled. Else, I will return error. In future, if we add different vendor, then we can add more code for vendor specific test in the else conditions.

sunghan-chang:
> My question is whether we need to differentiate between flat and app separation build in this case? In either case, rdp demo will work only if AMEBAD_TRUSTZONE is enabled. I understand your concern that in kernel size test_tz() depends on ARMV8_TRUSTZONE, but in user side rdp_demo() depends on AMEBAD_TRUSTZONE. I had mentioned this point before when I made this suggestion.

Then, let's remove CONFIG_BUILD_FLAT related codes in the app. Let's use ioctl only for both flat and app separation.

> However, ARMV8_TRUSTZONE will get automatically enabled when AMEBAD_TRUSTZONE is enabled, if menuconfig utility is run. So, I think in this case, everything will work well. But in a future project, if we enable only ARMV8_TRUSTZONE, then the test_tz will return error. If we add a new vendor specific mechanism for test_tz, then it will work based on the new mechanism. This is what I meant when i said this.

Then, how about `depends on AMEBAD_TRUSTZONE if ARCH_CHIP_AMEBAD`?

kishore-sn:
> Then, let's remove CONFIG_BUILD_FLAT related codes in the app. Let's use ioctl only for both flat and app separation.

Done

> Then, how about depends on AMEBAD_TRUSTZONE if ARCH_CHIP_AMEBAD? Because another vendor code could be added as you said.

I could not change like this because if condition is not allowed in depends statement syntax. However, I have taken your suggestion and changed the dependency for rdp from AMEBAD_TRUSTZONE to ARMV8M_TRUSTZONE.

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

thapav (dismissed):

sunghan-chang:
1. You should not add 2nd commit here again. Please remove that from this PR.
2. To allow FPU enabled, could you make a commit for Kconfig as you pushed before?
```
config AMEBA....
   select ARCH_HAVE_FPU
```

kishore-sn:
This commit got pushed by mistake, while i was pushing a different branch. 
I will fix this PR once the #4613 is merged.

sunghan-chang:
@kishore-sn I merged PR #4613 . Could you update this?

os/arch/arm/src/amebad/Kconfig
```diff
@@ -13,6 +13,7 @@ choice
 config ARCH_CHIP_RTL8721CSM
 	bool "RTL8721CSM"
 	select ARCH_CORTEXM33
+	select ARCH_HAVE_FPU
```
sunghan-chang, line 16: Could you split commits for this? This is not a topic of rdp demo.

sunghan-chang (approved):

PR #4605, opened 2020-11-11
arch/armv8-m: modify tz_context initialization value to 0 in up_svcall.c

sunghan-chang:
The tz_context type is TZ_ModuleId_t which is the same as uint32_t.
So, null is not proper. This causes build warning as below:

CC:  armv8-m/up_svcall.c
armv8-m/up_svcall.c: In function 'up_svcall':
armv8-m/up_svcall.c:571:20: warning: assignment makes integer from pointer without a cast [-Wint-conversion]
   rtcb->tz_context = NULL;
                    ^

Let's change the value to 0 to resolve above.


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4610, opened 2020-11-17
os/LibTargets.mk: Add FPU enabled libraries for RTK

kishore-sn:
We have two prebuilt libraries from RTK.
os/board/rtl8721csm/src/libs/lib_wlan_fpu.a
os/board/rtl8721csm/src/libs/lib_wps_fpu.a

In RTK project it is possible to either enable or disable FPU based
on usecase scenarios. In either case, we need to use the prebuilt
libraries which are suitably compiled with proper compile options to
either enable or disable the FPU usage in code. So, we will maintain
who versions of these libraries and choose the required one based on
whether the CONFIG_ARCH_FPU flag is enabled or not.


jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4611, opened 2020-11-18
kernel/group: Always detach the group from the tcb if the tcb has a group

jeongarmy:
A group_leave function is called when a task or thread exits.
It decrements the reference count of the group and frees all resources of group if the reference count is zero.
Then a group is detached from the tcb regardless of reference count to avoid doing this again.


sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #4612, opened 2020-11-18
binary_manager, micomapp and wifiapp: adjust the period of alert message

sunghan-chang:
1. Every 10 sec, micom and wifi binaries prints the alive message.
It's not good to use TASH.
Let's adjust the period from 10 secs to 5 mins.
2. The binary_manager_reset_board() resets the board when CONFIG_BOARDCTL_RESET
is enabled, but it just waits after printing the message every 10 sec with
CONFIG_BOARDCTL_RESET disabled.
Message at every 10 sec disturbs checking the debug messages.
Let's increase the period from 10 secs to 5 mins.


Taejun-Kwon (approved):
I Love it

PR #4613, opened 2020-11-18
build/tools/amebad : Add non FPU image 3 binaries for RTK

kishore-sn:
The main TizenRT image can be built either with FPU enabled or FPU disabled.
In order to match this, we need to have 2 versions of image 3 binary with
FPU enabled and disabled state. The right version to be used will be selected
based on the CONFIG_ARCH_FPU flag.


os/board/rtl8721csm/src/libs/.gitignore
```diff
@@ -1,5 +0,0 @@
-!./lib_wlan.a
-!./lib_wps.a
-!./lib_wlan_fpu.a
-!./lib_wps_fpu.a
-!./cmse_implib.a
```
sunghan-chang, line 5: Why do you remove this file?

sunghan-chang (approved):

sunghan-chang:
You need to add xxx_fpu.a files in .gitignore.

kishore-sn:
> You need to add xxx_fpu.a files in .gitignore.

Done.

sunghan-chang (approved):

PR #4614, opened 2020-11-19
os/drivers: rename testcase to kernel_test

sunghan-chang:
This driver supports testing kernel functionalities from user apps.
The start of this module supported kernel testcase only so that it was named
testcase. But currently this is used to test other format of kernel test.
Let's rename testcase to kernel_test.


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4615, opened 2020-11-19
arch/arm: set the default value of FPU config as n

sunghan-chang:
Even a board supports FPU, it's not mandatory feature.
It might cause performance degradation so that it should be
defaultly disabled. When it needs, we can enable.


thapav (approved):
LGTM

Taejun-Kwon (approved):

PR #4617, opened 2020-11-20
os : Change zalloc and free to kmm_zalloc and kmm_free

jeongchanKim:
In kernel, we should use kmm_zalloc and kmm_free for using kernel heap.
So changed them.


sunghan-chang (approved):

Taejun-Kwon (approved):
Thanks for fixing.

PR #4622, opened 2020-11-23
apps/example/security_test : Change malloc to zalloc for initializing…

jeongchanKim:
… the test condition

There can be remained data in allocated memory chunk if that chunk is used before.
To avoid remained data for testcase, change malloc to zalloc.


jeongarmy (approved):
LGTM

thapav (approved):

sunghan-chang (approved):

PR #4626, opened 2020-11-25
arch/armv8-m : Fix Security test hang issue

kishore-sn:
During context switch, the secure context store is not happening sometimes.
Load secure context is getting called repeatedly without corresponding calls
to score secure context. This is resulting in corruption of secure area PSP.
The issue is fixed by making sure that secure context is stored for each
task which is switched OUT during a context switch.

### DIFF
--- a/os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
+++ b/os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
@@ -136,11 +136,6 @@ void up_unblock_task_without_savereg(struct tcb_s *tcb)
 		tcb->task_state = TSTATE_TASK_RUNNING;
 		tcb->flink->task_state = TSTATE_TASK_READYTORUN;
 
-#ifdef CONFIG_ARMV8M_TRUSTZONE
-		if (tcb->tz_context) {
-			TZ_StoreContext_S(tcb->tz_context);
-		}
-#endif
 	} else {
 		/* The new btcb was added in the middle of the ready-to-run list */
 
--- a/os/arch/arm/src/armv8-m/up_unblocktask.c
+++ b/os/arch/arm/src/armv8-m/up_unblocktask.c
@@ -135,8 +135,8 @@ void up_unblock_task(struct tcb_s *tcb)
 
 	if (sched_addreadytorun(tcb)) {
 #ifdef CONFIG_ARMV8M_TRUSTZONE
-		if (tcb->tz_context) {
-			TZ_StoreContext_S(tcb->tz_context);
+		if (rtcb->tz_context) {
+			TZ_StoreContext_S(rtcb->tz_context);
 		}
 #endif
 		/* The currently active task has changed! We need to do


sunghan-chang (approved):

os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
```diff
@@ -136,17 +136,17 @@ void up_unblock_task_without_savereg(struct tcb_s *tcb)
...
-		}
-#endif
 	} else {
 		/* The new btcb was added in the middle of the ready-to-run list */
 
 		tcb->task_state = TSTATE_TASK_READYTORUN;
 	}
 
+#ifdef CONFIG_ARMV8M_TRUSTZONE
+		if (rtcb->tz_context) {
```
sunghan-chang, line 146: Could you check indentation? (alignment?)

jeongchanKim (approved):
Looks good to me.

os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
kishore-sn: Fixed.

jeongchanKim (approved):
LGTM

os/arch/arm/src/armv8-m/up_blocktask.c
```diff
@@ -155,8 +155,8 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 
 	if (switch_needed) {
 #ifdef CONFIG_ARMV8M_TRUSTZONE
-		if (tcb->tz_context) {
-			TZ_StoreContext_S(tcb->tz_context);
+		if (rtcb->tz_context) {
```
jeongchanKim, line 158: Is it correct to change `tcb` to `rtcb`?
Because 'tcb' is a tcb which will be blocked. So I think we should store its tz_context.

kishore-sn: Reverted it.

sunghan-chang (approved):

os/arch/arm/src/armv8-m/up_unblocktask_withoutsavereg.c
```diff
@@ -136,17 +136,17 @@ void up_unblock_task_without_savereg(struct tcb_s *tcb)
...
 		/* The new btcb was added in the middle of the ready-to-run list */
 
 		tcb->task_state = TSTATE_TASK_READYTORUN;
 	}
 
+#ifdef CONFIG_ARMV8M_TRUSTZONE
+	if (rtcb->tz_context) {
+		TZ_StoreContext_S(rtcb->tz_context);
+	}
+#endif
```
jeongarmy, line 149: I think it should be called if a head of list is changed to tcb (in line 131-138), isn't it?

kishore-sn: Even if the head is not changed, we running the below context switch code.
In that case we will call load_context() always. 
So, corresponding call to store_context() also needs to be done always.

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4629, opened 2020-11-26
apps/examples/security_test/hal : Add dependency for security hal tes…

jeongchanKim:
…t on Flat build

Security Hal Test can run only with flat build. It cannot run from Loadable apps.
So add dependency.

### DIFF
--- a/apps/examples/security_test/hal/Kconfig
+++ b/apps/examples/security_test/hal/Kconfig
@@ -4,6 +4,8 @@
 #
 config EXAMPLES_SECURITY_HAL_TEST
 	bool "Security HAL Test"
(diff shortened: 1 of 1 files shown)


apps/examples/security_test/hal/Kconfig
```diff
@@ -4,6 +4,7 @@
 #
 config EXAMPLES_SECURITY_HAL_TEST
 	bool "Security HAL Test"
+	depends on BUILD_FLAT
```
jeongarmy, line 7: I think it need a dependency for SE also.

sunghan-chang (approved):

PR #4630, opened 2020-11-26
build/rtl8721csm: Increase the maximum size of mq message

jeongarmy:
Set the maximum size of mq message to a sufficiently large size for various tests.


jeongchanKim (dismissed):
LGTM

sunghan-chang:
I don't think we need to increase it for all of defconfigs. If you think this necessary, please let me know why you don't set another configure value with big value.

sunghan-chang (approved):

jeongchanKim (approved):
Looks good to me

PR #4633, opened 2020-11-30
Stkmon time print

jeongchanKim:


os/kernel/Kconfig
```diff
@@ -153,6 +153,7 @@ endif # !SCHED_TICKLESS
 
 config SYSTEM_TIME64
 	bool "64-bit system clock"
+	depends on LIBC_LONG_LONG
```
sunghan-chang, line 156: Let's use `select`. Because it does not have any other dependency so that `select` gives better usability.

lib/libc/stdio/lib_sscanf.c
```diff
@@ -643,9 +643,9 @@ int vsscanf(FAR const char *buf, FAR const char *fmt, va_list ap)
 				if (*buf) {
 					FAR char *endptr;
 					int errsave;
-					unsigned long tmplong;
+					unsigned long tmplong = 0;
 #ifdef CONFIG_LIBC_LONG_LONG
-					unsigned long long tmplonglong;
+					unsigned long long tmplonglong = 0;
```
sunghan-chang, line 648: Please add error logs in commit description.

jeongchanKim: Added. Thanks.

sunghan-chang (approved):

PR #4635, opened 2020-12-01
arch/arm/common/up_createstack : Add condition for checking kernel heap

jeongchanKim:
kernel_thread() API only uses kernel heap regardless of calling from where.
so add type checking for where to allocate stack memory.


sunghan-chang (approved):

sunghan-chang (dismissed):

sunghan-chang (commented):
Could you include <tinyara/sched.h> to use this type?

kishore-sn (approved):
Looks good.

jeongarmy (approved):
LGTM

sunghan-chang (approved):

PR #4642, opened 2020-12-03
Add functionality to dump Userfs partition from External Flash

amogh-samsung:
- Add smart MTD ioctl to retrieve all smartfs sectors for
  corresponding partition identifier.
- Add option to dump tool interface to dump external
  userfs partition.

### DIFF
--- a/os/board/common/crashdump.c
+++ b/os/board/common/crashdump.c
@@ -25,11 +25,14 @@
 #include <tinyara/arch.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <fcntl.h>
+#include <errno.h>
 #include <string.h>
 #include <arch/board/board.h>
 #include "up_internal.h"
 #include "up_watchdog.h"
 #include <sys/boardctl.h>
+#include <limits.h>
 
 #include <tinyara/mm/heap_regioninfo.h>
 #include <tinyara/mm/mm.h>
@@ -40,6 +43,7 @@
 #if defined(CONFIG_BOARD_DUMP_UART)
 #define RAMDUMP_HANDSHAKE_STRING	"TIZENRTRMDUMP"
 #define FSDUMP_HANDSHAKE_STRING	"TIZENRTFSDUMP"
+#define EXTFSDUMP_HANDSHAKE_STRING	"TIZENRTEXTFSD"
 #define TARGET_REBOOT_STRING		"TIZENRTREBOOT"
 #define HANDSHAKE_STR_LEN_MAX		(13)
 #define USERFS_PART_NAME		"user"
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
 	return 0;
 }
 
+static int get_ext_userfs_part(size_t *size, int *index)
+{
+#if defined(CONFIG_SECOND_FLASH_PART_NAME) && defined(CONFIG_SECOND_FLASH_PART_SIZE)
+	int i = 0;
+	int c = 0;
+	int x = 0;
+	size_t ext_userfs_size;
+	char second_flash_part_list[] = CONFIG_SECOND_FLASH_PART_NAME;
+	char second_flash_part_sizes[] = CONFIG_SECOND_FLASH_PART_SIZE;
+	char tmp_part_name[USERFS_PART_NAME_LEN + 1] = { '0' };
+
+	while (second_flash_part_list[i] != '\0') {
+		if (second_flash_part_list[i] != 'u') {
+			while (second_flash_part_list[i] != ',') {
+				i++;
+			}
+			c++;
+		} else {
+			while ((x < USERFS_PART_NAME_LEN) && (second_flash_part_list[i] != ',')) {
+				tmp_part_name[x] = second_flash_part_list[i];
+				x++;
+				i++;
+			}
+			tmp_part_name[x] = '\0';
+			if ((x == USERFS_PART_NAME_LEN) && (strncmp(tmp_part_name, USERFS_PART_NAME, USERFS_PART_NAME_LEN) == 0)) {
+				break;
+			}
+			if (second_flash_part_list[i] == ',') {
+				c++;
+			}
+		}
+
+		x = 0;
+		i++;
+	}
+
+	if (second_flash_part_list[i] == '\0') {
+		*size = 0;
+		*index = -1;
+		return ERROR;
+	}
+
+	*index = c;
+	i = 0;
+	while (second_flash_part_sizes[i] != '\0' && c) {
+		while (second_flash_part_sizes[i] != ',') {
+			i++;
+		}
+		c--;
+		i++;
+	}
+
+	if (second_flash_part_sizes[i] == '\0') {
+		*size = 0;
+		*index = -1;
+		return ERROR;
+	}
+
+	ext_userfs_size = 0;
+	while (second_flash_part_sizes[i] != ',' && second_flash_part_sizes[i] != '\0') {
+		ext_userfs_size *= 10;
+		ext_userfs_size += (second_flash_part_sizes[i] - 48);
+		i++;
+	}
+	ext_userfs_size *= 1024;
+	*size = ext_userfs_size;
+
+#else
+	*size = 0;
+	*index = -1;
+#endif
+
+	return 0;
+}
+
 static int send_ramdump(void)
 {
 	int i;
@@ -201,7 +280,7 @@ static int send_ramdump(void)
 
 static int send_fsdump(void)
 {
-	int i;
+	int i = 0;
 	uint8_t *ptr;
 	uint32_t userfs_start;
 	size_t size;
@@ -235,6 +314,66 @@ static int send_fsdump(void)
 		ptr++;
 		size--;
 	}
+	return 0;
+}
+
+static int send_extfsdump(void)
+{
+	int i = 0;
+	int fd;
+	int index;
+	uint8_t *ptr;
+	size_t size;
+	char ext_userfs_partition[CONFIG_PATH_MAX] = { '\0' };
+	char ext_fsdump_buf[CONFIG_MTD_SMART_SECTOR_SIZE + 1];
+
+	/*Send ACK for EXTERNAL FSDUMP Handshake */
+	up_lowputc('A');
+
+	i = get_ext_userfs_part(&size, &index);
+	if (i < 0) {
+		return ERROR;
+	}
+
+	/* Send external userfs partition size to host */
+	ptr = (uint8_t *)&size;
+	for (i = 0; i < sizeof(size); i++) {
+		up_lowputc((uint8_t)*ptr);
+		ptr++;
+	}
+
+	/* If extracted size is 0, it means there is no partition available to dump */
+	if (size == 0) {
+		return 0;
+	}
+
+	//TODO:- Currently using blockproxy to create a temporary BCH driver instance to
+	//access the block device. Should be changed to create a permanent BCH instance
+	//on device initialization which can also be used for other use cases.
+	snprintf(ext_userfs_partition, CONFIG_PATH_MAX, "/dev/smart1p%d", index);
+	fd = open(ext_userfs_partition, O_RDONLY);
+	if (fd < 0) {
+		return errno;
+	}
+
+	while (size) {
+		// ToDo: Revert to the flexible logic that searches sectors and
+		//       reads sector sizes stored in the sectors instead of
+		//               using CONFIG_MTD_SMART_SECTOR_SIZE.
+		memset(ext_fsdump_buf, 0, CONFIG_MTD_SMART_SECTOR_SIZE);
+		i = read(fd, ext_fsdump_buf, CONFIG_MTD_SMART_SECTOR_SIZE);
+		if (i < CONFIG_MTD_SMART_SECTOR_SIZE) {
+			return ERROR;
+		}
+
+		ptr = (uint8_t *)ext_fsdump_buf;
+		for (i = CONFIG_MTD_SMART_SECTOR_SIZE; i > 0; i--) {
+			up_lowputc((uint8_t)*ptr);
+			ptr++;
+		}
+		size -= CONFIG_MTD_SMART_SECTOR_SIZE;
+	}
+
 	return 0;
 }
 
@@ -298,6 +437,11 @@ static int dump_via_uart(void)
 				break;
 			}
 		/* Check handshake string for TARGET REBOOT signal */
+		} else if (strncmp(host_buf, EXTFSDUMP_HANDSHAKE_STRING, strlen(EXTFSDUMP_HANDSHAKE_STRING)) == 0) {
+			i = send_extfsdump();
+			if (i != 0) {
+				break;
+			}
 		} else if (strncmp(host_buf, TARGET_REBOOT_STRING, strlen(TARGET_REBOOT_STRING)) == 0) {
 			/* Acknowledge the reboot signal, then reboot device if CONFIG is enabled, otherwise simply exit the function */
 			up_lowputc('A');
--- a/tools/dump_tool/dump_tool.c
+++ b/tools/dump_tool/dump_tool.c
@@ -244,6 +244,50 @@ static int fsdump_recv(int dev_fd)
 	return 0;
 }
 
+static int extfsdump_recv(int dev_fd)
+{
+        int ret;
+        uint32_t ext_smartfs_size;
+        char ext_fs_bin[BINFILE_NAME_SIZE] = { '\0' };
+        FILE *fs_fp;
+
+        /* Receive Userfs flash partition size */
+        ret = b_read(dev_fd, (uint8_t *)&ext_smartfs_size, 4);
+        if (ret != 4) {
+                printf("Receiving Userfs size failed, ret = %d\n", ret);
+                return -1;
+        }
+
+	if (ext_smartfs_size == 0) {
+		printf("No external userfs partition found. Please check config\n");
+		return 0;
+	}
+
+        printf("\n=========================================================================\n");
+        printf("External filesystem size = %u\n", ext_smartfs_size);
+        printf("=========================================================================\n");
+
+        /* Create filesystem dump binary file name */
+        snprintf(ext_fs_bin, BINFILE_NAME_SIZE, "ext_fsdump_%08u.bin", ext_smartfs_size);
+
+        printf("\nReceiving external file system dump.....\n");
+
+        fs_fp = fopen(ext_fs_bin, "w");
+        if (fs_fp == NULL) {
+                printf("%s create failed\n", ext_fs_bin);
+                return -1;
+        }
+
+        ret = get_dump(dev_fd, fs_fp, ext_smartfs_size);
+        fclose(fs_fp);
+        if (ret != 0) {
+                printf("Unable to receive complete FS Dump!!!\n");
+                return -1;
+        }
+
+        return 0;
+}
+
 int main(int argc, char *argv[])
 {
 	int ret;
@@ -306,7 +350,7 @@ int main(int argc, char *argv[])
 	printf("Target device locked and ready to DUMP!!!\n");
 
 	while (1) {
-		printf("Choose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. Exit Dump Tool\n5. Reboot TARGET Device\n");
+		printf("Choose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. External Userfs Partition Dump\n5. Exit Dump Tool\n6. Reboot TARGET Device\n");
 		fflush(stdout);
 		scanf("%d", &choice);
 		getchar();
@@ -325,9 +369,11 @@ int main(int argc, char *argv[])
 			DUMP_FLAGS |= USERFSDUMP_FLAG;
 			break;
 		case 4:
-			DUMP_FLAGS |= EXIT_TOOL_FLAG;
+			DUMP_FLAGS |= EXTUSERFS_DUMP_FLAG;
 			break;
 		case 5:
+			break;
+		case 6:
 			DUMP_FLAGS |= REBOOT_DEVICE_FLAG;
 			break;
 		default:
@@ -338,6 +384,7 @@ int main(int argc, char *argv[])
 		if (DUMP_FLAGS & REBOOT_DEVICE_FLAG) {
 			printf("NOTE: - CONFIG_BOARD_ASSERT_AUTORESET needs to be enabled to reboot TARGET Device after a crash\n");
 			printf("Handsake for rebooting device may fail since device reboots before sending acknowledgement\n");
+
 			/* Send handshake signal for rebooting TARGET device */
 			strncpy(handshake_string, TARGET_REBOOT_STRING, HANDSHAKE_STR_LEN_MAX);
 			if (do_handshake(dev_fd, handshake_string) != 0) {
@@ -346,13 +393,15 @@ int main(int argc, char *argv[])
 			break;
 		}
 
-		if (DUMP_FLAGS & EXIT_TOOL_FLAG) {
+		/* If no flag is set, then exit the dump tool */
+		if (DUMP_FLAGS == CLEAR_DUMP_FLAGS) {
 			break;
 		}
 
 		if (DUMP_FLAGS & RAMDUMP_FLAG) {
 			printf("DUMPING RAM CONTENTS\n");
 			fflush(stdout);
+
 			/* Handshake for synchronization before RAMDUMP */
 			strncpy(handshake_string, RAMDUMP_HANDSHAKE_STRING, HANDSHAKE_STR_LEN_MAX);
 			if (do_handshake(dev_fd, handshake_string) != 0) {
@@ -377,6 +426,7 @@ int main(int argc, char *argv[])
 		if (DUMP_FLAGS & USERFSDUMP_FLAG) {
 			printf("\nDUMPING USERFS CONTENTS\n");
 			fflush(stdout);
+
 			/* Handshake for synchronization before FS Dump */
 			strncpy(handshake_string, FSDUMP_HANDSHAKE_STRING, HANDSHAKE_STR_LEN_MAX);
 			if (do_handshake(dev_fd, handshake_string) != 0) {
@@ -390,6 +440,27 @@ int main(int argc, char *argv[])
 				goto dump_err;
 			}
 			printf("\nFilesystem Dump received successfully\n");
+			fflush(stdout);
+		}
+
+		if (DUMP_FLAGS & EXTUSERFS_DUMP_FLAG) {
+			printf("\nMake sure to enable BCH Driver for External UserFS Partition Dump\n");
+			printf("\nDUMPING EXTERNAL USERFS DUMP PARTITION\n");
+			fflush(stdout);
+
+			/* Handshake for synchronization before External Userfs Dump */
+			strncpy(handshake_string, EXTFSDUMP_HANDSHAKE_STRING, HANDSHAKE_STR_LEN_MAX);
+			if (do_handshake(dev_fd, handshake_string) != 0) {
+				printf("Target handshake failed\n");
+				goto handshake_err;
+			}
+
+			ret = extfsdump_recv(dev_fd);
+			if (ret != 0) {
+				printf("External Userfs partition dump reception failed, ret: %d\n", ret);
+				goto dump_err;
+			}
+			printf("\nExternal Userfs partition dump received successfully\n");
 			fflush(stdout);
 		}
 	}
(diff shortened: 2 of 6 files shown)


os/board/common/crashdump.c
```diff
@@ -129,6 +133,76 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
 	return 0;
 }
 
+static int get_ext_userfs_part(size_t *size, int *index)
+{
+#if defined CONFIG_ARCH_BOARD_HAVE_SECOND_FLASH && defined CONFIG_SECOND_FLASH_PARTITION && (CONFIG_SECOND_FLASH_PARTITION == 'y')
```
jeongchanKim, line 138: Please change like below
```
#if defined(XXX) && defined(YYY) && (ZZ == AA) ...
```

os/board/common/crashdump.c
```diff
@@ -238,6 +312,48 @@ static int send_fsdump(void)
...
+	}
+
+	/* Send external userfs partition size to host */
+	ptr = (uint8_t *)&size;
+	for (i = 0; i < sizeof(size); i++) {
+		up_lowputc((uint8_t)*ptr);
+		ptr++;
+	}
+
+	/* If extracted size is 0, it menas ther eis no partition available to dump */
```
jeongchanKim, line 339: Please fix `ther eis` to `there is`.

os/board/common/crashdump.c
```diff
@@ -238,6 +312,48 @@ static int send_fsdump(void)
...
+	uint8_t *ptr;
+	size_t size;
+	char ext_userfs_partition[EXTUSERFS_PARTITION_NAME_SIZE + 1] = { '\0' };
+
+	/*Send ACK for EXTERNAL FSDUMP Handshake */
+	up_lowputc('A');
+
+	i = get_ext_userfs_part(&size, &index);
+	if (i < 0) {
+		return -1;
```
jeongchanKim, line 329: Please use `ERROR` instead of `-1`.

os/board/common/crashdump.c
```diff
@@ -238,6 +312,48 @@ static int send_fsdump(void)
...
+	/* If extracted size is 0, it menas ther eis no partition available to dump */
+	if (size == 0) {
+		return 0;
+	}
+
+	snprintf(ext_userfs_partition, EXTUSERFS_PARTITION_NAME_SIZE, "/proc/fs/smartfs/smart1p%d/dump_ext", index);
+
+	fd = open(ext_userfs_partition, O_RDONLY);
+	if (fd < 0) {
+		return -EIO;
```
jeongchanKim, line 348: Is there any using this errno after calling send_extfsdump()? If not, I'm not sure what is the difference between returning ERROR and -errno.

os/board/common/crashdump.c
```diff
@@ -238,6 +312,48 @@ static int send_fsdump(void)
...
+	}
+
+	/* Send dummy read request, no data will actually be returned. Procfs entry automatically sends data over tty */
+	i = read(fd, NULL, 63);
```
jeongchanKim, line 352: If `i` is not used anymore, you don't need to assign `i` from `read`.
You can just call `(void)read(xxx);`

os/board/common/crashdump.c
amogh-samsung: Since the tty connection is established by the HOST machine, returned ERROR signals cannot be tracked.
I still kept different error signals for consistency and ease of understanding.

os/board/common/crashdump.c
amogh-samsung: Ahh ohkkay, Thankyou

os/fs/driver/mtd/smart.c
```diff
@@ -521,6 +525,28 @@ static const struct block_operations g_bops = {
  ****************************************************************************/
 
 #ifdef CONFIG_DEBUG_FS
+#ifndef NXFUSE_HOST_BUILD
+static int smart_dumpext(FAR struct smart_struct_s *dev)
+{
+	int ret = OK;
```
jeongchanKim, line 531: I think we don't need to initialize the `ret`. Because it is always set before using.

amogh-samsung: I follow this because I think it is a good practice to safeguard from garbage initialisation.
I'd like to keep this line same.

os/board/common/crashdump.c
Taejun-Kwon: returning errno is correct, I think

os/fs/driver/mtd/smart.c
```diff
@@ -521,6 +525,28 @@ static const struct block_operations g_bops = {
  ****************************************************************************/
 
 #ifdef CONFIG_DEBUG_FS
+#ifndef NXFUSE_HOST_BUILD
+static int smart_dumpext(FAR struct smart_struct_s *dev)
```
Taejun-Kwon, line 529: smart level doesn't know storage is internal or external.
Please make common name.

os/fs/driver/mtd/smart.c
```diff
@@ -521,6 +525,28 @@ static const struct block_operations g_bops = {
...
+	int i;
+	int j;
+
+	for (i = 0; i < (dev->geo.neraseblocks * dev->sectorsPerBlk); i++) {
+		ret = MTD_BREAD(dev->mtd, i * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)dev->rwbuffer);
+		if (ret != dev->mtdBlksPerSector) {
+			return -EIO;
+		}
+
+		for (j = 0; j < dev->sectorsize; j++) {
```
Taejun-Kwon, line 541: Too deep dependency here. It should work stand alone.
smart should return buffer only, then pro will get the buffer filled with data.
then dumper should write through the up_lowputc.

os/fs/driver/mtd/smart.c
```diff
@@ -4895,6 +4921,12 @@ static int smart_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 		case SMART_DEBUG_CMD_DUMP_PSECTOR:
 			ret = smart_dumpsector(dev, (uint16_t)debug_data->debugdata, debug_data->debugcmd);
 			goto ok_out;
+
+		case SMART_DEBUG_CMD_DUMP_EXT:
```
Taejun-Kwon, line 4925: It should be changed too.

Taejun-Kwon (commented):
Anyway, due to limitation of procfs (can't return read data of specific sector)
You'd better use mtd operation, not procfs.

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
...
+	i = 0;
+	while (second_flash_part_sizes[i] != '\0' && c) {
+		while (second_flash_part_sizes[i] != ',') {
+			i++;
+		}
+		c--;
+                i++;
+        }
+
+        if (second_flash_part_sizes[i] == '\0') {
```
jeongchanKim, line 188: Please remove space after tab.

tools/dump_tool/dump_tool.c
```diff
@@ -244,6 +244,50 @@ static int fsdump_recv(int dev_fd)
...
+                return -1;
+        }
+
+	if (ext_smartfs_size == 0) {
+		printf("No external userfs partition found. Please check config\n");
+		return 0;
+	}
+
+        printf("\n=========================================================================\n");
+        printf("External filesystem size = %d\n", ext_smartfs_size);
```
jeongchanKim, line 267: `ext_smartfs_size` is unsigned type, so please print its size with `%u` instead of `%d`.

tools/dump_tool/dump_tool.c
```diff
@@ -244,6 +244,50 @@ static int fsdump_recv(int dev_fd)
...
+        /* Receive Userfs flash partition size */
+        ret = b_read(dev_fd, (uint8_t *)&ext_smartfs_size, 4);
+        if (ret != 4) {
+                printf("Receiving Userfs size failed, ret = %d\n", ret);
+                return -1;
+        }
+
+	if (ext_smartfs_size == 0) {
+		printf("No external userfs partition found. Please check config\n");
+		return 0;
```
jeongchanKim, line 263: Do you have any special reason to return 0 for this failure case? If not, I think `-1` is better.

os/board/common/crashdump.c
```diff
@@ -298,6 +419,11 @@ static int dump_via_uart(void)
 				break;
 			}
 		/* Check handshake string for TARGET REBOOT signal */
+		} else if (strncmp(host_buf, EXTFSDUMP_HANDSHAKE_STRING, strlen(EXTFSDUMP_HANDSHAKE_STRING)) == 0) {
```
jeongchanKim, line 422: I think using `strlen(EXTFSDUMP_HANDSHAKE_STRING) + 1` is better for security issue.

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
...
+	}
+
+	*index = c;
+	i = 0;
+	while (second_flash_part_sizes[i] != '\0' && c) {
+		while (second_flash_part_sizes[i] != ',') {
+			i++;
+		}
+		c--;
+                i++;
```
deepaksrma, line 185: maintain the same alignment.

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
...
+        }
+
+        if (second_flash_part_sizes[i] == '\0') {
+		*size = 2000;
+		*index = 20;
+                return -1;
+	}
+
+	ext_userfs_size = 0;
+        while (second_flash_part_sizes[i] != ',' && second_flash_part_sizes[i] != '\0') {
```
deepaksrma, line 195: same like above.

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
...
+                ext_userfs_size *= 10;
+                ext_userfs_size += (second_flash_part_sizes[i] - 48);
+                i++;
+        }
+        ext_userfs_size *= 1024;
+	*size = ext_userfs_size;
```
deepaksrma, line 201: same.

tools/dump_tool/dump_tool.c
```diff
@@ -392,6 +439,25 @@ int main(int argc, char *argv[])
 			printf("\nFilesystem Dump received successfully\n");
 			fflush(stdout);
 		}
+
+		if (DUMP_FLAGS & EXTUSERFS_DUMP_FLAG) {
+			printf("\nDUMPING EXTERNAL USERFS DUMP PARTITION\n");
+			fflush(stdout);
+			/* Handshake for synchronization before External Userfs Dump */
```
deepaksrma, line 446: can you give a space below and up?
it will increase the readability for comment.

os/fs/smartfs/smartfs_procfs.c
```diff
@@ -988,6 +990,24 @@ static ssize_t smartfs_dump_psector(FAR struct file *filep, FAR const char *buff
 	/* Return 0 this doesn't bring any data */
 	return 0;
 }
+
+static size_t smartfs_dump_ext(FAR struct file *filep, FAR char *buffer, size_t buflen)
```
deepaksrma, line 994: can you provide some description for this function

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
 	return 0;
 }
 
+static int get_ext_userfs_part(size_t *size, int *index)
```
deepaksrma, line 136: please provide the description for all of your function.
it will be easier to understand the code for a new person.

os/board/common/crashdump.c
amogh-samsung: Done, thanks

os/board/common/crashdump.c
amogh-samsung: Done, thanks

os/board/common/crashdump.c
amogh-samsung: Done

os/board/common/crashdump.c
amogh-samsung: Done

tools/dump_tool/dump_tool.c
amogh-samsung: Updated

tools/dump_tool/dump_tool.c
amogh-samsung: In case no External Userfs partition is there, I'd like the user to still have the option to continue with the dump tool and dump other partitions.
Hence, not returning -1 to avoid exiting the Dump Tool

os/board/common/crashdump.c
amogh-samsung: The very same was added base don earlier suggestions but ended up causing problems when comparing handshake strings for Mt7686_hdk.
So reverted it to strlen(<MACRO>) only

os/board/common/crashdump.c
amogh-samsung: Updated

os/board/common/crashdump.c
amogh-samsung: Updated

os/fs/driver/mtd/smart.c
amogh-samsung: I was considering the same but had 1 issue
Getting only "n" bytes at a time in the buffer, we have to request many many times to get complete partition data (2Mb or more)
So it will take a lot of time.

Also, the buffer memory will have to be allocated after the crash, so is that safe?

os/board/common/crashdump.c
amogh-samsung: Done

os/fs/driver/mtd/smart.c
amogh-samsung: Updated to your previously used name
"smart_dumpall"

os/fs/driver/mtd/smart.c
amogh-samsung: Changed to SMART_DEBUG_CMD_DUMP_ALL

os/board/common/crashdump.c
amogh-samsung: Since this file doesn't have any other function descriptions either,
I could take this up separately if required.
@sunghan-chang is addition of descriptions required here?

tools/dump_tool/dump_tool.c
amogh-samsung: Added suitable spaces

os/fs/smartfs/smartfs_procfs.c
amogh-samsung: Added description

os/board/common/crashdump.c
```diff
@@ -201,7 +280,7 @@ static int send_ramdump(void)
 
 static int send_fsdump(void)
 {
-	int i;
+	int i = 0;
```
jeongchanKim, line 283: Is this initialization needed? `i` will be always set from `get_userfs_address_size()`.

amogh-samsung: I follow this as a general code writing practice to avoid garbage values.
If this doesn't have any disadvantages, I'd like to keep this please.
Line 322 has been written in the same way.

Taejun-Kwon:
If you really want to use procfs, at least below step should be considered.

precondition : Refactor SMART_DEBUG_CMD_DUMP_PSECTOR & SMART_DEBUG_CMD_DUMP_LSECTOR to get buffer data instead of print sector. print can be done in procfs layer.

1. Get geometry information with 'BIOC_GETPROCFSD'
you can get sector size and something more. but if journaling is enabled, it should use dev->get.neraseblocks instead of dev->neraseblocks
2. Now we know size of sector, we can get specific sector's data through the SMART_DEBUG_CMD_DUMP_PSECTOR in loop
```
for (int i = 0; i < procfs_data->xyz; i++) {
   ioctl();
}
```

os/board/common/crashdump.c
```diff
@@ -40,10 +42,12 @@
 #if defined(CONFIG_BOARD_DUMP_UART)
 #define RAMDUMP_HANDSHAKE_STRING	"TIZENRTRMDUMP"
 #define FSDUMP_HANDSHAKE_STRING	"TIZENRTFSDUMP"
+#define EXTFSDUMP_HANDSHAKE_STRING	"TIZENRTEXTFSD"
 #define TARGET_REBOOT_STRING		"TIZENRTREBOOT"
 #define HANDSHAKE_STR_LEN_MAX		(13)
 #define USERFS_PART_NAME		"user"
 #define USERFS_PART_NAME_LEN		(4)
+#define PROCFS_ENTRY_NAME_LEN		(40)
```
Taejun-Kwon, line 50: We already have a definition reagarding length of name in limit.h
CONFIG_NAME_MAX..

os/fs/driver/mtd/smart.c
```diff
@@ -521,6 +524,27 @@ static const struct block_operations g_bops = {
  ****************************************************************************/
 
 #ifdef CONFIG_DEBUG_FS
+static int smart_dumpall(FAR struct smart_struct_s *dev, FAR char* buffer)
+{
+	int ret = OK;
+	uint32_t sector = 0;
+
+	if (dump_flag) {
```
Taejun-Kwon, line 532: If you apply my last comment regarding sector size, you can remove dump_flag here.

os/fs/driver/mtd/smart.c
```diff
@@ -4895,6 +4919,10 @@ static int smart_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 		case SMART_DEBUG_CMD_DUMP_PSECTOR:
 			ret = smart_dumpsector(dev, (uint16_t)debug_data->debugdata, debug_data->debugcmd);
 			goto ok_out;
+
+		case SMART_DEBUG_CMD_DUMP_ALL:
```
Taejun-Kwon, line 4923: this command doesn't bring all sectors just one time...
SMART_DEBUG_COPY_PSECTOR or something else to explain that is copy specific sector data from flash..

os/board/common/crashdump.c
```diff
@@ -129,6 +133,81 @@ static int get_userfs_address_size(uint32_t *userfs_start, size_t *userfs_size)
...
+#if defined(CONFIG_SECOND_FLASH_PART_NAME) && defined(CONFIG_SECOND_FLASH_PART_SIZE)
+	int i = 0;
+	int c = 0;
+	int x = 0;
+	size_t ext_userfs_size;
+	char second_flash_part_list[] = CONFIG_SECOND_FLASH_PART_NAME;
+	char second_flash_part_sizes[] = CONFIG_SECOND_FLASH_PART_SIZE;
+	char tmp_part_name[USERFS_PART_NAME_LEN + 1] = { '0' };
+
+	while (second_flash_part_list[i] != '\0') {
```
Taejun-Kwon, line 147: some of this is duplicated with code in partitons.c , I guess...
so can't we reuse it if we add signature of those functions to partition.h ??

amogh-samsung:
> If you really want to use procfs, at least below step should be considered.
> 
> precondition : Refactor SMART_DEBUG_CMD_DUMP_PSECTOR & SMART_DEBUG_CMD_DUMP_LSECTOR to get buffer data instead of print sector. print can be done in procfs layer.
> 
> 1. Get geometry information with 'BIOC_GETPROCFSD'
>    you can get sector size and something more. but if journaling is enabled, it should use dev->get.neraseblocks instead of dev->neraseblocks
> 2. Now we know size of sector, we can get specific sector's data through the SMART_DEBUG_CMD_DUMP_PSECTOR in loop
> 
> ```
> for (int i = 0; i < procfs_data->xyz; i++) {
>    ioctl();
> }
> ```

Procfs supports 2 functionalities - **dump_lsector/psector** and **dump for External Userfs Dump**
Where one uses the 'read' API and other uses 'write' API.
For both to be clubbed, changes would be required to the `struct mtd_smart_debug_data_s` as we would need to pass all 3
1. the mtd device object,
2. buffer for returning data, and
3. sector number.

os/board/common/crashdump.c
amogh-samsung: CONFIG_NAME_MAX often defined as 32 is barely enough for this procfs entry
`/proc/fs/smartfs/smart1p1/dump_all` -> 34 characters (35 including '\0' for snprintf)
`/proc/fs/smartfs/smart0p8/dump_lsector` -> 38 characters

os/fs/driver/mtd/smart.c
amogh-samsung: Lets rename this to `SMART_DEBUG_CMD_RETRIEVE_SECTOR` as it retrieves the data of one sector at a time

os/board/common/crashdump.c
amogh-samsung: The common part of code is one function as follows: -
```
static void move_to_next_part(const char **par)
{
                /* Move to next part information. */
                while (*(*par)++ != ',');
}
```
which parses the partition names/list till the next comma is encountered.

This can be made to fit with the requirement here.
However in addition to the new function `get_ext_userfs_part`, another pre-existing function `get_userfs_address_size` will make use of this.

So, I will make the changes for both and optimize in an upcoming commit.
Thankyou very much for pointing this out.

os/board/common/crashdump.c
Taejun-Kwon: No, there's limit value about file name & path name.
So this is unncessary.
please see limit.h one more time.

os/fs/smartfs/smartfs_procfs.c
```diff
@@ -988,6 +990,32 @@ static ssize_t smartfs_dump_psector(FAR struct file *filep, FAR const char *buff
...
 }
+
+/****************************************************************************
+ * Name: smartfs_dump_all
+ *
+ * Description: Performs the dump of the entire userfs partition in flash
+ *
+ ****************************************************************************/
+
+static size_t smartfs_dump_all(FAR struct file *filep, FAR char *buffer, size_t buflen)
```
Taejun-Kwon, line 1001: This doesn't dump all, renaming is needed.

os/fs/driver/mtd/smart.c
Taejun-Kwon: Why resolved this without any comment?
dump_flag is not necessary if we know about sectorsize, and we can get it through the procfs.
Anyway I think dump flag is not unsafe for multi thread... make sure that this should work with another module, not only with dump tool.

os/board/common/crashdump.c
amogh-samsung: Yes,
CONFIG_NAME_MAX doesn't suffice,
But CONFIG_PATH_MAX will, lets use that

os/fs/smartfs/smartfs_procfs.c
amogh-samsung: Function names changed to `smartfs_retrieve_sector_data` to align them with the command name `SMART_DEBUG_CMD_RETRIEVE_PSECTOR`
Procfs entry name retained as `dump_all` since this name signifies that this entry will be used to get the entire partition's dump

Taejun-Kwon:
Could you please check below step?
1. open '/dev/smart%d%sd%d'
2. call ioctl with defined constants what you added.
3. If smart_ioctl called and it works or not.
This approach is control block driver  drirectly.

os/fs/smartfs/smartfs_procfs.c
amogh-samsung: Procfs entries combined for optimization

os/fs/driver/mtd/smart.c
amogh-samsung: Done by mistake, sorry.
Updated now.

os/board/common/crashdump.c
```diff
@@ -238,6 +317,64 @@ static int send_fsdump(void)
...
+	fd = open(ext_userfs_partition, O_RDONLY);
+	if (fd < 0) {
+		return errno;
+	}
+
+	/* Send dummy read request, no data will actually be returned. Procfs entry automatically sends data over tty */
+	while (size) {
+		ext_fsdump_buf[1] = (uint8_t)count;
+		ext_fsdump_buf[0] = (uint8_t)(count >> 8); //Buffer now holds the sector number whose data is to be retrieved.
+		i = read(fd, ext_fsdump_buf, CONFIG_MTD_SMART_SECTOR_SIZE);
```
Taejun-Kwon, line 361: Actually sector size is decided after read header of mtd sector, if it is empty, then we use this config.
We have a TODO list there, please see TODO at the start of smart_scan in smart.c
So here also TODO need to be added.

os/fs/driver/mtd/smart.c
```diff
@@ -513,14 +513,41 @@ static const struct block_operations g_bops = {
...
+ * Description: Write specific sector data to buffer and send back.
+ *
+ ****************************************************************************/
+
+static int smart_retrieve_sector(FAR struct smart_struct_s *dev, FAR char* buffer)
+{
+	int ret = OK;
+	uint16_t sector = 0;
+
+	sector = (uint8_t)buffer[0];
```
Taejun-Kwon, line 529: UINT8TOUINT16(buffer);

os/include/tinyara/fs/smart_procfs.h
```diff
@@ -70,7 +70,7 @@
 #define SMART_DEBUG_CMD_SHOW_LOGMAP       2
 #define SMART_DEBUG_CMD_DUMP_PSECTOR      3
 #define SMART_DEBUG_CMD_DUMP_LSECTOR      4
-#define SMART_DEBUG_DUMP_ALL              0xFFFF
+#define SMART_DEBUG_CMD_RETRIEVE_PSECTOR  5
```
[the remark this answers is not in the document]
amogh-samsung: Get is simple and short, I agree...
However, get would seem like "getting information _about_ the psector", not the sector data itself.
And, _retrieve_ would mean "getting back the data of the psector" which we are actually doing by requesting MTD driver.

os/board/common/crashdump.c
amogh-samsung: Done

os/fs/driver/mtd/smart.c
amogh-samsung: Updated, thankyou

os/board/common/crashdump.c
```diff
@@ -238,6 +317,68 @@ static int send_fsdump(void)
...
+		return 0;
+	}
+
+	snprintf(ext_userfs_partition, CONFIG_PATH_MAX, "/proc/fs/smartfs/smart1p%d/dump_psector", index);
+	fd = open(ext_userfs_partition, O_RDONLY);
+	if (fd < 0) {
+		return errno;
+	}
+
+	/* Send dummy read request, no data will actually be returned. Procfs entry automatically sends data over tty */
```
Taejun-Kwon, line 357: Think this comment is not correct now.

os/fs/driver/mtd/smart.c
```diff
@@ -513,14 +513,41 @@ static const struct block_operations g_bops = {
...
+ ****************************************************************************/
+
+static int smart_retrieve_sector(FAR struct smart_struct_s *dev, FAR char* buffer)
+{
+	int ret = OK;
+	uint16_t sector = 0;
+
+	sector = UINT8TOUINT16(buffer[0]);
+	sector = (sector << 8);
+	sector += UINT8TOUINT16(buffer[1]);
```
Taejun-Kwon, line 531: sector = UINT8TOUINT16(buffer);

os/fs/driver/mtd/smart.c
```diff
@@ -513,14 +513,41 @@ static const struct block_operations g_bops = {
...
+	if (sector >= (dev->geo.neraseblocks * dev->sectorsPerBlk)) {
+		return -EINVAL;
+	}
+
+	ret = MTD_BREAD(dev->mtd, sector * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)buffer);
+	if (ret != dev->mtdBlksPerSector) {
+		return -EIO;
+	}
+
+	return CONFIG_MTD_SMART_SECTOR_SIZE;
```
Taejun-Kwon, line 541: This value is not related to above MTD_BREAD, it had better return dev->mtdBlkPerSector * dev->sectorsize
anyway it is aligned to Todo in line 362 of crashdump.c
And please add more info to description regarding return value.

amogh-samsung: Updated

os/board/common/crashdump.c
amogh-samsung: Ahh yes
Updated

os/fs/driver/mtd/smart.c
amogh-samsung: Updated

os/board/common/crashdump.c
```diff
@@ -238,6 +317,63 @@ static int send_fsdump(void)
...
+		up_lowputc((uint8_t)*ptr);
+		ptr++;
+	}
+
+	/* If extracted size is 0, it means there is no partition available to dump */
+	if (size == 0) {
+		return 0;
+	}
+
+	snprintf(ext_userfs_partition, CONFIG_PATH_MAX, "/dev/smart1p%d", index);
```
Taejun-Kwon, line 350: hmm...it should be handled by config value.
Let's add TOTO here and fix it when you have some times.

amogh-samsung: Updated

Taejun-Kwon (approved):

PR #4643, opened 2020-12-03
mm/mm_heap/mm_realloc : Add logic for checking overflow of size

jeongchanKim:
MM_ALIGN_UP do an unsafe calculation that can cause integer wrap-around.
For avoiding integer wrap-around, check the size whether it is smaller than possible allocation size.

### DIFF
--- a/os/mm/mm_heap/mm_initialize.c
+++ b/os/mm/mm_heap/mm_initialize.c
@@ -127,6 +127,20 @@ void mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
 
 	heapbase = MM_ALIGN_UP((uintptr_t)heapstart);
 	heapend  = MM_ALIGN_DOWN((uintptr_t)heapstart + (uintptr_t)heapsize);
+	if ((heapbase < (uintptr_t)heapstart)) {
+		/* heapbase cannot be smaller than heapstart.
+		 * If this happens, heapbase can be overflowed from alignment.
+		 */
+		return;
+	}
+
+	if (heapbase >= (uintptr_t)heapend) {
+		/* heapbase is aligned up from heapstart,
+		 * and heapend is the summation of heapstart and heapsize.
+		 * If heapsize is very small, align-up address can be equal to or greater than heapend.
+		 */
+		return;
+	}
 	heapsize = heapend - heapbase;
 
 	mlldbg("Region %d: base=%p size=%u\n", IDX + 1, heapstart, heapsize);
--- a/os/mm/mm_heap/mm_realloc.c
+++ b/os/mm/mm_heap/mm_realloc.c
@@ -136,6 +136,13 @@ FAR void *mm_realloc(FAR struct mm_heap_s *heap, FAR void *oldmem, size_t size)
 		return NULL;
 	}
 
+	if (size > MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE) {
+		mdbg("Because of mm_allocnode, %u cannot be allocated. The maximum \
+			 allocable size is (MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE) \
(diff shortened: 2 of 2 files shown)


os/mm/mm_heap/mm_initialize.c
```diff
@@ -127,6 +127,13 @@ void mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
 
 	heapbase = MM_ALIGN_UP((uintptr_t)heapstart);
 	heapend  = MM_ALIGN_DOWN((uintptr_t)heapstart + (uintptr_t)heapsize);
```
davidfather, line 129: At line 111, heapsize should be checked whether it is non-zero positive values, for example, if (heapsize <= 0) return;

jeongchanKim: Modified. Thanks.

os/mm/mm_heap/mm_realloc.c
```diff
@@ -136,6 +136,13 @@ FAR void *mm_realloc(FAR struct mm_heap_s *heap, FAR void *oldmem, size_t size)
 		return NULL;
 	}
 
+	if (size > MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE) {
+		mdbg("Because of mm_allocnode, %u cannot be allocated. The maximum \
+			 allocable size is (MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE) \
+			 : %u\n.", size, (MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE));
+		return NULL;
```
davidfather, line 143: Why don't you free 'oldmem' before return?

jeongchanKim: I think `free` is not needed, because 
1) if `oldmem` allocates well, then already returned from line 126 or 128.
2) even if `oldmem` is not null, we don't need to free the original memory. free is only working when requested size is 0.(line 134)

davidfather: Abnormal requests need to be treated with some penalties.
Anyhow, you should describe how to treat this case.

davidfather: Please, check POSIX. We need to follow the standard.

jeongchanKim: There is a guide in POSIX documents like below.
```
If the space cannot be allocated, the object shall remain unchanged.
```
So I think we don't need to free the original.
And I think, regardless of POSIX description, `realloc` is for re-allocating new memory with new size. If we cannot allocate for any reason, we should not spoil the original memory.

os/mm/mm_heap/mm_initialize.c
jeongchanKim: I reverted. Because `heapsize` is `size_t` type, and `size_t` is unsigned int. So it cannot be negative.

sunghan-chang (dismissed):

os/mm/mm_heap/mm_initialize.c
```diff
@@ -127,6 +127,21 @@ void mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
...
+		 * If this happens, heapbase can be overflowed from alignment.
+		 */
+		return;
+	}
+
+	if (heapbase >= (uintptr_t)heapend) {
+		/* heapbase is aligned up from heapstart,
+		 * and heapend is the summation of heapstart and heapsize.
+		 * If heapsize is very small, align-up address can be equal to or greater than heapend.
+		 * In this case, heapbase cannot be the base address of heap.
```
sunghan-chang, line 141: This line has little bit different meaning against the statement.
This statement means there is no meaningful `size` for new heap region.
But this line says wrong `base address` of heap.

sunghan-chang (approved):

davidfather (approved):
Looks good to me.

PR #4644, opened 2020-12-04
arch/amebad/flash : Apply bytewrite function

Taejun-Kwon:
RTL8721x didn't implement bytewrite, so add it to increase performance

### DIFF
--- a/os/arch/arm/src/amebad/amebad_flash.c
+++ b/os/arch/arm/src/amebad/amebad_flash.c
@@ -267,12 +267,18 @@ static ssize_t amebad_read(FAR struct mtd_dev_s *dev, off_t offset, size_t nbyte
  * Name: amebad_write
  ************************************************************************************/
 
-#if defined(CONFIG_MTD_BYTE_WRITE)
+#ifdef CONFIG_MTD_BYTE_WRITE
 static ssize_t amebad_write(FAR struct mtd_dev_s *dev, off_t offset, size_t nbytes, FAR const uint8_t *buffer)
 {
 	FAR struct amebad_dev_s *priv = (FAR struct mtd_dev_s *)dev;
+	size_t addr;
+	ssize_t result;
+	addr = CONFIG_AMEBAD_FLASH_BASE + offset;
+	result = amebad_flash_write(addr, buffer, nbytes);
+	return result < 0 ? result : nbytes;
 }
 #endif
+
 
 /************************************************************************************
  * Name: amebad_ioctl


os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -267,13 +267,19 @@ static ssize_t amebad_read(FAR struct mtd_dev_s *dev, off_t offset, size_t nbyte
...
-#if defined(CONFIG_MTD_BYTE_WRITE)
+#ifdef CONFIG_MTD_BYTE_WRITE
 static ssize_t amebad_write(FAR struct mtd_dev_s *dev, off_t offset, size_t nbytes, FAR const uint8_t *buffer)
 {
 	FAR struct amebad_dev_s *priv = (FAR struct mtd_dev_s *)dev;
+	size_t addr;
+	ssize_t result;
+	addr = CONFIG_AMEBAD_FLASH_BASE + offset;
+	result = amebad_flash_write(addr, buffer, nbytes);
+	return result < 0 ? result : nbytes;
```
jeongchanKim, line 278: `nbytes` is unsigned type, `size_t`. Is it okay not to make type casting?

Taejun-Kwon: 1. it should return one of errno or written bytes. amebad_flash_write handles it.
2. ssize_t huge enough for TizenRT as a buffer. (I think enough for window or linux)
3. that is one of posix style(https://linux.die.net/man/3/write)

sunghan-chang:
@fengyunlai @ziyik @AdamZhang0124 Could you check this? Is this ok to support byte-write on the flash?

AdamZhang0124:
Hi @Taejun-Kwon,
Seems the bytewrite is applied with the same function as amebad_bwrite is using. I am not sure if it can increase the performance.  Could you please share your test result on the performance improvement? I am also testing it on my side, thanks!

Taejun-Kwon:
> Hi @Taejun-Kwon,
> Seems the bytewrite is applied with the same function as amebad_bwrite is using. I am not sure if it can increase the performance. Could you please share your test result on the performance improvement? I am also testing it on my side, thanks!
> 
> TJ : It is always faster than disabled situation because smartfs write 1 byte to change status of logical sector. When it disabled, it always write entire of MTD sector, that is 4 times of page-write. But with byte write, mostly it write one page. so Much faster .
> As I know you have test app already, you can compare it.

sunghan-chang:
Let's merge this first. When we find an issue, let's revert both this pr and defconfig pr.

sunghan-chang (approved):

AdamZhang0124:
> > Hi @Taejun-Kwon,
> > Seems the bytewrite is applied with the same function as amebad_bwrite is using. I am not sure if it can increase the performance. Could you please share your test result on the performance improvement? I am also testing it on my side, thanks!
> 
> It is always faster than disabled situation because smartfs write 1 byte to change status of logical sector. When it disabled, it always write entire of MTD sector, that is 4 times of page-write. But with byte write, mostly it write one page. so Much faster .
> As I know you have test app already, you can compare it.

Hi @Taejun-Kwon , I have tested the writing speed with the bytewrite enabled. From my test result, the speed does not have obvious improvement comparing to the disabled situation.
fs_test(bytewrite_on)
Is there any other MTD configurations to be done besides enabling bytewrite? Could you please check on this also? Thanks!
menuconfig

Taejun-Kwon:
> > > Hi @Taejun-Kwon,
> > > Seems the bytewrite is applied with the same function as amebad_bwrite is using. I am not sure if it can increase the performance. Could you please share your test result on the performance improvement? I am also testing it on my side, thanks!
> > 
> > 
> > It is always faster than disabled situation because smartfs write 1 byte to change status of logical sector. When it disabled, it always write entire of MTD sector, that is 4 times of page-write. But with byte write, mostly it write one page. so Much faster .
> > As I know you have test app already, you can compare it.
> 
> Hi @Taejun-Kwon , I have tested the writing speed with the bytewrite enabled. From my test result, the speed does not have obvious improvement comparing to the disabled situation.
> fs_test(bytewrite_on)
> Is there any other MTD configurations to be done besides enabling bytewrite? Could you please check on this also? Thanks!
> menuconfig


Actually when any of sector which is already commited becomes to release state, it should change 1bit to change state.
So if byte write is enabled, it just write 1byte.
Otherwise, it should follow below logic.
1. read whole of logical sector(current config set as 1k)
2. change offset of state in buffer.
3. write entire of logical sector

So with bytewrite it just write 1byte but without bytewrite it write 4pages(256b x 4).
If it doesn't improve performance, you should check your flash driver first.
Is it really write 1byte?



smartfs

PR #4645, opened 2020-12-04
configs/rtl8721csm/loadable_apps/defconfig:  Enable configs related to journaling

Taejun-Kwon:
Loadable config should use journaling to garuantees integrity of binary.
So add configs


sunghan-chang:
@Taejun-Kwon Could you write readable path on the title?
```
configs/rtl8721csm/loadable_apps/defconfig: enable journaling
```

sunghan-chang (approved):

PR #4646, opened 2020-12-04
Wifi on off

pillip8282:


framework/src/wifi_manager/wifi_manager_api.c
```diff
@@ -100,7 +100,7 @@ wifi_manager_result_e wifi_manager_init(wifi_manager_cb_s *wmcb)
 		return WIFI_MANAGER_INVALID_ARGS;
 	}
 
-	wifimgr_msg_s msg = {EVT_INIT_CMD, (void *)wmcb, NULL};
+	wifimgr_msg_s msg = {EVT_INIT_CMD, WIFI_MANAGER_FAIL, (void *)wmcb, NULL};
```
davidfather, line 103: The second parameter is the result whether the Wi-Fi manager successfully performs the request, which will be delivered asynchronously.
Then, what is the return value of this function? Is it just a result whether the request is successfully delivered to the Wi-Fi manager?
Then, we need to clarify this difference to improve API usability.

pillip8282: The return value of wifimgr_post_message is for checking that the msg is passed to wifi msg handler thread well. however this return value and second parameter are not separated well. so I fixed it.

framework/src/wifi_manager/wifi_manager_api.c
```diff
@@ -100,24 +109,25 @@ wifi_manager_result_e wifi_manager_init(wifi_manager_cb_s *wmcb)
...
 
 wifi_manager_result_e wifi_manager_deinit(void)
 {
 	sem_t signal;
 	sem_init(&signal, 0, 0);
-	wifimgr_msg_s msg = {EVT_DEINIT_CMD, NULL, &signal};
+	wifimgr_msg_s msg = {EVT_DEINIT_CMD, WIFI_MANAGER_FAIL, NULL, &signal};
 
-	wifi_manager_result_e res = wifimgr_post_message(&msg);
+	int res = wifimgr_post_message(&msg);
```
davidfather, line 125: If 'res' is less than zero, do you need to still wait 'msg.signal'?

pillip8282: if 'res' is less than zero, then it won't wait 'msg.signal'

pillip8282: no it doesn't need to wait signal. I'll fix it

framework/src/wifi_manager/wifi_manager_api.c
```diff
@@ -135,79 +152,73 @@ wifi_manager_result_e wifi_manager_set_mode(wifi_manager_mode_e mode, wifi_manag
...
 	sem_init(&signal, 0, 0);
-	wifimgr_msg_s msg = {EVT_SET_STA_CMD, NULL, &signal};
+	wifimgr_msg_s msg = {EVT_SET_STA_CMD, WIFI_MANAGER_FAIL, NULL, &signal};
 	if (mode == SOFTAP_MODE) {
 		msg.event = EVT_SET_SOFTAP_CMD;
 		msg.param = (void *)config;
 	}
 
-	wifi_manager_result_e res = wifimgr_post_message(&msg);
+	int res = wifimgr_post_message(&msg);
```
davidfather, line 161: Here the same issue mentioned above.
If 'res' is less than zero, do you need to still wait 'msg.signal'?

davidfather (approved):
Looks good to me.

sangwon03 (approved):
LGTM

PR #4647, opened 2020-12-04
os/board/rtl8721csm :  Fix MPU setting error in rtk board

kishore-sn:
MPU region 0 usage depends on some conditions in code. Sometimes it is not used,
but the region number is allocated for it. So region 0 goes waste. So, we move
the region number allocation into the if condition.

Also, we were using a hardcoded value of 1 for mpu_set_nregion_board_specific().
If rtk uses more than 1 region, then this will result in overwriting the board
specific region by the app region. So, we modify the hardcoded value and instead
use the current region number setting from the rtk board for reserving the board
mpu regions.

### DIFF
--- a/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
+++ b/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
@@ -31,6 +31,7 @@ SECTION(".data") u8* __ram_nocache_end__ = 0;
 SECTION(".data") u8* __psram_bss_start__ = 0;
 SECTION(".data") u8* __psram_bss_end__ = 0;
 #endif
+static u32 g_mpu_nregion_allocated;
 extern int main(void);
 extern u32 GlobalDebugEnable;
 struct _driver_call_os_func_map driver_call_os_func_map;
@@ -945,7 +946,6 @@ u32 app_mpu_nocache_init(void)
 	/* It is for wifi related buffer area, to prevent non-sync from happening.
 	*  Because DMA is involved, so it cannot be used together with cache.
 	*/
-	mpu_entry = mpu_entry_alloc();
 	if (wifi_config.km4_cache_enable) {
 		mpu_cfg.region_base = (uint32_t)__ram_nocache_start__;
 		mpu_cfg.region_size = __ram_nocache_end__-__ram_nocache_start__;
@@ -958,6 +958,7 @@ u32 app_mpu_nocache_init(void)
 	mpu_cfg.sh = MPU_NON_SHAREABLE;
 	mpu_cfg.attr_idx = MPU_MEM_ATTR_IDX_NC;
 	if (mpu_cfg.region_size >= 32) {
+		mpu_entry = mpu_entry_alloc();
 		mpu_region_cfg(mpu_entry, &mpu_cfg);
 	}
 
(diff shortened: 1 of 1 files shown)


os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1350,7 +1350,7 @@ extern void __libc_init_array(void);
 #ifdef CONFIG_PLATFORM_TIZENRT_OS
 #ifdef CONFIG_ARMV8M_MPU
 	/* Initialize number of mpu regions for board specific purpose */
-	mpu_set_nregion_board_specific(1);
+	mpu_set_nregion_board_specific(mpu_entry_alloc());
```
sangam-swami, line 1353: does mpu_entry_alloc() reserve one more region here?
because for region 0 it was mentioned entry_alloc reserve the region

kishore-sn: Every time we call entry_alloc(), it will reserve one region internally. 
For example if rtk is using below regions, 
xxx = entry_alloc() ==> region 0
psram = entry_alloc() ==> region 1

So, when we call entry_alloc() again, it returns value 2, which is correct for our mpu_set_nregion_board_specific() API.
After this point, realtek code is not allocating more regions.

sangam-swami: yes correct, to know the already allocated number we using secure allocation call.  The function name and the purpose we use seems mismatching. 
Can we get mpu_entry from app_mpu_nocache_init and pass here?

kishore-sn: Ok. I modified it to use a global variable.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -31,6 +31,7 @@ SECTION(".data") u8* __ram_nocache_end__ = 0;
 SECTION(".data") u8* __psram_bss_start__ = 0;
 SECTION(".data") u8* __psram_bss_end__ = 0;
 #endif
+u32 g_mpu_last_entry;
```
jeongchanKim, line 34: If this variable is used only in this file, please add static.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1010,6 +1011,7 @@ u32 app_mpu_nocache_init(void)
 	mpu_cfg.attr_idx = MPU_MEM_ATTR_IDX_WB_T_RWA;
 	mpu_region_cfg(mpu_entry, &mpu_cfg);
 
+	g_mpu_last_entry = mpu_entry + 1;
```
sunghan-chang, line 1014: This is not the last entry. It's (last entry + 1). Let's name it as `g_mpu_nregion_allocated`

sunghan-chang (approved):

PR #4651, opened 2020-12-08
arch/arm/common/up_mpuinit.c : Fix type casting and Add ifdef conditi…

jeongchanKim:
…on for mpuinitialize

1) mpu->nregion_board_specific is uint8_t type, so don't need to type cating to struct mpu_usages_s *.
2) up_mpuinitialize is only working when CONFIG_BUILD_PROTECTED.

### DIFF
--- a/os/arch/arm/src/common/up_mpuinit.c
+++ b/os/arch/arm/src/common/up_mpuinit.c
@@ -73,7 +73,7 @@ static struct mpu_usages_s g_mpu_region_usages;
 
 void mpu_region_initialize(struct mpu_usages_s *mpu)
 {
-	uint8_t offset = (struct mpu_usages_s *)mpu->nregion_board_specific;
+	uint8_t offset = mpu->nregion_board_specific;
 
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	offset += MPU_NUM_REGIONS;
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/common/up_mpuinit.c
```diff
@@ -149,7 +149,7 @@ void mpu_set_nregion_board_specific(uint8_t num)
  *   resources.
  *
  ****************************************************************************/
-
+#ifdef CONFIG_BUILD_PROTECTED
```
sunghan-chang, line 152: Let's remove this from the header.

davidfather (approved):
Looks good to me.

os/arch/arm/src/common/up_mpuinit.h
```diff
@@ -54,11 +54,10 @@
...
  *
  ****************************************************************************/
-
-#ifdef CONFIG_BUILD_PROTECTED
+#ifdef CONFIG_ARM_MPU  
 void up_mpuinitialize(void);
-#else
-#  define up_mpuinitialize()
+#else    
+#  define up_mpuinitialize()
```
sunghan-chang, line 60: Could you remove unnecessary spaces?

sunghan-chang (approved):

PR #4652, opened 2020-12-08
network/libc: fix getaddrinfo memory leak.

pillip8282:
netdev_ioctl() converts lwIP data type to socket data type to pass the
result of getaddrinfo(). but it doesn't free lwip data type moreover doesn't free it's member data.


davidfather (approved):
Looks good to m.e

sangwon03 (approved):
LGTM

PR #4653, opened 2020-12-08
Wifi pf common library support

kishore-sn:
Wifi PF library will be included in common binary. 
This PR provides a format for implementing the platform library and tests the calls to library function from the main app.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_data.c
+++ b/os/kernel/binary_manager/binary_manager_data.c
@@ -191,6 +191,13 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
 {
 	int bin_idx;
 
+	DEBUGASSERT(tcb);
+
+	if (!tcb->group) {
+		bmdbg("Failed to add pid %d to binlist. This task has already exited and group is NULL\n", tcb->pid);
+		return;
+	}
+
 	bin_idx = tcb->group->tg_binidx;
 
 	/* A binary index, bin_idx is greater than 0 only if tcb is a thread of user binary.


loadable_apps/loadable_sample/Kconfig
```diff
@@ -77,6 +77,13 @@ config EXAMPLES_MICOM_TIMER_INTERVAL
 		Interval is based on microsecond.
 endif
 
+config WIFI_PF
```
jeongchanKim, line 80: I think we'd better rename the config name for clarity.

loadable_apps/loadable_sample/Kconfig
```diff
@@ -77,6 +77,13 @@ config EXAMPLES_MICOM_TIMER_INTERVAL
...
+	bool "Support for Wifi platform library"
+	default n
+	depends on SUPPORT_COMMON_BINARY
+	---help---
+		Support for Wifi platform library.
```
jeongchanKim, line 85: Please add description more about what is platform library.

loadable_apps/loadable_sample/Kconfig
jeongchanKim, line 80: Why do you put this config in `Enable Test Scenarios` menu?
I think this config can be outside of test scenarios.

loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <tinyara/config.h>
```
jeongchanKim, line 22: This include is not needed.

loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,33 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
```
thapav, line 21: Please remove extra lines

loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,33 @@
...
+#include <tinyara/config.h>
+#include <stdio.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
```
thapav, line 27: Same as above

loadable_apps/loadable_sample/wifiapp/Makefile
```diff
@@ -62,6 +62,9 @@ endif
 ifneq ($(filter y,$(CONFIG_EXAMPLES_BINARY_UPDATE_TEST) $(CONFIG_EXAMPLES_UPDATE_AGING_TEST)),)
 SRCS += binary_update.c
 endif
+ifeq ($(CONFIG_WIFI_PLATFORM_LIB),y)
+LIBSRCS += wifi_pf.c
```
jeongchanKim, line 66: I think `wifi_pflib.c` or `wifi_pf_lib.c` is better.

loadable_apps/loadable_sample/Kconfig
```diff
@@ -107,4 +107,15 @@ config EXAMPLES_MICOM_TIMER_INTERVAL
...
 endif #EXAMPLES_LOADABLE_AUTOMATIC_TEST
+
+config WIFI_PLATFORM_LIB
+	bool "Support for Wifi platform library"
+	default n
+	depends on SUPPORT_COMMON_BINARY
+	---help---
+		Support for Wifi platform library. Wifi platform library
+		will contain wifi platform specific code and will included
+		in the common binary. This is done to allow us to perform
```
jeongchanKim, line 118: wifi app binary and wifi platform lib binary has dependency for update. Isn't it?

loadable_apps/loadable_sample/wifiapp/Makefile
kishore-sn: I named it wifi_pf because it can be part of library or main wifi binary. 
Just wanted to differentiate between wifi app code and pf code. 
Pls let me know what you think.

loadable_apps/loadable_sample/Kconfig
kishore-sn: Yes. There might be a dependency. 
But this config is just to enable the wifi pf library support. Else all code of wifi app and pf will be built as a single binary. 
Do you think we need to add a dependency here?

loadable_apps/loadable_sample/wifiapp/Makefile
jeongchanKim: Okay. I agree your opinion. Let's keep it.

loadable_apps/loadable_sample/Kconfig
jeongchanKim: No. I mean the explanation can be confusing. Because explanation said we can update the wifi app and pf independently.

kishore-sn: Oh ok. 
But we have made a separate library for wifi pf because we would like to update it separately. 
When we update wifi pf library, the app also needs to be reloaded, because the symbol addresses in the pf will change. 
But, the app does not need to be updated always. An app update is required only if some interface change happens.

jeongchanKim: Yes, I know. Could you add the description about this also? Then it is much better to understand.

loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,31 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
```
pillip8282, line 21: is it ok not to include tinyara/config.h ?

kishore-sn:
The PR got overwritten by mistake. 
I will open new PR again.

PR #4655, opened 2020-12-09
board/rtl8721csm: Change key_id used for WEP shared key auth method to 0

fengyunlai:
Change key_id from 1 to 0 for WEP shared key auth method connection according to RTK default SDK setting, also to fix connection issue found in compatibility test.

### DIFF
--- a/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
+++ b/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
@@ -445,7 +445,7 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 		password = ap_connect_config->passphrase;
 		ssid_len = strlen((const char *)ssid);
 		password_len = ap_connect_config->passphrase_length;
-		key_id = 1; //Foucus
+		key_id = 0;
 		if ((password_len != 5) && (password_len != 13)) {
 			ndbg("\n\rWrong WEP key length. Must be 5 or 13 ASCII characters.");
 			return -1;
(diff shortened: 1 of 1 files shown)


os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -445,7 +445,7 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 		password = ap_connect_config->passphrase;
 		ssid_len = strlen((const char *)ssid);
 		password_len = ap_connect_config->passphrase_length;
-		key_id = 1; //Foucus
+		key_id = 0;
```
sangam-swami, line 448: you can initialize with zero at declaration, as nowhere it is getting modified or assigned different values

jeongchanKim, line 448: I agreed Sangam's opinion. And in this function(`cmd_wifi_connect()`), all `key_id` are used as 0, so it is better to initialize at the first of the function.

jeongchanKim (approved):
LGTM

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
fengyunlai: I agree with you, thanks

fengyunlai: I agree with you, thanks

sunghan-chang (approved):

PR #4659, opened 2020-12-10
kernel : Remove unnecessary variable in tcb - ram_start

jeongchanKim:
ram_start in tcb is not needed anymore, because we allocate the loading app with its own way.

### DIFF
--- a/os/include/tinyara/sched.h
+++ b/os/include/tinyara/sched.h
@@ -620,8 +620,6 @@ struct tcb_s {
 
 	uint32_t uheap;			/* User heap object pointer */
 #ifdef CONFIG_APP_BINARY_SEPARATION
-	uint32_t ram_start;		/* Start address of RAM partition for this app */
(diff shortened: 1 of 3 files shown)


sunghan-chang (dismissed):

os/include/tinyara/sched.h
```diff
@@ -620,7 +620,6 @@ struct tcb_s {
 
 	uint32_t uheap;			/* User heap object pointer */
 #ifdef CONFIG_APP_BINARY_SEPARATION
-	uint32_t ram_start;		/* Start address of RAM partition for this app */
 	uint32_t ram_size;		/* Size of RAM partition for this app */
```
kishore-sn, line 623: We could also remove ram_size variable.

jeongchanKim: Yes, I agree. I will update with another PR for checking `ram_size` and `ramstart`.

sunghan-chang (approved):

PR #4661, opened 2020-12-10
apps/examples/wifi_manager_sample/wm_test.c: Fix auth type specified wm_test join issues.

fengyunlai:
Fix 2 parts: 1. wep_shared key not only support ASCII password, but also HEX password, while checking password length, add conditions.
                  2. In wifi_interactive_mode.c, cmd_wifi_connect() checked AES and auth type match. It is okay to skip, delete it to keep it same as RTK default wifi_interactive_mode.c, if keep this check, need to add for wpa, wpa12, wps for all cases.

### DIFF
--- a/apps/examples/wifi_manager_sample/wm_test.c
+++ b/apps/examples/wifi_manager_sample/wm_test.c
@@ -1036,9 +1036,9 @@ static int _wm_test_join(struct options *opt, int argc, char *argv[])
 	}
 
 	if (opt->auth_type == WIFI_MANAGER_AUTH_WEP_SHARED) {
-		if (strlen(argv[5]) == 13) {
+		if ((strlen(argv[5]) == 13) || (strlen(argv[5]) == 26)) {
 			opt->crypto_type = WIFI_MANAGER_CRYPTO_WEP_128;
-		} else if (strlen(argv[5]) == 5) {
+		} else if ((strlen(argv[5]) == 5) || (strlen(argv[5]) == 10)) {
 			opt->crypto_type = WIFI_MANAGER_CRYPTO_WEP_64;
 		} else {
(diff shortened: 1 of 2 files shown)


pillip8282 (approved):
lgtm

apps/examples/wifi_manager_sample/wm_test.c
```diff
@@ -1036,9 +1036,9 @@ static int _wm_test_join(struct options *opt, int argc, char *argv[])
 	}
 
 	if (opt->auth_type == WIFI_MANAGER_AUTH_WEP_SHARED) {
-		if (strlen(argv[5]) == 13) {
+		if ((strlen(argv[5]) == 13) || (strlen(argv[5]) == 26)) {
```
davidfather, line 1039: How can we discriminate hex from ascii?

fengyunlai: You can discriminate it by its password length, 10 or 26 means HEX, 5 or 13 means ASCII

davidfather: How about 'asdfg12345'? Is it correct to consider this as a hex value?

fengyunlai: "asdfg12345" is invalid wep password, because hex password can be only from 0~9, a~f,A~F. But the _wm_test_join() here only determine whether its WIFI_MANAGER_CRYPTO_WEP_128 or WIFI_MANAGER_CRYPTO_WEP_64. So the modification is okay.

jsdosa (approved):
Looks nice to me.

davidfather (approved):
Looks good to me.

PR #4662, opened 2020-12-11
drivers: Store filep to distinguish fds of poll waiter slots

jeongarmy:
Store filep when file is registered in a list of waiters for polling and check it when closing filep.


jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4666, opened 2020-12-11
Load apps into single RAM region

kishore-sn:
Load apps into single RAM region to make sure that all app sections are within 32MB address range.
Modify defconfig to use 2 heaps to allow the above change

### DIFF
--- a/os/arch/Kconfig
+++ b/os/arch/Kconfig
@@ -798,6 +798,16 @@ config RAM_KREGIONx_HEAP_INDEX
 		It means that its region can be handled to the index-th heap.
 		Index can be from 0 to (KMM_NHEAPS-1).
 
+config HEAP_INDEX_LOADED_APP
+	int "Heap index for allocating app elf sections and app heap"
+	default 0
+	depends on APP_BINARY_SEPARATION
+	---help---
+		This heap index is used for allocating elf sections and heap area
+		for loadable apps. This value must be set to point to a heap which
+		has sufficient size of contiguous memory to accomodate all the
+		app sections with their alignment requriements.
+
 config RAM_MALLOC_PRIOR_INDEX
 	int "prior region index(0 KMM_NHEAPS-1) of RAM used in malloc"
 	default 0
--- a/os/arch/arm/src/common/up_allocateheap.c
+++ b/os/arch/arm/src/common/up_allocateheap.c
@@ -161,6 +161,7 @@ void up_add_kregion(void)
 	kheap = kmm_get_heap();
 	for (region_cnt = 1; region_cnt < CONFIG_KMM_REGIONS; region_cnt++) {
 		if (kheap[kregionx_heap_idx[region_cnt]].mm_heapsize == 0) {
+			lldbg("Heap idx = %u start = 0x%x size = %d\n", kregionx_heap_idx[region_cnt], kregionx_start[region_cnt], kregionx_size[region_cnt]);
 			mm_initialize(&kheap[kregionx_heap_idx[region_cnt]], kregionx_start[region_cnt], kregionx_size[region_cnt]);
 			continue;
 		}
--- a/os/binfmt/libelf/libelf_addrenv.c
+++ b/os/binfmt/libelf/libelf_addrenv.c
@@ -70,6 +70,13 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+#ifndef CONFIG_HEAP_INDEX_LOADED_APP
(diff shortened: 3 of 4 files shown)


os/arch/Kconfig
```diff
@@ -798,6 +798,16 @@ config RAM_KREGIONx_HEAP_INDEX
...
 		Index can be from 0 to (KMM_NHEAPS-1).
 
+config APP_HEAP_INDEX
+	int "Heap index for allocating app elf sections and app heap"
+	default 0
+	depends on APP_BINARY_SEPARATION
+	---help---
+		This heap index is used for allocating elf section and heap area
+		for loadable apps. This value must be set to point to a heap which
+		has sufficiently size of contiguous memory to accomodate all the
```
jeongchanKim, line 808: sufficiently -> sufficient

jeongchanKim:
I have a query. This option is required only for RTL8721, so I'm not sure we can provide this option by default.

os/arch/Kconfig
```diff
@@ -798,6 +798,16 @@ config RAM_KREGIONx_HEAP_INDEX
 		It means that its region can be handled to the index-th heap.
 		Index can be from 0 to (KMM_NHEAPS-1).
 
+config APP_HEAP_INDEX
+	int "Heap index for allocating app elf sections and app heap"
+	default 0
+	depends on APP_BINARY_SEPARATION
+	---help---
+		This heap index is used for allocating elf section and heap area
```
jeongchanKim, line 806: elf sections

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -161,6 +161,7 @@ void up_add_kregion(void)
 	kheap = kmm_get_heap();
 	for (region_cnt = 1; region_cnt < CONFIG_KMM_REGIONS; region_cnt++) {
 		if (kheap[kregionx_heap_idx[region_cnt]].mm_heapsize == 0) {
+			lldbg("start = 0x%x size = %d\n", kregionx_start[region_cnt], kregionx_size[region_cnt]);
```
jeongchanKim, line 164: I think we'd better print the region num also.

kishore-sn:
> I have a query. This option is required only for RTL8721, so I'm not sure we can provide this option by default.

This option is required for all projects to make sure that we load all elf sections within 32MB region.

os/kernel/init/os_start.c
```diff
@@ -287,6 +287,20 @@ void os_start(void)
 {
 	int i;
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	ASSERT(CONFIG_APP_HEAP_INDEX < CONFIG_KMM_NHEAPS);
```
sunghan-chang, line 291: Let's make a build error instead of assert. Because they are predefined values, we can make a warning at build time. That's better usability I think.

sunghan-chang:
Could you resolve conflicts?

os/kernel/init/os_start.c
sunghan-chang: And for remain lines, we can move it into libelf_addrenv.c which uses CONFIG_APP_HEAP_INDEX.

os/tools/mkconfig.c
```diff
@@ -203,6 +203,10 @@ int main(int argc, char **argv, char **envp)
 	printf("#ifndef CONFIG_KMM_REGIONS\n");
 	printf("# define CONFIG_KMM_REGIONS 1\n");
 	printf("#endif\n\n");
+	printf("/* CONFIG_APP_HEAP_INDEX must be defined always */\n\n");
```
sunghan-chang, line 206: We can add this in libelf_addrenv.c like below
```
#ifndef CONFIG_APP_HEAP_INDEX
#define CONFIG_APP_HEAP_INDEX 0
#endif
```

os/arch/Kconfig
```diff
@@ -798,6 +798,16 @@ config RAM_KREGIONx_HEAP_INDEX
 		It means that its region can be handled to the index-th heap.
 		Index can be from 0 to (KMM_NHEAPS-1).
 
+config APP_HEAP_INDEX
```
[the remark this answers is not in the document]
kishore-sn: Done

os/kernel/init/os_start.c
kishore-sn: Done

os/tools/mkconfig.c
kishore-sn: Done

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #4667, opened 2020-12-14
os/board/rtl8721csm: Fix compile error when TrustZone not enable

ziyik:
Fix compile error when TrustZone not enable, do not call mpu_s_no_cache_init() when TrustZone is not enable.
mpu_s_no_cache_init() only exist in TrustZone

### DIFF
--- a/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
+++ b/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
@@ -1019,7 +1019,9 @@ u32 app_mpu_nocache_init(void)
 
 u32 app_mpu_s_nocache_init(void)
 {
+#ifdef CONFIG_AMEBAD_TRUSTZONE
 	mpu_s_no_cache_init();
+#endif
 }
 
 VOID app_vdd1833_detect(VOID)


jeongchanKim (approved):
LGTM

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1019,7 +1019,9 @@ u32 app_mpu_nocache_init(void)
 
 u32 app_mpu_s_nocache_init(void)
 {
+#ifdef CONFIG_AMEBAD_TRUSTZONE
```
sangam-swami, line 1022: you can put this condition at app_mpu_s_nocache_init call itself?

ziyik: The reason to put the condition here is to prevent compile error went TrustZone is not enable.
mpu_s_no_cache_init() API only exist image3, the Secure lib/image3.
If put this condition at app_mpu_s_nocache_init call itself compile error will still happen.
Unless I put the condition in both location.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -1019,7 +1019,9 @@ u32 app_mpu_nocache_init(void)
 
 u32 app_mpu_s_nocache_init(void)
```
sunghan-chang, line 1020: There is no return value for this function. Let's use void.

sunghan-chang (approved):

PR #4672, opened 2020-12-15
arch/arm/src/armv8-m: Fix build break with MPU Stack protection enabled.

thapav:
This patch fixes the following build break:
armv8-m/up_svcall.c: In function 'up_svcall':
armv8-m/up_svcall.c:277:23: error: 'tcb' undeclared (first use in this function)
   up_mpu_set_register(tcb->stack_mpu_regs);
                       ^~~


jeongchanKim (approved):
Looks good to me.

sunghan-chang (approved):

PR #4679, opened 2020-12-18
smartfs/journaling : Do not update last_recovery_seq if recovery failed

Taejun-Kwon:
Sometimes there is invalid journal which has checkin state but not checkout in the middle of journal list.
In this case last_recovery_seq is changed to sequence of invalid journal, then corrupt happend due to overwrite exist journals.
Hence fix it

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -5515,7 +5515,12 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
 					fdbg("Recovery Failed : ret : %d\n", ret);
 					return ret;
 				}
-				last_recovery_seq = dev->journal_seq;
+				/* If it was valid journal, update last_recovery_seq. Otherwise, update last_checkout_seq to use next journal */
+				if (ret == OK) {
+					last_recovery_seq = dev->journal_seq;
+				} else {
+					last_checkout_seq = dev->journal_seq;
+				}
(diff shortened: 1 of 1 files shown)


jeongchanKim (dismissed):
LGTM

jsdosa:
Looks nice to me.

os/fs/driver/mtd/smart.c
```diff
@@ -5515,7 +5515,12 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
 					fdbg("Recovery Failed : ret : %d\n", ret);
 					return ret;
 				}
-				last_recovery_seq = dev->journal_seq;
+				/* If it was valid journal, update last_recovery_seq. Otherwise, update last_checkout_seq to use next journal */
+				if (ret == OK) {
```
davidfather, line 5519: Does this 'if' statement mean the following one?
				if (ret == OK) {
					last_recovery_seq = dev->journal_seq;
				} else if (ret == -EINVAL) {
					last_checkout_seq = dev->journal_seq;
				} else {
					fdbg("Recovery Failed : ret : %d\n", ret);
					return ret;
				}

Taejun-Kwon: yes it is same but no advantage of line, I prepare this style, it is always easy to extend later.

davidfather (approved):
lgtm.

PR #4680, opened 2020-12-18
loadable_apps/wifiapp : Change user input variable type from char to int

jeongchanKim:
WID:10253106 Integer value received from 'fgetc' can't be safely converted to a character at wifiapp.c:89 without a test for EOF.

### DIFF
--- a/loadable_apps/loadable_sample/wifiapp/wifiapp.c
+++ b/loadable_apps/loadable_sample/wifiapp/wifiapp.c
@@ -59,7 +59,7 @@ int main(int argc, char **argv)
 int wifiapp_main(int argc, char **argv)
 #endif
 {
-	char ch;
+	int ch;
 	bool is_testing = true;
 
 #if defined(CONFIG_SYSTEM_PREAPP_INIT) && defined(CONFIG_APP_BINARY_SEPARATION)


loadable_apps/loadable_sample/wifiapp/wifiapp.c
```diff
@@ -83,35 +83,36 @@ int wifiapp_main(int argc, char **argv)
 #ifdef CONFIG_EXAMPLES_LOADABLE_MANUAL_TEST
 	while (is_testing) {
 		display_test_scenario();
-		ch = getchar();
-		switch (ch) {
+		while ((ch = getchar()) != EOF) {
```
Taejun-Kwon, line 86: I think additional while is not necessary, just break above while is enough if getchar returns EOF

sunghan-chang, line 86: The `ch` is char type so that after getting the return value of getchar, it can't compare with EOF.

jeongchanKim: Yes, but svace tool makes error about below.
```
Integer value received from 'fgetc' can't be safely converted to a character at wifiapp.c:89 without a test for EOF.
```
So I modified like this.

jeongchanKim: I changed the `ch` type to int. Thanks.

sunghan-chang (approved):

PR #4688, opened 2020-12-21
Modify RTK heap regions to fix auto update aging crash

kishore-sn:
This commit is for fixing the below issue in RTK board:
- Enable Auto Update Aging test in RTK loadable_apps config
- Run the auto aging test by starting the board
- After several (random number of) iterations there is a memfault

This commit makes following changes:
- Remove SRAM region from kregion list and reduce KMM heap regions to 1
- Comment out os_heap_init() is RTK board code

By making the above changes, we make sure that the kernel heap for RTK
board is created only in the PSRAM region.


thapav (approved):
LGTM

sunghan-chang (approved):

PR #4690, opened 2020-12-22
include/tinyara/net/if: add missing WPA3 type enum member for PR4597

fengyunlai:
Add WP3 missing enmum

Signed-off-by: fengyunlai <1063072208@qq.com>


pillip8282 (approved):
lgtm

PR #4696, opened 2020-12-22
arch/arm/Kconfig : Disable MPU based stack protection if selected v8m…

jeongchanKim:
… arch

ARMv8M MPU cannot permit to overlap the region.
So we cannot use MPU based stack protection on ARMv8M architecture.
We can refer to the belows:
https://developer.arm.com/documentation/100235/0100/The-Cortex-M33-Peripherals/Security-Attribution-and-Memory-Protection/Memory-Protection-Unit?lang=en
"When memory regions overlap, the processor generates a fault if a core access hits the overlapping regions."


sunghan-chang (approved):

davidfather (approved):
lgtm.

PR #4701, opened 2020-12-23
fs/procfs: Add /proc/meminfo entry

jeongarmy:


jeongchanKim:
This feature is almost same as `heapinfo`. We'd better discuss more what and how to use.

apps/system/utils/utils_free.c
```diff
@@ -54,24 +54,61 @@
...
+	int ret;
+	char *filepath;
+	char buf[FREE_BUFLEN];
+#if !defined(CONFIG_FS_AUTOMOUNT_PROCFS)
+	bool is_mounted;
+
+	is_mounted = false;
+
+	/* Mount Procfs to use */
+	ret = mount(NULL, PROCFS_MOUNT_POINT, PROCFS_FSTYPE, 0, NULL);
```
Taejun-Kwon, line 85: mount should be decided in board level, not here.

jeongchanKim:
As my understanding, we should follow the usage of linux.
But our meminfo is not compatible with linux. So I think meminfo is not a good name.

apps/system/utils/utils_free.c
```diff
@@ -56,22 +56,51 @@
...
+	printf(" Mem:   %11d%11d%11d%11d\n", data.arena, data.uordblks, data.fordblks, data.mxordblk);
 
+#if defined(CONFIG_BUILD_PROTECTED) && defined(CONFIG_MM_KERNEL_HEAP)
+	fd = open(MEMINFO_DRVPATH, O_RDWR);
+	if (fd < 0) {
+		printf("Open %s Fail, errno %d\n", MEMINFO_DRVPATH, get_errno());
+		return ERROR;
+	}
+
+	ret = ioctl(fd, MEMINFOIOC_KHEAP, (int)&data);
```
jeongchanKim, line 93: In this case, it includes user heaps also. Is it okay?
If then, I think Kmem is not a good expression.

os/drivers/meminfo/meminfo_drv.c
```diff
@@ -107,7 +130,10 @@ static int heapinfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		heapinfo_parse_heap(heap, option->mode, option->pid);
 		ret = OK;
 		break;
+#endif
 	default:
+		mdbg("Not supported\n");
+		ret = ENOSYS;
```
jeongchanKim, line 136: In this case, sometimes it cannot be handled with fail. Because in many cases, it checks the ioctl return with ERROR.
So I think, then we'd better return -ERRNO and check with negative when it checks.

os/include/tinyara/meminfo_drv.h
```diff
@@ -16,22 +16,20 @@
...
-#include <tinyara/binary_manager.h>
-#endif
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define HEAPINFO_DRVPATH     "/dev/heapinfo"
+#define MEMINFO_DRVPATH     "/dev/meminfo"
 
+#if defined(CONFIG_DEBUG_MM_HEAPINFO)
```
jeongchanKim, line 32: I'm not sure heapinfo structure is in meminfo header.

apps/system/utils/utils_free.c
```diff
@@ -56,22 +56,51 @@
...
+	if (ret == ERROR) {
+		printf("Meminfo ioctl Fail, errno %d\n", get_errno());
+		close(fd);
+		return ERROR;
+	}
 	printf("              total       used       free    largest\n");
-	printf("Data:   %11d%11d%11d%11d\n", data.arena, data.uordblks, data.fordblks, data.mxordblk);
+	printf("Kmem:	%11d%11d%11d%11d\n", data.arena, data.uordblks, data.fordblks, data.mxordblk);
```
jeongchanKim, line 100: For linux compatibility, only with `free` is good for showing the total.
If we want to expand, we'd better use options.

os/drivers/meminfo/meminfo_drv.c
```diff
@@ -20,70 +20,93 @@
...
+static int meminfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg);
+static ssize_t meminfo_read(FAR struct file *filep, FAR char *buffer, size_t len);
+static ssize_t meminfo_write(FAR struct file *filep, FAR const char *buffer, size_t len);
 
 /****************************************************************************
  * Private Data
  ****************************************************************************/
 
-struct heapinfo_dev_s {
+struct meminfo_dev_s {
```
jeongchanKim, line 47: Is this needed?

apps/system/utils/utils_free.c
jeongarmy: Right. I modified

os/include/tinyara/meminfo_drv.h
jeongarmy: Yes. I agree with your opinion. I will consider more.

apps/system/utils/utils_free.c
```diff
@@ -56,22 +56,48 @@
...
 
+#if defined(CONFIG_BUILD_PROTECTED)
+	fd = open(MEMINFO_DRVPATH, O_RDWR);
+	if (fd < 0) {
+		printf("Open %s Fail, errno %d\n", MEMINFO_DRVPATH, get_errno());
+		return ERROR;
+	}
+
+	ret = ioctl(fd, MEMINFOIOC_KHEAP, (int)&data);
+	if (ret == ERROR) {
```
Taejun-Kwon, line 86: Close first and use if state more explicitly.
```
close(fd);
if (ret != OK) {
   printf();
   return ERROR;
}
```

apps/system/utils/utils_heapinfo.c
```diff
@@ -277,12 +277,12 @@ int utils_heapinfo(int argc, char **args)
...
 	}
 #endif
-	heapinfo_fd = open(HEAPINFO_DRVPATH, O_RDWR);
+	heapinfo_fd = open(MEMINFO_DRVPATH, O_RDWR);
 	if (heapinfo_fd < 0) {
 		printf("Heapinfo Fail, %d.\n", get_errno());
 		return ERROR;
 	}
-	ret = ioctl(heapinfo_fd, HEAPINFOIOC_PARSE, (int)&options);
+	ret = ioctl(heapinfo_fd, MEMINFOIOC_PARSE, (int)&options);
```
Taejun-Kwon, line 285: same as above comment.

os/drivers/meminfo/meminfo_drv.c
```diff
@@ -98,7 +116,7 @@ static int heapinfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 			break;
 		}
 		if (heap == NULL) {
-			return ERROR;
+			return -EINVAL;
```
Taejun-Kwon, line 119: I think if option->heap_type is HEAPINFO_HEAP_TYPE_KERNEL, heap is not NULL all the time, so this state had better move to line 112~113.

apps/system/utils/utils_free.c
```diff
@@ -56,22 +56,48 @@
...
+#include <tinyara/fs/ioctl.h>
+#include <tinyara/meminfo_drv.h>
+#endif
 
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
 int utils_free(int argc, char **args)
 {
+#if defined(CONFIG_BUILD_PROTECTED) && defined(CONFIG_MM_KERNEL_HEAP)
```
sunghan-chang, line 72:
```
	struct mallinfo data;
#if defined(CONFIG_BUILD_PROTECTED) && defined(CONFIG_MM_KERNEL_HEAP)
	int fd;
	int ret;

	fd = open(MEMINFO_DRVPATH, O_RDWR);
	if (fd < 0) {
		printf("Open %s Fail, errno %d\n", MEMINFO_DRVPATH, get_errno());
		return ERROR;
	}
....
#else
..
#endif
```

build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -417,6 +417,7 @@ CONFIG_GPIO=y
 # CONFIG_RTC is not set
 # CONFIG_WATCHDOG is not set
 # CONFIG_TIMER is not set
+CONFIG_MEMINFO=y
```
sunghan-chang, line 420: Why do you enable only three defconfigs?

apps/system/utils/Kconfig
```diff
@@ -113,13 +113,15 @@ endif #!DISABLE_ENVIRON
...
 	bool "free"
 	default y
+	select MEMINFO if BUILD_PROTECTED
 	---help---
 		Show information about memory status
 
 config ENABLE_HEAPINFO
 	bool "heapinfo"
 	default y
 	depends on DEBUG_MM_HEAPINFO && FS_PROCFS
```
sunghan-chang, line 123: Even we use meminfo driver, do we still need FS_PROCFS ?

os/drivers/meminfo/Make.defs
```diff
@@ -15,17 +15,13 @@
...
 
-# Include heapinfo driver support
-
-DEPPATH += --dep-path heapinfo
-VPATH += :heapinfo
+# Include meminfo driver support
 
-endif
+DEPPATH += --dep-path meminfo
+VPATH += :meminfo
```
sunghan-chang, line 27: Let's cover them as well for `ifeq and endif`

apps/system/utils/utils_free.c
jeongarmy: It is good for safety but we can't know the errno returned from ioctl.

apps/system/utils/Kconfig
jeongarmy: Yes. It still use procfs for task information like pid, stack, name

os/include/tinyara/fs/ioctl.h
```diff
@@ -94,7 +94,7 @@
 #define _FOTABASE       (0x1900)	/* FOTA ioctl commands */
 #define _GPIOBASE       (0x2000)	/* GPIO ioctl commands */
 #define _TMBASE         (0x2100)	/* Task Management ioctl commands */
-#define _HEAPINFOBASE   (0x2200)	/* Heapinfo ioctl commands */
+#define _MMINFOBASE     (0x2200)	/* Mm ioctl commands */
```
jeongchanKim, line 97: I think `mminfo ioctl commands` looks better, because we have `mm` module independently.

os/mm/umm_heap/umm_mallinfo.c
```diff
@@ -103,6 +107,14 @@ struct mallinfo mallinfo(void)
 int mallinfo(struct mallinfo *info)
 {
 	int heap_idx;
+	DEBUGASSERT(info);
```
jeongchanKim, line 110: `mallinfo` is not an important functionality, so I think we don't need to make assert.
I think `return ERROR` is enough.

jeongchanKim (commented):
Please modify the commit msg also.
meminfo -> mminfo

apps/system/utils/utils_free.c
jeongarmy: modified

os/drivers/mminfo/mminfo_drv.c
```diff
@@ -118,14 +139,14 @@ static int heapinfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+ * Name: mminfo_drv_register
  *
  * Description:
- *   Register heapinfo driver path, HEAPINFO_DRVPATH
+ *   Register mminfo driver path, MMINFO_DRVPATH
  *
  ****************************************************************************/
 
-void heapinfo_drv_register(void)
+void mminfo_drv_register(void)
```
Taejun-Kwon, line 149: so then mminfo_register.

apps/system/utils/utils_free.c
```diff
@@ -56,22 +56,46 @@
...
+
+	fd = open(MMINFO_DRVPATH, O_RDWR);
+	if (fd < 0) {
+		printf("Open %s Fail, errno %d\n", MMINFO_DRVPATH, get_errno());
+		return ERROR;
+	}
 
+	ret = ioctl(fd, MMINFOIOC_KHEAP, (int)&data);
+	if (ret == ERROR) {
+		printf("Meminfo ioctl Fail, errno %d\n", get_errno());
```
jeongchanKim, line 85: Please modify `Meminfo` to `Mminfo`.

os/drivers/memory/mminfo.c
```diff
@@ -20,70 +20,88 @@
...
 
 /************************************************************************************
- * Name: heapinfo_ioctl
+ * Name: mminfo_ioctl
  *
  * Description: The ioctl method for heapinfo.
  *
  ************************************************************************************/
-static int heapinfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
+static int mminfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
```
Taejun-Kwon, line 76: if ret is initialized as a OK, you can reduce more line here.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -646,25 +650,52 @@ void binary_update_aging_test(void)
 {
 	int ret;
 	binary_update_info_t bin_info;
+#ifdef CONFIG_MMINFO
+	int fd;
+	struct mallinfo data;
+
+	fd = open(MMINFO_DRVPATH, O_RDWR);
+	if (fd <= 0) {
```
jeongchanKim, line 658: In `open()` API description, it says
```
 *   The new file descriptor is returned on success; -1 (ERROR) is returned
 *   on any failure the errno value set appropriately.
```
So I think we don't need to check with `<= 0`.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -646,25 +650,52 @@ void binary_update_aging_test(void)
...
+		printf("Open %s Fail, %d errno %d\n", MMINFO_DRVPATH, fd, get_errno());
+	}
+#endif
 
 	while (1) {
+#ifdef CONFIG_MMINFO
+		if (fd > 0) {
```
jeongchanKim, line 665: With above comment, `fd` can be `0`. Then we should change the comparison.

loadable_apps/loadable_sample/wifiapp/binary_update.c
```diff
@@ -646,25 +650,52 @@ void binary_update_aging_test(void)
...
 		ret = binary_update_reload(APP_NAME);
 		if (ret != OK) {
-			return;
+			break;
 		}
 	}
+
+	printf(" Binary update aging Test EXIT!! \n");
+#ifdef CONFIG_MMINFO
+	if (fd > 0) {
```
jeongchanKim, line 696: Same as above.

loadable_apps/loadable_sample/wifiapp/binary_update.c
jeongarmy: Right. but actually it returns a value which is larger than 2 on success.
That's because fd 0~2 is assigned for stdin/out/err when a thread is created.

sunghan-chang (approved):

PR #4707, opened 2020-12-24
arch/arm/src/amebad/: Fix more warnings.

fengyunlai:
Fix more warnings like incompatible type, unused variable, unintialized variable.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_irq.c
+++ b/os/arch/arm/src/amebad/amebad_irq.c
@@ -358,13 +358,13 @@ void up_irqinitialize(void)
 
 void up_disable_irq(int irq)
 {
-  uintptr_t regaddr;
-  uint32_t regval;
-  uint32_t bit;
 #if 1
 	  irq -= AMEBAD_IRQ_FIRST;
 	  __NVIC_DisableIRQ(irq);
 #else
+  uintptr_t regaddr;
+  uint32_t regval;
+  uint32_t bit;
   if (amebad_irqinfo(irq, &regaddr, &bit, NVIC_CLRENA_OFFSET) == 0)
     {
       /* Modify the appropriate bit in the register to disable the interrupt.
@@ -398,13 +398,13 @@ void up_disable_irq(int irq)
 
 void up_enable_irq(int irq)
 {
-	uintptr_t regaddr;
-	uint32_t regval;
-	uint32_t bit;
 #if 1
 	irq -= AMEBAD_IRQ_FIRST;
 	__NVIC_EnableIRQ(irq);
 #else
+	uintptr_t regaddr;
+	uint32_t regval;
+	uint32_t bit;
 	if (amebad_irqinfo(irq, &regaddr, &bit, NVIC_ENA_OFFSET) == 0)
 	{
 		/* Modify the appropriate bit in the register to enable the interrupt.
@@ -453,9 +453,6 @@ void up_ack_irq(int irq)
 #ifdef CONFIG_ARCH_IRQPRIO
 int up_prioritize_irq(int irq, int priority)
 {
-  uint32_t regaddr;
-  uint32_t regval;
-  int shift;
 
   DEBUGASSERT(irq >= AMEBAD_IRQ_MEMFAULT && irq < NR_IRQS &&
               (unsigned)priority <= NVIC_SYSH_PRIORITY_MIN);
@@ -464,6 +461,9 @@ int up_prioritize_irq(int irq, int priority)
   priority = priority >> (8U - __NVIC_PRIO_BITS);
   __NVIC_SetPriority(irq, priority); //use CMSIS for now
 #else
+  uint32_t regaddr;
+  uint32_t regval;
+  int shift;
   if (irq < AMEBAD_IRQ_FIRST)
     {
       /* NVIC_SYSH_PRIORITY() maps {0..15} to one of three priority
--- a/os/board/rtl8721csm/src/component/os/os_dep/include/osdep_service.h
+++ b/os/board/rtl8721csm/src/component/os/os_dep/include/osdep_service.h
@@ -179,7 +179,7 @@ typedef int (*event_handler_t)(char *buf, int buf_len, int flags, void *user_dat
 #define CONFIG_THREAD_COMM_SEMA
 struct task_struct {
 	const char *task_name;
-	_thread_hdl_ task;	/* I: workqueue thread */
+	pid_t task;	/* I: workqueue thread */
 
 #ifdef CONFIG_THREAD_COMM_SIGNAL
 	const char *name;	/* I: workqueue thread name */
--- a/os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
+++ b/os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
@@ -668,11 +668,7 @@ static int _tizenrt_create_task(struct task_struct *ptask, const char *name, u32
 	struct sched_param sparam;
 	int res = 0;
 
-	pthread_t *tid = _tizenrt_zmalloc(sizeof(pthread_t));
-	if (tid == NULL) {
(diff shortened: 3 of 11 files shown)


os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -716,10 +716,15 @@ static int _tizenrt_create_task(struct task_struct *ptask, const char *name, u32
 	return _FAIL;
 #else
 	int func_addr, ctx_addr;
-	pid_t pid;
 	char str_func_addr[9];
 	char str_ctx_addr[9];
 	char *task_info[3];
+
+	pthread_t *pid = (pthread_t *)_tizenrt_zmalloc(sizeof(pthread_t));
```
jeongchanKim, line 723: I'm not sure why you changed like this.
It is same as previous, but looks not good.
And if you want to use like this, please add `free` for failure or termination case.

fengyunlai: You are right Mr Kim, I fix it in new commit. We made these changes to fix warings of incompatible type. Basically I followed the way in #if USE_PTHREAD_MUTEX in the same API itself.

sunghan-chang: That's just integer value. You don't need to allocate the memory for that. This wastes the memory. And I can't find the free at success case. I mean when you delete or finish the task, there is no free for this allocation. (What you added is a failure case.)

fengyunlai: Thank you Mr Chang. Let me clarify on that: 
1. You are right, if I want to use malloc, I still need to modify _tizenrt_delete_task() to free the memory when deleting the task.
2. Why I made those changes are because in os\board\rtl8721csm\src\component\os\os_dep\include\osdep_service.h, line 182, for the struct task_struct, _thread_hdl_ task is actually a pointer which we assign value like this: ptask->task = pid; which is not correct if pid is an integer.

So do you think I can proceed to fix it by modifying _tizenrt_delete_task()?

os/arch/arm/src/amebad/amebad_irq.c
```diff
@@ -358,13 +358,13 @@ void up_irqinitialize(void)
 
 void up_disable_irq(int irq)
 {
-  uintptr_t regaddr;
-  uint32_t regval;
-  uint32_t bit;
 #if 1
```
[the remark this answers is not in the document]
fengyunlai: This is not needed. We are using __NVIC_DisableIRQ(irq); instead. But I can also put them back under condition #1 #else, under the #else part if you think it is better.

os/arch/arm/src/amebad/amebad_irq.c
```diff
@@ -398,13 +398,13 @@ void up_disable_irq(int irq)
 
 void up_enable_irq(int irq)
 {
-	uintptr_t regaddr;
-	uint32_t regval;
-	uint32_t bit;
 #if 1
```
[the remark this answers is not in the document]
fengyunlai: similar to above reply

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
sunghan-chang: 1. If you use the allocation way, it wastes memory. (Actually it requires an integer memory, but allocation way requires two integer memory.)
2. I don't think you can know the time to be deleted for all of threads you made. Then how will you call the delete function?

fengyunlai: 1. Dear Mr Chang. From my point of view, assigning an integer to a pointer is wrong usage. And that's why we need to fix it. So may I ask your advice? 
2. I understand that if we do not delete task on time, which might cause memory leakage with my modifications. But even with current branch's code, if we do not delete task, calling task_delete(pid), we still have other issues right? So for both case we need to make sure task creation and deletion are properly handled.

sunghan-chang: 1. PID is an integer value always. You don't need to keep it as pointer value. In my opinion, you should change the type in your structure.
```
typedef pid_t _thread_hdl_;
```
or
```
struct task_struct {
	const char *task_name;
	pid_t task;	/* I: workqueue thread */
```
2. Because you use a pointer and allocation way for pid value, task termination has a problem. If you use my suggestion as 1st, then there is no issue.
```
But even with current branch's code, if we do not delete task, calling task_delete(pid), we still have other issues right? So for both case we need to make sure task creation and deletion are properly handled.
```
In your code, yes. But you can't modify it with current pointer type I think.

fengyunlai: Sure Mr Chang, I checked our code, and changed our struct as adviced.

os/board/rtl8721csm/src/component/os/os_dep/include/osdep_service.h
```diff
@@ -179,7 +179,7 @@ typedef int (*event_handler_t)(char *buf, int buf_len, int flags, void *user_dat
 #define CONFIG_THREAD_COMM_SEMA
 struct task_struct {
 	const char *task_name;
-	_thread_hdl_ task;	/* I: workqueue thread */
+	pid_t task;	/* I: workqueue thread */
```
sunghan-chang, line 182: Because you change this type, you should check the code which use this variable like the tid in _tizenrt_create_task function. 
```
	pthread_t *tid = _tizenrt_zmalloc(sizeof(pthread_t));
	if (tid == NULL) {
		DBG_ERR("Failed to malloc for tid\n");
		goto err_exit;
	}
	res = pthread_create(tid, &attr, (pthread_startroutine_t) func, thctx);
==>
	pthread_t tid;
	res = pthread_create(&tid, &attr, (pthread_startroutine_t) func, thctx);
```

sunghan-chang:
```
	ptask->task = NULL;
==>
	ptask->task = -1;
```

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -1223,3 +1223,4 @@ const struct osdep_service_ops osdep_service = {
 	NULL,						// rtw_create_secure_context
 	NULL,						//rtw_get_current_TaskHandle
 };
+
```
jeongchanKim, line 1226: Please revert this.

fengyunlai: Yes

os/board/rtl8721csm/src/component/os/os_dep/include/osdep_service.h
fengyunlai: Sure, changed as adviced.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -53,13 +53,15 @@ extern unsigned int __PsramStackLimit;
 void os_heap_init(void){
 	kregionx_start[0] = (void *)HEAP_BASE;
 	kregionx_size[0] = (size_t)(HEAP_LIMIT - HEAP_BASE);
+#if CONFIG_KMM_REGIONS > 1
```
sunghan-chang, line 56: We already modify here. Let's revert this change.

fengyunlai: No problem

sunghan-chang (approved):

PR #4708, opened 2020-12-24
os/board/rtl8721csm : Enable power save mode

keerthanalex:
Enabled power save mode for TizenRT
Updated wlan and wps libs for power save mode enable with/without fpu


sunghan-chang:
Could you let us know wifi performance result with this pr?

sunghan-chang (approved):

keerthanalex:
> Could you let us know wifi performance result with this pr?

Updated the comment with result.

PR #4709, opened 2020-12-28
fs/smartfs : adjust log level regarding sector recovery

Taejun-Kwon:
Change log level to verbose about freeing node, and change to error regarding invalidate entry


jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4710, opened 2020-12-29
kernel/binary_manager_load : Modify read_size and file_size to unsign…

jeongchanKim:
…ed int

read_size and file_size cannot be negative, so changed to unsigned int.


jeongarmy (approved):
LGTM

sunghan-chang (approved):

PR #4716, opened 2020-12-30
[DO NOT MERGE]os/kernel/init: Add Idle stack MPU protection

thapav:
This patch secures a gaurd band within the idle stack and
configures it as read only region.

### DIFF
--- a/os/kernel/init/os_start.c
+++ b/os/kernel/init/os_start.c
@@ -79,6 +79,7 @@
 #ifdef CONFIG_DRIVERS_KERNEL_TEST
 #include  <tinyara/kernel_test_drv.h>
 #endif
+#include  <tinyara/mpu.h>
 
 #include  "sched/sched.h"
 #include  "signal/signal.h"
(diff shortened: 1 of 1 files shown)


os/kernel/init/os_start.c
```diff
@@ -364,10 +365,29 @@ void os_start(void)
 	g_idleargv[1]  = NULL;
 	g_idletcb.argv = g_idleargv;
 
+#ifdef CONFIG_MPU_STACK_OVERFLOW_PROTECTION
+
+	g_idletcb.cmn.stack_alloc_ptr = (void *)(g_idle_topstack - CONFIG_IDLETHREAD_STACKSIZE);
+
+	if(g_idletcb.cmn.stack_alloc_ptr) {
```
jeongchanKim, line 372: Please add space after `if`.

os/kernel/init/os_start.c
```diff
@@ -364,10 +365,29 @@ void os_start(void)
...
+		size_t size_of_stack;
+		size_t top_of_stack = (size_t)g_idletcb.cmn.stack_alloc_ptr;
+
+		top_of_stack = MPU_ALIGN_UP(top_of_stack);
+		size_of_stack = CONFIG_IDLETHREAD_STACKSIZE - CONFIG_MPU_STACK_GUARD_SIZE;
```
jeongchanKim, line 377: In this case, it is smaller than expected which user set. Is it okay?

os/kernel/init/os_start.c
```diff
@@ -364,10 +365,29 @@ void os_start(void)
...
+		g_idletcb.cmn.adj_stack_ptr = (void *)top_of_stack;
+		g_idletcb.cmn.adj_stack_size = size_of_stack;
```
jeongchanKim, line 379: Is it okay not to check that the size is power of two, especially ARMv7M?

os/kernel/init/os_start.c
```diff
@@ -79,6 +79,7 @@
 #ifdef CONFIG_DRIVERS_KERNEL_TEST
 #include  <tinyara/kernel_test_drv.h>
 #endif
+#include  <tinyara/mpu.h>
```
jeongchanKim, line 82: We can add `#ifdef` only for MPU supported.

PR #4717, opened 2020-12-31
Update Dump Tool (Realtek - rtl8721csm)

amogh-samsung:
- ARM Cortex M devices do not support register for frame pointer backtracking.
  Hence, build fails when CONFIG_FRAME_POINTER is selected.
  Add flag to omit frame pointer while building for rtl8721csm.
- For such cases, it is not possible to obtain exact call stack for the crash.
- Modify flash base address of rtl8721csm to proper value.
- Add NOTE to "How-To" file for dumps to declare limitation for ARM Cortex M devices.
- Modify formatting of filesystem start address print output to console.

### DIFF
--- a/tools/dump_tool/HowToUseDumpTool.md
+++ b/tools/dump_tool/HowToUseDumpTool.md
@@ -30,6 +30,11 @@ Debug Options -> Enable Debug Output Features to y
 ```
 Debug Options -> Enable backtracking using Frame pointer register  to y
 ```
+```
+NOTE: - For devices that use ARM Cortex M, backtracking of frame pointer is not supported.
+Hence, it is not possible to obtain exact call stack for the crash.
+Ramdump may not produce results as expected in such cases.
+```
 
 ## How to upload RAMDUMP-or-UserfsDUMP
 ### In Linux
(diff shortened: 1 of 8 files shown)


jeongchanKim (approved):
LGTM

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -30,6 +30,11 @@ Debug Options -> Enable Debug Output Features to y
 ```
 Debug Options -> Enable backtracking using Frame pointer register  to y
 ```
+```
+NOTE: - For devices that use ARM Cortex M, backtracking of frame pointer is not supported.
+Hence, it is not possible to get the pointer to stack frame for these devices.
```
thapav, line 35: I think following would be more appropriate:
"it is not possible to obtain exact call stack for the crash"

thapav: This can be modified in the commit description as well

amogh-samsung: Ohkay, thankyou.
Updated

thapav (approved):
LGTM

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4725, opened 2021-01-05
arch/amebad/flash : Add function to verify erased page

Taejun-Kwon:
If sudden power off happened during erase, sometimes target pages are not erased properly.
In product level, level of voltage can be increase for a while, then fs could treat it is correct.
To prevent this situation, verify logic must be added

### DIFF
--- a/os/arch/arm/src/amebad/amebad_flash.c
+++ b/os/arch/arm/src/amebad/amebad_flash.c
@@ -116,7 +116,8 @@ static ssize_t amebad_erase_page(size_t page)
 {
 	uint32_t address;
 	irqstate_t irqs;
-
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -116,7 +116,8 @@ static ssize_t amebad_erase_page(size_t page)
 {
 	uint32_t address;
 	irqstate_t irqs;
-
+	ssize_t ret = OK;
```
jeongchanKim, line 119: I think we don't need to initialize the `ret`.

sunghan-chang (approved):

PR #4727, opened 2021-01-06
mm/umm_mallinfo : Use First heap only when binary separation

jeongchanKim:
When CONFIG_APP_BINARY_SEPARATION, user heap only can be single heap.

### DIFF
--- a/os/mm/umm_heap/umm_mallinfo.c
+++ b/os/mm/umm_heap/umm_mallinfo.c
@@ -87,14 +87,19 @@
 
 struct mallinfo mallinfo(void)
 {
-	int heap_idx;
 	struct mallinfo info;
 #if CONFIG_KMM_NHEAPS > 1
 	memset(&info, 0, sizeof(struct mallinfo));
 #endif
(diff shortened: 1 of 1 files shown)


Taejun-Kwon (approved):

os/mm/umm_heap/umm_mallinfo.c
```diff
@@ -87,25 +87,33 @@
 
 struct mallinfo mallinfo(void)
 {
-	int heap_idx;
 	struct mallinfo info;
 #if CONFIG_KMM_NHEAPS > 1
 	memset(&info, 0, sizeof(struct mallinfo));
 #endif
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	mm_mallinfo(&BASE_HEAP[0], &info);
```
jeongarmy, line 95: I think BASE_HEAP is enough

jeongchanKim: For easy understanding, I will keep this. Thanks.

sunghan-chang (approved):

PR #4731, opened 2021-01-08
tools/memory: add README and add file existence check

sunghan-chang:
1. tools/memory: check existence of default ELF 
    When ELF option is not specified by user to g_var_profiler.sh, default ELF binary is used.
    But in some board default ELF binary is not built resulting into empty output file by the script.
    To avoid above scenario, we will check pre-existence of ELF binary.
2. tools/memory: add README for g_var_profiler.sh

### DIFF
--- a/tools/memory/g_var_profiler.sh
+++ b/tools/memory/g_var_profiler.sh
@@ -81,6 +81,10 @@ fi
 
 if [ -z ${ELF} ]; then
 	ELF="${MEMTOOL_PATH}/../../build/output/bin/tinyara"
+	if [ ! -f ${ELF} ]; then
+		echo "Unavailable Default ELF \"${ELF}\""
+		exit 1
+	fi
 fi
 
 if [ -z ${OUTPUT_FILE} ]; then
--- a/tools/memory/HowToUseGlobalVariableProfiler.md
+++ b/tools/memory/HowToUseGlobalVariableProfiler.md
@@ -1,0 +1,40 @@
+# How to use global variable profiler
+The *g_var_profiler.sh* script profiles static memory usages.  
+It generates the list of global variables which have greater size than given.
+
+# How to use
+Please find the usages with *-h* option.
+```bash
+tools/memory$ ./g_var_profiler.sh -h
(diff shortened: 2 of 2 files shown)


sunghan-chang:
Could you review the README?

tools/memory/g_var_profiler.sh
```diff
@@ -113,6 +117,6 @@ do
 	fi
 
 done < $temp_output_file
-rm $temp_output_file
+#rm $temp_output_file
```
sangam-swami, line 120: don't we need to remove this file?

tools/memory/g_var_profiler.sh
```diff
@@ -81,6 +81,10 @@ fi
 
 if [ -z ${ELF} ]; then
 	ELF="${MEMTOOL_PATH}/../../build/output/bin/tinyara"
+	if [ ! -f ${ELF} ]; then
+		echo "No valid ELF file"
```
sangam-swami, line 85: "No ELF file generated for this board" is this correct?

sangam-swami (commented):
1. tools/memory: 
When ELF option is not specified by user to g_var_profiler.sh, default ELF binary is used.
But in some board default ELF binary is not build resulting into empty output file by the script.

To avoid above scenario, we will check pre-existence of ELF binary.

tools/memory/g_var_profiler.sh
sunghan-chang: Oh it's my mistake. I will revert.

tools/memory/g_var_profiler.sh
sunghan-chang: Hm.. It's little bit different. It could be a case for no generated or could be a case for different file like realtek. Realtek has different file name, tinyara.axf. How about "${ELF} is not available"?

sangam-swami: yes looks fine..

sangam-swami (approved):
looks fine

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4733, opened 2021-01-08
Idle stack

sunghan-chang:
1. arch/arm,xtensa: move the location of up_getsp to arch.h 
    This commit moves the location of up_getsp function to arch.h to use in other module.
    And this renames xtensa_getsp to up_getsp to keep the same usages in other module.
2. kernel/init: compare the sp of the idle to tcb values
    The os_start sets idle thread's stack informations. But they are just calculated values, not real core stack pointer.
    This commit adds comparision real stack pointer to tcb values.


sangam-swami (approved):

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4735, opened 2021-01-11
arch/arm/src/amebad: Fix the 'extra CR after LF' issue

AdamZhang0124:
Add function LOGUART_PutChar_RAM

### DIFF
--- a/os/arch/arm/src/amebad/amebad_serial.c
+++ b/os/arch/arm/src/amebad/amebad_serial.c
@@ -374,6 +374,25 @@ static u32 uart_index_get(PinName tx)
 	return 3;
 }
 
+static void LOGUART_PutChar_RAM(u8 c)
+{
+	UART_TypeDef* LOGUART = UART2_DEV;
+	u32 CounterIndex = 0;
+
+	if (ConfigDebugClose == 1)
+		return;
+
+	while(1) {
+		CounterIndex++;
+		if (CounterIndex >=6540)
+			break;
+
+		if (LOGUART->LSR & RUART_LINE_STATUS_REG_THRE)
+			break;
+	}
+	LOGUART->RB_THR = c;
+}
+
 /****************************************************************************
  * Name: rtl8721d_up_setup
  *
@@ -723,7 +742,7 @@ void up_serialinit(void)
  ****************************************************************************/
 void up_lowputc(char ch)
 {
-	LOGUART_PutChar(ch);
+	LOGUART_PutChar_RAM(ch);
 }
 
 /****************************************************************************
--- a/os/arch/arm/src/amebad/amebad_serial.c
+++ b/os/arch/arm/src/amebad/amebad_serial.c
@@ -108,16 +108,19 @@
 #if defined(CONFIG_UART0_SERIAL_CONSOLE)
 #define CONSOLE_DEV             g_uart0port             /* UART0 is console */
 #define TTYS0_DEV               g_uart0port             /* UART0 is ttyS0 */
+#define CONSOLE                 UART0_DEV
 #define UART0_ASSIGNED  1
 #define HAVE_SERIAL_CONSOLE
 #elif defined(CONFIG_UART1_SERIAL_CONSOLE)
 #define CONSOLE_DEV             g_uart1port             /* UART1 is console */
 #define TTYS0_DEV               g_uart1port             /* UART1 is ttyS0 */
+#define CONSOLE                 UART1_DEV
 #define UART1_ASSIGNED  1
 #define HAVE_SERIAL_CONSOLE
 #elif defined(CONFIG_UART2_SERIAL_CONSOLE)
 #define CONSOLE_DEV             g_uart2port             /* UART2 is console */
 #define TTYS0_DEV               g_uart2port             /* UART2 is ttyS0 */
+#define CONSOLE                 UART2_DEV
 #define UART2_ASSIGNED  1
 #define HAVE_SERIAL_CONSOLE
 #else
@@ -158,6 +161,7 @@
 #define TTYS2_DEV               g_uart2port             /* UART2 is ttyS2 */
 #define UART2_ASSIGNED  1
 #endif
+#define CHAR_TIMEOUT 6540
 
 /****************************************************************************
  * Private Types
@@ -376,7 +380,7 @@ static u32 uart_index_get(PinName tx)
 
 static void LOGUART_PutChar_RAM(u8 c)
 {
-	UART_TypeDef* LOGUART = UART2_DEV;
+	UART_TypeDef* LOGUART = CONSOLE;
 	u32 CounterIndex = 0;
 
 	if (ConfigDebugClose == 1)
@@ -384,7 +388,7 @@ static void LOGUART_PutChar_RAM(u8 c)
 
 	while(1) {
 		CounterIndex++;
-		if (CounterIndex >=6540)
+		if (CounterIndex >= CHAR_TIMEOUT)
 			break;
 
 		if (LOGUART->LSR & RUART_LINE_STATUS_REG_THRE)
@@ -708,11 +712,14 @@ void up_serialinit(void)
 	UART_DeInit(UART2_DEV);
 	UART_ClearRxFifo(UART2_DEV);
 	UART_ClearTxFifo(UART2_DEV);
+
+#ifdef CONSOLE_DEV
 	CONSOLE_DEV.isconsole = true;
 	rtl8721d_up_setup(&CONSOLE_DEV);
 
 	/* Register the console */
 	uart_register("/dev/console", &CONSOLE_DEV);
+#endif
 
 	/* Register all UARTs */
 #ifdef TTYS0_DEV


sunghan-chang:
@AdamZhang0124 I don't know what the issue is and I can't find difference between new api and existed one because there is no code for existed. Could you share more details?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -374,6 +374,25 @@ static u32 uart_index_get(PinName tx)
...
+{
+	UART_TypeDef* LOGUART = UART2_DEV;
+	u32 CounterIndex = 0;
+
+	if (ConfigDebugClose == 1)
+		return;
+
+	while(1) {
+		CounterIndex++;
+		if (CounterIndex >=6540)
```
thapav, line 387: Coding rule violation*
>= 6540

thapav (commented):
Can you please confirm the same with UART logs or dump logs for loadable config

os/arch/arm/src/amebad/amebad_serial.c
sunghan-chang: @thapav I am sorry but we don't want to apply our coding rules for arch, board and external folders.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -723,7 +742,7 @@ void up_serialinit(void)
  ****************************************************************************/
 void up_lowputc(char ch)
 {
-	LOGUART_PutChar(ch);
+	LOGUART_PutChar_RAM(ch);
```
Taejun-Kwon, line 745: can't understand why RAM has added as a postfix.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -374,6 +374,25 @@ static u32 uart_index_get(PinName tx)
...
+
+	if (ConfigDebugClose == 1)
+		return;
+
+	while(1) {
+		CounterIndex++;
+		if (CounterIndex >=6540)
+			break;
+
+		if (LOGUART->LSR & RUART_LINE_STATUS_REG_THRE)
```
Taejun-Kwon, line 390: Ok, here it checks status of holding register, we don't have to disable interrupt first, and enable it when holding register is empty to prevent data corruption by data in uart buffer?

AdamZhang0124:
> @AdamZhang0124 I don't know what the issue is and I can't find difference between new api and existed one because there is no code for existed. Could you share more details?

Hi @sunghan-chang,
This PR is actually to fix the issue that @amogh-samsung recently reported. The original function 'LOGUART_PutChar' adds an extra "CR" for every "LF" detected, and the new api has removed this action.

sunghan-chang:
@amogh-samsung Could you confirm the operation?

amogh-samsung:
Hi @AdamZhang0124 , I thought adding a new function was the temporary fix.
Is it not possible to fix this at the lower level in LOGUART_PutChar?
If not, may I ask why the code to append extra CR after LF was originally added and why it needs to be retained?

amogh-samsung:
Also, with LOGUART_PutChar and LOGUART_PutChar_RAM
We only need one of up_putc and up_lowputc right?
Duplications can be avoided.

os/arch/arm/src/amebad/amebad_serial.c
amogh-samsung: Yes please, me too.
Could you help us understand please @AdamZhang0124

AdamZhang0124:
> Hi @AdamZhang0124 , I thought adding a new function was the temporary fix.
> Is it not possible to fix this at the lower level in LOGUART_PutChar?
> If not, may I ask why the code to append extra CR after LF was originally added and why it needs to be retained?

No, as we discussed in the email, the LOGUART_PutChar is in rom code and it is originally used by Amebad, just like what up_putc is doing. But when both actions are applied, it will have "double CR" issue.

AdamZhang0124:
> Also, with LOGUART_PutChar and LOGUART_PutChar_RAM
> We only need one of up_putc and up_lowputc right?
> Duplications can be avoided.

Hi @amogh-samsung,
In the email you mentioned that you need to call up_lowputc directly and asked us to figure out the reason of the extra CR in **up_lowputc**. Therefore, I thought you need to achieve the following two actions：
1) If call up_putc --> add 'CR' for 'LF';
2) If call up_lowputc --> only output 'LF' without 'CR'

However, if you only expected one CR before LF when calling **up_putc** and you don't need to directly call **up_lowputc**. Then removing the 'CR adding' part' in the up_putc and keep LOGUART_PutChar unchanged will be the right way to do it.

If that so, let me know and I will modify the PR, thanks.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -374,6 +374,25 @@ static u32 uart_index_get(PinName tx)
 	return 3;
 }
 
+static void LOGUART_PutChar_RAM(u8 c)
+{
+	UART_TypeDef* LOGUART = UART2_DEV;
```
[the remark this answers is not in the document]
AdamZhang0124: Not only UART2_DEV is allowed for logUART in TizenRT. This part of code is inherited from LOGUART_PutChar, and ROM is using UART2 to print logs. I will change that, thanks

os/arch/arm/src/amebad/amebad_serial.c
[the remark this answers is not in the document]
AdamZhang0124: It provides a timeout to make sure the pervious char has been sent.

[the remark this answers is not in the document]
thapav: May be we can add this to a #define rather than hardcoding.
Say 
#define CHAR_TIMEOUT 6540

amogh-samsung:
> @amogh-samsung Could you confirm the operation?

Apologies for the delay. I will confirm the working with these changes ASAP.

os/arch/arm/src/amebad/amebad_serial.c
amogh-samsung: Apologies for the repetition,
Could you please clarify why the new function is named as "LOGUART_PutChar_**RAM**"?
The significance of the work RAM?

amogh-samsung:
My apologies for the delayed response please.

> Hi @amogh-samsung,
> In the email you mentioned that you need to call up_lowputc directly and asked us to figure out the reason of the extra CR in **up_lowputc**. Therefore, I thought you need to achieve the following two actions：
> 
> 1. If call up_putc --> add 'CR' for 'LF';
> 2. If call up_lowputc --> only output 'LF' without 'CR'

Yes, this is what we desire to achieve. Agreed

> However, if you only expected one CR before LF when calling **up_putc** and you don't need to directly call **up_lowputc**. Then removing the 'CR adding' part' in the up_putc and keep LOGUART_PutChar unchanged will be the right way to do it.
> 
> If that so, let me know and I will modify the PR, thanks.

Now, we have 2 use-cases

1. Add extra CR to LF -> for this case we call up_putc, 
2. Do NOT add extra CR -> for this we call up_lowputc, with this PR, it should work fine

up_putc adds 1 CR to LF and calls up_lowputc, so ultimately up_putc will also call LOGUART_PutChar_RAM.
Doesn't that put LOGUART_PutChar out of use?

I think we would like to retain the code for adding extra CR in up_putc for consistency with functions for other platforms.
@sunghan-chang @thapav could you provide your opinions/help clarify on this please?

AdamZhang0124:
> Now, we have 2 use-cases
> 
> 1. Add extra CR to LF -> for this case we call up_putc,
> 2. Do NOT add extra CR -> for this we call up_lowputc, with this PR, it should work fine
> 
> up_putc adds 1 CR to LF and calls up_lowputc, so ultimately up_putc will also call LOGUART_PutChar_RAM.
> Doesn't that put LOGUART_PutChar out of use?

Yes, the LOGUART_PutChar is currently out of use here, as it gives extra 'CR'. We use LOGUART_PutChar_RAM instead to achieve the 1. and 2. operations.

amogh-samsung:
The following are the console logs of the dump_tool when tested with rtl8721csm(loadable_apps)
1. Subsequent Handshakes are **successful** because there is no left over data in the buffer.
2. The Dumped UserFS file and the downloaded SmartFS files are **identical** (Unable to attach the image here, has been shared over email).
3. Since contents **match** (no extra CRs) and buffers are **clean**, the issue seems to be **resolved**

```
test@VirtualBox ~/workspace/TizenRT/tools/dump_tool (master) $./dump_tool.sh 
Please enter serial port adapter:
For example: /dev/ttyUSB1 or /dev/ttyACM0
Enter:
/dev/ttyUSB0
Target device locked and ready to DUMP!!!
Choose from the following options:-
1. RAM Dump
2. Userfs Dump
3. Both RAM and Userfs dumps
4. Exit Dump Tool
5. Reboot TARGET Device
2

DUMPING USERFS CONTENTS
do_handshake: Target Handshake successful

=========================================================================
Filesystem start address = 08106000, filesystem size = 1024000
=========================================================================

Receiving file system dump.....
[==============================================================>]

Filesystem Dump received successfully
Choose from the following options:-
1. RAM Dump
2. Userfs Dump
3. Both RAM and Userfs dumps
4. Exit Dump Tool
5. Reboot TARGET Device
1
DUMPING RAM CONTENTS
do_handshake: Target Handshake successful

=========================================================================
Ramdump Region Options:
1. ALL
2. Region 0:	( Address: 0x10056b60, Size: 132256 )	 [Heap index = 0]
3. Region 1:	( Address: 0x02000000, Size: 1048576 )	 [Heap index = 0]
4. Region 2:	( Address: 0x02100000, Size: 2883584 )	 [Heap index = 1]
=========================================================================

Please enter desired ramdump option as below:
 	1 for ALL
 	2 for Region 0
 	25 for Region 0 & 3 ...
Please enter your input: 1

ramdump_recv: No. of Regions to be dumped received

Receiving ramdump......

=========================================================================
Dumping Region: 0, Address: 0x10056b60, Size: 132256bytes
=========================================================================
[========>]
=========================================================================
Dumping Region: 1, Address: 0x02000000, Size: 1048576bytes
=========================================================================
[================================================================>]
=========================================================================
Dumping Region: 2, Address: 0x02100000, Size: 2883584bytes
=========================================================================
[================================================================================================================================================================================>]
Ramdump received successfully..!
Choose from the following options:-
1. RAM Dump
2. Userfs Dump
3. Both RAM and Userfs dumps
4. Exit Dump Tool
5. Reboot TARGET Device
3
DUMPING RAM CONTENTS
do_handshake: Target Handshake successful

=========================================================================
Ramdump Region Options:
1. ALL
2. Region 0:	( Address: 0x10056b60, Size: 132256 )	 [Heap index = 0]
3. Region 1:	( Address: 0x02000000, Size: 1048576 )	 [Heap index = 0]
4. Region 2:	( Address: 0x02100000, Size: 2883584 )	 [Heap index = 1]
=========================================================================

Please enter desired ramdump option as below:
 	1 for ALL
 	2 for Region 0
 	25 for Region 0 & 3 ...
Please enter your input: 2

ramdump_recv: No. of Regions to be dumped received

Receiving ramdump......

=========================================================================
Dumping Region: 0, Address: 0x10056b60, Size: 132256bytes
=========================================================================
[========>]
Ramdump received successfully..!

DUMPING USERFS CONTENTS
do_handshake: Target Handshake successful

=========================================================================
Filesystem start address = 08106000, filesystem size = 1024000
=========================================================================

Receiving file system dump.....
[==============================================================>]

Filesystem Dump received successfully
Choose from the following options:-
1. RAM Dump
2. Userfs Dump
3. Both RAM and Userfs dumps
4. Exit Dump Tool
5. Reboot TARGET Device
4
Dump tool exits after successful operation
test@VirtualBox ~/workspace/TizenRT/tools/dump_tool (master) $./dump_tool.sh 
Please enter serial port adapter:
For example: /dev/ttyUSB1 or /dev/ttyACM0
Enter:
/dev/ttyUSB0
[sudo] password for test: 
Target device locked and ready to DUMP!!!
Choose from the following options:-
1. RAM Dump
2. Userfs Dump
3. Both RAM and Userfs dumps
4. Exit Dump Tool
5. Reboot TARGET Device
5
NOTE: - CONFIG_BOARD_ASSERT_AUTORESET needs to be enabled to reboot TARGET Device after a crash
Handsake for rebooting device may fail since device reboots before sending acknowledgement
do_handshake: Target Handshake successful
Dump tool exits after successful operation

```

amogh-samsung (approved):
LGTM
Data comparison screenshot and logs shared.
Please let me know if any other checks are needed to be completed before merging this PR.

Taejun-Kwon (approved):

PR #4737, opened 2021-01-11
os/include/tinyara/mm : Move heapinfo_total_info_t into mm_heapinfo_utils.c

jeongchanKim:
heapinfo_total_info_t can be used only in mm_heapinfo_utils.c
so moved it.

### DIFF
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -350,9 +350,6 @@ extern struct heapinfo_group_s heapinfo_group[HEAPINFO_USER_GROUP_NUM];
 extern struct heapinfo_group_info_s group_info[HEAPINFO_THREAD_NUM];
 #endif
 
-#if CONFIG_KMM_NHEAPS > 1
-extern heapinfo_total_info_t total_info;
-#endif
 #endif
 /* This describes one heap (possibly with multiple regions) */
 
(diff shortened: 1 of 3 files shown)


os/include/tinyara/mm/mm.h
```diff
@@ -351,6 +351,18 @@ extern struct heapinfo_group_info_s group_info[HEAPINFO_THREAD_NUM];
...
+	int total_heap_size;
+	int cur_free;
+	int largest_free_size;
+	int cur_dead_thread;
+	int sum_of_stacks;
+	int sum_of_heaps;
+	int cur_alloc_size;
+	int peak_alloc_size;
+};
+typedef struct heapinfo_total_info_s heapinfo_total_info_t;
```
Taejun-Kwon, line 364: 1. It should changed to um.. like a public structure? parser is only caller who calling this structure, you can move it to there to use it as a private structure.
2. Some of values in this structure had better change its type to size_t instead of int to maintain more huge number of size & prevent error.

jeongchanKim: 1. In mm_heapinfo_utils.c, there is also a variable, `heapinfo_total_info_t total_info;`.
For using here, it should be declared in header.
2. Okay. I will check and change the type with another PR. Thanks.

Taejun-Kwon: > 1. In mm_heapinfo_utils.c, there is also a variable, `heapinfo_total_info_t total_info;`.
>    For using here, it should be declared in header.
Actually I couldn't see it in heapinfo_utils.c is it newly added?
Alright, anyway it still can be hided but not major structure, so let it be.

Taejun-Kwon: then you can remove below extern, I think

PR #4738, opened 2021-01-12
os/binfmt/libelf_addrenv : Change alignment size for allocating user …

jeongchanKim:
…binary when ARMv8M

ARMv8M doesn't require addresses to be aligned to the size of the region.
It can only be aligned with power of two.


jsdosa (approved):
Looks nice to me.

sunghan-chang (approved):

PR #4741, opened 2021-01-12
os/kernel/debug/memdbg : Add build condition when CONFIG_DEBUG_MM_WARN is enabled

jeongchanKim:
1) memdbg is only meaningful when CONFIG_DEBUG_MM_WARN is enabled.
2) Modify comment about heap configuration

### DIFF
--- a/os/kernel/debug/Make.defs
+++ b/os/kernel/debug/Make.defs
@@ -20,7 +20,10 @@ ifeq ($(CONFIG_DEBUG_SYSTEM),y)
 CSRCS += sysdbg.c
 endif
 
-CSRCS += memdbg.c dbg_termination_info.c
+CSRCS += dbg_termination_info.c
+ifeq ($(CONFIG_DEBUG_MM_WARN),y)
+CSRCS += memdbg.c
+endif
 
 ifeq ($(CONFIG_ENABLE_STACKMONITOR)$(CONFIG_DEBUG),yy)
 CSRCS += stackinfo_save_terminated.c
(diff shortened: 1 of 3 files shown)


os/kernel/debug/Make.defs
```diff
@@ -20,7 +20,10 @@ ifeq ($(CONFIG_DEBUG_SYSTEM),y)
 CSRCS += sysdbg.c
 endif
 
-CSRCS += memdbg.c dbg_termination_info.c
+CSRCS += dbg_termination_info.c
+ifeq ($(CONFIG_DEBUG_MM_WARN),y)
```
Taejun-Kwon, line 24: I think verbose(info) is correct for this
Usually WARN level is for potentially harmful situations.

jeongchanKim: Yes, but we used warn intentionally. Because mostly info is not enabled, but we want to check the information for debug.

Taejun-Kwon: warn is higher level than info, so if warn is enabled, info also need to be enabled. Most of platform follows below.
trace->debug->info->warn->error->fatal

sunghan-chang:
You need to check the caller. The caller calls it without the conditional. But when mm warning config is disabled, that function is not built. It will cause build break.

os/kernel/debug/Make.defs
sunghan-chang: Not for us and actually opposite way.
Fatal -> error -> warning ....
Because lower level requires more memory space.

jeongchanKim:
> You need to check the caller. The caller calls it without the conditional. But when mm warning config is disabled, that function is not built. It will cause build break.

Modified. Thanks.

sunghan-chang (approved):

PR #4743, opened 2021-01-12
docs: Add API list to be ported.

thapav:
Our porting guide does not provide entire api list to be ported
This document provides the api list, references, file location
and header file which has the API prototypes.
A link to the document has also been added to the TizenRT PortingGuide.

### DIFF
--- a/docs/APIListToBePorted.md
+++ b/docs/APIListToBePorted.md
@@ -1,0 +1,207 @@
+# API List to be Ported
+
+## Contents
+- [API List](#api-list)
+- [API Description](#api-description)
+
+## API List
+1. up_flashinitialize()
+2. up_idle()
+3. up_enable_irq()
+4. up_netinitialize()
+5. up_serialinit()
+6. up_earlyserialinit()
+7. up_putc()
+8. up_getc()
+9. up_puts()
+10. up_timer_initialize()
+11. up_wdog_init()
+12. up_wdog_keepalive()
+13. up_watchdog_disable()
+14. up_i2cinitialize()
+15. up_spiinitialize()
+16. up_rtc_initialize()
+
+## API Description
+
+- Add low level chip specific API **up_flashinitialize()** to create an initialize MTD device instance. MTD devices are not registered in the file system, but are created as instances that can be bound to other functions (such as a block or character driver front end).
+```c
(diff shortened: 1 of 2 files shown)


docs/APIListToBePorted.md
```diff
@@ -0,0 +1,207 @@
...
+ * Input Parameters:
+ *   none
+ * Returned Value:
+ *   Valid MTD device structure reference on success; a NULL on failure
+```
+*File Location:*	"os/arch/arm/src/chip/chip_flash.c"
+
+*Prototype Location:*	"os/include/tinyara/fs/mtd.h"
+
+- Add low level chip specific API **up_idle()** to be executed when their is no other ready-to-run task. This is processor idle time and will continue until some interrupt occurs to cause a context switch from the idle task. Processing in this state may be processor-specific. e.g., this is where power management operations might be performed.
```
jeongchanKim, line 39: How about changing the line start from `e.g. this is where ~~~`?

docs/APIListToBePorted.md
```diff
@@ -0,0 +1,207 @@
...
+a. **up_putc()** : Output one byte on the serial console
+```c
+ * Prototype: int up_putc(int ch)
+ * Input Parameters:
+ *   ch - chatacter to output
+ * Returned Value:
+ *   sent character
+```
+b. **up_getc()** : Read one byte from the serial console
+```
```
jeongchanKim, line 115: We'd better marking this box as c language like line 107.

docs/APIListToBePorted.md
```diff
@@ -0,0 +1,207 @@
...
+ * Prototype: int up_getc(void)
+ * Input Parameters:
+ *   none
+ * Returned Value:
+ *   int value, -1 if error, 0~255 if byte successfully read
+```
+c. **up_puts()** : Output string on the serial console
+```
+ * Prototype: void up_puts(const char *str)
```
jeongchanKim, line 124: Same as above.

thapav: done

docs/APIListToBePorted.md
thapav: done

sunghan-chang (approved):

PR #4748, opened 2021-01-18
board/common: Manage kernel data for kernel binary update

jeongarmy:
- Manage kernel data for kernel binary update
 . version, index in use, partition size, num, count
- Support getinfo functions to get kernel data
 . partitions_get_kcount
 . partitions_get_kversion
 . partitions_get_download_kinfo

### DIFF
--- a/os/board/common/partitions.c
+++ b/os/board/common/partitions.c
@@ -30,6 +30,7 @@
 
 #include <tinyara/fs/mtd.h>
 #include <tinyara/fs/ioctl.h>
+
 #ifdef CONFIG_BINARY_MANAGER
 #include <tinyara/binary_manager.h>
 #endif
@@ -85,7 +86,7 @@ static void type_specific_initialize(int minor, FAR struct mtd_dev_s *mtd_part, 
 		|| !strncmp(types, "romfs,", 6)
 #endif
 #ifdef CONFIG_BINARY_MANAGER
-		|| !strncmp(types, "kernel,", 7) || !strncmp(types, "bin,", 4)
+		|| !strncmp(types, "kernel,", 7)
 #endif
 		) {
 			if (ftl_initialize(partno, mtd_part)) {
@@ -211,13 +212,13 @@ int configure_mtd_partitions(struct mtd_dev_s *mtd, struct partition_data_s *par
 
 		type_specific_initialize(minor, mtd_part, partno, types);
 
-#ifdef CONFIG_MTD_PARTITION_NAMES
-		configure_partition_name(mtd_part, (const char **)&names, &index, part_name);
 #ifdef CONFIG_BINARY_MANAGER
 		if (!strncmp(types, "kernel,", 7)) {
 			binary_manager_register_kpart(partno, partsize);
 		}
 #endif
+#ifdef CONFIG_MTD_PARTITION_NAMES
+		configure_partition_name(mtd_part, (const char **)&names, &index, part_name);
 #endif
 		move_to_next_part((const char **)&sizes);
 		move_to_next_part((const char **)&types);
--- a/os/kernel/binary_manager/Kconfig
+++ b/os/kernel/binary_manager/Kconfig
@@ -6,8 +6,8 @@
 config BINARY_MANAGER
 	bool "Enable Binary Manager"
 	default n
-	depends on APP_BINARY_SEPARATION && BINFMT_LOADABLE && !DISABLE_MQUEUE && !DISABLE_SIGNAL
-	depends on MTD_FTL && FLASH_PARTITION && MTD_PARTITION_NAMES
+	depends on !DISABLE_MQUEUE && !DISABLE_SIGNAL
+	depends on MTD_FTL && FLASH_PARTITION
 	---help---
 		This is kernel thread which manages binaries.
 		It loads/unloads binaries and recovers any fault occurs in a system.
@@ -29,6 +29,7 @@ config BM_PRIORITY_MIN
 config BINMGR_RECOVERY
 	bool "Enable Recovery Management"
 	default y
+	depends on APP_BINARY_SEPARATION && BINFMT_LOADABLE
 	---help---
 		Enable the recovery management.
 		If any fault occurs in a system, it will either restart binary or reboot the system
@@ -37,6 +38,9 @@ config BINMGR_RECOVERY
 config BINMGR_UPDATE
 	bool "Enable Binary Update"
 	default y
+	select BCH
+	select LIB_BOARDCTL
+	select BOARDCTL_RESET
 	---help---
 		Enables Binary Manager Update APIs.
 
--- a/os/kernel/binary_manager/binary_manager.c
+++ b/os/kernel/binary_manager/binary_manager.c
@@ -28,6 +28,7 @@
 #include <string.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <sys/boardctl.h>
 
 #include <tinyara/binary_manager.h>
 #ifdef CONFIG_BINMGR_RECOVERY
@@ -41,6 +42,11 @@
 /* The largest size of messages is (CONFIG_NAME_MAX + 64) for communication with binary manager */
 #error "CONFIG_MQ_MAXMSGSIZE should be greater than (CONFIG_NAME_MAX + 64)"
 #endif
+#ifndef CONFIG_BOARDCTL_RESET
+/* CONFIG_BOARDCTL_RESET should be enabled for kernel binary update */
+#error "CONFIG_BOARDCTL_RESET should be enabled for kernel binary update"
+#endif
+
 
 /****************************************************************************
  * Private Definitions
@@ -74,18 +80,21 @@ mqd_t binary_manager_get_mqfd(void)
  ****************************************************************************/
 int binary_manager(int argc, char *argv[])
 {
-	int ret;
 	int nbytes;
 #if !defined(CONFIG_DISABLE_SIGNALS)
 	sigset_t sigset;
 #endif
 	struct mq_attr attr;
 	binmgr_request_t request_msg;
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	int ret;
 
 	/* Scan user binary files and Register them */
 	binary_manager_scan_ubin_all();
 
-	ASSERT(binary_manager_get_kcount() > 0 && binary_manager_get_ucount() > 0);
+	ASSERT(binary_manager_get_ucount() > 0);
+#endif
+	ASSERT(binary_manager_get_kcount() > 0)
 
 	bmvdbg("Binary Manager STARTED\n");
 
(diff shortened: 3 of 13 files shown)


os/board/common/partitions.c
```diff
@@ -119,6 +122,24 @@ static void move_to_next_part(const char **par)
 		while (*(*par)++ != ',');
 }
 
+static void partitions_register_kernel(int part_num, int part_size)
```
deepaksrma, line 125: can you write some description for this function?

os/include/tinyara/partitions.h
```diff
@@ -0,0 +1,69 @@
...
+
+/* The devname format for kernel partition */
+#define KERNEL_DEVNAME_FMT         "/dev/mtdblock%d"
+
+/* The sturcture of partition information */
+struct part_info_s {
+	uint32_t part_size;         /* Partition size */
+	int8_t part_num;            /* Partition number */
+};
+typedef struct part_info_s part_info_t;
```
jeongchanKim, line 47: If this type is not used without `struct kernel_data_s`, we don't need to make a structure. We can merge into `struct kernel_data_s`.

os/include/tinyara/partitions.h
```diff
@@ -0,0 +1,69 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
```
jeongchanKim, line 3: Please change the license year.

os/board/common/partitions.c
```diff
@@ -146,6 +167,34 @@ static void configure_partition_name(FAR struct mtd_dev_s *mtd_part, const char
...
+
+uint32_t partitions_get_kcount(void)
+{
+	return kernel_data.part_count;
+}
+
+int partitions_get_download_kdata(kdata_download_t *kdata)
+{
+	if (kdata == NULL) {
+		lldbg("kdata is NULL\n");
```
jeongchanKim, line 183: We cannot understand the kdata. Please leave comment in more detail.

os/board/common/partitions.c
```diff
@@ -146,6 +167,34 @@ static void configure_partition_name(FAR struct mtd_dev_s *mtd_part, const char
...
+		return ERROR;
+	}
+
+	if (kernel_data.part_count > 1) {
+		kdata->available_size = kernel_data.part_info[kernel_data.inuse_idx ^ 1].part_size;
+		snprintf(kdata->path, KERNEL_PATH_MAX, KERNEL_DEVNAME_FMT, kernel_data.part_info[kernel_data.inuse_idx ^ 1].part_num);
+	} else {
+		kdata->available_size = 0;
+		lldbg("No additional partition to download new binary\n");
```
jeongchanKim, line 192: In this case, there is no additional partition, but returns `OK`. Is it okay?

os/board/common/partitions.c
```diff
@@ -211,13 +260,12 @@ int configure_mtd_partitions(struct mtd_dev_s *mtd, struct partition_data_s *par
 
 		type_specific_initialize(minor, mtd_part, partno, types);
 
-#ifdef CONFIG_MTD_PARTITION_NAMES
-		configure_partition_name(mtd_part, (const char **)&names, &index, part_name);
-#ifdef CONFIG_BINARY_MANAGER
 		if (!strncmp(types, "kernel,", 7)) {
-			binary_manager_register_kpart(partno, partsize);
+			partitions_register_kernel(partno, partsize);
```
thapav, line 264: If this API fails, it just returns with invalid part info logs (no return value)
Do we not need to take any further action here for kernel configuration failure and move on to next partition ?

os/kernel/binary_manager/binary_manager.c
```diff
@@ -143,14 +157,25 @@ int binary_manager(int argc, char *argv[])
 		case BINMGR_GET_INFO_ALL:
 			binary_manager_get_info_all(request_msg.requester_pid);
 			break;
-		case BINMGR_GET_STATE:
-			binary_manager_get_state_with_name(request_msg.requester_pid, (char *)request_msg.data.bin_name);
+		case BINMGR_CREATE_BIN:
+			binary_manager_create_entry(request_msg.requester_pid, request_msg.data.update_bin.bin_name, request_msg.data.update_bin.version);
 			break;
 		case BINMGR_UPDATE:
+			if (!strncmp("kernel", request_msg.data.bin_name, BIN_NAME_MAX)) {
```
sunghan-chang, line 164: In flat build, do we need to check this name?
If it is, then you need to add failure case. I mean that binary name is not "kernel".
If it isn't, then you can move this conditional into CONFIG_APP_BINARY_SEPARATION.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -49,35 +54,62 @@
 /* Binary table, the first data [0] is for Common Library. */
 static binmgr_uinfo_t bin_table[USER_BIN_COUNT + 1];
 static uint32_t g_bin_count;
+#endif
 
-/* Data for Kernel partitions */
+/* Data for Kernel partitions */
```
sunghan-chang, line 59: Could you remove a tab at end of line?

os/kernel/binary_manager/binary_manager.c
jeongarmy: Yes. I think we need to check name for invalid string on flat build.
I agree with your opinion but currently it doesn't return a result. I will think more.

os/kernel/binary_manager/binary_manager_binfile.c
```diff
@@ -222,12 +227,14 @@ int binary_manager_scan_ubin(int bin_idx)
  ****************************************************************************/
 int binary_manager_create_entry(int requester_pid, char *bin_name, int version)
 {
+#ifdef CONFIG_APP_BINARY_SEPARATION
```
jeongchanKim, line 230: This file is built only when `CONFIG_BINARY_MANAGER` is enabled. 
And `CONFIG_BINARY_MANAGER` has dependency with `CONFIG_APP_BINARY_SEPARATION`.
You'd better check the dependency of all configs related to binary manager and app separation.

os/kernel/binary_manager/binary_manager.c
```diff
@@ -41,6 +42,11 @@
 /* The largest size of messages is (CONFIG_NAME_MAX + 64) for communication with binary manager */
 #error "CONFIG_MQ_MAXMSGSIZE should be greater than (CONFIG_NAME_MAX + 64)"
 #endif
+#ifndef CONFIG_BOARDCTL_RESET
+/* CONFIG_BOARDCTL_RESET should be enabled for kernel binary update */
+#error "CONFIG_BOARDCTL_RESET should be enabled for reboot"
```
sunghan-chang, line 47: for reboot -> for kernel binary update

os/kernel/binary_manager/Kconfig
```diff
@@ -6,8 +6,10 @@
 config BINARY_MANAGER
 	bool "Enable Binary Manager"
 	default n
-	depends on APP_BINARY_SEPARATION && BINFMT_LOADABLE && !DISABLE_MQUEUE && !DISABLE_SIGNAL
-	depends on MTD_FTL && FLASH_PARTITION && MTD_PARTITION_NAMES
+	depends on !DISABLE_MQUEUE && !DISABLE_SIGNAL
+	depends on BCH && MTD_FTL && FLASH_PARTITION
+	select LIB_BOARDCTL
+	select BOARDCTL_RESET
```
sunghan-chang, line 12: Is here proper location? not BINMGR_UPDATE config?

os/kernel/binary_manager/binary_manager.c
```diff
@@ -161,16 +185,12 @@ int binary_manager(int argc, char *argv[])
...
 		default:
+			bmvdbg("Invalid cmd = %d\n", request_msg.cmd);
 			break;
 		}
 	}
 
-binary_manager_exit:
-	bmvdbg("Binary Manager EXITED\n");
-
-	mq_close(g_binmgr_mq_fd);
```
jeongchanKim, line 172: Is it okay not to close and unlink the mq? I know there is no normal case that binary manager is terminated, but it can be happened because of something not common. In this case, there will be kernel resource leak.

os/kernel/binary_manager/binary_manager.c
```diff
@@ -74,18 +80,25 @@ mqd_t binary_manager_get_mqfd(void)
  ****************************************************************************/
 int binary_manager(int argc, char *argv[])
 {
+#ifdef CONFIG_APP_BINARY_SEPARATION
 	int ret;
```
jeongchanKim, line 84: If we move this declaration under `request_msg`, we can merge the ifdef with `binary_manager_scan_ubin_all`.
It looks much simple.

os/kernel/binary_manager/binary_manager.c
```diff
@@ -74,18 +80,25 @@ mqd_t binary_manager_get_mqfd(void)
...
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
 	/* Scan user binary files and Register them */
 	binary_manager_scan_ubin_all();
+#endif
 
-	ASSERT(binary_manager_get_kcount() > 0 && binary_manager_get_ucount() > 0);
+	ASSERT(binary_manager_get_kcount() > 0)
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	ASSERT(binary_manager_get_ucount() > 0);
```
jeongchanKim, line 100: If we can check the user binary count first, then we also can merge the ifdef with the above.

os/kernel/binary_manager/binary_manager_getinfo.c
```diff
@@ -194,22 +201,36 @@ void binary_manager_get_info_with_name(int requester_pid, char *bin_name)
...
+		}
+		response_msg.result = BINMGR_OK;
+	}
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	else {
+		bin_count = binary_manager_get_ucount();
+		for (bin_idx = 1; bin_idx <= bin_count; bin_idx++) {
+			if (!strncmp(BIN_NAME(bin_idx), bin_name, BIN_NAME_MAX)) {
+				size = binary_manager_get_available_size(bin_idx);
+				if (size < 0) {
```
jeongchanKim, line 221: If the available size is 0, is it okay not to handle as fail?

jeongarmy: Yes. a size can be 0 on success(BINMGR_OK). It means operations are successful but a available space is not existing.

os/kernel/binary_manager/binary_manager.c
jeongarmy: yes. right. but there is no case that a loop for request messages of  binary manager although binary manager is failed.

sunghan-chang (approved):

PR #4745, opened 2021-01-13
os/board/rtl8721csm : Modify wifi driver skbuff to get better performance

keerthanalex:
Modified skbuff to 40 from 56 to get better performance
Updated wlan and wps libs with/without fpu

PR #4750, opened 2021-01-19
os/board/rtl8721csm : Tune Wi-Fi driver RX buffer count to get better performance for weak signal

keerthanalex:
Modified RX buffer count to get better performance for weak signal
Modified the condition of power save threshold to packet count
Updated wlan and wps libs with/without fpu


jeongchanKim (approved):
LGTM

sunghan-chang (approved):

keerthanalex:
Includes the change from #4745 as well.

keerthanalex:
@thapav, Please find the Wi-Fi throughput result as below

Shield Room | Latest(before #4750) | PR #4750
-- | -- | --
RX   throughput (Mbps) | ~29 | ~30

PR #4755, opened 2021-01-21
configs/rtl8721csm: set secure storage partition name and skip that at download

sunghan-chang:
1. config/rtl8721csm/download.sh: skip non-binary partition 
2. configs/rtl8721csm/*: set partition name for secure storage


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4762, opened 2021-01-26
os/mm/mm_heap/mm_getheap.c : Fix string copy size

jeongchanKim:
For security issue, it is better to copy the string based on its size.

### DIFF
--- a/os/mm/mm_heap/mm_getheap.c
+++ b/os/mm/mm_heap/mm_getheap.c
@@ -84,7 +84,8 @@ void mm_add_app_heap_list(struct mm_heap_s *heap, char *app_name)
 
 	node->is_active = true;
 	node->heap = heap;
-	strncpy(node->app_name, app_name, BIN_NAME_MAX);
+	strncpy(node->app_name, app_name, BIN_NAME_MAX - 1);
+	node->app_name[BIN_NAME_MAX - 1] = '\0';
 
 	/* Add the new heap node to the head of the list*/
 	dq_addfirst((dq_entry_t *)node, &app_heap_q);
--- a/apps/system/utils/utils_heapinfo.c
+++ b/apps/system/utils/utils_heapinfo.c
@@ -229,7 +229,8 @@ int utils_heapinfo(int argc, char **args)
 #ifdef CONFIG_APP_BINARY_SEPARATION
 		case 'b':
 			options.heap_type = HEAPINFO_HEAP_TYPE_BINARY;
-			strncpy(options.app_name, optarg, BIN_NAME_MAX);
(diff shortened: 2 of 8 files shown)


jeongarmy (dismissed):
LGTM

os/mm/mm_heap/mm_getheap.c
```diff
@@ -84,7 +84,7 @@ void mm_add_app_heap_list(struct mm_heap_s *heap, char *app_name)
 
 	node->is_active = true;
 	node->heap = heap;
-	strncpy(node->app_name, app_name, BIN_NAME_MAX);
+	strncpy(node->app_name, app_name, strlen(app_name) + 1);
```
sunghan-chang, line 87: I don't know why this change is necessary. I think previous is right.

os/mm/mm_heap/mm_getheap.c
```diff
@@ -144,7 +145,7 @@ struct mm_heap_s *mm_get_app_heap_with_name(char *app_name)
 
 	/* Search the heap */
 	while (node && node->is_active) {
-		if (strncmp(node->app_name, app_name, BIN_NAME_MAX) == 0) {
+		if (strncmp(node->app_name, app_name, BIN_NAME_MAX + 1) == 0) {
```
Taejun-Kwon, line 148: last is always null, BIN_NAME_MAX is correct, I think.

jeongchanKim: Let's assume that `node->app_name` is "abc", `app_name` is "abcd" and BIN_NAME_MAX is 3.
Then this comparison returns OK, because it tries to check from index 0 to 2.
But it is not correct. 
So I think `BIN_NAME_MAX + 1` is correct.

Taejun-Kwon: OK

an4967 (dismissed):
LGTM

Taejun-Kwon (dismissed):

os/mm/mm_heap/mm_getheap.c
```diff
@@ -43,7 +43,7 @@
 typedef struct app_heap_s {
 	struct app_heap_s *flink;
 	struct app_heap_s *blink;
-	char app_name[BIN_NAME_MAX];
```
sunghan-chang, line 46: There are many locations which use BIN_NAME_MAX for array size. But you modify only here. If this is necessary, you should check all of others. But is this really necessary?

jsdosa (approved):
Looks nice to me.

jsdosa (approved):
Looks nice to me

PR #4764, opened 2021-01-27
drivers/kernel_test: Get an address of uheap from tcb structure instead of BIN_LOADINFO

jeongarmy:
Get an address of uheap from tcb structure instead of BIN_LOADINFO.
That's because a BIN_LOADINFO() is defined when CONFIG_OPTIMIZE_APP_RELOAD_TIME is enabled only.

### DIFF
--- a/os/drivers/kernel_test/kernel_test_drv.c
+++ b/os/drivers/kernel_test/kernel_test_drv.c
@@ -153,13 +153,17 @@ static int kernel_test_drv_ioctl(FAR struct file *filep, int cmd, unsigned long 
 			* which belongs to any other app in the system. Here,
 			* we choose to return the address of the app heap
 			*/
-			uint32_t binid = sched_self()->group->tg_binidx;
-			binid = (binid + 1) % (binary_manager_get_ucount() + 1);
-			if (binid == 0) {
-				binid++;
+			uint32_t binidx = sched_self()->group->tg_binidx;
+			binidx = (binidx + 1) % (binary_manager_get_ucount() + 1);
+			if (binidx == 0) {
+				binidx++;
 			}
-
-			obj->addr = (volatile uint32_t *)BIN_LOADINFO(binid)->uheap;
+			struct tcb_s *tcb = (struct tcb_s *)sched_gettcb(BIN_ID(binidx));
+			if (tcb) {
+				obj->addr = (volatile uint32_t *)tcb->uheap;
+			} else {
+				ret = -ESRCH;
+			}
 			break;
 		}
 		default:


os/drivers/kernel_test/kernel_test_drv.c
```diff
@@ -153,13 +153,17 @@ static int kernel_test_drv_ioctl(FAR struct file *filep, int cmd, unsigned long
...
+			uint32_t binidx = sched_self()->group->tg_binidx;
+			binidx = (binidx + 1) % (binary_manager_get_ucount() + 1);
+			if (binidx == 0) {
+				binidx++;
+			}
+			struct tcb_s *tcb = (struct tcb_s *)sched_gettcb(BIN_ID(binidx));
+			if (tcb) {
+				obj->addr = (volatile uint32_t *)tcb->uheap;
+			} else {
+				ret = -EINVAL;
```
Taejun-Kwon, line 165: -EINVAL is returned when error occurs because of invalid parameter of function, so can't be acceptable here.
If there's no pid or process group, -ESRCH is good option, but I am not sure it is right value for here.

os/drivers/kernel_test/kernel_test_drv.c
```diff
@@ -153,13 +153,17 @@ static int kernel_test_drv_ioctl(FAR struct file *filep, int cmd, unsigned long
...
-			uint32_t binid = sched_self()->group->tg_binidx;
-			binid = (binid + 1) % (binary_manager_get_ucount() + 1);
-			if (binid == 0) {
-				binid++;
+			uint32_t binidx = sched_self()->group->tg_binidx;
+			binidx = (binidx + 1) % (binary_manager_get_ucount() + 1);
+			if (binidx == 0) {
+				binidx++;
+			}
+			struct tcb_s *tcb = (struct tcb_s *)sched_gettcb(BIN_ID(binidx));
```
jeongchanKim, line 161: It looks that we want to set `obj->addr` as its user heap address.
Every tcbs in a binary has its user heap address.
So I think we don't need to find the parent.
We can get the heap address with its own like below.
```
struct tcb_s *tcb = sched_self();
if (tcb) {
    obj->addr = (volatile uint32_t *)tcb->uheap;
}
```

jeongarmy: It is used to get the heap address of another binary so it gets a index of another binary here.
You can find the usage in mpu_test.c:58 and recovery:67.

deepaksrma (approved):
LGTM

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4765, opened 2021-01-28
heapinfo, messaging : Fix copy size and add '\0' to the end of string array

jeongarmy:
Fix string copy size and add '\0' to the end of string array for security issue.

### DIFF
--- a/apps/system/utils/utils_heapinfo.c
+++ b/apps/system/utils/utils_heapinfo.c
@@ -229,7 +229,8 @@ int utils_heapinfo(int argc, char **args)
 #ifdef CONFIG_APP_BINARY_SEPARATION
 		case 'b':
 			options.heap_type = HEAPINFO_HEAP_TYPE_BINARY;
-			strncpy(options.app_name, optarg, BIN_NAME_MAX);
+			strncpy(options.app_name, optarg, BIN_NAME_MAX - 1);
+			options.app_name[BIN_NAME_MAX - 1] = '\0';
 			heap_name = options.app_name;
 			break;
 #endif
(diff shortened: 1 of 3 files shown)


sijupunnnoose (dismissed):
LGTM

jeongchanKim (dismissed):
LGTM

apps/system/utils/utils_heapinfo.c
```diff
@@ -229,7 +229,8 @@ int utils_heapinfo(int argc, char **args)
 #ifdef CONFIG_APP_BINARY_SEPARATION
 		case 'b':
 			options.heap_type = HEAPINFO_HEAP_TYPE_BINARY;
-			strncpy(options.app_name, optarg, BIN_NAME_MAX);
+			strncpy(options.app_name, optarg, BIN_NAME_MAX - 1);
+			options.app_name[BIN_NAME_MAX - 1] = '\0';
```
sunghan-chang, line 233: Do you confirm that `optarg` has `BIN_NAME_MAX - 1` size? If not, it will have garbage value in the name.

jeongchanKim (commented):
Please check the `killall` also.

apps/system/utils/utils_heapinfo.c
jeongarmy: Actually it is meaningful when `optarg` is larger than `BIN_NAME_MAX - 1 only.
But I added this to avoid calling strlen every time for performance.

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4766, opened 2021-01-28
Reboot reason

jeongchanKim:


os/kernel/task/task_prctl.c
```diff
@@ -347,6 +351,28 @@ int prctl(int option, ...)
...
+	case PR_REBOOT_REASON_READ:
+	{
+		va_end(ap);
+		return reboot_reason_read();
+	}
+	case PR_REBOOT_REASON_WRITE:
+	{
+		reboot_reason_code_t reason = (reboot_reason_code_t)va_arg(ap, int);
+
+		reboot_reason_write(reason);
```
sijupunnnoose, line 365: Should we check the return value of reboot_reason_write? Function definition has a return value, can i assume write can fail?

os/kernel/task/task_prctl.c
```diff
@@ -347,6 +351,28 @@ int prctl(int option, ...)
...
+
+		va_end(ap);
+		return OK;
+	}
+	case PR_REBOOT_REASON_CLEAR:
+	{
+		reboot_reason_clear();
```
sijupunnnoose, line 372: ditto, else make function return void.

os/kernel/task/task_prctl.c
jeongchanKim: I modified the prototype as void return. Thanks.

os/kernel/task/task_prctl.c
jeongchanKim: I modified the prototype as void return. Thanks.

os/board/Kconfig
```diff
@@ -300,6 +300,15 @@ config BOARD_DUMP_UART
 		If selected it will allow sending RAM and FS dump via UART.
 endif
 
+config SYSTEM_REBOOT_REASON
```
sunghan-chang, line 303: Code and header are in arch, but config is in board. It looks strange. Let's move this into arch.

os/include/tinyara/reboot_reason.h
```diff
@@ -0,0 +1,34 @@
...
+ * Included Files
+ ********************************************************************************************/
+
+typedef enum {
+	REBOOT_SYSTEM_WATCHDOG		= 0, /* watchdog timeout */        
+	REBOOT_SYSTEM_DATAABORT		= 1, /* Data abort */
+	REBOOT_SYSTEM_PREFETCHABORT	= 2, /* Prefetch abort */
+	REBOOT_SYSTEM_MEMORYALLOCFAIL	= 3, /* Memory allocation failure */
+	REBOOT_SYSTEM_HW_RESET		= 4, /* HW power reset */
+	REBOOT_UNKNOWN 	 		= 32
```
sunghan-chang, line 31: Let's add two more reasons.
```
REBOOT_SYSTEM_WIFICORE_WATCHDOG = 11,
REBOOT_SYSTEM_WIFICORE_PANIC    = 12,
```

jeongchanKim: Added. Thanks.

sunghan-chang (approved):

PR #4771, opened 2021-02-03
os/board/rtl8721csm: Add Boot Reason

steamed-stuffed-bread:


os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -56,10 +56,15 @@
 #include <tinyara/config.h>
 
 #include <tinyara/reboot_reason.h>
+#include "ameba_soc.h"
 /****************************************************************************
  * Public functions
  ****************************************************************************/
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
+
+#define UPPER_REBOOT_SYSTEM_WATCHDOG	BIT(0)
```
jeongchanKim, line 65: How about define like below? Because it should be matched with common enum value in `tinyara/reboot_reason.h`
```
#define UPPER_REBOOT_SYSTEM_WATCHDOG BIT(REBOOT_SYSTEM_WATCHDOG)
```

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -70,12 +75,38 @@ void up_reboot_reason_init(void)
...
+	if (boot_reason == 0) {
+		temp_bootcfg |= UPPER_REBOOT_SYSTEM_HW_RESET;
+	}
+
+	/* KM4 or KM0 WDT reset */
+	else if ((boot_reason & BIT_BOOT_KM4WDG_RESET_HAPPEN) || (boot_reason & BIT_BOOT_WDG_RESET_HAPPEN)) {
+		temp_bootcfg |= UPPER_REBOOT_SYSTEM_WATCHDOG;
+	}
+
+	temp_bootcfg |= backup_reg;
```
jeongchanKim, line 98: In this case, is `temp_bootcfg` matched with `reboot_reason_code_t` in `tinyara/reboot_reason.h`?

steamed-stuffed-bread: Yes

jeongchanKim: Ok. Thanks.

steamed-stuffed-bread: I see that in reboot_reason.h, REBOOT_SYSTEM_HW_RESET is defined to be BIT(4).

jeongchanKim: Oh sorry, I mean for `REBOOT_SYSTEM_WATCHDOG` case.

steamed-stuffed-bread: Some boot reason you provided in reboot_reason.h is not supported by us. We thought that you want to write your own boot reason and read, that's why we provide BKUP_REG to achieve it. The common cases we are both handling are:
BIT_BOOT_KM4WDG_RESET_HAPPEN BIT_BOOT_WDG_RESET_HAPPEN - REBOOT_SYSTEM_WATCHDOG
HW RESET - REBOOT_SYSTEM_HW_RESET

That is why we have the if-else if in up_reboot_reason_read.
Thanks.

jeongchanKim: Okay. Thanks.

jeongchanKim (approved):
LGTM

jeongchanKim (commented):
I agreed the whole changes in this file, but we don't need to separate the commit, I think. 
Please merge the commits.

sunghan-chang:
```
jeongchanKim left a comment
I agreed the whole changes in this file, but we don't need to separate the commit, I think.
Please merge the commits.
```
@jeongchanKim @steamed-stuffed-bread I will use `squash and merge` option to merge this PR so that you don't need to consider above. (It will be one commit by system)

os/arch/arm/src/amebad/amebad_reboot_reason.c
sunghan-chang: In our usecase, we don't take combined two more reason. When we call `up_reboot_reason_read`, it should return only one value which is defined in tinyara/reboot_reason.h. Because reboot reason must be one.
The `sys_reset` Realtek provided is working with watchdog. The system will be reset by watchdog, but it's not real reboot reason. Because of some other reason, something calls sys_reset to reset. At that case, we will write the reason through `up_reboot_reason_write`.
Application should be the the same regardless of board so that please make this API to return just one reason.

sunghan-chang: Please find our API prototype. `reboot_reason_code_t up_reboot_reason_read(void)`. It's not u32 for return.

steamed-stuffed-bread: Does that mean we don't need to deal with the BIT_BOOT_KM4WDG_RESET_HAPPEH and the BIT_BOOT_WDG_RESET_HAPPEN and just leave it to you to take care of? How about HW reset? Do we set bit 4 or return 0 directly?

sunghan-chang: For HW Reset, 4. (Not BIT(4))
```
REBOOT_SYSTEM_HW_RESET		= 4,	/* HW power reset */
```
For Watchdog time out, 0. (Not Bit(0) to set LSB)
```
REBOOT_SYSTEM_WATCHDOG		= 0,	/* watchdog timeout */  
```

sunghan-chang: I think when `backup_reg = BKUP_Read(BKUP_REG1);` has some value, you can return it directly without checking `boot_reason = BOOT_Reason();`. When backup does not have, then you can check boot reason and then can return.

But we don't clear it all the time, simply reading backup and then boot reason could be return old reason.
How about reading the reason using the init function and saving it in static global variable? And read function returns just static global. Then we can prevent reading old reason.

steamed-stuffed-bread: Thank you for your suggestions, I will try it out.

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -56,26 +56,50 @@
...
 
 }
 
 reboot_reason_code_t up_reboot_reason_read(void)
 {
-	return 0;
+	u32 boot_reason = 0;
+
+	if (backup_reg) {
+
```
jeongchanKim, line 79: Please remove unnecessary empty line.

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -56,26 +56,50 @@
...
 #include <tinyara/reboot_reason.h>
+#include "ameba_soc.h"
 /****************************************************************************
  * Public functions
  ****************************************************************************/
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
+
+static reboot_reason_code_t backup_reg;
+
 void up_reboot_reason_init(void)
```
jeongchanKim, line 67: In this case, we'd better call the `up_reboot_reason_init()` when board bring up.
Please add it.

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -56,26 +56,50 @@
...
 {
-	/* If chip vendor needs to initialize for using reboot reason
-	 * chip specific initialization code will be here.
-	 */
+	/* Read the same backup register for the boot reason */
+	backup_reg = BKUP_Read(BKUP_REG1);
```
sunghan-chang, line 70: 1. After reading, it should be clear.
2. This init function should be called from some initialization function like board_initialize.

os/board/rtl8721csm/src/rtl8721csm_boot.c
```diff
@@ -236,6 +236,11 @@ void board_initialize(void)
 		wifi_config.wifi_app_ctrl_tdma == FALSE) {
 		SystemSetCpuClk(CLK_KM4_100M);
 	}
+
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	up_reboot_reason_init();
```
jeongchanKim, line 241: Please include the `reboot_reason.h` which is in os/arch/arm/include.

steamed-stuffed-bread: Is it <arch/reboot_reason.h> or <tinyara/reboot_reason.h> ? 
Thanks

sunghan-chang: <arch/reboot_reason.h>

jeongchanKim: I mean in arch. Because the prototype of `up_reboot_reason_init()` is in arch.

steamed-stuffed-bread: OK, thank you all.

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #4772, opened 2021-02-03
arch/arm/amebad_irq : Add reboot reason write for every fault cases

jeongchanKim:
In every fault cases, write the reboot reason as prefetch abort before calling PANIC.


sunghan-chang (commented):
up_hardfault and up_memfault are missing

jeongchanKim:
> up_hardfault and up_memfault are missing

Updated. Thanks.

sunghan-chang (approved):

PR #4774, opened 2021-02-04
apps/examples: Fix build failure with tc configuration

sangam-swami:
Fix build failure on seclink test case.

### DIFF
--- a/apps/examples/security_test/seclink/sl_test_utils.c
+++ b/apps/examples/security_test/seclink/sl_test_utils.c
@@ -55,7 +55,7 @@ int sl_test_malloc_buffer(hal_data *data, int buf_len)
 	if (!data->priv) {
 		free(data->data);
 		data->data_len = 0;
-		return -3
+		return -3;
 	}
 	data->priv_len = buf_len;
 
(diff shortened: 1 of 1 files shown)


jeongchanKim (commented):
Could you add Null check for `data` in `sl_test_malloc_buffer_priv()`?
All other functions check that, but only it doesn't.
It can be dereferencing.

sangam-swami:
> Could you add Null check for `data` in `sl_test_malloc_buffer_priv()`?
> All other functions check that, but only it doesn't.
> It can be dereferencing.

applied

apps/examples/security_test/seclink/sl_test_utils.c
```diff
@@ -64,9 +64,12 @@ int sl_test_malloc_buffer(hal_data *data, int buf_len)
 
 int sl_test_malloc_buffer_priv(hal_data *data, int buf_len)
 {
+	if (!data) {
+		return -1;
+	}
 	data->priv = (unsigned char *)zalloc(buf_len);
 	if (!data->priv) {
-		return -1;
+		return -2;
```
jeongchanKim, line 72: What is the meaning of `-1` and `-2`?
Just for checking where to fail?

jeongchanKim: Even if my assumption is correct, I don't know why we need this. Because we don't use/check the return value in anywhere.

sangam-swami: in sl_key_test they check only success return, so just for future checking I have added

jeongchanKim: Okay. Thanks.

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4775, opened 2021-02-04
os/mm: Add debug logic to find out heap corruption

kishore-sn:
Heap corruption can occur when data of one node overflows to the next node.
In order to detect heap corruption, we added some debug logic in mallinfo api.
We check the current node "preceding" size with previous node size. If it is
not a match, then previous node is corrupted. We print debug information about
the corrupted node and trigger an assert failure.

### DIFF
--- a/os/mm/mm_heap/mm_mallinfo.c
+++ b/os/mm/mm_heap/mm_mallinfo.c
@@ -86,7 +86,9 @@
 
 int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 {
-	struct mm_allocnode_s *node;
+	struct mm_freenode_s *node;
+	struct mm_freenode_s *prev;
+	struct mm_freenode_s *next;
 	size_t mxordblk = 0;
 	int    ordblks  = 0;		/* Number of non-inuse chunks */
 	size_t uordblks = 0;		/* Total allocated space */
@@ -111,8 +113,33 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 
 		mm_takesemaphore(heap);
 
-		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+		prev = NULL;
+		node = heap->mm_heapstart[region];
+		next = (struct mm_freenode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_freenode_s *)((char *)next + next->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
+
+			if ((prev && prev->size != node->preceding & ~MM_ALLOC_BIT) ||
+				(node->size != next->preceding & ~MM_ALLOC_BIT) ||
+				(!(node->preceding & MM_ALLOC_BIT) &&
+				((node->blink && node->blink->flink != node) ||
+				(node->flink && node->flink->blink != node)))
+			) {
+				mdbg("ERROR: Heap node corruption detected from one of following nodes\n");
+				if (prev) {
+					mdbg("Previous node addr = 0x%08x size = %u preceding size = %u\n", prev, prev->size, prev->preceding & ~MM_ALLOC_BIT);
+				}
+				mdbg("Current  node addr = 0x%08x size = %u preceding size = %u\n", node, node->size, node->preceding & ~MM_ALLOC_BIT);
+
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+				if (prev) {
+					mdbg("Previous node owner pid = %u, alloc_call_addr = 0x%08x\n", prev->pid, prev->alloc_call_addr);
+				}
+				mdbg("Current  node owner pid = %u, alloc_call_addr = 0x%08x\n", node->pid, node->alloc_call_addr);
+#endif
+				ASSERT(0);
+			}
 
 			/* Check if the node corresponds to an allocated memory chunk */
 
@@ -130,7 +157,6 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 		mm_givesemaphore(heap);
 
 		mvdbg("region=%d node=%p heapend=%p\n", region, node, heap->mm_heapend[region]);
-		DEBUGASSERT(node == heap->mm_heapend[region]);
 		uordblks += SIZEOF_MM_ALLOCNODE;	/* account for the tail node */
 	}
 #undef region


os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +112,19 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 
 		mm_takesemaphore(heap);
 
-		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+		prev = NULL;
```
sunghan-chang, line 115: Duplicated initialization

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +112,19 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; prev = node, node = (struct mm_allocnode_s *)((char *)node + node->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if (prev && (prev->size != (node->preceding & ~MM_ALLOC_BIT))) {
+				mdbg("ERROR: Heap node corruption detected\n");
+				mdbg("Corrupted node addr = 0x%08x size = %u\n", prev, prev->size);
```
sunghan-chang, line 121: Previous node is corrupting node, current node is corrupted node.
```
Corrupted node addr : node, size : node->size
Corrupting node addr : prev, size : prev->size
```
Am I right?

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +112,19 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+				mdbg("Corrupted node owner pid = %u, alloc_call_addr = 0x%08x\n", prev->pid, prev->alloc_call_addr);
+#endif
+				DEBUGASSERT(0);
```
sunghan-chang, line 125: Is debugassert enough? Because our mm is broken, we can run mm normally. So I think it should be assert, not debug assert.

sunghan-chang: In my opinion, we can use debug assert when it does not affect system operation even it's wrong. Then we can debug that with config_debug enabled.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +112,19 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 
 		mm_takesemaphore(heap);
 
-		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+		prev = NULL;
+		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; prev = node, node = (struct mm_allocnode_s *)((char *)node + node->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if (prev && (prev->size != (node->preceding & ~MM_ALLOC_BIT))) {
```
jeongchanKim, line 119: How about moving line 117 under this checking?
Because from line 119 to 126 means there is a corruption, then we don't need to show the current node's information.

os/mm/mm_heap/mm_mallinfo.c
jeongchanKim: I think corrupting is correct for this case.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: It needs to be initialized every time in the for loop for each memory region. (Line 106)

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Line 117 was present in the old code. I assumed it is for obtaining detailed logs of all nodes in the heap. If we move it under the if condition, then it will allow us to get the logs only for the corrupted node case. I think we can leave it at current position.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Node x got corrupted. This resulting in overflow into node x+1. 
So, the preceding value in node x+1 is wrong. Hence, the corruption is at node x.

So, I think @sunghan-chang is right. 
`Corrupting node addr : prev, size : prev->size`

But we are more concerned by who has caused the corruption during the debug stage. So, I think its better to print details of prev node only.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: True. I had a confusion about this. I have seen some places in code, where debugassert is used even in critical case. 
Let me change this to assert.

os/mm/mm_heap/mm_mallinfo.c
sunghan-chang:
```
Node x got corrupted. This resulting in overflow into node x+1.
So, the preceding value in node x+1 is wrong. Hence, the corruption is at node x.
```
Two cases can make above.
1. size of node x is corrupted.
2. preceding of node x + 1 is corrupted.
So, we should show both. (We don't know what is corrupted among them)

os/mm/mm_heap/mm_mallinfo.c
sunghan-chang: Then we can remove initialization in 90.
```
struct ... *prev;
```
is enough.

kishore-sn: ok

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: But, if node x size is corrupted, it is mostly because of overflow from node x-1.
So, it would be caught in the previous for loop iteration.

sunghan-chang: Right. I totally agree what you say. But it's one of the case.
When node x is ok but node (x+1) is corrupted for preceding value, then it will show the same assert.
What I mean is when we meet `size of node (x) != preceding of node (x+1)` assert
1. size of node (x) is broken, all of node (x+1) is ok. => node (x-1) overflow
2. all of node (x) is ok, preceding of node (x+1) is broken. => node (x) overflow
both cases show the same assert. But it is difficult to know what is broken and who overflows.

So, I think we should print 3 nodes information at assert case.

sunghan-chang: In my word, node (x+1) is next node of node (x). node (x-1) is previous node of node (x).

kishore-sn: Actually, we always try to find overflow by checking value of preceding variable. Not the size variable. 

1. preceding of node x is broken --> our logs show node x-1 information (address, owner ,etc)
2. preceding of node x+1 is broken --> our logs show node x information

In this way, our logs always show the information of the node which caused the overflow.
Also, pls check my changes in mm.h. I have made "preceding" as first member of the node struct. So, whenever there is a overflow from previous node, then preceding of current node will always get corrupted. So, we check 

`if (prev-> size != node->preceding)`
`      print(prev node details)
`

sunghan-chang: I think only `size` could be corrupted. Then, we can get the same assert by `if (prev-> size != node->preceding`.
But this case we need previous of prev.

sunghan-chang: If you think this is enough, then I will merge this immediately

jeongchanKim: I agreed with @sunghan-chang's opinion.
`size` can be corrupted even if preceding is okay.
So we'd better print/check the 3 nodes.

sunghan-chang: @kishore-sn So, could you let me know conclusion?

kishore-sn:
@sunghan-chang 
I have modified according to your suggestion to print all 3 nodes information when some issue is found in the heap. 
However, in this case, it is not possible to point out exactly which node was corrupted, and it is the developer responsibility to check the debug information and determine the corruption.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +113,31 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if (node->size != (next->preceding & ~MM_ALLOC_BIT)) {
+				mdbg("ERROR: Heap node corruption detected in one of following nodes\n");
+				if (prev) {
+					mdbg("Previous node addr = 0x%08x size = %u preceding size = %u\n", prev, prev->size, prev->preceding & ~MM_ALLOC_BIT);
+				}
+				mdbg("Current  node addr = 0x%08x size = %u preceding size = %u\n", node, node->size, node->preceding & ~MM_ALLOC_BIT);
```
Taejun-Kwon, line 128: mdbg display function name as prefix eh?
If I were you I would take below style.
```
#ifdef CONFIG_DEBUG_MM_ERROR
 dbg("Current  node addr = 0x%08x size = %u preceding size = %u\n", node, node->size, node->preceding & ~MM_ALLOC_BIT);
 dbg("Next     node addr = 0x%08x size = %u preceding size = %u\n", next, next->size, next->preceding & ~MM_ALLOC_BIT);
#endif
```

kishore-sn: But mdbg internally uses dbg.. and it prints the function name prefix.

sunghan-chang:
```c
#ifdef CONFIG_DEBUG_MM_ERROR
dbg(....);
#endif
```
and
```c
mdbg(....);
```
are exactly the same.

kishore-sn:
I have updated the commit to add some more checks for node corruption. 
However, there is no way to determine if current node is corrupted or next node is corrupted. 
So, we will print the debug info about prev and current nodes, one of which might have caused the corruption.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +113,33 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+		node = heap->mm_heapstart[region];
+		next = (struct mm_allocnode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if ((prev && prev->size != node->preceding & ~MM_ALLOC_BIT) ||
+				(node->size != next->preceding & ~MM_ALLOC_BIT) ||
+				(!(node->preceding & MM_ALLOC_BIT) &&
+				 (((struct mm_freenode_s*)node)->blink != prev || ((struct mm_freenode_s *)node)->flink != next))
```
sunghan-chang, line 126: This is wrong. The blink and flink is only for free node. If prev or next is allocated node, then blink or flink will not be them. This should be like this.
```
(((struct mm_freenode_s*)node)->blink->flink != node || ((struct mm_freenode_s *)node)->flink->blink != node))
```

kishore-sn: fixed.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +113,33 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+		prev = NULL;
+		node = heap->mm_heapstart[region];
+		next = (struct mm_allocnode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if ((prev && prev->size != node->preceding & ~MM_ALLOC_BIT) ||
+				(node->size != next->preceding & ~MM_ALLOC_BIT) ||
+				(!(node->preceding & MM_ALLOC_BIT) && !(prev->preceding & MM_ALLOC_BIT) && ((struct mm_freenode_s *)node)->blink != prev) ||
```
sunghan-chang, line 125: This is either wrong. Free node is sorted by size, it's not by location. So, even both are free node, node -> blink could not be prev.

sunghan-chang: And If both are free nodes, then they will be merged as one free node. This is checking for impossible case.

kishore-sn: Oh.. yes.. then we have to remove it..

sunghan-chang: (((struct mm_freenode_s*)node)->blink->flink != node || ((struct mm_freenode_s *)node)->flink->blink != node))

sunghan-chang: You can use this. This is what you want

kishore-sn: Done. Thanks.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +113,33 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+		prev = NULL;
+		node = heap->mm_heapstart[region];
+		next = (struct mm_allocnode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
 			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
+			if ((prev && prev->size != node->preceding & ~MM_ALLOC_BIT) ||
+				(node->size != next->preceding & ~MM_ALLOC_BIT) ||
+				(!(node->preceding & ~MM_ALLOC_BIT) &&
```
sunghan-chang, line 125: should be `MM_ALLOC_BIT`

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -111,9 +113,33 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
+				(((struct mm_freenode_s*)node)->blink->flink != node || ((struct mm_freenode_s *)node)->flink->blink != node))
```
sunghan-chang, line 126: Sorry, need to check null for `node->blink` and `node->flink` for start or end of free node llink

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #4776, opened 2021-02-04
apps/shell, telnetd: dynamically allocate shell command buffer

sunghan-chang:
Currently TASH uses global array variable to save the command.
It sets the size from the config to get the maximum number of command.
But it's not good due to belows.
1. User should consider the maximum number of shell commands.
  When user enables more application or utils, then user should check it.
  It's much uncomfortable.
2. When user does not use it fully, it wastes memory.
3. Usually global variables are located in internal RAM which has fastest performance.
  But shell command itself does not require performance.
  And this takes memory even TASH is exited.

This commit changes that TASH dynamically allocates memory for the commands instead of
using global array variable.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -18,13 +18,15 @@
 /// @file   tash_command.c
 /// @brief  functions to treate commands
 
-#include <pthread.h>
 #include <tinyara/config.h>
-#include <tinyara/ascii.h>
+
+#include <sys/types.h>
 #include <string.h>
 #include <stdio.h>
-#include <sys/types.h>
+#include <stdlib.h>
+#include <pthread.h>
 #include <sys/boardctl.h>
+#include <tinyara/ascii.h>
 #include <apps/shell/tash.h>
 #ifdef CONFIG_BUILTIN_APPS
 #include <apps/builtin.h>
@@ -37,11 +39,6 @@
 
 /* Following defines are fixed to avoid many configuration variables for TASH */
 #define TASH_CMD_MAXSTRLENGTH		(16) /* example: length of "ifconfig" command is 8 */
-#ifdef CONFIG_TASH_MAX_COMMANDS
-#define TASH_MAX_COMMANDS			CONFIG_TASH_MAX_COMMANDS
-#else
-#define TASH_MAX_COMMANDS			(32)
-#endif
 #ifdef CONFIG_TASH_CMDTASK_STACKSIZE
 #define TASH_CMDTASK_STACKSIZE		CONFIG_TASH_CMDTASK_STACKSIZE
 #else
@@ -85,7 +82,7 @@ struct tash_cmd_s {
 
 struct tash_cmd_info_s {
 	pthread_mutex_t tmutex;						/* Mutex for protection */
-	struct tash_cmd_s cmd[TASH_MAX_COMMANDS];
+	struct tash_cmd_s *cmd;
 	int count;									/* Number of TASH commands */
 };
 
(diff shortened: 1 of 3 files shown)


apps/shell/tash_command.c
```diff
@@ -638,18 +641,24 @@ bool tash_do_autocomplete(char *cmd, int *pos, bool double_tab)
...
+	if (tash_cmds_info.count == 0) {
+		new_cmd_buff = (struct tash_cmd_s *)malloc(sizeof(struct tash_cmd_s));
+	} else {
+		new_cmd_buff = (struct tash_cmd_s *)realloc(tash_cmds_info.cmd, sizeof(struct tash_cmd_s) * (tash_cmds_info.count + 1));
+	}
+
+	if (new_cmd_buff) {
+		tash_cmds_info.cmd = new_cmd_buff;
+	} else {
+		return -1; /* Memory Allocation Fail for new command */
```
jeongchanKim, line 659: How about printing the log we cannot install the cmd like before?

sunghan-chang: Done

apps/shell/tash_command.c
```diff
@@ -638,18 +641,25 @@ bool tash_do_autocomplete(char *cmd, int *pos, bool double_tab)
...
-		printf("Couldn't install cmd: (%s), Refer CONFIG_TASH_MAX_COMMANDS\n", str);
-		return -1;				/* MAX cmd count reached */
-	}
+	struct tash_cmd_s *new_cmd_buff;
 
 	/* Lock mutex */
 	pthread_mutex_lock(&tash_cmds_info.tmutex);
 
+	/* allocate the cmd buffer */
+	if (tash_cmds_info.count == 0) {
```
jeongchanKim, line 650: If command is already registered, we don't need to allocate the new.
So I think we'd better check the existence(line 664 to 669) before allocation or free the memory if already registered.

apps/shell/tash_command.c
```diff
@@ -638,18 +641,25 @@ bool tash_do_autocomplete(char *cmd, int *pos, bool double_tab)
...
+		new_cmd_buff = (struct tash_cmd_s *)malloc(sizeof(struct tash_cmd_s));
+	} else {
+		new_cmd_buff = (struct tash_cmd_s *)realloc(tash_cmds_info.cmd, sizeof(struct tash_cmd_s) * (tash_cmds_info.count + 1));
+	}
+
+	if (new_cmd_buff) {
+		tash_cmds_info.cmd = new_cmd_buff;
+	} else {
+		printf("TASH: memory allocation fail to register (%s) command\n", str);
+		return -1; /* Memory Allocation Fail for new command */
```
jeongchanKim, line 660: In this case,  we should unlock the mutex(`tash_cmds_info.tmutex`).

apps/shell/tash_command.c
```diff
@@ -638,24 +641,34 @@ bool tash_do_autocomplete(char *cmd, int *pos, bool double_tab)
...
 	pthread_mutex_lock(&tash_cmds_info.tmutex);
 
-	/* CHeck if cmd is already installed */
-	for (cmd_idx = 0; cmd_idx < tash_cmds_info.count; cmd_idx++) {
-		if (strncmp(str, tash_cmds_info.cmd[cmd_idx].str, TASH_CMD_MAXSTRLENGTH - 1) == 0) {
-			pthread_mutex_unlock(&tash_cmds_info.tmutex);
-			return -2;			/* CMD already installed */
+	/* allocate memory for new cmd */
+	if (tash_cmds_info.count == 0) {
+		new_cmd_buff = (struct tash_cmd_s *)malloc(sizeof(struct tash_cmd_s));
```
jeongchanKim, line 651: Please include `stdlib.h` for `malloc` and `realloc`.

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4782, opened 2021-02-05
Fix onoff driver

pillip8282:


os/fs/vfs/fs_select.c
```diff
@@ -265,6 +265,10 @@ int select(int nfds, FAR fd_set *readfds, FAR fd_set *writefds, FAR fd_set *exce
 	int ndx = _init_desc_list(nfds, readfds, writefds, exceptfds, pollset);
 
 	DEBUGASSERT(ndx == npfds);
+	if (ndx != npfds) {
+		set_errno(EINVAL);
```
Taejun-Kwon, line 269: we return EINVAL if any of parameter is invalid but both of ndx & npfds are not parameter of this function.
    This happened when fdset is invalid, then EBADF is correct value I think.

os/drivers/lwnl/lwnl.c
```diff
@@ -116,26 +115,19 @@ static const struct file_operations g_lwnl_fops = {
...
 static int lwnl_open(struct file *filep)
 {
 	LWNL_ENTER;
 	struct inode *inode = filep->f_inode;
 	struct lwnl_upperhalf_s *upper = inode->i_private;
-	int ret = -EMFILE;
 
+	LWNLDEV_LOCK(upper);
 	upper->crefs++;
+	LWNLDEV_UNLOCK(upper);
```
Taejun-Kwon, line 126: overflow can be happened here, crefs is just uint8_t..
So below is needed
```
int ref;
LWNLDEV_LOCK(upper);
ref = upper->crefs + 1;
if (ref == 0) {
    ret = -EMFILE; //please notice that crefs is not changed here.
    goto blahblah;
}
upper->crefs = refs; // now Update crefs here 

goto blahblah :
    LWNLDEV_UNLOCK(upper);
```

os/drivers/lwnl/lwnl.c
```diff
@@ -149,14 +141,14 @@ static int lwnl_close(struct file *filep)
 	int waiter_idx;
 #endif
 
+	LWNLDEV_LOCK(upper);
```
Taejun-Kwon, line 144: Please check definition of -ENOSYS
It can't be used here. Also please see below.
```
if (upper->crefs > 1) {
     upper->crefs--;
} else {
     /* Last fd will be closed here, so we can free all resources, here and lower */
}
```

os/drivers/lwnl/lwnl.c
```diff
@@ -169,69 +161,55 @@ static int lwnl_close(struct file *filep)
...
 
-
 static int lwnl_ioctl(struct file *filep, int cmd, unsigned long arg)
 {
 	LWNL_ENTER;
-
-	return 0;
+	int res = lwnl_add_listener(filep);
+	if (res < 0) {
+		res = -ENOSYS;
```
Taejun-Kwon, line 207: same as above. if there's no implemented code in ioctl , then it should return -ENOSYS
if lwnl_add_listener has no implemented code, it should return -ENOSYS
so res will be smaller than 0, it doesn't affect here.

pillip8282:
@Taejun-Kwon 
I resolved your comments in PR4842. please check it

PR #4784, opened 2021-02-05
tools/mkbinheader: Support kernel binary header

jeongarmy:
Support kernel binary header for binary update.
 - CRC32 : uint32_t, 4bytes
 - Header size : uint32_t, 4bytes
 - Binary size : uint32_t, 4bytes
 - Secure header size : uint32_t, 4bytes
 - Secure header : #bytes. It is board-specific values


os/Kconfig
```diff
@@ -268,7 +268,15 @@ config PASS1_OBJECT
 	default ""
 
 endif # Build 2-pass
-endmenu # Build Setup
+
+config VERSION_BUILD_DATE
+	int "Binary Build Date"
```
jeongchanKim, line 273: I think we'd better show the format in the prompt.
How about `Binary Build Date(YYMMDD)`?

os/tools/mkbinheader.py
```diff
@@ -102,10 +108,65 @@ def get_static_ram_size(bin_type):
...
     return static_ram_size
 
+############################################################################
+#
+# Kernel binary header information :
+#
+# total header size is (16 + #)bytes.
+# +-----------------------------------------------------------------------------------
+# | Header size | Binary Version |  Binary Size | Secure Header Size | Secure Header |
+# |   (4bytes)  |    (4byte)     |    (4byte)   |       (4byte)      |    (#bytes)   |
```
jeongchanKim, line 118: In user binary header, header size is 2bytes. But for kernel, it is 4bytes.
I think 2bytes is also enough to kernel. What do you think about?

build/configs/rtl8721csm/rtl8721csm_make_bin.sh
```diff
@@ -130,11 +132,15 @@ function concatenate_binary()
 		cp $BOOT_PATH/km4_boot_all.bin $BINDIR/km4_boot_all.bin
 		cat $BINDIR/xip_image2_prepend.bin $BINDIR/ram_2_prepend.bin $BINDIR/psram_2_prepend.bin > $BINDIR/km4_image2_all.bin
 		cat $GNUUTL/km0_image2_all.bin $BINDIR/km4_image2_all.bin > $BINDIR/km0_km4_image2.bin
+		${TOP_PATH}/os/tools/mkbinheader.py $BINDIR/km0_km4_image2.bin kernel $SECURE_HEADER_SIZE
```
sunghan-chang, line 135: Samsung header is not board specific. How about calling it from Makefile.unix after MAKE_BOARD_SPECIFIC_BIN?

os/tools/mkbinheader.py
```diff
@@ -102,10 +108,65 @@ def get_static_ram_size(bin_type):
     os.remove(STATIC_RAM_ESTIMATION)
     return static_ram_size
 
+############################################################################
+#
+# Kernel binary header information :
+#
+# total header size is (16 + #)bytes.
+# +-----------------------------------------------------------------------------------
```
sunghan-chang, line 116: where is crc?

os/Kconfig
```diff
@@ -268,7 +268,15 @@ config PASS1_OBJECT
 	default ""
 
 endif # Build 2-pass
-endmenu # Build Setup
+
+config VERSION_BUILD_DATE
```
sunghan-chang, line 272: This looks much similar with CONFIG_VERSION_BUILD_TIME.
This is for version but its format is date. So, let's just say BINARY_VERSION.
We can't guarantee the version will be the same as build date.

os/tools/mkbinheader.py
```diff
@@ -124,198 +185,212 @@ def get_static_ram_size(bin_type):
...
+    if priority > 0 :
+        BM_PRIORITY_MIN = priority
+
+    with open(file_path, 'rb') as fp:
+        # binary data copy to 'data'
+        data = fp.read()
+        file_size = fp.tell()
+        fp.close()
+
+        if binary_format == 'bin' or binary_format == 'BIN' :
```
jeongchanKim, line 276: If we separate the kernel case, is there `bin` type for user binary?
If not, we can remove the code which is related to bin type.

build/configs/rtl8721csm/rtl8721csm_make_bin.sh
jeongarmy: Yes, but each binary names can be different.

os/tools/mkbinheader.py
jeongarmy: CRC is generated by mkchecksum.py for whole binary including binary header after this script is executed.

os/tools/mkbinheader.py
```diff
@@ -116,206 +177,197 @@ def get_static_ram_size(bin_type):
...
+    binary_ver = sys.argv[6]
+    dynamic_ram_size = sys.argv[7]
+    main_stack_size = sys.argv[8]
+    main_priority = sys.argv[9]
+    comp_enabled = sys.argv[10]
+    comp_blk_size = sys.argv[11]
+    loading_priority = sys.argv[12]
+
+    # This path is only for dbuild.
+    elf_path_for_bin_type = 'root/tizenrt/build/output/bin/tinyara'
```
jeongchanKim, line 215: This is not needed anymore. Please remove it.

os/tools/mkbinheader.py
```diff
@@ -116,206 +177,197 @@ def get_static_ram_size(bin_type):
...
-        bin_type = BIN
-    elif binary_type == 'elf' or binary_type == 'ELF' :
-        bin_type = ELF
-    else : # Not supported.
-        bin_type = 0
-        print("Error : Not supported Binary Type")
-        sys.exit(1)
+    # In size command on linux, 4th value is the summation of text, data and bss.
+    # We will use this value for elf.
+    SIZE_CMD_SUMMATION_INDEX = 3
```
jeongchanKim, line 255: There is no usage of `SIZE_CMD_SUMMATION_INDEX `. Please remove it.

os/tools/mkbinheader.py
```diff
@@ -116,206 +177,197 @@ def get_static_ram_size(bin_type):
...
-    fp.write(data)
+############################################################################
+#
+# Generate headers for binary types
+#
+############################################################################
+file_path  = sys.argv[1]
+binary_type = sys.argv[2]
+if binary_type == 'kernel' :
+    make_kernel_binary()
```
jeongchanKim, line 367: This is not making binary.
How about changing `make_kernel_bin_header()` and `make_user_bin_header()`?
This script name is also mkbinheader.

build/configs/rtl8721csm/hello/Make.defs
```diff
@@ -210,6 +210,13 @@ define MAKE_BOARD_SPECIFIC_BIN
   $(TOPDIR)/../build/configs/rtl8721csm/rtl8721csm_make_bin.sh
 endef
 
+SECURE_HEADER_SIZE=20
+
+define MAKE_SAMSUNG_HEADER
+  $(TOPDIR)/../os/tools/mkbinheader.py $(TOPDIR)/../build/output/bin/km0_km4_image2.bin kernel ${SECURE_HEADER_SIZE}
```
sunghan-chang, line 216: I suggest two things.
1. Let's make new script file to call these two lines. 
   And please write entire structure of samsung header there. There is no total structure of samsung header anywhere.
2. Let's make these two lines can be used in other board. It means you should use some macro for binary like `SECURE_HEADER_SIZE`.
  It's easier to know what we should modify for new board.

os/tools/mkbinheader.py
```diff
@@ -116,206 +170,189 @@ def get_static_ram_size(bin_type):
...
+        static_ram_size = get_static_ram_size(bin_type)
+        cfg_path = CONFIG_DIR + '/.config'
+        if check_optimize_config(cfg_path) == True:
+            binary_ram_size = int(dynamic_ram_size)
+        else:
+            binary_ram_size = int(static_ram_size) + int(dynamic_ram_size)
+            binary_ram_size = roundup_power_two(binary_ram_size)
+
+        if float(kernel_ver) <= 0.0 :
+            kernel_ver = 3.0
```
jeongchanKim, line 302: In Kconfig and `make_kernel_binary_header()`, default kernel version is int type and its value is based on `CONFIG_BINARY_VERSION`. But you set the default version as 3.0 in here, even float type.
I think we should unify the version.

os/tools/mkbinheader.py
```diff
@@ -116,206 +170,189 @@ def get_static_ram_size(bin_type):
...
+        if binary_format == 'elf' or binary_format == 'ELF' :
+            bin_type = ELF
+        else : # Not supported.
+            bin_type = 0
+            print("Error : Not supported Binary Type")
+            sys.exit(1)
+
+        # Calculate RAM size
+        # Dynamic RAM size : user input of argv[6]
+        # Static RAM size : Extract from size command in linux(ONLY for elf)
```
jeongchanKim, line 270: There is no usage for using size command in linux anymore. Please remove this explanation.

os/tools/mkbinheader.py
```diff
@@ -116,206 +170,189 @@ def get_static_ram_size(bin_type):
...
+        if bin_type == ELF :
```
jeongchanKim, line 271: I think we can merge from line 271 to line 275 with from line 261 to line 266.
We can reduce the code then.

sunghan-chang:
@jeongarmy 
1. Could you check comments?
2. Is TizenRT booting well with this PR? Could you let me know whether we need to modify ld script or not? If we need, please make a separated commit for that to get review from Realtek.

os/tools/mkbinheader.py
```diff
@@ -94,20 +96,74 @@ def get_static_ram_size(bin_type):
...
-    os.remove(STATIC_RAM_ESTIMATION)
-    return static_ram_size
+############################################################################
+#
+# Kernel binary header information :
+#
+# total header size is (12 + #)bytes.
+# +-----------------------------------------------------------------------------------
+# | Header size | Binary Version |  Binary Size | Secure Header Size | Secure Header |
+# |   (2bytes)  |    (4byte)     |    (4byte)   |       (2byte)      |    (#bytes)   |
```
jeongchanKim, line 113: Kernel version is YYMMDD format as you said in Kconfig.
Then do we need 4bytes?

os/include/tinyara/binary_manager.h
```diff
@@ -142,17 +142,15 @@ struct binary_header_s {
 	uint32_t bin_ver;
 	uint32_t bin_ramsize;
 	uint32_t bin_stacksize;
-	float kernel_ver;
-	uint32_t jump_addr;
+	uint16_t kernel_ver;
```
jeongchanKim, line 145: If we rearrange the member variable, we can save the memory, Because of padding.
Please check it.

os/Kconfig
```diff
@@ -268,7 +268,14 @@ config PASS1_OBJECT
 	default ""
 
 endif # Build 2-pass
-endmenu # Build Setup
+
+config BINARY_VERSION
+	int "Binary Version(YYMMDD)"
```
Taejun-Kwon, line 273: Actually not all culture uses YYMMDD, for example england culture uses DDMMYY.
(so in this case binary manager occure error??)
So you should change description about this.
For example, Binary version(default value follows YYMMDD type, please make sure that high number of version is latest version) blah blah blah blah
As per my experience, this is very serious agenda in product level.

os/tools/mksamsungheader.py
```diff
@@ -0,0 +1,73 @@
...
+# total header size is (14 + #)bytes.
+# +----------------------------------------------------------------------------------------------
+# |  CRC32   | Header size | Binary Version |  Binary Size | Secure Header Size | Secure Header |
+# | (4bytes) |   (2bytes)  |    (2bytes)    |   (4bytes)   |      (2bytes)      |    (#bytes)   |
+# +----------------------------------------------------------------------------------------------
+#  * A CRC32 is checksum value for the entire binary from Header size to the end of binary.
+#  * A size of Secure Header depends on board.
+#
+#
+# 2. User binary header information on APP_BINARY_SEPARTION
```
sunghan-chang, line 38: This file does not make the header of user binary. No need to show this here. Let's remove this block.

os/tools/mksamsungheader.py
```diff
@@ -0,0 +1,73 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+############################################################################
+import os
+import sys
+
+TOOL_DIR = '/root/tizenrt/os/tools/'
```
sunghan-chang, line 22: `/root/tizenrt/` is not right for all the cases. That's why we use `test -d ${0%/*} && cd ${0%/*}; pwd\` in shell script to get current path.
Please do like shell script.

sunghan-chang (approved):

PR #4789, opened 2021-02-08
docs/APIListToBePorted.md : Add up_lowputc and reboot reason APIs

jeongchanKim:


docs/APIListToBePorted.md
```diff
@@ -21,6 +21,11 @@
 14. up_i2cinitialize()
 15. up_spiinitialize()
 16. up_rtc_initialize()
+17. up_lowputc()
```
sunghan-chang, line 24: How about locating this after up_putc? Because it's almost the same, user could be better to check.

sunghan-chang (approved):

PR #4790, opened 2021-02-08
configs/rtl8721csm: Enable debug configs for rtk

kishore-sn:
Enable CONFIG_DEBUG and CONFIG_DEBUG_ERROR for rtk defconfigs.
If this is not enabled, all the faults such as memfault, busfault,
etc are treated as unexpected irq and we dont get debug information
for them. These configs were previously disabled for wifi perf
issue, and since we found that wifi perf is not related to these
configs, we are re-enabling them.

### DIFF
--- a/build/configs/rtl8721csm/hello/defconfig
+++ b/build/configs/rtl8721csm/hello/defconfig
@@ -947,7 +947,50 @@ CONFIG_SCHED_LPWORKSTACKSIZE=2048
 #
 # Debug Options
 #
-# CONFIG_DEBUG is not set
+CONFIG_DEBUG=y
+CONFIG_DEBUG_ERROR=y
+# CONFIG_DEBUG_WARN is not set
+# CONFIG_DEBUG_VERBOSE is not set
+
+#
+# Subsystem Debug Options
+#
+# CONFIG_DEBUG_BINFMT is not set
+# CONFIG_DEBUG_BINMGR is not set
+# CONFIG_DEBUG_FS is not set
+# CONFIG_DEBUG_LIB is not set
+# CONFIG_DEBUG_MM is not set
+# CONFIG_DEBUG_NET is not set
+# CONFIG_DEBUG_PM is not set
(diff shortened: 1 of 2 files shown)


jeongchanKim (approved):
LGTM

sunghan-chang:
@kishore-sn Do we need to enable CONFIG_DEBUG or to remove CONFIG_DEBUG conditional from amebad_irq.c?

jeongarmy (approved):
LGTM

kishore-sn:
> @kishore-sn Do we need to enable CONFIG_DEBUG or to remove CONFIG_DEBUG conditional from amebad_irq.c?

I am really not sure. In every chipset irq init, the fault irqs are added under CONFIG_DEBUG.

sunghan-chang:
> I am really not sure. In every chipset irq init, the fault irqs are added under CONFIG_DEBUG.

@kishore-sn Yeah, it does. It looks Nuttx style. But we should think what is correct for us. If we don't want to debug, then we could not need the logs. That's why it does. But currently we should save the reboot reason and if we need to know exception information, then we should change the logic.
How about adding one common API to cover all of exception without config_debug? It shows just exception number and saves reboot reason. But if you think we need all of assert logs, then we should remove the conditional from xxx_irq.c

kishore-sn:
I think its better to get complete assert logs in case of memfault, usagefault and busfault. 
When any of these faults occur, we are entering the panic handler and it is end of execution. So, i think it must be ok to print complete debug information in these cases. So, I will move these 3 out of the config_debug. 

Would you like me to make this change in all chipsets?
Also, even if this change is done, I think it is a good idea to enable CONFIG_DEBUG. 
If you think this is ok, then please merge this PR. I will raise new PR for the above changes in xxx_irq.c

sunghan-chang (approved):

sunghan-chang:
> Would you like me to make this change in all chipsets?

@kishore-sn Yes, please.

PR #4793, opened 2021-02-09
os/kernel/binary_manager: Add NULL check in binary_manager_add_binlist

kishore-sn:
This commit is for crash issue in network_tc pthread. It is observed that
there is a busfault while running net_sendto_main(). This API creates some
pthreads. If we add some print statements in the pthread, then crash is not
reproduced. It is found from analysis that the crash is happening due to
NULL pointer access in binary_manager_add_binlist. This is because, the pthread
seems to have exited before the binary_manager_add_binlist() is called, and
hence the tcb->group member is NULL.

In order to fix this issue, we perform NULL check in binary_manager_add_binlist
and if group member in tcb is null, then we do not need to add the tcb to the
app's thread list.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_data.c
+++ b/os/kernel/binary_manager/binary_manager_data.c
@@ -191,6 +191,21 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
 {
 	int bin_idx;
 
+	if (!tcb) {
+		bmdbg("ERROR: tcb parameter is NULL\n");
+		return;
+	}
+
+	if (!tcb->group) {
+		/* It is possible that the task / pthread has already exited at the time of calling this function.
+		 * In this case, the group member will be NULL and will cause an exception if we try to dereference
+		 * the pointer. So, perform a null check here. If group is null, it is ok to not add the tcb to our
+		 * list because, task has already exited.
+		 */
+		bmdbg("Failed to add pid %d to binlist. This task has already exited and group is NULL\n", tcb->pid);
+		return;
+	}
+
 	bin_idx = tcb->group->tg_binidx;
 
 	/* A binary index, bin_idx is greater than 0 only if tcb is a thread of user binary.


os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -191,6 +191,13 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
 {
 	int bin_idx;
 
+	DEBUGASSERT(tcb);
+
+	if (!tcb->group) {
```
sijupunnnoose, line 196: Good to add a one line comment, why this check is important here.

deepaksrma (commented):
please change in commit msg, like below:
........if we add some print statements in the pthread_exit(), then crash ......

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -191,6 +191,13 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
 {
 	int bin_idx;
 
+	DEBUGASSERT(tcb);
```
jeongchanKim, line 194: For line 196, we'd better check the tcb not only when debug.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -191,6 +191,13 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
 {
```
jeongarmy, line 191: NULL checking is right but a scheduling is locked by sched_lock before calling task_activate in pthread_create. Is it possible that pthread is exited before calling it?

kishore-sn: Yes. But as per my understanding this is what happens:
Both parent and child tasks are created with default priority of 100.
- Parent task is running
- pthread_create
- activate task
- pthread_sem_take(&pjoin->data_sem, false); is called ... So parent gets blocked
- pthread_start for child is called
- Here, pthread_sem_give(&pjoin->data_sem) is called. This is supposed to unblock the parent. But since child is the current running task and both have same priority, the parent task gets added to the ready to run list just after the child task.
- So, the child task will continue to run, before the pthread_create of the parent is completed.

This is why, i think, its possible that the child pthread gets exited before the pthread_create is complete.

os/kernel/binary_manager/binary_manager_data.c
kishore-sn: Fixed.

os/kernel/binary_manager/binary_manager_data.c
kishore-sn: Done

kishore-sn:
> please change in commit msg, like below:
> ........if we add some print statements in the pthread_exit(), then crash ......

Done.

os/kernel/binary_manager/binary_manager_data.c
jeongarmy: Thank you so much. I understood

os/kernel/binary_manager/binary_manager_data.c
jeongarmy: I think it should not go panic because it is not fault case. A tcb can be NULL in normal scenario as you said.

kishore-sn: True. I removed the assert now.

sijupunnnoose (approved):
LGTM

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -191,6 +191,21 @@ void binary_manager_add_binlist(FAR struct tcb_s *tcb)
...
 	int bin_idx;
 
+	if (!tcb) {
+		bmdbg("ERROR: tcb parameter is NULL\n");
+		return;
+	}
+
+	if (!tcb->group) {
+		/* It is possible that the task / pthread has already exited at the time of calling this function.
+		 * In this case, the group member will be NULL and will cause an exception if we try to derefence
```
sangam-swami, line 201: dereference

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4797, opened 2021-02-10
build/configs/rtl8721csm: Update USERFS DEVNAME

AdamZhang0124:
USERFS Dev Name should be matched with index of flash partition list. UserFS is located at index 6 so that DEV name should be "smart0p6", not "smart0p5".


sangam-swami (approved):
Looks fine

sunghan-chang (approved):

PR #4791, opened 2021-02-09
mm/mm_heap : Fix build error and warnings

jeongchanKim:

PR #4799, opened 2021-02-12
New method to debug heap corruption

kishore-sn:
Heap corruption can happen when a node overflows into next node
in the heap list. However, even when heap is corrupted, the system
can run properly for some time. So, we provide heap corruption check
in following ways:

- Add heap corruption check api as part of heapinfo
- If heapinfo is not enabled, then we dont have any information about
the owner process of the heap node which caused the corruption.
- When user calls heapinfo for any app or kernel, the corresponding
heap is checked for corruption. All details about corruption are printed.
- When mallinfo api is called, heap corruption is checked and printed.
- In both cases, we will not ASSERT because, even with heap corruption,
the system can continue to operate unless the corrupted node is accessed.
- However, if the system gets ASSERT due to any other reason, we check
and print the heap corruption details for the corresponding heap.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -91,6 +91,8 @@
 #include <sys/boardctl.h>
 #endif
 
+#include <tinyara/mm/mm.h>
+
 #ifdef CONFIG_BINMGR_RECOVERY
 #include <stdbool.h>
 #include "binary_manager/binary_manager.h"
@@ -477,6 +479,19 @@ void up_assert(const uint8_t *filename, int lineno)
 #endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
+
+	lldbg("Checking kernel heap for corruption...\n");
+	if (mm_check_heap_corruption(g_kmmheap) == OK) {
+		lldbg("No heap corruption detected\n");
+	}
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	if (!is_kernel_fault) {
+		lldbg("Checking current app heap for corruption...\n");
+		if (mm_check_heap_corruption((struct mm_heap_s *)(this_task()->uheap)) == OK) {
+			lldbg("No heap corruption detected\n");
+		}
+	}
+#endif
 
 #if defined(CONFIG_BOARD_CRASHDUMP)
 	board_crashdump(up_getsp(), this_task(), (uint8_t *)filename, lineno);
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -104,6 +104,8 @@
 
 #ifdef CONFIG_BINMGR_RECOVERY
 bool abort_mode = false;
+#endif
+#ifdef CONFIG_APP_BINARY_SEPARATION
 extern uint32_t g_assertpc;
 #endif
 
@@ -456,7 +458,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 
-#ifdef CONFIG_BINMGR_RECOVERY
+#ifdef CONFIG_APP_BINARY_SEPARATION
 	uint32_t assert_pc;
 	bool is_kernel_fault;
 
@@ -472,7 +474,7 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	is_kernel_fault = is_kernel_space((void *)assert_pc);
 
-#endif  /* CONFIG_BINMGR_RECOVERY */
+#endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
 
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -106,12 +106,14 @@
 
 #ifdef CONFIG_BINMGR_RECOVERY
 bool abort_mode = false;
-extern uint32_t g_assertpc;
 extern struct tcb_s *g_faultmsg_sender;
 extern sq_queue_t g_faultmsg_list;
 extern sq_queue_t g_freemsg_list;
 #endif
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+extern uint32_t g_assertpc;
+#endif
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -386,6 +388,7 @@ static void up_dumpstate(void)
 
 	(void)usbtrace_enumerate(assert_tracecallback, NULL);
 #endif
+
 }
 #else
 #define up_dumpstate()
@@ -462,7 +465,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 
-#ifdef CONFIG_BINMGR_RECOVERY
+#ifdef CONFIG_APP_BINARY_SEPARATION
 	uint32_t assert_pc;
 	bool is_kernel_fault;
 
@@ -478,7 +481,7 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	is_kernel_fault = is_kernel_space((void *)assert_pc);
 
-#endif  /* CONFIG_BINMGR_RECOVERY */
+#endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
 
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -86,6 +86,8 @@
 #include <arch/board/board.h>
 #include <tinyara/sched.h>
 
+#include <tinyara/mm/mm.h>
+
 #include "sched/sched.h"
 #ifdef CONFIG_BOARD_ASSERT_AUTORESET
 #include <sys/boardctl.h>
@@ -484,6 +486,19 @@ void up_assert(const uint8_t *filename, int lineno)
 #endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
+
+	lldbg("Checking kernel heap for corruption...\n");
+	if (mm_check_heap_corruption(g_kmmheap) == OK) {
+		lldbg("No heap corruption detected\n");
+	}
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	if (!is_kernel_fault) {
+		lldbg("Checking current app heap for corruption...\n");
+		if (mm_check_heap_corruption((struct mm_heap_s *)(this_task()->uheap)) == OK) {
+			lldbg("No heap corruption detected\n");
+		}
+	}
+#endif
 
 #if defined(CONFIG_BOARD_CRASHDUMP)
 	board_crashdump(up_getsp(), this_task(), (uint8_t *)filename, lineno);
--- a/os/mm/mm_heap/mm_heapinfo_parse_heap.c
+++ b/os/mm/mm_heap/mm_heapinfo_parse_heap.c
@@ -69,6 +69,7 @@
  ****************************************************************************/
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
+
 /****************************************************************************
  * Name: heapinfo_parse
  *
@@ -104,6 +105,8 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 	size_t nodelist_size[MM_NNODES] = {0, };
 	FAR struct mm_freenode_s *fnode;
 #endif
+
+	ASSERT(mm_check_heap_corruption(heap) == OK);
 
 	/* initialize the heap, stack and nonsched resource */
 	nonsched_resource = 0;
--- a/os/mm/mm_heap/mm_mallinfo.c
+++ b/os/mm/mm_heap/mm_mallinfo.c
@@ -61,6 +61,7 @@
 #include <unistd.h>
 #include <tinyara/mm/mm.h>
 #include <tinyara/sched.h>
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -86,9 +87,7 @@
 
 int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 {
-	struct mm_freenode_s *node;
-	struct mm_freenode_s *prev;
-	struct mm_freenode_s *next;
+	struct mm_allocnode_s *node;
 	size_t mxordblk = 0;
 	int    ordblks  = 0;		/* Number of non-inuse chunks */
 	size_t uordblks = 0;		/* Total allocated space */
@@ -113,36 +112,11 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 
 		mm_takesemaphore(heap);
 
-		prev = NULL;
-		node = (struct mm_freenode_s *)heap->mm_heapstart[region];
-		next = (struct mm_freenode_s *)((char *)node + node->size);
-
-		for (; node < (struct mm_freenode_s *)heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_freenode_s *)((char *)next + next->size)) {
-			mvdbg("region=%d node=%p size=%u preceding=%u (%c)\n", region, node, node->size, (node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
-
-			if ((prev && prev->size != (node->preceding & ~MM_ALLOC_BIT)) ||
-				(node->size != (next->preceding & ~MM_ALLOC_BIT)) ||
-				(!(node->preceding & MM_ALLOC_BIT) &&
-				((node->blink && node->blink->flink != node) ||
-				(node->flink && node->flink->blink != node)))
-			) {
-				mdbg("ERROR: Heap node corruption detected from one of following nodes\n");
-				if (prev) {
-					mdbg("Previous node addr = 0x%08x size = %u preceding size = %u\n", prev, prev->size, prev->preceding & ~MM_ALLOC_BIT);
-				}
-				mdbg("Current  node addr = 0x%08x size = %u preceding size = %u\n", node, node->size, node->preceding & ~MM_ALLOC_BIT);
-
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-				if (prev) {
-					mdbg("Previous node owner pid = %u, alloc_call_addr = 0x%08x\n", prev->pid, prev->alloc_call_addr);
-				}
-				mdbg("Current  node owner pid = %u, alloc_call_addr = 0x%08x\n", node->pid, node->alloc_call_addr);
-#endif
-				ASSERT(0);
-			}
+		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+			mvdbg("region=%d node=%p size=%u preceding=%u (%c)\n", region, node, node->size,
+				(node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 
 			/* Check if the node corresponds to an allocated memory chunk */
-
 			if ((node->preceding & MM_ALLOC_BIT) != 0) {
 				uordblks += node->size;
 			} else {
@@ -157,6 +131,7 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 		mm_givesemaphore(heap);
 
 		mvdbg("region=%d node=%p heapend=%p\n", region, node, heap->mm_heapend[region]);
+		ASSERT(node == heap->mm_heapend[region]);
 		uordblks += SIZEOF_MM_ALLOCNODE;	/* account for the tail node */
 	}
 #undef region
--- a/os/arch/arm/src/armv7-m/up_hardfault.c
+++ b/os/arch/arm/src/armv7-m/up_hardfault.c
@@ -84,7 +84,7 @@
 
 #define INSN_SVC0        0xdf00	/* insn: svc 0 */
 
-#ifdef CONFIG_BINMGR_RECOVERY
+#ifdef CONFIG_APP_BINARY_SEPARATION
 uint32_t g_assertpc;
 #endif
 /****************************************************************************
@@ -118,7 +118,7 @@ int up_hardfault(int irq, FAR void *context, FAR void *arg)
 	uint32_t *regs = (uint32_t *)context;
 #endif
 
-#ifdef CONFIG_BINMGR_RECOVERY
+#ifdef CONFIG_APP_BINARY_SEPARATION
 	g_assertpc = regs[REG_R15];
 #endif
 	/* Get the value of the program counter where the fault occurred */
--- a/os/arch/arm/src/armv7-m/up_svcall.c
+++ b/os/arch/arm/src/armv7-m/up_svcall.c
@@ -97,7 +97,7 @@
 #define svcdbg(...)
 #endif
 
-#ifdef CONFIG_BINMGR_RECOVERY
+#ifdef CONFIG_APP_BINARY_SEPARATION
 extern uint32_t g_assertpc;
 #endif
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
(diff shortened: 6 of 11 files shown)


os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -105,6 +203,8 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 	FAR struct mm_freenode_s *fnode;
 #endif
 
+	heapinfo_check_corruption(heap);
```
jeongchanKim, line 206: In this case, we walk the whole heap twice. How about merging this with the line from 219 to 286?
Or how about exiting the parse heap when there is a corruption? Because if there is corrupted node, it is not useful to see the whole heap.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -113,36 +116,11 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-				if (prev) {
-					mdbg("Previous node owner pid = %u, alloc_call_addr = 0x%08x\n", prev->pid, prev->alloc_call_addr);
-				}
-				mdbg("Current  node owner pid = %u, alloc_call_addr = 0x%08x\n", node->pid, node->alloc_call_addr);
-#endif
-				ASSERT(0);
-			}
+		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+			mvdbg("region=%d node=%p size=%p preceding=%p (%c)\n", region, node, node->size,
```
jeongchanKim, line 120: Please print the `node->size` with `%u` instead of `%p`?

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,11 +64,109 @@
...
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
+
+/****************************************************************************
+ * Name: heapinfo_check_corruption
+ *
+ * Description:
+ *   This function walk through heap and prints information about corrupt node.
+ ****************************************************************************/
+int heapinfo_check_corruption(struct mm_heap_s *heap)
```
jeongchanKim, line 105: Please add `mm_givesemaphore` for success and fail cases.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,11 +64,109 @@
...
  * Public Functions
  ****************************************************************************/
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
+
+/****************************************************************************
+ * Name: heapinfo_check_corruption
+ *
+ * Description:
+ *   This function walk through heap and prints information about corrupt node.
```
jeongchanKim, line 103: walk -> walks

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,11 +64,109 @@
...
 
+/****************************************************************************
+ * Public data
+ ****************************************************************************/
+
+typedef enum
+{
+	TYPE_CORRUPTED,
+	TYPE_OVERFLOWED,
+} node_type_t;
```
jeongchanKim, line 75: In TizenRT, we declare the enum type with `xxx_e` at the end.
So, please change the type name to `node_type_e` instead of `node_type_t`.

sunghan-chang: _t is correct for typedef. But our coding rule recommends splitting the enum definition and type definition.

apps/system/utils/utils_heapinfo.c
```diff
@@ -291,6 +284,13 @@ int utils_heapinfo(int argc, char **args)
 	}
 	close(heapinfo_fd);
 
+#ifdef CONFIG_BUILD_PROTECTED
```
jeongchanKim, line 287: Why did you move this? For clarifying which heap to see, it is better to print before calling parse heap(`ioctl(heapinfo_fd, MMINFOIOC_PARSE, (int)&options)`).

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -113,36 +116,11 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
...
-				}
-				mdbg("Current  node addr = 0x%08x size = %u preceding size = %u\n", node, node->size, node->preceding & ~MM_ALLOC_BIT);
-
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-				if (prev) {
-					mdbg("Previous node owner pid = %u, alloc_call_addr = 0x%08x\n", prev->pid, prev->alloc_call_addr);
-				}
-				mdbg("Current  node owner pid = %u, alloc_call_addr = 0x%08x\n", node->pid, node->alloc_call_addr);
-#endif
-				ASSERT(0);
```
sunghan-chang, line 141: Could you let me know why you remove this from mm_mallinfo?
If because this is duplicated, then we should revert removing previous assert at line 138.
```c
mvdbg("region=%d node=%p heapend=%p\n", region, node, heap->mm_heapend[region]);
DEBUGASSERT(node == heap->mm_heapend[region]);
uordblks += SIZEOF_MM_ALLOCNODE;	/* account for the tail node */
```

When you add debugassert, I want to know your opinion for debugassert and assert for that.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,11 +64,109 @@
...
+			if (prev && prev->size != (node->preceding & ~MM_ALLOC_BIT)) {
+				mdbg("ERROR: Heap node corruption detected\n");
+				dump_node(prev, TYPE_OVERFLOWED);
+				dump_node(node, TYPE_CORRUPTED);
+				return -1;
+			} else if (node->size != (next->preceding & ~MM_ALLOC_BIT)) {
+				mdbg("ERROR: Heap node corruption detected. Possible corruption at following 2 nodes:\n");
+				dump_node(prev, TYPE_OVERFLOWED);
+				dump_node(node, TYPE_CORRUPTED);
+				dump_node(next, TYPE_CORRUPTED);
```
sunghan-chang, line 144: We can meet this statement with two cases.
1. prev overflowed, node corrupted
2. node overflowed, next corrupted (prev is ok)

But this statement shows that prev is overflowed. It could cause misunderstanding.

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -478,10 +485,19 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	is_kernel_fault = is_kernel_space((void *)assert_pc);
 
-#endif  /* CONFIG_BINMGR_RECOVERY */
+#endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
```
sunghan-chang, line 492: I don't think CONFIG_DEBUG_MM_HEAPINFO is mandatory for this.
We can't know the owner information which is almost mandatory for debug.
But without that config, it's helpful to know whether heap is corrupted or not.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,11 +64,109 @@
...
+		 */
+
+		mm_takesemaphore(heap);
+
+		prev = NULL;
+		node = heap->mm_heapstart[region];
+		next = (struct mm_allocnode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
+			if (prev && prev->size != (node->preceding & ~MM_ALLOC_BIT)) {
```
sunghan-chang, line 135: To understand easier and faster, how about using some macros?
```
#define MM_PREV_NODE_SIZE(x)  ((x)->preceding & ~MM_ALLOC_BIT)
#define IS_ALLOCATED_NODE(x)   ((x)->preceding & MM_ALLOC_BIT)
#define IS_FREE_NODE(x)               (!IS_ALLOCATED_NODE(x))
```

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Fixed.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Fixed

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Done.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: I want to keep the check corruption api separate. Also, even if there is a heap corruption, it is possible that we can continue to execute until the corrupted node actually causes a problem. So, I think its ok to print the corrupted node info and then proceed to print the heapinfo as requested by the user. If there is a problem in some node, and heapinfo does not work properly, then it is ok because we have already printed the corrupted node info.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: This was in original code. But you are right. I will change it.

apps/system/utils/utils_heapinfo.c
kishore-sn: Sure.. I had something else in mind when i moved this down. 
But, maybe its not required. I will revert it.

os/mm/mm_heap/mm_mallinfo.c
jeongchanKim: I modified the original in https://github.com/Samsung/TizenRT/pull/4791, Anyway Thanks.

os/arch/arm/src/armv8-m/up_assert.c
kishore-sn: But, I have added heapinfo_check_corruption() API in heapinfo file. 
It will not get built when heapinfo is not enabled. If you think its ok to not have owner info at some times, then i will move this code to mallinfo file and call the api from other places.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Since we are checking for heap corruption, i think we dont need this check. 
A condition of last node in the loop not matching the heapend will occur only if one of the nodes in the middle is corrupted. 
But, i think we could instead add DEBUGASSERT in the heap corruption check api so that during debugging, we will halt if heap corruption is seen. But, in the release binary, we will let the execution to continue even with heap corruption. 
Please let me know if its ok.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Done

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Fixed.

os/mm/mm_heap/mm_mallinfo.c
sunghan-chang: Even we can know heap corruption at mm_mallinfo, we can't know it at that time.
We will get assert some time later, or it could cause some strange symptom without assert.
mm_mallinfo is not mandatory API so that adding one conditional does not affect performance degradation. Then I think why not.

os/arch/arm/src/armv8-m/up_assert.c
sunghan-chang: Location can't be a problem. We can move it to where we want.
We need to decide whether we will show memory corruption without debug information or not.
In my opinion, it will be definitely helpful. We can figure out the problem. And then we can do next step.
But if we don't show, then we should find what causes the assert.

sunghan-chang:
@kishore-sn Could you check violations of coding rules and build break?

kishore-sn:
I moved DEBUGASSERT() out of mm_check_heap_corruption(), because it results in recursive corruption.

`mm_check_heap_corruption() -> DEBUGASSERT() -> assert handler -> mm_check_heap_corruption(`

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -73,9 +87,112 @@
...
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
+			if (prev && prev->size != MM_PREV_NODE_SIZE(node)) {
+				mdbg("#########################################################################################\n");
+				mdbg("ERROR: Heap node corruption detected\n");
+				dump_node(prev, TYPE_OVERFLOWED);
+				dump_node(node, TYPE_CORRUPTED);
+				mdbg("#########################################################################################\n");
+				mm_givesemaphore(heap);
+				return -1;
+			} else if (node->size != MM_PREV_NODE_SIZE(next)) {
```
Taejun-Kwon, line 150: One question, node is not null all the time?

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -73,9 +87,112 @@
  * Private Functions
  ****************************************************************************/
 
+static void dump_node(struct mm_allocnode_s *node, node_type_t type)
+{
+	if (type == TYPE_CORRUPTED) {
```
Taejun-Kwon, line 92: Very minor comment, When you checking type of enum, you'd better use switch-case for expandability.
If there's no chance that expand node_type_e, please ignore this comment.

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -73,9 +87,112 @@
...
+		 */
+
+		mm_takesemaphore(heap);
+
+		prev = NULL;
+		node = heap->mm_heapstart[region];
+		next = (struct mm_allocnode_s *)((char *)node + node->size);
+
+		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
+			if (prev && prev->size != MM_PREV_NODE_SIZE(node)) {
```
Taejun-Kwon, line 142: All branch return -1, you'd better use below style.
```
if (a) {
   return -1;
}

if (b) {
   return -1;
}
...
```

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: What I meant is that, if there is a corruption in any node, it will assert at line 218. 
`DEBUGASSERT(mm_check_heap_corruption(heap) == OK)`

There will be no need to go till end of heap. 
But, if the for loop actually goes till end of loop without asserting at line 218, then it means there is no corruption. 
So, this check is redundant. 

`DEBUGASSERT(node == heap->mm_heapend[region]);`

os/arch/arm/src/armv8-m/up_assert.c
kishore-sn: Done.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Node is just an address in the heap. It cannot be NULL.

os/mm/mm_heap/mm_mallinfo.c
Taejun-Kwon: Depth is not quiet deep here, so I agree with your answer.
But please understand that above style is very good habit for readability.

os/mm/mm_heap/mm_mallinfo.c
sunghan-chang: Oh I didn't find line 217. But in my opinion, adding `ASSERT(node == heap->mm_heapend[region]);` is better instead adding line 217.
1. This is working only when both config_debug and config_debug_mm_heapinfo are enabled. I want to check heap corruption anytime. (I mean it should work without any conditional).
2. This causes walking entire nodes twice. If you want to add this here, then you should merge this and second for loop.

The number of checking entire node is
|    | Adding one line at end | Current |
|---------------------------|------|-----|
| No Memory Corrupted | 1 | 2 |
| Some Memory Corrupted | 2 | 3 |

kishore-sn: > This is working only when both config_debug and config_debug_mm_heapinfo are enabled. I want to check heap corruption anytime. (I mean it should work without any conditional).

Sorry, I missed removing the heapinfo condition. Also, I forgot about DEBUGASSERT. I modified it as below. So, it will always check corruption:

`        if (mm_check_heap_corruption(heap) != OK) {`
`                DEBUGASSERT(0);`
`        }`

> This causes walking entire nodes twice. If you want to add this here, then you should merge this and second for loop.

I understand your concern. But, I thought it would be better to separate it into a different API which can be easily called from assert handler, heapinfo and any other location as we require. If I merge the new API into mallinfo, then we will have problem of recursive assertion:

`mallinfo -> corruption detected -> debug assert -> assert handler -> mallinfo`

Even if we dont use debugassert for corruption, its possible that corrupted node might cause assertion in mallinfo and then assert handler will again call mallinfo recursively. So, I feel its better to keep it as a separate API.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -472,10 +476,17 @@ void up_assert(const uint8_t *filename, int lineno)
...
+#endif  /* CONFIG_APP_BINARY_SEPARATION */
 
 	up_dumpstate();
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	if (!is_kernel_fault) {
+		mm_check_heap_corruption((struct mm_heap_s *)(this_task()->uheap));
+	} else
+#endif
+	mm_check_heap_corruption(g_kmmheap);
```
Taejun-Kwon, line 488: To make same level of tab, 
```
#endif
{
      mm_check_heap_corruption(g_kmmheap);
}

File: os/mm/mm_heap/mm_mallinfo.c (hunk 7 above)
Comment (sunghan-chang) [reply os/mm/mm_heap/mm_mallinfo.c:141]: I completely agree about separate API. But I don't want to call it from mm_mallinfo.
In mm_mallinfo, just check simple one conditional, `ASSERT(node == heap->mm_heapend[region]);`.
If there is corruption, then it will call assert and assert calls your new API.
In this PR, what I want is removing calling `mm_check_heap_corruption` in mm_mallinfo and adding above at end of the function.

File: os/mm/mm_heap/mm_mallinfo.c [hunk 15]
```diff
@@ -99,6 +213,10 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 #define region 0
 #endif
 
+	if (mm_check_heap_corruption(heap) != OK) {
+		DEBUGASSERT(0);
```
sunghan-chang, line 217: DEBUGASSERT is meaningful when CONFIG_DEBUG is enabled. When it is disabled, DEBUGASSERT is the same as nothing. I think this case means heap is already corrupted. So ASSERT is right. not debugassert.

os/arch/arm/src/armv7-m/up_assert.c
kishore-sn: Done.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: OK. Done. 
I was under the assumption that we can let the system run even if heap is corrupted, until it causes some assert by accessing the corrupted node. But, I agree that it will be difficult to debug in such case. So, its good to change it to ASSERT.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Done.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -105,6 +114,11 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 	FAR struct mm_freenode_s *fnode;
 #endif
 
+	if (mm_check_heap_corruption(heap) != OK) {
+		ASSERT(0);
+	}
+
```
amogh-samsung, line 120: Do we need a double line space here?

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -64,6 +64,19 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+#define MM_PREV_NODE_SIZE(x)	((x)->preceding & ~MM_ALLOC_BIT)
+#define IS_ALLOCATED_NODE(x)	((x)->preceding & MM_ALLOC_BIT)
+#define IS_FREE_NODE(x)		(!IS_ALLOCATED_NODE(x))
```
amogh-samsung, line 69: Could you align these MACRO definitions please?

jeongchanKim (approved):
LGTM

jeongchanKim:
Please resolve the travis error.

kishore-sn:
> Please resolve the travis error.

The error seems to be coming in artik055s/audio config. 
And it is not related to any of my changes. It is kernel_tc_driver.c. Does this need to be fixed?

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Fixed.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: It is already aligned using tabs and looking find in vim editor. 
Not sure why the issue here. Will try to align again.

sunghan-chang:
@kishore-sn Could you confirm there is no error with #4802 ? Because I saw another error about typecasting for `blink->flink == node`.

sunghan-chang:
@kishore-sn The build break in kernel_test_drv was resolved by PR #4802 . Could you rebase yours for build test of this PR?

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -105,6 +114,10 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 	FAR struct mm_freenode_s *fnode;
 #endif
 
+	if (mm_check_heap_corruption(heap) != OK) {
+		ASSERT(0);
```
sunghan-chang, line 118: ASSERT already includes if conditional so that below is better.
```c
ASSERT(mm_check_heap_corruption(heap) == OK);
```

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -157,6 +246,7 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 		mm_givesemaphore(heap);
 
 		mvdbg("region=%d node=%p heapend=%p\n", region, node, heap->mm_heapend[region]);
+		DEBUGASSERT(node == heap->mm_heapend[region]);
```
sunghan-chang, line 249: As I said before, ASSERT is correct I think. From your reply, you've agreed. If right, could you change it to ASSERT?

os/mm/mm_heap/mm_mallinfo.c
```diff
@@ -73,9 +86,112 @@
  * Private Functions
  ****************************************************************************/
 
+static void dump_node(struct mm_allocnode_s *node, node_type_t type)
```
sunghan-chang, line 89: How about making new file for new added? Because they are not mm_mallinfo.

sunghan-chang: I think we can make mm_check_heap_corruption.c.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Done.

os/mm/mm_heap/mm_mallinfo.c
kishore-sn: Done.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Done.

kishore-sn:
In check heap corruption api, I changed mdbg to dbg, because if there is a heap corruption, we would like to get the details even if mdbg is not enabled.

sunghan-chang:
> In check heap corruption api, I changed mdbg to dbg, because if there is a heap corruption, we would like to get the details even if mdbg is not enabled.

Ok, that's good.

sunghan-chang (approved):

PR #4802, opened 2021-02-16
fix build error

deepaksrma:
- this commit fix the build error of commit 057dec00ab9911dbcbf43379e11fc923ab31d056

Signed-off-by: Deepak sharma <deepak.sh@samsung.com>

### DIFF
--- a/apps/examples/testcase/le_tc/network/Kconfig
+++ b/apps/examples/testcase/le_tc/network/Kconfig
@@ -53,6 +53,7 @@ config TC_NET_SOCKET
 config TC_NET_PBUF
 	bool "pbuf() api "
 	default n
+	select DRIVERS_KERNEL_TEST
 
 config TC_NET_SETSOCKOPT
 	bool "setsockopt() api"
--- a/os/drivers/kernel_test/Make.defs
+++ b/os/drivers/kernel_test/Make.defs
@@ -16,11 +16,10 @@
 #
 ###########################################################################
 
+ifeq ($(CONFIG_DRIVERS_KERNEL_TEST),y)
 include kernel_test$(DELIM)kernel$(DELIM)Make.defs
 include kernel_test$(DELIM)network$(DELIM)Make.defs
-
 CSRCS += kernel_test_drv.c
-
 DEPPATH += --dep-path kernel_test
 VPATH += :kernel_test
-
+endif


os/drivers/kernel_test/Make.defs
```diff
@@ -19,7 +19,11 @@
 include kernel_test$(DELIM)kernel$(DELIM)Make.defs
 include kernel_test$(DELIM)network$(DELIM)Make.defs
 
+ifeq ($(CONFIG_DRIVERS_KERNEL_TEST),y)
 CSRCS += kernel_test_drv.c
+else ifeq ($(CONFIG_TC_NET_PBUF),y)
+CSRCS += kernel_test_drv.c
```
pillip8282, line 25: CONFIG_DRIVERS_KERNEL_TEST and CONFIG_TC_NET_PBUF are different layer. so this code seems not good to me. You'd better to add dependency between both config in Kconfig. Please check 'depends on' and 'select' command in Kconfig

pillip8282 (changes requested):

os/drivers/kernel_test/Make.defs
deepaksrma: changed. check now

thapav (commented):
LGTM

Taejun-Kwon:
Problem is happened because pbuf using kernel_test_drv which added for kernel test.
So below is needed
1. rename kernel_test_drv to something common name for both of kernel & network. Of course another module can use it.
    Or network should have its own virtual driver for test
2.  If we allow kernel_test_drv for all other module, CONFIG_DRIVERS_KERNEL_TEST also need to be renamed.
3. Each of module should enable that config in their defconfig.

Taejun-Kwon (changes requested):
please see my comment.

apps/examples/testcase/le_tc/network/Kconfig
```diff
@@ -53,6 +53,7 @@ config TC_NET_SOCKET
 config TC_NET_PBUF
 	bool "pbuf() api "
 	default n
+	select DRIVERS_KERNEL_TEST
```
sunghan-chang, line 56: Please split the commit for this.

os/drivers/kernel_test/Make.defs
```diff
@@ -19,7 +19,9 @@
 include kernel_test$(DELIM)kernel$(DELIM)Make.defs
 include kernel_test$(DELIM)network$(DELIM)Make.defs
 
+ifeq ($(CONFIG_DRIVERS_KERNEL_TEST),y)
```
sunghan-chang, line 22: Please cover above  as well.
```
ifeq (...)
include
include
CSRCS +=
DEPPATH +=
VPATH +=
endif
```

sunghan-chang (commented):
And please cleanup including header in kernel_test_drv.c

sunghan-chang:
@Taejun-Kwon @deepaksrma I agree your opinion. Let's rename it but let's do that with another PR. Because it's not for resolving build break. Resolving build break is more important so that let's finish another thing first and merge this first.

sunghan-chang:
@deepaksrma Let's cleanup including header with another pr.

sunghan-chang (approved):

PR #4805, opened 2021-02-18
network/wifimanager: fix assert issue in wifi stress test

gexuyan:
Increase the priority of wifi msg handler thread from 100 to 101 to avoid the assert issue in wifi stress test.
Please note that this is a workaround for the assert issue in wifi stress test.

### DIFF
--- a/framework/src/wifi_manager/wifi_manager_msghandler.c
+++ b/framework/src/wifi_manager/wifi_manager_msghandler.c
@@ -67,7 +67,7 @@ static int _process_msg(int argc, char *argv[])
  */
 int wifimgr_run_msghandler(void)
 {
-	int tid = task_create("wifi msg handler", 100, 4096, (main_t)_process_msg, NULL);
+	int tid = task_create("wifi msg handler", 101, 4096, (main_t)_process_msg, NULL);
 	if (tid < 0) {
 		WM_ERR;
 		return -1;


Taejun-Kwon:
TizenRT follows round robin scheduling, so I think if there are many threads in product level, assert will occur again.
This is not good solution, please consider it more and more.

Taejun-Kwon (changes requested):
please see my comment

davidfather:
> TizenRT follows round robin scheduling, so I think if there are many threads in product level, assert will occur again.
> This is not good solution, please consider it more and more.

@pillip8282 I also agree with @Taejun-Kwon. This code is for all Wi-Fi chipsets, not for only RTK.

framework/src/wifi_manager/wifi_manager_msghandler.c
```diff
@@ -67,7 +67,7 @@ static int _process_msg(int argc, char *argv[])
  */
 int wifimgr_run_msghandler(void)
 {
-	int tid = task_create("wifi msg handler", 100, 4096, (main_t)_process_msg, NULL);
+	int tid = task_create("wifi msg handler", 101, 4096, (main_t)_process_msg, NULL);
```
davidfather, line 70: This temporary fix should be avoided because wifi_manager is for all Wi-Fi chipsets, not for only RTK.

gexuyan:
will raise another PR for this issue~

PR #4806, opened 2021-02-19
external/mbedtls: apply SECLINK changes in mbedtls alt

an4967:
- apply SECLINK changes in mbedtls alt

### DIFF
--- a/external/include/mbedtls/alt/common.h
+++ b/external/include/mbedtls/alt/common.h
@@ -25,8 +25,8 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define ECP_KEY_INDEX (1)
-#define RSA_KEY_INDEX (2)
+#define ECP_KEY_INDEX (32)
+#define RSA_KEY_INDEX (32)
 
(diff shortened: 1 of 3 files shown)


pillip8282 (approved):
lgtm

external/include/mbedtls/alt/common.h
```diff
@@ -25,8 +25,8 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define ECP_KEY_INDEX (1)
-#define RSA_KEY_INDEX (2)
+#define ECP_KEY_INDEX (32)
+#define RSA_KEY_INDEX (32)
```
davidfather, line 29: You need to explain why ECP and RSA have the same index here.

an4967: @davidfather 
In SE APIs, the index is divided into ROM & RAM areas.
The ROM area uses 0~31 index, the RAM area does 31~63. Hence, the stare index should be 32 for using ROM area.
Also, SE APIs can avoid duplicated index by checking empty slot, so the key indexs are okay even starts from same number.

davidfather (approved):
Looks good to me.

sunghan-chang (approved):

PR #4811, opened 2021-02-24
os/mm/mm_heap: Fix assertion fail during mm_check_heap_corruption

kishore-sn:
Assertion failure occurs when we try to aquire heap semaphore
in mm_check_heap_corruption. So, we add a check for whether we
are in irq mode, and try to aquire semaphore only when we are
not in irq mode.

### DIFF
--- a/os/mm/mm_heap/mm_check_heap_corruption.c
+++ b/os/mm/mm_heap/mm_check_heap_corruption.c
@@ -54,9 +54,9 @@
  * Included Files
  ****************************************************************************/
 #include <tinyara/config.h>
-#include <debug.h>
+#include <tinyara/arch.h>
 #include <tinyara/mm/mm.h>
-
+#include <debug.h>
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -127,7 +127,12 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
 		 * Retake the semaphore for each region to reduce latencies
 		 */
 
-		mm_takesemaphore(heap);
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
+		if (!up_interrupt_context())
+#endif
+		{
+			mm_takesemaphore(heap);
+		}
 
 		prev = NULL;
 		node = heap->mm_heapstart[region];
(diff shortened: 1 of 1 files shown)


sunghan-chang:
> However, it is not required to take a heap semaphore in any
case while checking for heap corruption. This is because, in this API
we just check each heap node for corruption and print some debug info
if corruption is detected.

Let me give you an example:
1. call mm_check_heap_corruption 
2. in the middle of it, context switched
3. new thread does mm so that some node is changed (heap is not corrupted, it's ok)
Then, can we always get right return? I think we can get wrong information.

sunghan-chang:
freenode A with size 10, next allocnode B with size 10.
mm_check_heap_corruption gets size of node A and then context switched (node, next and prev is fixed)
new thread allocates node A with size 4. so
allocnode A' wit size 4, freenode C with size 6, free node B with size 10 (A -> A' and C)

When mm_check_heap_corruption reruns, 
node->size = 4 but next->preceding = 10. Because node and next is fixed.

sunghan-chang:
@kishore-sn Could you check my example?

kishore-sn:
@sunghan-chang 
I understand what you mean. 
But in that case, we always require to take heap semaphore to perform heap check. 
So, heap check cannot be performed from any irq handler. 
But i also dont understand why we dont see this issue in case of recovery test, because even in that case, we are in irq mode due to memfault handler. I will check more on all these aspects.

jeongchanKim (approved):
LGTM

sunghan-chang:
@kishore-sn Because of that, some APIs include checking irq state and do different operation. Can we do the same?

kishore-sn:
Yes, I also thought of using up_interrupt_context().
But, in that case, if we do heap check in irq context, then will it cause a problem?

sunghan-chang:
It is used only in exception, not normal interrupt handler so that I think no problem.

kishore-sn:
@sunghan-chang 
I have changed accordingly.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -177,13 +197,24 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
 					dbg("Corrupted node flink(0x%08x) and next node blink(0x%08x) do not match\n", ((struct mm_freenode_s *)node)->flink,
 							((struct mm_freenode_s *)node)->flink->blink);
 					dbg("#########################################################################################\n");
-					mm_givesemaphore(heap);
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
```
sunghan-chang, line 200: Could you let know why user context does not require this?

sunghan-chang:
Could you check build break?

os/mm/mm_heap/mm_check_heap_corruption.c
kishore-sn: The API up_interruptcontext is not present in user side. 
Also, if we are currently executing in user side, then it is impossible to be in irq mode.

sunghan-chang (approved):

PR #4812, opened 2021-02-24
arch/arm : Add reboot_reason_trywrite in up_assert

jeongchanKim:
1) reboot_reason_trywrite checks that there is already written reboot reason. If no, write REBOOT_SYSTEM_USER_INTENDED.
2) In up_assert, if there is user intended assert, not data abort or prefetch abort, calls reboot_reason_trywrite().

### DIFF
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -88,6 +88,9 @@
 
 #include <tinyara/mm/mm.h>
 
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+#include <arch/reboot_reason.h>
+#endif
 #include "sched/sched.h"
 #ifdef CONFIG_BOARD_ASSERT_AUTORESET
 #include <sys/boardctl.h>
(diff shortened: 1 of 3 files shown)


os/include/tinyara/arch.h
```diff
@@ -2179,6 +2179,10 @@ void up_wdog_keepalive(void);
 
 #endif
 
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
```
sunghan-chang, line 2182: Another reboot reason prototype is in os/arch/arm/include/reboot_reason.h. I think the same file is right for this.

os/arch/arm/src/common/up_reboot_reason.c
```diff
@@ -0,0 +1,38 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: reboot_reason_trywrite
+ ****************************************************************************/
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+void reboot_reason_trywrite(void)
```
sunghan-chang, line 31: The name can't represent the operation. How about `reboot_reason_write_user_intended`?

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -457,6 +457,10 @@ void up_assert(const uint8_t *filename, int lineno)
 {
 	board_led_on(LED_ASSERTION);
 
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	reboot_reason_trywrite();
```
sunghan-chang, line 461: Let's split commits for this change
1. add new api support on reboot reason
2. apply it on each chip

sunghan-chang:
@jeongchanKim
arch/arm : Add reboot_reason_write_user_intended function support
arch/armv8-m/up_assert : write reboot_reason_write_user_intended

sunghan-chang:
Title and content are not matched. Could you rewrite them like above?

sunghan-chang (approved):

PR #4813, opened 2021-02-24
/se/ameba: Modify RTL SE HAL Free API

ziyik:
- Add error prevention check in RTL SE Secure
- Modify SE HAL Free API
- Data pointer in non-secure should be free is non-secure, passing pointer to free in secure will cause error


os/se/ameba/security_ameba_wrapper_tz.c
```diff
@@ -201,16 +201,18 @@ int se_ameba_hal_deinit(void)
 int se_ameba_hal_free_data(hal_data *data)
 {
 	AWRAP_ENTER;
-	int ret = HAL_SUCCESS;
 
-	up_allocate_secure_context(CONFIG_SE_SECURE_CONTEXT_SIZE);
-	ret = ameba_hal_free_data(data);
```
sunghan-chang, line 207: Then, who calls this?

sunghan-chang: Is that moved into cmse_implib.a?

ziyik: Currently no one need to use this, this will pass data pointer into secure to free it.
But it is not needed, because in non-secure there is no secure pointer.
So it have be be remove and replace with normal free.
Previously it did not cause problem due to no one is using it.

sunghan-chang (approved):

PR #4815, opened 2021-02-25
arch/arm/src/stm32l4: Update arm conditional flag for stm32l4 board.

thapav:
This patch updates the stm32l4 board interrupt conditional with
appropriate arm cortex flag for unification across different chipsets.

### DIFF
--- a/os/arch/arm/src/stm32l4/stm32l4_irq.c
+++ b/os/arch/arm/src/stm32l4/stm32l4_irq.c
@@ -373,7 +373,7 @@ void up_irqinitialize(void)
    * Fault handler.
    */
 
-#ifdef CONFIG_ARM_MPU
+#ifdef CONFIG_ARMV7M_MPU
   irq_attach(STM32L4_IRQ_MEMFAULT, up_memfault, NULL);
   up_enable_irq(STM32L4_IRQ_MEMFAULT);
 #endif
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/stm32l4/stm32l4_irq.c
```diff
@@ -373,7 +373,7 @@ void up_irqinitialize(void)
    * Fault handler.
    */
 
-#ifdef CONFIG_ARM_MPU
+#ifndef CONFIG_ARMV7M_MPU
```
sangam-swami, line 376: is it with #ifdef?

sangam-swami: can you detail the problem for this PR

sunghan-chang: I don't know why previous is wrong. Could you let know detail as @sangam-swami said?

thapav: Oh yes!, corrected that.
It is currently using ARM_MPU config, whereas when compared to other chipset codes, cortex specific flag should be used.

sunghan-chang: We can use both ARM_MPU and ARMV7M_MPU. No difference between them. I don't know this is necessary. If you want to be the same as others, please write that even we can use ARM_MPU, we will change it for unification in commit description.

thapav: Yes that is the purpose, I have modified the description.

sunghan-chang (approved):

PR #4822, opened 2021-03-02
os/Makefile.unix : Update the version to 3.1

jeongchanKim:


os/Makefile.unix
```diff
@@ -300,7 +300,7 @@ tools/mkversion$(HOSTEXEEXT):
 
 $(TOPDIR)/.version: force_build
 	echo "create .version file"; \
-	tools/version.sh -v 3.0 .version; \
+	tools/version.sh -v 3.1 .version; \
```
sunghan-chang, line 303: Could you do this for Makefile.win as well?

PR #4823, opened 2021-03-02
network/lwip: fix ipv6 kconfig

pillip8282:
ICMPv6 and neighbor discovery don't have type. so it was removed after
menuconfig.

### DIFF
--- a/os/net/lwip/configs/icmp6/Kconfig
+++ b/os/net/lwip/configs/icmp6/Kconfig
@@ -3,9 +3,10 @@
 # see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
 #
 
-menu "ICMP6 Support"
+menu "ICMPv6 (RFC 4443)"
 
 config NET_IPv6_ICMP
+	bool
 	depends on NET_IPv6
 	default y
 	---help---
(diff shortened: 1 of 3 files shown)


os/net/lwip/configs/nd/Kconfig
```diff
@@ -3,9 +3,10 @@
 # see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
 #
 
-menu "ND Support"
+menu "Neighbor Discovery (RFC 4861)"
 
 config NET_IPv6_ND
```
sunghan-chang, line 8: There is no code for this. Why do you need this config?

os/net/lwip/configs/icmp6/Kconfig
```diff
@@ -3,9 +3,10 @@
 # see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
 #
 
-menu "ICMP6 Support"
+menu "ICMPv6 (RFC 4443)"
 
 config NET_IPv6_ICMP
```
sunghan-chang, line 8: same as previous

os/net/lwip/configs/nd/Kconfig
pillip8282: for compatibility. there are many files to do fix.  and it's not fully tested. I'll remove it later

os/net/lwip/configs/icmp6/Kconfig
pillip8282: for compatibility. there are many files to do fix.  and it's not fully tested. I'll remove it later

sunghan-chang (approved):

PR #4824, opened 2021-03-02
os/arch/arm/amebad_reboot_reason : Add return of REBOOT_REASON_INITIA…

jeongchanKim:
…LIZED in up_reboot_reason_read()

If there is no written reboot reason and it is not hw reset or watchdog reset, then return REBOOT_REASON_INITIALIZED.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_reboot_reason.c
+++ b/os/arch/arm/src/amebad/amebad_reboot_reason.c
@@ -55,6 +55,7 @@
 
 #include <tinyara/config.h>
 
+#include <stdbool.h>
 #include <tinyara/reboot_reason.h>
 #include "ameba_soc.h"
 /****************************************************************************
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -79,6 +79,9 @@ reboot_reason_code_t up_reboot_reason_read(void)
 	if (backup_reg != REBOOT_REASON_INITIALIZED) {
 		return backup_reg;
 	} else {
+		if (BKUP_Read(BKUP_REG1) == REBOOT_REASON_INITIALIZED) {
```
sunghan-chang, line 82: I don't know why this is necessary. Could you explain more?

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -110,4 +111,13 @@ void up_reboot_reason_clear(void)
 	up_reboot_reason_write(REBOOT_REASON_INITIALIZED);
 	backup_reg = REBOOT_REASON_INITIALIZED;
 }
+
+bool up_reboot_reason_is_written(void)
+{
+	if (BKUP_Read(BKUP_REG1) != REBOOT_REASON_INITIALIZED) {
+		return true;
```
sunghan-chang, line 118: Let's include the header for `true` and `false`.

jeongchanKim: It is added, `stdbool.h`.

sunghan-chang (approved):

PR #4825, opened 2021-03-03
apps/testcase/kernel/tc_umm_heap : Change sleep to waitpid and Remove…

jeongchanKim:
… unnecessary task_delete

1) For waiting umm testcase task, change sleep to waitpid.
2) task_delete(0) terminates its own, but not needed in this case. So remove it.


apps/examples/testcase/le_tc/kernel/tc_umm_heap.c
```diff
@@ -318,8 +318,10 @@ static int umm_task(int argc, char *argv[])
 
 int umm_heap_main(void)
 {
-	task_create("umm_task", 150, 2048, umm_task, (char * const *)NULL);
-	sleep(1);
+	int pid;
+	int stat_loc;
+	pid = task_create("umm_task", 150, 2048, umm_task, (char * const *)NULL);
+	waitpid(pid, &stat_loc, 0);
```
pillip8282, line 324: Is checking return of waitpid() unnecessary?

sunghan-chang (approved):

sunghan-chang (approved):

PR #4837, opened 2021-03-05
board/rtl8721csm/bluetooth: Use macro to control bt trace log & fw log switch

Pangwawaa:


sunghan-chang:
Could you use the same title between commit and pr?
If you write commit title like pr, then you don't need to manually modify PR title.
In history point of view, commit is used, not pr so that we think commit is important.

sunghan-chang:

Pangwawaa:
> Could you use the same title between commit and pr?
> If you write commit title like pr, then you don't need to manually modify PR title.
> In history point of view, commit is used, not pr so that we think commit is important.

I have modified the commit message.

pillip8282 (approved):
lgtm

davidfather (approved):
lgtm.

PR #4842, opened 2021-03-09
drivers/lwnl: fix lwnl return type

pillip8282:
fix lwnl return type

### DIFF
--- a/os/drivers/lwnl/lwnl.c
+++ b/os/drivers/lwnl/lwnl.c
@@ -28,7 +28,6 @@
 #include <errno.h>
 #include <debug.h>
 #include <net/if.h>
-#include <tinyara/kmalloc.h>
 #include <tinyara/fs/fs.h>
 #include <tinyara/lwnl/lwnl.h>
 #include "lwnl_evt_queue.h"
@@ -122,6 +121,11 @@ static int lwnl_open(struct file *filep)
 	struct lwnl_upperhalf_s *upper = inode->i_private;
 
 	LWNLDEV_LOCK(upper);
+	/*
+	 * crefs type is uint8 so overflow might happen.
+	 * However there are a few modules which use lwnl.
+	 * So checking overflow is not needed now.
+	 */
 	upper->crefs++;
 	LWNLDEV_UNLOCK(upper);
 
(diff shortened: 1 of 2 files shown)


os/fs/vfs/fs_select.c
```diff
@@ -266,7 +266,7 @@ int select(int nfds, FAR fd_set *readfds, FAR fd_set *writefds, FAR fd_set *exce
 
 	DEBUGASSERT(ndx == npfds);
 	if (ndx != npfds) {
-		set_errno(EINVAL);
+		set_errno(EBADF);
```
jeongarmy, line 269: How about keeping EINVAL? In posix, a select set errno as EINVAL when the nfds argument is less than 0 or greater than FD_SETSIZE.

pillip8282: ndx != npfds doesn't mean that nfds is negative or  exceeds the RLIMIT_NOFILE resource limit (see getrlimit(2)). Hence we don't need to modify it

os/drivers/lwnl/lwnl.c
```diff
@@ -122,6 +121,11 @@ static int lwnl_open(struct file *filep)
 	struct lwnl_upperhalf_s *upper = inode->i_private;
 
 	LWNLDEV_LOCK(upper);
+	/*
+	 * crefs type is uint8 so overflow might happen.
+	 * However there are a few modules which use lwnl.
+	 * So checking overflow is not needed now.
```
davidfather, line 127: Do you mean that overflow checking will be done later by some modules which use lwnl?

os/drivers/lwnl/lwnl.c
```diff
@@ -204,7 +212,7 @@ static int lwnl_ioctl(struct file *filep, int cmd, unsigned long arg)
 	LWNL_ENTER;
 	int res = lwnl_add_listener(filep);
 	if (res < 0) {
-		res = -ENOSYS;
+		res = -EBADF;
```
davidfather, line 215: In this function, there are two cases where the return type is set to '-EBADF'. How could you tell them apart?

os/drivers/lwnl/lwnl.c
pillip8282: I'll modify it if many modules use lwnl. However only wi-fi manager and iotivity access lwnl now. So I think we don't need to add overflow logic now.

os/drivers/lwnl/lwnl.c
pillip8282: According to ioctl manual page. ioctl can return following errno values.
---
EBADF  fd is not a valid file descriptor.

       EFAULT argp references an inaccessible memory area.

       EINVAL request or argp is not valid.

       ENOTTY fd is not associated with a character special device.

       ENOTTY The specified request does not apply to the kind of object that the file descriptor fd references
---
There is no relevant errno value for this. If you have good idea, I'll apply it.

davidfather (approved):
Looks good to me.

PR #4843, opened 2021-03-09
board/rtl8721csm/flash: Add flow control mechanism for 256BWrite

AdamZhang0124:
1. Add flow control mechanism on 256BWrite function 
2. Move the new 256BWrite function to RAM 
3. Use the updated function instead of FLASH_TxData256B in flash_stream_write

### DIFF
--- a/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_common/rtl8721d_flash_ram.c
+++ b/os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_common/rtl8721d_flash_ram.c
@@ -541,4 +541,83 @@ void FLASH_ClockSwitch(u32 Source, u32 Protection)
 	}
 #endif
 }
+
+/**
+* @brief  This function is used to write data to flash in OneBitMode and User Mode.
+* @param  StartAddr: Start address in flash from which SPIC writes.
+* @param  DataPhaseLen: the number of bytes that SPIC sends in Data Phase.
+* @param  pData: pointer to a byte array that is to be sent.
+* @note
+*		- page program(256B) time typical is 0.7ms: BaudRate=2.9Mbps, so one bit mode is enough.
+*		- use FLASH_SW_CS_Control to protect flash write
+*		- flow control is added to prevent fifo from overflow
+* @retval none
+*/
+IMAGE2_RAM_TEXT_SECTION
+void FLASH_TxData256B_RAM(u32 StartAddr, u32 DataPhaseLen, u8* pData)
+{
+	SPIC_TypeDef *spi_flash = SPIC;
+	u32 tx_num= 0;
+	u32 ctrl0 = spi_flash->ctrlr0;
+	u8 addrbyte[4];
+	u8 fifo_alarm_level = 28;
+
+	assert_param(DataPhaseLen <= 256); /* max len is 256 */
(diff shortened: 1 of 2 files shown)


sunghan-chang:
@jeongarmy Could you verify resolving the issue for file operation with network?

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_common/rtl8721d_flash_ram.c
```diff
@@ -541,4 +541,83 @@ void FLASH_ClockSwitch(u32 Source, u32 Protection)
...
+void FLASH_TxData256B_RAM(u32 StartAddr, u32 DataPhaseLen, u8* pData)
+{
+	SPIC_TypeDef *spi_flash = SPIC;
+	u32 tx_num= 0;
+	u32 ctrl0 = spi_flash->ctrlr0;
+	u8 addrbyte[4];
+	u8 fifo_alarm_level = 28;
+
+	assert_param(DataPhaseLen <= 256); /* max len is 256 */
+	assert_param((DataPhaseLen % 4) == 0); /* len should be 4 byte aligned */
```
Taejun-Kwon, line 566: I have a question, 
Few months ago I requested to you that check amebad_write which I added to support byte write option, but you didn't reply to me.
Actually amebad_write finally calling flash_stream_write, buf size is not always aligned with 4byte.
Smartfs use sector size as 1024, so size could be 1~1024.
Let's assume that size is 1023, 
In this case it will call this function 3times and then FLASH_TxData12B will be called.
So my question is that FLASH_TxData12B works properly without not 4byte aligned buffer??
Of course address also need to be checked.

AdamZhang0124: Hi @Taejun-Kwon,
I have reviewed the PR 4644 you are referring to.
For the reply you posted after the merge, I thought it was an explanation for my pervious reply, sorry about that.

Let me explain your doubt together at here,

If writing 1023B using flash_stream_write, it will do as follow:
1. Do 256B write for 3 times
2. Do 12B write for 21 times
For the part that cannot use 12B to write:
3. Do 4B write until less than 4B (here does not use)
4. If less than 4B to be written, read 4B and replace the first 1/2/3B  
5. Do the last 4B write

So that it is always 4B aligned.


As for the 'performance not change', my current understanding is that:
For block write, the mtd will cost time to read the whole page, then only replace the part to be written and pass to flash_stream_write. The flash_stream_write will only need to do 256B write once.
For byte write, it saves the time in the mtd part. But in flash_stream_write, it will use the combination of 256B/12B/4B to write, and it will cost more time at here.

For example:
If writing 255B,
Block write: MTD->flash_stream_write->256B write 1 time.
Byte write: MTD->flash_stream_write->12B write 21 times, 4B read 1time,4B write 1 time.
And this is the rough reason of the 'performance not change'.

If the byte write function need to be further investigated to exploit the performance. May need you to start a new request for it.

Many thanks!

sunghan-chang (approved):

sunghan-chang:
@AdamZhang0124 Once @jeongarmy verifies, I will merge.

AdamZhang0124:
@sunghan-chang Noted with thanks!

PR #4848, opened 2021-03-10
Add debug support for armv8-m crashes

thapav:
This PR :
1) Adds display symbol debug support
2) Adds is_kernel_text_space() API
3) Modifies kernel text start & end address variables

### DIFF
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -503,8 +503,10 @@ ifeq ($(CONFIG_UBOOT_UIMAGE),y)
 	fi
 endif
 ifeq ($(CONFIG_DEBUG_DISPLAY_SYMBOL),y)
+	cp ${OUTBIN_DIR}/System.map ${TOOLSDIR}/fs/contents-smartfs/rtl8721csm/base-files/System.map
 	cp ${OUTBIN_DIR}/System.map ${TOOLSDIR}/fs/contents-romfs/System.map
 else
+	$(call DELFILE, ${TOOLSDIR}/fs/contents-smartfs/rtl8721csm/base-files/System.map)
 	$(call DELFILE, ${TOOLSDIR}/fs/contents-romfs/System.map)
 endif
 
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -75,6 +75,7 @@
 #include <sys/types.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <assert.h>
 #include <debug.h>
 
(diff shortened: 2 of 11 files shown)


os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
...
+
+static void up_text_dump(uint32_t sp, uint32_t stack_base)
+{
+	uint32_t stack;
+
+	for (stack = sp & ~0x1f; stack < stack_base; stack += 32) {
+		int i;
+		uint32_t *ptr = (uint32_t *)stack;
+
+		for(i = 0; i < 8; i++) {
```
jeongchanKim, line 245: Please add space after `for`.

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
...
+
+			/* Does the current stack pointer lie within the
+			 * kernel text section ?
+			 */
+			if(is_kernel_text_space((void *)ptr[i])) {
```
jeongchanKim, line 250: Please add space after `if`.

os/arch/arm/src/common/up_checkspace.c
```diff
@@ -38,11 +38,23 @@ extern uint32_t _sdata;
...
 extern uint32_t _sbss;
 extern uint32_t _ebss;
+extern uint32_t _ram_kernel_text_start;
+extern uint32_t _ram_kernel_text_end;
+
+bool is_kernel_text_space(void *addr)
+{
+	/* Check if the address lies in the kernel text section range */
+	if ((addr >= (void *)&_stext && addr <= (void *)&_etext) ||
+		(addr >= (void *)&_ram_kernel_text_start && addr <= (void *)&_ram_kernel_text_end)) {
```
jeongchanKim, line 48: Are `_ram_kernel_text_start` and `_ram_kernel_text_end` working even there is no ram text space?

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
...
+		mid = (first + last) / 2;
+	}
+	if (first > last) {
+		lldbg("symbol is not found in system map\n");
+	}
+
+	/* Close the file */
+	fclose(pFile);
+
+	return;
```
jeongchanKim, line 230: This is not needed. Please remove it.

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
...
+	int word;
+	int ch_in_word;
+	int read_line;
+	int ch_in_line;
+
+	pFile = fopen("/mnt/System.map", "r");
+
+	/* Check if file System.map exists */
+	if (pFile == NULL) {
+		lldbg("Could not open file: /rom/System.map\n");
```
jeongchanKim, line 162: Please change `/rom/System.map` to `/mnt/System.map`. It is much easier to understand.

sangam-swami:
@thapav 
Can you address build failure

os/Makefile.unix
```diff
@@ -503,8 +503,10 @@ ifeq ($(CONFIG_UBOOT_UIMAGE),y)
 	fi
 endif
 ifeq ($(CONFIG_DEBUG_DISPLAY_SYMBOL),y)
+	cp ${OUTBIN_DIR}/System.map ${TOOLSDIR}/fs/contents-smartfs/rtl8721csm/base-files/System.map
```
sunghan-chang, line 506: I don't agree adding board-specific in the common file. Let's find another way.

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
  * Private Functions
  ****************************************************************************/
 
+#ifdef CONFIG_DEBUG_DISPLAY_SYMBOL
+/****************************************************************************
+ * Name: print_dumpstack_symbol
+ * Below API works if there is existance of System.map file in smart fs
```
sunghan-chang, line 140: When do you add map file in rom? when in smartfs? How to configure?

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -133,6 +134,129 @@ extern sq_queue_t g_freemsg_list;
...
+	int mid;
+	char line[128] = { '\0' };
+	char data[6][128] = { {'\0', '\0'} };
+	char c;
+	int word;
+	int ch_in_word;
+	int read_line;
+	int ch_in_line;
+
+	pFile = fopen("/mnt/System.map", "r");
```
sunghan-chang, line 158: I don't agree using this kind function in assert. We can't guarantee the operation. When inode is broken, then it will cause next assert or wrong information.

sunghan-chang:
@sangam-swami @kishore-sn @thapav I strongly disagree calling another APIs like fopen, fxxxx in assert state. Because we don't know what is broken. That's why we directly call hw API like up_getc not uart API. 
We can check some data is broken like heap, stack or so on. But we should not call the API which works with some data.

os/Makefile.unix
```diff
@@ -503,8 +503,10 @@ ifeq ($(CONFIG_UBOOT_UIMAGE),y)
...
 	cp ${OUTBIN_DIR}/System.map ${TOOLSDIR}/fs/contents-romfs/System.map
 else
+	$(call DELFILE, ${TOOLSDIR}/fs/contents-smartfs/rtl8721csm/base-files/System.map)
```
jeongarmy, line 509: I think we need to add a dependency of smartfs and romfs for these commands

os/include/tinyara/arch.h
```diff
@@ -2179,6 +2179,15 @@ void up_wdog_keepalive(void);
 
 #endif
 
+/****************************************************************************
+ * Name: is_kernel_text_space
+ *
+ * Description:
+ *   Check the address is in kernel text space or not
```
abhishek-samsung, line 2186: Please add if after check. "Check if the address is in kernel text space or not"

PR #4850, opened 2021-03-11
[Changed] change ble include rtl8721

taehwan-hwang:
change include path for ble rtl8721 : `da_include/ble`, `da_include/da_core`


sangwon03 (approved):
LGTM

PR #4856, opened 2021-03-16
build/tools/amebad/bootloader/ : Secure boot modifications sync.

fengyunlai:
Modify bootloader for secure boot modifciations.


sunghan-chang:
@fengyunlai Could you let me know what modification is?

sunghan-chang:
@fengyunlai Is it applied that OTA1 region boots when both OTA regions have the same version?

sunghan-chang:
> @fengyunlai Is it applied that OTA1 region boots when both OTA regions have the same version?

Self-answer. I found that was merged.

sunghan-chang (approved):

PR #4867, opened 2021-03-22
lib/libc: Fix testcase failure in tc_rtc

kishore-sn:
tc_rtc fails for settime and rdtime ioctl. This is because, we set
year value as 3. According to unix standard, this value is considered
as year 1903, which is in turn less than EPOCH year of 1970. In such
case, we need to follow unix standard and add 100 to the year value.
So, all year values from 1900 to 1969 will be converted to 2000 to
2069. We update the tc_rtc accordingly.

### DIFF
--- a/apps/examples/testcase/le_tc/drivers/tc_rtc.c
+++ b/apps/examples/testcase/le_tc/drivers/tc_rtc.c
@@ -20,6 +20,7 @@
 /// @brief Test Case Example for rtc driver
 #include <tinyara/config.h>
 #include <tinyara/rtc.h>
+#include <tinyara/time.h>
 #include <stdio.h>
 #include <fcntl.h>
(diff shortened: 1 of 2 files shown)


jeongchanKim (commented):
Could you split the commit?
I think it is better to maintain the code.

jeongchanKim (approved):
LGTM

deepaksrma (commented):
please change in the commit msg of 1st commit as
 _According `to`_
and 
_unix standard suggests that all year values
from `1900` to 1969 must be considered as values between 2000 to
2069 respectively._

also change in the commit msg of 2nd commit as
_The tc_driver_rtc_ioctl fails due to year value being out `of` range._

sunghan-chang (approved):

PR #4868, opened 2021-03-22
os/arch/arm: Prevent recursive abort due to heap check

kishore-sn:
Recursive abort issue happened since the first assertion was caused
due to invalid heap semaphore, and in assert handler we again try
to check heap corruption which will try to aquire the invalid heap
semaphore. Modify heap corruption check to NOT use semaphore during
abort_mode. Also, we would like heap check to run always in case of
assert, irrespective of whether somebody else is holding the sem.

### DIFF
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -109,8 +109,9 @@
 #include "up_internal.h"
 #include "mpu.h"
 
-#ifdef CONFIG_BINMGR_RECOVERY
 bool abort_mode = false;
+
+#ifdef CONFIG_BINMGR_RECOVERY
 extern struct tcb_s *g_faultmsg_sender;
 extern sq_queue_t g_faultmsg_list;
 extern sq_queue_t g_freemsg_list;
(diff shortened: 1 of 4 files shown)


os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -119,6 +119,8 @@ extern sq_queue_t g_freemsg_list;
 #ifdef CONFIG_APP_BINARY_SEPARATION
 extern uint32_t g_assertpc;
 #endif
+
+bool recursive_abort = false;
```
thapav, line 123: Do you think we should check for abort_mode flag as well at the start of up_assert().
This will prevent any other recursive abort scenarios as well ? (Similar to armv7-r)

kishore-sn: The current code has the same effect. Basically recursive_abort needs to be set to true in assert handler.

sunghan-chang:
@kishore-sn Could you update this PR to restart travis?
And you only added it on armv8-m. No problem on 7m?

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -109,8 +109,9 @@
 #include "up_internal.h"
 #include "mpu.h"
 
-#ifdef CONFIG_BINMGR_RECOVERY
 bool abort_mode = false;
+
+#ifdef CONFIG_BINMGR_RECOVERY
```
sunghan-chang, line 114: armv7-m also needs this change.

sunghan-chang: armv7-r as well

kishore-sn: Fixed.

sunghan-chang (approved):

PR #4871, opened 2021-03-25
apps/testcase/filesystem : Add auto calculation of mount devname

jeongchanKim:
In commit bca7c3826e60e, mountpoint is calculcated automatically. So we don't need to define every board's mountpoint.

### DIFF
--- a/apps/examples/testcase/le_tc/filesystem/fs_main.c
+++ b/apps/examples/testcase/le_tc/filesystem/fs_main.c
@@ -54,6 +54,12 @@
 #include "tc_common.h"
 #include "tc_internal.h"
 
+#ifdef CONFIG_AUTOMOUNT_USERFS
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#include <tinyara/os_api_test_drv.h>
+#endif
+
 /****************************************************************************
  * Definitions
  ****************************************************************************/
@@ -62,6 +68,13 @@
 #define SEEK_DEF		3
 #define SEEK_OFFSET		6
 #define MOUNT_DIR "/fsmnt/"
+#define FS_PATH_MAX 15
+
+#ifdef CONFIG_AUTOMOUNT_USERFS
+static char *TMP_MOUNT_DEV_DIR;
+#else
+#define TMP_MOUNT_DEV_DIR "/dev/smart1"
+#endif
 
 #ifdef CONFIG_FS_SMARTFS
 #ifdef CONFIG_SMARTFS_MULTI_ROOT_DIRS
@@ -170,6 +183,10 @@ static void vfs_rmdir(const char *dirpath);
 
 #ifndef CONFIG_DISABLE_ENVIRON
 extern int mount_show(void);
+#endif
+
+#ifdef CONFIG_AUTOMOUNT_USERFS
+static char devname[CONFIG_SMARTFS_MAXNAMLEN];
 #endif
 
 #if defined(CONFIG_PIPES) && (CONFIG_DEV_PIPE_SIZE > 11)
@@ -6103,6 +6120,29 @@ static void tc_fs_driver_ramdisk_ops_p(void)
 }
 #endif
 
+#ifdef CONFIG_AUTOMOUNT_USERFS
+char *get_fs_mount_devname(void)
+{
+	return devname;
+}
+
+void get_userfs_devname(void)
+{
+	int partno;
+	int fd;
+	fd = open(OS_API_TEST_DRVPATH, O_WRONLY);
+	TC_ASSERT_GEQ("open", fd, 0);
+
+	partno = ioctl(fd, TESTIOC_GET_FS_PARTNO, 0);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", partno, 0, close(fd));
+
+	snprintf(devname, FS_PATH_MAX, "/dev/smart0p%d", partno);
+	TMP_MOUNT_DEV_DIR = get_fs_mount_devname();
+
+	close(fd);
+}
+#endif
+
 #ifdef CONFIG_BUILD_KERNEL
 int main(int argc, FAR char *argv[])
 #else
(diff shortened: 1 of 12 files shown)


apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -172,6 +173,30 @@ static void vfs_rmdir(const char *dirpath);
...
+{
+	/* Move to next part information. */
+	while (*(*par)++ != ',');
+}
+
+static int fs_get_mount_partno(void)
+{
+	int partno = 0;
+
+	char *partname = CONFIG_FLASH_PART_NAME;
```
Taejun-Kwon, line 187: In my opinion, we'd better avoid access internal config where in lower level.
Like a pipe or something related to syscall, that is ok.
But this case had better hided to user lvl, because it using by mtd level which never open to user.
Does previous version has some problems?

apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -6113,6 +6137,26 @@ int tc_filesystem_main(int argc, char *argv[])
 		return ERROR;
 	}
 
+#ifdef CONFIG_AUTOMOUNT_USERFS
+	int partno;
+	int fd;
+	fd = open(OS_API_TEST_DRVPATH, O_WRONLY);
+	if (fd < 0) {
```
thapav, line 6144: Can we use TC_ASSERT_XX instead to check for failures??

apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -6113,6 +6137,26 @@ int tc_filesystem_main(int argc, char *argv[])
...
+		printf("ERROR : Fail to open OS API test driver %d\n", get_errno());
+		return ERROR;
+	}
+
+	partno = ioctl(fd, TESTIOC_GET_FS_PARTNO, 0);
+	if (partno < 0) {
```
thapav, line 6150: Can we use TC_ASSERT_XX instead to check for failures??

apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -6103,6 +6120,13 @@ static void tc_fs_driver_ramdisk_ops_p(void)
 }
 #endif
 
+#ifdef CONFIG_AUTOMOUNT_USERFS
+void set_fs_mount_devname(char **path)
```
sunghan-chang, line 6124: 1. It looks "get" is right.
2. How about delivering the value with return value?
 I mean `char *get_fs_mount_devname(void)`.

```
#ifdef CONFIG_AUTOMOUNT_USERFS
static char *TMP_MOUNT_DEV_DIR;
#else
#define TMP_MOUNT_DEV_DIR "/dev/smart1"
#endif

void itc_fs_main(void)
{
#ifdef CONFIG_AUTOMOUNT_USERFS
	TMP_MOUNT_DEV_DIR = get_fs_mount_devname();
#endif
}
```

apps/examples/testcase/le_tc/filesystem/fs_main.c
```diff
@@ -6313,6 +6350,10 @@ int tc_filesystem_main(int argc, char *argv[])
 #ifdef CONFIG_ITC_FS
 	itc_fs_main();
 #endif
+
+#ifdef CONFIG_AUTOMOUNT_USERFS
+	close(fd);
```
sunghan-chang, line 6355: Is there any reason to keep this fd until end of tc?
In my opinion, we could close it after getting the value immediately.

os/drivers/os_api_test/filesystem/test_filesystem.c
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+#include <tinyara/config.h>
+#include <common.h>
+#include <sys/types.h>
```
sunghan-chang, line 20: This looks unnecessary.

os/board/common/partitions.c
```diff
@@ -163,6 +163,21 @@ static void move_to_next_part(const char **par)
 		while (*(*par)++ != ',');
 }
 
+int get_partition_num(char *part)
+{
+	int partno = 0;
+
+	char *partname = CONFIG_FLASH_PART_NAME;
+	while (*partname) {
+		if (!strncmp(partname, part, strlen(part) + 1)) {
```
sunghan-chang, line 172: You should check it with `strlen(part)` and then you should check `,` at next step. Because user will not give the part with `,`.
I mean user will deliver "userfs" not "userfs,".
Then this will return always false.

os/drivers/Makefile
```diff
@@ -66,6 +66,8 @@ ifeq ($(CONFIG_TASK_MANAGER),y)
 CFLAGS += -I $(TOPDIR)/kernel
 endif
 
+CFLAGS += -I $(TOPDIR)/board/common
```
sunghan-chang, line 69: Let's move this into `os/drivers/os_api_test/filesystem/Make.defs`

os/drivers/os_api_test/filesystem/Make.defs
```diff
@@ -0,0 +1,26 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_EXAMPLES_TESTCASE_FILESYSTEM)$(CONFIG_AUTOMOUNT_USERFS),yy)
+CSRCS += test_filesystem.c
+endif
```
sunghan-chang, line 21: Let's cover all of lines.
```
ifeq (...)
CSRCS +=

CFLAGS +=

DEPPATH +=
VPATH +=
endif
```

sunghan-chang (approved):

PR #4873, opened 2021-03-25
apps/system: Support flash read test

jeongarmy:
This reads flash by size from offset and prints them.
Usage :
TASH >> flash_test read <offset> <size>


apps/system/utils/utils_flash.c
```diff
@@ -0,0 +1,151 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+int utils_flash(int argc, char **args)
+{
+	off_t offset;
+	int read_size;
+
+	if (argc >= 4 && !strncmp(args[1], "read", strlen("read") + 1)) {
```
jeongchanKim, line 139: Is there any case that `argc` is greater than 4?

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,102 @@
...
+	offset = 0;
+	remaining = size;
 
+	printf("Read 0x%x: ", addr);
+
+	while (remaining > 0) {
+		read_size = BUF_SIZE < remaining ? BUF_SIZE : remaining;
+		ret = MTD_READ(dev_mtd, addr + offset, read_size, buffer);
+		if (ret <= 0) {
+			printf("Read Failed : %d\n", ret);
```
jeongchanKim, line 111: In this case, we'd better free the `buffer`.

apps/system/utils/utils_flash.c
```diff
@@ -0,0 +1,151 @@
...
+	remaining = size;
+
+	printf("Read 0x%x: ", addr);
+
+	while (remaining > 0) {
+		read_size = BUF_SIZE < remaining ? BUF_SIZE : remaining;
+		ret = MTD_READ(dev_mtd, addr + offset, read_size, buffer);
+		if(ret <= 0) {
+			printf("Read Failed : %d\n", ret);
+			return ERROR;
```
jeongchanKim, line 112: Same as above.

apps/system/utils/utils_flash.c
```diff
@@ -0,0 +1,151 @@
...
+
+	offset = 0;
+	remaining = size;
+
+	printf("Read 0x%x: ", addr);
+
+	while (remaining > 0) {
+		read_size = BUF_SIZE < remaining ? BUF_SIZE : remaining;
+		ret = MTD_READ(dev_mtd, addr + offset, read_size, buffer);
+		if(ret <= 0) {
```
jeongchanKim, line 110: Please add space after `if`.

jeongchanKim (commented):
`utils_flash` and `flash_test` have almost same code.
Is it possible to merge them?

apps/system/flash_test/Makefile
```diff
@@ -100,7 +100,7 @@ else
   INSTALL_DIR = $(BIN_DIR)
 endif
 
-CONFIG_XYZ_PROGNAME ?= flash_eraseall$(EXEEXT)
+CONFIG_XYZ_PROGNAME ?= flash_test$(EXEEXT)
```
jeongchanKim, line 103: I think you missed to change `XYZ`.

apps/system/flash_test/Kconfig
```diff
@@ -3,9 +3,9 @@
 # see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
 #
 
-config SYSTEM_FLASH_ERASEALL
-	bool "FLASH eraseall tool"
+config SYSTEM_FLASH_TEST
```
sunghan-chang, line 6: There is no test stub. It only has read operation. Let's rename it.
 `SYSTEM_FLASH_READ`

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,105 @@
...
 {
-	/* Argument given? */
+	off_t offset;
+	int read_size;
 
-	if (argc < 2) {
-		fprintf(stderr, "usage: flash_eraseall flash_block_device\n");
-		return -1;
-	}
+	if (argc == 4 && !strncmp(argv[1], "read", strlen("read") + 1)) {
```
sunghan-chang, line 145: "read" is constant. Do you need to use strlen even it's not a variable?

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,105 @@
...
+	}
+
+	offset = 0;
+	remaining = size;
 
+	printf("Read 0x%x: ", addr);
+
+	while (remaining > 0) {
+		read_size = BUF_SIZE < remaining ? BUF_SIZE : remaining;
+		ret = MTD_READ(dev_mtd, addr + offset, read_size, buffer);
```
sunghan-chang, line 109: This can't be used for loadable. Is there no way to do on both?

apps/system/flash_test/flash_test.c
jeongchanKim, line 145: If you support only read option, is checking argv[1] needed?
Or do you consider expansion?

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,105 @@
...
+ * Private Functions
  ****************************************************************************/
+static int flash_read(unsigned long addr, int size)
+{
+	int ret;
+	int offset;
+	int buf_idx;
+	int read_size;
+	int remaining;
+	FAR uint8_t *buffer;
```
jeongchanKim, line 82: For using uint8_t, please include `stdint.h`.

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,105 @@
...
+	FAR struct mtd_dev_s *dev_mtd = NULL;
+
+	if (addr == 0 || size <= 0) {
+		printf("Invalid arguments, offset 0x%x, read size %d\n", addr, size);
+		return ERROR;
+	}
+
+	dev_mtd = up_flashinitialize();
```
jeongchanKim, line 90: `up_flashinitialize` is also only for flat build. Could you add conditions about this?

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,105 @@
...
+ ****************************************************************************/
 #ifdef CONFIG_BUILD_KERNEL
 int main(int argc, FAR char *argv[])
 #else
-int flash_eraseall_main(int argc, char *argv[])
+int flash_test_main(int argc, char *argv[])
 #endif
 {
-	/* Argument given? */
+	off_t offset;
```
jeongchanKim, line 142: Please add `sys/types.h` for `off_t`.

apps/system/flash_test/flash_test.c
jeongarmy: I already added that condition in Kconfig.
`depends on MTD && BUILTIN_APPS && !BUILD_PROTECTED && !BUILD_KERNEL`

apps/system/flash_test/flash_test.c
jeongarmy: Yes. I think some tests for flash can be added here.

apps/system/flash_test/flash_test.c
jeongarmy: No.. There is no way to support them now.
For that, all partitions should be registered to some device like ftl to access with standard I/O (open, read, ).

apps/system/flash_test/Kconfig
```diff
@@ -3,9 +3,9 @@
 # see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
 #
 
-config SYSTEM_FLASH_ERASEALL
-	bool "FLASH eraseall tool"
+config SYSTEM_FLASH_READ
+	bool "FLASH read test tool"
```
sunghan-chang, line 7: Let's remove 'test' from the sentence. I don't think flash read is a test.

sunghan-chang: For the folder name, how about `flash_opt` if you have function extension?

apps/system/flash_test/flash_test.c
```diff
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- * Copyright 2016 Samsung Electronics All Rights Reserved.
+ * Copyright 2016, 2021 Samsung Electronics All Rights Reserved.
```
sunghan-chang, line 3: You didn't keep original functionality so that I think we make a new function, not reorganize.
I think we can use 2021 only and can remove nuttx license from the c file.

apps/system/flash_test/flash_test.c
```diff
@@ -53,44 +53,108 @@
...
+		offset = strtoul(argv[2], NULL, 16);
+		read_size = (int)atoi(argv[3]);
 
-	/* Do the job */
+		/* Read an address in flash by size */
+		return flash_read(offset, read_size);
+	}
 
-	flash_eraseall(argv[1]);
+	printf("Usage: flash_test read <offset> <size>\n");
```
jeongchanKim, line 156: Please change `read` to `READ_CMD`.

sunghan-chang (approved):

sunghan-chang:
Let's modify other comments later.

apps/system/flash_test/flash_test.c
jeongarmy: I added `free` to avoid memory leak. Thank you

PR #4874, opened 2021-03-26
Sdaiot 98 fix scan type missing issue

gexuyan:
Fix the issue: Cannot type the password in Easy setup due to the incorrect AP scan result.
Add the missing security types in scan result handler.


an4967 (approved):
It works well

jsdosa (approved):
Looks nice to me.

PR #4880, opened 2021-03-30
os/board/rtl8721csm: add wpa tkip aes mixed security mode in wifi sca…

gexuyan:
…n result

Add the wpa tkip aes mixed mode in wifi scan result


gexuyan:
rename the commit message

PR #4883, opened 2021-04-01
net/lwip/sockets.c : Add Null checking for get_socket

jeongchanKim:
WID:10916217 Return value of a function 'get_socket' is dereferenced at sockets.c:585 without checking, but it is usually checked for this function (23/24).

### DIFF
--- a/os/net/lwip/src/api/sockets.c
+++ b/os/net/lwip/src/api/sockets.c
@@ -492,6 +492,7 @@ int alloc_socket(struct netconn *newconn, int accepted)
 		}
 	}
 	SYS_ARCH_UNPROTECT(lev);
+	kmm_free(sock);
 	return -1;
 }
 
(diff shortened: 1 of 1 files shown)


os/net/lwip/src/api/sockets.c
```diff
@@ -575,6 +575,12 @@ int lwip_accept(int s, struct sockaddr *addr, socklen_t *addrlen)
 	LWIP_ASSERT("invalid socket index", (newsock >= LWIP_SOCKET_OFFSET) && (newsock < NUM_SOCKETS + LWIP_SOCKET_OFFSET));
 	LWIP_ASSERT("newconn->callback == event_callback", newconn->callback == event_callback);
 	nsock = get_socket(newsock, getpid());
+	if (nsock == NULL) {
+		netconn_delete(newconn);
+		free_socket(nsock, 1);
```
an4967, line 580: As 'nsock' is already NULL, "free_socket(nsock, 1)" would occurs an error.
In 'free_socket', It is using 'nsock' without checking NULL.

jeongchanKim: Thanks. I made mistake. I wanted to free `newsock`, so fixed.

an4967 (approved):
LGTM

pillip8282 (approved):
lgtm

PR #4884, opened 2021-04-02
libc/lib_strlen : Add NULL check for string

jeongchanKim:
If passed param 's' is NULL, it cannot be calculated.
So add null checking.

### DIFF
--- a/lib/libc/string/lib_strlen.c
+++ b/lib/libc/string/lib_strlen.c
@@ -66,6 +66,9 @@
 size_t strlen(const char *s)
 {
 	const char *sc;
+	if (s == NULL) {
+		return 0;
+	}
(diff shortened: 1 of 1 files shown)


an4967 (approved):
LGTM

lib/libc/string/lib_strlen.c
```diff
@@ -66,6 +66,7 @@
 size_t strlen(const char *s)
 {
 	const char *sc;
+	if (s == NULL) return 0;
```
sunghan-chang, line 69: Let's follow our coding rules.
```c
if (xxx) {
    yyy;
}
```

an4967 (approved):
LGTM

sunghan-chang (approved):

PR #4885, opened 2021-04-05
apps/testcase/tc_rtc.c : Remove unnecessary rtc unlink

jeongchanKim:
1) After running rtc tc, it unlinks the rtc driver.
In this case, we cannot use rtc anymore.
So remove unlink when finishing rtc tc.
2) The 3rd param of ioctl is unsigned long type, so add typecasting for build warning.


thapav (approved):
LGTM

sunghan-chang (approved):

PR #4891, opened 2021-04-08
Add support for Wifi PF library in common binary

kishore-sn:
Add support for Wifi PF library in common binary and fix related issues.


loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,31 @@
+/****************************************************************************
+ *
+ * Copyright 2020 Samsung Electronics All Rights Reserved.
```
jeongchanKim, line 3: Please change to 2021.

loadable_apps/loadable_sample/wifiapp/wifi_pf.c
```diff
@@ -0,0 +1,31 @@
...
+ ****************************************************************************/
+#include <stdio.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+void wifi_pf_test(void)
+{
+	printf("Wifi platform init Success!\n");
+	return;
```
jeongchanKim, line 30: This `return` is not needed.

loadable_apps/loadable_sample/wifiapp/wifiapp.c
```diff
@@ -70,6 +74,10 @@ int wifiapp_main(int argc, char **argv)
 
 	printf("This is WIFI App\n");
 
+#ifdef CONFIG_WIFI_PLATFORM_LIB
+	wifi_pf_test();
```
jeongchanKim, line 78: How about moving this before line 75?
Because normally platform code initialization is called before application code.

framework/src/wifi_manager/wifi_manager_msghandler.c
```diff
@@ -37,9 +37,11 @@ static handler_queue g_wifi_message_queue;
  * External functions
  */
 extern wifi_manager_result_e wifimgr_handle_request(wifimgr_msg_s *msg);
+#ifdef CONFIG_BUILD_FLAT
 #ifdef CONFIG_VIRTUAL_WLAN
 extern void vwifi_start(void);
 #endif
+#endif
```
pillip8282, line 45: please move vwifi_start() call to apps/system/init/init.c
below 

#if defined(CONFIG_WIFI_MANAGER)
	(void)wifimgr_run_msghandler();
#endif
#ifdef CONFIG_VIRTUAL_WLAN
 	vwifi_start();
 #endif

loadable_apps/loadable_sample/wifiapp/wifiapp.c
kishore-sn: Done.

framework/src/wifi_manager/wifi_manager_msghandler.c
kishore-sn: Done. 
But, this will still result in error during protected build. So, I have added BUILD_FLAT check for it. 
Please try to find a different method to call kernel api from app side, either with syscall or ioctl.

jeongchanKim (approved):
LGTM

PR #4892, opened 2021-04-09
kernel/clock_initialize : Add initialization of init_time

jeongchanKim:
tm_isdst in init_time variable is used from mktime when CONFIG_LIBC_LOCALTIME is enabled.
Without initialization, it can have trash variable, so it can make wrong calculation in mktime.


davidfather (approved):
Initializing local variables before using is recommended.

sunghan-chang (approved):

PR #4893, opened 2021-04-09
mm/mm_heap/mm_calloc : Check integer wrap-around

jeongchanKim:
The size which is n * elem_size can be integer wrap-around.
In this case, mm_calloc tries to allocate wrong, small memory and user can access invalid memory.

### DIFF
--- a/os/mm/mm_heap/mm_calloc.c
+++ b/os/mm/mm_heap/mm_calloc.c
@@ -82,13 +82,21 @@ FAR void *mm_calloc(FAR struct mm_heap_s *heap, size_t n, size_t elem_size)
 {
 	FAR void *ret = NULL;
 
-	if (n > 0 && elem_size > 0) {
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-		ret = mm_zalloc(heap, n * elem_size, caller_retaddr);
-#else
-		ret = mm_zalloc(heap, n * elem_size);
-#endif
+	if ((n == 0) || (elem_size == 0)) {
+		mdbg("Invalid Parameter, n : %u, elem_size : %u\n", n, elem_size);
+		return NULL;
 	}
+
+	if (n > (MMSIZE_MAX / elem_size)) {
+		mdbg("Parameter n(%u) should be smaller than (MMSIZE_MAX / elem_size)(%u), \
+			because multiplication of n and elem_size cannot overflow the size_t.\n", n, (MMSIZE_MAX / elem_size));
+		return NULL;
+	}
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	ret = mm_zalloc(heap, n * elem_size, caller_retaddr);
+#else
+	ret = mm_zalloc(heap, n * elem_size);
+#endif
 
 	return ret;
 }


os/mm/mm_heap/mm_calloc.c
```diff
@@ -81,12 +81,25 @@ FAR void *mm_calloc(FAR struct mm_heap_s *heap, size_t n, size_t elem_size)
...
 	FAR void *ret = NULL;
+	size_t size;
 
 	if (n > 0 && elem_size > 0) {
+		if (n > (MMSIZE_MAX / elem_size)) {
+			mdbg("Parameter n(%u) should be smaller than %u, \
+				because multiplication of n and elem_size cannot overflow the size_t.\n", n, elem_size);
+			return NULL;
+		}
+		size = n * elem_size;
```
Taejun-Kwon, line 92: if state in line 86 is not true, size is not used. svace will treat this as minor issue.

os/mm/mm_heap/mm_calloc.c
```diff
@@ -81,12 +81,25 @@ FAR void *mm_calloc(FAR struct mm_heap_s *heap, size_t n, size_t elem_size)
...
+		if (size <= elem_size || size > MM_ALIGN_DOWN(MMSIZE_MAX) - SIZEOF_MM_ALLOCNODE) {
```
Taejun-Kwon, line 93:
```
if (n > 0 && elem_size > 0) {
    ....
    size = n * elem_size;
}
```
so, size never can be smaller than elem_size, but can be same as elem_size if n is 1
if return NULL is correct when n is 1, you'd better modify if state in line 86.

sunghan-chang: `n` can be 1. It's not a problem. You should remove `=`. But by line 87, size can't be smaller as Taejun said.
This line is unnecessary.  
`size > MM_ALIGN....` is duplicated, it's in zalloc function.

os/mm/mm_heap/mm_calloc.c
```diff
@@ -83,6 +83,11 @@ FAR void *mm_calloc(FAR struct mm_heap_s *heap, size_t n, size_t elem_size)
 	FAR void *ret = NULL;
 
 	if (n > 0 && elem_size > 0) {
+		if (n > (MMSIZE_MAX / elem_size)) {
```
Taejun-Kwon, line 86: How about below style?
```
if ((n == 0) || (elem_size ==0)) {
        return NULL;
}

if (n > (MMSIZE_MAX / elem_size) {
        mdbg();
        return NULL;
}
	ret = mm_zalloc(heap, n * elem_size, caller_retaddr);
#else
	ret = mm_zalloc(heap, n * elem_size);
#endif

return ret;
```

os/mm/mm_heap/mm_calloc.c
```diff
@@ -82,13 +82,21 @@ FAR void *mm_calloc(FAR struct mm_heap_s *heap, size_t n, size_t elem_size)
...
 
-	if (n > 0 && elem_size > 0) {
+	if ((n == 0) || (elem_size == 0)) {
+		mdbg("Invalid Parameter, n : %u, elem_size : %u\n", n, elem_size);
+		return NULL;
+	}
+
+	if (n > (MMSIZE_MAX / elem_size)) {
+		mdbg("Parameter n(%u) should be smaller than %u, \
+			because multiplication of n and elem_size cannot overflow the size_t.\n", n, elem_size);
```
sunghan-chang, line 92: Maybe second `%u` is `(MMSIZE_MAX / elem_size)`, not `elem_size`. Could you check it?
Because `n` could be bigger than `elem_size`, it' does not cause a problem.

sunghan-chang (approved):

PR #4900, opened 2021-04-14
app/system/sekey_test: implement secure key tester

an4967:
- ECDSA Key sign & verify test
- Certificate parsing test

### DIFF
--- a/apps/system/sekey_test/Kconfig
+++ b/apps/system/sekey_test/Kconfig
@@ -1,0 +1,27 @@
+#
+# For a description of the syntax of this configuration file,
+# see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
+#
+
+config SYSTEM_SEKEY_TEST
+	bool "Secure Key Tester"
+	default n
+	depends on SECURITY_LINK
+	---help---
+		Enable the SE Key & Cert Test
+
+if SYSTEM_SEKEY_TEST
+
+config ECDSA_KEY_SLOT
+	string "ECDSA Key slot list"
+	default "0"
+	---help---
+		Comma separated list of ecdsa keys.
+
+config CERT_SLOT
+	string "Certificate slot list"
+	default "1,2,3"
+	---help---
+		Comma separated list of certificates (PEM, DER).
+
+endif
--- a/apps/system/sekey_test/sekey_test.c
+++ b/apps/system/sekey_test/sekey_test.c
@@ -1,0 +1,243 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * apps/system/sekey_test/sekey_test.c
+ *
+ *   Copyright (C) 2008, 2011-2012 Gregory Nutt. All rights reserved.
+ *   Author: Gregory Nutt <gnutt@nuttx.org>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name NuttX nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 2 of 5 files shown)


apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,239 @@
...
+	ptr = strtok(cert_temp, ",");
+	while (ptr != NULL) {
+		cert_test[cert_cnt++] = atoi(ptr);
+		ptr = strtok(NULL, ",");
+	}
+
+	t_hash.data = malloc(hash_len);
+	if (t_hash.data == NULL) {
+		res = -1;
+		goto exit;
```
jeongchanKim, line 138: How about printing the log about failure?

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,239 @@
...
+
+	for (i = 0; i < cert_cnt; i++) {
+		memset(buf, 0, sizeof(buf));
+		memset(cert_out.data, 0, cert_out_size);
+		cert_out.data_len = cert_out_size;
+
+		printf("[Cert] Get Cert Slot#%d ... ", cert_test[i]);
+		ret = sl_get_certificate(shnd, cert_test[i], &cert_out, &hres);
+		if (ret != SECLINK_OK || hres != HAL_SUCCESS) {
+			printf("[ FAIL ] / ret : %d, hres : %d \n", ret, hres);
```
jeongchanKim, line 190: We cannot distinguish the failure between line 190 and 201.
How about remaining the reason in detail?

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,239 @@
...
+#ifdef CONFIG_ECDSA_KEY_SLOT
+#define ECDSA_KEY_SLOT CONFIG_ECDSA_KEY_SLOT
+#else
+#define ECDSA_KEY_SLOT "0,"
+#endif
+
+#ifdef CONFIG_CERT_SLOT
+#define CERT_SLOT CONFIG_CERT_SLOT
+#else
+#define CERT_SLOT "1,"
```
jeongchanKim, line 73: The default of `CONFIG_CERT_SLOT` is "1,2,3," in Kconfig.
How about matching the default?

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,239 @@
...
+	int ret;
+	int res = 0;
+	char *ptr;
+	sl_ctx shnd;
+	hal_result_e hres = 0;
+	hal_ecdsa_mode ecdsa_mode;
+	hal_data t_hash = {
+		0,
+	};
+	int hash_len = 32;
```
jeongchanKim, line 95: If `hash_len` is not changed in the code, how about using `#define HASH_LEN 32`?

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,239 @@
...
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <stdlib.h>
```
jeongchanKim, line 59: Please include `string.h` also for `strtok`.

apps/system/sekey_test/sekey_test.c
an4967: In this case, there are several types of errors so we cannot explain the reason in one sentence.
The only thing that users can know is which command is not working in the board.
For more details, the user should check 'hres' value at seclink.

apps/system/sekey_test/sekey_test.c
an4967: Done

apps/system/sekey_test/sekey_test.c
an4967: Done

apps/system/sekey_test/sekey_test.c
an4967: Done

apps/system/sekey_test/sekey_test.c
an4967: Done

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+	sl_ctx shnd;
+	hal_result_e hres = 0;
+	hal_ecdsa_mode ecdsa_mode;
+	hal_data t_hash = {
+		0,
+	};
+	hal_data sign = {
+		0,
+	};
+	int sign_len = 1024;
```
jeongchanKim, line 101: Please change `sign_len` to `#define` like HASH_LEN.

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+
+	mbedtls_x509_crt cert = {
+		0,
+	};
+	hal_data cert_out = {
+		0,
+	};
+	int cert_out_size = 1024;
```
jeongchanKim, line 109: Same as above.

jsdosa (dismissed):
looks nice to me.

jeongchanKim (approved):
LGTM

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+#define ECDSA_KEY_SLOT "0"
+#endif
+
+#ifdef CONFIG_CERT_SLOT
+#define CERT_SLOT CONFIG_CERT_SLOT
+#else
+#define CERT_SLOT "1,2,3"
+#endif
+
+#define HASH_LEN 		32
```
sunghan-chang, line 77: You should use spaces between definition and value.
```
#define HASH_LEN (spaces) 32
```
same as line 78

sunghan-chang (commented):
There is no usages sentence. Could you add `--help` option to show how to use?

jeongchanKim (approved):
LGTM

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * apps/system/sekey_test/sekey_test.c
+ *
+ *   Copyright (C) 2008, 2011-2012 Gregory Nutt. All rights reserved.
```
sunghan-chang, line 21: We don't need to add Nuttx License. Let's remove.

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+
+int sekey_test_main(int argc, char *argv[])
+{
+	int i;
+	int ret;
+	int res = 0;
+	char *ptr;
+	sl_ctx shnd;
+	hal_result_e hres = 0;
+	hal_ecdsa_mode ecdsa_mode;
```
sunghan-chang, line 96: 1. Is this checking ecdsa only? If it is, how about renaming it as ecdsa key test?
2. We have several keys so that we need to test them all. Is it possible to add others as well?

pillip8282 (commented):
All security related test are implemented apps/example/security_test/. How about move this to there?

apps/system/sekey_test/sekey_test.c
```diff
@@ -0,0 +1,243 @@
...
+
+	ret = sl_init(&shnd);
+	if (ret != SECLINK_OK) {
+		printf("[SECLINK] sl_init fail \n");
+		return -1;
+	}
+
+	ptr = strtok(key_temp, ",");
+	while (ptr != NULL) {
+		key_test[key_cnt++] = atoi(ptr);
```
jeongarmy, line 127: I think you need to check `key_cnt` and `cert_cnt` values with 128(buffer size).
That's why it can make an overflow here if user set `CONFIG_CERT_SLOT` or `CONFIG_ECDSA_KEY_SLOT` as a long list with larger than buffer size.
Please add checking them with a range of buffer.
And you'd better to define values like `#define KEY_CNT 128` and `#define CERT_CNT 128` and use them when checking the range for readability and security.

apps/system/sekey_test/Kconfig
```diff
@@ -0,0 +1,27 @@
+#
+# For a description of the syntax of this configuration file,
+# see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
+#
+
+config SYSTEM_SEKEY_TEST
+	bool "Secure Key Tester"
+	default n
+	depends on SECURITY_LINK
```
jeongarmy, line 9: I think you need to add a dependency of mbedtls, CONFIG_NET_SECURITY_TLS.
If not, it brings build error because  this example uses mbedtls APIs.
I think `select CONFIG_NET_SECURITY_TLS` is better because user can enable this config only if user want to run this examples.
Please check this and add `depends on ` or `select` as your opinion here.

apps/system/sekey_test/sekey_test.c
an4967: @jeongarmy 
Good point.  I will fix my code

apps/system/sekey_test/Kconfig
an4967: @jeongarmy 
Ok. I will apply 'select' with mbedtls

PR #4905, opened 2021-04-16
testcase/network: Add new poll testcases

deepaksrma:


apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+
+	args.fd = sd;
+	args.fds = &fds;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_GEQ("lwip_poll", ret, 0);
+	TC_SUCCESS_RESULT();
+
+	args.setup = 0;
```
jeongchanKim, line 79: Is this needed? Because `args` is local variable.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	args.fd = sd;
+	args.fds = &fds;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_GEQ("lwip_poll", ret, 0);
```
jeongchanKim, line 76: I think we'd better use TC_ASSERT_GEQ_CLEANUP for closing `sd`.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	struct poll_test_args args;
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
+
+	args.fd = sd;
+	args.fds = NULL;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_LEQ("lwip_poll", ret, -1);
```
jeongchanKim, line 108: Please close `sd` when it fails.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	args.fds = &fds;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_GEQ("lwip_poll", ret, 0);
+	TC_SUCCESS_RESULT();
+
+	args.setup = 0;
+
+	tc_net_poll_teardown_p(fd, &args);
```
jeongchanKim, line 81: Is it okay not to call from failure case?

os/net/lwip/src/api/sockets.c
```diff
@@ -444,6 +444,9 @@ struct lwip_sock *trycheck_selwait_socket(int s, pid_t pid)
 {
 	struct lwip_sock *sock;
 	sock = tryget_socket(s, pid);
+	if(sock == NULL) {
```
jeongchanKim, line 447: Please add space after `if`.

os/drivers/os_api_test/network/test_net_poll.c
```diff
@@ -0,0 +1,54 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+static int test_poll(unsigned long arg)
+{
+	int ret = -1;
+	poll_test_args_t *args = (poll_test_args_t *)arg;
+
+	ret = lwip_poll(args->fd, args->fds, args->setup);
```
jeongchanKim, line 40: `args` can be dereferencing. Please check it before using.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	args.fds = 0;
+
+	args.fd = sd;
+	args.fds = &fds;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_GEQ("lwip_poll", ret, 0);
+	TC_SUCCESS_RESULT();
+	close(sd);
```
jeongchanKim, line 141: Same as above.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: yes, because if setup is failed then we can't call to teardown.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: yes. because setup =1 is for poll_setup and 0 is for teardown.

apps/examples/testcase/le_tc/network/tc_net_poll.c
jeongchanKim: Oh okay. I got it.

apps/examples/testcase/le_tc/network/tc_net_poll.c
jeongchanKim: Ok. Thanks for checking.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	struct pollfd fds[1];
+	struct poll_test_args args;
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
+
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	args.fd = sd;
+	args.fds = &fds;
```
sangam-swami, line 72: can you type case here to remove the warning messages

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	struct pollfd fds[1];
+	struct poll_test_args args;
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
+
+	fds[0].fd = sd;
+	args.fds = 0;
+
+	args.fd = sd;
+	args.fds = &fds;
```
sangam-swami, line 135: can you type case here to remove the warning messages

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+	int ret = -1;
+	int sd = -1;
+	struct pollfd fds[1];
+	struct poll_test_args args;
+
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	args.fd = -1;
+	args.fds = &fds;
```
sangam-swami, line 164: can you type case here to remove the warning messages

sangam-swami:
you can add description for PR when you raise.
"No description provided."

os/net/lwip/src/api/sockets.c
```diff
@@ -444,6 +444,9 @@ struct lwip_sock *trycheck_selwait_socket(int s, pid_t pid)
 {
 	struct lwip_sock *sock;
 	sock = tryget_socket(s, pid);
+	if (sock == NULL) {
```
sangam-swami, line 447: is this scenario created or tested?

deepaksrma: without checking this, few new TCs may stuck here.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: corrected now.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	args.fd = sd;
+	args.fds = fds;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_GEQ("lwip_poll", ret, 0);
```
jeongchanKim, line 76: Please check the close of fd.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,183 @@
...
+
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	args.fd = -1;
+	args.fds = fds;
+	args.setup = 0;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_LEQ("lwip_poll", ret, -1);
```
jeongchanKim, line 168: Is it okay not to close the fd?

os/net/lwip/src/include/lwip/sockets.h
```diff
@@ -475,6 +476,13 @@ typedef struct ip_mreq {
 #define SELWAIT_T u8_t
 #endif
 
+/* use in poll testcases */
+struct poll_test_args {
```
jeongchanKim, line 480: How about moving this structure definition into `os_api_test_drv.h`?
Because it is only used for testcase.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: we have not created any socket here, so no need to close.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: done

os/net/lwip/src/include/lwip/sockets.h
```diff
@@ -58,6 +58,7 @@
 #include <string.h>				/* for memset used by FD_ZERO */
 #include <fcntl.h>
 #include <errno.h>
+#include <poll.h>
```
pillip8282, line 61: lwip doesn't need to include poll.h

pillip8282 (commented):
Calling setup and teardown must be happened inside poll function. So I don't understand why did you distinguish poll(setup), poll(teardown). Could you describe the purpose of this test more detail?

deepaksrma:
> Calling setup and teardown must be happened inside poll function. So I don't understand why did you distinguish poll(setup), poll(teardown). Could you describe the purpose of this test more detail?

because if a problem occurred then it should be distinguishable whether the problem occurred in setup or teardown. but if you think that both should be inside the poll, then I'll change this.

pillip8282:
I 

> > Calling setup and teardown must be happened inside poll function. So I don't understand why did you distinguish poll(setup), poll(teardown). Could you describe the purpose of this test more detail?
> 
> because if a problem occurred then it should be distinguishable whether the problem occurred in setup or teardown. but if you think that both should be inside the poll, then I'll change this.

I think that testing poll() in multi-thread environment would be better. Because many applications use poll in multi-thread env, and many issues are reported. For example thread1 call poll while thread2 close the socket which is used for poll().

deepaksrma:
> I
> 
> > > Calling setup and teardown must be happened inside poll function. So I don't understand why did you distinguish poll(setup), poll(teardown). Could you describe the purpose of this test more detail?
> > 
> > 
> > because if a problem occurred then it should be distinguishable whether the problem occurred in setup or teardown. but if you think that both should be inside the poll, then I'll change this.
> 
> I think that testing poll() in multi-thread environment would be better. Because many applications use poll in multi-thread env, and many issues are reported. For example thread1 call poll while thread2 close the socket which is used for poll().

done. please check now.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,244 @@
...
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sd < 0) {
+		printf("socket fail %s:%d", __FUNCTION__, __LINE__);
+		return NULL;
+	}
+
+	args.fd = fd;
+	args.sd = sd;
+	pthread_create(&thread1, NULL, wait_event, (void *)&args);
```
jeongarmy, line 92: I think you need to check result of pthread_create.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,244 @@
...
+
+	close(sd);
```
jeongarmy, line 94: You passed sd to wait_event thread via args and wait_event uses sd.
But sd is closed here. I think it doesn't work well. Is it right?

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,244 @@
...
+*/
+
+static void tc_net_poll_p(int fd)
+{
+	int ret = -1;
+	int sd = -1;
+	struct pollfd fds[1];
+	struct poll_test_args args;
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
```
jeongarmy, line 115: Need to check a sd value with TC_ASSERT_GEQ.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,244 @@
...
+		printf("socket fail %s:%d", __FUNCTION__, __LINE__);
+		return;
+	}
+
+	args.fd = sd;
+	args.fds = NULL;
+	args.setup = 1;
+
+	ret = ioctl(fd, TESTIOC_NET_POLL, &args);
+	TC_ASSERT_LEQ_CLEANUP("lwip_poll", ret, -1, close(sd));
```
jeongarmy, line 161: TC_ASSERT_LEQ_CLEAN_UP is right? I think TC_ASSERT_GEQ_CLEANUP("lwip_poll", ret, 0, close(sd)); is right.

jeongarmy: Oh, I missed negative test. I think TC_ASSERT_EQ_CLEANUP("lwip_poll", ret, -1, close(sd)); is right.
That's because ioctl should return -1 in this case.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: yes, it is correct because we are testing this thing, means when one thread try for poll and meanwhile other thread close the socket.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: could you clear your point in detail?
how can we check sd value with TC_ASSERT_GEQ?

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: i don't think, there is any need to check result of pthread_create because if thread is not created so function will not be called, then socket will be closed and that TC will not execute.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,244 @@
...
+
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <poll.h>
+#include <pthread.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netutils/netlib.h>
+#include <lwip/sockets.h>
```
pillip8282, line 31: it's prohibited to include lwip headers in application layer.

os/drivers/os_api_test/os_api_test_drv.c
```diff
@@ -191,6 +191,14 @@ static int os_api_test_drv_ioctl(FAR struct file *filep, int cmd, unsigned long
 		ret = test_net_pbuf(cmd, arg);
 		break;
 #endif
+
+#ifdef CONFIG_TC_NET_POLL
+	/* Run the test case for poll   */
+	case TESTIOC_NET_POLL:
+	ret = test_net_poll(cmd, arg);
```
pillip8282, line 198: please check indentation

pillip8282 (commented):
You don't need to test lwip_poll directly. You can test it if you open socket descriptors and pass it to select(). Instead of testing lwip_poll directly, you'd better to test it API which provides to application layer(e.g select,() socket())

deepaksrma:
> You don't need to test lwip_poll directly. You can test it if you open socket descriptors and pass it to select(). Instead of testing lwip_poll directly, you'd better to test it API which provides to application layer(e.g select(), socket(), close())

Done, please check now.

abhishek-samsung (approved):
LGTM

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,182 @@
...
+	fds[0].events = POLLIN;
+
+	ret = poll(fds, 1, 1000);
+	TC_ASSERT_LEQ("lwip_poll", ret, -1);
+	TC_SUCCESS_RESULT();
+}
+
+int net_poll_main(void)
+{
+#ifndef CONFIG_DISABLE_POLL
```
sunghan-chang, line 170: You should make a dependency with tc config instead of this conditional.
```
config TC_NET_POLL
	bool "lwip_poll() api"
	default n
	depends on !DISABLE_POLL
```

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,182 @@
...
+	fds[0].fd = sd;
+	fds[0].events = POLLIN;
+
+	ret = poll(fds, 1, 1000);
+	tc_net_multi_thread_poll_p(ret, sd);
+
+	return NULL;
+}
+
+void *multi_thread_poll(void *arg)
```
sunghan-chang, line 64: Let's add "static".

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,182 @@
...
+* @postcondition        :
+*/
+
+static void tc_net_multi_thread_poll_p(int ret, int sd)
+{
+	TC_ASSERT_LEQ_CLEANUP("lwip_poll", ret, 0, close(sd));
+	TC_SUCCESS_RESULT();
+}
+
+void *wait_event(void *arg)
```
sunghan-chang, line 49: same as previous, static

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,182 @@
...
+	int sd = -1;
+	pthread_t thread1;
+
+	sd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sd < 0) {
+		printf("socket fail %s:%d", __FUNCTION__, __LINE__);
+		return NULL;
+	}
+
+	pthread_create(&thread1, NULL, wait_event, (void *)&sd);
```
sunghan-chang, line 75: There is no pthread_detach or pthread_join. It's mandatory to prevent memory leak after pthread_create.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,182 @@
...
+
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <poll.h>
+#include <pthread.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netutils/netlib.h>
+#include <tinyara/os_api_test_drv.h>
```
sunghan-chang, line 31: looks unnecessary.

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: done

apps/examples/testcase/le_tc/network/tc_net_poll.c
deepaksrma: done. please check.

apps/examples/testcase/le_tc/network/tc_net_poll.c
```diff
@@ -0,0 +1,178 @@
...
+	TC_ASSERT_GEQ("socket", sd, 0);
+
+	fds[0].fd = sd;
+	fds[0].events = 0;
+
+	ret = poll(fds, 1, 1000);
+	TC_ASSERT_GEQ_CLEANUP("lwip_poll", ret, 0, close(sd));
+
+	close(sd);
+	TC_ASSERT_GEQ("lwip_poll_tc_close", ret, 0);
```
jeongchanKim, line 138: You didn't set the `ret` again after checking in line 135.
Why do you check the `ret` here again?
You mean to return from `close(sd)`?

deepaksrma: corrected

kripal-samsung (commented):
LGTM

PR #4907, opened 2021-04-19
Fix max MPU region issue and modify CONFIG_NAME_MAX

abhishek-samsung:
Fix max MPU region issue. Modiy CONFIG_NAME_MAX value to 32.


kishore-sn (approved):
LGTM

sangam-swami (dismissed):
looks good

thapav (dismissed):

jeongchanKim (approved):
LGTM

apps/system/init/init.c
```diff
@@ -69,6 +69,11 @@
 extern void iotjs_register_cmds(void);
 #endif
 
+#ifdef CONFIG_BUILD_FLAT
```
sunghan-chang, line 72:
```

The vwifi_start symbol is defined in kernel. So, when loadable_apps config
is used, this symbol cannot be accessed directly from app binary, and needs
to be accessed via syscall or ioctl. As a temporary fix, we check for
CONFIG_BUILD_FLAT in the framework code to avoid linker error.
```
There is no reason why you moved this from framework to preapp. Could you let me know and write it on commit description?

sunghan-chang (commented):
First commit title is `common/....` but I don't know which common is. There are several `common` folders. Let's add `arch` or `arm`.
```
arch/common/...
or
arm/common/...
````

apps/system/init/init.c
sunghan-chang: Pilseob made PR #4908 for this. Let's remove this commit from this PR.

abhishek-samsung: Okay. Done.

sunghan-chang (approved):

PR #4908, opened 2021-04-19
network/wi-fi: move vwifi initialization

pillip8282:
vwifi was initialized a user space. So it caused build fail if
loadable configuration was set. Therefore change the location of
initialization of vwifi device.

### DIFF
--- a/os/drivers/wireless/virtual/vdev.c
+++ b/os/drivers/wireless/virtual/vdev.c
@@ -54,6 +54,9 @@ static trwifi_result_e vdev_start_sta(struct netdev *dev);
 static trwifi_result_e vdev_stop_softap(struct netdev *dev);
 static trwifi_result_e vdev_set_autoconnect(struct netdev *dev, uint8_t check);
 
+static int vdev_linkoutput(struct netdev *dev, void *buf, uint16_t dlen);
+static int vdev_set_multicast_list(struct netdev *dev, const struct in_addr *group, netdev_mac_filter_action action);
+
 /****************************************************************************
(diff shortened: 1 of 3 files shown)


os/drivers/wireless/virtual/vdev.c
```diff
@@ -213,13 +216,109 @@ int _vwifi_create_msgqueue(int *fd)
...
+		}
+
+		res = vwifi_handle_message(msg.req);
+		if (res < 0) {
+			VWIFI_ERROR(res);
+			break;
+		}
+		_send_signal(&msg);
+	}
+	return;
```
jeongchanKim, line 295: I think this return is not needed.

pillip8282: I think this is trivial. so I won't remove it

os/drivers/wireless/virtual/vdev.c
```diff
@@ -213,13 +216,109 @@ int _vwifi_create_msgqueue(int *fd)
 	return 0;
 }
 
+/*
+ * Internal Function
+ */
+static struct netdev* vdev_register_dev(int sizeof_priv)
```
davidfather, line 222: When is the input param 'sizeof_priv' used? If it is useless, please remove it.

jeongchanKim (dismissed):
LGTM

jeongchanKim:
Please check the coding rule.

jeongchanKim (approved):
LGTM

an4967 (approved):
LGTM

PR #4909, opened 2021-04-21
apps/system/utils : Remove not available API and Modify its utils

jeongchanKim:


apps/system/utils/utils_kill.c
```diff
@@ -235,24 +238,62 @@ int utils_kill(int argc, char **args)
...
+
+	asprintf(&filepath, "%s/%s/%s", PROCFS_MOUNT_POINT, entryp->d_name, "stat");
+	ret = utils_readfile(filepath, buf, KILLALL_BUFLEN, arg, killall_send_signal);
+	if (ret < 0) {
+		printf("Failed to read %s\n", filepath);
+		free(filepath);
+		return ERROR;
+	}
+
+	free(filepath);
```
amogh-samsung, line 282: Extra empty lines needed both before and after this one?

apps/system/utils/utils_kill.c
```diff
@@ -235,24 +238,62 @@ int utils_kill(int argc, char **args)
...
+
+	return OK;
```
amogh-samsung, line 284: Maybe you could unify the control flow of this function?

```
{
    int ret = OK;
    ret = utils_readfile();;
    if (ret < 0) {
        printf();
        ret = ERROR;
    }
    
    free(filepath);
    return ret;
}
```

apps/system/utils/utils_kill.c
jeongchanKim: Removed. Thanks.

apps/system/utils/utils_kill.c
```diff
@@ -58,11 +58,14 @@
 #include <signal.h>
 #include <sched.h>
 #include <pthread.h>
+#include "utils_proc.h"
 
 #if defined(CONFIG_ENABLE_KILLALL) && (CONFIG_TASK_NAME_SIZE <= 0)
 #error If you want to use killall, CONFIG_TASK_NAME_SIZE should be > 0
 #endif
 
+#define KILLALL_BUFLEN 128
```
sangam-swami, line 67: can be put under config_enable_killall to avoid warnings?

apps/system/utils/Kconfig
```diff
@@ -167,7 +167,7 @@ config ENABLE_KILL
 config ENABLE_KILLALL
 	bool "killall"
 	default y
-	depends on !BUILD_PROTECTED
+	depends on FS_PROCFS && !FS_PROCFS_EXCLUDE_PROCESS
```
jeongarmy, line 170: Why does it have a dependency on procfs? The killall sends signal to all tasks only. It doesn't use procfs. please check.

apps/system/utils/utils_proc.c
```diff
@@ -93,7 +93,7 @@ int utils_proc_pid_foreach(procentry_handler_t handler)
 	return ret;
 }
 
-int utils_readfile(FAR const char *filepath, char *buf, int buflen, utils_print_t print_func)
+int utils_readfile(FAR const char *filepath, char *buf, int buflen, void *arg, utils_handler_t handler)
```
jeongarmy, line 96: How about changing a sequence of handler and args for better understanding?
That's because args is argument passed to handler, not used in this function.

sunghan-chang (approved):

PR #4910, opened 2021-04-21
apps/shell/tash_security : Modify wrong comparison of write size

jeongchanKim:
If it writes passwd_prompt properly, it can be equal to sizeof(passwd_prompt).
So the return of write is equal to sizeof(passwd_prompt) for successful case.


sangam-swami (approved):
looks fine

sunghan-chang:
@jeongchanKim It's not `unnecessary`, it's `wrong`. Could you update commit title?

sunghan-chang (approved):

PR #4914, opened 2021-04-22
shell/Kconfig: add description in TASH_PASSWORD_SHA256

sunghan-chang:
The password which is given in the config is hashed by SHA256
so that it should be 32 bytes. This commit describes it in the
help tab.

### DIFF
--- a/apps/shell/Kconfig
+++ b/apps/shell/Kconfig
@@ -58,17 +58,17 @@ config TASH_SCRIPT
 		See "sh" on TASH help command.
 
 config SECURED_TASH
-    bool "Enable Secure shell"
-    depends on NET_SECURITY_TLS
-    default n
+	bool "Enable Secure shell"
+	depends on NET_SECURITY_TLS
+	default n
 
 if SECURED_TASH
 
 config TASH_PASSWORD_SHA256
-    string "SHA256 encrypted password"
-    default "0000"
-    ---help---
-        Enable password to enter TASH.
+	string "SHA256 encrypted password"
+	default "0000"
+	---help---
+		Enable password to enter TASH.
 endif
 
 endif
--- a/apps/shell/Kconfig
+++ b/apps/shell/Kconfig
@@ -65,10 +65,12 @@ config SECURED_TASH
 if SECURED_TASH
 
 config TASH_PASSWORD_SHA256
-	string "SHA256 encrypted password"
-	default "0000"
+	string "SHA256 encrypted password (32 Bytes)"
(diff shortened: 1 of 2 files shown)


jeongchanKim (approved):
LGTM

apps/shell/Kconfig
```diff
@@ -65,10 +65,11 @@ config SECURED_TASH
 if SECURED_TASH
 
 config TASH_PASSWORD_SHA256
-	string "SHA256 encrypted password"
+	string "SHA256 encrypted password (32 Bytes)"
 	default "0000"
 	---help---
-		Enable password to enter TASH.
+		Enable password to enter TASH. The password is hashed by SHA256 so that
+		the length of this config should be 32 bytes (64 characters).
```
sangam-swami, line 72: it it fixed 32 byte or can be <= 32 bytes?

apps/shell/Kconfig
```diff
@@ -58,17 +58,18 @@ config TASH_SCRIPT
...
+	default n
 
 if SECURED_TASH
 
 config TASH_PASSWORD_SHA256
-    string "SHA256 encrypted password"
-    default "0000"
-    ---help---
-        Enable password to enter TASH.
+	string "SHA256 encrypted password (32 Bytes)"
```
sangam-swami, line 68: the commit title can be "add description for" ?

apps/shell/Kconfig
sunghan-chang: fixed 32 bytes

apps/shell/Kconfig
sunghan-chang: done

apps/shell/Kconfig
```diff
@@ -58,17 +58,18 @@ config TASH_SCRIPT
...
+	default "0000"
```
sangam-swami, line 69: can we set default 32 bytes here

sunghan-chang: I don't know that we need to add default value here. Because it's password, someone who wants to enable this feature can't use default value.
Could you let me know your opinion? Is it better to add for example?

sangam-swami: I feel better don't add any default password, we can give example...
if user not set any password then feature can't be used

sangam-swami (approved):
looks good

jeongarmy (approved):
LGTM

thapav (approved):
LGTM

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4915, opened 2021-04-23
shell: tidyup secured TASH

sunghan-chang:
1. add CONFIG_TASH_PASSWORD_SHA256 length check
  It should be 32 bytes.
2. covert CONFIG_TASH_PASSWORD_SHA256 char to hex once
  It should be converted to SHA256 hex value to compare input password.
  But it's fixed value so that we can convert it one time.
  And this commit removes unnecessary string copy.
  2 times (sprintf, strncpy) -> 1 time (strncpy)
3. use some definitions for readability
  SHA256_HASH_nCHAR : number of SHA256 hash character to verify the config
  SHA256_HASH_nBYTE : size of SHA256 data to set array
  CONVERT_nCHAR : number of character to convert char to hex at once
4. reduce local variables size
  212 bytes -> 98 bytes

### DIFF
--- a/apps/shell/tash_security.c
+++ b/apps/shell/tash_security.c
@@ -18,6 +18,7 @@
 /// @file   tash_security.c
 /// @brief  functions to check secure tash command
 
+#include <tinyara/config.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


apps/shell/tash_security.c
```diff
@@ -49,26 +69,15 @@ void tash_check_security(int fd)
...
-
-		mbedtls_sha256((unsigned char *)line_buff, len, result, 0);
+		mbedtls_sha256((unsigned char *)line_buff, len, input_sha256_hex, 0);
 
 		tash_free(line_buff);
 
-		if (!strncmp((char *)sha256_hex, (char *)result, 32)) {
+		if (!strncmp((char *)pw_sha256_hex, (char *)input_sha256_hex, 32)) {
 			break;
 		} else if (len != 0) {
```
jeongarmy, line 78: It is better to move this `len!=0` to line 71 and change `else` here. How about this?

apps/shell/tash_security.c
```diff
@@ -25,50 +25,68 @@
...
-	unsigned char result[64];
-	unsigned char sha256_hex[64];
 	int len = 0;
-	const char passwd_prompt[] = "PASSWORD>>";
-	int count = 0;
+	char pw_sha256_char[SHA256_HASH_nCHAR + 1] = { 0, };
+	unsigned char pw_sha256_hex[SHA256_HASH_BYTE];
+	unsigned char input_sha256_hex[SHA256_HASH_nBYTE];
+	char *psha256_char;
+	unsigned long covert_value;
```
jeongchanKim, line 41: I think this variable name can be `convert_value`.

apps/shell/tash_security.c
```diff
@@ -25,50 +25,68 @@
...
+	int sha256_hex_idx;
+	int retry_count = 0;
+
+	if (strlen(CONFIG_TASH_PASSWORD_SHA256) != SHA256_HASH_nCHAR) {
```
jeongchanKim, line 45: Please include `tinyara/config.h` for `CONFIG_TASH_PASSWORD_SHA256`.

apps/shell/tash_security.c
```diff
@@ -18,57 +18,76 @@
...
 
-		mbedtls_sha256((unsigned char *)line_buff, len, result, 0);
+		/* Convert user input to SHA256 hash value */
+		mbedtls_sha256((unsigned char *)line_buff, len, input_sha256_hex, 0);
 
 		tash_free(line_buff);
 
-		if (!strncmp((char *)sha256_hex, (char *)result, 32)) {
+		/* Confirm the password */
+		if (!strncmp((char *)pw_sha256_hex, (char *)input_sha256_hex, 32)) {
```
jeongchanKim, line 86: How about using `SHA256_HASH_nBYTE` instead of 32?

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #4918, opened 2021-04-27
build/configs/rtl8721csm/: Extend kernel size from 1792 to 2304

fengyunlai:
For BLE app usage, extend kernel size from 1792 to 2304, change bootloader and defconfig.


jeongchanKim:
Did you test with loadable config?

fengyunlai:
> Did you test with loadable config?

No I didn't, but since I took 512*2 = 1024 KB flash size, I drop 3904 to 2880 to not surpass the flash size.

deepaksrma (commented):
could you mention the reason of increasing the kernel size in the commit msg?

fengyunlai:
> could you mention the reason of increasing the kernel size in the commit msg?

Sure, I just mentioned it as for BLE app usage

sunghan-chang:
Because we merged PR #4947 , this PR is unnecessary. Let's close.

PR #4919, opened 2021-04-28
Add framework and external to common binary 

abhishek-samsung:
Added framework and external to common binary. Task Manager shares a variable with apps/builtin to register the builtin
apps information such as the entry point. With commit "Adding Framework and External to Common Binary", calls from Framework to Apps is not be possible. To resolve this issue, adding an API in Task Manager to initialize the above shared variables. This API is called just after the Task Manager is launched.

### DIFF
--- a/framework/include/task_manager/task_manager.h
+++ b/framework/include/task_manager/task_manager.h
@@ -32,6 +32,7 @@
 #include <sys/types.h>
 #include <signal.h>
 #include <pthread.h>
+#include <apps/builtin.h>
 
 /**
  * @brief Task State which managed by Task Manager
@@ -102,6 +103,22 @@ enum tm_defined_broadcast_msg {
 	TM_BROADCAST_MSG_MAX = TM_BROADCAST_SYSTEM_MSG_MAX,
 #endif
 };
+
+/**
+ * @brief Builtin Application Info Structure
+ */
+struct tm_builtin_info_s {
+	char name[CONFIG_TASK_NAME_SIZE];
+	TASK_CALLBACK entry;
+	int exectype;
+	int priority;
+	int stacksize;
+};
+typedef struct tm_builtin_info_s tm_builtin_info_t;
+
+extern tm_builtin_info_t *tm_builtin_list;
+extern int tm_builtin_list_cnt;
+
 /**
  * @brief Application Info Structure
  */
@@ -161,6 +178,14 @@ extern "C" {
  * @endcond
  */
 int task_manager(int argc, char *argv[]);
+/**
+ * @brief Initialize the variables tm_builtin_list and builtin_list_cnt
+ * @details @b #include <task_manager/task_manager.h>\n
+ * This API can initialize the variables tm_builtin_list and builtin_list_cnt.\n
+ * @param[in] builtin_list that contains information about the builtin apps
+ * @param[in] length of builtin_list
+ */
+int task_manager_populate_builtin_list(const builtin_info_t builtin_apps_list[], int length);
 /**
  * @brief Request to register a built-in task
  * @details @b #include <task_manager/task_manager.h>\n
--- a/framework/src/task_manager/task_manager_core.c
+++ b/framework/src/task_manager/task_manager_core.c
@@ -35,7 +35,6 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/ioctl.h>
-#include <apps/builtin.h>
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/clock.h>
 #include <tinyara/task_manager_drv.h>
@@ -83,6 +82,12 @@ static int task_manager_pid;
 		tmvdbg("Registered handle %d\n", handle);                       \
 	} while (0)
 
+
+/****************************************************************************
+ * Public Definitions
+ ****************************************************************************/
+tm_builtin_info_t *tm_builtin_list;
+int tm_builtin_list_cnt;
 
 /****************************************************************************
  * Public Functions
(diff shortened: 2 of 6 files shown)


framework/src/task_manager/task_manager_core.c
```diff
@@ -1495,6 +1500,20 @@ static int taskmgr_init_task_manager(void)
 	return OK;
 }
 
+void task_manager_populate_builtin_list(const builtin_info_t builtin_list[], int builtin_list_count)
+{
+	tm_builtin_list = TM_ALLOC(builtin_list_count * sizeof(tm_builtin_info_t));
```
dineshdas-samsung, line 1505: Can we please check tm_builtin_list is null ?

deepaksrma (commented):
change the commit msg like this:

`framework/task_manager: Fix for Framework accessing App side code`

because we need to tell in the commit where we are making changes.

same for other commit.
and 1 question, why did you split the commit?

framework/include/task_manager/task_manager.h
```diff
@@ -102,6 +103,21 @@ enum tm_defined_broadcast_msg {
 	TM_BROADCAST_MSG_MAX = TM_BROADCAST_SYSTEM_MSG_MAX,
 #endif
 };
+/**
```
amogh-samsung, line 106: Could you please leave a line's space after the previous structure definition?

framework/src/task_manager/task_manager_core.c
```diff
@@ -1495,6 +1500,20 @@ static int taskmgr_init_task_manager(void)
...
+	for (int i = 0; i < builtin_list_count; i++) {
+		strncpy(tm_builtin_list[i].name, builtin_list[i].name, CONFIG_TASK_NAME_SIZE);
+		tm_builtin_list[i].entry = builtin_list[i].entry;
+		tm_builtin_list[i].exectype = builtin_list[i].exectype;
+		tm_builtin_list[i].priority = builtin_list[i].priority;
+		tm_builtin_list[i].stacksize = builtin_list[i].stacksize;
+	}
+
+	builtin_list_cnt = builtin_list_count;
```
amogh-samsung, line 1514: Possible to make these two variable names a bit more distinct?
"cnt" vs "count" could be confusing in the future.

framework/src/task_manager/task_manager_permission.c
```diff
@@ -32,7 +31,7 @@ bool taskmgr_is_permitted(int handle, pid_t pid)
 	int builtin_cnt;
 	int permission = TM_PERMISSION(handle);
 
-	builtin_cnt = get_builtin_list_cnt();
+	builtin_cnt = builtin_list_cnt;
```
amogh-samsung, line 34: Similarly, here as well.
The two variable names in use are fairly close to each other.

amogh-samsung (commented):
Could you shorten the commit description in addition to the redefining of the commit message as suggested by @deepaksrma 
My suggestion would be -

> -Task Manager shares a variable to register the builtin apps information.
> -However, calls from Framework to Apps are not yet possible.
> -Hence, add an API in Task Manager to initialize the shared variable.
> -This API is called just after the Task Manager is launched.

You can decide better based on the context.

jeongchanKim:
I understand why you changed like this.
But I think, with loadable configuration, apps folder can be into common binary also.
Because real user apps will be in loadable_apps, and the apps in apps folder are almost about system utils.
(For example, TASH is in apps folder, and it should be into common binary.)
How do you think about?

kishore-sn:
@jeongchanKim 
If apps can be moved into common binary, then this change wont be necessary. 
Since, we have been keeping the apps in the app binary till now, we thought of making this change.

framework/src/task_manager/task_manager_core.c
```diff
@@ -1495,6 +1500,20 @@ static int taskmgr_init_task_manager(void)
 	return OK;
 }
 
+void task_manager_populate_builtin_list(const builtin_info_t builtin_list[], int builtin_list_count)
+{
+	tm_builtin_list = TM_ALLOC(builtin_list_count * sizeof(tm_builtin_info_t));
+	for (int i = 0; i < builtin_list_count; i++) {
```
thapav, line 1506: declare i in the start of the function separately

framework/src/task_manager/task_manager_core.c
Abhishek-05: Done

framework/src/task_manager/task_manager_core.c
Abhishek-05: Done

framework/src/task_manager/task_manager_core.c
Abhishek-05: Changed the global variable name to tm_builtin_list_cnt

framework/src/task_manager/task_manager_permission.c
Abhishek-05: Done

framework/include/task_manager/task_manager.h
Abhishek-05: Done

framework/src/task_manager/task_manager_core.c
abhishek-samsung: Done.

framework/include/task_manager/task_manager.h
abhishek-samsung: Done.

framework/src/task_manager/task_manager_core.c
abhishek-samsung: Changed the global variable name to tm_builtin_list_cnt .

framework/src/task_manager/task_manager_permission.c
abhishek-samsung: Done.

framework/src/task_manager/task_manager_core.c
abhishek-samsung: Done.

jeongchanKim:
> If apps can be moved into common binary, then this change wont be necessary.
> Since, we have been keeping the apps in the app binary till now, we thought of making this change.

Yes, I know.
I discussed with @sunghan-chang that apps folder can be into common binary. 
The reason is what I said before.
If you have another opinion, please let me know.

PR #4920, opened 2021-04-29
os/board/rtl8721csm : Modified TrustZone linker script to maintain wifi performance

keerthanalex:
Moved data/bss/COMMON to psram from sram to prevent data/bss/COMMON of newly added codes to be automatically loaded into sram
Performance related data/bss/COMMON files are kept on SRAM
Verified the performance on hello config, same as before


an4967 (approved):
It works well

PR #4922, opened 2021-04-29
[DO NOT MERGE] Fix TASH functions and handle '/' characters in path string according to POSIX standard.

amogh-samsung:
- Fix smartfs_finddirentry() returns number of bytes read from flash when path ends in extra '/' character due to retention of last
  assigned value to return variable.
- Repeated '/' in path should be ignored to align with POSIX standard.
- For Eg: A) paths = "/mnt/test1" and = "/mnt/test1/" and B )paths = "/mnt/test1" and "/mnt//test1" should be considered
  same while looking for the entry respectively..
- Fix behavior of "rm -r", should only delete directory paths

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -517,7 +517,6 @@ static ssize_t smartfs_read(FAR struct file *filep, char *buffer, size_t buflen)
 	bytesread = 0;
 	while (bytesread != buflen) {
 		/* Test if we are at the end of data */
-
 		if (sf->currsector == SMARTFS_ERASEDSTATE_16BIT) {
 			/* Break and return the number of bytes we read (may be zero) */
 
@@ -550,7 +549,6 @@ static ssize_t smartfs_read(FAR struct file *filep, char *buffer, size_t buflen)
 		}
 #endif
 		/* Calculate the number of bytes to read into the buffer */
-
 		bytestoread = bytesinsector - (sf->curroffset - (uint16_t)sizeof(struct smartfs_chain_header_s));
 		if (bytestoread + bytesread > buflen) {
 			/* Truncate bytesto read based on buffer len */
@@ -588,7 +586,6 @@ static ssize_t smartfs_read(FAR struct file *filep, char *buffer, size_t buflen)
 	}
 
 	/* Return the number of bytes we read */
-
 	ret = bytesread;
 
 errout_with_semaphore:
@@ -651,6 +648,10 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 
 	header = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
 	byteswritten = 0;
+
+	/* As long as the current file pointer positions for writing is less than the total data length in the file, we are doing and overwrite operation
+	 * When the file pointer positins surpasses the file's data length, it becomes an append operation which will be handled by a separate function
+	 */
 	while ((sf->filepos < sf->entry.datalen) && (buflen > 0)) {
 		/* Overwriting data caused by a seek, etc.  In this case, we need
 		 * to check if the write causes the file length to be extended
@@ -662,7 +663,6 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 
 		smartfs_setbuffer(&readwrite, sf->currsector, sf->curroffset, fs->fs_llformat.availbytes - sf->curroffset, (uint8_t *)&buffer[byteswritten]);
 		/* Limit the write based on available data to write */
-
 		if (readwrite.count > buflen) {
 			readwrite.count = buflen;
 		}
@@ -671,7 +671,6 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 
 		if (readwrite.count > sf->entry.datalen - sf->filepos) {
 			/* Limit the write length so we write to the current EOF. */
-
 			readwrite.count = sf->entry.datalen - sf->filepos;
 		}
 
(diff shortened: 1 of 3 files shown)


jeongchanKim (dismissed):
LGTM

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1193,13 +1202,13 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 					/* Test if the name matches */
 
 					if (strncmp(entry->name, fs->fs_workbuffer, fs->fs_llformat.namesize) == 0) {
-						/* We found it!  If this is the last segment entry,
-						 * then report the entry.  If it isn't the last
-						 * entry, then validate it is a directory entry and
+						/* We found it!  If this is the last segment entry, then report the entry
+						 * If this segment is followed only by and extra '/' character in the end,
```
dineshdas-samsung, line 1206: I guess it will be  "only by an extra" rather than " only by and extra " .

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1246,8 +1255,8 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 										readwrite.count = fs->fs_llformat.availbytes;
 										readwrite.buffer = (uint8_t *)fs->fs_chunk_buffer;
 
-										ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-										if (ret < 0) {
+										tmp = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+										if (tmp < 0) {
 											fdbg("Error reading sector %d header, ret : %d\n", readwrite.logsector, ret);
```
abhishek-samsung, line 1260: Can you please verify if the ret value printed above should be tmp instead of ret ?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1235,8 +1244,8 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 									/* Read the next sector of the file */
 
 									smartfs_setbuffer(&readwrite, dirsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
-									ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-									if (ret < 0) {
+									tmp = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+									if (tmp < 0) {
 										fdbg("Error in sector chain at %d, ret : %d\n", readwrite.logsector, ret);
```
abhishek-samsung, line 1249: Can you please verify if the ret value printed above should be tmp instead of ret ?

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Rectified, thanks

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Updated, thanks

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Updated, thanks

os/fs/smartfs/smartfs.h
```diff
@@ -416,6 +416,8 @@ uint32_t smartfs_rdle32(FAR const void *val);
 
 void smartfs_wrle32(uint8_t *dest, uint32_t val);
 
+bool is_path_last_slash(const char* path);
```
abhishek-samsung, line 419: Can you please change "char* path" to "char *path" to maintain consistency.

abhishek-samsung, line 419: How about changing the function name from "is_path_last_slash" to "path_ends_with_slash" ?

os/fs/smartfs/smartfs_utils.c
```diff
@@ -300,6 +300,30 @@ void smartfs_wrle32(uint8_t *dest, uint32_t val)
...
+ *
+ * Returned Value:
+ *   boolean "true" if path ends with '/', otherwise "false"
+ *
+ ****************************************************************************/
+
+bool is_path_last_slash(const char* path)
+{
+	int i;
+	for (i = 0; path[i] != '\0'; i++);
```
jeongchanKim, line 320: How about using `strlen`? It is almost same as libc `strlen`, but some arch specific code have much faster `strlen`.

amogh-samsung: Updated

amogh-samsung:
[CHECKPOINT]

The following issues have been taken care of so far: -
1. "`rm -r <>`" command works for file entries also but it should not
2. "`mkdir /mnt/test/`" command should work, i.e. even if there is an extra '/' at the end of the path string
3. Extra '/' characters in the path string should be handled according to POSIX standards while traversing for the entry.
    For Eg: - `/mnt/test//test1/` <=> `/mnt/test/test1`
4. Extra '/' character at the end of a path string is valid only as long as the path represents a directory entry

PENDING: -
1. `cd ../` does not work from TASH
2. working of "`mkdir /mnt//test`" is not ascertained, i.e. extra '/' in the middle

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1588,6 +1604,13 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		goto errout_with_semaphore;
 	}
 
+	/* If the relpath ends in '/' but the entry is not a directory, report an error */
+	if ((oldrelpath[strlen(oldrelpath) - 1] == '/') && ((oldentry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_FILE)) {
```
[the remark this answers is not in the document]
amogh-samsung: We see if a path is requested with last character '/' and path does not belong to a directory,
We get an error, so I tried to replicate the behavior.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1108,6 +1109,14 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 			seglen++;
 			ptr++;
 		}
+
+		/* Paths like "/mnt/test//file1" should be treated the same as "/mnt/test/file1" */
```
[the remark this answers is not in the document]
amogh-samsung: Yes, that should also work.
In-line with Linux behavior

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1375,10 +1383,18 @@ static int smartfs_unlink(struct inode *mountpt, const char *relpath)
 		/* The name exists -- validate it is a file, not a dir */
 
 		if ((entry.flags & SMARTFS_DIRENT_TYPE) == SMARTFS_DIRENT_TYPE_DIR) {
+			fdbg("%s is a directory\n", relpath);
```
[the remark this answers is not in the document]
amogh-samsung: Noted, will update the same

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1375,10 +1383,18 @@ static int smartfs_unlink(struct inode *mountpt, const char *relpath)
...
 			ret = -EISDIR;
 			goto errout_with_semaphore;
 		}
 
+		/* If the path does not represent a directory but the "relpath" string ends in '/', report an error */
+		if (relpath[strlen(relpath) - 1] == '/') {
+			fdbg("%s is not a directory\n", relpath);
```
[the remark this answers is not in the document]
amogh-samsung: Noted, will update the same

os/fs/smartfs/smartfs_smart.c
```diff
@@ -1588,6 +1604,13 @@ int smartfs_rename(struct inode *mountpt, const char *oldrelpath, const char *ne
 		goto errout_with_semaphore;
 	}
 
+	/* If the relpath ends in '/' but the entry is not a directory, report an error */
```
dineshdas-samsung, line 1607: It will be better if we replace relpath in comment with oldrelpath to avoid confusion?

amogh-samsung:
This PR is similar to #4928 
Both offer a solution to the same bugs with a different version of the changes.
4922 includes changes to fs layer majorly to handle path discrepancies.
4928 includes changes to vfs layer majorly to handle path discrepancies.

apps/system/utils/fscmd.c
```diff
@@ -1084,10 +1085,26 @@ static int tash_rm(int argc, char **args)
...
+				ret = -errno;
+				goto errout;
+			}
+
+			if (!S_ISDIR(st.st_mode)) {
+				FSCMD_OUTPUT("%s is not a directory\n", fullpath);
+				ret = -ENOTDIR;
+				goto errout;
+			}
+
```
Taejun-Kwon, line 1107: As Per I reviewed, this is very duplicated code, anyway please see below information..
I think -r should remove directory..

Usage: rm [OPTION]... [FILE]...
Remove (unlink) the FILE(s).

  -f, --force           ignore nonexistent files and arguments, never prompt
  -i                    prompt before every removal
  -I                    prompt once before removing more than three files, or
                          when removing recursively; less intrusive than -i,
                          while still giving protection against most mistakes
      --interactive[=WHEN]  prompt according to WHEN: never, once (-I), or
                          always (-i); without WHEN, prompt always
      --one-file-system  when removing a hierarchy recursively, skip any
                          directory that is on a file system different from
                          that of the corresponding command line argument
      --no-preserve-root  do not treat '/' specially
      --preserve-root   do not remove '/' (default)
  -r, -R, --recursive   remove directories and their contents recursively
  -d, --dir             remove empty directories
  -v, --verbose         explain what is being done
      --help     display this help and exit
      --version  output version information and exit

amogh-samsung:
Obsolete

PR #4924, opened 2021-04-30
Support boot parameters for kernel

jeongarmy:
- os/tools: Support boot parameters for kernel
- build/configs/rtl8721csm: Set partition for boot parameters and Define MAKE_BOOTPARAM

### DIFF
--- a/build/configs/rtl8721csm/hello/defconfig
+++ b/build/configs/rtl8721csm/hello/defconfig
@@ -243,9 +243,9 @@ CONFIG_ARCH_BOARD_HAVE_FLASH=y
 # Board-Partition Options
 #
 CONFIG_FLASH_PARTITION=y
-CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,1048,"
-CONFIG_FLASH_PART_TYPE="none,none,none,none,none,kernel,kernel,smartfs,"
-CONFIG_FLASH_PART_NAME="km0_bl,km4_bl,reserved,ss,system_data,kernel,kernel,userfs,"
+CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,4080,8,"
+CONFIG_FLASH_PART_TYPE="none,none,none,none,none,kernel,kernel,smartfs,bootparam,"
+CONFIG_FLASH_PART_NAME="km0_bl,km4_bl,reserved,ss,system_data,kernel,kernel,userfs,bootparam,"
 # CONFIG_ARCH_BOARD_HAVE_SECOND_FLASH is not set
 CONFIG_AUTOMOUNT=y
 CONFIG_AUTOMOUNT_USERFS=y
--- a/build/configs/rtl8721csm/rtl8721csm_download.sh
+++ b/build/configs/rtl8721csm/rtl8721csm_download.sh
@@ -75,6 +75,7 @@ function get_executable_name()
 		km0_bl) echo "km0_boot_all.bin";;
 		km4_bl) echo "km4_boot_all.bin";;
 		kernel|ota) echo "km0_km4_image2.bin";;
+		bootparam) echo "bootparam.bin";;
 		userfs) echo "rtl8721csm_smartfs.bin";;
 		*) echo "No Binary Match"
 		exit 1
@@ -250,6 +251,10 @@ download_specific_partition()
 
 	echo ""
 	echo "Download $exe_name COMPLETE!"
+
+	if [[ "${parts[$partidx]}" == "kernel" ]];then
+		download_specific_partition "bootparam"
+	fi
 }
 
 download_all()
--- a/build/configs/rtl8721csm/rtl8721csm_download.sh
+++ b/build/configs/rtl8721csm/rtl8721csm_download.sh
@@ -251,10 +251,6 @@ download_specific_partition()
 
 	echo ""
 	echo "Download $exe_name COMPLETE!"
-
-	if [[ "${parts[$partidx]}" == "kernel" ]];then
-		download_specific_partition "bootparam"
-	fi
 }
 
 download_all()
@@ -369,6 +365,10 @@ case $1 in
 		;;
 	erase|ERASE)
 		erase $1 $2
+		;;
+	kernel|KERNEL)
+		download_specific_partition $1
+		download_specific_partition "bootparam"
 		;;
 	*)
 		download_specific_partition $1
--- a/os/tools/mkbootparam.py
+++ b/os/tools/mkbootparam.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+############################################################################
+#
+# Copyright 2021 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+############################################################################
+import os
+import sys
+import struct
+
+config_file_path = os.path.dirname(__file__) + '/../.config'
+bootparam_file_path = os.path.dirname(__file__) + '/../../build/output/bin/bootparam.bin'
+
+SIZE_OF_BOOTPARAM_PART = 8192
+
+###########################################################################################
+#
+# This script generates Boot Parameters for kernel.
+#
+# Boot Parameter information :
+# +---------------------------------------------------------------------+
+# | Active Idx |  First Address  | Second Address  |      Reserved      |
+# |  (1 byte)  |    (4 bytes)    |    (4 bytes)    | (8Kbytes - 9bytes) |
+# +---------------------------------------------------------------------+
+#  * The Boot Parameters (8KB) should be downloaded and be located at the end of a flash.
+#  * The "Active Idx" is a index indicating which address value to use.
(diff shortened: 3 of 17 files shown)


os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+	for line in lines:
+		if config in line:
+			value = (line.split("=")[1])
+                        value = value.replace('"','').replace('\n','')
+			break
+	return value;
+
+def make_bootparam():
+    print "========== Making boot param start =========="
+    SIZE_OF_INDEX = 1
```
sunghan-chang, line 55: SIZE_OF_KERNEL_INDEX

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+    SIZE_OF_KERNEL_FIRST_ADDR = 4
+    SIZE_OF_KERNEL_SECOND_ADDR = 4
+
+    bootparam_size = SIZE_OF_INDEX + SIZE_OF_KERNEL_FIRST_ADDR + SIZE_OF_KERNEL_SECOND_ADDR
+
+    names = get_config_value(config_path, "CONFIG_FLASH_PART_NAME=").split(",")
+    sizes = get_config_value(config_path, "CONFIG_FLASH_PART_SIZE=").split(",")
+    names = filter(None, names)
+    sizes = filter(None, sizes)
```
sunghan-chang, line 64: Is this filter necessary?

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Calulate partition size of boot parameters
+    partition_size = 0
+    for index, name in enumerate(names):
+        if name == "bootparam":
+            partition_size = int(sizes[index]) * 1024
```
sunghan-chang, line 70: Isn't it fixed 8 KB?

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+############################################################################
+
+def get_config_value(file_name, config):
+    with open(file_name, 'r+') as f:
+        lines = f.readlines()
+	found = False
+	for line in lines:
+		if config in line:
+			value = (line.split("=")[1])
+                        value = value.replace('"','').replace('\n','')
```
jeongchanKim, line 49: Please check the alignment.

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+    SIZE_OF_KERNEL_SECOND_ADDR = 4
+
+    bootparam_size = SIZE_OF_INDEX + SIZE_OF_KERNEL_FIRST_ADDR + SIZE_OF_KERNEL_SECOND_ADDR
+
+    names = get_config_value(config_path, "CONFIG_FLASH_PART_NAME=").split(",")
+    sizes = get_config_value(config_path, "CONFIG_FLASH_PART_SIZE=").split(",")
+    names = filter(None, names)
+    sizes = filter(None, sizes)
+
+    # Calulate partition size of boot parameters
```
jeongchanKim, line 66: Calulate -> Calculate

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+# |  (1 byte)  |    (4 bytes)    |    (4 bytes)    |
+# +------------------------------------------------+
+#  * This data is written to the last of partition with size 8Kbytes.
+#
+############################################################################
+
+def get_config_value(file_name, config):
+    with open(file_name, 'r+') as f:
+        lines = f.readlines()
+	found = False
```
jeongchanKim, line 45: Is this `found` needed? There is no usage.

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+	for line in lines:
+		if config in line:
+			value = (line.split("=")[1])
+                        value = value.replace('"','').replace('\n','')
+			break
+	return value;
+
+def make_bootparam():
+    print "========== Making boot param start =========="
```
jeongchanKim, line 54: I think `Start to make boot param` is much better.

build/configs/rtl8721csm/rtl8721csm_download.sh
```diff
@@ -250,6 +251,10 @@ download_specific_partition()
 
 	echo ""
 	echo "Download $exe_name COMPLETE!"
+
+	if [[ "${parts[$partidx]}" == "kernel" ]];then
```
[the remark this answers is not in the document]
jeongarmy: When kernel is downloaded, boot parameters should be downloaded also.
That's because boot parameters include kernel information and they can be changed.

os/tools/mkbootparam.py
jeongarmy: It is for removing empty element from a list. The empty string is caused by `,` at the end of config value.

os/tools/mkbootparam.py
jeongarmy: I want to make sure right configuration in a script. I added a condition line 81.

build/configs/rtl8721csm/hello/defconfig
```diff
@@ -243,9 +243,9 @@ CONFIG_ARCH_BOARD_HAVE_FLASH=y
 # Board-Partition Options
 #
 CONFIG_FLASH_PARTITION=y
-CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,1048,"
-CONFIG_FLASH_PART_TYPE="none,none,none,none,none,kernel,kernel,smartfs,"
-CONFIG_FLASH_PART_NAME="km0_bl,km4_bl,reserved,ss,system_data,kernel,kernel,userfs,"
+CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,1048,8,"
```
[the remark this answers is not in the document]
jeongarmy: Right. I will modify

build/configs/rtl8721csm/hello/defconfig
```diff
@@ -243,9 +243,9 @@ CONFIG_ARCH_BOARD_HAVE_FLASH=y
 # Board-Partition Options
 #
 CONFIG_FLASH_PARTITION=y
-CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,1048,"
-CONFIG_FLASH_PART_TYPE="none,none,none,none,none,kernel,kernel,smartfs,"
-CONFIG_FLASH_PART_NAME="km0_bl,km4_bl,reserved,ss,system_data,kernel,kernel,userfs,"
+CONFIG_FLASH_PART_SIZE="16,8,16,472,8,1792,1792,1048,3032,8,"
+CONFIG_FLASH_PART_TYPE="none,none,none,none,none,kernel,kernel,smartfs,none,bootparam,"
+CONFIG_FLASH_PART_NAME="km0_bl,km4_bl,reserved,ss,system_data,kernel,kernel,userfs,free,bootparam,"
```
sunghan-chang, line 248: free? How about increase userfs instead of adding free partition?
If you don't agree, let's name it as reserved.

build/configs/rtl8721csm/rtl8721csm_download.sh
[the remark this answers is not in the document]
sunghan-chang: If this is necessary, it should be done before printing "Download complete".
But this function is only to download specific binary. I think this is not good option to support bootparam in function point of view.
I suggest adding conditional in argument parsing at line 366.
```
case $1 in
	all|ALL)
		download_all
		;;
	erase|ERASE)
		erase $1 $2
		;;
	kernel/KERNEL)
		download_specific_partition $1
		download_specific_partition bootparam
	*)
		download_specific_partition $1
		;;
esac
```

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,113 @@
...
+    with open(file_name, 'r+') as fp:
+        lines = fp.readlines()
+        found = False
+        for line in lines:
+            if config in line:
+                value = (line.split("=")[1])
+                value = value.replace('"','').replace('\n','')
+                found = True
+                break
+        fp.close()
```
jeongchanKim, line 52: In python, `with` will close the file automatically. So you don't nee to close manually.

os/tools/mkbootparam.py
sunghan-chang: 1. If we don't remove empty element, what is the problem?
2. Can we guarantee or give guidance that if name is none, size should be none as well?
   If we don't size and name could be mismatched, then it will cause a problem.

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,101 @@
...
+output_path = os.path.dirname(__file__) + '/../../build/output/bin/'
+file_path = output_path + 'bootparam.bin'
+
+FLASH_START_ADDR = 0x8000000
+
+############################################################################
+#
+# This script generates Boot Parameters for kernel.
+#
+# Boot Parameter information :
```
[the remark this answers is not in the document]
sunghan-chang: And you need to explain the "active idx" about what it means.
```
The value 0 in the "active idx" means the "first address" will be loaded or is loaded.
The value 1 in the "active idx" means the "second address" will be loaded or is loaded.
```

os/tools/mkbootparam.py
```diff
@@ -0,0 +1,111 @@
...
+# language governing permissions and limitations under the License.
+#
+############################################################################
+import os
+import sys
+import struct
+
+config_file_path = os.path.dirname(__file__) + '/../.config'
+output_path = os.path.dirname(__file__) + '/../../build/output/bin/'
+file_path = output_path + 'bootparam.bin'
```
jeongchanKim, line 25: How about renaming the value to indicate which file path it is?

os/tools/mkbootparam.py
jeongarmy: 1. An error occurs about null element in line 76.
2. I added verifying size about what you said in line 88.

sunghan-chang (approved):

PR #4925, opened 2021-05-02
network/dns: fix req_lwip_data

pillip8282:
The structure req_lwip_data contains lwip data structure. Therefore
application have to access lwip data type. It'll cause build error
if loadable option is enabled. So replace lwip data type to struct
sockaddr which is in socket api.

### DIFF
--- a/external/dhcpc/dhcpc.c
+++ b/external/dhcpc/dhcpc.c
@@ -488,8 +488,6 @@ int g_dhcpc_state;
  ****************************************************************************/
 static int dhcpc_request(void *handle, struct dhcpc_state *presult)
 {
-	int sock;
-	int ioctl_ret;
 	int idx;
 	if (!handle) {
 		ndbg("ERROR : handle must not be null\n");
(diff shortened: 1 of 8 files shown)


external/dhcpc/dhcpc.c
```diff
@@ -688,39 +686,19 @@ static int dhcpc_request(void *handle, struct dhcpc_state *presult)
...
-		ioctl_ret = ioctl(sock, SIOCLWIP, (unsigned long)&req);
-		if (ioctl_ret == ERROR) {
-			ndbg("dnsclient : ioctl() failed with errno: %d\n", errno);
-			close(sock);
-			return ioctl_ret;
+		char dns_buffer[32] = {0,};
+		dns_addr.sin_addr = presult->dnsaddr[idx];
+		int res = netlib_setdnsserver((struct sockaddr *)&dns_addr, -1);
+		if (res < 0) {
+			ndbg("Set DNS server fail idx(%d)\n", idx);
```
jeongchanKim, line 696: Even there is a fail case, this function returns OK. Is it okay?

os/net/netmgr/netmgr_ioctl_lwip.c
```diff
@@ -59,6 +59,31 @@ static struct netif *_netdev_dhcp_dev(FAR const char *intf)
...
+	if (index > 0) {
+		if (index >= CONFIG_NET_DNS_MAX_SERVERS) {
+			return -2;
+		} else {
+			dns_setserver(index, &dns_addr);
+			return 0;
+		}
+	}
+	/* Set DNS server to available slot.*/
+	for (int i = 0; i < CONFIG_NET_DNS_MAX_SERVERS; i++) {
```
dineshdas-samsung, line 76: Is it possible to please declare i at the start of the function separately?

pillip8282: it's possible but I don't think it's necessary.

sangwon03 (approved):
LGTM

PR #4928, opened 2021-05-04
[DO NOT MERGE] Fix "rm -r" function and normalize path string with extra '/' characters.

amogh-samsung:
- Fix smartfs_finddirentry() returns number of bytes read from flash when path ends in extra '/' character due to retention of last
  assigned value to return variable.
- Repeated '/' in path should be ignored to align with POSIX standard.
- For Eg: A) paths = "/mnt/test1" and = "/mnt/test1/" and B )paths = "/mnt/test1" and "/mnt//test1" should be considered
  same while looking for the entry respectively..
- Fix behavior of "rm -r", should only delete directory paths

### DIFF
--- a/apps/system/utils/fscmd.c
+++ b/apps/system/utils/fscmd.c
@@ -1084,10 +1084,26 @@ static int tash_rm(int argc, char **args)
 {
 	char *fullpath;
 	int ret = OK;
+	struct stat st;
 
 	if (args[2]) {
 		fullpath = get_fullpath(args[2]);
 		if (strncmp(args[1], "-r", strlen(args[1])) == 0) {
+			/* The request is to delete an entire directory recursively
+			   Check if the entry requested is not a directory and report
+			 */
+			if (stat(fullpath, &st) < 0) {
+				FSCMD_OUTPUT("stat failed with %s, error = %d\n", fullpath, errno);
+				ret = -errno;
+				goto errout;
+			}
+
+			if (!S_ISDIR(st.st_mode)) {
+				FSCMD_OUTPUT("%s is not a directory\n", fullpath);
+				ret = -ENOTDIR;
+				goto errout;
+			}
+
 			ret = delete_entry(fullpath);
 		} else {
 			FSCMD_OUTPUT("Usage: rm [-r] [FILE/DIR]\n");
@@ -1097,15 +1113,14 @@ static int tash_rm(int argc, char **args)
 		ret = unlink(fullpath);
 		if (ret != OK) {
 			FSCMD_OUTPUT(CMD_FAILED, args[0], "unlink");
-			fscmd_free(fullpath);
-			return ret;
+			ret = -errno;
+			goto errout;
 		}
 		FSCMD_OUTPUT("%s deleted\n", fullpath);
 	}
 
-	if (fullpath) {
-		fscmd_free(fullpath);
-	}
+errout:
+	fscmd_free(fullpath);
 
 	return ret;
 }
--- a/os/fs/vfs/fs_norm_path.c
+++ b/os/fs/vfs/fs_norm_path.c
@@ -1,0 +1,74 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/fs/fs.h>
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Variables
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: normalize_path
+ *
+ * Description: Normalize the path if it has extra '/' characters in the
+ *  middle or at the end.
+ *
+ ****************************************************************************/
+
+void normalize_path(const char *path, char **ret_path)
+{
+	int i;
+	int j = 1;
+	char *tmp = *ret_path;
+
+	/* Copy the first character anyways */
+	tmp[0] = path[0];
+
+	for (i = 1; path[i] != '\0'; i++) {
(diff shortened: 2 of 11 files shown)


apps/system/utils/fscmd.c
```diff
@@ -1084,10 +1084,26 @@ static int tash_rm(int argc, char **args)
 {
 	char *fullpath;
 	int ret = OK;
+	struct stat st;
 
 	if (args[2]) {
 		fullpath = get_fullpath(args[2]);
 		if (strncmp(args[1], "-r", strlen(args[1])) == 0) {
+			/* The request is to delete an entire directory recursively
+			   Check if the entry requested is not a direcotyr and report
```
dineshdas-samsung, line 1093: Can you please check the spelling of directory ?

amogh-samsung: Rectified, thanks

apps/system/utils/fscmd.c
```diff
@@ -1097,12 +1113,12 @@ static int tash_rm(int argc, char **args)
...
 			FSCMD_OUTPUT(CMD_FAILED, args[0], "unlink");
-			fscmd_free(fullpath);
-			return ret;
+			goto errout;
 		}
 		FSCMD_OUTPUT("%s deleted\n", fullpath);
 	}
 
+errout:
 	if (fullpath) {
```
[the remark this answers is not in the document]
amogh-samsung: Thanks for pointing that out :)

deepaksrma (commented):
LGTM

apps/system/utils/fscmd.c
```diff
@@ -1084,10 +1084,26 @@ static int tash_rm(int argc, char **args)
...
 
 	if (args[2]) {
 		fullpath = get_fullpath(args[2]);
 		if (strncmp(args[1], "-r", strlen(args[1])) == 0) {
+			/* The request is to delete an entire directory recursively
+			   Check if the entry requested is not a directory and report
+			 */
+			if (stat(fullpath, &st) < 0) {
+				FSCMD_OUTPUT("stat failed with %s\n", fullpath);
+				ret = ERROR;
```
jeongchanKim, line 1097: In `stat`, it sets the errno before returning failure case. So I think we'd better get the errno and set the `ret` as errno.

apps/system/utils/fscmd.c
```diff
@@ -1097,15 +1113,13 @@ static int tash_rm(int argc, char **args)
 		ret = unlink(fullpath);
 		if (ret != OK) {
 			FSCMD_OUTPUT(CMD_FAILED, args[0], "unlink");
-			fscmd_free(fullpath);
-			return ret;
+			goto errout;
```
jeongchanKim, line 1116: `unlink` also sets the errno on failure case. How about setting the `ret` as -errno?

apps/system/utils/fscmd.c
```diff
@@ -1084,10 +1084,26 @@ static int tash_rm(int argc, char **args)
...
+			/* The request is to delete an entire directory recursively
+			   Check if the entry requested is not a directory and report
+			 */
+			if (stat(fullpath, &st) < 0) {
+				FSCMD_OUTPUT("stat failed with %s\n", fullpath);
+				ret = ERROR;
+				goto errout;
+			}
+
+			if (!S_ISDIR(st.st_mode)) {
```
jeongchanKim, line 1101: In this case, is `unlink` not needed?
In `delete_entry()`, there is an unlink with same case.

amogh-samsung: This case checks if the path given along with rm -r does not represent a directory, then the command is invalid.

apps/system/utils/fscmd.c
amogh-samsung: Yes, updated

apps/system/utils/fscmd.c
amogh-samsung: Updated, thanks

apps/system/utils/fscmd.c
```diff
@@ -1097,15 +1114,14 @@ static int tash_rm(int argc, char **args)
...
+			ret = -errno;
+			goto errout;
 		}
 		FSCMD_OUTPUT("%s deleted\n", fullpath);
 	}
 
-	if (fullpath) {
-		fscmd_free(fullpath);
-	}
+errout:
```
thapav, line 1123: Is the if(fullpath) check not necessary anymore since its been deleted ?

amogh-samsung: Old comment on the same line.
fscmd_free(path) checks if path is NULL internally.

jeongchanKim (dismissed):
LGTM

thapav (dismissed):
LGTM

apps/system/utils/fscmd.c
```diff
@@ -1084,10 +1085,26 @@ static int tash_rm(int argc, char **args)
...
 	int ret = OK;
+	struct stat st;
 
 	if (args[2]) {
 		fullpath = get_fullpath(args[2]);
 		if (strncmp(args[1], "-r", strlen(args[1])) == 0) {
+			/* The request is to delete an entire directory recursively
+			   Check if the entry requested is not a directory and report
+			 */
+			if (stat(fullpath, &st) < 0) {
```
Taejun-Kwon, line 1096: Below check code already duplicated, delete_entry also checking this, they are almost same.

amogh-samsung: `delete_entry()` is being called recursively for all subentries in the directory.
Sometimes the entry is a file and sometimes a directory, but we cannot check if it is the parent entry for which the function is invoked.
So, despite `stat` already being invoked in `delete_entry()`, I had to add a check here...

amogh-samsung:
This PR is similar to #4922 
Both offer a solution to the same bugs with a different version of the changes.
4922 includes changes to fs layer majorly to handle path discrepancies.
4928 includes changes to vfs layer majorly to handle path discrepancies.

os/fs/vfs/fs_open.c
```diff
@@ -137,7 +141,15 @@ int vopen(FAR const char *path, int oflags, va_list ap)
 
 	/* Get an inode for this file */
 
-	inode = inode_find(path, &relpath);
+	if (strncmp(path, "/", 1) != 0) {
+		tmp = (char *)kmm_malloc(strlen(path));
```
dineshdas-samsung, line 145: What if tmp is NULL ? Can we please handle this case ?

os/fs/vfs/fs_rmdir.c
```diff
@@ -111,13 +113,19 @@ int rmdir(FAR const char *pathname)
...
+	const char *norm_path = NULL;
+	char *tmp = NULL;
 	int errcode;
 
 	/* Get an inode for this file.  inode_find() automatically increments the
 	 * reference count on the inode if one is found.
 	 */
 
-	inode = inode_find(pathname, &relpath);
+	tmp = (char *)kmm_malloc(strlen(pathname));
```
dineshdas-samsung, line 124: Same as above.

os/fs/vfs/fs_norm_path.c
```diff
@@ -0,0 +1,74 @@
...
+ ****************************************************************************/
+
+void normalize_path(const char *path, char **ret_path)
+{
+	int i;
+	int j = 1;
+	char *tmp = *ret_path;
+
+	/* Copy the first character anyways */
+	tmp[0] = path[0];
```
dineshdas-samsung, line 60: Can you please check ?
Since, there was also no check while calling  normalize_path(path, &tmp) for tmp==NULL;

amogh-samsung:
Under reconsideration

PR #4929, opened 2021-05-04
apps/examples/wifi_manager_sample : Fix stress test1 crash during exit (ref: PR4888)

s1-vivek:
FileName: apps/examples/wifi_manager_sample/wm_test_stress_test1.c
Refactored stress 1 test case to handle error cases and do proper de-init.
This prevents error propagation from 1 iteration to another.

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_pthread.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_pthread.c
@@ -1539,6 +1539,72 @@ static void tc_pthread_pthread_setcanceltype(void)
 }
 #endif
 
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+static void *test_mutex_consistent(void *param)
+{
+	int ret = -1;
+
+	ret = pthread_mutex_lock(&g_mutex);
+	if (ret == EOWNERDEAD) {
+		/* make the mutex consistent, which became inconsistent because holding thread has been dead*/
+		pthread_mutex_consistent(&g_mutex);
+		pthread_mutex_unlock(&g_mutex);
+	}
+
+	return (void *)ret;
+}
+
+/**
+* @fn                   :tc_pthread_pthread_mutex_consistent
+* @brief                :This tc tests pthread_mutex_consistent()
+* @Scenario             :the pthread_mutex_consistent() function can be used to mark the state protected by the mutex referenced by mutex as consistent again
+*                        and this situation arises, If an owner of a robust mutex terminates while holding the mutex,
+*                        the mutex becomes inconsistent. In this case, the mutex does not become normally usable again until the state is marked consistent.
+*                        after marking consistent next thread can acquire the lock but will receive EOWNERDEAD.
+* @API'scovered         :pthread_mutex_consistent
+* @Preconditions        :none
+* @Postconditions       :none
+* @return               :void
+*/
+static void tc_pthread_pthread_mutex_consistent(void)
+{
+	int ret;
+	void *status;
+	pthread_t thread1;
+	pthread_t thread2;
+	pthread_t thread3;
+	pthread_mutexattr_t mutex_attr;
+
+	pthread_mutexattr_init(&mutex_attr);
+	pthread_mutexattr_setrobust(&mutex_attr, PTHREAD_MUTEX_ROBUST);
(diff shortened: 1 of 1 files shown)


s1-vivek:
> Hard code values should be replaced with #define or const
> Other changes are fine.

As discussed the changes are fine as per the coding guidelines.

deepaksrma (commented):
could you paste the error logs in commit msg?
and why it was happening and how did you remove that?

vibhor-m:
> could you paste the error logs in commit msg?
> and why it was happening and how did you remove that?

Commit amend done.

jeongchanKim (approved):
LGTM

apps/examples/wifi_manager_sample/wm_test_stress_test1.c
```diff
@@ -122,20 +121,20 @@ static void wm_get_apinfo(wifi_manager_ap_config_s *apconfig)
...
 	ST_START_TEST;
+	ST_EXPECT_EQ(WIFI_MANAGER_SUCCESS, wifi_manager_deinit());
 	ST_END_TEST;
 }
 
 TEST_F(init)
 {
 	ST_START_TEST;
-	ST_EXPECT_EQ(WIFI_MANAGER_SUCCESS, wifi_manager_deinit());
+	ST_EXPECT_EQ(WIFI_MANAGER_SUCCESS, wifi_manager_init(&g_wifi_callbacks));
```
pillip8282, line 137: the purpose of TEST(init) is check not only deinit result but also time duration of deinit. So deinit must be in TEST_F(init) function.

apps/examples/wifi_manager_sample/wm_test_stress_test1.c
```diff
@@ -146,22 +145,22 @@ TEST_SETUP(connect)
...
 	ST_END_TEST;
 }
 
 TEST_F(connect)
 {
 	ST_START_TEST;
-	ST_EXPECT_EQ(WIFI_MANAGER_SUCCESS, wifi_manager_deinit());
+	wifi_manager_ap_config_s apconfig;
+	wm_get_apinfo(&apconfig);
+	ST_EXPECT_EQ(WIFI_MANAGER_SUCCESS, wifi_manager_connect_ap(&apconfig));
```
pillip8282, line 163: same as above

pillip8282 (changes requested):
I think you misunderstood the purpose of this test cases. deinit which is tested function must be inside TEST_F.

s1-vivek:
> I think you misunderstood the purpose of this test cases. deinit which is tested function must be inside TEST_F.

will do changes and add deinit in TEST_F

apps/examples/wifi_manager_sample/wm_test_stress_test1.c
s1-vivek: purpose of connect, connect_done, disconnect and scan is to measure time of these respective operations.
Is it required to add deinit in TEST_F and add time of deinit execution in total TC execution time


> same as above

purpose of connect, connect_done, disconnect and scan is to measure time of these respective operations.
Is it required to add deinit in TEST_F and add time of deinit execution in total TC execution time

s1-vivek: > same as above

purpose of connect, connect_done, disconnect and scan is to measure time of these respective operations.
Is it required to add deinit in TEST_F and add time of deinit execution in total TC execution time?

vibhor-m: stress1 -> Connect_done ->
description: call deinit() after receiving connect done event

Originally code
TEST_SETUP -> init, connect ap,
TEST_F -> deinit
only deinit time was measured.

Modified code:
TEST_SETUP -> init,
TEST_F -> , connect ap
TEST_TEARDOWN -> deint
only connect_ap time is measured.

If required. we can add deinit in TEST_F.

pillip8282:
> > I think you misunderstood the purpose of this test cases. deinit which is tested function must be inside TEST_F.
> 
> will do changes and add deinit in TEST_F

@s1-vivek 
Yes, the purpose of this TC is to measure deinit() accuracy and performance. So you have to add deinit in TEST_F

PR #4934, opened 2021-05-12
include/sys/boardctl.h : Add macro of boardctl

jeongchanKim:
To reset the board through sw, we should call boardctl(BOARDIOC_RESET, EXIT_SUCCESS/FAILURE).
For easy use, add macro as board_sw_reset()

### DIFF
--- a/os/include/sys/boardctl.h
+++ b/os/include/sys/boardctl.h
@@ -58,6 +58,7 @@
  ****************************************************************************/
 #include <tinyara/config.h>
 #include <stdint.h>
+#include <stdlib.h>
 
 #include <tinyara/fs/ioctl.h>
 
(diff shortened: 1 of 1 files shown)


os/include/sys/boardctl.h
```diff
@@ -169,5 +170,7 @@ int boardctl(unsigned int cmd, uintptr_t arg);
 }
 #endif
 
+#define board_sw_reset()	boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
```
sunghan-chang, line 173: 1. Need to remove `;` 
2. Don't we need to use `do { } while(0)` ?

os/include/sys/boardctl.h
```diff
@@ -58,6 +58,7 @@
  ****************************************************************************/
 #include <tinyara/config.h>
 #include <stdint.h>
+#include <stdlib.h>
```
sunghan-chang, line 61: Why do you need to add this for this commit?

jeongchanKim: `stdlib.h` is needed for `EXIT_SUCCESS`.

sunghan-chang (approved):

PR #4939, opened 2021-05-18
os/board/rtl8721csm: Implement Secure FTL

ziyik:
- Implement Secure FTL
- Due to image3 is not update, so Secure FTL is default set to off.
- To turn on Secure FTL please define FTL_SECURE as 1. (#define FTL_SECURE 1)

### DIFF
--- a/os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
+++ b/os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
@@ -123,6 +123,65 @@ extern bool ftl_page_erase(struct Page_T *p);
 void ftl_mapping_table_init(void);
 uint16_t read_mapping_table(uint16_t logical_addr);
 
+#define FTL_SECURE 1
+#if defined(CONFIG_AMEBAD_TRUSTZONE) && (FTL_SECURE == 1)
+/* FTL Secure Related */
+/* Secure efuse key location */
+#define FTL_KEY_ADDR 0x150
+#define FTL_SECURE_CONTEXT_SIZE 4096
+
+typedef struct {
+	void (*device_lock)(uint32_t);
+	void (*device_unlock)(uint32_t);
+	void (*setstatusbits)(uint32_t);
+	uint32_t (*save_to_storage)(void*, uint16_t, uint16_t);
+	uint32_t (*load_from_storage)(void*, uint16_t, uint16_t);
+} ns_ftl_func_struc;
+#endif
+
+/* Flash Status Bit */
+#define FLASH_STATUS_BITS 0x2c
+uint32_t backup_state = 0;
+//uint32_t cur_state = 0;
+
+static void ftl_ns_setstatusbits(uint32_t NewState)
+{
+	device_mutex_lock(RT_DEV_LOCK_FLASH);
+	FLASH_Write_Lock();
+	FLASH_SetStatusBits(FLASH_STATUS_BITS, NewState);
+	FLASH_Write_Unlock();
+	device_mutex_unlock(RT_DEV_LOCK_FLASH);
+}
+
+static void ftl_setstatusbits(uint32_t NewState)
+{
+	flash_t flash;
+
+	device_mutex_lock(RT_DEV_LOCK_FLASH);
+	if (!NewState) {	/* If to disable */
+		backup_state = flash_get_status(&flash);	/* Read State */
+		//FTL_PRINTF(FTL_LEVEL_INFO, "[ftl] backup_state: %x \n", backup_state);
+		backup_state = backup_state & 0xFF;		/* Only compare status bits */
+		if (FLASH_STATUS_BITS == backup_state) {	/* State is enable */
+			FLASH_Write_Lock();
+			FLASH_SetStatusBits(FLASH_STATUS_BITS, NewState);	/* Clear */
+			FLASH_Write_Unlock();
+			//cur_state = flash_get_status(&flash);	/* Read State */
+			//FTL_PRINTF(FTL_LEVEL_INFO, "[ftl] cur_state: %x \n", cur_state);
+		}
+	} else {
+		if (FLASH_STATUS_BITS == backup_state) {	/* State is enable */
+			FLASH_Write_Lock();
+			FLASH_SetStatusBits(FLASH_STATUS_BITS, NewState); /* Set State */
+			FLASH_Write_Unlock();
+			//cur_state = flash_get_status(&flash);	/* Read State */
+			//FTL_PRINTF(FTL_LEVEL_INFO, "[ftl] cur_state: %x \n", cur_state);
+		}
+		backup_state = 0;	/* Clear backup state */
+	}
+	device_mutex_unlock(RT_DEV_LOCK_FLASH);
+}
+
 uint32_t ftl_page_read(struct Page_T *p, uint32_t index)
 {
     uint32_t rdata = 0;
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@ziyik Could you let me know what this PR is?

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -1013,9 +1067,41 @@ uint32_t ftl_read(uint16_t logical_addr, uint32_t *value)
...
+	uint32_t ret = 0;
+	uint8_t* tmp_buff = NULL;
+
+	tmp_buff = rtw_zmalloc(size);
+	if (tmp_buff == NULL) {
+		return 0x10;
+	}
+
+	up_allocate_secure_context(FTL_SECURE_CONTEXT_SIZE);
+	ret = ameba_ftl_save_to_storage(tmp_buff, pdata_tmp, offset, size);
```
thapav, line 1083: Can we check for return value here and handle error scenarios ?

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -1013,9 +1067,41 @@ uint32_t ftl_read(uint16_t logical_addr, uint32_t *value)
...
+	rtw_mfree(tmp_buff, size);
+	return ret;
+}
+
+uint32_t ftl_secure_load_from_storage(void *pdata_tmp, uint16_t offset, uint16_t size)
+{
+	uint32_t ret = 0;
+
+	up_allocate_secure_context(FTL_SECURE_CONTEXT_SIZE);
+	ret = ameba_ftl_load_from_storage(pdata_tmp, offset, size);
```
thapav, line 1095: Same as above

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
ziyik: The error scenarios will be handle on the caller side.
We can print error log here is want to.

thapav: yes, I think that would be good for debugging purpose

ziyik: Added log for error scenarios, return value will be print if error occurs.

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -123,6 +123,56 @@ extern bool ftl_page_erase(struct Page_T *p);
 void ftl_mapping_table_init(void);
 uint16_t read_mapping_table(uint16_t logical_addr);
 
+#define FTL_SECURE 0
+#if defined(CONFIG_AMEBAD_TRUSTZONE) && (FTL_SECURE == 1)
```
jeongarmy, line 127: You defined a config for secure FTL here.
User needs to change this value in this file directly whenever enabling or disabling this feature.
Could you add this config to Kconfig for user to enable or disable this via menuconfig?
Is there any reason that you define and set value 0 here?

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -1013,9 +1067,47 @@ uint32_t ftl_read(uint16_t logical_addr, uint32_t *value)
...
+	up_free_secure_context();
+
+	return ret;
+}
+#endif
+
 // return 0 success
 // return !0 fail
-uint32_t ftl_save_to_storage(void *pdata_tmp, uint16_t offset, uint16_t size)
+uint32_t ftl_save_to_storage_i(void *pdata_tmp, uint16_t offset, uint16_t size)
```
jeongarmy, line 1110: What does `i` means? I think this function name is bad to understand.
Please rename it to name related to the actual operation for better understanding.

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -1554,6 +1682,12 @@ uint32_t ftl_init(uint32_t u32PageStartAddr, uint8_t pagenum)
     ftl_recover_from_power_lost();
     g_free_page_count = ftl_get_free_page_count();
 
+#if defined(CONFIG_AMEBAD_TRUSTZONE) && (FTL_SECURE == 1)
+	if (0 != ftl_secure_init()) {
+		FTL_PRINTF(FTL_LEVEL_ERROR, "ftl_secure_init fail!");
```
jeongarmy, line 1687: I think it should be an error code here.
That's because ftl_init always returns 0 although ftl_secure_init is failed here. Is it ok?
It can bring some errors in codes that run after ftl_init.
Please check and update this.

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
ziyik: When FTL_SECURE is define as 1, FTL will be secure it will make use of Secure API which is in TrustZone.
But this PR did not include that, is enable Secure FTL without updating TrustZone image i will create compile error.
So i turn it off by default.

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -1406,6 +1514,26 @@ uint32_t ftl_save(void *pdata, uint16_t offset, uint16_t size)
...
+	in_func.device_lock = device_mutex_lock;
+	in_func.device_unlock = device_mutex_unlock;
+	in_func.setstatusbits = ftl_ns_setstatusbits;
+	in_func.save_to_storage = ftl_save_to_storage_i;
+	in_func.load_from_storage = ftl_load_from_storage_i;
+
+	up_allocate_secure_context(FTL_SECURE_CONTEXT_SIZE);
+	ret = ameba_ftl_secure_init(FTL_KEY_ADDR, &in_func);
+	up_free_secure_context();
+}
```
jeongarmy, line 1534: The return type of this function is uint32_t. I think you need to add `return ret`.

sunghan-chang:
@ziyik Could you let us know why this is necessary?

ziyik:
> @ziyik Could you let us know why this is necessary?

This is a feature requested before, if without this feature FTL data is not encrypt. 
Plain data/key will be store in Flash, the feature will encrypt the data in Secure and Store in Flash. (with Secure encryption and Flash write protection.)

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
ziyik: It is for non-secure API, save or load without encryption.
I will rename it to ftl_non_secure_save_to_storage()

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
ziyik: Added error code return and print.

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
ziyik: Added return.

an4967 (approved):
LGTM

an4967 (changes requested):
@ziyik 
Can you test with 'test1' of (https://github.com/an4967/TizenRT/blob/check_ble/apps/examples/hello/hello_main.c)?

It still be stuck.

an4967 (approved):
LGTM

sunghan-chang (approved):

PR #4941, opened 2021-05-25
testcase/kernel/pthread: add pthread_mutex_consistent TC

deepaksrma:
- add new TC to test pthread_mutex_consistent()

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_pthread.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_pthread.c
@@ -1539,6 +1539,72 @@ static void tc_pthread_pthread_setcanceltype(void)
 }
 #endif
 
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+static void *test_mutex_consistent(void *param)
+{
+	int ret = -1;
+
+	ret = pthread_mutex_lock(&g_mutex);
+	if (ret == EOWNERDEAD) {
+		/* make the mutex consistent, which became inconsistent because holding thread has been dead*/
+		pthread_mutex_consistent(&g_mutex);
+		pthread_mutex_unlock(&g_mutex);
+	}
+
+	return (void *)ret;
+}
+
+/**
+* @fn                   :tc_pthread_pthread_mutex_consistent
+* @brief                :This tc tests pthread_mutex_consistent()
+* @Scenario             :the pthread_mutex_consistent() function can be used to mark the state protected by the mutex referenced by mutex as consistent again
+*                        and this situation arises, If an owner of a robust mutex terminates while holding the mutex,
+*                        the mutex becomes inconsistent. In this case, the mutex does not become normally usable again until the state is marked consistent.
+*                        after marking consistent next thread can acquire the lock but will receive EOWNERDEAD.
+* @API'scovered         :pthread_mutex_consistent
+* @Preconditions        :none
+* @Postconditions       :none
+* @return               :void
+*/
+static void tc_pthread_pthread_mutex_consistent(void)
+{
+	int ret;
+	void *status;
+	pthread_t thread1;
+	pthread_t thread2;
+	pthread_t thread3;
+	pthread_mutexattr_t mutex_attr;
+
+	pthread_mutexattr_init(&mutex_attr);
+	pthread_mutexattr_setrobust(&mutex_attr, PTHREAD_MUTEX_ROBUST);
+	pthread_mutex_init(&g_mutex, &mutex_attr);
+
+	/* creates the 1st thread to lock mutex */
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	TC_ASSERT_EQ("pthread_create", ret, 0);
+	pthread_join(thread1, &status);
+	TC_ASSERT_EQ_CLEANUP("test_mutex_consistent", (int)status, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* create 2nd thread to lock the mutex again, which is inconsistent, it will acquire the lock with an error return of EOWNERDEAD.
+	 * as mutex is in inconsistent state so this thread will create it consistent by calling pthread_mutex_consistent(). */
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	TC_ASSERT_EQ("pthread_create", ret, 0);
+	pthread_join(thread2, &status);
+	TC_ASSERT_EQ_CLEANUP("test_mutex_consistent", (int)status, EOWNERDEAD, pthread_mutex_unlock(&g_mutex));
+
+	/* Now thread3 is used to check whether mutex is working fine or not */
+	ret = pthread_create(&thread3, NULL, &test_mutex_consistent, NULL);
+	TC_ASSERT_EQ("pthread_create", ret, 0);
+	pthread_join(thread3, &status);
+	TC_ASSERT_EQ_CLEANUP("test_mutex_consistent", (int)status, 0, pthread_mutex_unlock(&g_mutex));
+
+	pthread_mutex_unlock(&g_mutex);
+
+	TC_SUCCESS_RESULT();
+}
+
+#endif
+
 /****************************************************************************
  * Name: pthread_main
  ****************************************************************************/
@@ -1561,6 +1627,9 @@ int pthread_main(void)
 	tc_pthread_pthread_mutex_init();
 	tc_pthread_pthread_mutex_destroy();
 	tc_pthread_pthread_mutex_lock_unlock_trylock();
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+	tc_pthread_pthread_mutex_consistent();
+#endif
 	tc_pthread_pthread_once();
 	tc_pthread_pthread_yield();
 	tc_pthread_pthread_cond_signal_wait();


apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+* @return               :void
+*/
+static void *test_mutex_consistent(void *param)
+{
+	int ret = -1;
+	ret = pthread_mutex_lock(&g_mutex);
+	if (ret != 0) {
+		printf("pthread_mutexlock was not successful, errno = %d", ret);
+	}
+}
```
sunghan-chang, line 1562: The return type of this function is void pointer so that you need to add `return NULL;` at the end of the function.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread1, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
+
+
+	ret = pthread_kill(thread1, 0);
+	if (ret != 0) {
+		printf("Thread 1 has died.\n");
+	}
```
sunghan-chang, line 1582: Could you let me know the purpose of this? This looks unnecessary because thread1 will be dead already.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_t thread1;
+	pthread_t thread2;
+
+	pthread_mutexattr_init(&g_mutexattr);
+	pthread_mutexattr_setrobust(&g_mutexattr, PTHREAD_MUTEX_ROBUST);
+	pthread_mutex_init(&g_mutex, &g_mutexattr);
+
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread1, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
```
sunghan-chang, line 1576: This checks pthread_create result. Let's change "pthread_mutex_consistent" to "pthread_create".

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_mutex_consistent(&g_mutex);
+
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread2, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
+
+	pthread_mutex_unlock(&g_mutex);
+
+	TC_SUCCESS_RESULT();
+}
```
sunghan-chang, line 1593: Could you add some comments to explain what this tc does?
just example
```
// create 1st pthread to lock mutex

// kill 1st pthread without mutex unlock

// call pthread_mutex_con....

// create 2nd pthread to lock the same mutex

// 2nd mutex lock is successfully done (It means pthread_mutex_con... works well)
```

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1554,6 +1607,9 @@ int pthread_main(void)
 	tc_pthread_pthread_set_get_schedparam();
 #if CONFIG_NPTHREAD_KEYS > 0
 	tc_pthread_pthread_key_create_delete_set_getspecific();
+#endif
+#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE
+	tc_pthread_pthread_mutex_consistent();
```
sunghan-chang, line 1612: How about moving this near another mutex tests?

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
 #endif
 
+/**
+* @fn                   :tc_pthread_pthread_mutex_consistent
+* @brief                :This tc tests pthread_mutex_consistent()
+* @Scenario             :the pthread_mutex_consistent() function can be used to mark the state protected by the mutex referenced by mutex as consistent again
+*                        and this situtaion arises, If an owner of a robust mutex terminates while holding the mutex,
+*                        the mutex becomes inconsistent and the next thread that acquires the mutex lock shall be notified of the state.
+*                        In this case, the mutex does not become normally usable again until the state is marked consistent.
+* @API'scovered         :pthread_setcanceltype
```
sunghan-chang, line 1550: You need check this.

sunghan-chang (changes requested):
Please check commit title format again. I already told you about that.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+
+
+	ret = pthread_kill(thread1, 0);
+	if (ret != 0) {
+		printf("Thread 1 has died.\n");
+	}
+
+	pthread_mutex_consistent(&g_mutex);
+
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
```
jeongchanKim, line 1586: I think we'd better make two pthreads independently.
1st is locking the mutex and finished,
2nd is trying to lock the previous mutex.
And check that return of `pthread_mutex_lock` is `EOWNERDEAD` or not from 2nd pthread.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_mutexattr_init(&g_mutexattr);
+	pthread_mutexattr_setrobust(&g_mutexattr, PTHREAD_MUTEX_ROBUST);
+	pthread_mutex_init(&g_mutex, &g_mutexattr);
+
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread1, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
+
+
+	ret = pthread_kill(thread1, 0);
```
jeongchanKim, line 1579: `pthread_kill` is for sending signal to the pthread.
So if you want to terminate the `thread1`, I think you should send signo `9`, instead of signo `0`.
How about using `pthread_cancel` if you want to terminate?

apps/examples/testcase/le_tc/kernel/tc_pthread.c
jeongchanKim, line 1550: Please change this to `pthread_mutex_consistent`.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+* @Postconditions       :none
+* @return               :void
+*/
+static void *test_mutex_consistent(void *param)
+{
+	int ret = -1;
+	ret = pthread_mutex_lock(&g_mutex);
+	if (ret != 0) {
+		printf("pthread_mutexlock was not successful, errno = %d", ret);
+	}
```
jeongchanKim, line 1561: Please add `return` of this function.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+* @Preconditions        :none
+* @Postconditions       :none
+* @return               :void
+*/
+static void *test_mutex_consistent(void *param)
+{
+	int ret = -1;
+	ret = pthread_mutex_lock(&g_mutex);
+	if (ret != 0) {
+		printf("pthread_mutexlock was not successful, errno = %d", ret);
```
jeongchanKim, line 1560: We don't want to use any `printf` in tcs, without TC_ASSERT macros.
Please change to use TC_ASSERT macros.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -66,6 +66,7 @@ static pthread_t g_thread1;
 static pthread_t g_thread2;
 
 static pthread_mutex_t g_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_mutexattr_t g_mutexattr;
```
jeongchanKim, line 69: Is this needed to locate the global variable? I think local variable is enough.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_mutex_init(&g_mutex, &g_mutexattr);
+
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread1, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
+
+
+	ret = pthread_kill(thread1, 0);
+	if (ret != 0) {
+		printf("Thread 1 has died.\n");
```
jeongchanKim, line 1581: Same as above. Please use TC_ASSERT macro.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	}
+
+	pthread_mutex_consistent(&g_mutex);
+
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread2, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
```
jeongchanKim, line 1588: For failure case, please use `TC_ASSERT_EQ_CLEANUP` for unlocking the used mutex.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_join(thread1, NULL);
+	TC_ASSERT_EQ("pthread_mutex_consistent", ret, 0);
+
+
+	ret = pthread_kill(thread1, 0);
+	if (ret != 0) {
+		printf("Thread 1 has died.\n");
+	}
+
+	pthread_mutex_consistent(&g_mutex);
```
jeongchanKim, line 1584: This tc is for `pthread_mutex_consistent`.
So I think we should check the return of this.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
jeongchanKim, line 1576: For every TC_ASSERT macro, the first parameter is for the API name which makes failure.
So in this case, it should be `pthread_create`.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
jeongchanKim, line 1612: I think we'd better move this tc after `tc_pthread_pthread_mutex_lock_unlock_trylock()`.

deepaksrma: yeah making changes

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
 }
 #endif
 
+/**
```
jeongchanKim, line 1543: These descriptions are for `tc_pthread_pthread_mutex_consistent()`, not `test_mutex_consistent()`.
So I think we'd better move these lines to line 1563.

jeongchanKim, line 1543: `pthread_mutex_consistent` API has dependency with `CONFIG_PTHREAD_MUTEX_UNSAFE`.
So we should enclose this tc with `#ifndef CONFIG_PTHREAD_MUTEX_UNSAFE`.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1540,58 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+* @fn                   :tc_pthread_pthread_mutex_consistent
+* @brief                :This tc tests pthread_mutex_consistent()
+* @Scenario             :the pthread_mutex_consistent() function can be used to mark the state protected by the mutex referenced by mutex as consistent again
+*                        and this situtaion arises, If an owner of a robust mutex terminates while holding the mutex,
```
jeongchanKim, line 1547: Plese fix the typo : situtaion -> situation.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: will change in pthread_cancel. thanks

sunghan-chang: Why? Could you let me know the reason? (I left the same comment above) Why do you need pthread_kill or pthread_cancel?

jeongchanKim: @sunghan-chang I agree with your opinion. We don't need `pthread_cancel`, because `thread1` will be terminated automatically.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: done

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: corrected

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: yes sunghan, you are right. no need of pthread_cancel() or pthread_kill() because pthread_exit() is already called from the pthread function.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: corrected now.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1539,60 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	static pthread_mutexattr_t mutex_attr;
+
+	pthread_mutexattr_init(&mutex_attr);
+	pthread_mutexattr_setrobust(&mutex_attr, PTHREAD_MUTEX_ROBUST);
+	pthread_mutex_init(&g_mutex, &mutex_attr);
+
+	/* creates the 1st thread to lock mutex */
+	ret = pthread_create(&thread1, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread1, &status);
+	TC_ASSERT_EQ_CLEANUP("pthread_create", ret, 0, pthread_mutex_unlock(&g_mutex));
```
jeongchanKim, line 1577: If pthread is not created well, we don't need to call `pthread_join()`. 
So I think we can move this line before calling `pthread_join()`.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1539,60 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", (int)status, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* make the mutex consistent, which became inconsistent because holding thread is dead*/
+	ret = pthread_mutex_consistent(&g_mutex);
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", ret, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* create 2nd thread to lock the mutex again, which is consistent now and if lock is succeed and this tc works*/
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	pthread_join(thread2, &status);
+	TC_ASSERT_EQ_CLEANUP("pthread_create", ret, 0, pthread_mutex_unlock(&g_mutex));
```
jeongchanKim, line 1587: Same as above.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
jeongchanKim: Okay. How about adding 2nd testcase scenario also?
Because, it is special to return `EOWNERDEAD` from `pthread_mutex_lock()`, so it is much better testcase for testing inconsistent mutex.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1539,60 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+* @Postconditions       :none
+* @return               :void
+*/
+static void tc_pthread_pthread_mutex_consistent(void)
+{
+	int ret;
+	void *status;
+	pthread_t thread1;
+	pthread_t thread2;
+	static pthread_mutexattr_t mutex_attr;
```
sunghan-chang, line 1568: No need to be `static`. Let's remove it.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: sorry, there is no first case. today I checked so TC was failing but yesterday it was pasing.
so TC is for only 2nd case.

jeongchanKim (dismissed):
LGTM

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1539,61 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+	pthread_join(thread1, &status);
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", (int)status, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* make the mutex consistent, which became inconsistent because holding thread is dead*/
+	ret = pthread_mutex_consistent(&g_mutex);
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", ret, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* create 2nd thread to lock the mutex again, which is consistent now and try to acquire the lock, it should be succeed and should return EOWNERDEAD*/
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	TC_ASSERT_EQ_CLEANUP("pthread_create", ret, 0, pthread_mutex_unlock(&g_mutex));
```
sunghan-chang, line 1586: Is this necessary even you successfully call pthread_mutex_consistent in this case?

sunghan-chang (commented):
Let's write the title like
```
testcase/kernel/pthread: add pthread_mutex_consistent TC
```
Then we can clearly know what this commit is without opening the commit.

apps/examples/testcase/le_tc/kernel/tc_pthread.c
deepaksrma: I think it can be better to check for the return value from pthread_create.
what you think, should I remove it or keep it?

sunghan-chang: Oh sorry. I missed exact word.
Is `pthread_mutex_unlock(&g_mutex)` necessary even you successfully call pthread_mutex_consistent in this case?
I don't mean that line itself.
What I said is
TC_ASSERT_EQ_CLEANUP -> TC_ASSERT_EQ.
Because unlock looks unnecessary.

sunghan-chang: I agree checking the return value from pthread_create.

jeongchanKim: I think `pthread_mutex_unlock` is needed. Because 2nd thread took the mutex.
`g_mutex` is used from another testcase, so It is better to unlock.

sunghan-chang: `pthread_mutex_unlock` in this line is executed when 2nd pthread_create is failed. It means the mutex is not locked. That's why I think it's unnecessary.

deepaksrma: yes sunghan, you are right.
i have made the required changes.

jeongchanKim (dismissed):
LGTM

an4967 (dismissed):
LGTM

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1539,6 +1539,61 @@ static void tc_pthread_pthread_setcanceltype(void)
...
+
+	/* make the mutex consistent, which became inconsistent because holding thread is dead*/
+	ret = pthread_mutex_consistent(&g_mutex);
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", ret, 0, pthread_mutex_unlock(&g_mutex));
+
+	/* create 2nd thread to lock the mutex again, which is consistent now and try to acquire the lock, it should be succeed and should return EOWNERDEAD*/
+	ret = pthread_create(&thread2, NULL, &test_mutex_consistent, NULL);
+	TC_ASSERT_EQ("pthread_create", ret, 0);
+	pthread_join(thread2, &status);
+	TC_ASSERT_EQ_CLEANUP("pthread_mutex_consistent", (int)status, EOWNERDEAD, pthread_mutex_unlock(&g_mutex));
```
sunghan-chang, line 1588: Is this right to return EOWNERDEAD? Doesn't it return EOWNERDEAD without calling `pthread_mutex_consistent`?

jeongchanKim: Sunghan's review is correct... In this TC, we called `pthread_mutex_consistent()`, so second `pthread_mutex_lock()` will not return `EOWNERDEAD`. It can successfully acquire the mutex.
So I think we need those two testcases.
Scenario1.
1) 1st thread terminates with locking the mutex.
2) 2nd thread tries to lock the mutex, then `pthread_mutex_lock()` returns `EOWNERDEAD`.
Scenario2.
1) 1st thread terminates with locking the mutex.
2) call `pthread_mutex_consistent()`
3) 2nd thread tries to lock the mutex, and it works normally.

deepaksrma: it is corrected now.

- When the owner of the mutex terminates without unlocking the mutex, the mutex is unlocked. The next owner of this mutex acquires the mutex with an error return of EOWNERDEAD.
- The new owner of this mutex should make the state protected by the mutex consistent. This state might have been left inconsistent when the previous owner terminated.
- If the new owner is able to make the state consistent, call pthread_mutex_consistent() for the mutex before unlocking the mutex. This marks the mutex as consistent and subsequent calls to pthread_mutex_lock() and pthread_mutex_unlock() will behave in the normal manner.

jeongchanKim (approved):
Looks good to me.

sunghan-chang (approved):

PR #4945, opened 2021-05-27
network/lwnl: modify lwnl to decouple wi-fi

pillip8282:
lwnl event queue is depedent to wi-fi events. so other
connectivities(etc. BLE) is difficult to use lwnl. This patch removes
all wi-fi related value in lwnl event queue. And this add device type
in lwnl_cb_status because events are come from various devices(wi-fi, BLE)


os/drivers/lwnl/Make.defs
```diff
@@ -18,7 +18,12 @@
 
 # Include lwnl drivers
 ifeq ($(CONFIG_LWNL80211),y)
-	CSRCS += lwnl.c lwnl_evt_queue.c lwnl_ioctl.c lwnl_netdev.c
+CSRCS += lwnl.c lwnl_evt_queue.c
+
+ifeq ($(CONFIG_NET_NETMGR), y)
```
jeongchanKim, line 23: If you want to add `lwnl_ioctl.c` and `lwnl_netdev.c` only when `CONFIG_NET_NETMGR` is not enabled,
how about using `ifneq`?

os/include/tinyara/net/if/wifi.h
```diff
@@ -6,6 +6,29 @@
...
+	lwnl_postmsg(LWNL_DEV_BLE, evt, buffer, buf_len)
+
+/*  if serialization is failed then driver should let pass the fail event to applications*/
+#define LWNL_POST_WIFISCANMSG(evt, scanlist)							\
+	do {																\
+		if (evt == LWNL_SCAN_DONE) {										\
+			uint8_t *buffer = NULL;										\
+			int lwnl_res = trwifi_serialize_scaninfo(&buffer, scanlist); \
+			if (lwnl_res < 0) {											\
+				LWNL_POST_WIFIMSG(LWNL_SCAN_FAILED, NULL, 0);				\
```
jeongchanKim, line 22: In this case, is it okay not to free the `buffer`?

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
```diff
@@ -97,7 +98,11 @@ static inline void LWNL_SET_MSG(wifimgr_msg_s *msg, wifimgr_evt_e event,
 
 static int _lwnl_call_event(int fd, lwnl_cb_status status, int len)
 {
-	switch (status) {
+	if (status.type != LWNL_DEV_WIFI) {
+		return 0;
```
jeongchanKim, line 102: In the below, failure case returns `-1`. Is it not the failure case?

os/include/tinyara/net/if/wifi.h
pillip8282: yes it's ok

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
pillip8282: it's not failure case.

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
```diff
@@ -143,10 +148,11 @@ int lwnl_fetch_event(int fd, void *buf, int buflen)
 {
 	lwnl_cb_status status;
 	uint32_t len;
-	char type_buf[8] = {0,};
+	char type_buf[sizeof(lwnl_cb_status) + sizeof(uint32_t)] = {0,};
+	size_t buf_len = sizeof(lwnl_cb_status) + sizeof(uint32_t);
```
jeongarmy, line 152: How about using the value with `#define XX_SIZE` like below?
```
#define XX_SIZE (sizeof(lwnl_cb_status) + sizeof(uint32_t))

char type_buf[XX_SIZE] = {0,};
...
int nbytes = read(fd, (char *)type_buf, XX_SIZE);
```

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
```diff
@@ -143,10 +148,11 @@ int lwnl_fetch_event(int fd, void *buf, int buflen)
...
 	handler_msg *hmsg = (handler_msg *)buf;
 
-	int nbytes = read(fd, (char *)type_buf, 8);
+	int nbytes = read(fd, (char *)type_buf, buf_len);
 	if (nbytes < 0) {
```
jeongarmy, line 156: It is error if only nbytes is negative integer with current condition, `if (nybytes < 0)`.
A value of nbytes can be 0 or positive integer not buf_len(8) like 3, 5.
It is invalid but not handled here and can make side effect later.
I think it should read all 8 bytes to succeed in this case.
So how about changing it to `if (nbytes != buf_len)`?

os/drivers/lwnl/lwnl_evt_queue.c
```diff
@@ -183,39 +158,19 @@ int lwnl_add_event(lwnl_cb_status type, void *buffer)
...
-			evt->data.status = LWNL_SCAN_FAILED;
-			evt->data.data = NULL;
-			evt->data.data_len = 0;
-			break;
+	evt->data.status = type;
+	if (buffer) {
+		char *output = kmm_malloc(buf_len);
+		if (!output) {
+			LWQ_ERR;
+			return -3;
```
jeongarmy, line 166: Some functions in this file return negative value like `-3, -2`, not defined error code.
How about making and using defined error code with enum?

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
pillip8282: file descriptor type is lwnl. if succeeds to read then it guarantees the read size. so it doesn't need to consider above case.

os/drivers/lwnl/lwnl_evt_queue.c
pillip8282: defining error code will improve readability. but it is not relate to this PR. So I'll modify it later.

os/net/netmgr/netdev_wifi.c
```diff
@@ -32,63 +32,63 @@ int netdev_handle_wifi(struct netdev *dev, lwnl_req cmd, void *data, uint32_t da
 {
 	lwnl_result_e res = LWNL_FAIL;
 	lldbg("T%d cmd(%d) (%p) (%d)\n", getpid(), cmd, data, data_len);
-	switch (cmd) {
-	case LWNL_INIT:
+	switch (cmd.type) {
+	case LWNL_REQ_WIFI_INIT:
```
jeongarmy, line 36: This change is good to understand.

os/net/netmgr/netdev_wifi.c
```diff
@@ -99,3 +99,30 @@ int netdev_handle_wifi(struct netdev *dev, lwnl_req cmd, void *data, uint32_t da
...
+
+int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
+{
+	trwifi_scan_list_s *item = scan_list;
+	int cnt = 0, total = 0;
+	while (item) {
+		item = item->next;
+		cnt++;
+	}
+	total = cnt;
```
jeongarmy, line 111: I think you can put `sizeof(trwifi_ap_scan_info_s) * cnt` to `total` here.
A `total` is used for total size as calculation with `sizeof(trwifi_ap_scan_info_s)`, not used alone anymore.
These changes can get good performance by removing unnecessary repetition of function calls.
You can change these like below:
```
total_size = sizeof(trwifi_ap_scan_info_s) * total;
*buffer = (uint8_t *)kmm_malloc(total_size);
...
return total_size;
```

os/net/netmgr/netdev_wifi.c
pillip8282: thanks

os/net/netmgr/netdev_wifi.c
```diff
@@ -99,3 +99,32 @@ int netdev_handle_wifi(struct netdev *dev, lwnl_req cmd, void *data, uint32_t da
 	}
 	return res;
 }
+
+int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
+{
+	trwifi_scan_list_s *item = scan_list;
+	int32_t cnt = 0, total = 0;
```
jeongchanKim, line 106: In TizenRT Coding rule, each variable should be declared in the new line.

apps/examples/lwnl_sample/lwnl_sample.c
```diff
@@ -75,7 +75,7 @@ static void *event_listener(void *arg)
 			}
 			memcpy(&status, buf, sizeof(lwnl_cb_status));
 			memcpy(&len, buf + sizeof(lwnl_cb_status), sizeof(uint32_t));
-			printf("[%d] status (%d) len(%d)\n", idx, status, len);
+			printf("[%d] status (%d) (%d) len(%d)\n", idx, status.type, status.evt, len);
```
jeongarmy, line 78: How about adding descriptions to log message like `status type (%d) evt (%d)` for easy debugging?

os/drivers/wireless/virtual/vdev_handler.c
```diff
@@ -29,31 +29,32 @@
...
+		trwifi_scan_list_s *scanlist = vwifi_get_scan_list();
+		LWNL_POST_WIFISCANMSG(LWNL_SCAN_DONE, scanlist);
 	} else {
-		lwnl_postmsg(evt, NULL);
+		LWNL_POST_WIFIMSG(evt.evt, NULL, 0);
 	}
 }
 
 static void _generate_disconnect(int argc, char *argv[])
 {
```
jeongarmy, line 47: I think we need to check `argc` for safety.
If not, a fault can be occurred because of invalid access when invalid arguments are passed.
Could you please check and add handling arguments?
```
if (argc <= XX) {
   return;
}
```

os/drivers/lwnl/lwnl_evt_queue.h
```diff
@@ -26,7 +26,7 @@
 
 struct lwnl_event;
 struct lwnl_queue;
-int lwnl_add_event(lwnl_cb_status type, void *buffer);
+int lwnl_add_event(lwnl_cb_status type, void *buffer, uint32_t buf_len);
```
amogh-samsung, line 29: Could you couple this function declaration with the others below instead of here, with the variable and object declarations please?

os/net/netmgr/netdev_wifi.c
```diff
@@ -99,3 +99,32 @@ int netdev_handle_wifi(struct netdev *dev, lwnl_req cmd, void *data, uint32_t da
...
+	if (!(*buffer)) {
+		ndbg("malloc fail %d\n", total);
+		return -1;
+	}
+
+	item = scan_list;
+	cnt = 0;
+	while (item) {
+		memcpy(*buffer + (item_size * cnt), &item->ap_info,
+			   item_size);
```
amogh-samsung, line 125: Can this be pushed back to one line please?

apps/examples/lwnl_sample/lwnl_sample.c
pillip8282: I think printing raw value is enough to debug.

os/drivers/wireless/virtual/vdev_handler.c
pillip8282: Only one function calls this function and it's static so adding exception handling code is not necessary.

os/drivers/lwnl/lwnl_evt_queue.h
pillip8282: I'll move the declaration of function. and I'll add description of the functions in another PR later.

PR #4949, opened 2021-06-02
build/configs/common_download.sh : Change ERASE ALL to erase userfs also

jeongchanKim:
./dbuild.sh ERASE ALL will erase whole partitions even userfs


sangam-swami (approved):
Looks good

sunghan-chang (approved):

PR #4952, opened 2021-06-03
fs/smartfs : Execute low level format when automount failed

Taejun-Kwon:
If automount failed, there should be internal error in smartfs,
For example, mount with journal enabled smartfs but smartfs which
disabled crc already written.

### DIFF
--- a/os/board/common/partitions.c
+++ b/os/board/common/partitions.c
@@ -34,6 +34,13 @@
 #include <tinyara/fs/mksmartfs.h>
 #endif
 
+#ifdef CONFIG_BOARDCTL_RESET
+#include <sys/boardctl.h>
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+#include <tinyara/reboot_reason.h>
+#endif
+#endif
+
 #include <tinyara/fs/mtd.h>
 #include <tinyara/fs/ioctl.h>
 
@@ -317,9 +324,29 @@ void automount_fs_partition(partition_info_t *partinfo)
 		ret = mount(fs_devname, "/mnt", "smartfs", 0, NULL);
 		if (ret != OK) {
 			lldbg("ERROR: mounting '%s' failed, errno %d\n", fs_devname, get_errno());
+
+#ifdef CONFIG_BOARDCTL_RESET
+			/* Mount was failed, so we'll try low level format forcely */
+			lldbg("Low level format will be done, then board will reset\n");
+
+#ifdef CONFIG_SMARTFS_MULTI_ROOT_DIRS
+			ret = mksmartfs(fs_devname, 1, true);
+#else
+			ret = mksmartfs(fs_devname, true);
+#endif
+			if (ret != OK) {
+				lldbg("ERROR: mksmartfs on %s failed again\n", fs_devname);
+				return;
+			} 
+			/* After llformat, Reset to free all resources & initialize dev of smartfs */
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+			up_reboot_reason_write(REBOOT_SYSTEM_USERFS_FORMAT);
+#endif
+			boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+#endif /* CONFIG_BOARDCTL_RESET */
 		}
 	}
-#endif
+#endif /* CONFIG_AUTOMOUNT_USERFS */
 
 #ifdef CONFIG_AUTOMOUNT_ROMFS
 	snprintf(fs_devname, FS_PATH_MAX, "/dev/mtdblock%d", partinfo->romfs_partno);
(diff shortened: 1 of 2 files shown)


os/board/common/partitions.c
```diff
@@ -34,6 +34,15 @@
 #include <tinyara/fs/mksmartfs.h>
 #endif
 
+#if defined(CONFIG_LIB_BOARDCTL) && defined(CONFIG_BOARDCTL_RESET)
+#include <sys/boardctl.h>
+#endif
+
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
```
jeongchanKim, line 41: How about moving these lines into the line from 37 to 39?
Because, reboot reason is used only when we need to reset the board.

os/board/common/partitions.c
```diff
@@ -34,6 +34,15 @@
 #include <tinyara/fs/mksmartfs.h>
 #endif
 
+#if defined(CONFIG_LIB_BOARDCTL) && defined(CONFIG_BOARDCTL_RESET)
```
jeongchanKim, line 37: `CONFIG_BOARDCTL_RESET` is enabled only when `CONFIG_LIB_BOARDCTL` is enabled.
So I think we can check only with `CONFIG_BOARDCTL_RESET`.

os/board/common/partitions.c
```diff
@@ -317,9 +326,29 @@ void automount_fs_partition(partition_info_t *partinfo)
...
+#ifdef CONFIG_SMARTFS_MULTI_ROOT_DIRS
+			ret = mksmartfs(fs_devname, 1, true);
+#else
+			ret = mksmartfs(fs_devname, true);
+#endif
+			if (ret != OK) {
+				lldbg("ERROR: mksmartfs on %s failed again\n", fs_devname);
+				return;
+			} 
+			/* After llformat, Reset to free all resources & initialialize smartfs dev */
```
jeongchanKim, line 343: Please modify `initialialize` to `initialize`.
And what is the meaning of llformat?

os/board/common/partitions.c
```diff
@@ -317,9 +326,29 @@ void automount_fs_partition(partition_info_t *partinfo)
...
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+			prctl(PR_REBOOT_REASON_WRITE, REBOOT_SYSTEM_USERFS_FORMAT);
```
jeongchanKim, line 345: This is os code, so you can use `up_reboot_reason_write(REBOOT_SYSTEM_USERFS_FORMAT);` directly.

os/board/common/partitions.c
Taejun-Kwon: ok good opinion.

os/board/common/partitions.c
Taejun-Kwon: modified, llformat is common expression, it means that low level format (physical format)

os/board/common/partitions.c
Taejun-Kwon: thanks.

jeongchanKim (approved):
LGTM

sunghan-chang (dismissed):

os/board/common/partitions.c
```diff
@@ -317,9 +324,29 @@ void automount_fs_partition(partition_info_t *partinfo)
...
+			/* Mount was failed, so we'll try low level format forcely */
+			lldbg("Low level format will be done, then board will reset\n");
+
+#ifdef CONFIG_SMARTFS_MULTI_ROOT_DIRS
+			ret = mksmartfs(fs_devname, 1, true);
+#else
+			ret = mksmartfs(fs_devname, true);
+#endif
+			if (ret != OK) {
+				lldbg("ERROR: mksmartfs on %s failed again\n", fs_devname);
```
amogh-samsung, line 338: Would it be right to say: -

> mksmartfs failed AGAIN

Since the reason for calling mksmartfs again is failure of `mount `function and not failure of `mksmartfs`

How about the statement: -

> redo/retry mksmartfs failed after mount failure

os/board/common/partitions.c
```diff
@@ -317,9 +324,29 @@ void automount_fs_partition(partition_info_t *partinfo)
...
+				return;
+			} 
+			/* After llformat, Reset to free all resources & initialize dev of smartfs */
```
amogh-samsung, line 341: Is it alright to mention `llformat` here?
Since that is an internal function called at the lower level, maybe not known to everybody.
How about mentioning: -

> After mksmartfs, reset to free....

After reformatting, resetting the device should now be able to mount the smartfs partition successfully, right?

amogh-samsung (commented):
Implementation and logic are all good.

Request you to kindly modify the commit description please: -

> If mount() function for smartfs fails, there could be an internal error in the smartfs partition.
> For example, mount() is requested with journaling enabled, but smartfs format with CRC
> option disabled is already written to the partition. Hence, re-format is required.

os/board/common/partitions.c
sangam-swami: current statement looks fine, as ret is return of mksmartfs not mount, mount ret value already printed

Taejun-Kwon:
We took another way, hence close this.

PR #4953, opened 2021-06-03
arch/arm/amebad_reboot_reason : Move up_reboot_reason_init() into up_…

jeongchanKim:
…initialize()

1. Move up_reboot_reason_init() into up_initialize()
2. Add log about previous reboot reason
3. Change to update the previous reboot reason when initializing the reboot reason, not reading the reason.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_reboot_reason.c
+++ b/os/arch/arm/src/amebad/amebad_reboot_reason.c
@@ -63,22 +63,17 @@
  ****************************************************************************/
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
 
-static reboot_reason_code_t backup_reg;
+static reboot_reason_code_t reboot_reason;
 
-void up_reboot_reason_init(void)
-{
-	/* Read the same backup register for the boot reason */
-	backup_reg = BKUP_Read(BKUP_REG1);
-	BKUP_Write(BKUP_REG1, REBOOT_REASON_INITIALIZED);
-
-}
-
-reboot_reason_code_t up_reboot_reason_read(void)
+static reboot_reason_code_t up_reboot_reason_get_hw_value(void)
 {
 	u32 boot_reason = 0;
 
-	if (backup_reg != REBOOT_REASON_INITIALIZED) {
-		return backup_reg;
+	/* Read the same backup register for the boot reason */
+	boot_reason = BKUP_Read(BKUP_REG1);
+
+	if (boot_reason != REBOOT_REASON_INITIALIZED) {
+		return boot_reason;
 	} else {
 		/* Read AmebaD Boot Reason, WDT and HW reset supported */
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -65,36 +65,44 @@
...
 	}
 
-	return REBOOT_UNKNOWN;
+	backup_reg = REBOOT_UNKNOWN;
+}
+
+reboot_reason_code_t up_reboot_reason_init(void)
+{
+	/* Read the same backup register for the boot reason */
+	backup_reg = BKUP_Read(BKUP_REG1);
```
sunghan-chang, line 97: This should be in the get_hw_val function.

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -65,36 +65,44 @@
...
+	up_reboot_reason_get_hw_value();
+	BKUP_Write(BKUP_REG1, REBOOT_REASON_INITIALIZED);
+	return backup_reg;
+}
+
+reboot_reason_code_t up_reboot_reason_read(void)
+{
+	return backup_reg;
```
sunghan-chang, line 105: The variable name can be renamed as reboot_reason. Because it saves other value as well.

os/arch/arm/src/amebad/amebad_reboot_reason.c
```diff
@@ -65,36 +65,44 @@
...
+	backup_reg = REBOOT_UNKNOWN;
+}
+
+reboot_reason_code_t up_reboot_reason_init(void)
+{
+	/* Read the same backup register for the boot reason */
+	backup_reg = BKUP_Read(BKUP_REG1);
+	up_reboot_reason_get_hw_value();
+	BKUP_Write(BKUP_REG1, REBOOT_REASON_INITIALIZED);
+	return backup_reg;
```
sunghan-chang, line 100: It looks strange that the function returns backup_reg without any code controlled. How about getting return value from get_hw_val and saving it in this and then return?

os/arch/arm/src/common/up_initialize.c
```diff
@@ -165,6 +169,12 @@ void up_initialize(void)
 
 	current_regs = NULL;
 
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	up_reboot_reason_init();
+	reboot_reason_code_t reason = up_reboot_reason_read();
+	lldbg("[Reboot Reason] : %d\n", reason);
```
sunghan-chang, line 175: It looks new variable is unnecessary. How about below?
```
lldbg("....", up_reboot_reason_read());
```

sunghan-chang (approved):

PR #4955, opened 2021-06-04
os/board/rtl8721csm: Disable all rtl8721csm BLE debug function by default

ziyik:
- Disable flash bit check to disable all BLE debug function

### DIFF
--- a/os/arch/arm/src/amebad/Kconfig
+++ b/os/arch/arm/src/amebad/Kconfig
@@ -73,6 +73,10 @@ config FTL_ENABLED
 	bool "Enable FTL"
 	default n
 
+config AMEBAD_BLE_DEBUG
+	bool "Enable BLE Debug"
+	default n
+
 config AMEBAD_BLE_PERIPHERAL
 	bool "Enable Amebad BLE Peripheral"
 	default n
--- a/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/inc/bt_board.h
+++ b/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/inc/bt_board.h
@@ -79,7 +79,11 @@ extern "C" {
 #define EFUSE_SW_DRIVER_DEBUG_LOG    BIT6
 #define EFUSE_SW_RSVD2               BIT7
 
+#ifdef CONFIG_AMEBAD_BLE_DEBUG
 #define CHECK_SW(x)                  (HAL_READ32(SPI_FLASH_BASE, FLASH_SYSTEM_DATA_ADDR + 0x28) & x)
+#else
+#define CHECK_SW(x)                  1
+#endif
 
 extern uint8_t rltk_wlan_is_mp(void);
 


os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/inc/bt_board.h
```diff
@@ -79,7 +79,7 @@ extern "C" {
 #define EFUSE_SW_DRIVER_DEBUG_LOG    BIT6
 #define EFUSE_SW_RSVD2               BIT7
 
-#define CHECK_SW(x)                  (HAL_READ32(SPI_FLASH_BASE, FLASH_SYSTEM_DATA_ADDR + 0x28) & x)
+#define CHECK_SW(x)                  1 /*(HAL_READ32(SPI_FLASH_BASE, FLASH_SYSTEM_DATA_ADDR + 0x28) & x)*/
```
sangam-swami, line 82: Can we use any debug config check for BT ?
like
#ifdef bt_debug
#define check_sw(x) ..
#else
#define check_sw(x) ..

ziyik: Okay, I will add a debug config for it.
it will be depend on AMEBAD_BLE.

os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/inc/bt_board.h
```diff
@@ -79,7 +79,11 @@ extern "C" {
 #define EFUSE_SW_DRIVER_DEBUG_LOG    BIT6
 #define EFUSE_SW_RSVD2               BIT7
 
+#ifdef CONFIG_BLE_DEBUG
 #define CHECK_SW(x)                  (HAL_READ32(SPI_FLASH_BASE, FLASH_SYSTEM_DATA_ADDR + 0x28) & x)
+#else
+#define CHECK_SW(x)                  1
```
jeongchanKim, line 85: Why do you remain 1 when `CONFIG_BLE_DEBUG` is not enabled?
How about remaining empty instead of 1?

os/arch/arm/src/amebad/Kconfig
```diff
@@ -73,6 +73,10 @@ config FTL_ENABLED
 	bool "Enable FTL"
 	default n
 
+config BLE_DEBUG
```
sunghan-chang, line 76: How about adding prefix as `AMEBAD`?
Because
1. All of others have prefix.
2. When we see defconfig, it could cause misunderstanding like it's common module config.

os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/inc/bt_board.h
ziyik: Other will call this CHECK_SW(x) and expect a return value, 0 mean enable, 1 disable.
So it have to return as 1 to disable all debug function.

os/arch/arm/src/amebad/Kconfig
ziyik: Your are right, i added AMEBAD prefix.

sunghan-chang: It's not added. Could you check again?

an4967 (changes requested):
@ziyik
Please check the "AMEBAD" prefix once again

os/arch/arm/src/amebad/Kconfig
ziyik: Sorry, my mistake, the commit did not add the changes file.
It is added now.

an4967 (approved):
LGTM

PR #4961, opened 2021-06-07
mtd/jedec : Add logic to verify erase

Taejun-Kwon:
When sudden power off happened, voltage level goes to down but
it can be grow up suddenly in product level.
In this case some of bits are still '0' but erase ops returns OK
To prevent this situation, verification is needed

### DIFF
--- a/os/fs/driver/mtd/mtd_jedec.c
+++ b/os/fs/driver/mtd/mtd_jedec.c
@@ -173,6 +173,7 @@ static inline void jedec_pagewrite(struct jedec_dev_s *priv, FAR const uint8_t *
 /* MTD driver methods */
 
 static int jedec_erase(FAR struct mtd_dev_s *dev, off_t startblock, size_t nblocks);
+static int jedec_is_erased(FAR struct jedec_dev_s *dev, off_t block);
 static ssize_t jedec_bread(FAR struct mtd_dev_s *dev, off_t startblock, size_t nblocks, FAR uint8_t *buf);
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/mtd_jedec.c
```diff
@@ -535,6 +536,70 @@ static inline void jedec_bytewrite(struct jedec_dev_s *priv, FAR const uint8_t *
 }
 #endif
 
+/************************************************************************************
+ * Name: jedec_is_erase
```
amogh-samsung, line 540: Would you consider renaming this function to `jedec_is_erased` or `is_jedec_erased` please?

os/fs/driver/mtd/mtd_jedec.c
```diff
@@ -535,6 +536,70 @@ static inline void jedec_bytewrite(struct jedec_dev_s *priv, FAR const uint8_t *
...
+	DEBUGASSERT(priv);
+	geo = priv->geo;
+
+	if (geo.subsectorshift > 0) {
+		erasesize = (1 << geo.subsectorshift);
+	} else {
+		erasesize = (1 << geo.sectorshift);
+	}
+
+	buffer = kmm_malloc(erasesize * sizeof(uint8_t));
```
sangam-swami, line 560: can we avoid multiple times allocation and free, as called in while (blocksleft > 0)

Taejun-Kwon: yeah good idea, but most of case this being called just one time. so I will keep current state, but I will improve whole of structure in future..

sunghan-chang (approved):

PR #4962, opened 2021-06-07
travis: add temporary travis account

an4967:
- add temporary travis account for pull limits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
     - TIZENRT_IMG_VERSION=latest
   
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
     - TIZENRT_IMG_VERSION=latest
   
 before_install:
+- echo "1111122222" | docker login -u "tizenrtpull" --password-stdin
```
jeongchanKim, line 21: I think we'd better hide the password.

jeongchanKim:
I think we can use `tizenrtpull` id permanently. So the commit msg can be like below.
`travis: add tizenrtpull id for managing the rate limit separately.`

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4967, opened 2021-06-09
wqueue : Modify to get hpwork, lpwork and usrwork from each getXXX APIs

jeongchanKim:
g_hpwork, g_lpwork, g_usrwork and g_usrsem/g_usrmutex are global variable, so it is not good to use directly.
So add getXXX APIs for each global variables.


thapav (approved):
LGTM

abhishek-samsung (dismissed):
LGTM

sungkyu0 (dismissed):
LGTM

sunghan-chang (dismissed):

PR #4968, opened 2021-06-09
network/netmgr: initialize allocated netif memory to 0

pillip8282:
if it's not zero then some module(ex. dhcpc) can access invalid
address of netif which is not initialized.

### DIFF
--- a/os/net/netmgr/netdev_lwip.c
+++ b/os/net/netmgr/netdev_lwip.c
@@ -782,7 +782,7 @@ static int lwip_init_nic(struct netdev *dev, struct nic_config *config)
 		return -1;
 	}
 
-	char *rnetif = (char *)kmm_malloc(sizeof(struct netif) + sizeof(struct netdev *));
+	char *rnetif = (char *)kmm_zalloc(sizeof(struct netif) + sizeof(struct netdev *));
 	if (!rnetif) {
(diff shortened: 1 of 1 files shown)


os/net/netmgr/netdev_lwip.c
```diff
@@ -782,7 +782,7 @@ static int lwip_init_nic(struct netdev *dev, struct nic_config *config)
 		return -1;
 	}
 
-	char *rnetif = (char *)kmm_malloc(sizeof(struct netif) + sizeof(struct netdev *));
+	char *rnetif = (char *)kmm_calloc(1, sizeof(struct netif) + sizeof(struct netdev *));
```
jeongchanKim, line 785: If you only want to initialize the memory after allocation, how about using `kmm_zalloc` instead of `kmm_calloc`?
`kmm_calloc` also initializes the memory, but it has more depth for allocation, so `kmm_zalloc` is better, I think.

jeongchanKim (approved):
LGTM

sangam-swami (approved):
looks good

thapav (approved):
LGTM

dineshdas-samsung:
LGTM

davidfather (approved):
Looks good to me.

PR #4969, opened 2021-06-09
os/board/rtl8721csm : Limit maximum connection to 1 in Soft AP mode

keerthanalex:
Added API to set maximum connection to one in Soft AP mode


sangwon03 (approved):
LGTM

PR #4973, opened 2021-06-10
board/common/partitions.c : Fix fs minor setting

jeongchanKim:
minor information of partition should be set for all cases, not only ftl_init


jeongarmy (dismissed):
LGTM

sunghan-chang (approved):

PR #4975, opened 2021-06-11
wifi/driver: add return value in cmd_wifi_off

pillip8282:
add return value in cmd_wifi_off

### DIFF
--- a/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
+++ b/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
@@ -789,12 +789,7 @@ int8_t cmd_wifi_off(void)
 		}
 	}
 	RTW_API_INFO("\n\rWIFI Mode Change instead of WIFI reload\r\n");
-
-#if 0
-	if (!wifi_off())
-		return 0;
-	return -1;
-#endif
+	return 0;
 #endif
 }
 


an4967 (approved):
LGTM

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -794,6 +794,8 @@ int8_t cmd_wifi_off(void)
 	if (!wifi_off())
 		return 0;
 	return -1;
+#else
```
davidfather, line 797: Why don't you remove '#if 0' statement, which is useless?

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -789,12 +789,7 @@ int8_t cmd_wifi_off(void)
 		}
 	}
 	RTW_API_INFO("\n\rWIFI Mode Change instead of WIFI reload\r\n");
-
-#if 0
-	if (!wifi_off())
-		return 0;
-	return -1;
-#endif
+	return 0;
```
jeongchanKim, line 792: How about moving this `return` after `#endif` in line 793?
Because for `##if CONFIG_ENABLE_P2P` condition, there is no return statement.

jeongchanKim, line 792: And this is not related to this PR, but it is better to add `#ifndef CONFIG_ENABLE_P2P` for the variable `mode` in line 766, because that is only used when `CONFIG_ENABLE_P2P` is not enabled.

pillip8282: I think it's better. however it's fixed by realtek wi-fi driver engineer.

davidfather (approved):
Looks good to me.

gexuyan:
Sorry that previous modifications didn't include return value in cmd_wifi_off. This modification looks good to me.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
gexuyan: Yes this way is better, or can move "int mode;" after line 778. Do you suggest me submitting a new pr for this modification? Thank you.

PR #4976, opened 2021-06-11
os/board/rtl8721csm: modify error flag for AP mac filter case

gexuyan:
Modify error flag in wifi_disconn_hdl for AP mac filter case from RTW_UNKNOWN to RTW_CONNECT_FAIL


j11591 (approved):
LGTM

pillip8282 (approved):
lgtm

PR #4980, opened 2021-06-15
apps/examples/testcase/le_tc/kernel: Add more task test cases 

thapav:
[Code Coverage] This patch adds few more TC's to task test cases.

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_task.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_task.c
@@ -50,9 +50,11 @@
 #define TC_TASK_CREATE_FAIL    1
 #define TC_REPARENT_FAIL       2
 
-static int tc_reparent_chk = 0;
 static int g_callback;
+#ifdef CONFIG_SCHED_HAVE_PARENT
 static int main_pid;
+static int tc_reparent_chk = 0;
+#endif
 #ifndef CONFIG_BUILD_PROTECTED
 static volatile int task_cnt;
 static volatile pid_t ppid;
--- a/apps/examples/testcase/le_tc/kernel/tc_task.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_task.c
@@ -416,6 +416,31 @@ static void tc_task_prctl(void)
 	TC_ASSERT_EQ_ERROR_CLEANUP("prctl", strncmp(getname, setname, CONFIG_TASK_NAME_SIZE), 0, get_errno(), prctl(PR_SET_NAME, oldname, 0, 0, 0));
 
 	prctl(PR_SET_NAME, oldname, 0, 0, 0);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	int reboot_reason_test_val = 100;
+
+	/* Clear, Write & Read the written reboot reason value */
+
+	ret_chk = prctl(PR_REBOOT_REASON_CLEAR);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+	ret_chk = prctl(PR_REBOOT_REASON_WRITE, reboot_reason_test_val);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+	ret_chk = prctl(PR_REBOOT_REASON_READ);
+	/* This check is only for a positive return value.
+	 * Reason being board cannot be rebooted for CTC evaluation
+	 * and the value written before a reboot only reflects after reboot */
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
+
+	/* Clear the latest reboot_reason_test_val reboot reason */
+
+	ret_chk = prctl(PR_REBOOT_REASON_CLEAR);
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
+#endif
+
 	TC_SUCCESS_RESULT();
 }
 


apps/examples/testcase/le_tc/kernel/tc_libc_sched.c
```diff
@@ -108,6 +108,26 @@ static void tc_libc_sched_task_setcanceltype(void)
 
 	TC_SUCCESS_RESULT();
 }
+
```
deepaksrma, line 111: remove extra line

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -37,6 +37,8 @@
 #include <tinyara/os_api_test_drv.h>
 #include "tc_internal.h"
 #include <semaphore.h>
+#include <tinyara/preference.h>
+#include <../arch/arm/include/reboot_reason.h>
```
sangam-swami, line 41: why do we need reboot_reason.h?

apps/examples/testcase/le_tc/kernel/tc_sched.c
```diff
@@ -661,6 +661,9 @@ static void tc_sched_task_setcanceltype(void)
 	TC_ASSERT_EQ_CLEANUP("task_setcanceltype", ret_chk, OK, goto errout);
 	TC_ASSERT_EQ_CLEANUP("task_setcanceltype", tcb->flags & TCB_FLAG_CANCEL_DEFERRED, defferred_flag, goto errout);
 
+
```
deepaksrma, line 664: you can remove this extra line

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +417,108 @@ static void tc_task_prctl(void)
 	TC_ASSERT_EQ_ERROR_CLEANUP("prctl", strncmp(getname, setname, CONFIG_TASK_NAME_SIZE), 0, get_errno(), prctl(PR_SET_NAME, oldname, 0, 0, 0));
 
 	prctl(PR_SET_NAME, oldname, 0, 0, 0);
+
```
sangam-swami, line 420: Can we add return check for above prctl call

deepaksrma (commented):
can you mention the exact no of TCs, which you are adding?
that would be more descriptive.

apps/examples/testcase/le_tc/kernel/tc_sched.c
```diff
@@ -661,6 +661,9 @@ static void tc_sched_task_setcanceltype(void)
...
+	task_testcancel();
```
jeongchanKim, line 665: Is this needed for testing `task_setcanceltype()`?

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +417,108 @@ static void tc_task_prctl(void)
...
+#ifdef CONFIG_MESSAGING_IPC
```
jeongchanKim, line 421: Those(messaging, preference, task monitor) are not kernel functionality, they are frameworks.
So there are their own samples. We can use them to check the coverage, we don't need to test from kernel testcases.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +417,108 @@ static void tc_task_prctl(void)
...
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+#endif
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	int reboot_reason_test_val = 100;
+
+	/* Read the previous reboot reason */
+
+	ret_chk = prctl(PR_REBOOT_REASON_READ);
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
```
jeongchanKim, line 508: For some cases, read value from reboot reason cannot be OK.
So I think we'd better test like below.
```
1. clear the reboot reason
2. write some value into reboot reason
3. try to read the reboot reason and check that the value is matched to wrote from No.2
```

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -520,6 +625,30 @@ static void tc_task_task_init(void)
...
+* @return               :void
+*/
+#ifdef CONFIG_TASK_MONITOR
+static void tc_task_task_monitor_utils(void)
+{
+	int ret_chk;
+	g_callback = ERROR;
+
+	ret_chk = task_monitor_register(SEC_2);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
```
jeongchanKim, line 644: In this case, the 1st parameter can be `task_monitor_register`, instead of `prctl`.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -520,6 +625,30 @@ static void tc_task_task_init(void)
...
+* API's covered         :task_monitor_register, task_monitor_update_status
+* Preconditions         :none
+* Postconditions        :none
+* @return               :void
+*/
+#ifdef CONFIG_TASK_MONITOR
+static void tc_task_task_monitor_utils(void)
+{
+	int ret_chk;
+	g_callback = ERROR;
```
jeongchanKim, line 641: Is this needed? Why do you set the `g_callback` variable?

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -541,6 +670,9 @@ int task_main(void)
 	tc_task_task_reparent();
 #endif
 	tc_task_task_init();
+#ifdef CONFIG_TASK_MONITOR
+	tc_task_task_monitor_utils();
```
jeongchanKim, line 674: TC name can be like below.
`tc_<module>_<API_name>`
So it can be `tc_task_task_monitor_register`.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +417,108 @@ static void tc_task_prctl(void)
...
+	/* Read the previous reboot reason */
+
+	ret_chk = prctl(PR_REBOOT_REASON_READ);
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
+
+	/* Write test reboot reason */
+
+	ret_chk = prctl(PR_REBOOT_REASON_WRITE, reboot_reason_test_val);
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
+	//Board  cannot be rebooted for CTC test case evaluation
```
jeongchanKim, line 514: I think we don't need to add this description about CTC.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +417,108 @@ static void tc_task_prctl(void)
...
+	ret_chk = prctl(PR_MSG_SAVE, portname, pid, param.sched_priority);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+	/* Removing receiver information */
+
+	ret_chk = prctl(PR_MSG_REMOVE, portname);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+
+#endif
+#ifdef CONFIG_TASK_MONITOR
```
jeongchanKim, line 453: This is duplicated with `tc_task_task_monitor_utils()`. So I think this is not needed.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -520,6 +625,30 @@ static void tc_task_task_init(void)
...
+* Postconditions        :none
+* @return               :void
+*/
+#ifdef CONFIG_TASK_MONITOR
+static void tc_task_task_monitor_utils(void)
+{
+	int ret_chk;
+	g_callback = ERROR;
+
+	ret_chk = task_monitor_register(SEC_2);
```
jeongchanKim, line 643: We cannot use task monitor API from user space.
I think, because of this, you made the tc in prctl.
Please remove this function.

apps/examples/testcase/le_tc/kernel/tc_sched.c
thapav: No, it is for covering task_testcancel() API.

jeongchanKim: Yes, but `task_testcancel()` can be covered from tc_libc_sched.c which you made.
Isn't it?

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +416,114 @@ static void tc_task_prctl(void)
...
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	int reboot_reason_test_val = 100;
+
+	/* Clear, Write & Read the written reboot reason value */
+
+	ret_chk = prctl(PR_REBOOT_REASON_CLEAR);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+	ret_chk = prctl(PR_REBOOT_REASON_WRITE, reboot_reason_test_val);
+	TC_ASSERT_EQ("prctl", ret_chk, OK);
+	ret_chk = prctl(PR_REBOOT_REASON_READ);
```
jeongchanKim, line 511: Is this tc passed?
When read the reboot reason, it returns the reboot reason value, not OK.
So I think we should check 100 instead of OK.
Please check this.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -414,6 +416,114 @@ static void tc_task_prctl(void)
...
+	TC_ASSERT_GEQ("prctl", ret_chk, OK);
+
+	/* Write new test reboot reason */
+
+	reboot_reason_test_val = 123;
+
+	ret_chk = prctl(PR_REBOOT_REASON_WRITE, reboot_reason_test_val);
```
jeongchanKim, line 518: Why do you write two times?

apps/examples/testcase/le_tc/kernel/tc_sched.c
thapav: So every time we enable/disable the CONFIG_CANCELLATION_POINTS ,
either of these will get executed. Hence the duplication.

apps/examples/testcase/le_tc/kernel/tc_task.c
thapav: Like I said earlier, currently it is only supported by realtek board. So to avoid unnecessary ASSERTION during testcase check, I have added GEQ check.  It can be updated later on.
What is your opinion ?

jeongchanKim: If a board will not support the reboot reason, then we should not enable `CONFIG_SYSTEM_REBOOT_REASON`.
In that case, this tc will not be tested. (It will be build error if a boart doesn't support the reboot reason, but enabling that config. Because in `prctl`, we cannot find the `up_reboot_reason_xxx()`s.)
So I think we don't need to consider

thapav: As discussed offline, there is issue with reboot reason return value in the module itself. (as per reboot reason sample app on latest code)
I will update the handling of the test case return value to:
`TC_ASSERT_EQ("prctl", ret_chk, reboot_reason_test_val)`

jeongchanKim: As discussed offline, reboot reason works well. Please modify the tc.
If you have any problem, please let me know.

apps/examples/testcase/le_tc/kernel/tc_task.c
sunghan-chang: Could you check @jeongchanKim 's comment ?

sunghan-chang:
@thapav Could you check @jeongchanKim 's comment?

apps/examples/testcase/le_tc/kernel/tc_task.c
thapav: The reboot reason value written to the BKUP_REG only reflects after reboot. Otherwise it shows the previous reboot reason value. So I have mentioned in the comments the reason for a TC_ASSERT_GEQ check and did not modify it. This TC is only to cover code.

thapav:
@sunghan-chang 
Apologies for the late response, modified the code. Please check.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -37,6 +37,7 @@
 #include <tinyara/os_api_test_drv.h>
 #include "tc_internal.h"
 #include <semaphore.h>
+#include <tinyara/preference.h>
```
jeongchanKim, line 40: Please remove unnecessary header.

apps/examples/testcase/le_tc/kernel/tc_libc_sched.c
```diff
@@ -108,6 +108,25 @@ static void tc_libc_sched_task_setcanceltype(void)
...
+ *                        cancelability is disabled
+ * @API'scovered         :task_testcancel
+ * @Preconditions        :none
+ * @Postconditions       :none
+ * @return               :void
+ */
+
+static void tc_libc_sched_task_testcancel(void)
+{
+	task_testcancel();
```
sunghan-chang, line 126: This was removed by https://github.com/Samsung/TizenRT/commit/3ffed1349c59b477213d9246c5982065be3b15bf#diff-aafb30cd3a9000323b9006de33be0fdeafa0e0489b84f161984a05605e1a0960. TC should include atleast one TC_ASSERT_XX macro. But this is not. Let's remove. How about removing function itself?

thapav: yes if the function is not there , we can remove

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4983, opened 2021-06-16
binfmt/binfmt_loadbinary.c : Fix build error about undeclared variabl…

jeongchanKim:
…e for some configs

Variable nregion is used not only #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME, but also #else case.


davidfather (approved):
Looks good to me.

thapav (approved):
LGTM

sunghan-chang (approved):

PR #4985, opened 2021-06-17
os/arch/arm: Enable Nested Interrupts on RTK board

kishore-sn:
Enable nested interrupts on RTK board. Add code for handling nested
interrupts in common exception hanlder. Enable the necessary CONFIGS
under CORTEX M33.

If lazy FPU is enabled, then the handling of interrupts is faster as
the saving of FPU registers is deferred. However, there is no mandatory
dependency between lazy FPU and nested interrupt handling. Currently,
the configs for FPU and LAZYFPU have been disabled in RTK.

Nested interrupts have been verified with and without lazyfpu.


build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -182,8 +184,9 @@ CONFIG_BOARD_LOOPSPERMSEC=22222
 # Interrupt options
 #
 CONFIG_ARCH_HAVE_INTERRUPTSTACK=y
-CONFIG_ARCH_INTERRUPTSTACK=0
-# CONFIG_ARCH_HAVE_NESTED_INTERRUPT is not set
+CONFIG_ARCH_INTERRUPTSTACK=1024
```
jeongchanKim, line 187: Is using interrupt stack mandantory?

kishore-sn: According to the Kconfig it is mandatory.

build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -102,6 +103,7 @@ CONFIG_REG_STACK_OVERFLOW_PROTECTION=y
 #
 CONFIG_ARMV8M_HAVE_ICACHE=y
 CONFIG_ARMV8M_HAVE_DCACHE=y
+# CONFIG_ARMV8M_LAZYFPU is not set
```
jeongchanKim, line 106: You said that enabling FPU, the handling of interrupts is faster.
Why don't we enable then?
What is the trade-off?

jeongchanKim (commented):
Could you update the test result about nested interrupt functionality in commit msg?

build/configs/rtl8721csm/loadable_apps/defconfig
kishore-sn: FPU is not enabled in RTK. So, we dont need to use LAZYFPU setting. 
Also, the drawback of LAZYFPU is that we cannot use floating point operations in interrupt handler because previous FPU state is not saved.

kishore-sn:
> Could you update the test result about nested interrupt functionality in commit msg?

Done

sunghan-chang (dismissed):

sunghan-chang:
@kishore-sn Sorry for delay. Let us verify DA basic operation first and then let's merge.

os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -202,14 +202,14 @@ exception_common:
...
-#ifndef CONFIG_ARCH_HIPRI_INTERRUPT
+#if !defined(CONFIG_ARCH_HIPRI_INTERRUPT) && !defined(CONFIG_ARCH_NESTED_INTERRUPT)
 	/* Disable interrupts, select the stack to use for interrupt handling
 	 * and call up_doirq to handle the interrupt
 	 */
 
 	cpsid	i						/* Disable further interrupts */
 
-#else
+#elif !defined(CONFIG_ARCH_NESTED_INTERRUPT)
```
sangam-swami, line 212: why can't we put CONFIG_ARCH_NESTED_INTERRUPT in #else?

os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -376,12 +408,15 @@ exception_common:
 	.bss
 	.global	g_intstackalloc
 	.global	g_intstackbase
+	.global	g_nestlevel
```
sangam-swami, line 411: is there any limitation on interrupt nesting level?

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: No.. First case if when nested and basepri are both undefined. Second case is when basepri is defined but nested is undefined. Third case below is for only nested being defined.

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: No. There is no limit. But when nestlevel becomes zero, then it means we are returning to process context and so context unstacking will be done.

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: @sangam-swami 
I modified according to your comments.

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4989, opened 2021-06-18
dbuild: Support build test with dbuild

jeongarmy:
Support build test with dbuild to allow users to check build errors before pushing commits

### DIFF
--- a/os/dbuild.sh
+++ b/os/dbuild.sh
@@ -105,6 +105,7 @@ function SELECT_OPTION()
 			if [ "${STATUS}" == "BUILT" ]; then
 				echo "  \"d. Download\""
 			fi
+			echo "  \"t. Build Test\""
 			echo "  \"x. Exit\""
 			echo ======================================================
 			read SELECTED_START
@@ -132,13 +133,15 @@ function SELECT_OPTION()
 		6|smartfs)
 			BUILD smartfs
 			;;
-
 		d|download)
 			if [ "${STATUS}" == "BUILT" ]; then
 				STATUS=PREPARE_DL
 			else
 				echo "No output file"
 			fi
+			;;
+		t|test)
+			BUILD_TEST
 			;;
 		x|exit)
 			exit 1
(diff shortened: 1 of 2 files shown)


os/dbuild.sh
```diff
@@ -154,6 +157,14 @@ function SELECT_OPTION()
 	fi
 }
 
+function BUILD_TEST()
+{
+	echo "Build Test START!! Docker Image Version : ${DOCKER_VERSION}"
```
sunghan-chang, line 162: Why is docker image version necessary?

os/dbuild.sh
```diff
@@ -154,6 +157,14 @@ function SELECT_OPTION()
...
+	pushd ${OSDIR} > /dev/null
+	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged tizenrt/tizenrt:${DOCKER_VERSION} bash -c "./tools/build_test.sh"
```
sunghan-chang, line 164: If we make this separately, when we need to modify docker command, we should modify two different lines. How about making a new function for this and call the same function at here and build function for maintainability?

os/tools/build_test.sh
```diff
@@ -55,3 +55,12 @@ for target in ${build_targets[@]}; do
...
 done
+
+	echo "===================================="
+	echo "======== Clean build output ========"
+	echo "===================================="
+	cd ${OSDIR}; ${CLEAN_CMD}
+
+	echo "===================================="
+	echo "====== Build Test All Success ======"
+	echo "===================================="
```
sunghan-chang, line 66: 1. This is not a part of "adding build test option in dbuild". Let's split commits.
2.  The dbuild script prints start of build test, but this script does end. It looks strange.

os/dbuild.sh
jeongarmy: will modify them with another PR.

sunghan-chang (approved):

PR #4990, opened 2021-06-18
fs/smartfs: Fix smartfs_finddirentry() might return unexpected values

amogh-samsung:
- "ret" variable gets assigned read byte count from FS_IOCTL (fs->fs_llformat.availbytes).
- This value might be returned by the function but it should only return from {OK, error code}
- Hence, check the return value from FS_IOCTL, if negative, assign to ret and proceed to
  return, otherwise store in temp variable only.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -245,22 +245,22 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 			goto errout_with_buffer;
 		}
 
-		/* TODO: Test open mode based on the file mode */
+		/* The file exists already.
+		 * If the file has been requested to be opened in TRUNCATE mode, then shrink the file.
+		 * However, if APPEND mode has also been requested, do not shrink the file to 0.
+		 */
 
-		/* The file exists.  Check if we are opening it for O_CREAT or
-		 * O_TRUNC mode and delete the sector chain if we are. */
+		/* TODO: This method for checking the open modes is not in full accordance with POSIX standard.
+		 *       POSIX standard truncates the file i.e. shrinks to 0 irrespective of other included flags.
+		 */
 
-		if ((oflags & (O_CREAT | O_TRUNC)) != 0) {
-			/* Don't truncate if open for APPEND */
-
-			if (!(oflags & O_APPEND)) {
-				/* Truncate the file to length 0 as part of the open */
-				ret = smartfs_shrinkfile(fs, sf, 0);
-				if (ret < 0) {
-					goto errout_with_buffer;
-				}
-			}
-		}
+		if (((oflags & O_TRUNC) != 0) && ((oflags & O_APPEND) == 0)) {
+			/* Truncate the file to length 0 as part of the open */
+			ret = smartfs_shrinkfile(fs, sf, 0);
+			if (ret < 0) {
+				goto errout_with_buffer;
+                        }
+                }
 	} else if (ret == -ENOENT) {
 		/* The file does not exist.  Were we asked to create it? */
 		if ((oflags & O_CREAT) == 0) {


os/fs/smartfs/smartfs_utils.c
```diff
@@ -1163,8 +1164,8 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 				/* Read the next directory in the chain */
 
 				smartfs_setbuffer(&readwrite, dirsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
-				ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-				if (ret < 0) {
+				tmp = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+				if (tmp < 0) {
```
sangam-swami, line 1168: what will be the "ret" value in case of errout? we need to return proper error ret value.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1235,9 +1236,9 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 									/* Read the next sector of the file */
 
 									smartfs_setbuffer(&readwrite, dirsector, 0, sizeof(struct smartfs_chain_header_s), (uint8_t *)fs->fs_rwbuffer);
-									ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
-									if (ret < 0) {
-										fdbg("Error in sector chain at %d, ret : %d\n", readwrite.logsector, ret);
+									tmp = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
+									if (tmp < 0) {
```
sangam-swami, line 1240: instead of tmp, can we use ret only and assign ret = o or -ENOENT in this case

Taejun-Kwon (commented):
This is wrong, instead of result value, this should return one of errno.

amogh-samsung:
The problem currently faced is: -
ret = FS_IOCTL() returns the number of bytes read for BIOC_READSECT command.
This byte count is held in ret variable which may be returned by smartfs_finddirentry().
However, smartfs_finddirentry is not expected to return values other than {OK, error codes (-ve)}

So, we should make sure that the function does not return the byte count of BIOC_READSECT.
We will hold the return value in tmp variable and if negative, copy the value in ret and return the same.

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1163,8 +1164,8 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 				/* Read the next directory in the chain */
 
 				smartfs_setbuffer(&readwrite, dirsector, 0, fs->fs_llformat.availbytes, (uint8_t *)fs->fs_rwbuffer);
-				ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
```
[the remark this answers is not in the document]
amogh-samsung: Updated the same, thanks

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1163,8 +1164,8 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
...
-				if (ret < 0) {
+				tmp = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
```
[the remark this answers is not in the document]
amogh-samsung: Updated and commented below, please review.

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Updated for proper return values

os/fs/smartfs/smartfs_utils.c
amogh-samsung: Updated for return values in case IOCTL fails, please review

Taejun-Kwon (commented):
problem is ret is returned after FS_IOCTL, so you should find somewhere we can add below,
'ret = ok'

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1053,6 +1053,7 @@ int smartfs_unmount(struct smartfs_mountpt_s *fs)
 int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *direntry, const char *relpath)
 {
 	int ret = -ENOENT;
+	int tmp;
```
jeongchanKim, line 1056: Please rename the variable name which can be easily undestood.

Taejun-Kwon:
I think title & description should be changed.

PR #4991, opened 2021-06-18
board/common: Fix string comparison length for flash partition

amogh-samsung:
- "timezone," string of length 9 should be matched with incoming
  partition type string but strncmp checks for string of length 10.
- Hence, the partition types never match and timzeone romfs mounting
  never takes place.
- Change comparison length in strncmp for "timezone," from 10 to 9

### DIFF
--- a/os/board/common/partitions.c
+++ b/os/board/common/partitions.c
@@ -114,7 +114,7 @@ static int type_specific_initialize(FAR struct mtd_dev_s *mtd_part, int partno, 
 	}
 #endif
 #ifdef CONFIG_LIBC_ZONEINFO_ROMFS
-	else if (!strncmp(types, "timezone,", 10)) {
+	else if (!strncmp(types, "timezone,", 9)) {
 		do_ftlinit = true;
 		save_timezone_partno = true;
 	}


os/board/common/partitions.c
```diff
@@ -114,7 +114,7 @@ static int type_specific_initialize(FAR struct mtd_dev_s *mtd_part, int partno,
 	}
 #endif
 #ifdef CONFIG_LIBC_ZONEINFO_ROMFS
-	else if (!strncmp(types, "timezone,", 10)) {
+	else if (!strncmp(types, "timezone,", 9)) {
```
dineshdas-samsung, line 117: This seems valid.

deepaksrma (commented):
LGTM

pillip8282 (approved):
lgtm

os/board/common/partitions.c
sunghan-chang: "timezone,a" will be return true. Is it ok?

sunghan-chang:
@amogh-samsung I don't know why this is necessary. This commit says "fix 10 to 9". But that's all. Please fill more description.

deepaksrma:
> @amogh-samsung I don't know why this is necessary. This commit says "fix 10 to 9". But that's all. Please fill more description.

we made these changes because without fixing this, we are getting error during mounting the zoneinfo.img.
error was:
`mount: ERROR: Failed to find block driver /dev/mtdblock0
`
after digging into this error, we found that the problem lies in the parttitions.c
like:

```
ifdef CONFIG_MTD_FTL
        bool do_ftlinit = false;
        if (!**strncmp(types, "ftl,", 4)**
#ifdef CONFIG_BINARY_MANAGER
        || !**strncmp(types, "kernel,", 7)**
#endif
        ) {
                do_ftlinit = true;
        }
#ifdef CONFIG_FS_ROMFS
        else if (!**strncmp(types, "romfs,", 6)**) {
                do_ftlinit = true;
                save_romfs_partno = true;
        }
#endif
```

@amogh-samsung as sunghan said correctly, please mention the error in the commit msg.

amogh-samsung:
> > @amogh-samsung I don't know why this is necessary. This commit says "fix 10 to 9". But that's all. Please fill more description.
> 
> we made these changes because without fixing this, we are getting error during mounting the zoneinfo.img.
> error was:
> `mount: ERROR: Failed to find block driver /dev/mtdblock0 `
> after digging into this error, we found that the problem lies in the parttitions.c
> please see below for more clarification, with below code, we can deduce that passed parameter in strncmp for timezone is not correct.
> 
> ```
> ifdef CONFIG_MTD_FTL
>         bool do_ftlinit = false;
>         if (!strncmp(types, "ftl,", 4)
> #ifdef CONFIG_BINARY_MANAGER
>         || !strncmp(types, "kernel,", 7)
> #endif
>         ) {
>                 do_ftlinit = true;
>         }
> #ifdef CONFIG_FS_ROMFS
>         else if (!strncmp(types, "romfs,", 6)) {
>                 do_ftlinit = true;
>                 save_romfs_partno = true;
>         }
> #endif
> #ifdef CONFIG_LIBC_ZONEINFO_ROMFS
>         else if (!strncmp(types, "timezone,", 10)) {
>                 do_ftlinit = true;
>                 save_timezone_partno = true;
>         }
> #endif
> ```
> 
> @amogh-samsung as sunghan said correctly, please mention the error in the commit msg.

A more accurate description of the error would be the inability to match the partition type "timezone," in CONFIG_PART_LIST with any given partition in board/common/partitions.c

This is because "timezone," -> has a length of 9 including the comma, but strncmp looks for one extra character (10 currently) so the match is never found.
Hence, bring it down to 9 for a match to be possible.

Will add the same in the commit description.

sunghan-chang:
@amogh-samsung @deepaksrma Thank you for explanation. Could you write what you say in the commit description? Then I will merge this.

sunghan-chang (approved):

PR #4993, opened 2021-06-21
apps/system/utils/tzselect: change the overwrite value in setenv

deepaksrma:
- in utils_tzselect.c, change the overwrite value as 1 in setenv() to select the TZ multiple times because if overwrite value is 0 then we can select the TZ only once, again overwriting is not allowed.

### DIFF
--- a/apps/system/utils/utils_tzselect.c
+++ b/apps/system/utils/utils_tzselect.c
@@ -44,6 +44,8 @@
 #define MAX_FILE_NAEM_LEN    10
 #define MAX_PATH_LEN         50
 
+#define TZ_SELECT_MULTIPLE_TIMES 1
+
 static char file_name[MAX_FILE_SIZE][MAX_FILE_NAEM_LEN];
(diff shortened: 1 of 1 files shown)


apps/system/utils/utils_tzselect.c
```diff
@@ -120,7 +120,7 @@ int utils_tzselect(int argc, char **args)
 	strncpy(path, TZDIR, MAX_PATH_LEN);
 	ret = search_tz(path);
 	if (ret) {
-		setenv("TZ", path, 0);
+		setenv("TZ", path, 1);
```
sunghan-chang, line 123: How about using definition to know what '1' mean?
```c
#define TZ_ENV_MULTIPLE_SELECTION 1

setenv("TZ", path, TZ_xxxx);
```

cf. please change the definition name with better naming.

deepaksrma: done

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #4996, opened 2021-06-22
apps/system/utils_date.c : Add timezone information in date command.

jeongchanKim:
Add timezone information in date command like below.
Dec 06 00:00:01 2011 Tue UTC

### DIFF
--- a/apps/system/utils/utils_date.c
+++ b/apps/system/utils/utils_date.c
@@ -119,7 +119,7 @@ static inline int day_of_week(FAR const char *abbrev)
 
 static inline int date_showtime(void)
 {
-	static const char format[] = "%b %d %H:%M:%S %Y %a";
+	static const char format[] = "%b %d %H:%M:%S %Y %a %Z";
 	struct timespec ts;
 	struct tm tm;
 	char timbuf[MAX_TIME_STRING];
--- a/apps/system/utils/utils_tzselect.c
+++ b/apps/system/utils/utils_tzselect.c
@@ -99,6 +99,7 @@ bool search_tz(char *dirpath)
 			break;
 		}
 	}
+	printf("\n");
 
 	strncat(dirpath, "/", MAX_PATH_LEN);
 	if (1 <= num && num <= folder_size) {
--- a/lib/libc/time/lib_strftime.c
+++ b/lib/libc/time/lib_strftime.c
@@ -156,6 +156,7 @@ static const char *const g_monthname[12] = {
  *   %t     A tab character. (SU)
  *   %y     The year as a decimal number without a century (range 00 to 99).
  *   %Y     The year as a decimal number including the century.
+ *   %Z     The timezone name or abbreviation, or by no bytes if no timezone information exists.
  *   %%     A literal '%' character.
  *
  * Returned Value:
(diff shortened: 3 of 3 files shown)


apps/system/utils/utils_date.c
```diff
@@ -141,7 +141,24 @@ static inline int date_showtime(void)
 	/* Show the current time in the requested format */
 
 	(void)strftime(timbuf, MAX_TIME_STRING, format, &tm);
-	printf("%s\n", timbuf);
+	printf("%s ", timbuf);
+
+	char *zoneinfo = getenv("TZ");
+	if (zoneinfo == NULL) {
+		printf("UTC\n");
+	} else {
```
sangam-swami, line 149: is it tested on different timezone

jeongchanKim: Sure. I tested.

sangam-swami (dismissed):
Looks good

apps/system/utils/utils_date.c
```diff
@@ -141,7 +141,24 @@ static inline int date_showtime(void)
 	/* Show the current time in the requested format */
 
 	(void)strftime(timbuf, MAX_TIME_STRING, format, &tm);
-	printf("%s\n", timbuf);
+	printf("%s ", timbuf);
+
+	char *zoneinfo = getenv("TZ");
```
kishore-sn, line 146: It would be good to declare all variables at start of function.

apps/system/utils/utils_date.c
```diff
@@ -141,7 +141,24 @@ static inline int date_showtime(void)
...
+	if (zoneinfo == NULL) {
+		printf("UTC\n");
+	} else {
+		/* zoneinfo can be like "/etc/zoneinfo/XXX".
+		 * We only need XXX for showing zoneinfo. */
```
kishore-sn, line 151: Below logic might not work if the zoneinfo string is not "/etc/zoneinfo/XXX"
If it is just "XXX" then, it will not be printed because we are checking slash_cnt as 2.

thapav: Yes, even if it is later changed to a different format, it will be problematic. (xx/xx/etc/zoneinfo/XXX)
May be you can check for zoneinfo and extract the next string after it.

jeongchanKim: Yes Right. I will change. Thanks.

sungkyu0 (approved):
LGTM

thapav (approved):
LGTM

apps/system/utils/utils_date.c
```diff
@@ -141,7 +142,28 @@ static inline int date_showtime(void)
 	/* Show the current time in the requested format */
 
 	(void)strftime(timbuf, MAX_TIME_STRING, format, &tm);
-	printf("%s\n", timbuf);
+	printf("%s ", timbuf);
+
+	zoneinfo = getenv("TZ");
```
sunghan-chang, line 147: Is this right to add here? not in localtime? How do you think?

sunghan-chang: How about adding %Z support in strftime?
Because POSIX strftime supports.

sunghan-chang:
```
   Select :
    43Set time zone : /etc/zoneinfo/MET
    TASH>>date
    Dec 06 01:00:09 2011 Tue MET
```
Could you make line change for "43" and "Set time zone"?

jeongchanKim:
> d you make line change

Modified. Thanks.

sunghan-chang:
Don't you need to add timezone config to prevent getting invalid env?

jeongchanKim:
> Don't you need to add timezone config to prevent getting invalid env?

I added the condition for LIBC_LOCALTIME. Thanks.

sunghan-chang (approved):

PR #4997, opened 2021-06-22
external/dhcps: refactoring dhcps message queue

pillip8282:
refactoring dhcps message queue

### DIFF
--- a/external/dhcpd/dhcpd_lwip.c
+++ b/external/dhcpd/dhcpd_lwip.c
@@ -88,7 +88,6 @@
 #define DHCP_CLIENT_PORT         68
 
 static pthread_t g_dhcpd_tid = -1;
-static int g_dhcpd_term = 0;
 
 struct dhcp_join_data {
 	char *intf;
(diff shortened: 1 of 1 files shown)


external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+		if (mq_stat.mq_curmsgs == 0) {
+			nvdbg("dhcps dhcps msg queue is empty\n");
+			return;
+		}
+		int nbytes = mq_timedreceive(mq, msg, DHCPD_MQ_LEN, 0, &ts);
+		if (nbytes == -1) {
+			if (errno == ETIMEDOUT) {
+				break;
+			} else {
+				nvdbg("dhcps empty dhcps message queue errno %d\n", errno);
```
davidfather, line 166: Why don't you print the msg received here?

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+static void _empty_mq(mqd_t mq)
+{
+	nvdbg("dhcps empty dhcps message queue\n");
+	char msg[DHCPD_MQ_LEN];
+	struct timespec ts;
+	int status = clock_gettime(CLOCK_REALTIME, &ts);
+	if (status != 0) {
+		ndbg("dhcps ERROR clock_gettime failed\n");
+	}
+	ts.tv_nsec += 5000000; // 5 milli seconds
```
davidfather, line 153: You should set ts.tv_sec to zero as well.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
 	struct mq_attr attr;
 	attr.mq_maxmsg = DHCPD_MQ_MAX_LEN;
 	attr.mq_msgsize = DHCPD_MQ_LEN;
 	attr.mq_flags = 0;
 	attr.mq_curmsgs = 0;
 
+	*mq = mq_open(DHCPD_MQ_NAME, O_RDWR | O_CREAT, 0666, &attr);
+	if (*mq == (mqd_t)ERROR) {
+		ndbg("dhcps Failed to open mq\n");
+	}
```
davidfather, line 185: You should add 'return' here.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+}
+
+static void _empty_mq(mqd_t mq)
+{
+	nvdbg("dhcps empty dhcps message queue\n");
+	char msg[DHCPD_MQ_LEN];
+	struct timespec ts;
+	int status = clock_gettime(CLOCK_REALTIME, &ts);
+	if (status != 0) {
+		ndbg("dhcps ERROR clock_gettime failed\n");
```
jeongchanKim, line 151: Is it okay to continue the below when fail to get time?

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+static int _send_mqmsg(mqd_t mq, char *msg, size_t msglen, int prio)
 {
-	dhcp_join_data_s *data = (dhcp_join_data_s *)arg;
-	int ret = ERROR;
-	int sockfd = 0;
-	struct req_lwip_data req;
+	nvdbg("dhcps send dhcps msg size %d\n", msglen);
+	while (1) {
+		int mq_ret = mq_send(mq, msg, msglen, prio);
+		if (mq_ret == -1) {
```
j11591, line 110: How about changing -1 and 0 to ERROR and OK?

external/dhcpd/dhcpd_lwip.c
```diff
@@ -263,31 +336,26 @@ int dhcp_server_stop(char *intf)
...
+	nvdbg("dhcps send terminate msg\n");
+	char msg[DHCPD_MQ_LEN] = {0}; // terminate thread message
+	msg[0] = 0;
+	int mq_ret = _send_mqmsg(md, msg, DHCPD_MQ_LEN, 100);
 	if (mq_ret == 0) {
+		nvdbg("dhcps send success\n");
 		pthread_join(g_dhcpd_tid, NULL);
 	} else {
-		ndbg("send mq fail (errno %d)\n", errno);
+		ndbg("dhcps send mq fail\n");
```
davidfather, line 351: What if the termination message is not delivered?

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+	}
+	ts.tv_nsec += 5000000; // 5 milli seconds
+	while (1) {
+		struct mq_attr mq_stat;
+		mq_getattr(mq, &mq_stat);
+		if (mq_stat.mq_curmsgs == 0) {
+			nvdbg("dhcps dhcps msg queue is empty\n");
+			return;
+		}
+		int nbytes = mq_timedreceive(mq, msg, DHCPD_MQ_LEN, 0, &ts);
```
jeongchanKim, line 161: In every loop, it consumes the time. So most of the cases are okay, but sometimes it cannot wait 5ms accurately.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+	nvdbg("dhcps create queue\n");
 	struct mq_attr attr;
 	attr.mq_maxmsg = DHCPD_MQ_MAX_LEN;
 	attr.mq_msgsize = DHCPD_MQ_LEN;
 	attr.mq_flags = 0;
 	attr.mq_curmsgs = 0;
 
+	*mq = mq_open(DHCPD_MQ_NAME, O_RDWR | O_CREAT, 0666, &attr);
+	if (*mq == (mqd_t)ERROR) {
+		ndbg("dhcps Failed to open mq\n");
```
jeongchanKim, line 184: How about printing the error number also for failure case?

external/dhcpd/dhcpd_lwip.c
pillip8282: this message should be delivered.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -263,31 +336,26 @@ int dhcp_server_stop(char *intf)
...
-
-	mqd_t md = mq_open(DHCPD_MQ_NAME, O_RDWR | O_CREAT, 0666, &attr);
-	if (md == (mqd_t)ERROR) {
-		ndbg("Failed to open mq\n");
-	}
-	char msg[2] = {0, 0}; // terminate thread message
-	int mq_ret = mq_send(md, msg, 2, 100);
+	nvdbg("dhcps send terminate msg\n");
+	char msg[DHCPD_MQ_LEN] = {0}; // terminate thread message
+	msg[0] = 0;
```
jeongchanKim, line 345: `msg[0]` is already filled with 0 from line 344.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,113 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+{
+	dhcp_join_data_s *data = (dhcp_join_data_s *)arg;
+	int ret = ERROR;
+	int sockfd = 0;
+	struct req_lwip_data req;
+	mqd_t md = (mqd_t)-1;
+
+	_create_mq(&md);
+
+	_empty_mq(md);
```
jeongchanKim, line 207: If `_create_mq()` fails, we cannot do `_empty_mq()` anymore.
So I think we'd better check the return of `_create_mq()` before calling `_empty_mq()`.

external/dhcpd/dhcpd_lwip.c
pillip8282: If there're many messages then mq_timedreceive will return message immediately. So I think 5ms is ok. and 5ms doesn't affect to system performance. If you have any idea please let me know.

davidfather: I also think 5ms is too short to read all pending messages.
Otherwise, why don't you move the timer setting logic into the loop.

jeongchanKim: I also think we'd better set the interval in the loop.

external/dhcpd/dhcpd_lwip.c
pillip8282: if _create_mq() failed then system will be restarted. because without dhcp server application can't be succeeded. so I think reboot system or adding assert() would be needed.

external/dhcpd/dhcpd_lwip.c
davidfather: I agree. Let's go with this.

external/dhcpd/dhcpd_lwip.c
pillip8282: it only returns when ETIMEOUT is returned. if mq_timedreceive() returns message than it tries to read mq message again.

jeongchanKim: As discussed on-line, please move time setting into the loop.

external/dhcpd/dhcpd_lwip.c
davidfather: I agree. If the mq creation fails, intended assert is necessary.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,125 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+
+	while (1) {
+		struct mq_attr mq_stat;
+		mq_getattr(mq, &mq_stat);
+		if (mq_stat.mq_curmsgs == 0) {
+			nvdbg("dhcps dhcps msg queue is empty\n");
+			return;
+		}
+
+		struct timespec ts;
```
jeongchanKim, line 157: `ts` should be initialized.

external/dhcpd/dhcpd_lwip.c
```diff
@@ -103,29 +102,125 @@ typedef struct dhcp_join_data dhcp_join_data_s;
...
+			ndbg("dhcps critical error %d %d\n", mq_ret, msglen);
+		}
+	}
+	return 0;
+}
+
+static void _empty_mq(mqd_t mq)
+{
+	nvdbg("dhcps empty dhcps message queue\n");
+	char msg[DHCPD_MQ_LEN];
```
j11591, line 147: msg is not initialized.

pillip8282: Does buffer have to be initialized before read it?

external/dhcpd/dhcpd_lwip.c
pillip8282: Does ts should be initialized before pass it clock_gettime? I don't think clock_gettime reads ts.

jeongchanKim: Oh right. It looks okay. Thanks.

davidfather (approved):
Looks good to me now.

PR #4999, opened 2021-06-22
external/dhcps: add log to dhcps event handler

pillip8282:
add log to dhcps event handler

### DIFF
--- a/external/dhcpd/dhcpd_lwip.c
+++ b/external/dhcpd/dhcpd_lwip.c
@@ -108,10 +108,11 @@ static int _send_mqmsg(mqd_t mq, char *msg, size_t msglen, int prio)
 	while (1) {
 		int mq_ret = mq_send(mq, msg, msglen, prio);
 		if (mq_ret == ERROR) {
-			if (errno == EAGAIN || errno == EINTR) {
+			int err_no = get_errno();
+			if (err_no == EINTR || err_no == EAGAIN) {
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


external/dhcpd/dhcpd_lwip.c
```diff
@@ -129,13 +129,15 @@ static int _recv_mqmsg(mqd_t mq, char *msg, size_t msglen, int *prio)
 			if (errno == EAGAIN && errno == EINTR) {
 				continue;
 			}
-			ndbg("dhcps _recv_mqmsg error %d\n", errno);
+			ndbg("dhcps error %d\n", errno);
```
jeongchanKim, line 132: How about like below? Then we can reduce the time.
Because `errno` tries to read the error number from tcb on every time.
```
int err_no = get_errno();
if (err_no == EAGAIN && err_no == EINTR) {
    continue;
}
ndbg("xxx %d\n", err_no);
```

external/dhcpd/dhcpd_lwip.c
```diff
@@ -129,13 +129,15 @@ static int _recv_mqmsg(mqd_t mq, char *msg, size_t msglen, int *prio)
 			if (errno == EAGAIN && errno == EINTR) {
```
jeongchanKim, line 129: Please check the `&&` condition. You want to check with `||`?

davidfather (approved):
Looks good to me.

PR #5004, opened 2021-06-23
mq, socket close fail log

jeongchanKim:


os/net/lwip/src/api/sockets.c
```diff
@@ -393,7 +393,12 @@ struct lwip_sock *get_socket(int s, pid_t pid)
 	list = tcb->group->tg_socketlist;
 	sock = (struct lwip_sock *)list.sl_sockets[s].sock;
 
-	if (!sock || !sock->conn) {
+	if (!sock) {
```
sunghan-chang, line 396: Is this enough conditional? Don't you need to compare to `s`?

os/net/lwip/src/api/sockets.c
```diff
@@ -393,7 +393,12 @@ struct lwip_sock *get_socket(int s, pid_t pid)
...
+		LWIP_DEBUGF(SOCKETS_DEBUG, ("get_socket(%d): Socket is not in this thread's group.\n", s + LWIP_SOCKET_OFFSET));
+		set_errno(EBADF);
```
pillip8282, line 398: I think why sock is null would have various reason. So just logging  "sock is null" is enough

os/net/lwip/src/api/sockets.c
jeongchanKim: It tries to get the socket from its task_group based on 's'. So I think it is enough.

os/net/lwip/src/api/sockets.c
jeongchanKim: Modified. Thanks.

pillip8282 (approved):
lgtm

PR #5006, opened 2021-06-23
arch/arm/src/amebad: Update serial_api

AdamZhang0124:
This commit updates serial_init process and removes the redundant code.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_serial.c
+++ b/os/arch/arm/src/amebad/amebad_serial.c
@@ -255,6 +255,7 @@ static char g_uart2txbuffer[CONFIG_UART2_TXBUFSIZE];
 #define RTL8721D_UART0_IRQ	(50)
 #define RTL8721D_UART1_IRQ	(51)
 #define RTL8721D_UART_LOG_IRQ	(19)
+#define RTL8721D_UARTLP_IRQ	(25)
 
 #ifdef CONFIG_RTL8721D_UART0
 static struct rtl8721d_up_dev_s g_uart0priv = {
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -417,7 +418,6 @@ static int rtl8721d_up_setup(struct uart_dev_s *dev)
 	serial_baud(sdrv[uart_index_get(priv->tx)], priv->baud);
 	serial_format(sdrv[uart_index_get(priv->tx)], priv->bits, priv->parity, priv->stopbit);
 	serial_set_flow_control(sdrv[uart_index_get(priv->tx)], priv->FlowControl, priv->rts, priv->cts);
-	serial_enable(sdrv[uart_index_get(priv->tx)]);
```
jeongchanKim, line 420: The only difference in changes is removing this line. Other changes are redundant code or tidy up the code. 
Is it okay we don't have this line?

AdamZhang0124: Yes, the corresponding peripheral clock functions are enabled in the bootloader, therefore it is redundant to enable again at here. Thanks!

jeongchanKim: Ok. Thanks for checking. I will approve it.

jeongchanKim (approved):
LGTM

sunghan-chang:
Because we use uart1 on internal usages and it's much important peri, let's check first.

sunghan-chang (approved):

PR #5007, opened 2021-06-24
os/driver/wireless/realtek: fix imxrt1020 build error

pillip8282:
fix imxrt1020 build error


os/drivers/wireless/realtek/wifi_util_interface/rtk_drv.c
```diff
@@ -53,7 +53,7 @@
 		vddbg(RTKDRV_TAG "%s:%d\n", __FILE__, __LINE__); \
 	} while (0)
 
-
+struct netdev *g_rtk_netdev;
```
jeongarmy, line 56: How about adding `static` if it is used in this file only?

pillip8282: I'll modify it later.

sunghan-chang (approved):

PR #5009, opened 2021-06-24
rtl8721csm: Update trustzone images for FTL stroage encryption

jeongarmy:
Added APIs for FTL storage encryption
- uint32_t wrap_ameba_ftl_secure_init(uint32_t key_addr, void *in_func);
- uint32_t wrap_ameba_ftl_save_to_storage(uint8_t* tmp_buff, void *pdata, uint16_t offset, uint16_t size);
- uint32_t wrap_ameba_ftl_load_from_storage(void *pdata, uint16_t offset, uint16_t size);


sunghan-chang:
This is merged into PR #5034 so that let's close this.

PR #5010, opened 2021-06-24
apps/testcase/kernel: add new IRQ TC

deepaksrma:
- add new IRQ TC to test fin_wait().

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/Make.defs
+++ b/apps/examples/testcase/le_tc/kernel/Make.defs
@@ -152,6 +152,9 @@ endif
 ifeq ($(CONFIG_TC_KERNEL_MEMORY_SAFETY),y)
   CSRCS += tc_memory_safety.c
 endif
+ifeq ($(CONFIG_TC_KERNEL_IRQ),y)
+  CSRCS += tc_irq.c
+endif
 ifeq ($(CONFIG_ITC_KERNEL_ENVIRON),y)
   CSRCS += itc_environ.c
 endif
--- a/apps/examples/testcase/le_tc/kernel/kernel_tc_main.c
+++ b/apps/examples/testcase/le_tc/kernel/kernel_tc_main.c
@@ -191,6 +191,9 @@ int tc_kernel_main(int argc, char *argv[])
 	memory_safety_main();
 #endif
 
+#ifdef CONFIG_TC_KERNEL_IRQ
+	irq_main();
+#endif
 #ifdef CONFIG_ITC_KERNEL_ENVIRON
 	itc_environ_main();
 #endif
--- a/apps/examples/testcase/le_tc/kernel/tc_irq.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_irq.c
@@ -1,0 +1,83 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @file tc_irq.c
+
+/// @brief Test Case Example for fin_wait API
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <tinyara/timer.h>
+#include <tinyara/irq.h>
+#include "tc_internal.h"
+
+# define TIMER_DEVNAME  "/dev/timer0"
+# define INTERVAL	100
+
+/**
+ * @fn                  :tc_irq
+ * @brief               :test the fin_wait using timer
+ * @scenario            :fin_wait waits for kernel irq event, so here we will generate the irq using timer.
+ * API's covered        :fin_wait
+ * Preconditions        :none
+ * Postconditions       :none
+ * @return              :void
+ */
+
+static void tc_irq_fin_wait(void)
+{
+	struct timer_notify_s notify;
+	notify.arg   = NULL;
+	notify.pid   = (pid_t)getpid();
+	char path[256];
+	int ret = 0;
+
+	snprintf(path, _POSIX_PATH_MAX, TIMER_DEVNAME);
+	int fd = open(path, O_RDONLY);
+	TC_ASSERT_GEQ("open", fd, 0);
+
+	ret = ioctl(fd, TCIOC_NOTIFICATION, (unsigned long)((uintptr_t)&notify));
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	ret = ioctl(fd, TCIOC_SETTIMEOUT, INTERVAL);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	/* Start the timer */
+	ret = ioctl(fd, TCIOC_START, 0);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	ret = fin_wait();
+	TC_ASSERT_EQ_CLEANUP("fin_wait", ret, 0, close(fd));
+
+	/* stop the timer */
+	ret = ioctl(fd, TCIOC_STOP, 0);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	close(fd);
+
+	TC_SUCCESS_RESULT();
+}
+
+int irq_main(void)
+{
+	tc_irq_fin_wait();
+
+	return 0;
+}
--- a/apps/examples/testcase/le_tc/kernel/Kconfig
+++ b/apps/examples/testcase/le_tc/kernel/Kconfig
@@ -47,6 +47,7 @@ config TC_KN_ALL
 	select TC_KERNEL_TIMER
 	select TC_KERNEL_UMM_HEAP
 	select TC_KERNEL_WORK_QUEUE if SCHED_WORKQUEUE
+	select TC_KERNEL_IRQ if TIMER
 	select ITC_KERNEL_ENVIRON
 	select ITC_KERNEL_LIBC_PTHREAD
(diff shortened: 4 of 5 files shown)


apps/examples/testcase/le_tc/kernel/kernel_tc_main.c
```diff
@@ -191,6 +191,9 @@ int tc_kernel_main(int argc, char *argv[])
 	memory_safety_main();
 #endif
 
+#ifdef CONFIG_TIMER
```
sunghan-chang, line 194: Need to check the config.
```
CONFIG_TIMER -> CONFIG_TC_KERNEL_IRQ
```

apps/examples/testcase/le_tc/kernel/Make.defs
```diff
@@ -152,6 +152,9 @@ endif
 ifeq ($(CONFIG_TC_KERNEL_MEMORY_SAFETY),y)
   CSRCS += tc_memory_safety.c
 endif
+ifeq ($(CONFIG_TIMER),y)
```
sunghan-chang, line 155: same as previous

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,81 @@
...
+ * @fn                  :tc_irq
+ * @brief               :test the fin_wait using timer
+ * @scenario            :none
+ * API's covered        :fin_wait
+ * Preconditions        :none
+ * Postconditions       :none
+ * @return              :void
+ */
+
+#ifdef CONFIG_TIMER
```
sunghan-chang, line 44: We should make a dependency with timer in Kconfig, not here. Because even user enables irq tc, because of missing timer enabled, nothing is executed. It makes user confused.
And when we enable irq tc config and disable timer config, this makes compilation error.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,81 @@
...
+
+	ret = fin_wait();
+	TC_ASSERT_EQ_CLEANUP("fin_wait", ret, 0, close(fd));
+
+	TC_SUCCESS_RESULT();
+}
+
+int irq_main(void)
+{
+	tc_fin_wait();
```
sunghan-chang, line 77: Let's use `tc_irq_fin_wait` for the same naming policy.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,81 @@
...
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @file tc_irq.c
+
+/// @brief Test Case Example for Errno API
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <string.h>
```
sunghan-chang, line 24: This looks unnecessary.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,81 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @file tc_irq.c
+
+/// @brief Test Case Example for Errno API
```
sunghan-chang, line 21: Please check the description.

apps/examples/testcase/le_tc/kernel/tc_irq.c
deepaksrma: thanks for pointing it out. corrected now. please check now.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+{
+	struct timer_notify_s notify;
+	notify.arg   = NULL;
+	notify.pid   = (pid_t)getpid();
+	int intval = 100;
+	char path[256];
+	int dev = 0;
+	int ret = 0;
+
+	snprintf(path, _POSIX_PATH_MAX, TIMER_DEVNAME, dev);
```
jeongchanKim, line 53: If we test the `timer0` only, how about defining `TIMER_DEVNAME` as "/dev/timer0`"?

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+ * Postconditions       :none
+ * @return              :void
+ */
+
+static void tc_irq_fin_wait(void)
+{
+	struct timer_notify_s notify;
+	notify.arg   = NULL;
+	notify.pid   = (pid_t)getpid();
+	int intval = 100;
```
jeongchanKim, line 48: Please declare this value through `#define INTERVAL`.
Then we can test with different interval easily.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	/* Start the timer */
+	ret = ioctl(fd, TCIOC_START, 0);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	ret = fin_wait();
+	TC_ASSERT_EQ_CLEANUP("fin_wait", ret, 0, close(fd));
+
+	TC_SUCCESS_RESULT();
```
jeongchanKim, line 70: For successful case, we need `close(fd)` also.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @file tc_irq.c
+
+/// @brief Test Case Example for fin_wait API
+#include <tinyara/config.h>
+#include <stdio.h>
+#include <errno.h>
```
jeongchanKim, line 24: I think this is not needed.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+ * Preconditions        :none
+ * Postconditions       :none
+ * @return              :void
+ */
+
+static void tc_irq_fin_wait(void)
+{
+	struct timer_notify_s notify;
+	notify.arg   = NULL;
+	notify.pid   = (pid_t)getpid();
```
jeongchanKim, line 47: Please include `unistd.h` for using `getpid()`.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,78 @@
...
+#include <tinyara/timer.h>
+#include <tinyara/irq.h>
+#include "tc_internal.h"
+
+#  define TIMER_DEVNAME  "/dev/timer%d"
+
+/**
+ * @fn                  :tc_irq
+ * @brief               :test the fin_wait using timer
+ * @scenario            :none
```
jeongchanKim, line 36: Please explain the tc scenario also.

apps/examples/testcase/le_tc/kernel/tc_irq.c
deepaksrma: forgot to close. thanks for noticing it. corrected now.

apps/examples/testcase/le_tc/kernel/tc_irq.c
deepaksrma: @sunghan-chang i have added CONFIG_TIMER again because if we enable the CONFIG_TIMER then CONFIG_TC_KERNEL_IRQ would be enabled and written in .config but when we disable CONFIG_TIMER then CONFIG_TC_KERNEL_IRQ is still in the .config then it gives the compilation errors.
let me know your views.

apps/examples/testcase/le_tc/kernel/tc_irq.c
```diff
@@ -0,0 +1,81 @@
...
+	TC_ASSERT_GEQ("open", fd, 0);
+
+	ret = ioctl(fd, TCIOC_NOTIFICATION, (unsigned long)((uintptr_t)&notify));
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	ret = ioctl(fd, TCIOC_SETTIMEOUT, INTERVAL);
+	TC_ASSERT_GEQ_CLEANUP("ioctl", ret, 0, close(fd));
+
+	/* Start the timer */
+	ret = ioctl(fd, TCIOC_START, 0);
```
jeongchanKim, line 64: Is it okay not to stop the timer for both successful and failure cases?

deepaksrma: corrected now.

apps/examples/testcase/le_tc/kernel/tc_irq.c
sunghan-chang: @deepaksrma It's impossible to disable CONFIG_TIMER even CONFIG_TC_KERNEL_IRQ is enabled. When you disable CONFIG_TIMER, then CONFIG_TC_KERNEL_IRQ will be disable automatically. That's why we make a dependency with TIMER in Kconfig. I think you consider manual modification for .config. But it should not be. 
```
#
# Automatically generated file; DO NOT EDIT.
# TinyAra Configuration
#
```
This is the first sentence of defconfig and .config. We should use menuconfig to modify configs.

deepaksrma: `When you disable CONFIG_TIMER, then CONFIG_TC_KERNEL_IRQ will be disabled automatically.`
i was also expecting the same but this is not happening and that is my point.

once you enable the CONFIG_TIMER then CONFIG_TC_KERNEL_IRQ is also enabled and now suppose if i disable CONFIG_TIMER still i see `CONFIG_TC_KERNEL_IRQ=y` in .config, which should not have happened, then it is resulting into compilation errors.

P.S. I am not modifying .config manually, i am modifying with menuconfig only.

jeongchanKim: I'm not sure why you said like above.
When I test, if I disable the `CONFIG_TIMER`, then `CONFIG_TC_KERNEL_IRQ` is also disabled properly.
Please check again.

deepaksrma: CONFIG_TC_KERNEL_IRQ was not disabled properly because
i wrote `select TC_KERNEL_IRQ` 
instead of
`select TC_KERNEL_IRQ if TIMER`

now corrected and working fine.

sunghan-chang (approved):

PR #5011, opened 2021-06-25
apps/examples/rtc : Add rtc accuracy test example

jeongchanKim:
This example runs forever with printing the log on every 5mins.
With this log, we can check the time difference which rtc runs well or not.

### DIFF
--- a/apps/examples/rtc/rtc_main.c
+++ b/apps/examples/rtc/rtc_main.c
@@ -1,0 +1,222 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * apss/examples/rtc/rtc_main.c
+ *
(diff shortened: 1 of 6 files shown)


apps/examples/rtc/rtc_main.c
```diff
@@ -0,0 +1,150 @@
...
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int main(int argc, FAR char *argv[])
+#else
+int rtc_main(int argc, char *argv[])
+#endif
+{
+	printf("=== RTC Accuracy Test ===\n");
+	printf(" * This test will run forever !!! *\n");
```
sunghan-chang, line 147: Let's support "stop" like stack monitor.

apps/examples/rtc/rtc_main.c
```diff
@@ -0,0 +1,222 @@
...
+	/* Initialize the RTC time as 0 */
+	ret = ioctl(fd, RTC_SET_TIME, (unsigned long)&prev_time);
+	if (ret < 0) {
+		printf("ERROR : Fail to set RTC init time.\n");
+		close(fd);
+		return NULL;
+	}
+
+	sched_lock();
+	while (rtctest_started) {
```
sunghan-chang, line 123: When disable_signal config is enabled, this function is called with this variable 0. Is it ok?

apps/examples/rtc/rtc_main.c
```diff
@@ -0,0 +1,222 @@
...
+ * rtc_main
+ ****************************************************************************/
+
+#ifdef CONFIG_BUILD_KERNEL
+int main(int argc, FAR char *argv[])
+#else
+int rtc_main(int argc, char *argv[])
+#endif
+{
+#ifndef CONFIG_DISABLE_SIGNALS
```
sunghan-chang, line 166: Could you let me know why you cover with disable_signals config? not disable_pthread?

apps/examples/rtc/rtc_main.c
jeongchanKim: Modified. Thanks.

sunghan-chang (approved):

PR #5014, opened 2021-06-28
os: Change a type of CONFIG_BOARD_BUILD_DATE to string

jeongarmy:
This config can be values string with 0 like 000101 because it indicates build date which has a "YYMMDD" format.
So change a type from int to string.


jeongchanKim (approved):
LGTM

sangwon03 (approved):
LGTM

PR #5015, opened 2021-06-28
os/se/ameba: Increase Secure Context Size for TrustZone

ziyik:
- Increase default Secure Context Size for TrustZone (Ready for ED25519)


jeongchanKim (commented):
Could you leave comment in commit msg why you increase the stack size?

ziyik:
> Could you leave comment in commit msg why you increase the stack size?

I have modify the commit message and PR title for better understanding

sunghan-chang (approved):

PR #5020, opened 2021-06-30
build/tools/amebad: Add km0_appliccation asm and map

ziyik:
- Added asm and map of km0_appliccation for reference


jeongchanKim (approved):
LGTM

sunghan-chang:
@ziyik Could you let me know why this is necessary? To debug km0 bootloader?

sunghan-chang:
@ziyik They should be synced with km0 bin file. Right?

ziyik:
> @ziyik Could you let me know why this is necessary? To debug km0 bootloader?

This is nor for Bootloader, it is for km0_image2_all.bin.
Yes for debug purpose, in case hardfault location is in km0_image2_all.bin we can locate the hardfault location.

ziyik:
> @ziyik They should be synced with km0 bin file. Right?

Yes, it is sync with km0_image2_all.bin, I use the same version as km0_image2_all.bin.
In my local I generate km0_image2_all.bin again to ensure it is same as current version, and used it asm and map.

sunghan-chang (approved):

PR #5023, opened 2021-07-01
network/lwnl: modify lwnl handler

pillip8282:
modify lwnl handler

### DIFF
--- a/os/net/netmgr/netdev_lwnl.c
+++ b/os/net/netmgr/netdev_lwnl.c
@@ -85,11 +85,16 @@ int netdev_req_handle(const char *msg, size_t msg_len)
 {
 	lwnl_msg *lmsg = (lwnl_msg *)msg;
 	struct netdev *dev = NULL;
-	dev = (struct netdev *)nm_get_netdev(lmsg->name);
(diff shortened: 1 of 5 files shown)


os/net/netmgr/netdev_lwnl.c
```diff
@@ -85,13 +85,18 @@ int netdev_req_handle(const char *msg, size_t msg_len)
...
+
+	if (!strncmp((const char *)lmsg->name, LWNL_INTF_NAME, strlen(LWNL_INTF_NAME) + 1)) {
 		int *res = (int *)lmsg->result;
 		*res = _handle_common(lmsg);
 		return 0;
 	}
 
+	dev = (struct netdev *)nm_get_netdev(lmsg->name);
+	if (!dev) {
+		return -ENOSYS;
```
jeongchanKim, line 97: How about returning as `-ENODEV`?

pillip8282: actually according to manual page write() doesn't return either enodev and enosys. so I think enosys would be ok

an4967 (approved):
It works well

PR #5024, opened 2021-07-01
include/tinyara/reboot_reason.h : Add macro for reboot reason

jeongchanKim:
To write reboot reason, user should call prctl(PR_REBOOT_REASON_WRITE, code) and include both sys/prctl.h and tinyara/reboot_reason.h
For easy use, add macro.

### DIFF
--- a/os/include/tinyara/reboot_reason.h
+++ b/os/include/tinyara/reboot_reason.h
@@ -21,19 +21,28 @@
 /********************************************************************************************
  * Included Files
  ********************************************************************************************/
+#include <sys/prctl.h>
 
 typedef enum {
-	REBOOT_REASON_INITIALIZED	= 0,
-	REBOOT_SYSTEM_DATAABORT		= 1,	/* Data abort */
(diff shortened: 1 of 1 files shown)


os/include/tinyara/reboot_reason.h
```diff
@@ -21,19 +21,32 @@
...
+	REBOOT_UNKNOWN                  = 99,
 } reboot_reason_code_t;
 
+#define WRITE_REBOOT_REASON(x) do {                                       \
+					prctl(PR_REBOOT_REASON_WRITE, x); \
+				} while (0)                               \
+
+#define READ_REBOOT_REASON() do {                                     \
+					prctl(PR_REBOOT_REASON_READ); \
+				} while (0)                           \
```
sunghan-chang, line 46: This will give compilation error.
User will use like below:
```c
     int x = READ_REBOOT_REASON();
```
But it's actually,
```c
    int x = do { prctl(); } while(0);
```
It's wrong.

This should be
```h
#define READ_R_R() prctl()
```

sunghan-chang (approved):

PR #5025, opened 2021-07-01
Reboot Reason for Binary Recovery failure

jeongchanKim:


os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -81,6 +84,9 @@ static faultmsg_t g_prealloc_faultmsg[FAULTMSG_COUNT];
 static void binary_manager_reset_board(void)
 {
 #ifdef CONFIG_BOARDCTL_RESET
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	WRITE_REBOOT_REASON(REBOOT_SYSTEM_BINARY_RECOVERYFAIL);
```
sunghan-chang, line 88: This should be saved regardless of automatic board reset. Let's take out of the conditional.

jeongchanKim: For else case, the board will not be rebooted.
So if user wants to reboot the board, hw reset is required.
Then the reboot reason is marked as hw reset.
Thanks.

sunghan-chang: 1. That's the limitation of rtl board only.
2. Even reboot reason is changed in rtl board, logically we should save proper reboot reason regardless of the config.

jeongchanKim: Modified. Thanks.

sunghan-chang (approved):

PR #5028, opened 2021-07-02
arch/boardctl.c : Add reboot debug message

jeongchanKim:
Board Reboot is important event, so it is better to remain the log.

Signed-off-by: jeongchanKim <initialkjc@gmail.com>

### DIFF
--- a/os/arch/boardctl.c
+++ b/os/arch/boardctl.c
@@ -57,11 +57,14 @@
 
 #include <sys/types.h>
 #include <sys/boardctl.h>
+#include <debug.h>
+#include <sched.h>
 #include <stdint.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


os/arch/boardctl.c
```diff
@@ -143,6 +145,9 @@ int boardctl(unsigned int cmd, uintptr_t arg)
 	 * DEPENDENCIES:  Board logic must provide board_reset
 	 */
 	case BOARDIOC_RESET:
+		/* Add 100ms delay for flushing another logs like printf. */
+		up_mdelay(100);
```
sunghan-chang, line 149: Don't we need to do sched lock?

sunghan-chang (approved):

PR #5031, opened 2021-07-02
os/board/rtl8721csm: add connection fail reason code in softap hdl

gexuyan:
To check reason code meaning, please refer to Reason codes (IEEE 802.11-2007, 7.3.1.7, Table 7-22) in ieee802_11_defs.h


os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
```diff
@@ -1596,8 +1596,8 @@ static void wifi_ap_sta_assoc_hdl( char* buf, int buf_len, int flags, void* user
 #if defined(CONFIG_PLATFORM_TIZENRT_OS)
 	rtk_reason_t reason;
 	memset(&reason, 0, sizeof(rtk_reason_t));
-	if (strlen(buf) >= 17) {			  // bssid is a 17 character string
-		memcpy(&(reason.bssid), buf, 17); // Exclude null-termination
+	if (buf_len >= 6) {
```
sangam-swami, line 1599: why this check is needed?

os/board/rtl8721csm/src/component/common/api/wifi/wifi_util.c
```diff
@@ -1071,7 +1071,7 @@ void wext_wlan_indicate(unsigned int cmd, union iwreq_data *wrqu, char *extra)
 				else if(!memcmp(IW_EVT_STR_STA_ASSOC, extra, strlen(IW_EVT_STR_STA_ASSOC)))
 					wifi_indication(WIFI_EVENT_STA_ASSOC, wrqu->data.pointer, wrqu->data.length, 0);
 				else if(!memcmp(IW_EVT_STR_STA_DISASSOC, extra, strlen(IW_EVT_STR_STA_DISASSOC)))
-					wifi_indication(WIFI_EVENT_STA_DISASSOC, wrqu->addr.sa_data, sizeof(null_mac), 0);
+					wifi_indication(WIFI_EVENT_STA_DISASSOC, wrqu->addr.sa_data, sizeof(null_mac) + 2, 0);
```
sangam-swami, line 1074: can we use macro for "2" reason code ?

j11591:
We requested linkup and linkdown handler are called after connection not assoc similar to STA connection.
As we known, connected callback is called after EAPOL. is it right???
we hope join callback is called after EAPOL similar to STA connection.

os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
gexuyan: hi @sangam-swami, this is just to check buf is not empty, but don't do check also ok.

os/board/rtl8721csm/src/component/common/api/wifi/wifi_util.c
gexuyan: sure can

gexuyan:
> We requested linkup and linkdown handler are called after connection not assoc similar to STA connection.
> As we known, connected callback is called after EAPOL. is it right???
> we hope join callback is called after EAPOL similar to STA connection.

hi @j11591 , ok I see. Actually for STA mode, there are two callbacks, wifi_connected_hdl() and wifi_handshake_done_hdl(). wifi_connected_hdl() happens after assoc, and wifi_handshake_done_hdl() happens after EAPOL 4way handshake. Here we have two callbacks is because:  in WEP mode, there is no EAPOL handshake process. If we only set callback after EAPOL, then if we connect in WEP mode which doesn't go through EAPOL, there is no callback msg.
Currently for SoftAP mode, there is wifi_ap_sta_assoc_hdl() happens after Assoc, I can add another callback hdl to be called after SoftAP EAPOL. Am I understanding correctly? Thank you.

j11591:
> wifi_handshake_done_hdl

Is it impossible to use only one join callback that if softap is open or wep, callback is called after assoc, and if softap need EAPOL, then callback is called after EAPOL.

os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
thapav: Can we not hardcode for certain string length and check otherwise ?

os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
```diff
@@ -1617,8 +1617,10 @@ static void wifi_ap_sta_disassoc_hdl( char* buf, int buf_len, int flags, void* u
 #if defined(CONFIG_PLATFORM_TIZENRT_OS)
 	rtk_reason_t reason;
 	memset(&reason, 0, sizeof(rtk_reason_t));
-	if (strlen(buf) >= 17) { // bssid is a 17 character string
-		memcpy(&(reason.bssid), buf, 17);
+	if (buf_len >= 8) { //mac address len 6; reason code len 2
```
thapav, line 1620: Same as above. Maye be a #define for checking would be better

gexuyan:
will raise new pr for this issue

PR #5032, opened 2021-07-04
build/configs/usbrule.sh : Modify descriptions about supporting which…

jeongchanKim:
… boards

This file is not only for Artik05x, but also for stm32.

Signed-off-by: jeongchanKim <initialkjc@gmail.com>

### DIFF
--- a/build/configs/usbrule.sh
+++ b/build/configs/usbrule.sh
@@ -17,8 +17,8 @@
 #
 ###########################################################################
 #
-# File   : artik05x_usbrule.sh
-# Description : usb rule script for ARTIK 05X
+# File   : usbrule.sh
+# Description : usb rule script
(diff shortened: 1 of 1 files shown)


build/configs/usbrule.sh
```diff
@@ -17,8 +17,8 @@
 #
 ###########################################################################
 #
-# File   : artik05x_usbrule.sh
-# Description : usb rule script for ARTIK 05X
+# File   : usbrule.sh
+# Description : usb rule script for ARTIK 05X and STM32 Boards
```
sunghan-chang, line 21: This is working on imxrt as well. I think we had better remove `for artik05x......`.

sunghan-chang (approved):

PR #5043, opened 2021-07-08
apps/examples/le_tc/kernel/tc_mqueue.c : Add missing return statement

jeongchanKim:
In tc_mqueue_mq_timedsend_timedreceive_failurechecks, return statement is missed.
So duplicated mq_close prints the below ERRORs.
mq_close_group: ERROR: mqdes(xxxxx) is not in this thread's group.


apps/examples/testcase/le_tc/kernel/tc_mqueue.c
```diff
@@ -847,6 +847,7 @@ static void tc_mqueue_mq_timedsend_timedreceive_failurechecks(void)
 	TC_ASSERT_GEQ("timedsend_test", mq_close(g_timedsend_mqfd), 0);
 
 	TC_SUCCESS_RESULT();
+	return;
```
sunghan-chang, line 850: Don't you need to do unlink?

sunghan-chang (commented):
You should revert file permission to 644

apps/examples/testcase/le_tc/kernel/tc_mqueue.c
jeongchanKim: Added. Thanks.

sunghan-chang (approved):

PR #5046, opened 2021-07-12
Reboot Reason String

jeongchanKim:


os/arch/arm/src/common/up_initialize.c
```diff
@@ -171,7 +171,7 @@ void up_initialize(void)
 
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
 	up_reboot_reason_init();
-	lldbg("[Reboot Reason] : %d\n", up_reboot_reason_read());
+	lldbg("[Reboot Reason] : %s\n", GET_REBOOT_REASON_STR(up_reboot_reason_read()));
```
amogh-samsung, line 174: Could you explain the preference of using a macro here instead of the function name please?

jeongchanKim: We are using macros for other reboot reason usages.
So I think it is much better to match the usage.

PR #5048, opened 2021-07-12
board/rtl8721csm/bluetooth: Do not start adv after set adv parameters

Pangwawaa:
Do not start adv after set adv parameters


an4967 (approved):
LGTM

PR #5049, opened 2021-07-12
security/seclink: add hal test case

pillip8282:
add hal test case


external/stress_tool/st_perf.c
```diff
@@ -40,6 +40,22 @@
 /**
  * Inner Function
  */
+void _smoke_print_interim_result(st_smoke *smoke, st_elapsed_time *duration)
+{
+	char *title = smoke->func->tc_name;
```
jeongarmy, line 45: I think it is better to check whether parameters are NULL or not like below to avoid memory dereferencing.
```
if (!smoke || !smoke->func || duration) {
  return;
}
```

pillip8282: it's internal function, so smoke and smoke func can't be null.

an4967 (approved):
LGTM

PR #5050, opened 2021-07-12
apps/testcase/systemio/itc: add new UART TCs

deepaksrma:
- add new TCs to test UART rxavailable, txready and txempty APIs.

Signed-off-by: Deepak sharma <deepak.sh@samsung.com>


apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -774,6 +774,76 @@ static void itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p(void)
...
+	ret = iotbus_uart_stop(h_uart);
+	TC_ASSERT_EQ("iotbus_uart_stop", ret, IOTBUS_ERROR_NONE);
+
+	TC_SUCCESS_RESULT();;
+
+}
+
+/**
+* @testcase         itc_systemio_iotbus_uart_txready
+* @brief            To check if the tranmsit fifo is not full
```
abhishek-samsung, line 803: Please change tranmsit to transmit.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -774,6 +774,76 @@ static void itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p(void)
...
+
+	ret = iotbus_uart_stop(h_uart);
+	TC_ASSERT_EQ("iotbus_uart_stop", ret, IOTBUS_ERROR_NONE);
+
+	TC_SUCCESS_RESULT();
+}
+
+/**
+* @testcase         itc_systemio_iotbus_uart_txempty
+* @brief            To check if the tranmsit fifo is empty
```
abhishek-samsung, line 826: Please change tranmsit to transmit.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
+/****************************************************************************
```
abhishek-samsung, line 1: Please add a Samsung's copy right also.

jeongchanKim (commented):
Please split the commit as 2.
1st is for changing the framework and lib, 2nd is for tc.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: tcrxavailable
+ *
+ * Descripton:
```
dineshdas-samsung, line 72: Description =>typo

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -774,6 +774,76 @@ static void itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p(void)
...
+	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
+	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	ret = iotbus_uart_rxavailable(h_uart);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_rxavailable", ret, IOTBUS_ERROR_NONE, iotbus_uart_stop(h_uart));
+
+	ret = iotbus_uart_stop(h_uart);
+	TC_ASSERT_EQ("iotbus_uart_stop", ret, IOTBUS_ERROR_NONE);
+
+	TC_SUCCESS_RESULT();;
```
sangam-swami, line 797: remove extra ;

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+ *	TIOCS_AVAIL: check if there is data in rx fifo.
+ *	TIOCS_READY: check if tranmsit fifo is not full.
+ *	TIOCS_EMPTY: check if transmit fifo is empty.
+ *
+ * Input Parameters:
+ *   fd  - The 'fd' argument is an open file descriptor associated with a terminal.
+ *   cmd - ioctl argument.
+ *
+ * Returned Value:
+ *   Upon successful completion, 0 is returned. Otherwise, -1 is returned.
```
jeongchanKim, line 83: For failure case, negative errno will be returned. Isn't it?

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+
+	case TIOCS_READY:
+		ret = ioctl(fd, TIOCS_READY, 0);
+		break;
+
+	case TIOCS_EMPTY:
+		ret = ioctl(fd, TIOCS_EMPTY, 0);
+		break;
+	default:
+		ret = -EINVAL;
```
jeongchanKim, line 104: In `ioctl`, if there is no matched cmd, it returns `-ENOTTY` instead of `-EINVAL`.
Which is correct?

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: tcrxavailable
+ *
+ * Descripton:
+ *   calls the ioctl ased on cmd,
```
dineshdas-samsung, line 73: ased =>based?

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+ * Returned Value:
+ *   Upon successful completion, 0 is returned. Otherwise, -1 is returned.
+ *
+ ****************************************************************************/
+
+int tccheckfifo(int fd, int cmd)
+{
+	int ret;
+
+	switch (cmd) {
```
jeongchanKim, line 91: All cases just call the ioctl with the `cmd`.
If there is no other case for calling `ioctl` with another parameters,
how about just calling `ioctl(fd, cmd, 0);` instead of checking with switch-case?
Because in `ioctl`, it checks that `cmd` is valid or not.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+	case TIOCS_AVAIL:
+		ret = ioctl(fd, TIOCS_AVAIL, 0);
```
jeongarmy, line 93: On success, True(1) is returned by rtl8721d_up_rxavailable in amebad_serial.c.
This is different from description `Upon successful completion, 0 is returned. Otherwise, -1 is returned.` in line 83.
Please check and modify mismatched description or return value.

framework/src/iotbus/iotbus_uart.c
```diff
@@ -176,6 +177,69 @@ int iotbus_uart_stop(iotbus_uart_context_h hnd)
...
+	struct _iotbus_uart_s *handle;
+
+	if (!hnd || !hnd->handle) {
+		return IOTBUS_ERROR_INVALID_PARAMETER;
+	}
+
+	handle = (struct _iotbus_uart_s *)hnd->handle;
+	fd = handle->fd;
+
+	ret = tccheckfifo(fd, TIOCS_READY);
```
jeongchanKim, line 214: `tccheckfifo` returns the negative errno for failure case, but this api returns only `IOTBUS_ERROR_UNKNOWN`.
In this case, the reason why it fails is disappeared.

lib/libc/termios/lib_tccheckfifo.c
jeongarmy: Please check TIOCS_READY, TIOCS_EMPTY also.
The ioctl functions of them return true(1) by rtl8721d_up_txready, rtl8721d_up_txempty.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
+/****************************************************************************
+ * libc/termios/lib_tcrxavailable.c
```
jeongchanKim, line 2: Please modify the file name.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,109 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: tcrxavailable
```
jeongchanKim, line 70: Please modify the function name.

framework/src/iotbus/iotbus_uart.c
```diff
@@ -176,6 +177,69 @@ int iotbus_uart_stop(iotbus_uart_context_h hnd)
...
+	if (ret < 0) {
```
jeongarmy, line 215: How about using specific error code like IOTBUS_ERROR_DEBICE_NOT_READY for easier debugging?

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
jeongarmy, line 797: Please remove duplicated `;`

lib/libc/termios/lib_tccheckfifo.c
jeongarmy, line 73: based*

lib/libc/termios/lib_tccheckfifo.c
jeongarmy, line 2: Please change filename to lib_tccheckfifo.c

lib/libc/termios/lib_tccheckfifo.c
jeongarmy, line 70: Please change this to tccheckfifo

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -774,6 +774,76 @@ static void itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p(void)
 	TC_SUCCESS_RESULT();
 }
 
+/**
```
jeongarmy, line 777: I think it is better to separate this commit to implementations of iotbus functions and TCs.

lib/libc/termios/lib_tccheckfifo.c
deepaksrma: for passed cases it will return 1, otherwise 0.

lib/libc/termios/lib_tccheckfifo.c
deepaksrma: done

lib/libc/termios/lib_tccheckfifo.c
deepaksrma: thanks, corrected now.

framework/src/iotbus/iotbus_uart.c
deepaksrma: will return a specific error now.

framework/src/iotbus/iotbus_uart.c
deepaksrma: done

lib/libc/termios/lib_tccheckfifo.c
deepaksrma: now, rtl8721d_up_rxavailable is failing, so will check this now.

framework/include/iotbus/iotbus_error.h
```diff
@@ -60,6 +60,9 @@ typedef enum {
 	IOTBUS_ERROR_DEVICE_FAIL = -1103, /* Device is Not ready to use */
 	IOTBUS_ERROR_QUEUE_FULL = -1104, /* Queue is full */
 	IOTBUS_ERROR_OUT_OF_MEM = -1105, /* Heap is full */
+	IOTBUS_ERROR_RXFIFO_EMPTY = -1106, /*Recieve fifo is empty*/
```
abhishek-samsung, line 63: Please change Recieve --> Receive

framework/include/iotbus/iotbus_error.h
```diff
@@ -60,6 +60,9 @@ typedef enum {
...
+	IOTBUS_ERROR_TXFIFO_FULL = -1107, /*transmit fifo is full*/
+	IOTBUS_ERROR_TXFIFO_NOT_EMPTY = -1108, /*transmit fifo is not empty*/
```
amogh-samsung, line 65: Could you align all the macro descriptions please?

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -774,6 +774,75 @@ static void itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p(void)
...
+* @testcase         itc_systemio_iotbus_uart_rxavailable
+* @brief            To check if there is data in rx fifo.
+* @scenario         if data in rx fifo
+* @apicovered       iotbus_uart_rxavailable
+* @precondition     initializes uart_context
+* @postcondition    closes uart_context
+*/
+static void itc_systemio_iotbus_uart_rxavailable(void)
+{
+	int ret = IOTBUS_ERROR_NONE;
```
abhishek-samsung, line 787: Do we need to initialize ret here?

deepaksrma: it's better to initialize, I think.

framework/src/iotbus/iotbus_uart.c
```diff
@@ -176,6 +177,69 @@ int iotbus_uart_stop(iotbus_uart_context_h hnd)
...
+
+	if (!hnd || !hnd->handle) {
+		return IOTBUS_ERROR_INVALID_PARAMETER;
+	}
+
+	handle = (struct _iotbus_uart_s *)hnd->handle;
+	fd = handle->fd;
+
+	ret = tccheckfifo(fd, TIOCS_EMPTY);
+	if (ret < 1) {
```
jeongchanKim, line 236: I think it is better to check with `true`, not with `1`. Because `ioctl(fd, TIOCS_EMPTY, ,,,)` returns `true` for success case.
```
if (ret != true)
```

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,72 @@
...
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: tccheckfifo
+ *
+ * Description:
+ *   calls the ioctl.
```
jeongchanKim, line 58: Please explain much detailed what to do.

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,72 @@
...
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+
+#include <sys/ioctl.h>
+
+#include <termios.h>
+#include <errno.h>
```
jeongchanKim, line 30: I think this is not needed.

sunghan-chang:
@deepaksrma Please let me know the plan of this PR.

deepaksrma:
> @deepaksrma Please let me know the plan of this PR.

will update this after the merging of following PR :
https://github.com/Samsung/TizenRT/pull/5101

lib/libc/termios/lib_tccheckfifo.c
```diff
@@ -0,0 +1,72 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+
+#include <sys/ioctl.h>
+
```
amogh-samsung, line 28: Could we avoid multiple extra line spaces in between header file inclusions?

os/include/termios.h
```diff
@@ -319,6 +319,10 @@ int tcflow(int fd, int action);
  * @internal
  */
 int tcflush(int fd, int cmd);
+/**
```
amogh-samsung, line 322: Could you add a line space before this line please?

lib/libc/termios/lib_tccheckfifo.c
deepaksrma: corrected now.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -175,6 +175,8 @@
  */
 
 static serial_t* sdrv[MAX_UART_INDEX + 1] = {NULL, NULL, NULL, NULL, NULL}; //uart 0~3, uart2 is configured as log uart
+static bool g_tc_rxavailable;
+static bool g_tc_rxavailable_res;	// used to hold the result of rxavailable TC
```
sunghan-chang, line 179: I don't agree adding codes here for tc.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -865,6 +942,10 @@ int itc_uart_main(void)
 	itc_systemio_iotbus_uart_set_mode_write_read_p();
 	itc_systemio_iotbus_uart_set_mode_flush_write_read_p();
 	itc_systemio_iotbus_uart_set_flow_mode_flush_write_read_p();
+	itc_systemio_iotbus_uart_rxavailable();
+/*	itc_systemio_iotbus_uart_txready();
+	itc_systemio_iotbus_uart_txempty();
```
sunghan-chang, line 947: Why do you comment them out?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -571,6 +573,22 @@ static int rtl8721d_up_ioctl(FAR struct uart_dev_s *dev, int cmd, unsigned long
...
+		ret = g_tc_rxavailable_res;
+		break;
+
+	case TIOCS_READY:
+		ret = rtl8721d_up_txready(dev);
+		break;
+
+	case TIOCS_EMPTY:
+		ret = rtl8721d_up_txempty(dev);
+		break;
```
sunghan-chang, line 590: 1. How about other boards like imxrt? Does it return invalid or not supported or already impleneted?
2. Are they only used in the tc for coverage? Then let's cover them with TC config.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -571,6 +573,22 @@ static int rtl8721d_up_ioctl(FAR struct uart_dev_s *dev, int cmd, unsigned long
 		serial_control_loopback(sdrv[uart_index_get(priv->tx)], *(bool *)arg);
 		break;
 
+	case TIOCS_AVAIL:
+		g_tc_rxavailable = 1;
+		g_tc_rxavailable_res = 0;
+		rtl8721d_up_send(dev, 'a');
+		sleep(1);
+		ret = g_tc_rxavailable_res;
```
sunghan-chang, line 581: I don't know what you want to do. Could you leave comment to explain?

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
deepaksrma: mistakenly added, will uncomment

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: for rxavailable TC, steps are:

- rxavailable returns true when there is data in rx fifo and when data comes on the tx line in loopback mode then it enables the RxIrq and it goes to uart_recvchars(), where it checks whether data is available in rx fifo or not.
- so to check the data in rx fifo, we need to send the data first, which we have sent through rtl8721d_up_send() and before sending the data you need to enable the g_tc_rxavailable so that g_tc_rxavailable_res can collect the result for our sent data and when it collects the result then immediately turn off the g_tc_rxavailable, which means it won't collect the result further.

so, this algorithm came to my mind to test rxavailable(), if you have other steps, which can be better, please suggest.
I'll make comment once we agree on the algo.

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: please check my comment below and let me know what you think?

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: @kishore-sn @sangam-swami @jeongchanKim please review.

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: fill add this in other boards after finalizing for this board

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -865,6 +942,10 @@ int itc_uart_main(void)
 	itc_systemio_iotbus_uart_set_mode_write_read_p();
```
amogh-samsung, line 942: Could you please add a "TODO" comment for line 939
`//itc_systemio_iotbus_uart_write_read_flush_read_p();// TASH being hanged on execution of the testcase`
If this particular testcase is not resolved and is left to be resolved later.

os/arch/arm/src/amebad/amebad_serial.c
sunghan-chang: 1. I understand they are "begin" and "end" of tc to get valid data. Am I right?
2. It's the loopback test. There is no way to get another data. Why do you need begin and end?
3. Even it's loopback mode, sender and receiver should be implemented using existed APIs without more variables. If it's impossible, it means our APIs are not enough to use serial functionality. Then we should add them as public API, not tc.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -175,6 +175,8 @@
  */
 
 static serial_t* sdrv[MAX_UART_INDEX + 1] = {NULL, NULL, NULL, NULL, NULL}; //uart 0~3, uart2 is configured as log uart
+static bool g_tc_rxavailable;
```
kishore-sn, line 178: Please add a comment explaining the meaning of this variable.

os/arch/arm/src/amebad/amebad_serial.c
kishore-sn: How about changing this implementation as follows:
Remove all global variables. 

`case TIOCS_AVAIL:`
`rtl8721d_up_send(dev, 'a');`
	`	sleep(1);`
	`	ret = rtl8721d_up_rxavailable(dev);`
`break;`

kishore-sn: On the other hand, I think we can change it further to make the code much cleaner.

In the driver we do:
`case TIOCS_AVAIL:`
`ret = rtl8721d_up_rxavailable(dev);`
`break;`

In the test case file we do:
`iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);`
`	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);`
`		ret = iotbus_uart_write(h_uart, sz_input_text, sizeof(sz_input_text));`
`ret = iotbus_uart_rxavailable(h_uart);`

deepaksrma: we can not do like this:
```
case TIOCS_AVAIL:
rtl8721d_up_send(dev, 'a');
sleep(1);
ret = rtl8721d_up_rxavailable(dev);
break;
```

because when we sent the data in loopback mode then it comes on the RX FIFO immediately and generates the interrupt,
and calls uart_recvchars() where it checks whether data is available in RX FIFO or not, if data is available then it removes the data from RX FIFO and copies it to buffer so the following lines won't work, because when it comes to rtl8721d_up_rxavailable(), then data is not in RX FIFO because it has been transferred to buffer.
```
ret = rtl8721d_up_rxavailable(dev);
break;
```

kishore-sn: ok. 
Then we need to keep a global variable. But its better to keep a single global variable and protect it with a semaphore. Add a new ioctl to reset global variable. 

`bool g_rxavail'
'sem_t g_rxavail_sem`

`TCIOC_RXAVAIL_RESET:`
`take g_rxavail_sem`
`g_rxavail = false`
`give g_rxavail_sem`

`TCIOC_RXAVAIL:`
`take g_rxavail_sem`
`ret = g_rxavail;`
`give g_rxavail_sem`

And in rtl8721d_up_rxavailable()

`take g_rxavail_sem`
`g_rxavail = ret;`
`give g_rxavail_sem`

kishore-sn: ok. 
Then we need to keep a global variable. But its better to keep a single global variable and protect it with a semaphore. Add a new ioctl to reset global variable. 

`bool g_rxavail'
'sem_t g_rxavail_sem`

`TCIOC_RXAVAIL_RESET:`
`take g_rxavail_sem`
`g_rxavail = false`
`give g_rxavail_sem`

`TCIOC_RXAVAIL:`
`take g_rxavail_sem`
`ret = g_rxavail;`
`give g_rxavail_sem`

And in rtl8721d_up_rxavailable()

`take g_rxavail_sem`
`g_rxavail = ret;`
`give g_rxavail_sem`

sunghan-chang:
@sangam-swami Could you manage this PR?

PR #5054, opened 2021-07-13
CircleCI: Enable circleCI for build test

jeongarmy:
Add config.yml to run circleCI


.circleci/config.yml
```diff
@@ -0,0 +1,126 @@
...
+            - checkout_code
+      - esp_wrover_kit/hello_with_tash:
+          requires:
+            - checkout_code
+      - imxrt1020-evk/loadable_elf_apps:
+          requires:
+            - checkout_code
+      - rtl8721csm/loadable_apps:
+          requires:
+            - checkout_code
```
sunghan-chang, line 125: As far as I know, they are all separate jobs. It causes huge docker pull. How about executing build_test only?

sunghan-chang (commented):
Is it done for circle ci setting itself?

jeongchanKim (commented):
We already have travis ci. 
If we need both, please add comment why we need.
If not, please remove one of them.

jeongarmy:
@jeongchanKim We will replace travisCI to circleCI.

.circleci/config.yml
jeongarmy: OK. I agree with your opinion. Please check modified codes.

an4967 (approved):
LGTM

jeongchanKim (commented):
Please split the commit. One for adding circle ci, the other is for removing travis ci.

an4967 (approved):
LGTM

PR #5061, opened 2021-07-14
os/arch/arm/src/amebad: Fix mixed uart log issue

AdamZhang0124:
This PR fixes the mixed uart log issue while multiple functions use log_uart at the same time.


jeongchanKim (commented):
Could you add the commit message what is the change about?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -114,7 +114,7 @@
 #elif defined(CONFIG_UART1_SERIAL_CONSOLE)
 #define CONSOLE_DEV             g_uart1port             /* UART1 is console */
 #define TTYS0_DEV               g_uart1port             /* UART1 is ttyS0 */
-#define CONSOLE                 UART1_DEV
+#define CONSOLE                 UART3_DEV
```
jeongchanKim, line 117: All other defines are related to uart1, but this define is uart3.
Is it correct? And if then, why we should set like this?

jeongarmy:
@AdamZhang0124 Could you update this commit with rebase?

os/arch/arm/src/amebad/amebad_serial.c
sunghan-chang: What we tested does not include this change. Could you let me know why this is suddenly added?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -672,7 +672,27 @@ static bool rtl8721d_up_txready(struct uart_dev_s *dev)
...
+	}
+	else
+#elif defined(CONFIG_UART2_SERIAL_CONSOLE)
+	if (uart_index_get(priv->tx) == 2){
+		while (!serial_writable(sdrv[uart_index_get(priv->tx)]));
+		return true;
+	}
+	else
+#endif
+		return (serial_writable(sdrv[uart_index_get(priv->tx)]));
```
sunghan-chang, line 695: To make it align, let's do as below.
```c
{
      return (...);
}
```

os/arch/arm/src/amebad/amebad_serial.c
jeongchanKim: @sunghan-chang In our test code, I missed this change....
@AdamZhang0124 , Anyway, in our test, we can successfully run without this change.
Is this needed?

AdamZhang0124: Hi @sunghan-chang  @jeongchanKim,
This is an notation mismatch I recently found.
The TizenRT uart1 actually maps to Ameba's uart3.
It won't affect the function.

jeongchanKim: Thanks for checking :)

jeongarmy:
@AdamZhang0124 Could you update this commit with rebase master?

AdamZhang0124:
Hi @sunghan-chang @jeongchanKim @jeongarmy ,
Please allow me to close this PR and raise another one, all the discussed questions will be fixed in the new PR, thanks!

PR #5065, opened 2021-07-14
network/wifi: remove mac address field in wifi manager

pillip8282:
remove mac address field in wifi manager.
netlib_getmacaddr() return mac address of NIC.

### DIFF
--- a/framework/src/wifi_manager/wifi_manager_info.c
+++ b/framework/src/wifi_manager/wifi_manager_info.c
@@ -30,15 +30,13 @@
 struct _wifimgr_info {
 	char ssid[WIFIMGR_SSID_LEN + 1];             // SSID of Connected AP if mode is a station
 	char softap_ssid[WIFIMGR_SSID_LEN + 1];      // SoftAP SSID if mode is a soft ap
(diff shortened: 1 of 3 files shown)


framework/src/wifi_manager/wifi_manager_info.c
```diff
@@ -30,15 +30,13 @@
...
-	char mac_address[WIFIMGR_MACADDR_LEN];	   // MAC address of wifi interface
 	int rssi;                  // It is only used for a station mode
 	wifimgr_state_e state;
 	wifimgr_state_e prev_state;
 };
 typedef struct _wifimgr_info _wifimgr_info_s;
 
-static _wifimgr_info_s g_manager_info = {{0}, {0}, {0}, 0};
-
+static _wifimgr_info_s g_manager_info = {{0}, {0}, 0, WIFIMGR_NONE, WIFIMGR_NONE};
```
thapav, line 39: why iarestate & prev_state values added now and not before  ?
Any particular reason?

pillip8282: there is no reason. I think it's mistake.

PR #5078, opened 2021-07-20
Add SPI driver build break fixes

thapav:
#Patch 1
apps/examples/testcase/ta_tc/systemio: Fix build breaks when SPI driver TC's are enabled
This patch adds fix for:
1) Missing declarations of SPI driver itc:
	int itc_spi_main(void);
2) Config dependency for unused API's when CONFIG_SPI_EXCHANGE is disabled.
3) SPI bus for UTC to 0.

#Patch 2
apps/examples/testcase/ta_tc/systemio/itc: Rename API in SPI ITC
Fix typo SPI test case ITc and rename to itc_systemio_spi_write_recv_transfer_p

### DIFF
--- a/apps/examples/testcase/ta_tc/systemio/itc/itc_internal.h
+++ b/apps/examples/testcase/ta_tc/systemio/itc/itc_internal.h
@@ -37,4 +37,6 @@
 **********************************************************/
 char *Expect_Value(int, int);
 
+int itc_spi_main(void);
+
 #endif /* __EXAMPLES_TESTCASE_SYSTEMIO_ITC_INTERNAL_H */
--- a/apps/examples/testcase/ta_tc/systemio/utc/utc_spi.c
+++ b/apps/examples/testcase/ta_tc/systemio/utc/utc_spi.c
@@ -29,7 +29,7 @@
 
 iotbus_spi_context_h spi;
 
-unsigned int bus = 1;
+static unsigned int bus = 0;
 
 unsigned char txbuf[64] = { 0, };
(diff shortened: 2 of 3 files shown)


thapav:
@sunghan-chang @sangam-swami @jeongchanKim 
Please review.

apps/examples/testcase/ta_tc/systemio/utc/utc_spi.c
```diff
@@ -29,7 +29,7 @@
 
 iotbus_spi_context_h spi;
 
-unsigned int bus = 1;
+unsigned int bus = 0;
```
jeongchanKim, line 32: Please add static if this variable is used within this file.
Please check another variables also.

apps/examples/testcase/ta_tc/systemio/itc/itc_internal.h
```diff
@@ -37,4 +37,6 @@
 **********************************************************/
 char *Expect_Value(int, int);
 
+int itc_spi_main(void);
```
jeongchanKim, line 40: This prototype is for itc, not utc.
So I think this change can be into another commit which you made for itc.

thapav: The commit to which it is currently added fixes all build breaks. Therefore it is better to keep it here.
The other commit is for renaming an API.

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #5083, opened 2021-07-22
network/netmgr: add event handler task

pillip8282:
add event handler task not to driver create new thread if wi-fi event occurs.

### DIFF
--- a/os/net/netmgr/netdev_wifi.c
+++ b/os/net/netmgr/netdev_wifi.c
@@ -35,7 +35,74 @@
 			res = (dev->t_ops.wl->method)param;	\
 		}										\
 	} while (0)
+
+#define LOCK_EVTQUEUE(queue)								\
+	do {													\
+		if (sem_wait(&queue.lock) != 0) {					\
+			int derr_no = get_errno();						\
+			if (derr_no == EINTR) {							\
+				continue;									\
+			} else {										\
+				NET_LOGE(TAG, "lock fail %d\n", derr_no);	\
+				assert(0);									\
+			}												\
+		}													\
+	} while (0)
+
+#define UNLOCK_EVTQUEUE(queue)							\
+	do {												\
+		if (sem_post(&queue.lock) != 0) {				\
+			NET_LOGE(TAG, "unlock fail %d\n", errno);	\
+			assert(0);									\
+		}												\
+	} while (0)
+
+#define WAIT_EVTQUEUE(queue)										\
+	do {															\
+		if (sem_wait(&queue.signal) != 0) {							\
+			int derr_no = get_errno();								\
+			if (derr_no == EINTR) {									\
+				continue;											\
+			} else {												\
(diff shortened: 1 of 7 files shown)


os/net/netmgr/netdev_wifi.c
```diff
@@ -184,11 +251,74 @@ int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
...
+
+	NET_LOGE(TAG, "critical error\n", errno);
+	return NULL;
+}
+
+int trwifi_run_handler(void)
+{
+	if (sem_init(&g_queue.lock, 0, 1) != 0) {
+		NET_LOGE(TAG, "init lock fail %d\n", errno);
+		return -1;
```
deepaksrma, line 309: How about `return ERROR` instead of `return -1`

os/net/netmgr/netdev_wifi.c
```diff
@@ -184,11 +251,74 @@ int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
...
+	}
+	if (sem_init(&g_queue.signal, 0, 0) != 0) {
+		NET_LOGE(TAG, "init signal fail %d\n", errno);
+		return -1;
```
deepaksrma, line 313: same as above.

os/net/netmgr/netdev_wifi.c
```diff
@@ -184,11 +251,74 @@ int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
...
+	}
+
+	sys_thread_t tid;
+	tid = kernel_thread("netmgr_event_handler", NETMGR_EVTHANDLER_PRIO,
+						NETMGR_EVTHANDLER_STACKSIZE, _trwifi_event_handler, NULL);
+	if (tid == ERROR) {
+		NET_LOGE(TAG, "critical error %d\n", errno);
+		return -2;
```
deepaksrma, line 321: Here, how about `return EINVAL` ?

os/net/netmgr/netdev_wifi.c
pillip8282: I think ERROR is used to interact to system. because it's defined in sys/types.h. So in this case, returning -1 is right because it's only called by netmgr inside.

os/net/netmgr/netdev_wifi.c
pillip8282: I think ERROR is used to interact to system. because it's defined in sys/types.h. So in this case, returning -1 is right because it's only called by netmgr inside.

os/net/netmgr/netdev_wifi.c
pillip8282: I think EINVAL is for vfs operation. In this case, trwifi_run_handler() is called by netmgr inside. So returning -2 seems right.

an4967 (approved):
LGTM

PR #5088, opened 2021-07-26
framework/ble_manager: add 'set_adv_type' api

an4967:
- add 'ble_addr' struct
- add 'set_adv_type' api


os/net/blemgr/bledev.c
```diff
@@ -413,17 +412,22 @@ int bledev_handle(struct bledev *dev, lwnl_req cmd, void *data, uint32_t data_le
...
+	case LWNL_REQ_BLE_SET_ADV_TYPE:
 	{
-		uint8_t *bd_addr = { 0, };
+		trble_adv_type_e adv_type;
+		trble_addr *addr;
 
+		lwnl_msg_params param = { 0, };
 		if (data != NULL) {
-			bd_addr = (uint8_t *)data;
+			memcpy(&param, data, data_len);
```
jeongarmy, line 422: The maximum size of `param` is LWNL_MAX_PARAM.
In lwnl.h (tinyara/lwnl)
```
typedef struct {
	uint8_t count;
	void *param[LWNL_MAX_PARAM];
} lwnl_msg_params;
```
So I think you need to compare data_len with MWNL_MAX_PARAM before calling memcpy or put LWNL_MAX_PARAM instead of data_len.
Please check it to avoid wrong memory access

an4967: If you see 'ble_manager_lwnl.c', the value of  'data_len' is always the size of 'lwnl_msg_params'.
'data_len' is under the control of the BLE manager strictly, so there is no worry about wrong access of memory.

pillip8282 (approved):
lgtm

PR #5089, opened 2021-07-26
os/board/rtl8721csm : Improve Wi-Fi join performance

keerthanalex:
For connecting to same AP after leave/disconnect, include partial scan to improve the join performance
Store Wi-Fi connection SSID/channel to do partial scan, about ~1 sec improvement in connection time compared to default

### DIFF
--- a/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
+++ b/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
@@ -182,6 +182,14 @@ extern void cmd_p2p_connect(int argc, char **argv);
 extern u32 CmdDumpWord(IN u16 argc, IN u8 *argv[]);
 extern u32 CmdWriteWord(IN u16 argc, IN u8 *argv[]);
 #endif
+
+struct wlan_connect_info {
+	unsigned char ssid[TRWIFI_SSID_LEN + 1];
+	unsigned char channel;
+};
+static struct wlan_connect_info wifi_ap_info;
+static void wlan_store_connect_info(void);
+
 #ifdef CONFIG_CONCURRENT_MODE
 static void cmd_wifi_sta_and_ap(int argc, char **argv)
 {
@@ -415,6 +423,24 @@ int8_t cmd_wifi_ap(trwifi_softap_config_s *softap_config)
 	return ret;
 }
 
+void wlan_store_connect_info(void)
+{
+	u8 *ifname[1] = {(u8*)WLAN0_NAME};
+	rtw_wifi_setting_t setting;
+
+	memset(&setting, 0, sizeof(rtw_wifi_setting_t));
+	if (wifi_get_setting((const char*)ifname[0],&setting)) {
+		RTW_API_INFO("\r\n %s():wifi_get_setting fail", __func__);
+		return;
+	}
+
+	if (setting.mode != RTW_MODE_AP) {
+		rtw_memset(&wifi_ap_info,0,sizeof(struct wlan_connect_info));
+		rtw_memcpy(&wifi_ap_info.ssid, setting.ssid, strlen((char const*)setting.ssid) + 1);
+		wifi_ap_info.channel = setting.channel;
+	}
+}
+
 int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 {
 	int ret;
@@ -427,6 +453,9 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 	int key_id = 0;
 	void *semaphore;
 	int security_retry_count = 0;
+
+	uint32_t	channel;
+	uint8_t     pscan_config;
 
 	wifi_utils_ap_auth_type_e auth = ap_connect_config->ap_auth_type;
 	wifi_utils_ap_crypto_type_e crypto = ap_connect_config->ap_crypto_type;
@@ -488,6 +517,17 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 		return -1;
 	}
 
+	if (rtw_memcmp(ssid, &wifi_ap_info.ssid, strlen(ssid))) {
+		channel = wifi_ap_info.channel;
+		channel &= 0xff;
+		pscan_config = PSCAN_ENABLE | PSCAN_FAST_SURVEY;
+		ret = wifi_set_pscan_chan((uint8_t *)&channel, &pscan_config, 1);
+		rtw_memset(&wifi_ap_info,0,sizeof(struct wlan_connect_info));
+		if (ret < 0) {
+			RTW_API_INFO("\n\rset pscan failed");
+		}
+	}
+
 	ret = wifi_connect(ssid,
 					   security_type,
 					   password,
@@ -499,6 +539,8 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 	if (ret) {
 		ndbg("\n\rrtk_wifi_connect failed\n");
 		return -1;
+	} else {
+		wlan_store_connect_info();
 	}
 
 	return 0;


os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -415,6 +422,20 @@ int8_t cmd_wifi_ap(trwifi_softap_config_s *softap_config)
...
+{
+	u8 *ifname[1] = {(u8*)WLAN0_NAME};
+	rtw_wifi_setting_t setting;
+
+	if(wifi_get_setting((const char*)ifname[0],&setting) || setting.mode == RTW_MODE_AP){
+		RTW_API_INFO("\r\n %s():wifi_get_setting fail or ap mode", __func__);
+		return;
+	}
+
+	rtw_memcpy(&wifi_ap_info.ssid, setting.ssid, strlen((char const*)setting.ssid));
```
pillip8282, line 435: strlen doesn't contain null character. So it can cause side effect. So you should copy strlen(ssid) + 1

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -488,6 +512,17 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 		return -1;
 	}
 
+	if (rtw_memcmp(ssid, &wifi_ap_info.ssid, strlen(ssid))) {
+		channel = wifi_ap_info.channel;
```
pillip8282, line 516: If connected AP changes channel later(SSID is same), then does it connect to the AP? Please check mentioned case.

pillip8282 (commented):
wifi_ap_info is global variable. So I worried race condition of wifi_ap_info because there is no protection for it. In upper layer only wifi_msg_handler will access wifi driver by using trwifi_ops interface. However I don't know how does wi-fi driver treat this value. Could you tell me how is wifi_ap_info safe to race condition?

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: > If connected AP changes channel later(SSID is same), then does it connect to the AP? Please check mentioned case.

@pillip8282 , If the channel is modified after leave and tries to connect to same AP, there will be connect fail for the 1st connect after channel modification with the implementation. The pscan will start the scan with previous channel, it will fail first but we memset "wifi_ap_info" after every pscan irrespective of success/fail. 
Since it is a fail case, the profile will be empty inside "wlan_store_connect_info()". For the second round, it does full scan and connects normally & stores the latest profile (same SSID and modified channel information) for the subsequent rounds.
We asked Eric to check the behavior of TP2x for the same case and he confirmed that it will also have one connect fail after channel modification.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: > strlen doesn't contain null character. So it can cause side effect. So you should copy strlen(ssid) + 1

Done.

keerthanalex:
> wifi_ap_info is global variable. So I worried race condition of wifi_ap_info because there is no protection for it. In upper layer only wifi_msg_handler will access wifi driver by using trwifi_ops interface. However I don't know how does wi-fi driver treat this value. Could you tell me how is wifi_ap_info safe to race condition?

@pillip8282 , The added "wifi_ap_info" struct and  "wlan_store_connect_info" function are all specific only to "wifi_interactive_mode.c" file. 
1) "wifi_ap_info" struct  will be "memset"  inside "cmd_wifi_connect" after every partial scan irrespective of fail/success.
2) "wlan_store_connect_info" function which assigns the value to  "wifi_ap_info" struct will be called only inside "cmd_wifi_connect" for success case and it gets the value from "wifi_get_setting()".
3) static is added to both struct & function.
Do we need to provide any more information?

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -415,6 +423,20 @@ int8_t cmd_wifi_ap(trwifi_softap_config_s *softap_config)
 	return ret;
 }
 
+void wlan_store_connect_info(void)
+{
+	u8 *ifname[1] = {(u8*)WLAN0_NAME};
+	rtw_wifi_setting_t setting;
+
+	if(wifi_get_setting((const char*)ifname[0],&setting) || setting.mode == RTW_MODE_AP){
```
jeongchanKim, line 431: If `wifi_get_setting()` works well, then this if condition will be passed regardless of checking `setting.mode`.
Checking `setting.mode` is working only when `wifi_get_setting()` is failed.
But `setting.mode` cannot be `RTW_MODE_AP` without `wifi_get_setting()`.
So I think checking `setting.mode` in this condition is not necessary.
If needed, it is better to check after this if condition.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -415,6 +423,20 @@ int8_t cmd_wifi_ap(trwifi_softap_config_s *softap_config)
 	return ret;
 }
 
+void wlan_store_connect_info(void)
+{
+	u8 *ifname[1] = {(u8*)WLAN0_NAME};
+	rtw_wifi_setting_t setting;
```
jeongchanKim, line 429: I think it is better to initialize the `setting` for checking `setting.mode` properly.

keerthanalex:
@pillip8282 @jeongchanKim do you have any more comments? I'll update altogether in one commit.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
pillip8282: Then upper layer(mostly wifi_manager) must call trwifi_connect_ap() twice in that case?
1) tp1x leaves an AP.
2) AP change its channel
3) tp1x connect to an AP via trwifi_connect_ap()
4) A wi-fi driver returns error
5) wifi_manager calls trwifi_connect_ap() again
6) A wi-fi driver returns success.
Is above sequence right?
If so, how about to modify trwifi_connect_ap() logic even though tp2x does same.
If pscan failed with previous channel, how about to do full scan not to return errors to wifi_manager immediately.

pillip8282:
> 1. irrespective

static doesn't matter. even though global variable was defined to static if two threads(or task) access the variable then race condition could happen. So please check that what task access the wifi_ap_info. tp1x driver runs several tasks(ex. rtw_recv_tasklet). A task "wifi msg handler" accessed wifi_ap_info because it calls cmd_wifi_connect(). please check that other tasks which access this and analyze possibility of race condition.

jeongchanKim:
> @pillip8282 @jeongchanKim do you have any more comments? I'll update altogether in one commit.

I don't have more. Please check the above comments :)

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: > Then upper layer(mostly wifi_manager) must call trwifi_connect_ap() twice in that case?
> 
> 1. tp1x leaves an AP.
> 2. AP change its channel
> 3. tp1x connect to an AP via trwifi_connect_ap()
> 4. A wi-fi driver returns error
> 5. wifi_manager calls trwifi_connect_ap() again
> 6. A wi-fi driver returns success.
> 
> Is above sequence right?
> If so, how about to modify trwifi_connect_ap() logic even though tp2x does same.
> If pscan failed with previous channel, how about to do full scan not to return errors to wifi_manager immediately.
> To reduce time, I think suggested logic will be fast.

@pillip8282, do you suggest  "5. wifi_manager calls trwifi_connect_ap() again" in case of channel modification scenario? 
The current implementation is,
1.tp1x leaves an AP.
2.AP change its channel
3.tp1x connect to an AP via trwifi_connect_ap()
4.A wi-fi driver returns error which in turn returns error to wifi_manager 

Proposed logic:
1.tp1x leaves an AP.
2.AP change its channel
3.tp1x connect to an AP via trwifi_connect_ap()
4.A wi-fi driver returns error
5.wifi_manager calls trwifi_connect_ap() again
6.A wi-fi driver returns success which in turn returns success to wifi_manager

Is that correct?

keerthanalex:
> > "wifi_ap_info" struct will be "memset" inside "cmd_wifi_connect" after every partial scan irrespective of fail/success.
> > "wlan_store_connect_info" function which assigns the value to "wifi_ap_info" struct will be called only inside  "cmd_wifi_connect" for success case and it gets the value from "wifi_get_setting()".
> > static is added to both struct & function.
> > Do we need to provide any more information?
> 
> static doesn't matter. even though global variable was defined to static if two threads(or task) access the variable then race condition could happen. So please check that what task access the wifi_ap_info. tp1x driver runs several tasks(ex. rtw_recv_tasklet). A task "wifi msg handler" accessed wifi_ap_info because it calls cmd_wifi_connect(). please check that other tasks which access this and analyze possibility of race condition.

@pillip8282, currently we have only one instance of calling cmd_wifi_connect(). As you mentioned for  "wm_test join" , "wifi msg handler" task access "wifi_ap_info" because it calls "wifi_netmgr_utils_connect_ap" which in turn calls "cmd_wifi_connect()". 
Do we have protection in upper layer? If so, then "cmd_wifi_connect()" doesn't need any protection.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: The `setting` is memset to '0' and The condition of  `setting.mode` is placed after `wifi_get_setting()`.

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: Modified according to the suggestion.

jeongchanKim (approved):
LGTM

keerthanalex:
> LGTM

@jeongchanKim @pillip8282 
since we have one more PR #5129 related to join performance, what is the status of this PR?

keerthanalex:
@pillip8282, PR  #5129 has been merged to TizenRT master. What is the status of this PR?

keerthanalex:
Duplicate to #5129 , so closing this PR.

PR #5091, opened 2021-07-26
arch/arm/src/armv8-m/up_svcall.c : Add missing TZ_LoadContext_S when …

jeongchanKim:
…restore context

Before restoring the tcb context, we should load the secure context of next running tcb.


kishore-sn (approved):
Looks good.

sangam-swami (approved):
looks fine

sunghan-chang (approved):

PR #5008, opened 2021-06-24
build/tools/amebad/bootloader/: Recovery feature for boot param bootloader

fengyunlai:
Recovery feature for boot param bootloader.

PR #5097, opened 2021-07-29
kernel/binary_manager: Update booted kernel information from bootparam

jeongarmy:
Update booted kernel binary information from boot parameter data without scanning all kernel binaries.

This PR will be updated with commits about boot parameter recovery, #5008 soon.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_bootparam.c
+++ b/os/kernel/binary_manager/binary_manager_bootparam.c
@@ -1,0 +1,152 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <tinyara/config.h>
+#include <debug.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+
+#include <tinyara/binary_manager.h>
+
+#include "binary_manager.h"
+
+/* Data for Boot parameters */
+static binmgr_bpinfo_t bp_info;
+
+#define BP_SEEK_OFFSET(index)   (index == 0 ? 0 : BOOTPARAM_PARTSIZE / 2)
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+/****************************************************************************
+ * Name: binary_manager_register_bppart
+ *
+ * Description:
+ *	 This function registers the partition of boot parameters.
+ *
+ ****************************************************************************/
+void binary_manager_register_bppart(int part_num, int part_size)
+{
+	if (part_num < 0 || part_size != BOOTPARAM_PARTSIZE) {
+		bmdbg("ERROR: Invalid part info : num %d, size %d\n", part_num, part_size);
+		return;
+	}
+
+	bp_info.part_num = part_num;
+
+	bmvdbg("[BOOTPARAM] part num %d\n", part_num);
+}
+
+bool is_valid_bootparam(binmgr_bpdata_t *bp_data)
+{
+	if (!bp_data) {
+		bmdbg("Boot param is NULL\n");
+		return false;
+	} else if (bp_data->format_ver == 0 || bp_data->active_idx >= binary_manager_get_kdata()->part_count) {
+		bmdbg("Invalid data. ver: %u, active index: %u, addresses: %x, %x\n", bp_data->version, bp_data->active_idx, bp_data->address[0], bp_data->address[1]);
+		return false;
+	} else if (bp_data->crc_hash != crc32((uint8_t *)bp_data + CHECKSUM_SIZE, sizeof(binmgr_bpdata_t) - CHECKSUM_SIZE)) {
+		bmdbg("Invalid crc32 value, crc32 %u, ver: %u, active index: %u, addresses: %x, %x\n", bp_data->crc_hash, bp_data->version, bp_data->active_idx, bp_data->address[0], bp_data->address[1]);
+		return false;
+	}
+
+	return true;
+}
+
+/*****************************************************************************************************
+ * Name: binary_manager_scan_bootparam
+ *
+ * Description:
+ *	 This function updates the latest boot parameter information by scanning all boot parameters.
+ *
+ *****************************************************************************************************/
+int binary_manager_scan_bootparam(void)
+{
+	int fd;
+	int ret;
+	int bp_idx;
+	int latest_idx = -1;
+	uint32_t latest_ver = 0;
+	binmgr_bpdata_t bp_data[BOOTPARAM_COUNT];
+	char bp_devpath[BINARY_PATH_LEN];
+
+	if (bp_info.part_num < 0) {
+		bmdbg("Invalid Bootparam partition Num : %d\n", bp_info.part_num);
+		return BINMGR_INVALID_PARAM;
+	}
+
+	/* Open dev to access boot parameters */
+	snprintf(bp_devpath, BINARY_PATH_LEN, BINMGR_DEVNAME_FMT, bp_info.part_num);
+	fd = open(bp_devpath, O_RDWR, 0666);
+	if (fd < 0) {
+		bmdbg("ERROR: Get a fd of bootparam, errno %d\n", errno);
+		return BINMGR_OPERATION_FAIL;
+	}
+
+	for (bp_idx = 0; bp_idx < BOOTPARAM_COUNT; bp_idx++) {
+		ret = lseek(fd, BP_SEEK_OFFSET(bp_idx), SEEK_SET);
+		if (ret < 0) {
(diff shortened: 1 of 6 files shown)


os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	for (bp_idx = 0; bp_idx < BOOTPARAM_COUNT; bp_idx++) {
+		ret = lseek(fd, BP_SEEK_OFFSET(bp_idx), SEEK_SET);
+		if (ret < 0) {
+			bmdbg("ERROR: Seek Failed, errno %d\n", errno);
+			return BINMGR_OPERATION_FAIL;
+		}
+		ret = read(fd, (FAR uint8_t *)&bp_data[bp_idx], sizeof(binmgr_bpdata_t));
+		bmvdbg("Bootparam[%d] ver: %u, active index: %u, addresses: %x, %x\n", bp_idx, bp_data[bp_idx].version, bp_data[bp_idx].active_idx, bp_data[bp_idx].address[0], bp_data[bp_idx].address[1]);
```
jeongchanKim, line 107: How about printing the log after checking that read properly or not?

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,145 @@
...
+	bp_info.part_num = part_num;
+
+	bmvdbg("[BOOTPARAM] part num %d\n", part_num);
+}
+
+bool is_valid_bootparam(binmgr_bpdata_t *bp_data)
+{
+	if (!bp_data) {
+		bmdbg("Boot param is NULL\n");
+		return false;
```
jeongchanKim, line 63: For boolean type, please include `stdbool.h`.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,145 @@
...
+	if (fd < 0) {
+		bmdbg("ERROR: Get a fd of bootparam, %d\n", fd);
+		return BINMGR_OPERATION_FAIL;
+	}
+
+	for (bp_idx = 0; bp_idx < BOOTPARAM_COUNT; bp_idx++) {
+		ret = lseek(fd, BP_SEEK_OFFSET(bp_idx), SEEK_SET);
+		if (ret < 0) {
+			bmdbg("ERROR: Seek Failed, errno %d\n", errno);
+			return BINMGR_OPERATION_FAIL;
```
jeongchanKim, line 104: Is it okay not to close the `fd` for failure case?

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,145 @@
...
+			continue;
+		} else if (!is_valid_bootparam(&bp_data[bp_idx])) {
+			continue;
+		}
+
+		/* Update the latest version and index */
+		if (latest_ver < bp_data->version) {
+			latest_ver = bp_data->version;
+			latest_idx = bp_idx;
+			bmdbg("Latest Version %d, index %d\n", latest_ver, latest_idx);
```
jeongchanKim, line 119: I think it is better to use `bmvdbg` instead of `bmdbg`.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,147 @@
...
+/****************************************************************************
+ * Name: binary_manager_register_bppart
+ *
+ * Description:
+ *	 This function registers the partition of boot parameters.
+ *
+ ****************************************************************************/
+void binary_manager_register_bppart(int part_num, int part_size)
+{
+	if (part_num < 0 || part_size != BOOTPARAM_PARTSIZE) {
```
jeongchanKim, line 50: If this function works only when `part_size` is same as `BOOTPARAM_PARTSIZE`, passing the size is needed?
And boot param partition size is checked when build time through `mkbootparam.py`.
So I think checking the `part_size` is not needed.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,147 @@
...
+		bmdbg("ERROR: Invalid part info : num %d, size %d\n", part_num, part_size);
+		return;
```
jeongchanKim, line 52: What will happen if registering `bppart` fails?
Is it okay for binary manager?
Why I'm asking, it just prints the error log, but do nothing. Can binary manager distinguish the success or handling differently?

jeongarmy: Binary manager doesn't do anything now except for returning error. I will add handling that case.

os/kernel/binary_manager/binary_manager_bootparam.c
davidfather: What if part_num = 0? I don't think this case is successful.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,147 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+
+#include <tinyara/binary_manager.h>
+
+#include "binary_manager.h"
+
+/* Data for Boot parameters */
+static binmgr_bpinfo_t bp_info;
```
davidfather, line 34: This static variable seems to be accessed by several APIs each of which can be called by separate thread.
How can this structure be protected from accessing by multiple threads?

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,147 @@
...
+/****************************************************************************
+ * Name: binary_manager_get_bpdata
+ *
+ * Description:
+ *	 This function gets boot parameter data.
+ *
+ ****************************************************************************/
+binmgr_bpdata_t *binary_manager_get_bpdata(void)
+{
+	return &bp_info.bp_data;
```
davidfather, line 146: Before accessing data, why don't you check that registration is done first?

os/kernel/binary_manager/binary_manager_bootparam.c
jeongarmy: Before refactoring about separation of BM operations like loading, reloading to each threads, binary manager only accessed this variable so any protection was unnecessary.
But now this can be accessed as you said and should be protected.
I will add codes for protection with another PR. Thank you for good point I missed.

davidfather: Do you guarantee that only the bin manager (i.e., a single thread) manipulate this variable?

jeongarmy: Yes. I guarantee that with current implementation.
There are many things to be implemented about bootparam.
I will check and add codes for protection if it is needed at that time.

os/kernel/binary_manager/binary_manager_bootparam.c
jeongarmy: Yes, but I added this to check old binary which bootparam is not applied.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,147 @@
...
+	int latest_idx = -1;
+	uint32_t latest_ver = 0;
+	binmgr_bpdata_t bp_data[BOOTPARAM_COUNT];
+	char bp_devpath[BINARY_PATH_LEN];
+
+	/* Open dev to access boot parameters */
+	snprintf(bp_devpath, BINARY_PATH_LEN, BINMGR_DEVNAME_FMT, bp_info.part_num);
+	fd = open(bp_devpath, O_RDWR, 0666);
+	if (fd < 0) {
+		bmdbg("ERROR: Get a fd of bootparam, %d\n", fd);
```
jeongchanKim, line 97: How about printing the `errno` instead of `fd`?
Because `open` returns just `ERROR` when fail-case, but it is not helpful.

os/kernel/binary_manager/binary_manager_data.c
```diff
@@ -183,30 +155,25 @@ void binary_manager_register_kpart(int part_num, int part_size)
...
+			return true;
 		}
 	}
-	bmvdbg("Latest version [%u] %u\n", kernel_info.inuse_idx, kernel_info.version);
 
-	/* Found valid binary */
-	if (kernel_info.version != 0) {
-		return true;
-	}
 	return false;
```
jeongchanKim, line 177: Even if it returns false because of failure to scan the boot param, the log message says it is from parsing partition list.
It is not good to debug when fail.
Please distinguish the log message for failure case.

jeongarmy: I think it is enough because each cases of failure print error logs.

jeongchanKim: I mean, even the failure is from bootparam, but the error logs says it is from parsing of partition list.
```
		if (!is_found_kpart) {
			lldbg("ERROR!! Not found kernel partitions because parsing a partition list is failed.\n");
			lldbg("Please check logs from configure_mtd_partitions and whether the partition 'kernel' exists in CONFIG_FLASH_PART_TYPE.\n");
		}
```
Do you think it is enough?

jeongarmy: I added logs for failures of bootparam scanning.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -0,0 +1,152 @@
...
+		/* Update the latest version and index */
+		if (latest_ver < bp_data->version) {
+			latest_ver = bp_data->version;
+			latest_idx = bp_idx;
+			bmvdbg("Latest Version %d, index %d\n", latest_ver, latest_idx);
+		}
+	}
+	close(fd);
+
+	if (latest_ver < 0 || latest_idx < 0) {
```
jeongchanKim, line 130: `latest_ver` is from `bp_data->version` and `bp_data->version` is unsigned type.
Even `latest_ver` is also unsigned type.
So it cannot be smaller than 0.

sunghan-chang (approved):

PR #5101, opened 2021-08-02
app/examples/testcase/systemio/itc: fix the UART Read/write TCs

deepaksrma:
commit 1: changes in iotbus to add loopback mode in UART.
commit 2: add loopback mode in UART.
commit 3: refactor the ITC UART TCs

### DIFF
--- a/apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
+++ b/apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
@@ -25,6 +25,7 @@
  ****************************************************************************/
 #include <tinyara/config.h>
 #include <iotbus_uart.h>
+#include <termios.h>
 #include "itc_internal.h"
 #include "iotbus_error.h"
 #include <string.h>
@@ -270,10 +271,14 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
 void itc_systemio_iotbus_uart_write_read_p(void)
 {
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	ret = iotbus_uart_write(h_uart, sz_input_text, sizeof(sz_input_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart));
@@ -282,8 +287,11 @@ void itc_systemio_iotbus_uart_write_read_p(void)
 
 	ret = iotbus_uart_read(h_uart, sz_output_text, sizeof(sz_output_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart));
-
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text, sz_output_text), 0, iotbus_uart_stop(h_uart));
+
+	// turn off the loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	ret = iotbus_uart_stop(h_uart);
 	TC_ASSERT_EQ("iotbus_uart_stop", ret, IOTBUS_ERROR_NONE);
@@ -325,11 +333,15 @@ void itc_systemio_iotbus_uart_flush_p(void)
 static void itc_systemio_iotbus_uart_write_read_flush_p(void)
 {
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	ret = iotbus_uart_write(h_uart, sz_input_text, sizeof(sz_input_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart));
@@ -339,6 +351,10 @@ static void itc_systemio_iotbus_uart_write_read_flush_p(void)
 	ret = iotbus_uart_read(h_uart, sz_output_text, sizeof(sz_output_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text, sz_output_text), 0, iotbus_uart_stop(h_uart));
+
+	// turn off the loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	ret = iotbus_uart_flush(h_uart);
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_flush", ret, IOTBUS_ERROR_NONE, iotbus_uart_stop(h_uart));
@@ -360,8 +376,8 @@ static void itc_systemio_iotbus_uart_write_read_flush_p(void)
 static void itc_systemio_iotbus_uart_write_read_flush_p_multi_handle(void)
 {
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text1[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
-	char sz_input_text2[BUF_LEN] = "UART READ/WRITE ITC TESTING1!";
+	char sz_input_text1[BUF_LEN] = "UART R/W TEST1";
+	char sz_input_text2[BUF_LEN] = "UART R/W TEST2";
 	char sz_output_text[BUF_LEN];
 
 	iotbus_uart_context_h h_uart1 = iotbus_uart_init(DEVPATH);
@@ -370,10 +386,11 @@ static void itc_systemio_iotbus_uart_write_read_flush_p_multi_handle(void)
 	iotbus_uart_context_h h_uart2 = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ_CLEANUP("iotbus_uart_init", h_uart2, NULL, iotbus_uart_stop(h_uart1));
 
-	ret = iotbus_uart_write(h_uart1, sz_input_text1, sizeof(sz_input_text1));
-	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart1, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 
-	ret = iotbus_uart_write(h_uart2, sz_input_text2, sizeof(sz_input_text2));
+	ret = iotbus_uart_write(h_uart1, sz_input_text1, sizeof(sz_input_text1));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 
 	usleep(MICROSECOND);
@@ -382,9 +399,18 @@ static void itc_systemio_iotbus_uart_write_read_flush_p_multi_handle(void)
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text1, sz_output_text), 0, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 
+	ret = iotbus_uart_write(h_uart2, sz_input_text2, sizeof(sz_input_text2));
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
+
+	usleep(MICROSECOND);
+
 	ret = iotbus_uart_read(h_uart2, sz_output_text, sizeof(sz_output_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text2, sz_output_text), 0, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
+
+	// turn off the loopback mode
+	ret = iotbus_uart_control_loopback(h_uart2, LOOPBACK_OFF);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 
 	ret = iotbus_uart_flush(h_uart1);
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_flush", ret, IOTBUS_ERROR_NONE, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
@@ -453,7 +479,7 @@ static void itc_systemio_iotbus_uart_set_flow_write_read_p(void)
 {
 	int i_size = 4;
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 	int rtscts[4][2] = { {1, 0}, {0, 1}, {1, 1}, {0, 0} };
 	int index = 0;
@@ -463,6 +489,10 @@ static void itc_systemio_iotbus_uart_set_flow_write_read_p(void)
 
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	for (index = 0; index < i_size; index++) {
 		ret = iotbus_uart_set_flowcontrol(h_uart, rtscts[index][0], rtscts[index][1]);
@@ -487,6 +517,10 @@ static void itc_systemio_iotbus_uart_set_flow_write_read_p(void)
 			count_fail_read++;
 		}
 	}
+	// turn off loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
+
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write failed count", count_fail_write, 0, iotbus_uart_stop(h_uart));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read failed count", count_fail_read, 0, iotbus_uart_stop(h_uart));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_set_flowcontrol failed count", count_fail_flowcontrol, 0, iotbus_uart_stop(h_uart));
@@ -509,7 +543,7 @@ static void itc_systemio_iotbus_uart_set_flow_flush_write_read_p(void)
 {
 	int i_size = 4;
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 	int rtscts[4][2] = { {1, 0}, {0, 1}, {1, 1}, {0, 0} };
 	int index = 0;
@@ -520,6 +554,10 @@ static void itc_systemio_iotbus_uart_set_flow_flush_write_read_p(void)
 
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	for (index = 0; index < i_size; index++) {
 		ret = iotbus_uart_set_flowcontrol(h_uart, rtscts[index][0], rtscts[index][1]);
@@ -552,6 +590,10 @@ static void itc_systemio_iotbus_uart_set_flow_flush_write_read_p(void)
 		}
 	}
 
+	// turn off loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
+
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_flush failed count", count_fail_flush, 0, iotbus_uart_stop(h_uart));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write failed count", count_fail_write, 0, iotbus_uart_stop(h_uart));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read failed count", count_fail_read, 0, iotbus_uart_stop(h_uart));
@@ -576,7 +618,7 @@ static void itc_systemio_iotbus_uart_set_mode_write_read_p(void)
 	int i_bytesize = 8;
 	int i_stop_bits = 1;
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 	int mode[] = { IOTBUS_UART_PARITY_NONE, IOTBUS_UART_PARITY_EVEN, IOTBUS_UART_PARITY_ODD };
 	char *mode_str[3] = { "IOTBUS_UART_PARITY_NONE", "IOTBUS_UART_PARITY_EVEN", "IOTBUS_UART_PARITY_ODD" };
@@ -588,6 +630,10 @@ static void itc_systemio_iotbus_uart_set_mode_write_read_p(void)
 
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret < 0, false, iotbus_uart_stop(h_uart));
 
 	for (index = 0; index < i_modes; index++) {
 		ret = iotbus_uart_set_mode(h_uart, i_bytesize, mode[index], i_stop_bits);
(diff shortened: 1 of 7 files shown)


lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+ *   termiosp - The termiosp argument is a pointer to a termios structure.
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void loopback_en(int fd, int cmd, FAR struct termios *termiosp)
+{
+	ioctl(fd, cmd, (unsigned long) termiosp);
```
jeongchanKim, line 67: Please remove space after type casting.

jeongchanKim (commented):
I think it is better to divide the commit between iotbus changes and libc changes.

jeongchanKim (commented):
Other boards like artik, imxrt don't need this functionality?

jeongchanKim (commented):
Is there any way to overcome the size limitation?

framework/src/iotbus/iotbus_uart.c
```diff
@@ -388,7 +390,15 @@ int iotbus_uart_write(iotbus_uart_context_h hnd, const char *buf, unsigned int l
 	}
 
 	fd = handle->fd;
+
+	tio.loopback_en = 1;
```
jeongchanKim, line 394: How about use `true` or `false`?

sangam-swami:
the PR is for verification of uart read and write tc's, can we change the heading

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -564,6 +564,13 @@ static int rtl8721d_up_ioctl(FAR struct uart_dev_s *dev, int cmd, unsigned long
 		rtl8721d_up_rxint(dev, priv->rxint_enable);
 		break;
 
+	case TIOCLOOPBACK:
+		if (!termiosp) {
```
jeongchanKim, line 568: All cases check that `termiosp` is NULL or not.
How about making it as common at line 510?

lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+ *   fd  - The 'fd' argument is an open file descriptor associated with a terminal.
+ *   cmd - ioctl argument.
+ *   termiosp - The termiosp argument is a pointer to a termios structure.
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void loopback_en(int fd, int cmd, FAR struct termios *termiosp)
```
jeongarmy, line 65: Please change function name to `loopback_enable` or something for readability.

lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <sys/ioctl.h>
+#include <termios.h>
+
+/****************************************************************************
```
jeongarmy, line 29: Please remove unnecessary comments.

lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Description:
+ *   calls the ioctl to enable/ disable the loopback mode of UART
+ *
+ * Input Parameters:
+ *   fd  - The 'fd' argument is an open file descriptor associated with a terminal.
+ *   cmd - ioctl argument.
+ *   termiosp - The termiosp argument is a pointer to a termios structure.
+ *
+ * Returned Value:
```
jeongarmy, line 60: Is there any reason that return type of this function is `void`?
A `ioctl` executes some parameters like fd, args for control something.
The result depends on incoming parameters so it can be failed according to them.
At that time, we can't know whether enabling or disabling loopback is succeed or not.
Please change return type from `void` to `int` to be like ioctl if there is no special reason.

lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: loopback_en
+ *
+ * Description:
+ *   calls the ioctl to enable/ disable the loopback mode of UART
+ *
+ * Input Parameters:
+ *   fd  - The 'fd' argument is an open file descriptor associated with a terminal.
```
sangam-swami, line 56: no need to mention The 'fd' argument, the meaning is enough

lib/libc/termios/lib_loopback_en.c
```diff
@@ -0,0 +1,68 @@
...
+ *   cmd - ioctl argument.
```
sangam-swami, line 57: We can add , command used to control uart loopback mode?

os/include/termios.h
```diff
@@ -358,6 +359,10 @@ int tcsendbreak(int fd, int duration);
  */
 int tcsetattr(int fd, int options, FAR const struct termios *termiosp);
 
+/*
+ * enable/disable the loopback mode.
+ */
+void loopback_en(int fd, int cmd, FAR struct termios *termiosp);
```
sangam-swami, line 365: function name suggest enable but description mentions both enable and disable feature..
can you change the function name

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: i think, we should not do that because in future, suppose if someone wants to send 0 in termios and just wants to work on cmd then it will return error.
so i think we should not change this.

lib/libc/termios/lib_loopback_en.c
deepaksrma: corrected.

lib/libc/termios/lib_loopback_en.c
deepaksrma: corrected

lib/libc/termios/lib_loopback_en.c
deepaksrma: i think, that won't be good idea, because we should explain the all args of a function and suppose if we define 2 args in comments so someone may be confused after seeing the comments and actual args.

please let me know what you think

os/include/tinyara/serial/tioctl.h
```diff
@@ -202,6 +202,7 @@
 #define TIOCCLRSIG      _TIOC(0x002f)	/* Clear signal to handle Interrupt */
 #define TIOCTXSTATE     _TIOC(0x0030)	/* Get State of TX */
 #define TIOCRXSTATE     _TIOC(0x0031)	/* Get State of RX */
+#define TIOCLOOPBACK    _TIOC(0x0032)	/* set loopback mode */
```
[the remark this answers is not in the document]
deepaksrma: comment was not appropriate for this cmd. corrected now.
we can not rename it like `TIOCSETLO`  because we doing both operations with this cmd, enable and disable.

deepaksrma:
> Is there any way to overcome the size limitation?

can you please tell me you are talking about which size limitation??

deepaksrma:
> Other boards like artik, imxrt don't need this functionality?

i didn't find the loopback support bit in artik and imxrt user manual, so can't add this to these boards as of now.
i'll add this in mtk.

os/include/termios.h
```diff
@@ -358,6 +358,10 @@ int tcsendbreak(int fd, int duration);
  */
 int tcsetattr(int fd, int options, FAR const struct termios *termiosp);
 
+/*
+ * enable/disable the loopback mode.
+ */
+void loopback(int fd, int cmd, FAR struct termios *termiosp);
```
thapav, line 365: Please check the type.
Return type > int

lib/libc/termios/lib_loopback.c
```diff
@@ -0,0 +1,53 @@
...
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
```
sunghan-chang, line 25: Unnecessary. Let's remove.

framework/src/iotbus/iotbus_uart.c
```diff
@@ -388,10 +390,24 @@ int iotbus_uart_write(iotbus_uart_context_h hnd, const char *buf, unsigned int l
 	}
 
 	fd = handle->fd;
+
+	tio.loopback = true;
+	ret = loopback(fd, TIOCLOOPBACK, &tio);
```
sunghan-chang, line 395: I don't know why user should give this value. When user calls the loopback function, this is fixed and must be given. Then we can do like below
```c
  ret = loopback(fd, &tio);

int loopback(int fd, struct termios *termiosp)
{
	return ioctl(fd, TIOCLOOPBACK, (unsigned long)termiosp);
}
```

sunghan-chang: And same as tio argument.
It only uses the loopback member. Then we can just give int value, not structure.
```c
in os/include/termios.h

/* Constants for use with loopback() */

#define TCLBON    0				/* Enable loopback */
#define TCLBOFF     1				/* Disable loopback */

in loopback.c

int loopback(int fd, int cmd)
{
	return ioctl(fd, TIOCLOOPBACK, (unsigned long)&cmd);
}
```

lib/libc/termios/lib_loopback.c
```diff
@@ -0,0 +1,53 @@
...
+ *   cmd - ioctl command used to control uart loopback mode.
+ *   termiosp - The termiosp argument is a pointer to a termios structure.
+ *
+ * Returned Value:
+ *   Upon successful completion, 0 is returned. Otherwise, -1 is returned and
+ *   errno is set to indicate the error.
+ *
+ ****************************************************************************/
+
+int loopback(int fd, int cmd, FAR struct termios *termiosp)
```
sunghan-chang, line 50: loopback is too common. Many many module can use loopback. 
How about adding `tc` as prefix like other termios functions?

os/board/rtl8721csm/src/component/common/mbed/hal/serial_api.h
```diff
@@ -143,6 +143,16 @@ void serial_irq_handler(serial_t *obj, uart_irq_handler handler, uint32_t id);
...
 
+/**
+  * @brief  Enable/Disable UART loopback mode for testing.
+  * @param  obj: uart object define in application software.
+  * @param  enable/disable the loopback mode
+  * 		@arg 0 disable
+  *		@arg 1 enable
+  * @retval none
+  */
+void serial_enable_loopback(serial_t *obj, bool enable);
```
sunghan-chang, line 154: This function can disable loopback so that function name could be mismatched.
How about `serial_control_loopback`?

os/include/termios.h
```diff
@@ -261,6 +261,7 @@ struct termios {
 	 */
 
 	speed_t c_speed;			/* Input/output speed (non-POSIX) */
+	bool loopback;				/* enable/disable loopback mode */
```
sunghan-chang, line 264: To use `bool` type, you should include stdbool.h.

sunghan-chang (commented):
Could you add loopback testcase and then add function description like below in termios.h?
```h
/* Set the parameters associated with the terminal */
/**
 * @ingroup TERMIOS_KERNEL
 * @brief set the parameters associated with the terminal
 * @details @b #include <termios.h> \n
 * POSIX API (refer to : http://pubs.opengroup.org/onlinepubs/9699919799/)
 * @since TizenRT v1.0
 */
int tcsetattr(int fd, int options, FAR const struct termios *termiosp);
```

framework/src/iotbus/iotbus_uart.c
```diff
@@ -388,10 +390,24 @@ int iotbus_uart_write(iotbus_uart_context_h hnd, const char *buf, unsigned int l
...
+	if (ret < 0) {
```
jeongarmy, line 396: How about adding logs for error cases for easy debugging?
That's because it returns IOTBUS_ERROR_UNKNOWN for failures of loopback, write.
With current implementation, we should put test logs to find failed point whenever this function is failed.

lib/libc/termios/lib_loopback.c
jeongarmy: I agree with chang's opinion. loopback is too common and ambiguous.
Please add prefix as chang's opinion or verb which indicates what it does

deepaksrma: corrected

os/board/rtl8721csm/src/component/common/mbed/hal/serial_api.h
deepaksrma: corrected

framework/src/iotbus/iotbus_uart.c
deepaksrma: i am returning IOTBUS_ERROR_UNKNOWN because we are setting a bit and it will not return any error unless ioctl returns any error and in ioctl we don't know which kind of error it can return.
you can suggest to me what kind of error we can print the logs here?

framework/src/iotbus/iotbus_uart.c
deepaksrma: done

lib/libc/termios/lib_tcloopback.c
```diff
@@ -0,0 +1,53 @@
+/****************************************************************************
+ *
+ * libc/termios/lib_loopback.c
```
jeongchanKim, line 3: Please change the file name in description.

os/board/rtl8721csm/src/component/common/mbed/hal/serial_api.h
```diff
@@ -143,6 +143,16 @@ void serial_irq_handler(serial_t *obj, uart_irq_handler handler, uint32_t id);
...
 
+/**
+  * @brief  Enable/Disable UART loopback mode for testing.
+  * @param  obj: uart object define in application software.
+  * @param  enable/disable the loopback mode
+  * 		@arg 0 disable
+  *		@arg 1 enable
+  * @retval none
+  */
+void serial_control_loopback(serial_t *obj, bool enable);
```
jeongchanKim, line 154: Please include `stdbool.h`.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -259,6 +259,38 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
...
+
+	ret = iotbus_uart_write(h_uart, sz_input_text, sizeof(sz_input_text));
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart));
+
+	usleep(MICROSECOND);
+
+	ret = iotbus_uart_read(h_uart, sz_output_text, sizeof(sz_output_text));
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart));
+
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text, sz_output_text), 0, iotbus_uart_stop(h_uart));
```
jeongchanKim, line 286: It is better to use `strncmp` instead of `strcmp`.

os/board/rtl8721csm/src/component/common/mbed/hal/serial_api.h
deepaksrma: Done

framework/src/iotbus/iotbus_uart.c
jeongarmy: I meant iotbus_uart_write, not ioctl.
If iotbus_uart_write returns IOTBUS_ERROR_UNKNOWN, user can't know where an error occurred because all failures of tcloopback and write return same value.
So I wanted to add some logs. For examples,
ret = tcloopback(fd, &loopback);
if (ret < 0) {
   ibdbg("tcloopback FAILED, %d\n", ret);
  return IOTBUS_ERROR_UNKNOWN;
}
...
ret = write(fd, buf, length);
if (ret < 0) {
  ibdbg("Write ## FAILED, fd %d, errno %d\n", fd, errno);
  return IOTBUS_ERROR_UNKNOWN;
}

deepaksrma: corrected now

lib/libc/termios/lib_tcloopback.c
```diff
@@ -0,0 +1,53 @@
+/****************************************************************************
```
Taejun-Kwon, line 1: Should we make library file only for tc?? I think ioctl is enough, library is unnecessary.

os/include/tinyara/serial/tioctl.h
```diff
@@ -202,6 +202,7 @@
 #define TIOCCLRSIG      _TIOC(0x002f)	/* Clear signal to handle Interrupt */
 #define TIOCTXSTATE     _TIOC(0x0030)	/* Get State of TX */
 #define TIOCRXSTATE     _TIOC(0x0031)	/* Get State of RX */
+#define TIOCLOOPBACK    _TIOC(0x0032)	/* Enable/Disable loopback mode */
```
Taejun-Kwon, line 205: For most of constant value, we usually follow linux's, there's TIOCM_LOOP in linux. Please check it.

lib/libc/termios/lib_tcloopback.c
deepaksrma: ok. you mean I should remove this file from here and should call ioctl from iotbus??
if this is the case then I did so to maintain symmetry because tc files are also there in lib.
please let me know what you think now??

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -373,15 +405,17 @@ static void itc_systemio_iotbus_uart_write_read_flush_p_multi_handle(void)
...
-	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
-
 	usleep(MICROSECOND);
 
 	ret = iotbus_uart_read(h_uart1, sz_output_text, sizeof(sz_output_text));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 	TC_ASSERT_EQ_CLEANUP("iotbus_uart_read", strcmp(sz_input_text1, sz_output_text), 0, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
 
+	ret = iotbus_uart_write(h_uart2, sz_input_text2, sizeof(sz_input_text2));
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart1); iotbus_uart_stop(h_uart2));
```
jeongchanKim, line 415: We usually use `TC_ASSERT_LEQ_CLEANUP` or `TC_ASSERT_GEQ_CLEANUP` for this case.
How about changing?

lib/libc/termios/lib_tcloopback.c
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Description:
+ *   calls the ioctl to enable/ disable the loopback mode of UART
+ *
+ * Input Parameters:
+ *   fd  - The 'fd' argument is an open file descriptor associated with a terminal.
+ *   arg - ioctl arg used to control uart loopback mode.
+ *
+ * Returned Value:
+ *   Upon successful completion, 0 is returned. Otherwise, -1 is returned and
```
jeongchanKim, line 45: When fail case, this function returns negative errno instead of setting the errno.
Please change the description.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -259,6 +259,38 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
...
 
+/**
+* @testcase         itc_systemio_iotbus_uart_loopback_p
+* @brief            check the uart loopback mode support
+* @scenario         check the uart loopback mode support
+* @apicovered       iotbus_uart_write, iotbus_uart_read
+* @precondition     initializes uart_context
+* @postcondition    closes uart_context
+*/
+void itc_systemio_iotbus_uart_loopback_p(void)
```
kishore-sn, line 270: How is this api different from below api "itc_systemio_iotbus_uart_write_read_p"?
We are performing same operations like opening same device, write, read and compare. The only difference I see is the text being written.

lib/libc/termios/lib_tcloopback.c
```diff
@@ -0,0 +1,53 @@
...
+ *   errno is set to indicate the error.
+ *
+ ****************************************************************************/
+
+int tcloopback(int fd, bool *arg)
+{
+	return ioctl(fd, TIOCLOOPBACK, (unsigned long)arg);
+}
```
kishore-sn, line 53: We dont need separate api and separate lib file for just one ioctl call. We can call ioctl directly from framework file.

framework/src/iotbus/iotbus_uart.c
```diff
@@ -388,10 +390,27 @@ int iotbus_uart_write(iotbus_uart_context_h hnd, const char *buf, unsigned int l
 	}
 
 	fd = handle->fd;
+
+	loopback = TCLBON;
+	ret = tcloopback(fd, &loopback);
```
kishore-sn, line 395: This framework api is used for normal uart write operation. But, if you set loopback always, then normal uart writes will always fail. Instead, we need to set loopback only if we are performing loopback test. I suggest that you must add new framework apis like iotbus_uart_loopback_enable and iotbus_uart_loopback_disable and perform the loopback enable/ disable only in such case.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
deepaksrma: I think we should leave it, as it is. because in the whole file they have used like this and if we change for one line then it won't look good and it does not affect the result so we can keep it same.
let me know still what you think?

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
```diff
@@ -684,6 +684,24 @@ void serial_irq_set(serial_t *obj, SerialIrq irq, uint32_t enable)
...
 
+/**
+  * @brief  Enable/Disable UART loopback mode.
+  * @param  obj: uart object define in application software.
+  * @param  enable/disable the loopback mode
+  *             @arg 0 disable
+  *             @arg 1 enable
+  * @retval none
+  */
+void serial_control_loopback(serial_t *obj, bool enable)
```
kishore-sn, line 695: I see you have implemented loopback for rtk driver. What about other boards?

kishore-sn (changes requested):

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
jeongchanKim: Ok fine. Let's keep it on as it is.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
deepaksrma: yes, this is not different from mentioned API. 
I created this because sunghan wanted to create an API to check loopback so that is the only way to check loopback, so after discussing with sangam, i created this TC.

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
deepaksrma: as i mentioned above
I didn't find the loopback support bit in artik and imxrt user manual, so don't know whether they support loopback or not so couldn't add this to these boards as of now.
i'll add this in mtk after merging of this PR.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -259,6 +262,55 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
...
+	char sz_output_text[BUF_LEN];
+	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
+	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	if (ret < 0) {
+		iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	}
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_control_loopback", ret, false, iotbus_uart_stop(h_uart));
```
kishore-sn, line 286: Here, the variable ret is of int type. So, you must not compare it to false. You must check if it is less than zero.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -259,6 +262,55 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
...
+	char sz_input_text[BUF_LEN] = "UART LB TEST";
+	char sz_output_text[BUF_LEN];
+	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
+	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
+
+	// turn on loopback mode
+	ret = iotbus_uart_control_loopback(h_uart, LOOPBACK_ON);
+	if (ret < 0) {
+		iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	}
```
kishore-sn, line 285: The above check is not required. If ret is less than zero, it means loopback setting has failed. So, we will stop the uart and end the test.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -259,6 +262,55 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
...
+	TC_ASSERT_EQ_CLEANUP("iotbus_uart_write", ret < 0, false, iotbus_uart_stop(h_uart));
+
+	usleep(MICROSECOND);
+
+	ret = iotbus_uart_read(h_uart, sz_output_text, sizeof(sz_output_text));
+	if (ret < 0) {
+		iotbus_uart_control_loopback(h_uart, LOOPBACK_OFF);
+	}
```
kishore-sn, line 299: I think its not a good way to call loopback setting each time. It is better to implement like follows:
- the default state of uart is loopback OFF
- So, we need to set loopback to OFF in uart_start and uart_stop api in framework
- Then we dont need to set loopback OFF every time if there is any error.

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -270,19 +322,37 @@ void itc_systemio_iotbus_uart_set_flowcontrol_p(void)
 void itc_systemio_iotbus_uart_write_read_p(void)
 {
 	int ret = IOTBUS_ERROR_NONE;
-	char sz_input_text[BUF_LEN] = "UART READ/WRITE ITC TESTING!";
+	char sz_input_text[BUF_LEN] = "UART R/W TEST";
 	char sz_output_text[BUF_LEN];
 	iotbus_uart_context_h h_uart = iotbus_uart_init(DEVPATH);
 	TC_ASSERT_NEQ("iotbus_uart_init", h_uart, NULL);
 
+	// turn on loopback mode
```
kishore-sn, line 330: Why do we need loopback setting in all these api's?
There is a separate API for loopback test. All these apis are supposed to be normal uart test without loopback. If we set loopback options in all these test cases, then how will they work for boards which do not have uart loopback?

kishore-sn (changes requested):

os/include/tinyara/serial/tioctl.h
```diff
@@ -202,6 +202,7 @@
 #define TIOCCLRSIG      _TIOC(0x002f)	/* Clear signal to handle Interrupt */
 #define TIOCTXSTATE     _TIOC(0x0030)	/* Get State of TX */
 #define TIOCRXSTATE     _TIOC(0x0031)	/* Get State of RX */
+#define TIOCM_LOOP	_TIOC(0x0032)	/* Enable/Disable loopback mode */
```
sunghan-chang, line 205: How about keep the naming policy?
```
TIOCLOOPBACK
```

os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
```diff
@@ -684,6 +684,24 @@ void serial_irq_set(serial_t *obj, SerialIrq irq, uint32_t enable)
...
+{
+	UART_TypeDef* UARTx = UART_DEV_TABLE[obj->uart_idx].UARTx;
+	if (enable) {
+		UARTx->MCR |= BIT(4);
+	}
+	else {
```
sunghan-chang, line 701: Let's use the same formatting.
I can see  
```
}  else {
```
, not 
```
}
else {
```.

File: os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c [hunk 35]
```diff
@@ -684,6 +684,24 @@ void serial_irq_set(serial_t *obj, SerialIrq irq, uint32_t enable)
 	}
 }
 
+/**
+  * @brief  Enable/Disable UART loopback mode.
+  * @param  obj: uart object define in application software.
+  * @param  enable/disable the loopback mode
```
sunghan-chang, line 690: param name is missing.
```
* @param enable: enable/disable the loopback mode

File: os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c (hunk 28 above)
Comment (sunghan-chang) [reply os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c:695]: As Jeongchan said. `bool` requires stdbool.h.

File: framework/src/iotbus/iotbus_uart.c [hunk 36]
```diff
@@ -371,6 +373,27 @@ int iotbus_uart_read(iotbus_uart_context_h hnd, char *buf, unsigned int length)
 	return ret;
 }
 
+int iotbus_uart_control_loopback(iotbus_uart_context_h hnd, bool mode)
```
sunghan-chang, line 376: I don't know why you implement the loopback on iotbus.

os/include/tinyara/serial/tioctl.h
sunghan-chang: Why we support the driver is keeping the same app code regardless of BSP.
To guarantee above, we should guide which value is loopback on and which value is off.
Could you add the definitions of LOOPBACK_ON and OFF?

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
```diff
@@ -37,6 +37,9 @@
 #define DEVPATH "/dev/ttyS2"
 #endif
 
+#define LOOPBACK_ON	1	//set the loopback mode for UART
```
sunghan-chang, line 40: Please use public definitions instead of local definitions in the app. Please find my previous comment.

sunghan-chang:
@sangam-swami @deepaksrma I am sorry but I have no plan to maintain iotbus. Could you let me know why you changed the commit to iotbus implementation?

os/include/tinyara/serial/tioctl.h
deepaksrma: i changed the name because taejun suggested that:

`For most of constant value, we usually follow linux's, there's TIOCM_LOOP in linux. Please check it.`

now, let me know what you think?

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
deepaksrma: `Why do we need loopback setting in all these api's?`
uart write, read TCs will pass only if loopback mode is supported.

`All these apis are supposed to be normal uart test without loopback`
No. without loopback,how can we test write and read? so loopback needs to be enabled, wherever write and read is used.

`how will they work for boards which do not have uart loopback?`
if loopback mode is not supported by any board then we can not test write read on that board.

kishore-sn: Ok. So, if there is no way to test uart without loopback, then these changes look good. 
But, in that case, we dont need separate api called `itc_systemio_iotbus_uart_loopback_p` because we are testing the exact same functionality in this api `itc_systemio_iotbus_uart_write_read_p`

kishore-sn (changes requested):

apps/examples/testcase/ta_tc/systemio/itc/itc_uart.c
deepaksrma: done

kishore-sn (approved):
Looks ok

jeongchanKim (dismissed):
LGTM

os/include/termios.h
```diff
@@ -227,6 +227,11 @@
 #define TCOOFF    2				/* Suspend output */
 #define TCOON     3				/* Restart output */
 
+/* Control the loopback mode */
+
+#define LOOPBACK_ON	1	/* Enable the Loopback mode */
+#define LOOPBACK_OFF	0	/* Disable the Loopback mode*/
```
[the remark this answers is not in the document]
deepaksrma: done

sunghan-chang (approved):

PR #5110, opened 2021-08-09
kernel : Change errno for wd_create fail case

jeongchanKim:
wd_create fails when it cannot allocate the wd.
So the errno of wd creation failure can be ENOMEM.


os/kernel/mqueue/mq_timedreceive.c
```diff
@@ -230,7 +230,7 @@ ssize_t mq_timedreceive(mqd_t mqdes, FAR char *msg, size_t msglen, FAR int *prio
 
 	rtcb->waitdog = wd_create();
 	if (!rtcb->waitdog) {
-		set_errno(EINVAL);
+		set_errno(ENOMEM);
```
sunghan-chang, line 233: Please add description of ENOMEM in the function description.

Taejun-Kwon (approved):

os/kernel/mqueue/mq_timedreceive.c
jeongchanKim: Added. Thanks.

sunghan-chang (approved):

PR #5111, opened 2021-08-09
arch/arm/amebad_timer_lowerhalf.c : Remove wrong return of amebad_gpt_handler

jeongchanKim:
1)   The return type of amebad_gpt_handler() is void, so it cannot return -ENODEV.
2)  'ret' is initialized as -EINVAL, so it is not need to be reassigned for some cases.


jeongarmy (approved):
LGTM

sunghan-chang:
@jeongchanKim I don't know why you say like below.
```
If there is no running timer, we cannot get the status of that timer, so just return OK.
```

kishore-sn (approved):
Looks good.

sunghan-chang (approved):

PR #5112, opened 2021-08-11
external/mbedtls: fix overwritten value

pillip8282:
hres is overwritten by sl_remove_key. In that case it can cause unexpected
behavior. So keeping the result in other variable.

### DIFF
--- a/external/mbedtls/alt/pk_wrap_alt.c
+++ b/external/mbedtls/alt/pk_wrap_alt.c
@@ -838,6 +838,7 @@ int eckey_verify_wrap(void *ctx, mbedtls_md_type_t md_alg, const unsigned char *
 	t_sig.data_len = sig_len;
 
 	ret = sl_ecdsa_verify_md(shnd, ecdsa_mode, &t_hash, &t_sig, key_idx, &hres);
+	hal_result_e hhres = hres;
 	free(t_hash.data);
 	free(t_sig.data);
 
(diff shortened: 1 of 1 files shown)


external/mbedtls/alt/pk_wrap_alt.c
```diff
@@ -846,7 +847,7 @@ int eckey_verify_wrap(void *ctx, mbedtls_md_type_t md_alg, const unsigned char *
 
 FREE:
 	sl_deinit(shnd);
-	if (ret != SECLINK_OK || hres != HAL_SUCCESS) {
+	if (ret != SECLINK_OK || hres != HAL_SUCCESS || hhres != HAL_SUCCESS) {
```
an4967, line 850: Is it okay without defining hhres such as Line:748?

Taejun-Kwon: Basically this is happened each of function in seclink always return SECLINK_OK.
So param hal_result_e should be checked all the time.
anyway each of step doesn't have a any dependency, and it doesn't jump to anywhere, so I think below will work.
```
ret = sl_ecdsa_verify_md(shnd, ecdsa_mode, &t_hash, &t_sig, key_idx, &hres);
if (ret != SECLINK_OK) { // but this will never be happened
	ret = MBEDTLS_ERR_RSA_UNSUPPORTED_OPERATION;
}

if (hres != HAL_SUCESS) {
	ret = MBEDTLS_ERR_RSA_UNSUPPORTED_OPERATION;
}

free(t_hash.data);
free(t_sig.data);

FREE_WITH_KEY:
	(void)sl_remove_key(shnd, key_type, key_idx, &hres);
	if (hres != HAL_SUCESS) {
		ret = MBEDTLS_ERR_RSA_UNSUPPORTED_OPERATION;
	}

FREE:
	sl_deinit(shnd);
	return ret;
```

PR #5113, opened 2021-08-11
board/rtl8721csm/bluetooth: Check active conn before read/write/notify

Pangwawaa:
Check active conn before read/write/notify


jeongchanKim (approved):
LGTM

an4967 (approved):
LGTM

PR #5115, opened 2021-08-12
arch/arm/src/armv7-r : Add task monitor functionality

jeongchanKim:
Task monitor is kernel thread which monitor the tasks/pthreads.
It checks if registered tasks/pthreads are alive.
It was implemented in v7m and v8m, so implement to v7-r also.

### DIFF
--- a/os/arch/arm/src/armv7-r/arm_syscall.c
+++ b/os/arch/arm/src/armv7-r/arm_syscall.c
@@ -275,6 +275,9 @@ uint32_t *arm_syscall(uint32_t *regs)
 
 		regs = (uint32_t *)regs[REG_R1];
 		DEBUGASSERT(regs);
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/armv7-r/arm_unblocktask.c
```diff
@@ -71,6 +71,10 @@
 #include <tinyara/debug/sysdbg.h>
 #endif
 
+#ifdef CONFIG_TASK_MONITOR
+#include "task_monitor/task_monitor_internal.h"
```
sunghan-chang, line 75: Could you let me know why this is necessary?

abhishek-samsung (approved):
LGTM

sunghan-chang (approved):

PR #5116, opened 2021-08-13
framework/wifi_manager: fix parameter validation in scan api

pillip8282:
Both channel and ssid_length are unsigned int. So they're never less
than zero.

### DIFF
--- a/framework/include/wifi_manager/wifi_manager.h
+++ b/framework/include/wifi_manager/wifi_manager.h
@@ -40,6 +40,7 @@ extern "C" {
 #define WIFIMGR_MACADDR_LEN        6
 #define WIFIMGR_MACADDR_STR_LEN    17
 #define WIFIMGR_SSID_LEN           32
+#define WIFIMGR_2G_CHANNEL_MAX     14
 #define WIFIMGR_PASSPHRASE_LEN     64
 /**
  * @brief <b> wifi MAC/PHY standard types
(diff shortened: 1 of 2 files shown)


framework/src/wifi_manager/wifi_manager_api.c
```diff
@@ -189,13 +189,12 @@ wifi_manager_result_e wifi_manager_scan_ap(wifi_manager_scan_config_s *config)
 {
 	NET_LOGI(TAG, "-->\n");
 	if (config) {
-		if (config->channel < 0 || config->channel > 14) {
+		if (config->channel > 14) {
```
jeongchanKim, line 192: There is `WIFIMGR_SSID_LEN` for `ssid_length`.
How about defining the channel length also?

framework/include/wifi_manager/wifi_manager.h
```diff
@@ -386,7 +386,7 @@ wifi_manager_result_e wifi_manager_disconnect_ap(void);
  *     if both SSID and channel are not set in config then it'll return error.
  *     wifi_manager check whether SSID is set by ssid_length in config. In the same way
  *     not checking channel is 0, but checking channel is greater than 0 and less than 15 in 2.5GHz.
```
jeongchanKim, line 388: In the code, we checked the channel length with `14`. But in the description it says `15`.
Please check this.

jeongchanKim, line 388: And we do not check that `channel is greater than 0` anymore.

pillip8282: wifi_manager checks it with following condition.
if (config->ssid_length == 0 && config->channel == 0) {

pillip8282: the valid range of 2.4G channel is from 1 to 14.

jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):
LGTM

an4967 (approved):
LGTM

PR #5117, opened 2021-08-13
external/webserver: Handling persistent connections in webserver and null pointer exception: [Do Not Merge]

amitpurwar:
This implement the Handling persistent connections in webserver and resolve null pointer exception

### DIFF
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -223,7 +223,7 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					} else {
 						HTTP_LOGD("unknown version\n");
 					}
-				} else { /* If this is called by webclient */
+				} else if(response) { /* If this is called by webclient */
 					http_separate_status_line(buf + len->sentence_start, &protocol, &response->status, response->phrase);
 					HTTP_LOGD("Response Status : %d\n", response->status);
 					HTTP_LOGD("Response Phrase : %s\n", response->phrase);
@@ -263,7 +263,9 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					}
 					if (strcmp(key, "Content-Length") == 0) {
 						len->content_len = HTTP_ATOI(value);
-						response->total_len = len->content_len;
+						if(response){
+							response->total_len = len->content_len;
+						}
 
 						HTTP_LOGD("This request contains contents, length : %d\n", len->content_len);
 					}
@@ -316,15 +318,17 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					HTTP_LOGD("[BODY_END]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d\n", buf_len, len->sentence_start, len->message_len, len->content_len);
 					//HTTP_LOGD("All body readed : \n%s\n", *body);
 
-					response->total_len = len->content_len;
+					if(response) {
+						response->total_len = len->content_len;
 
-					if (is_header == true) {
-						response->entity_len = buf_len - len->sentence_start;
-						response->entity = buf + len->sentence_start;
-					} else {
-						response->entity_len = buf_len;
-						response->entity = buf;
-					}	
+						if (is_header == true) {
+							response->entity_len = buf_len - len->sentence_start;
+							response->entity = buf + len->sentence_start;
+						} else {
+							response->entity_len = buf_len;
+							response->entity = buf;
+						}
+					}
 
 					read_finish = true;
 				} else {
@@ -362,20 +366,22 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					}
 					len->message_len += buf_len;
 
-					response->total_len = len->content_len;
+					if(response) {
+						response->total_len = len->content_len;
 
-					if (is_chunked) {
-						if (response->entity) {
-							response->entity = *body;
-							response->entity_len = len->content_len;
+						if (is_chunked) {
+							if (response->entity) {
+								response->entity = *body;
+								response->entity_len = len->content_len;
+							}
+						} else if (is_header == true && !is_chunked) {
+							response->entity_len = buf_len - len->sentence_start;
+							response->entity = buf + len->sentence_start;
+							HTTP_LOGD("[HEADER_BODY]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d entity_len : %d\n", buf_len, len->sentence_start, len->message_len, len->content_len, response->entity_len);
+						} else {
+							response->entity = buf;
+							response->entity_len = buf_len;
 						}
-					} else if (is_header == true && !is_chunked) {
-						response->entity_len = buf_len - len->sentence_start;
-						response->entity = buf + len->sentence_start;
-						HTTP_LOGD("[HEADER_BODY]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d entity_len : %d\n", buf_len, len->sentence_start, len->message_len, len->content_len, response->entity_len);
-					} else {
-						response->entity = buf;
-						response->entity_len = buf_len;
 					}
 					
 					HTTP_LOGD("[BODY]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d\n", buf_len, len->sentence_start, len->message_len, len->content_len);
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -94,6 +94,8 @@ pthread_addr_t http_handle_client(pthread_addr_t arg)
 			}
 		}
 #endif
+
+#ifdef CONFIG_NETUTILS_WEBSOCKET
 		http_keyvalue_list_init(&request_params);
 		result = http_recv_and_handle_request(p, &request_params);
 		http_keyvalue_list_release(&request_params);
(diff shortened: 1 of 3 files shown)


external/webserver/http_client.c
```diff
@@ -263,7 +263,9 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					}
 					if (strcmp(key, "Content-Length") == 0) {
 						len->content_len = HTTP_ATOI(value);
-						response->total_len = len->content_len;
+						if(response){
```
abhishek-samsung, line 266: Please follow coding rules. "if (condition) {".

abhishek-samsung (commented):
Please change the title of the commits to include the path to the file in which the changes were made.

external/webserver/http_client.c
```diff
@@ -103,6 +105,29 @@ pthread_addr_t http_handle_client(pthread_addr_t arg)
...
+			} else {
+				HTTP_LOGD("Client %d  in normal case.\n", sock_fd, p->keep_alive);
+
+				// Close client socket if connection is not persistent one.
+				if (p->keep_alive == 0) {
+					HTTP_LOGD("Client %d closing.\n", sock_fd);
+					close(p->client_fd);
+				}
+			}
+		} while(p->keep_alive);
```
jeongchanKim, line 128: Please add space after `while`.

external/webserver/http_client.c
```diff
@@ -237,7 +263,9 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					}
 					if (strcmp(key, "Content-Length") == 0) {
 						len->content_len = HTTP_ATOI(value);
-						response->total_len = len->content_len;
+						if(response){
+							response->total_len = len->content_len;
```
jeongchanKim, line 267: Is it okay not to check `len` is NULL or not before using?

external/webserver/http_client.c
```diff
@@ -197,7 +223,7 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					} else {
 						HTTP_LOGD("unknown version\n");
 					}
-				} else { /* If this is called by webclient */
+				} else if(response) { /* If this is called by webclient */
```
amogh-samsung, line 226: could you add a space after "else if" please?

external/webserver/http_client.c
```diff
@@ -290,15 +318,17 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					HTTP_LOGD("[BODY_END]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d\n", buf_len, len->sentence_start, len->message_len, len->content_len);
 					//HTTP_LOGD("All body readed : \n%s\n", *body);
 
-					response->total_len = len->content_len;
+					if(response) {
```
amogh-samsung, line 321: A single space after "if" please

external/webserver/http_client.c
```diff
@@ -336,20 +366,22 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					}
 					len->message_len += buf_len;
 
-					response->total_len = len->content_len;
+					if(response) {
```
amogh-samsung, line 369: A space after "if" please?

external/webserver/http_client.h
```diff
@@ -50,6 +50,8 @@ struct http_client_t {
 	mbedtls_ssl_context       tls_ssl;
 	mbedtls_net_context       tls_client_fd;
 #endif
+
```
amogh-samsung, line 53: Could we avoid this extra line space please?

external/webserver/http_client.c
```diff
@@ -103,6 +105,29 @@ pthread_addr_t http_handle_client(pthread_addr_t arg)
...
+			if (result != HTTP_OK) {
+				HTTP_LOGD("Client %d  in error case.\n", sock_fd);
+				break;
+			} else {
+				HTTP_LOGD("Client %d  in normal case.\n", sock_fd, p->keep_alive);
+
+				// Close client socket if connection is not persistent one.
+				if (p->keep_alive == 0) {
+					HTTP_LOGD("Client %d closing.\n", sock_fd);
+					close(p->client_fd);
```
Taejun-Kwon, line 125: Here client_fd is closed, but it doesn't have to close it after main loop finished? I mean line 130.

external/webserver/http_client.c
```diff
@@ -103,6 +105,29 @@ pthread_addr_t http_handle_client(pthread_addr_t arg)
 		} else {
 			HTTP_LOGD("Client %d  in normal case.\n", sock_fd);
 		}
+#else
+		do {
```
Taejun-Kwon, line 109: How about below style?
```
do_keep_alive :
		http_keyvalue_list_init(&request_params);
		result = http_recv_and_handle_request(p, &request_params);
		http_keyvalue_list_release(&request_params);
		if (result != HTTP_OK) {
			HTTP_LOGD("Client %d  in error case.\n", sock_fd);
		} else {
			HTTP_LOGD("Client %d  in normal case.\n", sock_fd);
		}
#ifndef CONFIG_NETUTILS_WEBSOCKET
		/* If result is OK & it need to be keep alive, try once again */
		if (result == HTTP_OK && p->keep_alive) {
			goto do_keep_alive;
		}
		HTTP_LOGD("Client %d closing.\n", sock_fd);
		close(p->client_fd);
#endif
```

external/webserver/http_client.c
Taejun-Kwon: If response is null, is this function working fine? Or, you should add null check & return routine regarding parameters at the start of function.
Especially, response referenced by this function in many places.

external/webserver/http_client.c
```diff
@@ -501,6 +534,15 @@ int http_recv_and_handle_request(struct http_client_t *client, struct http_keyva
 			goto errout;
 		}
 	}
+
+	// Check "Connection" header value
+	conn_type = http_keyvalue_list_find(request_params, "Connection");
+	if (!strcmp(conn_type, "Keep-Alive")) {
```
Taejun-Kwon, line 540: Let's use strncmp

external/webserver/http_client.c
```diff
@@ -501,6 +534,15 @@ int http_recv_and_handle_request(struct http_client_t *client, struct http_keyva
 			goto errout;
 		}
 	}
+
+	// Check "Connection" header value
+	conn_type = http_keyvalue_list_find(request_params, "Connection");
```
Taejun-Kwon, line 539: http_keyvalue_list_find returns value which matched with key, then I think conn_type is not good name.

external/webserver/http_client.c
```diff
@@ -544,11 +586,10 @@ int http_recv_and_handle_request(struct http_client_t *client, struct http_keyva
 		}
 		pthread_setname_np(ws->thread_id, "websocket handle server");
 		pthread_detach(ws->thread_id);
-	} else
-#endif
-	{
+	} else {
 		close(client->client_fd);
 	}
+#endif
```
Taejun-Kwon, line 592: same as above opinion, then when client_fd is closed?

external/webserver/http_client.c
```diff
@@ -336,20 +366,22 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
...
-							response->entity_len = len->content_len;
+						if (is_chunked) {
+							if (response->entity) {
+								response->entity = *body;
+								response->entity_len = len->content_len;
+							}
+						} else if (is_header == true && !is_chunked) {
+							response->entity_len = buf_len - len->sentence_start;
+							response->entity = buf + len->sentence_start;
+							HTTP_LOGD("[HEADER_BODY]buf_len : %d len->sentence_start : %d len->message_len : %d len->content_len %d entity_len : %d\n", buf_len, len->sentence_start, len->message_len, len->content_len, response->entity_len);
```
dineshdas-samsung, line 380: Is it possible to break the line for better readability ?

pillip8282 (commented):
Could you describe what cases cause null pointer exception?
Please add above description to commit message.

external/webserver/http_client.c
```diff
@@ -94,6 +94,8 @@ pthread_addr_t http_handle_client(pthread_addr_t arg)
 			}
 		}
 #endif
+
+#ifdef CONFIG_NETUTILS_WEBSOCKET
```
pillip8282, line 98: why did you add CONFIG_NETUTILS_WEBSOCKET in here. could you comment the reason?

external/webserver/http_client.c
pillip8282: is it ok to call strcmp when conn_type is null?

external/webserver/http_client.c
HarishSRIB: len is pointer to local structure in http_recv_and_handle_request so it will not be NULL.

external/webserver/http_client.c
HarishSRIB: response is NULL only when this function called from webserver. but in webserver case client is not NULL so this block will not hit. response will never be NULL when this methiod called from webclient.

external/webserver/http_client.h
```diff
@@ -50,6 +50,8 @@ struct http_client_t {
...
+	int keep_alive;
```
jeongchanKim, line 54: As checked, this variables set only 0 or 1. 
How about using bool type instead of int?
It is much better to read and can save the memory.

os/net/lwip/src/api/sockets.c
```diff
@@ -388,6 +388,11 @@ struct lwip_sock *get_socket(int s, pid_t pid)
 	}
 
 	tcb = sched_gettcb(pid);
+	if (tcb == NULL) {
+		printf("pid: %d[%d] tcb is null\n", (int)pid, getpid());
```
sunghan-chang, line 392: 1. Let's use lwip's debug way instead of printf.
2. line 396, DEBUGASSERT(tcb) is the same as you added when CONFIG_DEBUG is enabled. If you want to add this instead of line 396, you should remove line 396 to resolve duplicated. And as line 396 does, you should check `tcb->group` as well.
3. I think you can remove `for external/webserver` from the title. This code is not for only external/webserver.

sunghan-chang (commented):
There are 5 commits but I think they are not 5 topics. You should not make a new commit to apply comment. It should be merged into original commit. It's written in the commit rules.

sunghan-chang:
Please check commit rules again. You don't follow some rules.

PR #5120, opened 2021-08-17
Support prodconfig

jeongchanKim:
TASH>>prodconfig
HW Debug : On
Download : Downloader & Hotfix Available

### DIFF
--- a/apps/system/utils/Kconfig
+++ b/apps/system/utils/Kconfig
@@ -140,6 +140,13 @@ config HEAPINFO_USER_GROUP_LIST
 endif
 endif
 
+config ENABLE_PRODCONFIG
+	bool "prodconfig"
+	depends on PRODCONFIG
+	default n
+	---help---
+		print product configuration like swd.
+
 config ENABLE_TZSELECT
 	bool "tzselect"
 	default n
--- a/apps/system/utils/systemcmd.c
+++ b/apps/system/utils/systemcmd.c
@@ -52,6 +52,9 @@ const static tash_cmdlist_t system_cmds[] = {
 #if defined(CONFIG_ENABLE_KILLALL)
 	{"killall",  utils_killall,      TASH_EXECMD_SYNC},
 #endif
+#if defined(CONFIG_ENABLE_PRODCONFIG)
+	{"prodconfig", utils_prodconfig, TASH_EXECMD_SYNC},
+#endif
 #if defined(CONFIG_ENABLE_PS)
 	{"ps",       utils_ps,           TASH_EXECMD_SYNC},
 #endif
--- a/apps/system/utils/utils_prodconfig.c
+++ b/apps/system/utils/utils_prodconfig.c
@@ -1,0 +1,67 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+#include <errno.h>
+#include <stdio.h>
+#include <fcntl.h>
+
+#include <tinyara/prodconfig.h>
+#include <tinyara/fs/ioctl.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name : utils_prodconfig
+ *
+ * Description:
+ *   It reads the hw configurations from prodconfig drivers
+ *   and prints the status like swd enable/disable.
+ *
+ ****************************************************************************/
+int utils_prodconfig(int argc, char **args)
+{
+	int prodconfig_fd;
+	char result = 0;
+	prodconfig_fd = open(PRODCONFIG_DRVPATH, O_RDWR);
+	if (prodconfig_fd < 0) {
+		printf("Fail to open prodconfig, errno %d.\n", get_errno());
+		return ERROR;
+	}
+
+	(void)read(prodconfig_fd, &result, sizeof(char));
+	if (result & SWD_ENABLED) {
+		printf("HW Debug : On\n");
+	} else {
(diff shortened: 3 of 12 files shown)


apps/system/utils/utils_hwconfig.c
```diff
@@ -0,0 +1,90 @@
...
+#include <stdio.h>
+#include <fcntl.h>
+
+#include <tinyara/hwconfig.h>
+#include <tinyara/fs/ioctl.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+int utils_hwconfig(int argc, char **args)
```
deepaksrma, line 66: could you give a description of this to understand it better?

jeongchanKim: Added. Thanks.

apps/system/utils/utils_hwconfig.c
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * apps/system/utils/utils_hwconfig.c
```
sunghan-chang, line 19: Nuttx license is unnecessary. Let's remove.

apps/system/utils/Kconfig
```diff
@@ -140,6 +140,13 @@ config HEAPINFO_USER_GROUP_LIST
 endif
 endif
 
+config ENABLE_HWCONFIG
```
sunghan-chang, line 143: This is too common name.  How about `PRODCONFIG`?

apps/system/utils/utils_hwconfig.c
```diff
@@ -0,0 +1,99 @@
...
+{
+	int hwconfig_fd;
+	int ret;
+	hwconfig_fd = open(HWCONFIG_DRVPATH, O_RDWR);
+	if (hwconfig_fd < 0) {
+		printf("Fail to open hwconfig, errno %d.\n", get_errno());
+		return ERROR;
+	}
+
+	ret = ioctl(hwconfig_fd, HWCONFIGIOC_SWD, 0);
```
sunghan-chang, line 85: How about using `read` instead of `ioctl`? Because it has only read operation.

apps/system/utils/utils_hwconfig.c
```diff
@@ -0,0 +1,99 @@
...
+	if (ret == SWD_ENABLED) {
```
sunghan-chang, line 86: downloader and hw debug should be separated.

apps/system/utils/systemcmd.c
```diff
@@ -43,6 +43,9 @@ const static tash_cmdlist_t system_cmds[] = {
 #if defined(CONFIG_ENABLE_HEAPINFO)
 	{"heapinfo", utils_heapinfo,     TASH_EXECMD_SYNC},
 #endif
+#if defined(CONFIG_ENABLE_PRODCONFIG)
+	{"prodconfig", utils_prodconfig,     TASH_EXECMD_SYNC},
```
sunghan-chang, line 47: Let's list it by Alphabet order

apps/system/utils/utils_hwconfig.c
sunghan-chang: how about this?

jeongchanKim: We can read one of `SWD` or `DOWNLOAD` separately.
So I think ioctl is better.

os/board/rtl8721csm/src/rtl8721csm_boot.c
```diff
@@ -85,6 +91,29 @@ extern FAR struct gpio_lowerhalf_s *amebad_gpio_lowerhalf(u32 pinname, u32 pinmo
 /************************************************************************************
  * Public Functions
  ************************************************************************************/
+#ifdef CONFIG_PRODCONFIG
+int up_prodswd(void)
```
sunghan-chang, line 95: 1. They are the same. Let's keep only one function. It could be `up_check_swd`.
2. The driver calls swd and download checks separately. But only this board can handle them using one function, up_check_swd. Let's redirect them to up_check_swd.

apps/system/utils/utils_prodconfig.c
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	ret = ioctl(prodconfig_fd, PRODCONFIGIOC_DOWNLOAD, 0);
+	if (ret == DOWNLOAD_ENABLED) {
+		printf("Download : Downloader & Hotfix Available\n");
+	} else if (ret == DOWNLOAD_DISABLED) {
+		printf("Download : Hotfix Only Available\n");
+	} else {
+		printf("Fail to check download config.\n");
+	}
```
sunghan-chang, line 67: They will be used only in this util so that we can use `read` operation instead of `ioctl`. If we use the read, no need additional definitions in ioctl.

apps/system/utils/Kconfig
```diff
@@ -140,6 +140,13 @@ config HEAPINFO_USER_GROUP_LIST
 endif
 endif
 
+config ENABLE_PRODCONFIG
+	bool "prodconfig"
+	depends on PRODCONFIG
+	default y
```
sunghan-chang, line 146: This functionality is not mandatory. Let's set it as n.

os/board/rtl8721csm/src/rtl8721csm_boot.c
```diff
@@ -85,6 +91,25 @@ extern FAR struct gpio_lowerhalf_s *amebad_gpio_lowerhalf(u32 pinname, u32 pinmo
...
  * Public Functions
  ************************************************************************************/
+#ifdef CONFIG_PRODCONFIG
+static void up_check_prod(char *value)
+{
+	u8 prod_disable;
+	EFUSERead8(0, EFUSE_SEC_CONFIG_ADDR1, &prod_disable, L25EOUTVOLTAGE);
+	if (prod_disable == 0xff) {
+		*value |= SWD_ENABLED;
+		*value |= DOWNLOAD_ENABLED;
```
sunghan-chang, line 101: Is this necessary to be bit handling?
This function should be implemented by chipset vendor. It's not good to understand to port.
```
int up_check_prodswd(void);
int up_check_proddownload(void);
{
 return up_check_prod(void);
}
```
It's ok to bit handling in the driver side because chipset vendor doesn't need to know.

os/drivers/Kconfig
```diff
@@ -418,6 +418,14 @@ config MMINFO
 		This selection enables building of the "upper-half" mminfo
 		driver.
 
+config PRODCONFIG
+	bool "Product config Driver Support"
+	default y
```
sunghan-chang, line 423: same as previous

sunghan-chang (approved):

PR #5121, opened 2021-08-17
tools/version.sh : Cut the commit hash as 7 strings

jeongchanKim:
Commit hash is very long, so cut it as 7 strings.


sunghan-chang:
How about below to give more exact information of the commit?
```
tools/version.sh : get first 7 chars of commit hash 

7 characters are enough to identify the commit, let's get only 7 chars instead of full hash to save memory and printing time.
```

sunghan-chang (approved):

PR #5127, opened 2021-08-19
apps/system/utils/Kconfig : Add dependency of PROCFS in StackMonitor

jeongchanKim:
StackMonitor only works when FS_PROCFS is enabled.

### DIFF
--- a/apps/system/utils/Kconfig
+++ b/apps/system/utils/Kconfig
@@ -181,7 +181,7 @@ config ENABLE_PS
 config ENABLE_STACKMONITOR
 	bool "Stack monitor"
 	default n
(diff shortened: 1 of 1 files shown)


apps/system/utils/Kconfig
```diff
@@ -182,6 +182,7 @@ config ENABLE_STACKMONITOR
 	bool "Stack monitor"
 	default n
 	depends on !FS_PROCFS_EXCLUDE_PROCESS
+	depends on FS_PROCFS
```
sunghan-chang, line 185: How about merging with above like others?
```
depends on FS_PROCFS && !FS_xxxxx
```

sunghan-chang (approved):

PR #5129, opened 2021-08-19
os/board/rtl8721csm : Improve Wi-Fi join performance for Easy setup

keerthanalex:
The full scan result is stored by default for 3 mins(SCAN_TIMER_DURATION) and timer is used to track it
The channel information is retrieved by comparing stored scan result with AP details from application
The partial scan is set up according to channel information which improves the join performance

### DIFF
--- a/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
+++ b/os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
@@ -414,7 +414,7 @@ int8_t cmd_wifi_ap(trwifi_softap_config_s *softap_config)
 	return ret;
 }
 
-int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
+int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg, uint32_t ap_channel)
 {
 	int ret;
 
@@ -426,6 +426,7 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 	int key_id = 0;
 	void *semaphore;
 	int security_retry_count = 0;
+	uint8_t pscan_config;
 
 	trwifi_ap_auth_type_e auth = ap_connect_config->ap_auth_type;
 	ssid = ap_connect_config->ssid;
@@ -495,6 +496,15 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 	if (wifi_on(RTW_MODE_STA) < 0) {
 		ndbg("\n\rERROR: Wifi on failed!");
 		return -1;
+	}
+
+	if ((ap_channel >= 1) && (ap_channel <= 13)) {
+		pscan_config = PSCAN_ENABLE | PSCAN_FAST_SURVEY;
+		ret = wifi_set_pscan_chan((uint8_t *)&ap_channel, &pscan_config, 1);
+
+		if (ret < 0) {
+			RTW_API_INFO("\n\rset pscan failed");
+		}
 	}
 
 	ret = wifi_connect(ssid,
--- a/os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
+++ b/os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
@@ -25,6 +25,8 @@
 #include <tinyara/lwnl/lwnl.h>
 #include <tinyara/net/if/wifi.h>
 #include <tinyara/netmgr/netdev_mgr.h>
+#include "freertos/wrapper.h"
+#include "osdep_service.h"
 /* WLAN CONFIG ---------------------------------------------------------------*/
 #define RTK_OK          0		/*!< RTK_err_t value indicating success (no error) */
 #define RTK_FAIL        -1		/*!< Generic RTK_err_t code indicating failure */
@@ -82,15 +84,88 @@ static trwifi_scan_list_s *g_scan_list;
 static int g_scan_num;
 extern struct netdev *ameba_nm_dev_wlan0;
 
+#define SCAN_TIMER_DURATION 180000
+
+struct ap_scan_list {
+	unsigned int channel;
+	char ssid[TRWIFI_SSID_LEN + 1];
+	unsigned int ssid_length;
+	int rssi;
+	trwifi_ap_auth_type_e ap_auth_type;
+	trwifi_ap_crypto_type_e ap_crypto_type;
+};
+typedef struct ap_scan_list ap_scan_list_s;
+
+static ap_scan_list_s *saved_scan_list = NULL;
+static uint32_t scan_number = 0;
+static _timer scan_timer;
+static _mutex scanlistbusy;	//protects shared variables saved_scan_list,scan_number and scan_timer
+
+static void _scan_timer_handler(void *FunctionContext)
+{
+	vddbg("scan Timer expired : release saved scan list\r\n");
+
+	rtw_mutex_get(&scanlistbusy);
+	vddbg("scan Timer expired : sizeof(ap_scan_list_s) =%d, scan_number=%d \r\n", sizeof(ap_scan_list_s), scan_number);
+
+	if (saved_scan_list)
+		rtw_mfree((unsigned char *)saved_scan_list, sizeof(ap_scan_list_s) * scan_number);
+	scan_number = 0;
+	rtw_del_timer(&(scan_timer));
+	rtw_mutex_put(&scanlistbusy);
+}
+
+static int save_scan_list(trwifi_scan_list_s *p_scan_list)
+{
+	rtw_mutex_get(&scanlistbusy);
+	// If application calls scan before scan result free(before timeout), release scan list and cancel timer
+	if (scan_number) {
+		if (saved_scan_list)
+			rtw_mfree((unsigned char *)saved_scan_list, sizeof(ap_scan_list_s) * scan_number);
+		rtw_cancel_timer(&(scan_timer));
+		vddbg("scan is called before timeout\r\n");
+	} else {
+		rtw_init_timer(&(scan_timer), &(scan_timer), _scan_timer_handler, &(scan_timer), "dynamic_chk_timer");
+		vddbg("Start scan timer\n");
+	}
+
+	rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
+
+	scan_number	 = 0;
+	saved_scan_list = (ap_scan_list_s *)rtw_malloc(sizeof(ap_scan_list_s) * g_scan_num);
+	if (saved_scan_list == NULL) {
+		rtw_mutex_put(&scanlistbusy);
+		return RTW_NOMEM;
+	}
+
+	while(p_scan_list) {
+		strncpy(saved_scan_list[scan_number].ssid, p_scan_list->ap_info.ssid, p_scan_list->ap_info.ssid_length);
+		saved_scan_list[scan_number].ssid_length = p_scan_list->ap_info.ssid_length;
+		saved_scan_list[scan_number].channel = p_scan_list->ap_info.channel;
+		saved_scan_list[scan_number].rssi = p_scan_list->ap_info.rssi;
+		saved_scan_list[scan_number].ap_auth_type = p_scan_list->ap_info.ap_auth_type;
+		saved_scan_list[scan_number].ap_crypto_type = p_scan_list->ap_info.ap_crypto_type;
+		p_scan_list = p_scan_list->next;
+		scan_number++;
+	}
+
+	if (scan_number != g_scan_num)
+		vddbg("Total scan list temp = %d, g_scan = %d\n", scan_number, g_scan_num);
+	rtw_mutex_put(&scanlistbusy);
+
+	return RTW_SUCCESS;
+}
+
 static void _free_scanlist(void)
 {
 	while (g_scan_list) {
 		trwifi_scan_list_s *cur = g_scan_list;
 		g_scan_list = g_scan_list->next;
-		rtw_mfree(cur, sizeof(trwifi_scan_list_s));
+		rtw_mfree((unsigned char *)cur, sizeof(trwifi_scan_list_s));
 	}
 	g_scan_num = 0;
 }
+
 rtw_result_t app_scan_result_handler(rtw_scan_handler_result_t *malloced_scan_result)
 {
 	trwifi_scan_list_s *scan_list;
@@ -244,6 +319,12 @@ rtw_result_t app_scan_result_handler(rtw_scan_handler_result_t *malloced_scan_re
 		}
 	} else {
 		nvdbg("SCAN DONE: Calling wifi_scan_result_callback\r\n");
+		int res;
+		trwifi_scan_list_s *p_scan_list = g_scan_list;
+
+		res = save_scan_list(p_scan_list);
+		if (res != RTW_SUCCESS)
+			vddbg("\r\nFail to malloc scan_list\r\n");
 		TRWIFI_POST_SCANEVENT(ameba_nm_dev_wlan0, LWNL_EVT_SCAN_DONE, (void *)g_scan_list);
 		_free_scanlist();
 		if (g_scan_list) {
@@ -370,6 +451,7 @@ trwifi_result_e wifi_netmgr_utils_init(struct netdev *dev)
 		}
 		g_mode = RTK_WIFI_STATION_IF;
 		wuret = TRWIFI_SUCCESS;
+		rtw_mutex_init(&scanlistbusy);
 	} else {
 		ndbg("Already %d\n", g_mode);
 	}
@@ -384,6 +466,7 @@ trwifi_result_e wifi_netmgr_utils_deinit(struct netdev *dev)
 	if (ret == RTK_STATUS_SUCCESS) {
 		g_mode = RTK_WIFI_NONE;
 		wuret = TRWIFI_SUCCESS;
+		rtw_mutex_free(&scanlistbusy);
 	} else {
 		ndbg("[RTK] Failed to stop STA mode\n");
 	}
@@ -422,6 +505,7 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
 	}
 
 	int ret;
+	uint32_t ap_channel;
 	wuret = TRWIFI_FAIL;
 
 	if (g_mode == RTK_WIFI_SOFT_AP_IF) {
@@ -436,7 +520,24 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
 		}
 	}
 
-	ret = cmd_wifi_connect(ap_connect_config, arg);
+	ap_channel = 0xffff;
+
+	rtw_mutex_get(&scanlistbusy);
+	if (scan_number) {
+		int i;
+		for (i = 0 ; i < scan_number ; i++) {
+			if((strncmp(saved_scan_list[i].ssid, ap_connect_config->ssid, ap_connect_config->ssid_length) == 0)
+				&& (ap_connect_config->ap_auth_type == saved_scan_list[i].ap_auth_type)
+				&& (ap_connect_config->ap_crypto_type == saved_scan_list[i].ap_crypto_type)) {
+				ap_channel = saved_scan_list[i].channel;
+			vddbg("[RTK] Scanned AP info : ssid = %s, auth = %d, crypt = %d, channel = %d\n", ap_connect_config->ssid, ap_connect_config->ap_auth_type, ap_connect_config->ap_crypto_type, ap_channel);
+				break;
+			}
+		}
+	}
+	rtw_mutex_put(&scanlistbusy);
+
+	ret = cmd_wifi_connect(ap_connect_config, arg, ap_channel);
 	if (ret != RTK_STATUS_SUCCESS) {
 		ndbg("[RTK] WiFiNetworkJoin failed: %d, %s\n", ret, ap_connect_config->ssid);
 		return wuret;
--- a/os/board/rtl8721csm/src/component/os/tizenrt/rtk_wifi_utils.h
+++ b/os/board/rtl8721csm/src/component/os/tizenrt/rtk_wifi_utils.h
@@ -87,7 +87,7 @@ extern int8_t WiFiRegisterLinkCallback(rtk_network_link_callback_t link_up, rtk_
 extern int8_t cmd_wifi_on(WiFi_InterFace_ID_t interface_id);
 extern int8_t cmd_wifi_off(void);
 extern int wifi_scan_networks(rtw_scan_result_handler_t results_handler, void* user_data);
-extern int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg);
+extern int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg, uint32_t ap_channel);
 extern int wifi_get_mac_address(char * mac);
 extern int wifi_is_connected_to_ap(void);
 extern int wifi_get_rssi(int *pRSSI);


os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
```diff
@@ -488,6 +490,15 @@ int8_t cmd_wifi_connect(trwifi_ap_config_s *ap_connect_config, void *arg)
 		return -1;
 	}
 
+	if ((ap_channel >= 1) && (ap_channel <= 13)) {
+		pscan_config = PSCAN_ENABLE | PSCAN_FAST_SURVEY;
+		ret = wifi_set_pscan_chan((uint8_t *)&ap_channel, &pscan_config, 1);
+
+		if (ret < 0) {
+			RTW_API_INFO("\n\rset pscan failed");
```
jeongchanKim, line 498: Is it okay not to return as fail like `-1`?

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,86 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+};
+typedef struct ap_scan_list ap_scan_list_s;
+
+static ap_scan_list_s *saved_scan_list;
+static uint32_t scan_number = 0;
+uint32_t ap_channel;
+
+#include "freertos/wrapper.h"
+#define SCAN_TIMER_DURATION 180000
+_timer scan_timer;
```
jeongchanKim, line 102: If this variable is used only in this file, please add `static`.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,86 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+		rtw_init_timer(&(scan_timer), &(scan_timer), _scan_timer_handler, &(scan_timer), "dynamic_chk_timer");
+		ndbg("Start scan timer\n");
+	}
+
+	rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
+
+	scan_number	 = 0;
+	saved_scan_list = (ap_scan_list_s *)rtw_zmalloc(sizeof(ap_scan_list_s) * g_scan_num);
+	if (saved_scan_list == NULL)
+		return RTW_NOMEM;
```
jeongchanKim, line 132: Is it okay not to stop or cancel the timer?

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,86 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+		saved_scan_list[scan_number].ap_auth_type = p_scan_list->ap_info.ap_auth_type;
+		saved_scan_list[scan_number].ap_crypto_type = p_scan_list->ap_info.ap_crypto_type;
+		p_scan_list = p_scan_list->next;
+		scan_number++;
+	}
+
+	if (scan_number != g_scan_num)
+		ndbg("Total scan list temp = %d, g_scan = %d\n", scan_number, g_scan_num);
+
+#if 0
```
jeongchanKim, line 148: If these lines are not used anymore, please remove them.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: The stop timer "rtw_cancel_timer()" has been used for two cases.
1) The timer is cancelled when it searches for specific AP detail from stored scan result by comparing SSID and set the timer again once if the timer is not deleted -> `wifi_netmgr_utils_connect_ap`
2) If the application calls scan before scan result free(before timeout), release scan list and cancel timer -> `save_scan_list()`

os/board/rtl8721csm/src/component/common/api/wifi_interactive_mode.c
keerthanalex: The `wifi_set_pscan_chan()` set the partial scan and driver scans only that channels set by this API.
If this API returns '-1' due to some reason, it only means driver will not do partial scan and continue with full scan.
So we don't really have to stop connect wifi just because `wifi_set_pscan_chan()` is failed.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,81 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+		ndbg("scan is called before timeout\r\n");
+	} else {
+		rtw_init_timer(&(scan_timer), &(scan_timer), _scan_timer_handler, &(scan_timer), "dynamic_chk_timer");
+		ndbg("Start scan timer\n");
+	}
+
+	rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
+
+	scan_number	 = 0;
+	saved_scan_list = (ap_scan_list_s *)rtw_zmalloc(sizeof(ap_scan_list_s) * g_scan_num);
```
sunghan-chang, line 130: Do you need to use zmalloc? After getting the pointer, you input values for entire members.
I think in your usecase, just malloc is better for the performance. Could you check?

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -437,6 +509,25 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
 		}
 	}
 
+	ap_channel = 0xffff;
+
+	if (scan_number) {
+		int i;
+		rtw_cancel_timer(&(scan_timer));
```
j11591, line 516: Why does it cancel timer when it try to connect???

j11591:
"The specific AP detail along with channel information is retrieved from stored scan result by comparing SSID"
What is the specific AP detail along with channel information??? specific scan???

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: It is to make sure that the timer is not expired during the ap connection.
if the timer is running, it may have possibility of timer expired.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: I'll change it to malloc and update with other review comments.

keerthanalex:
> "The specific AP detail along with channel information is retrieved from stored scan result by comparing SSID"
> What is the specific AP detail along with channel information??? specific scan???

I meant that when application wants to connect to particular SSID, it may not send all the parameters (SSID,PWD,Auth,crypto,channel) required for wifi connection. Channel information is required to set the partial scan. In our implementation, once we get the specific SSID from application we search for it's related information from our scan result.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
j11591: > 		if (scan_number)
			rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
Then it is possible to extend lifetime of scan_list forever with connect try.
How about using mutex for shared resource (saved_scan_list)??

keerthanalex: @j11591 , The rtw_enter_critical/rtw_exit_critical API's can be added to prevent scan timer timeout(which frees up  saved_scan_list and delete the timer). Are there any more review comments? I'll update all in single commit.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
davidfather: @sunghan-chang @keerthanalex I think the ssid length varies, so that zmalloc might be better for cleaning used ssids or any garbages.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,81 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+		saved_scan_list[scan_number].channel = p_scan_list->ap_info.channel;
+		saved_scan_list[scan_number].rssi = p_scan_list->ap_info.rssi;
+		saved_scan_list[scan_number].ap_auth_type = p_scan_list->ap_info.ap_auth_type;
+		saved_scan_list[scan_number].ap_crypto_type = p_scan_list->ap_info.ap_crypto_type;
+		p_scan_list = p_scan_list->next;
+		scan_number++;
+	}
+
+	if (scan_number != g_scan_num)
+		ndbg("Total scan list temp = %d, g_scan = %d\n", scan_number, g_scan_num);
```
davidfather, line 146: Could you explain why this case is checked? When does this case happen? Do you need any measures to resolve this situation?

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
davidfather: @keerthanalex Before canceling the timer, I think you need to store the remaining duration and to reset the timer with this duration after connecting AP.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @davidfather, `g_scan_num` is the total number of list on `p_scan_list`. We call `save_scan_list()` after the getting the final number of `g_scan_num` and before `TRWIFI_POST_SCANEVENT()`. 
Here with this condition, we are just double checking whether we have copied the whole of `p_scan_list` to `saved_scan_list` by comparing both it's number. The `g_scan_num = 0;` will also be done after `TRWIFI_POST_SCANEVENT()` inside `_free_scanlist();`. Hence we don't have to take any measures for this case.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Then I'll retain `rtw_zmalloc` in this case.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: The rtw_enter_critical/rtw_exit_critical API's may be excessive, so we plan to go with mutex to protect the `saved_scan_list`.
@davidfather, we only need to protect the list from getting deleted by timeout. Hence we can skip stop/reset timer and use mutex for this case.
But we would to like to get suggestion on adding `rtw_mutext_init` where it needs to be called only once before using `rtw_mutext_get()` and `rtw_mutext_put()`.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
sunghan-chang: @davidfather @keerthanalex There is the member, ssid_length so that it should be ok with non-initialized buffer.
```
        strncpy(saved_scan_list[scan_number].ssid, p_scan_list->ap_info.ssid, p_scan_list->ap_info.ssid_length);
        saved_scan_list[scan_number].ssid_length = p_scan_list->ap_info.ssid_length;
```
If we should use zmalloc, then I think `ssid_length` is unnecessary.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,81 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+	}
+
+	rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
+
+	scan_number	 = 0;
+	saved_scan_list = (ap_scan_list_s *)rtw_zmalloc(sizeof(ap_scan_list_s) * g_scan_num);
+	if (saved_scan_list == NULL)
+		return RTW_NOMEM;
+
+    while(p_scan_list) {
```
pillip8282, line 134: it's space indentation. please fix change to tab.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -437,6 +509,25 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
...
 	}
 
+	ap_channel = 0xffff;
+
+	if (scan_number) {
+		int i;
+		rtw_cancel_timer(&(scan_timer));
+		for (i = 0 ; i < scan_number ; i++) {
+			if(strncmp(saved_scan_list[i].ssid, ap_connect_config->ssid, ap_connect_config->ssid_length) == 0) {
+				ap_connect_config->ap_auth_type = saved_scan_list[i].ap_auth_type;
```
pillip8282, line 519: An application might pass ap_auth_type and crypto_type. according to this, driver overwrite application value. I think this is wrong. So you have to compare ap_auth type and crypto type too.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -437,6 +509,25 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
 		}
 	}
 
+	ap_channel = 0xffff;
```
pillip8282, line 512: how about to pass it as function parameter. the use of global variable should be minimized.

pillip8282 (changes requested):
if driver callback(handle scan result) is executing save_scanlist and it writes to saved_scanlist, while timer run scan_timer_handle and free saved_scanlist. then it cause unexpected behavior. So you should protect saved_scanlist. Please check the case.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -83,15 +83,81 @@ static wifi_utils_scan_list_s *g_scan_list;
...
+		rtw_mfree((unsigned char *)saved_scan_list, sizeof(ap_scan_list_s) * scan_number);
+		rtw_cancel_timer(&(scan_timer));
+		ndbg("scan is called before timeout\r\n");
+	} else {
+		rtw_init_timer(&(scan_timer), &(scan_timer), _scan_timer_handler, &(scan_timer), "dynamic_chk_timer");
+		ndbg("Start scan timer\n");
+	}
+
+	rtw_set_timer(&(scan_timer), SCAN_TIMER_DURATION);
+
```
keerthanalex, line 128: >  if driver callback(handle scan result) is executing save_scanlist and it writes to saved_scanlist, while timer run scan_timer_handle and free saved_scanlist. then it cause unexpected behavior. So you should protect saved_scanlist. Please check the case.

@pillip8282 , The `saved_scan_list` struct has been used in three cases 
1) The write is done inside `save_scan_list()`, but we start the timer with `SCAN_TIMER_DURATION` before every write.
So there won't be immediate any timeout which will free the struct to interrupt the write function and in my opinion there is no need to protect the write to `saved_scan_list` struct. what is your suggestion?

2) The search is done inside `wifi_netmgr_utils_connect_ap()` and there is chance of timeout handler to free the struct during search. We already planned to add mutex to protect the `saved_scan_list` here but we would to like to get suggestion on adding `rtw_mutext_init` where it needs to be called only once before using `rtw_mutext_get()` and `rtw_mutext_put()`. How about mutex init during `wifi_on()`?

3) The free is done inside `_scan_timer_handler()`.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @sunghan-chang, then we need to follow either one of the below
1) Retain zmalloc and remove `ssid_length` from struct `ap_scan_list`. The `ssid_length` parameter is not really used anywhere else apart from this place. 

- For writing to `ssid` of struct `ap_scan_list`, the `p_scan_list` ssid_length is used.
- In case of search in `wifi_netmgr_utils_connect_ap()`, the `ap_connect_config` ssid_length is used.

2) Modify  zmalloc to malloc and retain `ssid_length`. But as I mentioned in above point, the `ssid_length` is not used for any operation.
which one do you suggest?

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
pillip8282: * About case 1 consider this scenario.
1) wifi_manager requests scan.
2) a driver calls save_scan_list and set a timer
3) after 3minutes later, a timer is expired. so the timer execute rtw_mfree in _scan_timer_handler().
4) context switching, wifi_manager request scan.
5) driver callback execute save_scan_list, and _scan_timer_handler wasn't finished it'll call rtw_mfree again.

* I think that initializing a mutex in wifi_netmgr_utils_init(struct netdev *dev) and deinit it in wifi_netmgr_utils_deinit()

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @pillip8282, Does that mean we only search the full scan list for channel information to set the partial scan to improve the performance?
In this case, how do we handle the case of modification of channel before the expiry of scan duration(refresh of scan list)? Because it will set the partial scan with previous channel and results in wifi_connect failure. 

**Suggested solution:** since we only pick the channel from scan list, we can remove that particular entry in case of wifi_connect failure. So that for the second call from application, it will skip the partial scan part and does the normal wifi connection.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @pillip8282, Ok I understand the scenario and will implement mutex as per your suggestion.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @pillip8282  Modified as suggested, please help to review.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: > An application might pass ap_auth_type and crypto_type. according to this, driver overwrite application value. I think this is wrong. So you have to compare ap_auth type and crypto type too.

Modified as suggested, Please help to review.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
sunghan-chang: It's not proper to suggest ssid_length usages because I am not in charge of network. I just want to check necessity of zmalloc. But in this function point of view, I don't think zmalloc is necessary.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -82,15 +82,98 @@ static trwifi_scan_list_s *g_scan_list;
...
+static ap_scan_list_s *saved_scan_list;
+static uint32_t scan_number = 0;
+static uint32_t ap_channel;
+
+#include "freertos/wrapper.h"
+#include "osdep_service.h"
+#define SCAN_TIMER_DURATION 180000
+static _timer scan_timer;
+static _mutex scanlistbusy;
```
pillip8282, line 103: it would be better to comment what shared variables does this protect to.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -436,6 +527,23 @@ trwifi_result_e wifi_netmgr_utils_connect_ap(struct netdev *dev, trwifi_ap_confi
 		}
 	}
 
+	ap_channel = 0xffff;
+
+	if (scan_number) {
+		int i;
+		rtw_mutex_get(&scanlistbusy);
+		for (i = 0 ; i < scan_number ; i++) {
+			if((strncmp(saved_scan_list[i].ssid, ap_connect_config->ssid, ap_connect_config->ssid_length) == 0)
```
pillip8282, line 536: if scan_number is not protected then this situation might be happened.

1. wifi_manager requests scan.
2. a driver calls save_scan_list and set a timer
3. after 3minutes later, a timer is expired. so the timer execute rtw_mfree in _scan_timer_handler().
4. before assign scan_number = 0 in _scan_timer_handler() context switching happen.
5. wifi_manager request connect. and go inside if(scan_number) because scan_number is not 0.
6. connec function would access invalid memory (saved_scan_list[i]).

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
pillip8282: you misunderstood my point. I think make ap_channel to a local variable. and pass it to cmd_wifi_connect() like this.
```
ret = cmd_wifi_connect(ap_connect_config, arg, ap_channel);
```

pillip8282 (changes requested):
it would be better to consider how to protect to shared variable(global variable) in multi-thread environment.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @pillip8282 , Modified the API as per suggestion and made ap_channel as local variable.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @pillip8282, Modified to protect three shared variables of latest added code with `_mutex scanlistbusy`

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: @sunghan-chang, modified to malloc.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
sunghan-chang: @pillip8282 Could you confirm this?

pillip8282:
@keerthanalex 
Now we're testing your PR. after it passed our TCs we'll merge it.

keerthanalex:
> @keerthanalex
> Now we're testing your PR. it'll take time because we'll run stress test and etc. after it passed our TCs we'll merge it.



> @keerthanalex
> Now we're testing your PR. it'll take time because we'll run stress test and etc. after it passed our TCs we'll merge it.

Ok, thank you.

keerthanalex:
> @keerthanalex
> Now we're testing your PR. it'll take time because we'll run stress test and etc. after it passed our TCs we'll merge it.

Ok @pillip8282 , thank you.

j11591:
@keerthanalex 
Sorry for late test...
I have tested join performance.
There are 4 type join performance. Please check it and tell the difference.

1. First join time without scan after reboot : about 2.4s 
2. Join time after second try without scan : about 1.78s
- Test scenario is reboot -> join(2.4s) -> disconnect -> join(1.78s) -> disconnect -> join(1,78s) ->.....

3. First join time after reboot and scan : about 0.99s
4. Join time after second try after scan : about 0.34s
- Test scenario is reboot -> scan -> join(0.99s) -> disconnect -> join(0.34s) -> disconnect -> join(0.34s) ->.....

According to my request, if there is scan information at least once, 
it is expected that only two types join performance will occur because the scan information will be used to access it. 
It is expected that 1 case will take a long time and the remaining 2.3.4 case will take the same short time. 
Please explain why there are four cases.

keerthanalex:
> @keerthanalex Sorry for late test... I have tested join performance. There are 4 type join performance. Please check it and tell the difference.
> 
> 1. First join time without scan after reboot : about 2.4s
> 2. Join time after second try without scan : about 1.78s
> 
> * Test scenario is reboot -> join(2.4s) -> disconnect -> join(1.78s) -> disconnect -> join(1,78s) ->.....
> 
> 3. First join time after reboot and scan : about 0.99s
> 4. Join time after second try after scan : about 0.34s
> 
> * Test scenario is reboot -> scan -> join(0.99s) -> disconnect -> join(0.34s) -> disconnect -> join(0.34s) ->.....
> 
> According to my request, if there is scan information at least once, it is expected that only two types join performance will occur because the scan information will be used to access it. It is expected that 1 case will take a long time and the remaining 2.3.4 case will take the same short time. Please explain why there are four cases.

@j11591, sorry for the late response. We already clarified the reason to Eric for the time difference between join after reboot and disconnect.

j11591 (dismissed):
LGTM

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -82,15 +82,90 @@ static trwifi_scan_list_s *g_scan_list;
 static int g_scan_num;
 extern struct netdev *ameba_nm_dev_wlan0;
 
+#include "freertos/wrapper.h"
```
pillip8282, line 85: how about to move include to line 28. I think gathering header will be better to maintain code.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -82,15 +82,90 @@ static trwifi_scan_list_s *g_scan_list;
...
+	unsigned int channel;
+	char ssid[TRWIFI_SSID_LEN + 1];
+	unsigned int ssid_length;
+	int rssi;
+	trwifi_ap_auth_type_e ap_auth_type;
+	trwifi_ap_crypto_type_e ap_crypto_type;
+};
+typedef struct ap_scan_list ap_scan_list_s;
+
+static ap_scan_list_s *saved_scan_list;
```
pillip8282, line 99: how about to initialize the variable.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -82,15 +82,90 @@ static trwifi_scan_list_s *g_scan_list;
...
+static uint32_t scan_number = 0;
+static _timer scan_timer;
+static _mutex scanlistbusy;	//protects shared variables saved_scan_list,scan_number and scan_timer
+
+static void _scan_timer_handler(void *FunctionContext)
+{
+	ndbg("scan Timer expired : release saved scan list\r\n");
+
+	rtw_mutex_get(&scanlistbusy);
+	ndbg("scan Timer expired : sizeof(ap_scan_list_s) =%d, scan_number=%d \r\n", sizeof(ap_scan_list_s), scan_number);
```
pillip8282, line 109: I requested eric.kim that wi-fi driver code use vddbg instead of ndbg during reviewing. So could you change ndbg to vddbg?

keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

os/board/rtl8721csm/src/component/os/tizenrt/rtk_netmgr.c
keerthanalex: Done.

pillip8282 (approved):
LGTM

PR #5133, opened 2021-08-20
apps/example/fs_test: Clean examples for fs test

jeongarmy:
- Unify duplicated examples for same feature
- Rename ambiguous names of functions and variables
- Remove usages of global variable for safety
- Use enum type for options for scalability
- Fix wrong result checking
- Clean codes


apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+	while (total_write < nbytes) {
+		ret = write(fd, buf + total_write, nbytes - total_write);
+		if (ret != nbytes - total_write) {
+			printf("%s errno: %d\n", __FUNCTION__, errno);
+			close(fd);
+			return ERROR;
+		}
+		total_write += ret;
+	}
+	timediff = get_time() - time;
```
sunghan-chang, line 126: The get_time works with the gettimeofday. But it works based on the tick. I don't know it's proper resolution of this test. Could you check the unit of this test?

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+	case TEST_OPT_WRITEREMOVE:
+		ret = write_remove_test(filepath, buf, size, iteration);
+		break;
+
+	default:
+		break;
+	}
+	free(buf);
+
+	if (ret == OK) {
```
sunghan-chang, line 294: If the `operation` has higher than MAX, the `ret` has garbage value. 
It should be initialized with ERROR or, should be set at the default case.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+		printf("%s", USAGE);
+		return -1;
+	}
+
+	if (strlen(argv[5]) != 1 || (g_printopt != 'y' && g_printopt != 'n')) {
+		printf("%s", USAGE);
+		return -1;
+	}
+
+	run_fs_test(operation, filepath, size, iteration);
```
sunghan-chang, line 334: I don't know this should be separated. The main and this have almost the same purpose.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+#endif
+{
+	int size;
+	int operation;
+	int iteration;
+	char *filepath;
+
+	printf("FS R/W Performance Test!\n");
+
+	if (argc <= 5) {
```
sunghan-chang, line 312: How about 7 or higher for argc?

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+{
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
+	return (tv.tv_sec * 1000 + tv.tv_usec / 1000);
+}
+
+static void print_log(const char *format, ...)
+{
+	va_list ap;
```
jeongchanKim, line 74: For using `va_list`, please add `stdarg.h`.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+		if (ret != nbytes - total_read) {
+			printf("%s errno: %d\n", __FUNCTION__, errno);
+			close(fd);
+			return ERROR;
+		}
+		total_read += ret;
+	}
+	timediff = get_time() - time;
+	g_total_time += timediff;
+	print_log("read:%dms, ", timediff);
```
jeongchanKim, line 161: For long type, it is better to print that value with `%ld` instead of `%d`.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+/****************************************************************************
+ * smartfs_test
+ ****************************************************************************/
+
+static inline long get_time(void)
+{
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
+	return (tv.tv_sec * 1000 + tv.tv_usec / 1000);
```
jeongchanKim, line 69: Is this intended result?
If `tv.tv_sec` is 1, and `tv.tv_usec` is 10, it means the result of `gettimeofday` is 1s10us.
But with your calculation, the result of `get_time()` indicates 1000s 0.001us.
I don't know why you need this time.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+		if (ret != nbytes - total_write) {
+			printf("%s errno: %d\n", __FUNCTION__, errno);
+			close(fd);
+			return ERROR;
+		}
+		total_write += ret;
+	}
+	timediff = get_time() - time;
+	g_total_time += timediff;
+	print_log("write:%dms, ", timediff);
```
jeongchanKim, line 128: I don't know the result of `get_time` is based on ms.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+
+	time = get_time();
+	ret = unlink(filepath);
+	if (ret < 0) {
+		printf("%s errno: %d\n", __FUNCTION__, errno);
+		return ERROR;
+	}
+	timediff = get_time() - time;
+	g_total_time += timediff;
+	print_log("remove:%dms, ", timediff);
```
jeongchanKim, line 182: Same as above. Please check the format and the unit.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+	operation = atoi(argv[1]);
+	filepath = argv[2];
+	size = atoi(argv[3]);
+	iteration = atoi(argv[4]);
+	g_printopt = argv[5][0];
+
+	if (operation < TEST_OPT_WRITEREAD || operation >= TEST_OPT_MAX ||
+		size < 0 || iteration < 1 || iteration > FS_RW_ITERATION_MAX) {
+		printf("%s", USAGE);
+		return -1;
```
jeongchanKim, line 326: How about using `ERROR` instead of `-1`?

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+	if (argc <= 5) {
+		printf("%s", USAGE);
+		return 0;
+	}
+
+	operation = atoi(argv[1]);
+	filepath = argv[2];
+	size = atoi(argv[3]);
+	iteration = atoi(argv[4]);
+	g_printopt = argv[5][0];
```
jeongchanKim, line 321: You checked the `argc` with smaller than equal to 5, but you used the index as 5 in `argv[][]`.
It is not correct.
If `argc` is really smaller than equal to 5, then we should not access the `argv` with index 5.
If not, please change the `argc` checking logic.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,337 @@
...
+ ****************************************************************************/
+#define USAGE                                                           \
+	"\n FS R/W Performance Test\n"                                      \
+	"\n Usage :     smartfs_test <operation> <path> <size> <iteration> <print_option>\n"  \
+	"\nOperation :  1.Read/Write / 2.Read / 3.Write / 4. Write/Remove \n"                  \
+	" Path :        Test File Path\n"                                   \
+	" Size :        Write/Read Data size\n"                             \
+	" Iteration :   Number of repetitions\n"                            \
+	" Detail :      Enable(y) or Disable(n) Printing Details Option\n"  \
+	"               (Printing is affects at R/W Performance)\n"         \
```
jeongchanKim, line 41: There are two verbs, `is` and `affects`.
I think `Printing affects ...` is better.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,332 @@
...
+			ret = read_test(filepath, buf, size, iteration);
+			break;
+		case TEST_OPT_READ:
+			ret = write_test(filepath, buf, size, iteration);
+			break;
+		case TEST_OPT_WRITEREMOVE:
+			ret = write_remove_test(filepath, buf, size, iteration);
+			break;
+		default:
+			ret = ERROR;
```
sunghan-chang, line 318: Because you already check the operation range at line 285, there is no way to be executed here. Let's remove `default`.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,332 @@
...
+			break;
+		}
+		free(buf);
+
+		if (ret == OK) {
+			printf("run_time: %ld ms\n", g_total_time);
+			return OK;
+		}
+		return ERROR;
```
sunghan-chang, line 327: Don't you need to feedback to user for failure?

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,329 @@
...
+			ret = write_remove_test(filepath, buf, size, iteration);
+			break;
+		}
+		free(buf);
+
+		if (ret == OK) {
+			/* Print measured run-time on success */
+			printf("run_time: %ld ms\n", g_total_time);
+		}
+		return OK;
```
sunghan-chang, line 324: Regardless of ret, it returns OK always. I asked you to add failure message. This is not what I want and this is wrong.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,331 @@
...
+			printf("%s", USAGE);
+			return ERROR;
+		}
+
+		if (strlen(argv[5]) != 1 || (g_printopt != 'y' && g_printopt != 'n')) {
+			printf("%s", USAGE);
+			return ERROR;
+		}
+
+		buf = (char *)malloc(size);
```
jeongchanKim, line 296: How about using `zalloc` instead of `malloc`?
Then we don't need to call `memset` in line 302.

apps/examples/fs_test/smartfs_test.c
```diff
@@ -0,0 +1,331 @@
...
+			break;
+		case TEST_OPT_WRITEREMOVE:
+			ret = write_remove_test(filepath, buf, size, iteration);
+			break;
+		}
+		free(buf);
+
+		if (ret == OK) {
+			/* Print measured run-time on success */
+			printf("run_time: %ld ms\n", g_total_time);
```
jeongchanKim, line 322: `g_total_time` is int type.
How about printing with `%d` or changing the type as long?

sunghan-chang (approved):

apps/examples/fs_test/smartfs_test.c
jeongarmy: The unit of this measurement is millisecond(ms). I updated the unit to logs like `print_log("write:%ldms, ", timediff);`

apps/examples/fs_test/smartfs_test.c
jeongarmy: Good. I updated this to `%ld` for long type g_total_time.

apps/examples/fs_test/smartfs_test.c
jeongarmy: I added the log for failure for user. Thank you

apps/examples/fs_test/smartfs_test.c
jeongarmy: I think your opinion is better. I modified.

apps/examples/fs_test/smartfs_test.c
jeongarmy: Right, I modified missing initialization. Thank you!

apps/examples/fs_test/smartfs_test.c
jeongarmy: Good opinion. I modified that things.

apps/examples/fs_test/smartfs_test.c
jeongarmy: I modified

apps/examples/fs_test/smartfs_test.c
jeongarmy: I modified

apps/examples/fs_test/smartfs_test.c
jeongarmy: Good. I changed to `ERROR`

PR #5140, opened 2021-08-24
Append the version to kernel bin name

jeongchanKim:


build/tools/amebad/image_tool/.gitignore
```diff
@@ -1,4 +1,4 @@
 ./km0_boot_all.bin
-./km0_km4_image2.bin
+./km0_km4_image2*.bin
```
thapav, line 2: Why change this ?
Do we only avae km0_km4_image2.bin generated for realtek ?

thapav (approved):
LGTM

build/configs/rtl8721csm/board_metadata.txt
```diff
@@ -23,7 +23,7 @@ DEFAULT_PORT="ttyUSB1"
 
 BL1="km0_boot_all"
 BL2="km4_boot_all"
-KERNEL="km0_km4_image2"
+KERNEL="kernel_tp1x"
 OTA="km0_km4_image2"
```
sangam-swami, line 27: no need to change OTA bin name?

build/configs/rtl8721csm/tc/Make.defs
```diff
@@ -210,10 +211,10 @@ define MAKE_BOARD_SPECIFIC_BIN
   $(TOPDIR)/../build/configs/rtl8721csm/rtl8721csm_make_bin.sh
 endef
 
-KERNEL_BINARY_PATH='$(TOPDIR)/../build/output/bin/km0_km4_image2.bin'
 SECURE_HEADER_SIZE=32
 FLASH_BASE_ADDRESS=0x8000000
 FLASH_SIZE=8388608
+KERNEL_BINARY_PATH='$(TOPDIR)/../build/output/bin/${KERNEL}_${CONFIG_BOARD_BUILD_DATE}.bin'
```
sangam-swami, line 217: KERNEL_BINARY_PATH is name of binary here not path, so we can rename?

jeongchanKim: It includes not only the binary name, but also the path.
We need the path also here, so I named like it, so it is better to keep this.
Thanks.

build/configs/rtl8721csm/hello/Make.defs
```diff
@@ -210,10 +211,10 @@ define MAKE_BOARD_SPECIFIC_BIN
   $(TOPDIR)/../build/configs/rtl8721csm/rtl8721csm_make_bin.sh
 endef
 
-KERNEL_BINARY_PATH='$(TOPDIR)/../build/output/bin/km0_km4_image2.bin'
 SECURE_HEADER_SIZE=32
 FLASH_BASE_ADDRESS=0x8000000
 FLASH_SIZE=8388608
+KERNEL_BINARY_PATH='$(TOPDIR)/../build/output/bin/${KERNEL}_${CONFIG_BOARD_BUILD_DATE}.bin'
```
sunghan-chang, line 217: Is it possible to use the result of `extract_binname`?

os/tools/extract_binname.sh
```diff
@@ -0,0 +1,29 @@
...
+BASE_DIR=$1
+CONFIG_FILE="${BASE_DIR}/.config"
+BUILD_DIR=${BASE_DIR}/../build
+OUTPUT_DIR=${BUILD_DIR}/output/bin
+
+source ${CONFIG_FILE}
+BOARD_DIR=${BUILD_DIR}/configs/${CONFIG_ARCH_BOARD}
+source ${BOARD_DIR}/board_metadata.txt
+
+KERNEL_BIN_PATH=`find ${OUTPUT_DIR} -name ${KERNEL}*`
```
sunghan-chang, line 28: For the file name, how about `get_kbinname.sh`?

build/configs/rtl8721csm/board_metadata.txt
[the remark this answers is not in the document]
jeongchanKim: I already changed. Thanks.

build/tools/amebad/image_tool/.gitignore
```diff
@@ -1,4 +1,4 @@
 ./km0_boot_all.bin
-./km0_km4_image2.bin
+./kernel_tp1x*.bin
```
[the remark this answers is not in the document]
jeongchanKim: Ok. I will change.

build/configs/rtl8721csm/hello/Make.defs
jeongchanKim: No, we cannot use from `extract_binname`.
Because `extract_binname` is working based on the result of build.
But `Make.def` is used to build.

os/tools/extract_binname.sh
jeongchanKim: Okay. I will change.

build/configs/rtl8721csm/tc/Make.defs
sunghan-chang: How about moving this into `define MAKE_SAMSUNG_HEADER`?
Then this will be set after build.

build/configs/rtl8721csm/hello/Make.defs
sunghan-chang: How about moving this into define MAKE_SAMSUNG_HEADER?
Then this will be set after build.

PR #5144, opened 2021-08-25
board/rtl8721csm/src/: Remove duplicated or unused cflags.

fengyunlai:
Remove duplicated or unused cflags for include of headers.


os/board/rtl8721csm/src/Makefile
```diff
@@ -204,12 +197,12 @@ ifeq ($(WINTOOL),y)
   CFLAGS += -I "${shell cygpath -w $(BOARD_SRCDIR)/common}"
   CFLAGS += -I "${shell cygpath -w $(BOARD_SRCDIR)/$(CONFIG_ARCH_BOARD)/src}"
 else
-  CFLAGS += -I$(ARCH_SRCDIR)/chip
+  #CFLAGS += -I$(ARCH_SRCDIR)/chip
```
jeongchanKim, line 200: How about removing instead of comment if it is not used anymore?

jeongchanKim (approved):
LGTM

PR #5147, opened 2021-08-26
arm/src/amebad : Buffer fix for mixed uart logs

abhishek-samsung:
lldbg logs are sent to xmit buffer when not in interrupt context to prevent mixed logs


os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,17 +786,130 @@ int up_lowgetc(void)
...
+			 * logs first.
+			 */
+
+			up_lowputc('\r');
+			up_lowputc('\n');
+
+			char msg_start[] = "***** xmit buffer content start (may contain lldbg logs) ****\n";
+			char msg_end[] = "***** xmit buffer content end ****\n";
+
+			for (int i = 0; msg_start[i] != '\0'; i++) {
```
dineshdas-samsung, line 839: It's better to initialize all the variables at the start...

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,17 +786,130 @@ int up_lowgetc(void)
...
+
+				up_lowputc(dev->xmit.buffer[dev->xmit.tail]);
+
+				/* Increment the tail index */
+				if (++(dev->xmit.tail) >= dev->xmit.size) {
+					dev->xmit.tail = 0;
+				}
+			}
+
+			for (int i = 0; msg_end[i] != '\0'; i++) {
```
dineshdas-samsung, line 854: same as above!

abhishek-samsung:
As using higher-level resources is not allowed in the lower level of the serial driver, this approach is not valid.

PR #5156, opened 2021-08-31
net/lwip: fix lwip pbuf initialization

pillip8282:
initialize pbuf->flags. other member of pbuf are initialized in
pbuf_alloc except flags. So it cause unexpected behavior when lwip
fails to make pbuf chain and free the pbuf.
The scenario is like this
1) pbuf_alloc
2) head of pbuf(p) chain is allocated.
3) fail to allocate chain pbuf.
4) lwip tries to deallocated pbuf p.
5) p->flags contains trash value. it can perform
pc->custom_free_function() in pbuf_free().


sangwon03 (approved):
LGTM

PR #5161, opened 2021-09-03
arch/arm/src/amebad: Buffer fix for mixed uart logs.

abhishek-samsung:
syslog(lldbg) logs are stored in a buffer until we reach end of the line and then the entire line is output by disabling uart interrupts to prevent mixed logs during application context.

example run:

[tc_libc_math_roundl] PASS
tc_libc_misc_lib_dumpbuffer (0x20083b4):

0000: 53414d53554e472d54697a656e5254                                    SAMSUNG-TizenRT
[t
tc_libc_misc_lib_dumpbuffer (0x20042d0):
c_li
0000: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
bc_m
0020: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
ath
0040: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
_sc
0060: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
alb
0080: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
n]
00a0: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
PAS
00c0: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................
S
00e0: 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a 0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a0a ................ ................

[tc_libc_math_scalbnf] PASS

[tc_libc_math_scalbnl] PASS

### DIFF
--- a/os/arch/arm/src/amebad/amebad_serial.c
+++ b/os/arch/arm/src/amebad/amebad_serial.c
@@ -114,7 +114,7 @@
 #elif defined(CONFIG_UART1_SERIAL_CONSOLE)
 #define CONSOLE_DEV             g_uart1port             /* UART1 is console */
 #define TTYS0_DEV               g_uart1port             /* UART1 is ttyS0 */
-#define CONSOLE                 UART3_DEV
+#define CONSOLE                 UART1_DEV
 #define UART1_ASSIGNED  1
 #define HAVE_SERIAL_CONSOLE
 #elif defined(CONFIG_UART2_SERIAL_CONSOLE)
@@ -672,29 +672,7 @@ static bool rtl8721d_up_txready(struct uart_dev_s *dev)
 {
 	struct rtl8721d_up_dev_s *priv = (struct rtl8721d_up_dev_s *)dev->priv;
 	DEBUGASSERT(priv);
-
-#if defined(CONFIG_UART0_SERIAL_CONSOLE)
-	if (uart_index_get(priv->tx) == 0){
-		while (!serial_writable(sdrv[uart_index_get(priv->tx)]));
-		return true;
-	}
-	else
-#elif defined(CONFIG_UART1_SERIAL_CONSOLE)
-	if (uart_index_get(priv->tx) == 3){
-		while (!serial_writable(sdrv[uart_index_get(priv->tx)]));
-		return true;
-	}
-	else
-#elif defined(CONFIG_UART2_SERIAL_CONSOLE)
-	if (uart_index_get(priv->tx) == 2){
-		while (!serial_writable(sdrv[uart_index_get(priv->tx)]));
-		return true;
-	}
-	else
-#endif
-	{
-		return (serial_writable(sdrv[uart_index_get(priv->tx)]));
-	}
+	return (serial_writable(sdrv[uart_index_get(priv->tx)]));
 }
 
 /****************************************************************************
--- a/os/arch/arm/src/amebad/amebad_serial.c
+++ b/os/arch/arm/src/amebad/amebad_serial.c
@@ -786,7 +786,16 @@ int up_lowgetc(void)
  *  sent character
  *
  ****************************************************************************/
-int up_putc(int ch)
+
+#define MAX_BUFFER_SIZE 1000
+extern bool abort_mode;
+static char buffer[MAX_BUFFER_SIZE];
+static int start = 0;
+static int end = 0;
+sem_t mutex;
+static bool init_mutex = true;
+
+int putc_helper(int ch)
 {
 	/* Check for LF */
 
@@ -797,6 +806,54 @@ int up_putc(int ch)
 	}
 
 	up_lowputc(ch);
+
+}
+
+int up_putc(int ch)
+{
+	if (abort_mode) {
+		/* board has crashed. so disable the uart interrupts
+		 * and procedd with up_lowputc directly
+		 */
+		struct uart_dev_s *dev = &CONSOLE_DEV;
+		uart_disabletxint(dev);
+		putc_helper(ch);
+	}
+	else if (getpid() == 0 || up_interrupt_context()) {
+		/* Direclty send it to the lower level */
+		putc_helper(ch);
+	}
+	else {
+		/* we are in application context */
+		buffer[end] = ch;
+		end = end + 1;
+
+		if (ch == '\n' || end == MAX_BUFFER_SIZE) {
+			/* we got a complete line to output or the buffer is full */
+			if (init_mutex) {
+				sem_init(&mutex, 1, 1);
+				init_mutex = false;
+			}
+
+			sem_wait(&mutex);
+			struct uart_dev_s *dev = &CONSOLE_DEV;
+			uart_disabletxint(dev);
+
+			/* print this entire line in a new line to avoid the line
+			 * starting in the middle of a line that is being output
+			 */
+			putc_helper('\n');
+
+			for (; start != end; start++) putc_helper(buffer[start]);
+
+			start = 0;
+			end = 0;
+
+			uart_enabletxint(dev);
+			sem_post(&mutex);
+
+		}
+	}
 	return ch;
 }
 


os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -819,6 +806,54 @@ int up_putc(int ch)
...
 
 	up_lowputc(ch);
+
+}
+
+int up_putc(int ch)
+{
+	if (abort_mode) {
+		/* board has crashed. so disable the uart interrupts
+		 * and procedd with up_lowputc directly
```
dineshdas-samsung, line 816: Typo "procedd" . Please correct it.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -819,6 +806,54 @@ int up_putc(int ch)
...
+		 */
+		struct uart_dev_s *dev = &CONSOLE_DEV;
+		uart_disabletxint(dev);
+		putc_helper(ch);
+	}
+	else if (getpid() == 0 || up_interrupt_context()) {
+		/* Direclty send it to the lower level */
```
amogh-samsung, line 823: Please correct "direclty" -> "directly"

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,7 +786,16 @@ int up_lowgetc(void)
...
  *
  ****************************************************************************/
-int up_putc(int ch)
+
+#define MAX_BUFFER_SIZE 1000
+extern bool abort_mode;
+static char buffer[MAX_BUFFER_SIZE];
+static int start = 0;
+static int end = 0;
+sem_t mutex;
```
jeongchanKim, line 795: Please add `static` if this variable is used only in this file.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -819,6 +806,54 @@ int up_putc(int ch)
...
+		uart_disabletxint(dev);
+		putc_helper(ch);
+	}
+	else if (getpid() == 0 || up_interrupt_context()) {
+		/* Direclty send it to the lower level */
+		putc_helper(ch);
+	}
+	else {
+		/* we are in application context */
+		buffer[end] = ch;
```
Taejun-Kwon, line 828: simply, buffer[end++] = ch;

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,7 +786,16 @@ int up_lowgetc(void)
  *  sent character
  *
  ****************************************************************************/
-int up_putc(int ch)
+
+#define MAX_BUFFER_SIZE 1000
```
Taejun-Kwon, line 790: Don't have to consider size of uart buffer?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,7 +786,16 @@ int up_lowgetc(void)
...
+extern bool abort_mode;
```
Taejun-Kwon, line 791: I think this is not good option, @sunghan-chang Is there any api or something to realize abort situation? or should we add this to dev structure? let us here your opinion.

os/arch/arm/src/amebad/amebad_serial.c
abhishek-samsung: This buffer is for holding the logs (lldbg) until we get to the end of the line and then the logs are output directly through up_lowputc. So I think we need not consider the size of uart buffer as this is not related to it.

os/arch/arm/src/amebad/amebad_serial.c
sunghan-chang: I agree @Taejun-Kwon 's opinion. But before this, I don't know why up_putc should be different with abort mode. Please let me know it first.

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -808,7 +786,16 @@ int up_lowgetc(void)
...
+static char buffer[MAX_BUFFER_SIZE];
+static int start = 0;
+static int end = 0;
+sem_t mutex;
+static bool init_mutex = true;
+
+int putc_helper(int ch)
```
sunghan-chang, line 798: What's the meaning of `helper` in the name?

os/arch/arm/src/amebad/amebad_serial.c
```diff
@@ -819,6 +806,54 @@ int up_putc(int ch)
 	}
 
 	up_lowputc(ch);
+
+}
+
+int up_putc(int ch)
```
sunghan-chang, line 812: This is the function of the chipset vendor. Is there no way to do this functionality on common code?

os/arch/arm/src/amebad/amebad_serial.c
abhishek-samsung: When assert fails during application context, both printf and lldbg will output at the same time. So by disabling uart interrupts during this time, we can prevent mixed logs. As abort mode is set to true after the failure of debug assert, I used it.

os/arch/arm/src/amebad/amebad_serial.c
abhishek-samsung: Initially, I named the function up_putc_helper, and then since up shouldn't be used in the name of this function, I changed it to putc_helper.

os/arch/arm/src/amebad/amebad_serial.c
abhishek-samsung: These changes can be moved to serial.c. The reason for making the changes here is since the problem is with only rtk boards. If this method is fine, I will move these changes to serial.c.

sunghan-chang: The serial.c does not work with printf and dbg. How would you move there?

os/arch/arm/src/amebad/amebad_serial.c
sunghan-chang: Do you mean that printf for threads is working in assert mode? If we add controlling the interrupt always, what's the problem? The stm has it in up_putc.

abhishek-samsung: True. both stm and imxrt have disabled interrupts in up_putc while outputting each character. During assert failure, we won't wait for \n, instead, we print the logs directly (This is what I implemented) so that in the future even if we miss \n in assert failure logs the logs won't be stuck in the buffer. I also have tried disabling uart interrupts while outputting each character but it didn't resolve the issue (still mixed logs occurred). So in order to prevent mixed logs we need to disable the uart interrupts till the whole line is output.

os/arch/arm/src/amebad/amebad_serial.c
abhishek-samsung: A new function can be declared like up_lineputc in arch.h and defined in serial.c, which is similar to the up_putc function in this PR. We can link this to lib_lowoutstream_putc (currently up_putc is linked to it).  This new function up_lineputc will use a buffer and will finally send the characters to up_putc to output once we get a whole line. Please let me know if this is fine.

sunghan-chang: Sorry for late reply.
The serial.c has the functions of uart driver. There is no other function. Could you check where other lib_xxxstream function?
Could you write simple ppt to show how it flows?

os/arch/arm/src/amebad/amebad_serial.c
deepaksrma: please correct the typo.

deepaksrma, line 823: please correct the typo.

PR #5164, opened 2021-09-07
os/fs/smartfs: Fix truncate discrepancy during file opening in SmartFs

amogh-samsung:
- Do not truncate an already existing file to 0 if O_CREAT flag is included for opening.
- Truncate the file if O_TRUNC flag is included and O_APPEND is not.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -245,20 +245,20 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 			goto errout_with_buffer;
 		}
 
-		/* TODO: Test open mode based on the file mode */
+		/* The file exists already.
+		 * If the file has been requested to be opened in TRUNCATE mode, then shrink the file.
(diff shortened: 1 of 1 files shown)


os/fs/smartfs/smartfs_smart.c
```diff
@@ -245,20 +245,20 @@ static int smartfs_open(FAR struct file *filep, const char *relpath, int oflags,
 			goto errout_with_buffer;
 		}
 
-		/* TODO: Test open mode based on the file mode */
-
-		/* The file exists.  Check if we are opening it for O_CREAT or
-		 * O_TRUNC mode and delete the sector chain if we are. */
+		/* The file exists already.
+		 * If the file has been requested to be opened in TRUNCATE mode, then shrink the file.
```
dineshdas-samsung, line 249: What about "If the file is opened in truncate mode, then shrink the file." ?

amogh-samsung: Since both sentences mean the same and are understandable enough, I'll keep it same for now, thanks.

Taejun-Kwon (approved):

PR #5168, opened 2021-09-08
drivers/lwnl: separate queue sem

an4967:
- separate ble & wifi queue sem
- add data pointer mode in lwnl_postmsg
(if data_len is negative, a queue copies pointer of data.)

### DIFF
--- a/os/drivers/lwnl/lwnl_evt_queue.c
+++ b/os/drivers/lwnl/lwnl_evt_queue.c
@@ -28,13 +28,13 @@
 #include "lwnl_evt_queue.h"
 #include "lwnl_log.h"
 
-#define LWQ_LOCK								\
+#define LWQ_LOCK(type)							\
 	do {										\
-		sem_wait(&g_wm_sem);					\
+		sem_wait(&g_queue_sem[type]);			\
 	} while (0)
-#define LWQ_UNLOCK								\
+#define LWQ_UNLOCK(type)						\
 	do {										\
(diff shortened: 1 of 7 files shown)


jeongchanKim (approved):
LGTM

os/drivers/lwnl/lwnl_evt_queue.c
```diff
@@ -253,14 +255,14 @@ int lwnl_get_event(struct file *filep, char *buf, int len)
...
-	LWQ_UNLOCK;
+	LWQ_UNLOCK(fp->type);
 	return written;
 }
 
 int lwnl_add_listener(struct file *filep, lwnl_dev_type type)
 {
 	LWNL_LOGI(TAG, "--> fp (%p) type (%d)", filep, type);
-	LWQ_LOCK;
+	LWQ_LOCK(type);
```
jeongarmy, line 265: I think we need to validate `type` value to avoid buffer overflow because this function is called via lwnl_ioctl by user.
So it can be out of range of g_queue_sem, LWNL_DEV_TYPE_MAX I think.
How about adding checking `type` value here or lwnl_ioctl?

pillip8282 (commented):
g_wm_sem is for protecting g_filep_list which can't be separated by dev type. So you should separate g_filep_list first.

an4967:
@pillip8282 
Okay, I will check g_filep_list architecture more. Thanks

PR #5171, opened 2021-09-09
os/Makefile: Enable binary compression for common binary

kishore-sn:
Common binary will be compressed with CONFIG_COMPRESSION_TYPE and
CONFIG_COMPRESSION_BLOCK_SIZE, if CONFIG_COMPRESSED_BINARY is enabled.

### DIFF
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -464,6 +464,10 @@ ifeq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
 		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)"; \
 		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)/base-files"; \
 	fi
+ifeq ($(CONFIG_COMPRESSED_BINARY),y)
+	$(Q) tools/compression/mkcompressimg $(CONFIG_COMPRESSION_BLOCK_SIZE) $(CONFIG_COMPRESSION_TYPE) $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME) $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME).comp
+	$(Q) mv $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME).comp $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME)
+endif
 	$(Q) cp $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME) $(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)/base-files/
 	$(Q) $(MAKE) smartfs
 endif
--- a/os/binfmt/elf.c
+++ b/os/binfmt/elf.c
@@ -235,12 +235,6 @@ static int elf_loadbinary(FAR struct binary_s *binp)
 	/* Initialize the ELF library to load the program binary. */
 	loadinfo.offset = binp->offset;
 	loadinfo.filelen = binp->filelen;
-	loadinfo.compression_type = binp->compression_type;
-#ifdef CONFIG_COMPRESSED_BINARY
-	if (binp->islibrary) {
-		loadinfo.compression_type = CONFIG_COMPRESSION_TYPE;
-	}
-#endif
 #ifdef CONFIG_APP_BINARY_SEPARATION
 	loadinfo.binp = binp;
 #endif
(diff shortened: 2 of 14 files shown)


os/binfmt/elf.c
```diff
@@ -236,6 +236,11 @@ static int elf_loadbinary(FAR struct binary_s *binp)
 	loadinfo.offset = binp->offset;
 	loadinfo.filelen = binp->filelen;
 	loadinfo.compression_type = binp->compression_type;
```
jeongchanKim, line 238: Can we merge those lines?
```
loadinfo.compression_type = binp->compression_type;
#ifdef CONFIG_COMPRESSED_BINARY
	if (binp->islibrary) {
		loadinfo.compression_type = CONFIG_COMPRESSION_TYPE;
	}
#endif
```
Because we can always set the compression type with `CONFIG_COMPRESSION_TYPE`.
```
#ifdef CONFIG_COMPRESSED_BINARY
    loadinfo.compression_type = CONFIG_COMPRESSION_TYPE;
#else
    loadinfo.compression_type = binp->compression_type;
```

os/Makefile.unix
```diff
@@ -464,6 +464,10 @@ ifeq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
 		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)"; \
 		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)/base-files"; \
 	fi
+ifeq ($(CONFIG_COMPRESSED_BINARY),y)
+	$(Q) tools/compression/mkcompressimg $(CONFIG_COMPRESSION_BLOCK_SIZE) $(CONFIG_COMPRESSION_TYPE) $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME) $(OUTBIN_DIR)/$(CONFIG_COMMON_BINARY_NAME).comp
```
jeongchanKim, line 468: It makes compressed binary with `.comp` and change the name to the origin.
Is it possible to make compressed binary with `CONFIG_COMMON_BINARY_NAME` directly?

os/binfmt/elf.c
kishore-sn: True.. Actually we dont need to set the compression type at all. We will always use CONFIG_COMPRESSION_TYPE if compression is enabled. So shall I remove the whole code related to setting compression type in code and binary header?

os/Makefile.unix
kishore-sn: No.. Its not possible. The compression utility requires output file to be different from input file.

jeongchanKim: Okay. Fine. Thanks.

os/binfmt/elf.c
jeongchanKim: I agree to remove redundant codes about setting compresstion type.
But even in binary header? Why?
I think we should check the header that binary is compressed or not. Isn't it?

kishore-sn: If `CONFIG_COMPRESSED_BINARY` is enabled, then we will always compress the binary using `CONFIG_COMPRESSION_TYPE`. It cannot be anything different. Then, why do we need to add this information in header?

jeongchanKim: Umm.. When kernel build, enable the `CONFIG_COMPRESSED_BINARY`.
Then I think binary manager can support both compressed and non-compressed binary simultaneously.
If we can distinguish both type, then it is much better to manage the binaries against memory constraint.
How do you think?

kishore-sn: In that case, it is ok. But currently in the code, if `CONFIG_COMPRESSED_BINARY` is enabled, then we always create compressed binary. Also, if this is the case, then it is possible that common binary also can be either compressed or uncompressed. But, there is no way to distinguish it.

jeongchanKim: As discussed through messenger, let's support same compression way for all binaries.
Please remove unnecessary codes which related to it.
Thanks :)

kishore-sn: Done

os/binfmt/libelf/libelf_cache.c
```diff
@@ -454,18 +436,17 @@ int elf_cache_read(int filfd, uint16_t binary_header_size, FAR uint8_t *buffer,
  *   OK (0) on Success
  *   Negative value on Failure
  ****************************************************************************/
-int elf_cache_init(int filfd, uint16_t offset, off_t filelen, uint8_t compression_type)
+int elf_cache_init(int filfd, uint16_t offset, off_t filelen)
 {
 	int ret = OK;
 
-	binfo("filfd: %d offset: %d filelen: %d compression_type: %d\n", filfd, offset, filelen, compression_type);
+	binfo("filfd: %d offset: %d filelen: %d\n", filfd, offset, filelen);
```
jeongchanKim, line 443: Please change `offset: %d` to `offset: %u`, because `offset` is unsigned type.

sunghan-chang: @kishore-sn Could you check Jeongchan's comment?

kishore-sn: Done

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #5176, opened 2021-09-15
wifi_manager: modify state changing in scanning.

pillip8282:
wifi_manager manages it's internal information in
wifi_manager_info.c. However it doesn't update its state after
scanning is done. Because wifimgr_set_info wasn't called.

### DIFF
--- a/framework/src/wifi_manager/wifi_manager_state.c
+++ b/framework/src/wifi_manager/wifi_manager_state.c
@@ -136,10 +136,13 @@ static char *wifimgr_state_str[] = {
 #define WIFIMGR_GET_STATE g_manager_info.state
 #define WIFIMGR_GET_PREVSTATE g_manager_info.prev_state
 #define WIFIMGR_STORE_PREV_STATE (g_manager_info.prev_state = g_manager_info.state)
-#define WIFIMGR_RESTORE_STATE								\
-	do {													\
-		g_manager_info.state = g_manager_info.prev_state;	\
-		g_manager_info.prev_state = WIFIMGR_NONE;			\
+#define WIFIMGR_RESTORE_STATE                             \
+	do {                                                  \
+		g_manager_info.state = g_manager_info.prev_state; \
+		g_manager_info.prev_state = WIFIMGR_NONE;         \
+		wifimgr_info_msg_s twmsg;                         \
+		twmsg.state = g_manager_info.state;               \
+		wifimgr_set_info(WIFIMGR_STATE, &twmsg);         \
 	} while (0)
 
 #define TAG "[WM]"


framework/src/wifi_manager/wifi_manager_state.c
```diff
@@ -136,10 +136,13 @@ static char *wifimgr_state_str[] = {
...
-	do {													\
-		g_manager_info.state = g_manager_info.prev_state;	\
-		g_manager_info.prev_state = WIFIMGR_NONE;			\
+#define WIFIMGR_RESTORE_STATE                             \
+	do {                                                  \
+		g_manager_info.state = g_manager_info.prev_state; \
+		g_manager_info.prev_state = WIFIMGR_NONE;         \
+		wifimgr_info_msg_s twmsg;                         \
+		twmsg.state = g_manager_info.state;               \
+		wifimgr_set_info(WIFIMGR_STATE, &twmsg);         \
```
jeongchanKim, line 145: In this case,
`g_manager_info.state` and `g_manager_info.prev_state` are set as same, the original `g_manager_info.prev_state`.

Let's assume, the original `g_manager_info.prev_state` is `A`.
```
line 141 : g_manager_info.state = A
line 142 : g_manager_info.prev_state = WIFIMGR_NONE
line 144 : twmsg.state  = A (Because of line 141)
line 145 : In wifimgr_set_info(), 
                 g_manager_info.prev_state = g_manager_info.state (== A)
                 g_manager_info.state = info->state(passed param) (== A)
```
Then, `g_manager_info.state` and `g_manager_info.prev_state` are all `A`, which is the original `g_manager_info.prev_state` .

So I think, this macro is as same as just setting `g_manager_info.state = g_manager_info.prev_state`.
How do you think?

pillip8282: g_manger_info in wifi_manager_state.c and g_manager_info in wifi_manager_info.c are different. So prev_state is not overwritten.

sangwon03 (approved):
LGTM

jeongchanKim (approved):
LGTM

framework/src/wifi_manager/wifi_manager_state.c
davidfather: @pillip8282 I understand two static variables have different purposes: one for state and the other for information.
However, it should've been different for better understanding.

davidfather (approved):
Looks good to me.

framework/src/wifi_manager/wifi_manager_state.c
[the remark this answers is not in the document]
pillip8282: I'll modify it to different PR.

PR #5178, opened 2021-09-16
framework/wifi_manager: Fix wrong event when mobile connect to softap

j11591:
If CONFIG_WIFIMGR_DISABLE_DHCPS is not set, g_msg is not set with LWNL_SOFTAP_STA_JOINED.

Even though g_msg is not set and value of g_msg is previous one, wifimgr_message_out call wifimgr_handle_request.

It makes wrong event like EVT_STA_DISCONNECTED or EVT_LEFT when mobile connect to SoftAP.

By using EVT_JOINED regardless of CONFIG_WIFIMGR_DISABLE_DHCPS, it can be fixed.


jeongchanKim (commented):
Could you leave the description into the commit msg?
Now, you left only in PR description.
But in this case, we cannot see the description with commit.

jeongchanKim (approved):
LGTM

thapav (approved):
LGTM

framework/src/wifi_manager/wifi_manager_lwnl_listener.c
```diff
@@ -112,9 +112,7 @@ static int _lwnl_call_event(int fd, lwnl_cb_status status, int len)
 		LWNL_SET_MSG(&g_msg, WIFIMGR_EVT_STA_DISCONNECTED, WIFI_MANAGER_FAIL, NULL, NULL);
 		break;
 	case LWNL_EVT_SOFTAP_STA_JOINED:
-#ifdef CONFIG_WIFIMGR_DISABLE_DHCPS
```
davidfather, line 115: @j11591 Could you elaborate why this modification is necessary?

j11591: In lwnl_fetch_event, (void)_lwnl_call_event(fd, status, len); is called.
g_msg is not set by not calling LWNL_SET_MSG, previous msg can be use for lwnl_fetch_event.

PR #5180, opened 2021-09-17
framework/ble_manager: add ble_addr param in scan filter struct

an4967:
- add ble_addr param in scan filter struct


pillip8282 (approved):
lgtm

PR #5183, opened 2021-09-20
os/board/rtl8721csm: Added Scan Timeout as BLE connect parameter

ziyik:
Take in Scan Timeout as parameter and use it during BLE connect


jeongchanKim (approved):
LGTM

an4967 (approved):
LGTM

PR #5186, opened 2021-09-23
os/board/rtl8721csm: Tune BT Coex case

ziyik:
- Tune BT Coex Case to unsure BT no delay with Slave Latency enable


an4967 (approved):
LGTM

PR #5187, opened 2021-09-23
os/Makefile.unix: use -f option to delete output bin folder 

sunghan-chang:
When loadable build configuration is built, the user folder under output/bin is newly created and
it should be deleted with clean command. But DELFILE definition can do only files, can't do folders.
It causes the error as below:

rm: cannot remove '../build/output/bin/user': Is a directory
Makefile.unix:629: recipe for target 'clean' failed
make: *** [clean] Error 1

This commit changes the command to delete output bin folder from DELFILE to DELDIR to resolve above.


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #5194, opened 2021-09-27
os/kernel/mqueue: increase NUM_INTERRUPT_MSGS

an4967:
- Because BLE is enabled, the number of interrupt is increased.


jeongchanKim (commented):
As we discussed, it is enough to change the normal mq.
Please close this PR.
Thanks.

an4967:
This is not needed

PR #5195, opened 2021-09-27
os/board/rtl8721csm: disable RX_SHORTCUT to avoid tcp crash

gexuyan:
disable RX_SHORTCUT to force large input packets to go through segmentation process.


os/board/rtl8721csm/src/component/common/drivers/wlan/realtek/include/autoconf.h
```diff
@@ -507,7 +507,7 @@ extern unsigned int g_ap_sta_num;
 		#undef RX_AGGREGATION
 		#define RX_AGGREGATION 1
 		#undef NOT_SUPPORT_40M
-		#define RX_SHORTCUT 1
+		//#define RX_SHORTCUT 1
```
jeongchanKim, line 510: How about removing this line if this is not used anymore?

gexuyan: ok can

davidfather: @gexuyan How large packet do you want to segment? Why does this forced segmentation help tcp to avoid crash?

gexuyan: hi @davidfather , the tcp crash is due to ip layer receive 60k bytes size packets and the bulky memory allocated for large packets will touch secure memory region, which will raise security hardfault and leads to crash. Originally, received packets will go through shortcut path which doesn't include segmentation procedure, after closing RX_SHORTCUT, received large packets will be segmented then don't require a continuous bulky memory.

sunghan-chang:
Could you merge your two commits?
You can use git rebase or git reset --soft.

And a minor question.
Do you use that definition like `#ifdef`? Then how about keeping the same policy of the definitions?
```
#undef RX_SHORTCUT
```

gexuyan:
Dear all, we found this modification may leads to a wlan performance degradation, so I will close this PR first. Thank you.

PR #5196, opened 2021-09-28
fs/smartfs : Erase old journal block forcely.

Taejun-Kwon:
When checkin failed due to powercut, and checkin was added to handle erase old journal block,
old journal block survided as a zombie.
There was no way to erase it, so block map has added to erase fully used journal block


os/fs/driver/mtd/smart.c
```diff
@@ -5581,6 +5598,15 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
 	} else {
 		dev->journal_seq = 0;
 	}
+	
+	for (int i = 0; i < dev->njournaleraseblocks; i++) {
+		/* Definately, below block should be erased */
```
jeongchanKim, line 5603: Please change to `definitely`.

os/fs/driver/mtd/smart.c
```diff
@@ -5581,6 +5598,15 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
...
+		fvdbg("current map : %d numberof journal : %d\n", i, dev->block_map[i]);
```
jeongchanKim, line 5604: How about printing the `block_map[i]` with unsigned format?
And I think adding space between `number` and `of` is better to read.

os/fs/driver/mtd/smart.c
```diff
@@ -5531,12 +5539,20 @@ static int smart_journal_scan(FAR struct smart_struct_s *dev, bool print_dump)
...
+	if (dev->block_map == NULL) {
+		fdbg("out of memory!!\n");
+		return -ENOMEM;
+	}
+	
 	for (dev->journal_seq = 0; dev->journal_seq < dev->njournalentries; dev->journal_seq++) {
 		ret = smart_journal_read_journal_log(dev, &log);
 		if (ret < 0) {
 			fdbg("Reading Journal Entry Failed! ret : %d journal_seq : %d\n", ret, dev->journal_seq);
 			return -EIO;
```
jeongchanKim, line 5553: Is it okay not to free the `dev->block_map` for failure case?

os/fs/driver/mtd/smart.c
amogh-samsung: Could you please change the line to: -
fvdbg("Current _Journal block_ : %d, number of _non-empty_ journals : %d\n", i, dev->block_map[i]);

amogh-samsung (commented):
Logic looks good

amogh-samsung:
Request you to please change the commit title and description
Since I believe this logic does a bigger job than just erasing old blocks.

"fs/smartfs: Add logic to check for possible un-erased old journal blocks and clear them

- A powercut during the shifting of active journal index from current to next block might leave the old block unerased.
- The system recovers with the active journal entry, but might not erase the old block which will cause corruption when re-allocated
- Check all Journal Eraseblocks during startup scan and erase those which are full and not currently in use."

os/fs/driver/mtd/smart.c
Taejun-Kwon: yes it will free when smartfs unmounted

Taejun-Kwon:
> Request you to please change the commit title and description Since I believe this logic does a bigger job than just erasing old blocks.
> 
> "fs/smartfs: Add logic to check for possible un-erased old journal blocks and clear them
> 
> * A powercut during the shifting of active journal index from current to next block might leave the old block unerased.
> * The system recovers with the active journal entry, but might not erase the old block which will cause corruption when re-allocated
> * Check all Journal Eraseblocks during startup scan and erase those which are full and not currently in use."

I will modify it based on your comment, but there should be only one missed block.

amogh-samsung:
> > Request you to please change the commit title and description Since I believe this logic does a bigger job than just erasing old blocks.
> > "fs/smartfs: Add logic to check for possible un-erased old journal blocks and clear them
> > 
> > * A powercut during the shifting of active journal index from current to next block might leave the old block unerased.
> > * The system recovers with the active journal entry, but might not erase the old block which will cause corruption when re-allocated
> > * Check all Journal Eraseblocks during startup scan and erase those which are full and not currently in use."
> 
> I will modify it based on your comment, but there should be only one missed block.

Agreed and understood, according to our code design, we only shift from current block to next
So only current block can be missed, not multiple.

Taejun-Kwon:
Will be tested until end of next weekend.

os/fs/driver/mtd/smart.c
```diff
@@ -5994,6 +6033,11 @@ int smart_initialize(int minor, FAR struct mtd_dev_s *mtd, FAR const char *partn
 		smart_free(dev, rootdirdev);
 	}
 #endif
+#ifdef CONFIG_MTD_SMART_JOURNALING
+	if (dev->block_map != NULL) {
+		kmm_free(dev->block_map);
```
jeongchanKim, line 6038: How about adding `dev->block_map = NULL;`?
In line 1186, it is added.

PR #5199, opened 2021-09-28
framework/ble_manager: improve memcpy speed

an4967:
- Test 3 cases of memcpy
1. use 'memcpy'
2. copy 1 byte unit
3. copy 4 bytes unit

The size of source  : 84 bytes (trble_scanned_device)
The number of trial : 10000 times

- The results
Test#1 :  8.01 ms
Test#2 : 21.15 ms
Test#3 :  6.36 ms

The speed of memcpy increased 15% by using 4 bytes unit.

### DIFF
--- a/framework/include/ble_manager/ble_client.h
+++ b/framework/include/ble_manager/ble_client.h
@@ -67,7 +67,7 @@ typedef struct {
 	uint8_t raw_data_length;
 	uint8_t resp_data[BLE_ADV_RESP_DATA_MAX_LEN];
 	uint8_t resp_data_length;
-} ble_scanned_device;
+} __attribute__((aligned(4), packed)) ble_scanned_device;
 
 typedef struct {
 	ble_conn_info conn_info;
--- a/os/include/tinyara/net/if/ble.h
+++ b/os/include/tinyara/net/if/ble.h
@@ -151,7 +151,7 @@ typedef struct {
 	uint8_t raw_data_length;
 	uint8_t resp_data[TRBLE_ADV_RESP_DATA_MAX_LEN];
 	uint8_t resp_data_length;
-} trble_scanned_device;
+} __attribute__((aligned(4), packed)) trble_scanned_device;
 
 typedef struct {
 	int size;
--- a/os/net/blemgr/bledev.c
+++ b/os/net/blemgr/bledev.c
@@ -57,7 +57,13 @@ int trble_scan_data_enque(trble_scanned_device *info)
 		*/
 		return -2;
 	}
-	memcpy(&g_scan_queue->queue[g_scan_queue->write_index], info, sizeof(trble_scanned_device));
+
+	int i;
+	uint32_t *u1 = (uint32_t *)info;
+	uint32_t *u2 = (uint32_t *)&g_scan_queue->queue[g_scan_queue->write_index];
+	for (i = 0; i < sizeof(trble_scanned_device) / sizeof(uint32_t); i++) {
+		*(u2 + i) = *(u1 + i);
+	}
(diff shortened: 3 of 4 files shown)


framework/include/ble_manager/ble_client.h
```diff
@@ -59,6 +59,7 @@ typedef struct {
 	bool is_secured_connect;
 } ble_conn_info;
 
+#pragma pack(push, 4)
```
jeongchanKim, line 62: As I know, for structure alignment in gcc, we should use `__attribute__ ((__packed__))`.
Please check this.

os/include/tinyara/net/if/ble.h
```diff
@@ -143,6 +143,7 @@ typedef struct {
 	bool is_secured_connect;
 } trble_conn_info;
 
+#pragma pack(push, 4)
```
jeongchanKim, line 146: Same as above.

framework/include/ble_manager/ble_client.h
jeongchanKim: And there is a trade-off for structure alignment.
If aligned, we can save the memory, but it takes much time as I know.
The purpose of this PR is for enhancing the speed, then we should check the trade-off.
Please check this.

an4967: @JeonginKim 
https://developer.arm.com/documentation/dui0491/i/Compiler-specific-Features/-pragma-pack-n-

With ARM GCC compiler, we can use '#pragma'.

an4967: When a structure is packed with 1 bytes, you are right.
As I know, the default value of packed size is 4 bytes. So there is no change with previous state.
I just added the packed size for ensuring the struct is multiple of 4

os/include/tinyara/net/if/ble.h
an4967: I replied above

framework/include/ble_manager/ble_client.h
jeongchanKim: Okay, understood. Thanks.

jeongchanKim (approved):
LGTM

framework/include/ble_manager/ble_client.h
```diff
@@ -67,7 +67,7 @@ typedef struct {
 	uint8_t raw_data_length;
 	uint8_t resp_data[BLE_ADV_RESP_DATA_MAX_LEN];
 	uint8_t resp_data_length;
-} ble_scanned_device;
+} __attribute__((aligned(4), pakced)) ble_scanned_device;
```
sangam-swami, line 70: typo *packed?

os/include/tinyara/net/if/ble.h
```diff
@@ -151,7 +151,7 @@ typedef struct {
 	uint8_t raw_data_length;
 	uint8_t resp_data[TRBLE_ADV_RESP_DATA_MAX_LEN];
 	uint8_t resp_data_length;
-} trble_scanned_device;
+} __attribute__((aligned(4), pakced)) trble_scanned_device;
```
sangam-swami, line 154: typo *packed?

an4967: Thanks, I fixed it

framework/include/ble_manager/ble_client.h
an4967: Thanks, I fixed it

jeongchanKim (approved):
LGTM

pillip8282 (approved):
LGTM

PR #5201, opened 2021-09-30
fs/smartfs : Check CRC of duplicate sectors

sungkyu0:
In the illegal case by power-loss,
when the multiple logical sectors are duplicated,
we compare the sequence number of each sector and select the newer sector.
Just in case, add CRC check for the newer sector.
If the newer sector has CRC error, then we use the older sector.


os/fs/driver/mtd/smart.c
```diff
@@ -2163,6 +2163,44 @@ static int smart_scan(FAR struct smart_struct_s *dev)
 #endif
 			}
 
+			fdbg("Duplicate Sector winner=%d, loser=%d\n", winner, loser);
```
amogh-samsung, line 2166: We've used the terms winner and loser in our code internally,
However, I'd request you to use other more proper terms for the debug statements please.
Somebody else might not be able to figure out what competition is happening here.

How about "active sector" and "inactive sector? instead of winner and loser respectively?

os/fs/driver/mtd/smart.c
```diff
@@ -2163,6 +2163,44 @@ static int smart_scan(FAR struct smart_struct_s *dev)
...
+
+#ifdef CONFIG_MTD_SMART_ENABLE_CRC
+			/* Check CRC of the winner sector just in case */
+
+			ret = MTD_BREAD(dev->mtd, winner * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)dev->rwbuffer);
```
amogh-samsung, line 2171: I think the header of the sector has already been read at line 2126, so maybe we can read just the remaining part of the sector here?

os/fs/driver/mtd/smart.c
```diff
@@ -2163,6 +2163,44 @@ static int smart_scan(FAR struct smart_struct_s *dev)
...
+			if (ret == dev->mtdBlksPerSector) {
```
amogh-samsung, line 2172: What if "ret != OK", i.e. the read request fails?
We should return an appropriate ERROR code here

Otherwise, continue with the functioning.

os/fs/driver/mtd/smart.c
```diff
@@ -2163,6 +2163,44 @@ static int smart_scan(FAR struct smart_struct_s *dev)
...
+				} else {
+					/* winner: origin(CRC error) -> sector
+					 * loser : sector            -> origin(CRC error)
+					 */
+
+					loser = winner;
+					winner = sector;
+				}
+
+				fdbg("Duplicate Sector winner=%d, loser=%d\n", winner, loser);
```
amogh-samsung, line 2200: Requesting you to modify this debug statement too.

amogh-samsung:
Requesting you to modify the commit title slightly as well: -
"fs/smartfs: Check CRC of duplicate sectors before confirming the active sector"

Something like this to indicate that we are checking CRC to confirm that the sector we have chosen as the winner has valid data.
I'd like to know your version of this please.

os/fs/driver/mtd/smart.c
Taejun-Kwon: No, winner can be changed based on sequence number, so here we should read winner's data again.

os/fs/driver/mtd/smart.c
Taejun-Kwon: I agree with @amogh-samsung, if ret is not same as mtdBlksPerSector, error happend in flash driver, and it will return errno.
```
if (ret != dev->mtdBlksPerSector) {
    return ret;
}

ret = smart_validate_crc(dev);
if (ret != OK) {
   ....
}
```

os/fs/driver/mtd/smart.c
sungkyu0: I changed

os/fs/driver/mtd/smart.c
sungkyu0: I changed

os/fs/driver/mtd/smart.c
sungkyu0: I changed

amogh-samsung:
> > Requesting you to modify the commit title slightly as well: - "fs/smartfs: Check CRC of duplicate sectors before confirming the active sector"
> > Something like this to indicate that we are checking CRC to confirm that the sector we have chosen as the winner has valid data. I'd like to know your version of this please.
> 
> I changed commit title

Hi, only the title of the PR has changed, not the commit title.
The PR title will not be visible in the Git Log,
You will need to update the commit title and then push please.

os/fs/driver/mtd/smart.c
```diff
@@ -2163,18 +2163,18 @@ static int smart_scan(FAR struct smart_struct_s *dev)
 #endif
 			}
 
-			fdbg("Duplicate Sector winner=%d, loser=%d\n", winner, loser);
+			fdbg("Duplicate Sector active_sector=%d, inactive_sector=%d\n", winner, loser);
```
amogh-samsung, line 2166: You can implement these changes in the first commit itself and re push please.
We don't need to add a second commit here.

amogh-samsung (changes requested):
Please do the modifications in the first commit itself, the second commit is not needed.
Please also modify the title of the commit itself instead of the PR

dineshdas-samsung:
Can we Please reduce the length of commit title (brief and concise)?

amogh-samsung (approved):
Looks Good

sungkyu0:
> Can we Please reduce the length of commit title (brief and concise)?

@dineshdas-samsung I changed commit title.
Thank you

os/fs/driver/mtd/smart.c
```diff
@@ -2163,6 +2163,44 @@ static int smart_scan(FAR struct smart_struct_s *dev)
...
 			}
 
+			fdbg("Duplicate Sector active_sector=%d, inactive_sector=%d\n", winner, loser);
+
+#ifdef CONFIG_MTD_SMART_ENABLE_CRC
+			/* Check CRC of the winner sector just in case */
+
+			ret = MTD_BREAD(dev->mtd, winner * dev->mtdBlksPerSector, dev->mtdBlksPerSector, (FAR uint8_t *)dev->rwbuffer);
+			if (ret != dev->mtdBlksPerSector) {
+				/* Validate the CRC of the read-back data */
```
kishore-sn, line 2173: This comment needs to be moved down i think.

PR #5203, opened 2021-10-04
os/board/rtl8721csm: process qos null data to avoid TCP receive crash

gexuyan:
make rx shortcut path stop sending qos null data to upper layer to avoid skb len indicated as negative value -10, -10 will be mapped into 65526 and leads to tcp receive crash.


pillip8282 (changes requested):
Please add commit description for others and for checking the related issue.

pillip8282 (approved):
LGTM

PR #5204, opened 2021-10-05
framework/ble_manager: adjust ble scan api

an4967:
- remove unused member in scanned device struct.
- separate scan APIs from client APIs
- add scan whitelist operation APIs


framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+			/* Already Exist */
+			break;
+		}
+
+		if (_whitelist_is_full() == true) {
+			ret = TRBLE_OUT_OF_MEMORY;
+			break;
+		}
+
+		if (g_scan_ctx.state != BLE_SCAN_STOPPED) {
```
jeongchanKim, line 340: How about moving this condition check at the top of the case statement like line 322?
Then we can save the time, if the state is not `BLE_SCAN_STOPPED`.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+			ret = TRBLE_INVALID_ARGS;
+			break;
+		}
+
+		if (_whitelist_is_exist(addr) != true) {
+			/* Already Deleted */
+			break;
+		}
+
+		if (g_scan_ctx.state != BLE_SCAN_STOPPED) {
```
jeongchanKim, line 368: Same as above.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+		blemgr_msg_params *param = (blemgr_msg_params *)msg->param;
+		ble_scan_filter *filter = (ble_scan_filter *)param->param[0];
+		ble_scan_callback_list *callbacks = (ble_scan_callback_list *)param->param[1];
+
+		if (filter != NULL) {
+			memcpy(&(g_scan_ctx.filter), filter, sizeof(ble_scan_filter));
+		}
+
+		if (callbacks == NULL) {
+			ret = TRBLE_INVALID_ARGS;
```
jeongchanKim, line 289: In this case, switch-case will be finished.
Is it okay to copy the `filter` into the `g_scan_ctx.filter` even for this case?

framework/src/ble_manager/ble_manager_state.c
an4967: By checking whitelist first, an user can get a exact result(already registered, whitelist full) regardless of state

framework/src/ble_manager/ble_manager_state.c
an4967: I commented above

framework/src/ble_manager/ble_manager_state.c
an4967: I fixed it

jeongchanKim: How about moving into the line 278?
Because, for failure case, we don't need to set the local variables also.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+			g_scan_ctx.state = priv_state;
+		}
+	} break;
+
+	case BLE_EVT_CMD_STOP_SCAN: {
+		BLE_STATE_CHECK;
+
+		ret = TRBLE_SUCCESS;
+		if (g_scan_ctx.state != BLE_SCAN_STARTED) {
+			break;
```
jeongchanKim, line 308: In this case, `ret` will be `TRBLE_SUCCESS`.
But I think this is failure case.
Or is it intended to return SUCCESS if trying to stop the already stopped?
Please check this.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
 		ret = ble_drv_conn_is_any_active(is_active);
 	} break;
 
+	// Scanner
+	case BLE_EVT_CMD_START_SCAN: {
+		BLE_STATE_CHECK;
+
+		if (g_scan_ctx.state != BLE_SCAN_STOPPED) {
```
Taejun-Kwon, line 274: So you should consider 2 things.
```
1. It is related to state, there should be lock mechanism.
2. State of framework should aligned with driver. Many of application can use framework, but driver will be used by only framework. So you should add some mechanism to get state of driver.
   For example, If application terminated suddenly and restarted it will init framework again. Then framework's state will be changed as a initial state, but driver still running.
```

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+			ret = TRBLE_BUSY;
+			break;
+		}
+
+		ble_scan_state_e priv_state = g_scan_ctx.state;
+		blemgr_msg_params *param = (blemgr_msg_params *)msg->param;
```
Taejun-Kwon, line 280: I think null state of msg & msg->param should be checked at the start of this function.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -197,6 +267,137 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
+			ret = TRBLE_INVALID_ARGS;
+			break;
+		}
+		memcpy(&(g_scan_ctx.callback), callbacks, sizeof(ble_scan_callback_list));
+
+		if (filter != NULL) {
+			memcpy(&(g_scan_ctx.filter), filter, sizeof(ble_scan_filter));
+		}
+
+		g_scan_ctx.state = BLE_SCAN_CHANGING;
```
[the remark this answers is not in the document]
an4967: In 'ble_manager_state', it is guaranteed that commands deliver in sequence.
So I think a lock is unnecessary.

framework/src/ble_manager/ble_manager_state.c
an4967: 1. Same as above.  BLE APIs are guaranteed to process in sequence.
2. Yes, you are right. I will raise a new PR for synchronizing state between driver and framework

framework/src/ble_manager/ble_manager_state.c
an4967: OK. I will fix it

framework/src/ble_manager/ble_manager_autoconnect.c
```diff
@@ -59,7 +59,7 @@ static void ble_auto_device_scanned_cb(ble_scanned_device *scanned_device)
 	char buf[BLE_MQ_SIZE] = { 0, };
 
 	if (scanned_device->adv_type == BLE_ADV_TYPE_DIRECT) {
-		if (memcmp(g_ctx->info.addr.mac, scanned_device->conn_info.addr.mac, BLE_BD_ADDR_MAX_LEN) == 0) {
+		if (memcmp(g_ctx->info.addr.mac, scanned_device->addr.mac, BLE_BD_ADDR_MAX_LEN) == 0) {
```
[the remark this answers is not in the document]
an4967: I will fix it

framework/src/ble_manager/ble_manager_lwnl.c
```diff
@@ -157,6 +157,37 @@ trble_result_e ble_drv_stop_scan()
 	return res;
 }
 
+trble_result_e ble_drv_scan_whitelist_add(trble_addr *addr)
+{
+	trble_result_e res = TRBLE_SUCCESS;
```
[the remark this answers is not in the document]
an4967: Yes, the WiFi & BLE managers are using lwnl for communicating between Kernel & User
They cannot work without lwnl.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -48,6 +58,61 @@ static blemgr_state_e g_manager_state = BLEMGR_UNINITIALIZED;
 		}                                                 \
 	} while (0)
 
+static bool _whitelist_delete(trble_addr *addr)
+{
+	int i;
+	for (i = 0; i < SCAN_WHITELIST_SIZE; i++) {
+		if (g_scan_whitelist[i].flag == SCAN_WHITELIST_IN_USE) {
```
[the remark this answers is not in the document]
an4967: I will fix it

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -48,6 +58,61 @@ static blemgr_state_e g_manager_state = BLEMGR_UNINITIALIZED;
...
+		}
+	}
+	return false;
+}
+
+static bool _whitelist_add(trble_addr *addr)
+{
+	int i;
+	for (i = 0; i < SCAN_WHITELIST_SIZE; i++) {
+		if (g_scan_whitelist[i].flag == SCAN_WHITELIST_EMPTY) {
```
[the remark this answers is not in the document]
an4967: Okay, I will fix it

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -190,15 +266,151 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
 		ret = ble_drv_conn_is_any_active(is_active);
 	} break;
 
+	// Scanner
+	case BLE_CMD_START_SCAN: {
+		BLE_STATE_CHECK;
+
+		if (g_scan_ctx.state == BLE_SCAN_STARTED) {
+			break;
+		} if (g_scan_ctx.state == BLE_SCAN_CHANGING) {
```
jeongchanKim, line 282: Do you mean `else if`?

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -190,15 +266,151 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
 		bool *is_active = (bool *)msg->param;
 		ret = ble_drv_conn_is_any_active(is_active);
 	} break;
 
+	// Scanner
+	case BLE_CMD_START_SCAN: {
+		BLE_STATE_CHECK;
+
+		if (g_scan_ctx.state == BLE_SCAN_STARTED) {
+			break;
```
jeongchanKim, line 281: I think we'd better set the `ret` as success.

an4967: I will fix it

jeongchanKim (approved):
LGTM

pillip8282 (approved):
LGTM

PR #5207, opened 2021-10-06
Makefile.unix : Generate Smartfs image

jeongchanKim:


os/Makefile.unix
```diff
@@ -469,6 +469,16 @@ endif
...
+ifneq ($(findstring "smartfs",$(CONFIG_FLASH_PART_TYPE)), "smartfs")
+	$(Q) if [ ! -d "$(TOPDIR)/../tools/fs/contents-smartfs" ]; then \
+		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs"; \
+		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)"; \
+		mkdir -p "$(TOPDIR)/../tools/fs/contents-smartfs/$(CONFIG_ARCH_BOARD)/base-files"; \
+	fi; \
+	$(MAKE) smartfs
+endif
+
+
```
amogh-samsung, line 481: Please remove one of two extra line spaces here

amogh-samsung, line 481: Also, these changes will always build a smartfs image when make command is executed?

os/Makefile.unix
```diff
@@ -469,6 +469,16 @@ endif
 endif
 endif
 
+ifneq ($(findstring "smartfs",$(CONFIG_FLASH_PART_TYPE)), "smartfs")
```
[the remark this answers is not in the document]
sunghan-chang: This is not to make smartfs image for user binary. It's just making empty smartfs image.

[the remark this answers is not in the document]
sunghan-chang: @jeongchanKim Could you check belows?
1. Let's call these with some config. I mean when user enables new config, then let's make empty smartfs image.
    User could make it with some content so that it should be disabled by user.
2. Because this is not for user binary, there is no reason to be in the pass1 target. Let's move into the post target.

os/fs/Kconfig
```diff
@@ -79,3 +79,10 @@ config SYSLOG_DEVPATH
 		(such as "/dev/ttyS1")/
 
 endif
+
+config GENERATE_FS_IMAGE
+	bool "Generate fs image"
+	default n
+	---help---
```
sunghan-chang, line 86: 1. Let's make a dependency with `FLASH_PARTITION`.
2. I don't know this location is proper or not. This is not the feature of smartfs itself. How about os/Kconfig? @Taejun-Kwon , Could you give me an opinion?
3. Could you describe about fs content? If no content added before build, empty smartfs image will be generated.

jeongchanKim: 1. Added.
2. I think it is related to fs, so I located in here.
3. Added.

jeongchanKim: I added the dependency of `FS_SMARTFS`.

sunghan-chang (approved):

PR #5211, opened 2021-10-08
framework/wifimanager: add convert function from trwifi to wifimgr

pillip8282:
add convert function from trwifi to wifimgr


j11591 (approved):
LGTM!!

PR #5212, opened 2021-10-08
lwip/config: remove lwip if name config

pillip8282:
interface name is managed by network manager in TizenRT 3.1. So these
configurations aren't needed anymore.

### DIFF
--- a/os/net/lwip/configs/Kconfig
+++ b/os/net/lwip/configs/Kconfig
@@ -227,24 +227,6 @@ config NET_LWIP_SNMP_MAX_TREE_DEPTH
 
 endif #NET_LWIP_SNMP
 
-menu "Interface Name"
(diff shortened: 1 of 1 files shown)


os/net/lwip/configs/Kconfig
```diff
@@ -227,24 +227,6 @@ config NET_LWIP_SNMP_MAX_TREE_DEPTH
 
 endif #NET_LWIP_SNMP
 
-menu "Interface Name"
-config NET_ETH_IFNAME
```
jeongchanKim, line 231: If these configs are not needed anymore, how about removing them from defconfigs also?

pillip8282: I'll modify it next PR

an4967 (approved):
LGTM

PR #5217, opened 2021-10-13
arch/amebad/flash : Add log about flash chip id

sungkyu0:
During flash initialize, chip id is printed.
It will be helpful for developer.

ex) #up_flashinitialize: Manufacturer : 161 memory type : 64 capacity : 23

### DIFF
--- a/os/arch/arm/src/amebad/amebad_flash.c
+++ b/os/arch/arm/src/amebad/amebad_flash.c
@@ -350,6 +350,9 @@ FAR struct mtd_dev_s *up_flashinitialize(void)
 		priv->mtd.write = amebad_write;
 #endif
 		priv->nsectors = AMEBAD_NSECTORS;
+		u8 chip_id[4];
+		flash_read_id(NULL, chip_id, 4);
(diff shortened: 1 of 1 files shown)


thapav (dismissed):
lgtm

os/arch/arm/src/amebad/amebad_flash.c
```diff
@@ -350,6 +350,9 @@ FAR struct mtd_dev_s *up_flashinitialize(void)
 		priv->mtd.write = amebad_write;
 #endif
 		priv->nsectors = AMEBAD_NSECTORS;
+		u8 chip_id[4];
+		flash_read_id(NULL, chip_id, 4);
+		lldbg("Manufacturer : %d memory type : %d capacity : %d\n", chip_id[0], chip_id[1], chip_id[2], chip_id[3]);
```
jeongchanKim, line 355: `chip_id` is unsigned type, so I think it is better to print with `%u` instead of `%d`.

sungkyu0: I changed code with your advice
Thank you :)

jeongchanKim (approved):
LGTM

PR #5221, opened 2021-10-17
Show all binaries' text address when asserted

jeongchanKim:


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -440,6 +440,7 @@ static int loadingall_thread(int argc, char *argv[])
 		bmdbg("Failed to load common binary, %d\n", ret);
 		return BINMGR_OPERATION_FAIL;
 	}
+	strncpy(BIN_NAME(BM_CMNLIB_IDX), CONFIG_COMMON_BINARY_NAME, sizeof(CONFIG_COMMON_BINARY_NAME));
```
jeongarmy, line 443: If user wants to enable common binary, user should add partition for it with `bin` type and `common` name like below.
```
CONFIG_FLASH_PART_TYPE="....bin,bin,"
CONFIG_FLASH_PART_NAME="...,common,common,"
```
If you just register the partition like this, you don't need to copy the binary name of common binary here because it is saved as `BM_CMNLIB_NAME` when partition is registered.
Please refer the `binary_manager_upart` in binary_manager_data.c:235
I will add reference or guidance about common binary usage.

jeongchanKim: Okay. I will remove this line.
But it is not good to use the common binary name as `BM_CMNLIB_NAME` even there is a config for common binary name, `CONFIG_COMMON_BINARY_NAME`.
How do you think?

PR #5222, opened 2021-10-18
docs: Update debug guidelines for Fault scenarios

kishore-sn:
Update debug guidelines for all fault scenarios in the HowToDebugACrash.md file.

### DIFF
--- a/docs/HowToDebugACrash.md
+++ b/docs/HowToDebugACrash.md
@@ -10,7 +10,8 @@ which resulted in the crash.
 - [Prerequisites](#prerequisites)  
 - [How to use addr2line](#how-to-use-addr2line)
 - [How to debug a crash with addr2line](#how-to-debug-a-crash-with-addr2line)
-- [Example debugging scenraio](#Example-debugging-scenario)
+- [Example debugging scenario](#Example-debugging-scenario)
+- [Understanding Debug messages during a fault](#Understanding-debug-messages-during-a-fault)
 
 ## Prerequisites
 1. To use addr2line, please enable below config.  
@@ -46,8 +47,8 @@ menuconfig -> Chip selection -> Dump stack on assertions
 ```
 
 ## How to use addr2line
-After build, use **addr2line** from linux shell.  
-Please refer to linux man page.  
+After build, use **addr2line** from Linux shell.  
+Please refer to Linux man page.  
 (https://man7.org/linux/man-pages/man1/addr2line.1.html)  
 Simple command is like below.  
 ```
@@ -63,7 +64,7 @@ For this to work, the ARM toolchain needs to be configured in the PATH:
 $ arm-none-eabi-addr2line -e [filepath] -a [address]
 ```
 
-However, if the toolchain path is not configured, or if you have mulitple versions of toolchain, then you may specify the complete path to the
+However, if the toolchain path is not configured, or if you have multiple versions of toolchain, then you may specify the complete path to the
 required addr2line tool as follows:
 ```
 $ /home/tools/gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-addr2line -e [filepath] -a [address]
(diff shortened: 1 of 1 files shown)


docs/HowToDebugACrash.md
```diff
@@ -161,3 +162,110 @@ Callstack from values in stackdump
...
+
+Armv8-m : https://developer.arm.com/documentation/ddi0553/latest
+
+### MemManage Fault Status Register (MMFSR)
+
+1. ```MMARVALID```
+
+Debug msg: Access violation occurred at address (MMFAR) : 0xAddress
+
+This message indicates that the address which caused the memfault is available to us. Check the printed address value. If the address is zero, it is probably caused due to a null pointer access. If the address value is non-zero, then the fault can the caused due to one of the following reasons.
```
abhishek-samsung, line 179: Please change "then the fault can 'the' caused due to" --> "then the fault can 'be' caused due to"

docs/HowToDebugACrash.md
```diff
@@ -161,3 +162,110 @@ Callstack from values in stackdump
...
+Debug msg: Data access violation occurred.
+
+Use the PC value and the address printed in the debug message to find which variable is causing an invalid data access as mention in point (1) above.
+
+
+3. ```IACCVIOL```
+
+Debug msg: Instruction access violation occurred while fetching instruction from an Execute Never (XN) region.
+
+Execution from an MPU region with no execute permission. Check the line number of PC and make sure that corresponding MPU region has the proper previlages.
```
jeongchanKim, line 208: `previlages` -> `privileges`

docs/HowToDebugACrash.md
```diff
@@ -161,3 +162,110 @@ Callstack from values in stackdump
...
+In both the above cases, the developer can use the addr2line tool along with the PC and LR values to find the line number in code, where the error has occured. Please note the following points:
+- The LR value points to the next instruction to be executed after return from a function call. So, the LR will point to a line number after a function in which the fault would have occured.
+- The LR and PC might not contain exact line numbers if the function was declared as "inline", or if some compiler optimization has caused the function to be inlined.
+
+
+2. ```DACCVIOL```
+
+Debug msg: Data access violation occurred.
+
+Use the PC value and the address printed in the debug message to find which variable is causing an invalid data access as mention in point (1) above.
```
dineshdas-samsung, line 201: mention - > mentioned

docs/HowToDebugACrash.md
```diff
@@ -161,3 +162,110 @@ Callstack from values in stackdump
...
+
+The only difference lies in the IMPRECISERR bit. This bit is set if an imprecise data access violation has occured. It means that the current PC value does NOT point to the instruction with caused the data access violation. The violation is caused by some previous instruction, but was delayed because memory accesses are not always synchronous. In order to debug this type of fault, the developer can check the previous few lines of code starting from the current PC value. The developer can also use the objdump tool to print the disassembly of the code at the current PC value and then check each assembly instruction which is either loading or storing values into the memory.
```
dineshdas-samsung, line 231: Did u mean  this -> instruction  "which "caused the data access violation?

docs/HowToDebugACrash.md
```diff
@@ -161,3 +162,110 @@ Callstack from values in stackdump
...
+
+5. ```MLSPERR```
+
+Debug msg: Error occurred during lazy state preservation of Floating Point unit registers.
+
+This bit also indicates an error during stacking of registers during exception handling.
+
+
+### Bus Fault Status Register (BFSR)
+A bus fault is vary similar to memory management fault as mentioned above. Any access of data or instructions from an invalid memory address can cause a bus fault. If the MPU is enabled, then most of such invalid accesses will result in a corresponding memfault. However, if the MPU is disabled, all invalid memory accesses result in bus fault. Bus faults can either be data abort or prefetch abort, based on whether the fault occured while accessing data or instructions. The BFARVALID, STKERR, UNSTKERR, LSPERR fields of this register are similar to the corresponding bits in the memfault scenario. The IBUSERR bit indicates an instruction access fault; whereas, the PRECISERR and IMPRECISERR bits indicate a data access violation. The PRECISERR is similar to DACCVIOL in the memfault register.
```
jeongchanKim, line 229: Is this intended to use `vary`, not `very`?

PR #5223, opened 2021-10-19
armv7-m,v8-m/up_assert.c : Change to check user binar…

jeongchanKim:
…y fault

1) Show all bins' text addresses when the fault is from user binary.
2) Check the user heap corruption when the fault is from user binary.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -126,6 +126,9 @@ extern uint32_t g_assertpc;
 #define CONFIG_BOARD_RESET_ON_ASSERT 0
 #endif
 
+#define IS_FAULT_IN_USER_THREAD  (fault_tcb->uheap != NULL)
+#define IS_FAULT_IN_USER_SPACE   (is_kernel_space((void *)assert_pc) == false)
+
 /****************************************************************************
  * Private Data
  ****************************************************************************/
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -489,7 +481,8 @@ void up_assert(const uint8_t *filename, int lineno)
 		lldbg("No heap corruption detected\n");
 	}
 #ifdef CONFIG_APP_BINARY_SEPARATION
-	if (!is_kernel_fault) {
+	if (fault_tcb->uheap != NULL) {
```
sunghan-chang, line 484: Is it possible to redefine belows for readability?
1. ault_tcb->uheap != NULL  <=  fault thread
2. is_kernel_space((void *)assert_pc   <= fault space

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -502,7 +498,7 @@ void up_assert(const uint8_t *filename, int lineno)
 #endif
 
 #ifdef CONFIG_BINMGR_RECOVERY
-	if (is_kernel_fault == false) {
+	if (IS_FAULT_IN_KERNEL_SPACE) {
```
sangam-swami, line 501: readability looks odd here..
is it return true if fault in kernel?

sunghan-chang (approved):

PR #5225, opened 2021-10-19
tools/dump_tool: Update the tool to accomodate loadable_apps crash point detection

thapav:
This patch adds the functionality where the crash logs are scanned for
address in the range of application text address range in addition to the
kernel crashes. If yes, it displays the point of crash for the respective
apps.

### DIFF
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -495,6 +495,7 @@ void up_assert(const uint8_t *filename, int lineno)
 			lldbg("No heap corruption detected\n");
 		}
 	}
+	lldbg("Assert location (PC) : %08x\n", g_assertpc);
 #endif
 
 #if defined(CONFIG_BOARD_CRASHDUMP)
--- a/tools/debug/debugsymbolviewer.py
+++ b/tools/debug/debugsymbolviewer.py
@@ -89,7 +89,7 @@ def print_symbol(search_addr):
 	next_addr = 0x00000000
 	l = 0
 	h = len(symbol_lookup_table)
-	mid = (l + h) / 2
+	mid = (l + h) // 2
 
 	while (l <= h):
 		addr = symbol_lookup_table[mid - 1][0]
@@ -104,7 +104,7 @@ def print_symbol(search_addr):
 		else:
 			l = mid + 1
 
-		mid = (l + h) / 2
+		mid = (l + h) // 2
 
 	if (l > h):
 		if debug:
@@ -169,7 +169,7 @@ if (__name__ == '__main__'):
 
     setup_symbol_table(systemmap_path)
 
-    print('----------------------------------  DEBUG DISPLAY SYMBOLS -------------------------------')
+    print('------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------')
     print('Dump_address\t Symbol_address\t  Symbol_name\tFile_name')
     parse_file(log_file)
     print('-----------------------------------------------------------------------------------------')
--- a/tools/debug/debugsymbolviewer.py
+++ b/tools/debug/debugsymbolviewer.py
@@ -19,9 +19,9 @@
 from __future__ import print_function
 from math import log
 import re
+import sys
 
-systemmap_path = '../../build/output/bin/System.map'		# System map file
-log_file = '../dump_tool/log_file'				# Crash log file
+bin_path = '../../build/output/bin/'		# System map file
 symbol_lookup_table = []	# Lookup table created from contents of system map file
 debug = False			# Variable to print debug logs
 pc_value = 0x00000000		# Program counter value
@@ -56,9 +56,9 @@ def setup_symbol_table(sm_file):
 					# Extract kernel text start & end addresses from the table
 					symbol = s[2].rstrip('\n')
 					if symbol == '_stext_flash':
-						flash_text_start_addr = hex(int(s[0], 16))
+						flash_text_start_addr = hex(int(s[0], 16) + int(offset,16))
 					if symbol == '_etext_flash':
-						flash_text_end_addr= hex(int(s[0], 16))
+						flash_text_end_addr= hex(int(s[0], 16) + int(offset,16))
 					if symbol == '_stext_ram':
 						is_ram_text = True
 						ram_text_start_addr = hex(int(s[0], 16))
@@ -67,13 +67,13 @@ def setup_symbol_table(sm_file):
 
 	if debug:
 		# Printing the Created Symbol table
-		print('~~~~~~~~~~~~~~~~~~~~~~~~ SYMBOL TABLE START ~~~~~~~~~~~~~~~~~~~~~')
+		print('\n~~~~~~~~~~~~~~~~~~~~~~~~ SYMBOL TABLE START ~~~~~~~~~~~~~~~~~~~~~')
 		for line in symbol_lookup_table:
 			print("{0:x} {1}".format(line[0], line[1]))
 		print('~~~~~~~~~~~~~~~~~~~~~~~~SYMBOL TABLE END   ~~~~~~~~~~~~~~~~~~~~~')
 
 		# Printing the kernel text address ranges in ram and flash areas
-		print('~~~~~~~~~~~~~~~~~~~~~~~~~~  KERNEL TEXT RANGE ~~~~~~~~~~~~~~~~~~~~~~~')
+		print('~~~~~~~~~~~~~~~~~~~~~~~~~~  TEXT ADDRESS RANGE ~~~~~~~~~~~~~~~~~~~~~~~')
 		print('flash_text_start_addr\t=\t', flash_text_start_addr)
 		print('flash_text_end_addr\t=\t', flash_text_end_addr)
 		if (is_ram_text):
@@ -88,7 +88,7 @@ def print_symbol(search_addr):
 	addr = 0x00000000
 	next_addr = 0x00000000
 	l = 0
-	h = len(symbol_lookup_table)
+	h = len(symbol_lookup_table) - 1
 	mid = (l + h) // 2
 
 	while (l <= h):
@@ -108,11 +108,11 @@ def print_symbol(search_addr):
 
 	if (l > h):
 		if debug:
-			print('Symbol not found for address: {0} in system map file'.format(hex(search_addr)))
+			print('Symbol not found for address: {0} in map file'.format(hex(search_addr)))
 
 
 # Function to check if address lies in the kernel text address range
-def is_kernel_text_address(address):
+def is_text_address(address):
 
 	# Check the address
 	if (address >= ram_text_start_addr and address < ram_text_end_addr) or (address >= flash_text_start_addr and address < flash_text_end_addr):
@@ -158,18 +158,30 @@ def parse_file(log_file):
 
 					stack_addr = int(sub_word,16)
 					# Check if the stack address lies in kernel text address range
-					if (is_kernel_text_address(hex(stack_addr))):
+					if (is_text_address(hex(stack_addr))):
 						#If yes, print it's corresponding symbol
+                                                stack_addr = stack_addr - int(offset,16)
 						print_symbol(stack_addr)
 
 
 #Execution starts here
 if (__name__ == '__main__'):
+
     # init
+    bin_path += sys.argv[1]
+    log_file = sys.argv[2]
+    offset = sys.argv[3]
+    is_kernel_crash = int(sys.argv[4])	# 1 if kernel crash, 0 in case of application crash
 
-    setup_symbol_table(systemmap_path)
+    if debug:
+        if (is_kernel_crash):
+           print("Is kernel crash : yes")
+        else:
+           print("Is application crash : yes")
+        print("Map file : " + bin_path)
+        print("Offset (text range start) : " + offset)
+    setup_symbol_table(bin_path)
 
-    print('------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------')
     print('Dump_address\t Symbol_address\t  Symbol_name\tFile_name')
     parse_file(log_file)
     print('-----------------------------------------------------------------------------------------')
--- a/tools/dump_tool/HowToUseDumpTool.md
+++ b/tools/dump_tool/HowToUseDumpTool.md
@@ -308,7 +308,7 @@ binary_manager_send_request
 
 ----------------------------------------------------------
 
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  DEBUG DISPLAY SYMBOLS ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------
 Dump_address	 Symbol_address	  Symbol_name	File_name
 0xe0072f3	 0xe007290 	  up_assert	/root/tizenrt/os/include/assert.h:211
 0xe04f0bn	 0xe04f028 	  __FUNCTION__.6571
@@ -340,7 +340,7 @@ os_start
 
 ----------------------------------------------------------
 
---------------------------- DEBUG SYMBOLS IN KERNEL TEXT RANGE -------------------------
+------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------
 Dump_address	 Symbol_address	  Symbol_name	File_name
 0xe0542b3	 0xe0542a3 	  __FUNCTION__.6620
 0xe0086cd	 0xe0086cc 	  lowoutstream_putc	/root/product/.tizenrt/lib/libc/stdio/lib_lowoutstream.c:76
@@ -379,7 +379,7 @@ sched_get_priority_max
 
 ----------------------------------------------------------
 
---------------------------- DEBUG SYMBOLS IN KERNEL TEXT RANGE -------------------------
+------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------
 Dump_address	 Symbol_address	  Symbol_name	File_name
 0xe006337	 0xe006310 	  lowvsyslog	/root/product/.tizenrt/os/include/syslog.h:263
 0xe0060c1	 0xe0060c0 	  lib_noflush	/root/product/.tizenrt/os/include/tinyara/streams.h:605
--- a/tools/dump_tool/HowToUseDumpTool.md
+++ b/tools/dump_tool/HowToUseDumpTool.md
@@ -265,6 +265,8 @@ Ramdump received successfully..!
 DumpParser Script provides two interfaces: CUI and GUI
 
 ### DumpParser using CUI
+
+#### To get call stack using RAM dump
 1. Run Ramdump Parser Script
 ```
 cd $TIZENRT_BASEDIR/tools/dump_tool/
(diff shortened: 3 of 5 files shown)


tools/dump_tool/ramdumpParser.py
```diff
@@ -44,12 +44,18 @@
...
+g_stext_app_wifi=0
+g_etext_app_wifi=0
 
 config_path = '../../os/.config'
 elf_path = '../../build/output/bin/tinyara'
 debug_cmd = 'addr2line'
 file_data = 'HeapInfo'
 heapinfo_path = '../../os/include/tinyara/mm/heapinfo_internal.h'
+MICOM_BIN_PATH='../../build/output/bin/micom_dbg'
+WIFI_BIN_PATH='../../build/output/bin/wifi_dbg'
```
sunghan-chang, line 58: We could use another name for the application. How about keeping them as default and providing the way to set the path and the name with opt?

sunghan-chang:
Could you add result logs of functionality added in the commit description?

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -272,6 +272,35 @@ python dumpParser.py -r $TIZENRT_BASEDIR/build/output/bin/ramdump_0x02020000_0x0
...
+```
+*************************************************************
+dump_file         : None
+log_file          : .log_file
+Tinyara_elf_file  : ../../build/output/bin/tinyara.axf
+*************************************************************
+
+Crash point in app is as follows:
+main
+/root/tizenrt/loadable_apps/loadable_sample/wifiapp/wifiapp.c:85
```
sangam-swami, line 294: in this example kernel and app crash is mixed, can you change to example for kernel only or app only

tools/dump_tool/ramdumpParser.py
sangam-swami: how about getting names from crash logs itself, if the elf_show_all_bin_addr print name as standard always

tools/dump_tool/ramdumpParser.py
```diff
@@ -498,6 +504,63 @@ def unwind_backtrace_with_framepointer(self, fp, sp, lr, pc, ss):
...
+                    # Last subword of word[2] contains the PC value
+                    pc_value = int(t[-1],16)
+                    #print('PC value : ', hex(pc_value))
+                    continue
+            # Read the app text address and size
+            if 'elf_show_all_bin_addr:' in line:
+                if '[micom]' in line:
+                    word = line.split(':')
+                    t = word[2].split(',')
+                    g_stext_app_micom = int(t[0], 16)
```
jeongchanKim, line 536: The name of app can be changed by user.
And we can consider the common binary also. 
So I think it is not good to fix the variable name as app name.
How about like this? `g_stext_app1`, `g_stext_app2`?

tools/dump_tool/ramdumpParser.py
jeongchanKim: We should consider the common binary also.

tools/dump_tool/ramdumpParser.py
```diff
@@ -498,6 +504,63 @@ def unwind_backtrace_with_framepointer(self, fp, sp, lr, pc, ss):
...
+    with open(log_file) as searchfile:
+        for line in searchfile:
+            if 'up_registerdump:' in line:
+                word = line.split(':')
+
+                # word[1] contains the register name i.e R0 or R8
+                reg = word[1].split()
+
+                if reg[0] == 'R8':
+                    t = word[2].split( )
```
jeongchanKim, line 520: Could you leave the comment that what is the `word[2]` indicates?

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -272,6 +272,35 @@ python dumpParser.py -r $TIZENRT_BASEDIR/build/output/bin/ramdump_0x02020000_0x0
 ```
 2. See the Output
 
+#### To display Debug Symbols
```
jeongchanKim, line 275: Could you add description about the option like `-e'?
There are many options, `-e`, `-t` and `-g`.
It is better to use if there is an explanation.

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -272,6 +272,35 @@ python dumpParser.py -r $TIZENRT_BASEDIR/build/output/bin/ramdump_0x02020000_0x0
...
+
+    ex)
+    $ python3 ramdumpParser.py -e ../../build/output/bin/tinyara.axf -t ./log_file
+
+Example Call Stack Output
+```
+*************************************************************
+dump_file         : None
+log_file          : .log_file
+Tinyara_elf_file  : ../../build/output/bin/tinyara.axf
```
jeongchanKim, line 289: Most of the other boards uses `tinyara` which is elf format.
Only rtl8721csm uses `.axf` format.
So I think it is better to guide to use `tinyara` instead of `tinyara.axf`.

tools/dump_tool/HowToUseDumpTool.md
thapav: These usages show up in case of a wrong argument. It is in the ramdumpParser script itself

tools/dump_tool/ramdumpParser.py
```diff
@@ -498,6 +504,113 @@ def unwind_backtrace_with_framepointer(self, fp, sp, lr, pc, ss):
...
+
+    print('')
+    print('----------------------------------------------------------')
+    print('App Crash point is as follows:')
+    address1 = hex(pc_value)
+    address2 = hex(lr_value)
+    result = 0
+
+    # Check for pc & lr values in app1 text address range
+    if (address1 >= g_stext_app1 and address1 < g_etext_app1):
```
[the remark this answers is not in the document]
jeongchanKim: We have a config for number of apps.
I think we can use this config and use the loop.

tools/dump_tool/ramdumpParser.py
```diff
@@ -40,16 +41,21 @@
...
+g_stext_ram = 0
+g_etext_ram = 0
+
+g_app_idx = 0 # To extract number of apps dynamically
+
+app_name = []
 
 config_path = '../../os/.config'
-elf_path = '../../build/output/bin/tinyara'
+elf_path = '../../build/output/bin/tinyara.axf'
```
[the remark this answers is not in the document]
sunghan-chang: You can read the ext from os/Make.defs.

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -265,13 +265,132 @@ Ramdump received successfully..!
...
+----------------------------------------------------------
+App Crash point is as follows:
+App name : app
+0x0000001a
+main
+/root/product/.tizenrt/loadable_apps/product_apps/default_app/app.c:45
+
+----------------------------------------------------------
+
+--------------------------- DEBUG SYMBOLS IN KERNEL TEXT RANGE -------------------------
```
jeongchanKim, line 346: As my understanding, the crash happens from app.
But in this log, it says `DEBUG SYMBOLS IN KERNEL TEXT RANGE`.
Please check this.

tools/debug/debugsymbolviewer.py
```diff
@@ -169,7 +169,7 @@ def parse_file(log_file):
 
     setup_symbol_table(systemmap_path)
 
-    print('----------------------------------  DEBUG DISPLAY SYMBOLS -------------------------------')
+    print('------------------------------- DEBUG SYMBOLS IN TEXT RANGE -----------------------------')
```
jeongchanKim, line 172: I think previous `DEBUG DISPLAY SYMBOLS` is enough.
Is there any special reason to print `TEXT RANGE`?

tools/dump_tool/HowToUseDumpTool.md
jeongchanKim: Please change the md file also based on your changes.

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -265,13 +265,132 @@ Ramdump received successfully..!
...
+PC_value	 Symbol_address	  Symbol_name	File_name
+0x218ec12	 0x2000000 	  __psram_image2_text_start__
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+```
+Example Call Stack Output for App crash is as follows:
+```
+*************************************************************
+dump_file         : None
+log_file          : log_file
+Tinyara_elf_file  : ../../../output/app_dbg
```
jeongchanKim, line 331: I think `elf_file` is enough.

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -373,3 +492,11 @@ Like BaudRate, StopBits, Parity, Databits, HardwareFlowControl.
...
         }
 ```
+## Abort parser vs dump parser tool
+
+1. The abort_parser prints all the valid addresses corresponding to the dump addresses in the stackdump retrieved from the dump logs using the addr2line command.
+
+2. The dump tool:
+a. Shows crash points corresponding to kernel/app crashes (using addr2line command).
+b. Shows symbols and addresses corresponding to the text address range in the stackdump for kernel/app stacks (using addr2line command).
+c. Shows the whole call stack corresponding to a crash in case of cortex R devices.
```
jeongchanKim, line 502: Can we use this even with `cortex R`?
In `ramdumpParser.py`, there is a condition like below
```
		# Get ARM arch family
		if ('CONFIG_ARCH_FAMILY="armv8-m"' in data) or ('CONFIG_ARCH_FAMILY="armv7-m"' in data):
			#If architecture is cortex M, then run debugsymbolviewer script
                        #It displays the debug symbols corresponding to all the addresses in the kernel text address range
			os.system("python ../debug/debugsymbolviewer.py")
			return None
```
If we can use with cortext R, what is the difference for using this tool between cortex M and R?

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -373,3 +492,11 @@ Like BaudRate, StopBits, Parity, Databits, HardwareFlowControl.
                 return -1;
         }
 ```
+## Abort parser vs dump parser tool
```
sunghan-chang, line 495: This is too Samsung internal specific. Noone understands what abort parser is. Let's remove this from this doc. Could you take two pics of each on confluence? how to execute and the output of the tool. And then write what is different and what is better.

tools/debug/debugsymbolviewer.py
thapav: Kishore had mentioned in one of the comments that its not clear which symbols are these exactly, hence renamed them

tools/dump_tool/HowToUseDumpTool.md
thapav: This is because we can't retrieve the whole call stack for cortex M.

00000000 <main>:
0:   e92d4800        push    {fp, lr}

Whereas we can retrieve it for cortex R >
 00000000 <main>:
 0:   b580            push    {r7, lr}

tools/dump_tool/HowToUseDumpTool.md
thapav: Updated both commit details and the md file

tools/debug/debugsymbolviewer.py
```diff
@@ -104,7 +104,7 @@ def print_symbol(search_addr):
 		else:
 			l = mid + 1
 
-		mid = (l + h) / 2
+		mid = (l + h) // 2
```
sunghan-chang, line 107: Is this right? two slash?

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -265,13 +265,169 @@ Ramdump received successfully..!
...
+----------------------------------------------------------
+App Crash point is as follows:
+App name : wifi
+0x00001492
+main
+/root/tizenrt/loadable_apps/loadable_sample/wifiapp/wifiapp.c:85
+
+0x00003367
+binary_manager_send_request
+/root/tizenrt/framework/src/binary_manager/binary_manager_interface.c:97
```
sunghan-chang, line 307: What is second info?

tools/debug/debugsymbolviewer.py
thapav: Yes this is floor division as per Python 3.x.

tools/dump_tool/HowToUseDumpTool.md
thapav: This shows the addr2line output based on R14 & R15 values

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -265,13 +265,169 @@ Ramdump received successfully..!
...
 cd $TIZENRT_BASEDIR/tools/dump_tool/
 python dumpParser.py -r $TIZENRT_BASEDIR/build/output/bin/ramdump_0x02020000_0x0210c800.bin -e $TIZENRT_BASEDIR/build/output/bin/tinyara -g 0
 ```
 2. See the Output
 
+#### To display Debug Symbols/Crash point using assert logs
+1. Copy crash logs
+    First copy the crash logs to a file named log_file in tools/dump_tool/`<log_file>`
+2. Run Script in tools/dump_tool path
+    $ python3 ramdumpParser.py -e `<Elf path>` -t `<Log file path>`
```
abhishek-samsung, line 281: Can you please add some more information here regarding what elf should be given? As I can see the app and common elf paths are generated using the app names that were fetched from the log file but still in the below few examples wifi_dbg and common_dbg were given as an argument.

thapav: This elf file information is for the rest of the dump parser functionality.
Getting call stack for cortex R devices, if below criteria is  is FALSE:
_if ('CONFIG_ARCH_FAMILY="armv8-m"' in data) or ('CONFIG_ARCH_FAMILY="armv7-m"' in data):_

But the exact crash point & debug display symbols do not need this information elf information just as you pointe dout, they are being fetched using the dynamic apps from the dump_logs

tools/dump_tool/HowToUseDumpTool.md
sunghan-chang: 1. As far as I know, pc register has current operation +4 (thumb) or +8 (arm) so that it does not show the location which causes the problem. Is there any need to print this?
2. No one knows what they mean. Please write they show info of R14 and R15.

thapav: 1. Yes, in case of no up_registerdump() we are getting approximate crash point using pc value only
2) Added appropriate description

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -265,13 +265,180 @@ Ramdump received successfully..!
...
+App name : app2
+0x000014d6
+main
+/root/tizenrt/loadable_apps/loadable_sample/wifiapp/wifiapp.c:121
+
+[Probable crash point corresponding to the R15 register value (PC) might be -4 or -8 bytes]
+App name : app2
+0x000014e5
+main
+/root/tizenrt/loadable_apps/loadable_sample/wifiapp/wifiapp.c:113
```
sunghan-chang, line 310: 1. I think non-familiar on ARM outweigh familiar on ARM. (I mean many peoples don't know ARM register itself)
How about writing the title like below?
```md
[ Caller - return address (LR) - of the function which makes the crash ]
```
2. I don't know what you intend at `(PC) might be -4 or -8 bytes`. Does user need to check -4 or -8 from the printed address? or Is it already checked in the printed?
3. Could you write the item name?
```md
app name : xx
code addr? symbol addr? : xx
function name : xx
file : xxx
```

tools/dump_tool/HowToUseDumpTool.md
```diff
@@ -272,6 +272,121 @@ python dumpParser.py -r $TIZENRT_BASEDIR/build/output/bin/ramdump_0x02020000_0x0
...
+0xe007733	 0xe0076e8 	  up_memfault	/root/tizenrt/os/arch/arm/src/armv8-m/up_memfault.c:116
+0x10012345	 0x1001232c 	  up_doirq	/root/tizenrt/os/arch/arm/src/armv8-m/up_doirq.c:91
+0x10012055	 0x10012014 	  exception_common	/root/tizenrt/os/arch/arm/src/armv8-m/up_exception.S:156
+0xe001e4f	 0xe001e4e 	  sem_freeholder	/root/tizenrt/os/kernel/semaphore/sem_holder.c:163
+
+PC_value	 Symbol_address	  Symbol_name	File_name
+0x218ec12	 0x2000000 	  __psram_image2_text_start__
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+```
+Example Call Stack Output for App crash is as follows:
```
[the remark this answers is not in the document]
sunghan-chang: What's difference between above and below?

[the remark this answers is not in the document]
thapav: Above result is on public repo, below one on ecode.

sunghan-chang:
There is a conflict. Could you resolve it?

tools/dump_tool/HowToUseDumpTool.md
sunghan-chang: As you know, my English is poor. When you take my suggestion, please re-write the sentence.

tools/dump_tool/HowToUseDumpTool.md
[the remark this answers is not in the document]
sunghan-chang: oh ok. then let's remove ecode sample. As I said before, no one knows what ecode is.

tools/dump_tool/HowToUseDumpTool.md
thapav: 1) [ Caller - return address (LR) - of the function which has caused the crash ] sounds ok ?
    Should I add this information as and when a match is found for PC & LR values after the existing log like below?
    [ Crash point corresponding to the R14 register value (LR) ]
    [ Caller - return address (LR) - of the function which has caused the crash ] 
2) User needs to check -4 or -8 from the printed address.
3) I will modify

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -491,6 +491,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	} else {
 		assert_pc = (uint32_t)g_assertpc;
 	}
+	lldbg("g_assertpc: %08x\n", g_assertpc);
```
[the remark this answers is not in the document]
thapav: 1) I will share these results.
2) Thank you for pointing it out, I think it is better to move it after the CONFIG_APP_BINARY_SEPARATION section.
3) So I kept it similar to below so that we can extract using the same symbol name:
        lldbg("sp:         %08x\n", sp);
        lldbg("stack base: %08x\n", ustackbase);
        lldbg("stack size: %08x\n", ustacksize);

It gives something as below:
up_assert: g_assertpc : 02100df9
up_dumpstate: sp:         02103d90
up_dumpstate: stack base: 02103e70
up_dumpstate: stack size: 00001fe4

We can rename it though. Please let me know what fits better.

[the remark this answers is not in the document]
sunghan-chang: sp, stack base, stack size are commonly used. but g_assertpc is not. asserted pc or something else is better I think.

tools/dump_tool/HowToUseDumpTool.md
sunghan-chang: > Should I add this information as and when a match is found for PC & LR values after the existing log like below?
[ Crash point corresponding to the R14 register value (LR) ]
[ Caller - return address (LR) - of the function which has caused the crash ]

Sorry I don't get your point.

loadable_apps/loadable_sample/micomapp/micomapp.c
```diff
@@ -36,6 +36,9 @@ int main(int argc, char **argv)
 int micomapp_main(int argc, char **argv)
 #endif
 {
+//volatile int *p = 0;
```
jeongchanKim, line 39: Please remove unnecessary lines.

os/kernel/init/os_start.c
```diff
@@ -605,6 +605,8 @@ void os_start(void)
 #endif
 
 	DEBUGVERIFY(os_bringup());
+volatile int *p = 0;
```
jeongchanKim, line 608: Same as above.

tools/debug/debugsymbolviewer.py
```diff
@@ -19,9 +19,10 @@
 from __future__ import print_function
 from math import log
 import re
+import sys
 
-systemmap_path = '../../build/output/bin/System.map'		# System map file
-log_file = '../dump_tool/log_file'				# Crash log file
+systemmap_path = '../../build/output/bin/'		# System map file
+log_file = '../dump_tool/log_file'			# Crash log file
```
jeongchanKim, line 25: I think it is better to take the argument, not fixed file name.

tools/dump_tool/ramdumpParser.py
```diff
@@ -498,6 +505,156 @@ def unwind_backtrace_with_framepointer(self, fp, sp, lr, pc, ss):
...
+                g_app_idx = g_app_idx + 1
+
+            # Get the number of applications loaded
+            if 'Assert location (PC) :' in line:
+                word = line.split(':')
+                #word[2] contains the g_assertpc value
+                g_assertpc = int(word[2].strip(),16)
+
+    if g_app_idx:
+        print('Number of applicaions :', g_app_idx)
```
jeongchanKim, line 555: applicaions -> applications

tools/debug/debugsymbolviewer.py
```diff
@@ -19,9 +19,10 @@
 from __future__ import print_function
 from math import log
 import re
+import sys
 
-systemmap_path = '../../build/output/bin/System.map'		# System map file
-log_file = '../dump_tool/log_file'				# Crash log file
+systemmap_path = '../../build/output/bin/'		# System map file
```
jeongchanKim, line 24: This path is not only for system map.
Let's rename it as output path or something general.

thapav: Modified

thapav:
@sunghan-chang 
Updated all the review comments.
Verified it with armv7-m & armv7-r as well. Shared the logs over email.

sunghan-chang (approved):

PR #5227, opened 2021-10-21
os/mm/Kconfig : Change default value of MM_ASSERT_ON_FAIL as y

jeongchanKim:
When fail to allocate the memory, it makes assertion by default.


sunghan-chang:
How about below for the title?

os/mm/Kconfig: change default value of xx config to y

PR #5228, opened 2021-10-21
board/common: Define flash configurations, FLASH_START_ADDR and FLASH_SIZE

jeongarmy:
Define flash configurations : start address of flash and flash size.
So mkbootparam.py can use them defined config.h without handling parameters.

### DIFF
--- a/os/arch/Kconfig
+++ b/os/arch/Kconfig
@@ -848,18 +848,6 @@ config RAM_DDR_SIZE
 		If the board uses external DDR RAM memory, you need to define
 		the size of DDR RAM
 
-
-if BOOT_RUNFROMFLASH && ARCH_USE_MMU
-
-config FLASH_VSTART
-	hex "Boot FLASH start address (virtual)"
-	default 0x0
-	---help---
-		The virtual start address of installed boot FLASH.  "Boot" FLASH
-		refers to the FLASH that you link program code into.
-
-endif # BOOT_RUNFROMFLASH && ARCH_USE_MMU
-
 config ARCH_HAVE_SDRAM
 	bool
 	default n


jeongchanKim (commented):
Please check the board metadata also.
It has same value, so we can merge them.

os/board/common/Kconfig
```diff
@@ -11,6 +11,19 @@ config ARCH_BOARD_HAVE_FLASH
...
+
+config FLASH_START_ADDR
+    hex
+    default 0x8000000
+    help
+        A start address of flash
+
+config FLASH_SIZE
+    int
+    default 8388608
```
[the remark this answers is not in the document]
abhishek-samsung: As the following changes are only for rtl8721csm, please check if it is better to add them in board/rtl8721csm/Kconfig. As adding it in common with these default values might cause some confusion while using other boards as they have different flash start addresses and sizes.

os/board/common/Kconfig
```diff
@@ -11,6 +11,19 @@ config ARCH_BOARD_HAVE_FLASH
 		This is enabled by board config.
 
 if ARCH_BOARD_HAVE_FLASH && (NFILE_DESCRIPTORS != 0)
+
+config FLASH_START_ADDR
```
sunghan-chang, line 15: 1. They should not be in `(NFILE_DESCRIPTORS != 0)`. Let's take them out.
2. If config is invisible, it is set as default value when menuconfig is executed. Then, when we run menuconfig, we should modify this value every time. So, you should put them in board-specific Kconfig.

build/configs/rtl8721csm/board_metadata.txt
```diff
@@ -26,5 +26,3 @@ BL2="km4_boot_all"
 KERNEL="km0_km4_image2"
 OTA="km0_km4_image2"
 BOOTPARAM="bootparam"
-
-FLASH_START_ADDR=0x08000000
```
abhishek-samsung, line 30: FLASH_START_ADDR is used in common_download.sh. Please modify common_download.sh accordingly.
https://github.com/Samsung/TizenRT/blob/77f1253131f4e8fafd827a0bf4accfbf219c6139/build/configs/common_download.sh#L404

PR #5229, opened 2021-10-22
configs/rtl8721csm/loadable_apps/defconfig : Enable binary compression

jeongchanKim:
1) Enable binary compression with block size as 16384
2) Change the range of compression block size


os/tools/compression/src/compress.c
```diff
@@ -36,7 +36,7 @@
 #include "miniz/miniz.h"
 #endif
 
-#define MAX_BLOCK_SIZE 8192
```
sunghan-chang, line 39: could you let me know why this is necessary?

PR #5230, opened 2021-10-22
tools/set_bininfo.py : Rename common and user binary name

jeongchanKim:
Rename common and user binary name based on board and the version
1) common : common_[board]_[version]
2) user : [app_name]_[board]_[version]


sunghan-chang:
Let's use app1 and app2 instead of wifi and micom

PR #5235, opened 2021-10-26
driver/flash : Improve performance of erase

Taejun-Kwon:
When FS process format procedure, it takes too much time because Bulk erase takes too much time.
So to reduce format time, check flash state first, and then erase if therer are some data on it.


Taejun-Kwon:
can't use this PR, it failed many times during the powercut.

PR #5247, opened 2021-10-29
kernel/binary_manager_load.c : Print loaded binary path

jeongchanKim:

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -118,7 +118,11 @@ static int binary_manager_load_binary(int bin_idx, char *path, load_attr_t *load
 	while (retry_count < BINMGR_LOADING_TRYCNT) {
 		ret = load_binary(bin_idx, path, load_attr);
 		if (ret >= 0) {
-			bmvdbg("%s Load '%s' success! pid = %d\n", BINARY_COMP_TYPE, load_attr->bin_name, ret);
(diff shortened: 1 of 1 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -118,7 +118,11 @@ static int binary_manager_load_binary(int bin_idx, char *path, load_attr_t *load
 	while (retry_count < BINMGR_LOADING_TRYCNT) {
 		ret = load_binary(bin_idx, path, load_attr);
 		if (ret >= 0) {
-			bmvdbg("%s Load '%s' success! pid = %d\n", BINARY_COMP_TYPE, load_attr->bin_name, ret);
+			if (bin_idx == 0) {
+				lldbg("%s Load 'common'(%s) success!\n", BINARY_COMP_TYPE, path);
+			} else {
+				lldbg("%s Load '%s'(%s) success! pid = %d\n", BINARY_COMP_TYPE, load_attr->bin_name, path, ret);
```
sunghan-chang, line 124: As I said before, if we need to show it always to user, let's use printf. lldbg should be used with specific cases like assert.

jeongchanKim: I will change all `lldbg` to `printf` in the binary manager files with another PR.
Thanks.

sunghan-chang (approved):

PR #5252, opened 2021-11-02
tools/compression/src: terminate build when compression fails

abhishek-samsung:

### DIFF
--- a/os/include/tinyara/binfmt/compression/compression.h
+++ b/os/include/tinyara/binfmt/compression/compression.h
@@ -27,6 +27,12 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#define COMP_SUCCESS            0
+#define COMP_NOT_SUPPORTED      1
+#define COMP_FAILURE            2
+#define COMP_INVALID_INPUT      3
+#define COMP_USAGE_ERROR        4
+
 /****************************************************************************
  * Public Types
  ****************************************************************************/
--- a/os/tools/compression/src/compress.c
+++ b/os/tools/compression/src/compress.c
@@ -42,7 +42,7 @@
 static void show_usage(const char *progname)
 {
 	fprintf(stderr, "USAGE: %s <block size> <compression format> <uncompressed file name> <compressed file name>\n", progname);
-	exit(1);
+	exit(COMP_USAGE_ERROR);
 }
 
 static void compress_file(int block_size, int type, char *in_file, char *out_file)
@@ -65,6 +65,7 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 #if CONFIG_COMPRESSION_TYPE == LZMA
 	long unsigned int propsSize = LZMA_PROPS_SIZE;
 #endif
+	int exit_code = COMP_FAILURE;
 
 	ret = stat((char *)in_file, &buf);
 	if (ret < 0) {
@@ -162,6 +163,7 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 		ret = LzmaCompress(&out_buf[LZMA_PROPS_SIZE], &writesize, read_buf, (block_size - readsize), out_buf, &propsSize, 0, 1<<13 , -1, -1, -1, -1, 1);
 		if (ret != SZ_OK) {
 			printf("LZMA Compress failed, ret = %d\n", ret);
+			goto error;
 		}
 
 		printf("==> lzma_compress %d writesize %lu\n", index, writesize);
@@ -171,12 +173,13 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 		ret = mz_compress(out_buf, &writesize, read_buf, (block_size - readsize));
 		if (ret != Z_OK) {
 			printf("Miniz Compress failed, ret = %d\n", ret);
+			goto error;
 		}
 		printf("==> miniz_compress %d writesize %lu\n", index, writesize);
 #else
 		printf("Compression for type %d not supported\n", CONFIG_COMPRESSION_TYPE);
 		printf("Set CONFIG_COMPRESSION_TYPE to %d, then generate mkcompressimg again for this type", CONFIG_COMPRESSION_TYPE);
-		exit(1);
+		exit(COMP_NOT_SUPPORTED);
 #endif
 		phdr->secoff[index + 1] = phdr->secoff[index] + writesize;
 
@@ -228,6 +231,8 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 		}
 	}
 
+	exit_code = COMP_SUCCESS; /* Compression is successful */
+
 error:
 	if (phdr) {
 		free(phdr);
@@ -244,6 +249,7 @@ error:
 	if (out_fd > 0) {
 		close(out_fd);
 	}
+	exit(exit_code);
 }
 
 /****************************************************************************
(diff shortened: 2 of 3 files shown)


os/tools/compression/src/compress.c
```diff
@@ -244,6 +250,9 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 	if (out_fd > 0) {
 		close(out_fd);
 	}
+	if (compression_fail) {
```
jeongchanKim, line 253: Is it okay not to delete the output file for compression?

os/tools/compression/src/compress.c
```diff
@@ -65,6 +66,7 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 #if CONFIG_COMPRESSION_TYPE == LZMA
 	long unsigned int propsSize = LZMA_PROPS_SIZE;
 #endif
+	bool compression_fail = false;
```
Taejun-Kwon, line 69: It's not related to this file but below 'stat' with same parameter seems that already checked in main function.

os/tools/compression/src/compress.c
```diff
@@ -162,6 +164,8 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 		ret = LzmaCompress(&out_buf[LZMA_PROPS_SIZE], &writesize, read_buf, (block_size - readsize), out_buf, &propsSize, 0, 1<<13 , -1, -1, -1, -1, 1);
 		if (ret != SZ_OK) {
 			printf("LZMA Compress failed, ret = %d\n", ret);
+			compression_fail = true;
```
Taejun-Kwon, line 167: Only 'compression fail' is special case ?? only this case we should exit?

os/tools/compression/src/compress.c
abhishek-samsung: the output file for compression is opened with O_TRUNC flag.
https://github.com/Samsung/TizenRT/blob/114810f9e6cf82f01086e0092c3cb961ff3dc1fc/os/tools/compression/src/compress.c#L80
So I think we need not delete the output file

os/tools/compression/src/compress.c
abhishek-samsung: Will add exit to all the failure cases. Thanks.

os/tools/compression/src/compress.c
abhishek-samsung: As you have suggested, I have removed the stat for the input file as it is already checked in the main function before this function is called.

jeongchanKim: All `goto error` cases are compression fail.
So I think we don't need to set and use this variable.
We just call `exit()` from goto statement.

abhishek-samsung: Hello @jeongchanKim, calling exit from goto statement causes successful compression cases also to fail as the statements in error are executed every time. Please let me know if I did understand your comment correctly. So I think it's better to use this boolean variable and call exit only during the fail cases.

abhishek-samsung: Hello @Taejun-Kwon , the stat buf here for the input file is used to get the size, hence I am not deleting the above lines. I will revert the changes made.

sunghan-chang: Because of naming, we feel they `goto error` and `compression_fail = true;` look duplicated. How about renaming `error:` to `exit:` or `finish:` or something like that?

abhishek-samsung: Done. I have changed error to exit.

jeongchanKim: Okay. I understood. Thanks.

jeongchanKim (dismissed):
LGTM

os/tools/compression/src/compress.c
```diff
@@ -65,41 +66,47 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 #if CONFIG_COMPRESSION_TYPE == LZMA
 	long unsigned int propsSize = LZMA_PROPS_SIZE;
 #endif
+	bool compression_fail = false;
```
davidfather, line 69: @abhishek-samsung Let me suggest one thing for readability. Instead of compression_fail, what about using compression_success?
Our purpose and intention is to make compression successful, not to get failed.
So, it might be better for readability to use compression_success = true/false rather than compression_fail = true/false.
It's up to you.

sunghan-chang: @abhishek-samsung Could you let us know your opinion about @davidfather 's comment?

abhishek-samsung: I will change compression_fail to compression_success as @davidfather suggested.

davidfather (approved):
Looks good to me.

os/tools/compression/src/compress.c
```diff
@@ -65,41 +66,48 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 #if CONFIG_COMPRESSION_TYPE == LZMA
 	long unsigned int propsSize = LZMA_PROPS_SIZE;
 #endif
+	bool compression_success = true;
```
kishore-sn, line 69: How about having a variable named exit_code? Instead of setting the success value at each step, we can initialize the exit code to 2, and change it to 0 only if compression is success?

`int exit_code = 2;`
`...`
`/* Compression logic */`
`...`
`exit_code = 0`
`error:`
`exit(exit_code);`

jeongchanKim (approved):
LGTM

os/tools/compression/src/compress.c
abhishek-samsung: Thanks for the suggestion Kishore. I have modified the code accordingly.

os/tools/compression/src/compress.c
```diff
@@ -65,6 +65,7 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 #if CONFIG_COMPRESSION_TYPE == LZMA
 	long unsigned int propsSize = LZMA_PROPS_SIZE;
 #endif
+	int exit_code = 2;
```
sunghan-chang, line 68: Is it possible to show the meaning of "2" ?

os/tools/compression/src/compress.c
```diff
@@ -228,6 +231,8 @@ static void compress_file(int block_size, int type, char *in_file, char *out_fil
 		}
 	}
 
+	exit_code = 0; /* Compression is successful */
```
sunghan-chang, line 234: same as 0

abhishek-samsung: I didn't understand. Please correct me If I am wrong. exit 0 indicates that the execution was successful. Should I add that the execution was successful in the comment.

jeongchanKim: Yes. We also know nomarlly 0 is indicates the success and other variables are failure.
But it is better to add comment for better understanding.

os/tools/compression/src/compress.c
sunghan-chang: Then how about defining them at public header to know the reason easily?

```
xxx.h
#define COMP_SUCCESS          0
#define COMP_INVALID_TYPE  1
#define COMP_FAILURE            2
#define COMP_NOT_SUPPORT 3
#define COMP_INVALID_INPUT 4

xxx.c
exit(COMP_SUCCESS);
exit(COMP_FAILURE);
...
```

os/tools/compression/src/compress.c
sunghan-chang: What you understood was correct :)

os/tools/compression/src/compress.c
abhishek-samsung: Done.

os/include/tinyara/binfmt/compression/compression.h
```diff
@@ -27,6 +27,13 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+#define COMP_SUCCESS            0
+#define COMP_INVALID_TYPE       1
```
jeongchanKim, line 31: I think we can substitute `COMP_INVALID_TYPE` to `COMP_NOT_SUPPORT`.
Because they are all for not supported type.

abhishek-samsung: changed COMP_INVALID_TYPE to COMP_NOT_SUPPORTED.

sunghan-chang (approved):

PR #5257, opened 2021-11-03
os/kernel/binary_manager : Change lldbg to printf

jeongchanKim:
To show always to user, it is better to use printf instead of lldbg.
So change lldbg to printf in binary manager.

### DIFF
--- a/os/kernel/binary_manager/binary_manager.c
+++ b/os/kernel/binary_manager/binary_manager.c
@@ -206,22 +206,22 @@ int binary_manager(int argc, char *argv[])
 	return 0;
 errout_with_nobinary:
 	while (1) {
-		lldbg("=============== !!ERROR!! ============== \n");
+		printf("=============== !!ERROR!! ============== \n");
 		if (!is_found_bootparam) {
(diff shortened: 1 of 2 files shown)


davidfather (approved):
lgtm.

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -92,7 +92,7 @@ static void binary_manager_reset_board(void)
 	(void)irqsave();
 	sched_lock();
 	for (;;) {
-		lldbg("\nASSERT!! Push the reset button!\n");
+		printf("\nASSERT!! Push the reset button!\n");
```
sunghan-chang, line 95: This is the assert code so that lldbg is correct. Let's revert this.

sunghan-chang: How about adding comment about the reason why here uses lldbg instead of printf?

jeongchanKim: Reverted and added the reason why changed to printf in the commit msg.
Thanks.

sunghan-chang (approved):

PR #5259, opened 2021-11-05
tools/set_bininfo.py: apply TizenRT Package extension - trpk on each output

sunghan-chang:
TizenRT adds its own header at front of the output so that it's not
proper to use previous extensions like bin, hex and so on.
This commit applies "trpk" for that.
"trpk" stands for TizenRT Package.

### DIFF
--- a/os/tools/set_bininfo.py
+++ b/os/tools/set_bininfo.py
@@ -21,7 +21,9 @@ import os
 import sys
 import string
 
-EXT_NAME = sys.argv[1]
+SOURCE_EXT_NAME = sys.argv[1]
+TARGET_EXT_NAME = "trpk" # TizenRT Package (TizenRT Header + output)
+
 
 # This script changes the kernel binary name as "kernel_[board]_[version]".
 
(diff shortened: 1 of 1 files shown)


jeongchanKim (approved):
LGTM

os/tools/set_bininfo.py
```diff
@@ -21,7 +21,9 @@
 import sys
 import string
 
-EXT_NAME = sys.argv[1]
+SOURCE_EXT_NAME = sys.argv[1]
+TARGE_EXT_NAME = "trpk"
```
jeongchanKim, line 25: How about changing the name as `TARGET_EXT_NAME`?

davidfather: @sunghan-chang Plz, explain what trpk stands for in commit description as well as in this file.
I also think TARGET is correct.

sunghan-chang: @jeongchanKim @davidfather Done

davidfather (approved):
lgtm.

PR #5262, opened 2021-11-09
mm/mm_heap : Make assert on allocation failure when checking all heap…

jeongchanKim:
… regions

If CONFIG_MM_ASSERT_ON_FAIL is enabled and there are more then 2 heap regions,
if fail to allocate the first heap, it makes assert even though we can allocate the memory into the other heap.
So changed to check all heaps and make assert if we cannot allocate anywhere.

### DIFF
--- a/os/mm/kmm_heap/kmm_calloc.c
+++ b/os/mm/kmm_heap/kmm_calloc.c
@@ -55,7 +55,7 @@
  ****************************************************************************/
 
 #include <tinyara/config.h>
-
+#include <debug.h>
 #include <tinyara/mm/mm.h>
 
 #ifdef CONFIG_MM_KERNEL_HEAP
@@ -84,6 +84,7 @@ static void *kheap_calloc(size_t n, size_t elem_size, size_t retaddr)
 		}
 	}
 
+	mm_manage_alloc_fail(kheap, CONFIG_KMM_NHEAPS, n * elem_size, KERNEL_HEAP);
 	return NULL;
 }
 
@@ -106,6 +107,7 @@ static void *kheap_calloc(size_t n, size_t elem_size, size_t retaddr)
 #if CONFIG_KMM_NHEAPS > 1
 void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 {
+	void *ret;
 	struct mm_heap_s *kheap;
 	if (heap_index >= CONFIG_KMM_NHEAPS || heap_index < 0) {
 		mdbg("kmm_calloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, CONFIG_KMM_NHEAPS);
@@ -115,9 +117,17 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 	kheap = kmm_get_heap();
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	ARCH_GET_RET_ADDRESS
-	return mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	ret = mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	if (ret == NULL) {
+		mm_manage_alloc_fail(&kheap[heap_index], 1, n * elem_size, KERNEL_HEAP);
+	}
+	return ret;
 #else
-	return mm_calloc(&kheap[heap_index], n, elem_size);
+	ret = mm_calloc(&kheap[heap_index], n, elem_size);
+	if (ret == NULL) {
+		mm_manage_alloc_fail(&kheap[heap_index], 1, n * elem_size, KERNEL_HEAP);
+	}
+	return ret;
 #endif
 }
 #endif
--- a/os/mm/mm_heap/mm_manage_allocfail.c
+++ b/os/mm/mm_heap/mm_manage_allocfail.c
@@ -1,0 +1,90 @@
+/****************************************************************************
+ *
+ * Copyright 2021 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <debug.h>
+#include <string.h>
+#include <stdlib.h>
+#include <tinyara/mm/mm.h>
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+#include <assert.h>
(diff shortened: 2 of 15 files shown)


os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -0,0 +1,65 @@
...
+ * Return Value: None
+ *
+ ************************************************************************/
+
+void mm_manage_alloc_fail(void)
+{
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_SYSTEM_REBOOT_REASON)
+	WRITE_REBOOT_REASON(REBOOT_SYSTEM_MEMORYALLOCFAIL);
+#endif
+	mdbg("Allocation failed, size %u\n", size);
```
sunghan-chang, line 57: Where does 'size' come from?

sunghan-chang: Need to pass the `size` using function argu. How about passing the heap info as well? I mean kernel heap? user heap? heap index? Because this is the debug message.

jeongchanKim: Changed to pass the size and others. Thanks.

sunghan-chang: Let's print the debug infors, heap, idx.

jeongchanKim: I will add to print the `heap`, but it is not good to print the `heapix`.
It is not the allocation failure index, it is just the last index of the heap.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -0,0 +1,67 @@
...
+ * Return Value: None
+ *
+ ************************************************************************/
+
+void mm_manage_alloc_fail(struct mm_heap_s *heap, int heapidx, size_t size)
+{
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_SYSTEM_REBOOT_REASON)
+	WRITE_REBOOT_REASON(REBOOT_SYSTEM_MEMORYALLOCFAIL);
+#endif
+	mdbg("Allocation failed, size %u, heap %x\n", size, heap);
```
sangam-swami, line 57: can we update heapidx as well?

jeongchanKim: I will add to print the heap, but it is not good to print the heapix.
It is not the allocation failure index, it is just the last index of the heap.

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -115,9 +117,17 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 	kheap = kmm_get_heap();
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	ARCH_GET_RET_ADDRESS
-	return mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	ret = mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	if (ret == NULL) {
```
sangam-swami, line 121: how about adding mdbg message here, to identify from which function allocation failing

jeongchanKim: Added. Thanks.

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -83,7 +83,13 @@ static void *kheap_calloc(size_t n, size_t elem_size, size_t retaddr)
 			return ret;
 		}
 	}
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+	lldbg("Allocation failed from kernel heap.\n");
+#else
+	mdbg("Allocation failed from kernel heap.\n");
+#endif
```
abhishek-samsung, line 90: Since there is a repetition of the above logs, how about moving these to ` mm_manage_alloc_fail` 
Something like this
```void mm_manage_alloc_fail(mm_heap_s *heap, heap_start_index, number_of_heaps, bool is_kernel_heap , size);```
so that we can print these logs accordingly, if the number of heaps is 1, we can print the lldbg("Allocation failed from %d-th user/kernel heap.\n", heap_index);
if greater than 1 we can print lldbg("Allocation failed from user/kernel heap.\n");

Comment (jeongchanKim) [reply os/mm/kmm_heap/kmm_calloc.c:90]: But in this case, we cannot distinguish that heap is from kernel or user.
So I added in here.

Comment (abhishek-samsung) [reply os/mm/kmm_heap/kmm_calloc.c:90]: we can add an argument to mm_manage_alloc_fail to determine if it is user heap or kernel heap as I mentioned in the comment above. Please see if this helps.

File: os/mm/kmm_heap/kmm_calloc.c [hunk 5]
```diff
@@ -115,9 +122,27 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
...
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	ARCH_GET_RET_ADDRESS
-	return mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	ret = mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
+	if (ret == NULL) {
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+		lldbg("Allocation failed from %d-th kernel heap.\n", heap_index);
+#else
+		mdbg("Allocation failed from %d-th kernel heap.\n", heap_index);
+#endif
```
sunghan-chang, line 131: Let's define below in mm header and use it.
```
mm header
#ifdef CONFIG_MM_ASSERT_ON_FAIL
#define mmfaildbg lldbg    // When CONFIG_MM_ASSERT_ON_FAIL is enabled, we can't use the buffer way like mdbg because of board assert.
#else
#define mmfaildbg mdbg
#endif

xxx.c
mmfaildbg("xxxxx");
```

sunghan-chang: or we can modify debug.h

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -0,0 +1,97 @@
...
+ *   If needed, write the reboot reason and show the heap status through heapinfo.
+ *   And if CONFIG_MM_ASSERT_ON_FAIL is enabled, make assert.
+ *
+ * Return Value: None
+ *
+ ************************************************************************/
+
+void mm_manage_alloc_fail(struct mm_heap_s *heap, int heapidx, size_t size, int heap_type)
+{
+	if (heapidx > 1) {
```
abhishek-samsung, line 61: I think heapidx should be equal to 1 here. Also if it is set to one, we will be always printing the failure at 1st heap. I think we should pass the index separately along with the number of heap regions.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -0,0 +1,91 @@
...
+	const char *heap_type_str = (heap_type == KERNEL_HEAP) ? kernel_str : user_str;
+	mfdbg("Allocation failed from %s heap.\n", heap_type_str);
+	mfdbg(" - requested size %u\n", size);
+
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+	struct mallinfo info;
+	memset(&info, 0, sizeof(struct mallinfo));
+	for (int idx = 0; idx < heapidx; idx++) {
+		mm_mallinfo(heap, &info);
+	}
```
abhishek-samsung, line 73: This loop looks redundant. Should something be changed inside the loop?

sunghan-chang (approved):

PR #5275, opened 2021-11-11
kernel/binary_manager: Optimize buffer size for crc checking

jeongarmy:
Binary manager allocates buffer with binary RAM size for binary crc checking.
But the space with this size can be not enough after some binaries are loaded.
So if this size is larger than the largest free size, set buffer size to (the largest free size / 2).

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -746,6 +746,13 @@ int binary_manager_read_header(int type, char *devpath, void *header_data, bool 
 			bin_size = ((common_binary_header_t *)header_data)->bin_size;
 			crc_hash = ((common_binary_header_t *)header_data)->crc_hash;
 		}
+		struct mallinfo mem;
+#ifdef CONFIG_CAN_PASS_STRUCTS
+		mem = kmm_mallinfo();
+#else
+		(void)kmm_mallinfo(&mem);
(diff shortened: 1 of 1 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -746,6 +746,13 @@ int binary_manager_read_header(int type, char *devpath, void *header_data, bool
 			bin_size = ((common_binary_header_t *)header_data)->bin_size;
 			crc_hash = ((common_binary_header_t *)header_data)->crc_hash;
 		}
+		struct mallinfo mem;
+#ifdef CONFIG_CAN_PASS_STRUCTS
+		mem = kmm_mallinfo();
+#else
+		(void)kmm_mallinfo(&mem);
+#endif
+		crc_bufsize = crc_bufsize < mem.mxordblk ? crc_bufsize : (mem.mxordblk / 2);
```
sunghan-chang, line 755: This looks too dangerous. Even if crc_bufsize == mem.mxordblk - 1, crc_bufsize will be used. Then other thread will be failed to allocate. How about checking half of mem.mxordblk?
```
crc_bufsize = crc_bufsize < (mem.mxordblk / 2) ? crc_bufsize : (mem.mxordblk / 2);
```

sunghan-chang (approved):

PR #5278, opened 2021-11-12
os/binfmt: Fix memfault during Common binary Update test

kishore-sn:
During update test, we try to unload and reload all binaries. In case
of app binary, we disbale the MPU regions based on the MPU regs stored
in the tcb of apps. But, in case of Common binary, the MPU regions are
not stored anywhere. As a result, the MPU for common binary is not
disabled. So, at the time of reloading common binary, there is a memfault
while trying to copy code from binary to the read-only area of the code
section.

Issue is fixed by storing the common binary MPU registers in the binary
data structure and disabling the regions at the time of unload module.

### DIFF
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -86,9 +86,6 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	int pid;
 	int errcode;
 	int ret;
-#if (defined(CONFIG_SUPPORT_COMMON_BINARY) && (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU)))
-	uint32_t com_bin_mpu_regs[MPU_REG_NUMBER * MPU_NUM_REGIONS];	/* We need 3 register values to configure each MPU region */
-#endif
 
 	/* Sanity check */
 	if (load_attr && load_attr->bin_size <= 0) {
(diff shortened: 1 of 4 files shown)


jeongchanKim (approved):
LGTM

os/binfmt/binfmt_unloadmodule.c
```diff
@@ -64,6 +64,7 @@
 
 #include <tinyara/kmalloc.h>
 #include <tinyara/binfmt/binfmt.h>
+#include <tinyara/mpu.h>
```
jeongarmy, line 67: Could you add conditions for mpu, `#if defined(CONFIG_SUPPORT_COMMON_BINARY) && defined(CONFIG_ARCH_MPU)`?

os/binfmt/binfmt_loadbinary.c
```diff
@@ -193,15 +190,15 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 		uint8_t nregion = mpu_get_nregion_info(MPU_REGION_COMMON_BIN);
 #ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
 		/* Get MPU register values for MPU regions */
-		mpu_get_register_config_value(&com_bin_mpu_regs[0], nregion - 3, (uintptr_t)bin->alloc[ALLOC_TEXT], bin->textsize, true,  true);
-		mpu_get_register_config_value(&com_bin_mpu_regs[3], nregion - 2, (uintptr_t)bin->alloc[ALLOC_RO],   bin->rosize,   true,  false);
-		mpu_get_register_config_value(&com_bin_mpu_regs[6], nregion - 1, (uintptr_t)bin->alloc[ALLOC_DATA], bin->ramsize,  false, false);
+		mpu_get_register_config_value(&bin->cmn_mpu_regs[0], nregion - 3, (uintptr_t)bin->alloc[ALLOC_TEXT], bin->textsize, true,  true);
```
sunghan-chang, line 193: In case of common bin, `bin` is `g_lib_binp`, not `load_attr->binp`. Because binary manager calls load_binary with null for 3rd argu. Right? 
If this is right, I suggest minor change.
When you call binary_manager_load_binary, let's give addr of `g_lib_binp` instead of null. Then we can keep the same usages of the function and can remove if condition here. It's difficult to know exact content.

sunghan-chang: Is it right `g_lib_binp` for common bin? then how about changing the name?

sunghan-chang (approved):

PR #5279, opened 2021-11-12
kernel/binary_manager: Update bootparam when the binary not written in bootparam is loaded

jeongarmy:
Binary manager loads binaries based on boot parameters.
But if the binary written in bootparam is invalid, it checks the binary in another partition.
If it is valid, binary manager loads the binary and updates bootparam.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -32,6 +32,7 @@
 #include <stdint.h>
 #include <string.h>
 #include <queue.h>
+#include <stdbool.h>
 #include <semaphore.h>
 #include <sys/types.h>
 #include <sys/boardctl.h>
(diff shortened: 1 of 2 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -161,6 +161,7 @@ static int binary_manager_load(int bin_idx)
 	int ret;
 	int bin_count;
 	load_attr_t load_attr;
+	bool need_update_bootparam;
```
jeongchanKim, line 164: For boolean type, please include `stdbool.h`.

jeongchanKim: In this file, most of the name of bootparam is bp.
How about changing this variable also `need_update_bp`?

sunghan-chang (approved):

PR #5281, opened 2021-11-12
external/webserver: resolves the crash and init buf 

amitpurwar:
This resolves the crash due to tcb and initialize the buffer

### DIFF
--- a/external/webserver/http.c
+++ b/external/webserver/http.c
@@ -98,6 +98,38 @@ pthread_addr_t http_server_handler(pthread_addr_t arg)
 	struct sockaddr_in client_addr;
 	struct mq_attr mqattr;
 	struct http_server_t *server = (struct http_server_t *)arg;
+	int reuse = 1;
+
+	/*
+	 * Initialize socket and bind, start listening
+	 */
+
+	server->listen_fd = socket(AF_INET, SOCK_STREAM, 0);
+	if (server->listen_fd < 0) {
+		HTTP_LOGE("Error: Cannot create socket!!\n");
+		return NULL;
+	}
+
+	if (setsockopt(server->listen_fd, SOL_SOCKET, SO_REUSEADDR, &reuse, sizeof(reuse)) < 0) {
+		HTTP_LOGE("Error: Cannot set socket option!!\n");
+	}
+
+	server->servaddr.sin_family = AF_INET;
+	server->servaddr.sin_port = HTTP_HTONS(server->port);
+	server->servaddr.sin_addr.s_addr = INADDR_ANY;
+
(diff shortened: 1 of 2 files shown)


os/net/lwip/src/api/sockets.c
```diff
@@ -388,7 +388,12 @@ struct lwip_sock *get_socket(int s, pid_t pid)
 	}
 
 	tcb = sched_gettcb(pid);
-	DEBUGASSERT(tcb && tcb->group);
+	if (tcb == NULL) {
```
pillip8282, line 391: I think that if tcb is null then system is corrupted. So it should be stop.

davidfather: @amitpurwar I don't understand why you are trying to add this kind of strange code.

davidfather (changes requested):
I think this debug code should be on your local PC. What do you say?

os/net/lwip/src/api/sockets.c
amitpurwar: we are analyzing the issue further and update the code accordingly

external/webserver/http.c
```diff
@@ -98,6 +98,39 @@ pthread_addr_t http_server_handler(pthread_addr_t arg)
 	struct sockaddr_in client_addr;
 	struct mq_attr mqattr;
 	struct http_server_t *server = (struct http_server_t *)arg;
+	int reuse = 1;
+
+
```
pillip8282, line 103: please remove additional line

pillip8282 (commented):
Could you explain the patch in more detail. I couldn't understand it.

pillip8282 (commented):
I have questions about commit message "external/werbserver: Resolve the webserver crash"
commit message explained that connection accept thread(http_server_handler) called accept, and inside of accept it calls get_socket with terminate pid. However lwip_accept calls get_socket with getpid(). Then it pass the pid of http_server_handler not httptest_cb. So It doesn't access termitated pid(httptest_cb). Please let me know if I missed something.

amitpurwar:
> getpid(

amitpurwar:
> getpid(

amitpurwar:
when I tested, getpid() from lwip_accept give PID of httptest_cb

pillip8282 (approved):
Thanks for explanation

pillip8282 (approved):

amitpurwar:
when I tested, getpid() from lwip_accept give PID of httptest_cb

> [davidfather](/davidfather)

@davidfather  : Can you please review

davidfather (approved):
lgtm.

PR #5283, opened 2021-11-12
os/board/rtl8721csm: wlan add set tx data rate api

gexuyan:
This api is added to limit wlan tx data rate to a fixed value, and user also can input 0xff to switch back to auto data rate adaptation.


davidfather:
@j11591 Could you test this new API?

davidfather:
@gexuyan does this API work in not only station mode but also SoftAP mode?

gexuyan:
> 

Dear @davidfather , currently this API is only for station mode. Thank you.

PR #5284, opened 2021-11-15
framework/ble_manager: fix BLE minor bug

an4967:
- fix BLE minor bug
- fix BLE build warnings

### DIFF
--- a/framework/src/ble_manager/ble_manager_state.c
+++ b/framework/src/ble_manager/ble_manager_state.c
@@ -466,7 +466,7 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
 		2. auto connection
 		*/
 		ret = TRBLE_SUCCESS;
-		ble_client_ctx *ctx = (ble_client_ctx *)msg->param;
+		ble_client_ctx_internal *ctx = (ble_client_ctx_internal *)msg->param;
 
 		if (ctx == NULL) {
 			ret = TRBLE_INVALID_ARGS;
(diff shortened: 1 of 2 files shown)


framework/src/ble_manager/ble_manager_state.c
```diff
@@ -910,6 +910,12 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
 			}
 		}
 
+		if (ctx == NULL) {
```
jeongchanKim, line 913: If `ctx` is NULL, is it needed to run from 890 to 911?
If not, it is better to move this NULL checking at line 890.

an4967: @jeongchanKim 
At line:907, ctx is set after matching BLE context table.
This line means a context is not found in the full table.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -466,7 +466,7 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
 		2. auto connection
 		*/
 		ret = TRBLE_SUCCESS;
-		ble_client_ctx *ctx = (ble_client_ctx *)msg->param;
+		ble_client_ctx_internal *ctx = (ble_client_ctx_internal *)msg->param;
```
Taejun-Kwon, line 469: you should check validation(null check)of msg & param right before switch-case, not here

an4967: 'msg' is guaranteed that it is not NULL.
It is always created in ble_msg_handler.

pillip8282 (approved):
LGTM

PR #5286, opened 2021-11-15
os/binfmt: Change logging type for information msgs

kishore-sn:
Too many logs coming when berr logging type is enabled. So, change
unwanted information msgs from berr to binfo type.

### DIFF
--- a/os/binfmt/elf.c
+++ b/os/binfmt/elf.c
@@ -120,54 +120,54 @@ static void elf_dumploadinfo(FAR struct elf_loadinfo_s *loadinfo)
 {
 	int i;
 
-	berr("LOAD_INFO:\n");
-	berr("  textalloc:    %08lx\n", (long)loadinfo->textalloc);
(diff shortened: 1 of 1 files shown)


jeongchanKim (approved):
LGTM

sunghan-chang (approved):

sunghan-chang:
I agree this PR. But I want to suggest few things for dump_module.
1. Let's merge printing loading infos into dump_module because of dupplication
   https://github.com/Samsung/TizenRT/blob/master/os/binfmt/binfmt_loadbinary.c#L224
2. Let's rename the function name because the function operation is printing
3. How about making as below?
```c
void dump_module(const struct binary_s *bin)
{
    elf_save_bin_section_addr(bin);
    print_loading_info(bin);  <--- need to change for proper name
}

sunghan-chang:
Because my comment is not for this PR, let's merge this now.

kishore-sn:
I could not reply yesterday as i was busy with the other issues. I will take up your suggestion when we perform refractoring of the loader code.

PR #5287, opened 2021-11-15
binfmt/binfmt_loadbinary.c : move elf_save_bin_section_addr to before…

abhishek-samsung:
… calling exec_module

if a higher priority app crashes on start, then elf_show_all_bin_addr() doesn't print the
details of the app as they are saved by elf_save_bin_section_addr after the app has started
executing. So moving elf_save_bin_section_addr to before calling exec_module.

### DIFF
--- a/os/binfmt/binfmt_loadbinary.c
+++ b/os/binfmt/binfmt_loadbinary.c
@@ -211,14 +211,15 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	heap_table[binary_idx] = bin->heapstart;
 #endif
 
+	elf_save_bin_section_addr(bin);
 	/* Start the module */
 	pid = exec_module(bin);
 	if (pid < 0) {
 		errcode = pid;
 		berr("ERROR: Failed to execute program '%s': %d\n", filename, errcode);
+		elf_delete_bin_section_addr(bin);
 		goto errout_with_unload;
 	}
-	elf_save_bin_section_addr(bin);
 
 	/* Print Binary section address & size details */
 


os/binfmt/binfmt_loadbinary.c
```diff
@@ -211,14 +211,15 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	heap_table[binary_idx] = bin->heapstart;
 #endif
 
+	elf_save_bin_section_addr(bin);
```
jeongchanKim, line 214: Is there any chance to use the app text address even though that app is not executed well?
If execution fails, that app doesn't do anything.

abhishek-samsung: Hello @jeongchanKim , these changes are for the case where the app crashes before saving the text address (as the app execution has started before calling elf_save_bin_section_addr) and thereby not getting the app text address in the crash logs. This is for the case when exec_module has passed but the app has crashed before saving the text addresses.

jeongchanKim: Oh executed app can have higher priority.
Agreed. Thanks.

jeongchanKim (approved):
LGTM

os/binfmt/binfmt_loadbinary.c
sangam-swami: > Oh executed app can have higher priority. Agreed. Thanks.

yes. 
Can you please test easysetup crash issue with this patch on product_refrigerator

sangam-swami (dismissed):
Looks good

thapav (approved):
LGTM

sunghan-chang (approved):

sunghan-chang:
> if the app crashes on start, then elf_show_all_bin_addr() doesn't print the details of the
app as they are saved by elf_save_bin_section_addr after the app has started executing. So
moving elf_save_bin_section_addr to before calling exec_module.

The description is valid with higher priority app. Could you add that in the description?

abhishek-samsung:
> > if the app crashes on start, then elf_show_all_bin_addr() doesn't print the details of the
> > app as they are saved by elf_save_bin_section_addr after the app has started executing. So
> > moving elf_save_bin_section_addr to before calling exec_module.
> 
> The description is valid with higher priority app. Could you add that in the description?

Added.

sunghan-chang (approved):

jeongchanKim (approved):
LGTM

PR #5288, opened 2021-11-15
mm/*mm_heap : Return NULL if requested size is 0 and fix index for mm…

jeongchanKim:
…_manage in xalloc_at

1) Return NULL if requested size is 0. In this case, we will not call mm_manage_alloc_fail.
2) For xalloc_at, check the heap within its index

### DIFF
--- a/os/mm/kmm_heap/kmm_calloc.c
+++ b/os/mm/kmm_heap/kmm_calloc.c
@@ -114,6 +114,10 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 		return NULL;
 	}
 
+	if (n == 0 || elem_size == 0) {
+		return NULL;
+	}
+
 	kheap = kmm_get_heap();
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	ARCH_GET_RET_ADDRESS
@@ -147,6 +151,9 @@ FAR void *kmm_calloc(size_t n, size_t elem_size)
 #else
 	size_t retaddr = 0;
 #endif
+	if (n == 0 || elem_size == 0) {
+		return NULL;
+	}
 	return kheap_calloc(n, elem_size, retaddr);
 }
 
--- a/os/mm/kmm_heap/kmm_calloc.c
+++ b/os/mm/kmm_heap/kmm_calloc.c
@@ -73,7 +73,7 @@ static void *kheap_calloc(size_t n, size_t elem_size, size_t retaddr)
 	void *ret;
 	struct mm_heap_s *kheap = kmm_get_heap();
 
-	for (heap_idx = 0; heap_idx < CONFIG_KMM_NHEAPS; heap_idx++) {
+	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 		ret = mm_calloc(&kheap[heap_idx], n, elem_size, retaddr);
 #else
@@ -84,7 +84,7 @@ static void *kheap_calloc(size_t n, size_t elem_size, size_t retaddr)
 		}
 	}
 
-	mm_manage_alloc_fail(kheap, CONFIG_KMM_NHEAPS, n * elem_size, KERNEL_HEAP);
+	mm_manage_alloc_fail(kheap, HEAP_START_IDX, HEAP_END_IDX, n * elem_size, KERNEL_HEAP);
 	return NULL;
 }
 
@@ -109,8 +109,8 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 {
 	void *ret;
 	struct mm_heap_s *kheap;
-	if (heap_index >= CONFIG_KMM_NHEAPS || heap_index < 0) {
-		mdbg("kmm_calloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, CONFIG_KMM_NHEAPS);
+	if (heap_index > HEAP_END_IDX || heap_index < HEAP_START_IDX) {
+		mdbg("kmm_calloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, HEAP_END_IDX);
 		return NULL;
 	}
 
@@ -119,13 +119,13 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 	ARCH_GET_RET_ADDRESS
 	ret = mm_calloc(&kheap[heap_index], n, elem_size, retaddr);
 	if (ret == NULL) {
-		mm_manage_alloc_fail(&kheap[heap_index], 1, n * elem_size, KERNEL_HEAP);
+		mm_manage_alloc_fail(&kheap[heap_index], heap_index, heap_index, n * elem_size, KERNEL_HEAP);
 	}
 	return ret;
 #else
 	ret = mm_calloc(&kheap[heap_index], n, elem_size);
 	if (ret == NULL) {
-		mm_manage_alloc_fail(&kheap[heap_index], 1, n * elem_size, KERNEL_HEAP);
+		mm_manage_alloc_fail(&kheap[heap_index], heap_index, heap_index, n * elem_size, KERNEL_HEAP);
 	}
 	return ret;
 #endif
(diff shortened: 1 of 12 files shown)


os/mm/umm_heap/umm_zalloc.c
```diff
@@ -92,13 +92,13 @@ void *zalloc_at(int heap_index, size_t size)
 	ARCH_GET_RET_ADDRESS
 	ret = mm_zalloc(&BASE_HEAP[heap_index], size, retaddr);
 	if (ret == NULL) {
```
jeongarmy, line 94: Please add checking size for zalloc also, `if (size == 0)`.

os/mm/umm_heap/umm_malloc.c
```diff
@@ -113,13 +113,13 @@ void *malloc_at(int heap_index, size_t size)
 	ARCH_GET_RET_ADDRESS
```
jeongarmy, line 113: Please add checking size for malloc_at also, if (size == 0).

jeongarmy, line 113: Please add checking size for malloc_at also, `if (size == 0)`.

jeongchanKim: Added.

os/mm/umm_heap/umm_zalloc.c
jeongchanKim: added.

os/mm/kmm_heap/kmm_realloc.c
```diff
@@ -99,13 +99,13 @@ void *kmm_realloc_at(int heap_index, void *oldmem, size_t size)
 	ARCH_GET_RET_ADDRESS
```
jeongarmy, line 99: Please add checking size for kmm_realloc_at also, `if (size == 0)`.

jeongchanKim: added

jeongarmy (approved):
LGTM

os/mm/umm_heap/umm_zalloc.c
```diff
@@ -158,6 +162,9 @@ static void *heap_zalloc(size_t size, int s, int e, size_t retaddr)
 
 FAR void *zalloc(size_t size)
 {
+	if (size == 0) {
```
sunghan-chang, line 165: mm_malloc and mm_realloc have the same conditional inside so that there are two duplicated checking. We need to remove one.

jeongchanKim: Yes, I will change mm_xalloc in other PR not to check the size which is smaller than 1.
Thanks.

os/mm/kmm_heap/kmm_memalign.c
```diff
@@ -97,18 +97,22 @@ void *kmm_memalign_at(int heap_index, size_t alignment, size_t size)
...
+		return NULL;
+	}
+
 	kheap = kmm_get_heap();
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	ARCH_GET_RET_ADDRESS
 	ret = mm_memalign(&kheap[heap_index], alignment, size, retaddr);
 	if (ret == NULL) {
-		mm_manage_alloc_fail(&kheap[heap_index], 1, size, KERNEL_HEAP);
+		mm_manage_alloc_fail(&kheap[heap_index], heap_index, heap_index + 1, size, KERNEL_HEAP);
```
sunghan-chang, line 109: Logically, end index should be `heap_index`. But you implemented it with `<`. The end is set as `heap_index + 1`. I think you should modify the function to compare with `<=` for end and you should give the same value for the start and the end.
```
mm_manage_alloc_fail(xxx, heap_index, heap_index, size. yyy);

void mm_manage_alloc_fail(xxx, int start_idx, int end_idx, size, yyy)
{
    for (i = start_idx; i<= end_idx; i++) {
    ...
   }
}
```

os/mm/kmm_heap/kmm_realloc.c
```diff
@@ -158,7 +166,7 @@ FAR void *kmm_realloc(FAR void *oldmem, size_t newsize)
 		}
 	}
 
-	mm_manage_alloc_fail(kheap_new, CONFIG_KMM_NHEAPS, newsize, KERNEL_HEAP);
+	mm_manage_alloc_fail(kheap_new, 0, CONFIG_KMM_NHEAPS, newsize, KERNEL_HEAP);
```
sunghan-chang, line 169: All of allocation function use the same value for the start and the end.
The start is 0, the end is CONFIG_KMM_NHEAPS.
How about using definitions for them? You can use them in two locations of each file.
This way gives
1. good readability
2. prevent mismatch

os/mm/kmm_heap/kmm_memalign.c
jeongchanKim: I want to check just `heap[heap_index]`. So I think start can be `heap_index`, end can be `heap_index + 1` and compare the end with `<`.
If modified to `<=`, we should pass the end param with `-1`.
For example, 
````
mm_manage_alloc_fail(kheap, 0, CONFIG_KMM_NHEAPS, size, KERNEL_HEAP);
```
In this case, we should pass `CONFIG_KMM_NHEAPS - 1`, instead of `CONFIG_KMM_NHEAPS`.

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -114,18 +114,22 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 		return NULL;
 	}
 
+	if (n * elem_size == 0) {
```
Taejun-Kwon, line 117: instead of multiply, I think below is better.
```
if (n == 0 || elem_size == 0) {
     mdbg("wrong arg, n : %d elem_size : %d\n");
     return NULL;
}
```
BTW, in this situation you don't have to set errno as -EINVAL?

jeongchanKim: It is better to set the errno, but there is no errno setting in all mm.
I will consider it later.
Thanks.

os/mm/kmm_heap/kmm_memalign.c
sunghan-chang: In my opinion, a function should have certain operation and it should be used easily without code analysis. When I see end index, it should be `heap_index` or `CONFIG KMM NHEAPS - 1`. When we see the code, then we know +1 due to <.

jeongchanKim: Changed. Thanks.

davidfather:
@jeongchanKim This commit has two different purposes. Could you split this commit into two commits?

os/mm/umm_heap/umm_zalloc.c
```diff
@@ -137,7 +141,7 @@ static void *heap_zalloc(size_t size, int s, int e, size_t retaddr)
 		}
```
sunghan-chang, line 141: Need to check line 133.
heap_idx < e;
->
heap_idx <= e;

os/mm/umm_heap/umm_realloc.c
```diff
@@ -157,6 +167,6 @@ FAR void *realloc(FAR void *oldmem, size_t size)
 		}
```
sunghan-chang, line 167: How about using new definitions for existed code, line 158?
```
for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
```
There are several locations like above.

os/mm/umm_heap/umm_zalloc.c
jeongchanKim: Modified. Thanks.

os/mm/umm_heap/umm_realloc.c
jeongchanKim: Changed. Thanks.

jeongchanKim:
> @jeongchanKim This commit has two different purposes. Could you split this commit into two commits?

Split the commit. Thanks.

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -109,8 +109,12 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 {
 	void *ret;
 	struct mm_heap_s *kheap;
-	if (heap_index >= CONFIG_KMM_NHEAPS || heap_index < 0) {
-		mdbg("kmm_calloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, CONFIG_KMM_NHEAPS);
+	if (heap_index > HEAP_END_IDX || heap_index < 0) {
```
sunghan-chang, line 112: How about HEAP_START_IDX instead of 0?

jeongchanKim: Modified. Thanks.

sunghan-chang (approved):

PR #5290, opened 2021-11-16
framework/ble_manager: fix minor bug in BLE manager

an4967:
- check mq_receive return value

### DIFF
--- a/framework/src/ble_manager/ble_manager_autoconnect.c
+++ b/framework/src/ble_manager/ble_manager_autoconnect.c
@@ -113,7 +113,18 @@ static void *_autocon_process(void *param)
 			}
 		}
 		memset(buf, 0, sizeof(buf));
-		nbytes = mq_receive(ctx->mqfd, buf, sizeof(buf), 0);
+		for (;;) {
+			nbytes = mq_receive(ctx->mqfd, buf, sizeof(buf), 0);
+			if (nbytes < 0) {
+				int err_no = get_errno();
+				if (err_no == EINTR || err_no == EAGAIN) {
+					continue;
+				}
+				BLE_LOG_ERROR("[BLEMGR] mq_receive fail(fd : %d , err : %d)", ctx->mqfd, errno);
+				goto finish_auto;
+			}
+			break;
+		}
 		evt = (ble_autocon_event)(buf[0]);
 		BLE_LOG_DEBUG("[BLEMGR] mqueue evt : %d\n", evt);
 


framework/src/ble_manager/ble_manager_autoconnect.c
```diff
@@ -114,6 +114,10 @@ static void *_autocon_process(void *param)
 		}
 		memset(buf, 0, sizeof(buf));
 		nbytes = mq_receive(ctx->mqfd, buf, sizeof(buf), 0);
+		if (nbytes < 0) {
```
pillip8282, line 117: How about to consider EAGAIN or EINTR. if those events happened then it doesn't need to finish.

Taejun-Kwon: 1. if returned size is smaller than size of buf, is it acceptable?
2. Instead of mq_receive, you should use mq_timedreceive..otherwise, it will wait forever when some kind of crash happened.
3. If you apply mq_timedreceive, you should handle -EPIPE case.

an4967: @pillip8282 
Thanks. I added errno checking process.

framework/src/ble_manager/ble_manager_autoconnect.c
```diff
@@ -114,6 +114,13 @@ static void *_autocon_process(void *param)
...
+			if (errno == EINTR || errno == EAGAIN) {
```
jeongchanKim, line 118: How about like below?
```
int err_no = get_errno();
if (err_no == EINTR || err_no == EAGAIN) {
...
}
BLE_LOG_ERROR("... ", ..., err_no);
```
`errno` calls a function, so with above, we can reduce the function calling.

framework/src/ble_manager/ble_manager_autoconnect.c
an4967: @Taejun-Kwon 
1. Yes. 'buf' is for getting one byte.
2. The mq will be managed by BLE_manager.
3. Yes. I will check that errno when I use mq_timedreceive.

framework/src/ble_manager/ble_manager_autoconnect.c
an4967: @jeongchanKim 
I will fix it. Thanks

framework/src/ble_manager/ble_manager_autoconnect.c
pillip8282: I think that If mq_receive return the values(eg. eintr) it should try to do mq_receive again not to go to while(). Because ble_client_start_scan is already done. Please check my comment.

davidfather: @an4967 Are there any possibilities to have nbytes = 0?

an4967: @pillip8282 
Yes, you are right. 
I will check EINTR & EAGAIN with loop.

pillip8282 (approved):
LGTM

PR #5292, opened 2021-11-16
mm/mm_manage_allocfail.c : Change to use printf for user level

jeongchanKim:
In user level, use printf instead of lldbg for assertion on mem allocation fail.


davidfather (approved):
lgtm.

sunghan-chang (approved):

PR #5304, opened 2021-11-19
mm/*mm_heap : Move ARCH_GET_RET_ADDRESS at the first of the function

jeongchanKim:
This macro should be placed at the first of the function.
If no, it is better to use "__builtin_return_address(0)" instead.

### DIFF
--- a/os/mm/umm_heap/umm_memalign.c
+++ b/os/mm/umm_heap/umm_memalign.c
@@ -87,6 +87,10 @@
 void *memalign_at(int heap_index, size_t alignment, size_t size)
 {
 	void *ret;
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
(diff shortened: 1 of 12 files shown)


os/mm/umm_heap/umm_memalign.c
```diff
@@ -87,6 +87,10 @@
 void *memalign_at(int heap_index, size_t alignment, size_t size)
 {
 	void *ret;
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	size_t caller_retaddr = 0;
+	ARCH_GET_RET_ADDRESS(aller_retaddr)
```
jeongarmy, line 92: please fix it to `caller_retaddr`

sunghan-chang (approved):

PR #5309, opened 2021-11-19
arch/armv7-m, armv8-m: do board reset with kernel heap corruption

sunghan-chang:
When kernel heap corruption is detected in assert, we can't keep kernel alive.
Currently it keeps and tries to reload user binary. Finally it causes kernel assert again.
This commit resets the board with kernel heap corruption in assert.


jeongchanKim (approved):
LGTM

sangwon03 (approved):
LGTM

PR #5313, opened 2021-11-23
src/armv7-m,armv8-m: Modify to get the correct assert pc

abhishek-samsung:
During the faults and assert failure in app space, g_assert_location is set
to appropriate value to indicate the location of the crash and is used during
up_assert. In case of app SW assert, the crash location is obtained from
R14 of the interrupt context and in case of HW fault in app, it is obtained
from R15 of the interrupt context saved on the stack. In the case of kernel
asserts, we get the assert pc from the return address (current LR value) in
up_assert.

Moved ARCH_GET_RET_ADDRESS definition out of CONFIG_DEBUG_MM_HEAPINFO
as it is used in up_assert.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -109,9 +109,8 @@
 #include "mpu.h"
 
 bool abort_mode = false;
-#ifdef CONFIG_APP_BINARY_SEPARATION
-extern uint32_t g_assertpc;
-#endif
+extern uint32_t system_exception_location;
+extern uint32_t user_assert_location;
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -127,7 +126,7 @@ extern uint32_t g_assertpc;
 #endif
 
 #define IS_FAULT_IN_USER_THREAD(fault_tcb)  ((void *)fault_tcb->uheap != NULL)
-#define IS_FAULT_IN_USER_SPACE(assert_pc)   (is_kernel_space((void *)assert_pc) == false)
+#define IS_FAULT_IN_USER_SPACE(asserted_location)   (is_kernel_space((void *)asserted_location) == false)
 
 /****************************************************************************
  * Private Data
@@ -453,9 +452,29 @@ void dump_all_stack(void)
 
 void up_assert(const uint8_t *filename, int lineno)
 {
+	/* ARCH_GET_RET_ADDRESS should always be
+	 * called at the start of the function */
+
+	size_t kernel_assert_location = 0;
+	ARCH_GET_RET_ADDRESS(kernel_assert_location)
+
 	board_led_on(LED_ASSERTION);
 
 	abort_mode = true;
+
+	uint32_t asserted_location;
+
+	/* Extract the PC value of instruction which caused the abort/assert */
+
+	if (system_exception_location) {
+		asserted_location = (uint32_t)system_exception_location;
+		system_exception_location = 0x0;	/* reset */
+	} else if (user_assert_location) {
+		asserted_location = (uint32_t)user_assert_location;
+		user_assert_location = 0x0;
+	} else {
+		asserted_location = (uint32_t)kernel_assert_location;
+	}
 
 #if CONFIG_TASK_NAME_SIZE > 0
 	lldbg("Assertion failed at file:%s line: %d task: %s\n", filename, lineno, this_task()->name);
@@ -482,25 +501,15 @@ void up_assert(const uint8_t *filename, int lineno)
 			lldbg("No heap corruption detected\n");
 		}
 	}
-	lldbg("Assert location (PC) : %08x\n", g_assertpc);
 #endif
+	lldbg("Assert location (PC) : %08x\n", asserted_location);
 
 #if defined(CONFIG_BOARD_CRASHDUMP)
 	board_crashdump(up_getsp(), this_task(), (uint8_t *)filename, lineno);
 #endif
 
 #ifdef CONFIG_BINMGR_RECOVERY
-	uint32_t assert_pc;
-
-	/* Extract the PC value of instruction which caused the abort/assert */
-
-	if (current_regs) {
-		assert_pc = current_regs[REG_R15];
-	} else {
-		assert_pc = (uint32_t)g_assertpc;
-	}
-
-	if (IS_FAULT_IN_USER_SPACE(assert_pc)) {
+	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
 		/* Recover user fault through binary manager */
 		binary_manager_recover_userfault();
 	} else
--- a/os/arch/arm/src/armv7-m/up_hardfault.c
+++ b/os/arch/arm/src/armv7-m/up_hardfault.c
@@ -85,9 +85,7 @@
 
 #define INSN_SVC0        0xdf00	/* insn: svc 0 */
 
-#ifdef CONFIG_APP_BINARY_SEPARATION
-uint32_t g_assertpc;
-#endif
+uint32_t system_exception_location;
 /****************************************************************************
  * Private Data
  ****************************************************************************/
@@ -115,13 +113,8 @@ uint32_t g_assertpc;
 
 int up_hardfault(int irq, FAR void *context, FAR void *arg)
 {
-#if defined(CONFIG_DEBUG_HARDFAULT) || !defined(CONFIG_ARMV7M_USEBASEPRI) || defined(CONFIG_BINMGR_RECOVERY)
 	uint32_t *regs = (uint32_t *)context;
-#endif
-
-#ifdef CONFIG_APP_BINARY_SEPARATION
-	g_assertpc = regs[REG_R15];
-#endif
+	system_exception_location = regs[REG_R15];
 	/* Get the value of the program counter where the fault occurred */
 
 #ifndef CONFIG_ARMV7M_USEBASEPRI
--- a/os/arch/arm/src/armv7-m/up_memfault.c
+++ b/os/arch/arm/src/armv7-m/up_memfault.c
@@ -75,6 +75,7 @@
 #include "nvic.h"
 #include "up_internal.h"
 
+extern uint32_t system_exception_location;
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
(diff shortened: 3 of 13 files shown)


os/arch/arm/src/armv7-m/up_memfault.c
```diff
@@ -121,6 +122,8 @@ int up_memfault(int irq, FAR void *context, FAR void *arg)
 	uint32_t *regs = (uint32_t *)context;
 	uint32_t cfsr = getreg32(NVIC_CFAULTS);
 	uint32_t mmfar = getreg32(NVIC_MEMMANAGE_ADDR);
+	g_assertpc = regs[REG_R15];
```
kishore-sn, line 125: In case of IACCVIOL, the PC value might be invalid. So let us set R14 for that case.

os/arch/arm/src/armv7-m/up_hardfault.c
```diff
@@ -115,13 +113,9 @@ uint32_t g_assertpc;
...
 int up_hardfault(int irq, FAR void *context, FAR void *arg)
 {
-#if defined(CONFIG_DEBUG_HARDFAULT) || !defined(CONFIG_ARMV7M_USEBASEPRI) || defined(CONFIG_BINMGR_RECOVERY)
 	uint32_t *regs = (uint32_t *)context;
-#endif
-
-#ifdef CONFIG_APP_BINARY_SEPARATION
 	g_assertpc = regs[REG_R15];
-#endif
+
```
deepaksrma, line 118: can you remove this unnecessary line

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -474,6 +472,18 @@ void up_assert(const uint8_t *filename, int lineno)
...
 	abort_mode = true;
 
+	uint32_t assert_pc;
+
+        /* Extract the PC value of instruction which caused the abort/assert */
+
+	if (g_assertpc) {
+		/* g_assertpc is set only during up_assert syscall or during faults*/
+		assert_pc = (uint32_t)g_assertpc;
+		g_assertpc = 0x0; /* reset */
```
deepaksrma, line 482: you can give a tab space between comment and code

os/arch/arm/src/armv8-m/up_hardfault.c
```diff
@@ -120,13 +118,9 @@ uint32_t g_assertpc;
...
 int up_hardfault(int irq, FAR void *context, FAR void *arg)
 {
-#if defined(CONFIG_DEBUG_HARDFAULT) || !defined(CONFIG_ARMV8M_USEBASEPRI) || defined(CONFIG_BINMGR_RECOVERY)
 	uint32_t *regs = (uint32_t *)context;
-#endif
-
-#ifdef CONFIG_APP_BINARY_SEPARATION
 	g_assertpc = regs[REG_R15];
-#endif
+
```
deepaksrma, line 123: remove this as well

kishore-sn:
Please modify the commit msg as follows:

During the faults and assert failure in app space, g_assertpc is
set to the appropriate value to indicate the location of the crash and is used
during up_assert. In case of app SW assert, the crash location is obtained from 
R14 of the interrupt context and in case of HW fault in app, it is obtained from 
R15 of the interrupt context saved on the stack. In the case of kernel asserts, we get the
assert pc from the return address (which is the current LR value) in up_assert.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -457,6 +455,18 @@ void up_assert(const uint8_t *filename, int lineno)
...
+	uint32_t assert_pc;
+
+	/* Extract the PC value of instruction which caused the abort/assert */
+
+	if (g_assertpc) {
+		/* g_assertpc is set only during up_assert syscall or during faults*/
+		assert_pc = (uint32_t)g_assertpc;
+		g_assertpc = 0x0; /* reset */
+	} else {
+		assert_pc = (uint32_t) __builtin_return_address(0);
```
jeongchanKim, line 467: As discussed, `__builtin_return_address` works only with some compilers, such as gcc.
Please add condition or replace it to `ARCH_GET_RET_ADDRESS` which is in mm.h

os/arch/arm/src/armv7-m/up_memfault.c
abhishek-samsung: Done

os/arch/arm/src/armv7-m/up_assert.c
abhishek-samsung: I have replaced builtin_return_address with ARCH_GET_RET_ADDRESS

jeongchanKim: Thanks for modification.
But as my test, we should call `ARCH_GET_RET_ADDRESS` at the first of the function, without calling any other function.
But in this code, there is `board_led_on`.
Please move the calling of `ARCH_GET_RET_ADDRESS` before calling `board_led_on`.

abhishek-samsung: I moved ARCH_GET_RET_ADDRESS to the start of the function. Thanks for pointing it out.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -457,6 +455,20 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	abort_mode = true;
 
+	uint32_t assert_pc;
```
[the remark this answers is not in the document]
abhishek-samsung: I have changed g_assertpc to g_assert_location and assert_pc to fault_location.

thapav:
Please mention the reason of moving ARCH_GET_ADDRESS definition out of CONFIG_DEBUG_MM_HEAPINFO in the commit description as well.

os/include/tinyara/mm/mm.h
```diff
@@ -254,6 +249,11 @@ typedef size_t mmaddress_t;		/* 32 bit address space */
 #error Unknown CONFIG_ARCH option, malloc debug feature wont work.
 #endif
 
+/* typedef is used for defining size of address space */
+
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+typedef size_t mmaddress_t;             /* 32 bit address space */
```
sangam-swami, line 255: pls change commit message 
ARCH_GET_ADDRESS  ==> ARCH_GET_RET_ADDRESS

abhishek-samsung: Thanks for pointing it out. Modified.

sunghan-chang:
Could you let us know how to test (confirm? verify?)?
Is it possible to verify this PR with existed exception example?
Could you let me know the result?

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -474,6 +478,18 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	abort_mode = true;
 
+	uint32_t fault_location;
+
+        /* Extract the PC value of instruction which caused the abort/assert */
+
+	if (g_assert_location) {
```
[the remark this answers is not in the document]
abhishek-samsung: 1. I will change `g_assert_location` to `system_exception_location` as suggested.
2. `caller_retaddr` can't be changed as the macro is defined with it.
https://github.com/Samsung/TizenRT/blob/316e807f7bd4b5345e42f1fb1a63c991e6981f16/os/include/tinyara/mm/mm.h#L244-L247
3. `asserted_location` for `fault_location`
    How about using `system_exception_location` here? 
    As it can contain both the assert failure location or the fault location, I think system_exception_location is better.
    I will change `g_assert_location` to `g_system_exception_location`.

Please let me know your thoughts.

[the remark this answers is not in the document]
sunghan-chang: 2. No. it's the same as `#define ARCH_GET_RET_ADDRESS(x) \`. It does not mean you should use `caller_retaddr`. It's the same as function argument. If function `A(int B, int C)` as prototype, you can call it as `A(F, G);`
3. What I mean from `system_exception_location` and `user_assert_location` is the location which system makes the exception and the location which user intends assert. But `fault_location` should cover both and it does not matter who (system or user) makes, it only concerns location where causes assertion. Whatever it is, it should include both cases, system makes exception and user intends assert.

os/arch/arm/src/armv7-m/up_svcall.c
```diff
@@ -515,6 +510,10 @@ int up_svcall(int irq, FAR void *context, FAR void *arg)
 			return INDEX_ERROR;
 		}
 
+		if (cmd == SYS_up_assert) {
+			g_assert_location = regs[REG_R14];
```
sunghan-chang, line 514: Please leave comment why this is necessary here.

os/arch/arm/src/armv8-m/up_assert.c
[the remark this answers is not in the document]
abhishek-samsung: 2. I will change caller_retaddr to `kernel_assert_location` as we use this only in the case of kernel assert failure
3. So as suggested we have `user_assert_location` and `system_exception_location`
    fault (user or kernel) -> fault_handler (here `system_exception_location` is set) -> up_assert
    user assert -> sv call (here `user_assert_location` is set) -> up_assert
    kernel assert -> up_assert (`kernel_assert_location` is set)
The final `fault_location` will be equal to `user_assert_location` or `kernel_assert_location` or `system_exception_location` .

abhishek-samsung:
> Could you let us know how to test (confirm? verify?)? Is it possible to verify this PR with existed exception example? Could you let me know the result?

I have tested this patch manually by adding asserts and faults in user and kernel.

Example (kernel assert failure test) :

add panic in kernel/init/os_start.c

```
605 #endif
606         PANIC();
607         DEBUGVERIFY(os_bringup());
```

get the pc value from the logs

up_assert: Checking kernel heap fororruption...
up_assert: No heap corruption detected
up_asser Assert location (PC) : `0x10010efd`

use add2line to find the line number

```
abhishek@abhishek-VirtualBox:~/work/public_fork/TizenRT/os$ arm-none-eabi-addr2line -a -f -e ../build/output/bin/tinyara.axf 0x10010efd
0x10010efd
os_start
/root/tizenrt/os/kernel/init/os_start.c:607
```
Similarly, I checked for the user side. To create a fault, I used the below code.

```
volatile int * p = 0;
* p = 10;
```

os/arch/arm/src/armv8-m/up_assert.c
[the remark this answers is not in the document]
sunghan-chang: Others are ok to me, but I don't know fault_location. It could be considered as location from faults.

sunghan-chang:
Is there any reason you used your own code, not existed sample?

os/arch/arm/src/armv8-m/up_assert.c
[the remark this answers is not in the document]
abhishek-samsung: Okay, I will use asserted_location for fault_location as suggested earlier.

abhishek-samsung:
> Is there any reason you used your own code, not existed sample?

Hi @sunghan-chang, by existing sample, I think you mean recovery test. Please correct me if I am wrong. If so, it only covers the user side fault and assert cases. So I had to check it with my own code for kernel side assert and faults, and also for the flat build.

sunghan-chang (approved):

PR #5314, opened 2021-11-23
kernel/binary_manager: Add deinit functions for binary reloading

jeongarmy:
Add binary_manager_deinit_modules function to call deinit functions of modules which need to clear or reset used resources like device state before loading again after binary unloading (i.e. network modules like LWIP)

### DIFF
--- a/os/kernel/binary_manager/Make.defs
+++ b/os/kernel/binary_manager/Make.defs
@@ -24,7 +24,7 @@ CSRCS += binary_manager.c binary_manager_getinfo.c binary_manager_response.c
 CSRCS += binary_manager_data.c binary_manager_bootparam.c
 
 ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
-CSRCS += binary_manager_load.c binary_manager_callback.c
+CSRCS += binary_manager_load.c binary_manager_callback.c binary_manager_deinit.c
 ifeq ($(CONFIG_BINMGR_RECOVERY),y)
 CSRCS += binary_manager_recovery.c
 endif # CONFIG_BINMGR_RECOVERY
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -393,44 +393,6 @@ static int binary_manager_terminate_binary(int bin_idx)
 	return OK;
 }
 
-#ifdef CONFIG_BINMGR_RECOVERY
-/****************************************************************************
- * Name: binary_manager_reload
- *
- * Description:
- *   This function will terminate all the task/thread created by the binary
- *   i.e input binary id.
- *   It is called after all children are excluded from scheduling by fault recovery.
- *   It terminates its children, unloads binary and then it will load the binary.
- *
- * Input parameters:
- *   binid   -   The pid of binary to be reload
- *
- * Returned Value:
- *   None
- *
- ****************************************************************************/
-static int binary_manager_reload(int bin_idx)
-{
-	int ret;
-
-	/* Terminate binary */
-	ret = binary_manager_terminate_binary(bin_idx);
-	if (ret != OK) {
-		bmdbg("Failed to terminate binary %s\n", BIN_NAME(bin_idx));
-		return BINMGR_OPERATION_FAIL;
-	}
-
-	/* Load binary */
-	ret = binary_manager_execute_loader(LOADCMD_LOAD, bin_idx);
-	if (ret != OK) {
-		bmdbg("Failed to load binary, bin_idx %d\n", bin_idx);
-		return BINMGR_OPERATION_FAIL;
-	}
-	return BINMGR_OK;
-}
-#endif
-
 /****************************************************************************
  * Name: loading_thread
  *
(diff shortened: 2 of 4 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -512,18 +474,26 @@ static int loadingall_thread(int argc, char *argv[])
  * Name: reloading_thread
  *
  * Description:
- *   This thread reloads binary with binary index.
+ *   This function terminates all the task/thread created by the binary.
+ *   If common binary is enabled, it terminates all user binaries and common binary.
+ *   Otherwise, it terminates a faulty binary only.
+ *   It is called after all children are excluded from scheduling by fault recovery.
+ *   It terminates all children of bianary, unloads binary and then it will load the binary.
```
thapav, line 481: binary*

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -533,8 +503,9 @@ static int reloading_thread(int argc, char *argv[])
 	 * So the common binary should be unloaded after termination of all user binaries
 	 * to clear used resources normally.
 	 */
+	load_cmd = LOADCMD_LOAD_ALL;
 
-	/* 1. Unload all user binaries */
+	/* Unload all user binaries */
```
sangam-swami, line 508: do we need to terminate all user and common binary, if fault is in one of user app?

os/kernel/binary_manager/Make.defs
```diff
@@ -24,7 +24,7 @@ CSRCS += binary_manager.c binary_manager_getinfo.c binary_manager_response.c
 CSRCS += binary_manager_data.c binary_manager_bootparam.c
 
 ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
-CSRCS += binary_manager_load.c binary_manager_callback.c
+CSRCS += binary_manager_load.c binary_manager_callback.c binary_manager_deinit.c
```
jeongchanKim, line 27: Is `binary_manager_deinit.c` needed even if there is no recovery?
I think deinit is needed only when binary manager tries to reload the modules.

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: Yes, all user binaries and common binary should be reloaded because they use same memory spaces.

os/kernel/binary_manager/Make.defs
jeongarmy: It is needed for binary update also.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -544,27 +515,34 @@ static int reloading_thread(int argc, char *argv[])
 		bmdbg("Terminate binary %d\n", bidx);
 	}
 
-	/* 2. Unload common binary */
+	/* Unload common binary */
```
sunghan-chang, line 518: Could you let me know why common should be terminated later? If it does not need, let's merge it into the for loop.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -544,27 +515,34 @@ static int reloading_thread(int argc, char *argv[])
...
 	ret = binary_manager_terminate_binary(BM_CMNLIB_IDX);
 	if (ret != OK) {
 		bmdbg("Failed to terminate common binary\n");
 		return BINMGR_OPERATION_FAIL;
 	}
+#else
+	load_cmd = LOADCMD_LOAD;
```
sunghan-chang, line 525: What's purpose of this?

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -544,27 +515,34 @@ static int reloading_thread(int argc, char *argv[])
...
-#else
-	/* argv[1] binary index for reloading */
-	int bin_idx = (int)atoi(argv[1]);
+	/* Deinitialize modules in kernel */
+	binary_manager_deinit_modules();
 
-	return binary_manager_reload(bin_idx);
-#endif
+	/* Create a loader to reload binary */
+	ret = binary_manager_execute_loader(load_cmd, bin_idx);
```
jeongchanKim, line 539: Most of the binary manager functions return `BINMGR_XXX`.
How about changing the return of `binary_manager_execute_loader` and the return check logic?

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: A loader is created and do something according to `loading type` like LOADCMD_LOAD, LOADCMD_LOAD_ALL.
If CONFIG_SUPPORT_COMMON_BINARY is enabled, we set `LOADCMD_LOAD_ALL` to load all binaries after termination of all.
Otherwise, we set `LOADCMD_LOAD` to load a faulty binary.

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: I will modify all later.

os/kernel/binary_manager/binary_manager_load.c
jeongarmy: I will check and modify it.

sunghan-chang (approved):

PR #5319, opened 2021-11-25
os/tools/set_bininfo.py : Set the app and common binary version based…

jeongchanKim:
… on their config

app and common binary have their own version in config.
(CONFIG_APP_BIN_VER, CONFIG_COMMON_BINARY_VERSION)
When build, set their name based on their own version, not build date.


an4967 (approved):
LGTM

PR #5323, opened 2021-11-26
tools: Refactor trap tool

thapav:
Remove kernel symbols added twice.
    Remove elf file argument to ramdumpParser.py
    Add elf file extension parsing to ramdumpParser.py
    Rectify HowToUseTrap.md log formatting.

### DIFF
--- a/tools/debug/debugsymbolviewer.py
+++ b/tools/debug/debugsymbolviewer.py
@@ -21,7 +21,7 @@ from math import log
 import re
 import sys
 
-bin_path = '../../build/output/bin/'	# System map file
+bin_path = '../../build/output/bin/'	# Output bin path
 symbol_lookup_table = []		# Lookup table created from contents of system map file
 debug = False				# Variable to print debug logs
 pc_value = 0x00000000			# Program counter value
@@ -161,7 +161,7 @@ def parse_file(log_file):
 					if (is_text_address(hex(stack_addr))):
 						#If yes, print it's corresponding symbol
                                                 stack_addr = stack_addr - int(offset,16)
-						print_symbol(stack_addr)
+                                                print_symbol(stack_addr)
 
 
 #Execution starts here
--- a/tools/trap/HowToUseTrap.md
+++ b/tools/trap/HowToUseTrap.md
@@ -44,7 +44,7 @@ Hardware Configuration > Board Selection -> Reset a board on assert status autom
 Device Drivers -> Block-to-character(BCH) Support
 ```
 
-## How to upload RAMDUMP-or-UserfsDUMP
+## How to upload RAMDUMP or UserfsDUMP
 ### In Linux
 With the DUMP Tool configured above, whenever the target board crashes because of an assert condition, it enters into PANIC mode, and displays the following message:
 ```
@@ -205,7 +205,7 @@ After you see this message, you can upload the ramdump by following the step bel
 ```
 Please enter COM port number & press enter : <ComPort>
 ```
-Once connection with the COM port is established, you will get following message:
+Once connection with the COM port is established, you will get following message:  
 --> Connection established.
 
 Now you can enter desired ramdump region from the options: (Multi-heap scenario)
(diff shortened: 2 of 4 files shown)


sunghan-chang:
Could you write what you refactor in the commit description?

tools/trap/HowToUseTrap.md
```diff
@@ -303,6 +303,7 @@ PC_value	 Symbol_address	  Symbol_name	File_name
 App Crash point is as follows:
 [ Caller - return address (LR) - of the function which has caused the crash ]
 
+App name          : app2
```
abhishek-samsung, line 306: Please change the usage. Now we need to pass the kernel elf file as an argument compulsorily. Currently, we can use any elf file according to the instructions above.
Example line 283 : $ python3 ramdumpParser.py -e ../../build/output/bin/app2_dbg -t ./log_file

thapav: Done

tools/trap/HowToUseTrap.md
```diff
@@ -280,7 +280,7 @@ cd $TIZENRT_BASEDIR/tools/trap/
     $ python3 ramdumpParser.py -e `<Elf path>` -t `<Log file path>`
 
     ex)
-    $ python3 ramdumpParser.py -e ../../build/output/bin/app2_dbg -t ./log_file
+    $ python3 ramdumpParser.py -e ../../build/output/bin/tinyara -t ./log_file
```
sunghan-chang, line 283: Regardless of crash point, can we use `tinyara` only?
If it is, why do we need to pass only this?

tools/trap/HowToUseTrap.md
```diff
@@ -373,7 +375,7 @@ PC_value	 Symbol_address	  Symbol_name	File_name
 ----------------------------------------------------------
 ```
 ex)
-$python3 ramdumpParser.py -e ../../build/output/bin/common_dbg -t ./log_file
+$python3 ramdumpParser.py -e ../../build/output/bin/tinyara -t ./log_file
```
sunghan-chang, line 378: same as previous

sunghan-chang (commented):
Could you check line 47 on md file, `## How to upload RAMDUMP-or-UserfsDUMP`?
Let's remove `-`s.

tools/trap/HowToUseTrap.md
thapav: Yes only tinyara can be used.
For app crash point detection we calculate app*_dbg based on the applications loaded for the configuration. (Line 592)

sunghan-chang: Then, why do we need to pass -e option even app can be done automatically?

sunghan-chang: If we need to pass the pass of bin folder for the location of elfs, then how about removing `tinyara` like `-e ../../build/output/bin`? If we can do that, then we had better change the option from -e to -l (?).

thapav: Only apps have an elf naming convention like <app_name>_dbg
For kernel elf varies > tinyara/tinyara.axf 
Hence using this option.

sunghan-chang: Kernel also has naming convention. Please find `EXEEXT` in os/Make.defs.
naming convention of tinyara
https://github.com/Samsung/TizenRT/blob/master/os/Makefile.unix#L228

sunghan-chang: Only rtl8721csm has the EXEEXT value as axf so that tinyara.axf is used. Other boards has nothing for that. That's why tinyara is used for other boards.

sunghan-chang: Even we change not to pass kernel elf, we need to pass the location because we can't guarantee this script can be executed at the same location.
Could you let me know your opinion for giving kernel elf and for giving location by user?

thapav:
@sunghan-chang Updated the PR. Kindly check
As per our discussion,
I have removed elf argument mandate from the script. The extension is also extracted from the Make.defs.
Log file naming convention is also flexible now.

sunghan-chang (dismissed):

sunghan-chang (commented):
1. change DumpParser -> TRAP
2. please check line 272. The explanation says run the script but box has changing the directory only.

thapav:
> 

modified

tools/trap/HowToUseTrap.md
```diff
@@ -262,32 +262,36 @@ Ramdump received successfully..!
...
+### TRAP using CUI
 
 #### To get call stack using RAM dump
-1. Run Ramdump Parser Script
+1. Change the directory
 ```
 cd $TIZENRT_BASEDIR/tools/trap/
 ```
-2. See the Output
+2. Run Ramdump Parser Script and see the Output
```
sunghan-chang, line 274: There is no guide to run ramdump parser script. Please add that like 1st step.

sunghan-chang (approved):

tools/trap/HowToUseTrap.md
```diff
@@ -262,32 +262,91 @@ Ramdump received successfully..!
...
 
 #### To get call stack using RAM dump
-1. Run Ramdump Parser Script
+1. Change the directory
 ```
 cd $TIZENRT_BASEDIR/tools/trap/
 ```
-2. See the Output
+2. Copy crash logs  
+    First copy the crash logs to a file in tools/trap/`<log_file>`
```
sunghan-chang, line 275: Do we need this 2nd step even this tab is `To get call stack using RAM dump`? If it is, why do you guide two different tabs with the same content? And what can we do with ramdump? I mean why do we need ramdump?

thapav: I think we can merge the two under one menu
"To get call stack"
Ramdump is used to retrieve call stack for scenarios when crash logs are not accessible and for the gui_ramdumpParser.py script.

tools/debug/debugsymbolviewer.py
```diff
@@ -161,7 +161,7 @@ def parse_file(log_file):
 					if (is_text_address(hex(stack_addr))):
 						#If yes, print it's corresponding symbol
                                                 stack_addr = stack_addr - int(offset,16)
-						print_symbol(stack_addr)
+                                                print_symbol(stack_addr)
```
sunghan-chang, line 164: Except those two lines, you used tabs for indentation. I don't know this change is necessary.

tools/trap/HowToUseTrap.md
sunghan-chang: Are you still modifying? There is no change on this.

PR #5324, opened 2021-11-29
kernel/binary_manager: Return all results for requested binaries from binary_manager_set_bootparam

jeongarmy:
The binary_manager_set_bootparam is called with binary group.
If one of them has no update, binary manager doesn't update bootparam and returns error.
For this case, return all results for requested group's binaries to know the reason of failure of bootparam update.

### DIFF
--- a/framework/src/binary_manager/binary_manager_update.c
+++ b/framework/src/binary_manager/binary_manager_update.c
@@ -41,9 +41,12 @@ binmgr_result_type_e binary_manager_set_bootparam(uint8_t type, binary_setbp_res
 		return BINMGR_INVALID_PARAM;
 	}
 
-	if (!BM_CHECK_GROUP(type, BINARY_KERNEL) && !BM_CHECK_GROUP(type, BINARY_USERAPP)
+	if (!BM_CHECK_GROUP(type, BINARY_KERNEL)
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	&& !BM_CHECK_GROUP(type, BINARY_USERAPP)
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	&& !BM_CHECK_GROUP(type, BINARY_COMMON)
+#endif
 #endif
 	) {
 		bmdbg("Invalid parameter %u\n", type);
--- a/framework/src/binary_manager/binary_manager_update.c
+++ b/framework/src/binary_manager/binary_manager_update.c
@@ -30,11 +30,16 @@
 #include <binary_manager/binary_manager.h>
 #include "binary_manager_internal.h"
 
-binmgr_result_type_e binary_manager_set_bootparam(uint8_t type)
+binmgr_result_type_e binary_manager_set_bootparam(uint8_t type, binary_setbp_result_t *update_result)
 {
 	binmgr_result_type_e ret;
 	binmgr_request_t request_msg;
 	binmgr_setbp_response_t response_msg;
+
+	if (update_result == NULL) {
+		bmdbg("update_result is NULL\n");
+		return BINMGR_INVALID_PARAM;
+	}
 
 	if (!BM_CHECK_GROUP(type, BINARY_KERNEL) && !BM_CHECK_GROUP(type, BINARY_USERAPP)
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
@@ -58,6 +63,15 @@ binmgr_result_type_e binary_manager_set_bootparam(uint8_t type)
 	ret = binary_manager_receive_response(&response_msg, sizeof(binmgr_setbp_response_t));
 	if (ret != BINMGR_OK) {
 		return ret;
+	}
+
+	if (response_msg.result != BINMGR_OK) {
+		bmdbg("Binary manager setbp FAIL %d\n", response_msg.result);
+		if (response_msg.result == BINMGR_ALREADY_UPDATED) {
+			/* Copy result data */
+			memset(update_result, 0, sizeof(binary_setbp_result_t));
+			memcpy(update_result, &response_msg.data, sizeof(binary_setbp_result_t));
+		}
 	}
 
 	return response_msg.result;
--- a/os/include/tinyara/binary_manager.h
+++ b/os/include/tinyara/binary_manager.h
@@ -67,12 +67,12 @@
 /* The lenght of user or kernel partition path */
 #define BINARY_PATH_LEN                  16
 
-/* Binary Type */
+/* Binary Type : Kernel, Common, User app */
 enum binary_type_e {
-	BINARY_KERNEL = 1,
-	BINARY_COMMON = 2,
-	BINARY_USERAPP = 3,
-	BINARY_TYPE_MAX,
+	BINARY_KERNEL = 0,
+	BINARY_COMMON = 1,
+	BINARY_USERAPP = 2,
+	BINARY_TYPE_MAX = 3,
 };
 
 /* Macros for binary grouping used for request to set bootparam */
@@ -165,6 +165,12 @@ struct common_binary_header_s {
 typedef struct common_binary_header_s common_binary_header_t;
 
 /* The structure of binary update information for kernel or user binaries */
+struct binary_setbp_result_s {
+	int result[BINARY_TYPE_MAX];
+};
+typedef struct binary_setbp_result_s binary_setbp_result_t;
+
+/* The structure of binary update information for kernel or user binaries */
 struct binary_update_info_s {
 	int available_size;
 	char name[BIN_NAME_MAX];
@@ -223,6 +229,7 @@ typedef struct binmgr_request_s binmgr_request_t;
 
 struct binmgr_setbp_response_s {
 	binmgr_result_type_e result;
+	binary_setbp_result_t data;
 };
 typedef struct binmgr_setbp_response_s binmgr_setbp_response_t;
 
--- a/os/kernel/Kconfig
+++ b/os/kernel/Kconfig
@@ -966,7 +966,6 @@ config IDLETHREAD_STACKSIZE
 config USERMAIN_STACKSIZE
 	int "Main thread stack size"
 	default 2048
-	depends on !BINARY_MANAGER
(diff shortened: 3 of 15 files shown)


os/include/tinyara/binary_manager.h
```diff
@@ -67,12 +67,12 @@
...
-	BINARY_KERNEL = 1,
-	BINARY_COMMON = 2,
-	BINARY_USERAPP = 3,
-	BINARY_TYPE_MAX,
+/* Binary Group Type : Kernel, Common, User app */
+enum binary_group_type_e {
+	BINARY_KERNEL = 0,
+	BINARY_COMMON = 1,
+	BINARY_USERAPP = 2,
+	BINARY_TYPE_COUNT = 3,
```
sunghan-chang, line 75: Previous naming looks better for me.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -268,22 +269,28 @@ void binary_manager_update_bootparam(int requester_pid, uint8_t type)
 		return;
 	}
 
+	response_msg.result = BINMGR_OK;
+	is_all_updatable = true;
+
 	/* Get current bootparam data and update version */
 	memcpy(&update_bp_data, binary_manager_get_bpdata(), sizeof(binmgr_bpdata_t));
 	update_bp_data.version++;
```
sunghan-chang, line 277: Regardless of this commit, bp version should be the last of update sequence. If not, when power off is happened at middle of the update, bootloader can check wrong bp first.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -294,42 +301,55 @@ void binary_manager_update_bootparam(int requester_pid, uint8_t type)
...
+				goto send_response;
 			}
 		}
 		if (!need_update) {
-			bmdbg("No binary to update\n");
-			goto send_response;
+			bmdbg("No App binaries to update\n");
+			is_all_updatable = false;
+			response_msg.data.result[BINARY_USERAPP] = BINMGR_ALREADY_UPDATED;
 		}
```
sunghan-chang, line 319: Why do you need this section separately? For kernel and common, you added them in `else  if`.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -60,6 +60,16 @@ binmgr_result_type_e binary_manager_set_bootparam(uint8_t type)
 		return ret;
 	}
 
+	if (response_msg.result != BINMGR_OK) {
+		bmdbg("Binary manager setbp FAIL %d\n", response_msg.result);
+		if (response_msg.result == BINMGR_ALREADY_UPDATED) {
+			/* Copy result data */
+			bmdbg("Binary manager setbp FAIL %d\n", response_msg.result);
+			memset(update_result, 0, sizeof(binary_setbp_result_t));
```
jeongchanKim, line 68: Is it okay not to check that `update_result` is NULL or not?'

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -268,22 +269,28 @@ void binary_manager_update_bootparam(int requester_pid, uint8_t type)
...
-			goto send_response;
-		} else if (!check_update) {
+		if (ret == BINMGR_OK) {
+			/* Update index for inactive partition */
+			update_bp_data.active_idx ^= 1;
+		} else if (ret == BINMGR_ALREADY_UPDATED || ret == BINMGR_NOT_FOUND) {
 			bmdbg("No binary to update\n");
+			is_all_updatable = false;
+			response_msg.data.result[BINARY_KERNEL] = BINMGR_ALREADY_UPDATED;
+
```
jeongchanKim, line 289: Please remove this empty line.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -60,6 +60,16 @@ binmgr_result_type_e binary_manager_set_bootparam(uint8_t type)
 		return ret;
 	}
 
+	if (response_msg.result != BINMGR_OK) {
+		bmdbg("Binary manager setbp FAIL %d\n", response_msg.result);
+		if (response_msg.result == BINMGR_ALREADY_UPDATED) {
+			/* Copy result data */
+			bmdbg("Binary manager setbp FAIL %d\n", response_msg.result);
```
jeongchanKim, line 67: This log is duplicated with line 64.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -268,22 +269,28 @@ void binary_manager_update_bootparam(int requester_pid, uint8_t type)
...
+	response_msg.result = BINMGR_OK;
+	is_all_updatable = true;
+
 	/* Get current bootparam data and update version */
 	memcpy(&update_bp_data, binary_manager_get_bpdata(), sizeof(binmgr_bpdata_t));
 	update_bp_data.version++;
 
 	if (BM_CHECK_GROUP(type, BINARY_KERNEL)) {
 		/* Update bootparam and Reboot if new kernel binary exists */
 		ret = binary_manager_check_kernel_update(&check_update);
```
jeongchanKim, line 281: In `os/kernel/binary_manager/binary_manager_data.c`,
there is no argument in the prototype of `binary_manager_check_kernel_update`.
Please remove the param if not needed.

framework/src/binary_manager/binary_manager_update.c
```diff
@@ -30,7 +30,7 @@
 #include <binary_manager/binary_manager.h>
 #include "binary_manager_internal.h"
 
-binmgr_result_type_e binary_manager_set_bootparam(uint8_t type)
+binmgr_result_type_e binary_manager_set_bootparam(uint8_t type, binary_setbp_result_t *update_result)
```
Taejun-Kwon, line 33: There are 2 options to make api of f/w clearly.
1. you can use binary_setbp_result_t as a type of function instead of binmgr_result_type_e.
2. Refine new struct as a result type which include both of result value.
Checking both of return value & param is not good options as a api of framework. simple is best.

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -294,42 +301,55 @@ void binary_manager_update_bootparam(int requester_pid, uint8_t type)
...
-		ret = binary_manager_check_user_update(BM_CMNLIB_IDX, &check_update);
-		if (ret != BINMGR_OK) {
-			bmdbg("Failed to check kernel update, %d\n", ret);
-			goto send_response;
-		} else if (!check_update) {
+		ret = binary_manager_check_user_update(BM_CMNLIB_IDX);
+		if (ret == BINMGR_OK) {
+			/* Update index for inactive partition */
+			update_bp_data.app_data[BIN_BPIDX(BM_CMNLIB_IDX)].useidx ^= 1;
+		} else if (ret == BINMGR_ALREADY_UPDATED || ret == BINMGR_NOT_FOUND) {
```
Taejun-Kwon, line 328: How about use something like a 'BINMGR_NO_AVAILABLE_UPDATE' instead of these 2 values?

os/kernel/binary_manager/binary_manager_bootparam.c
jeongarmy: That's because there may be multiple user binaries.

os/kernel/binary_manager/binary_manager_bootparam.c
jeongarmy: I separated the error codes to distinguish the cause of failure for user.

os/include/tinyara/binary_manager.h
jeongarmy: I reverted this change

framework/src/binary_manager/binary_manager_update.c
jeongarmy: I agree with your opinion. But if it returns arrays(binary_setbp_result_t), we need to allocate memory and return it then user should free the result. So I implemented with [out] parameter.
I will consider more later.

os/kernel/binary_manager/binary_manager_bootparam.c
jeongarmy: This value is a local variable and not written in flash until bootparam checking is done.
So I think this is not problem.

os/include/tinyara/binary_manager.h
```diff
@@ -67,12 +67,12 @@
...
-	BINARY_KERNEL = 1,
-	BINARY_COMMON = 2,
-	BINARY_USERAPP = 3,
-	BINARY_TYPE_MAX,
+/* Binary Group Type : Kernel, Common, User app */
+enum binary_group_type_e {
+	BINARY_KERNEL = 0,
+	BINARY_COMMON = 1,
+	BINARY_USERAPP = 2,
+	BINARY_TYPE_MAX = 3,
```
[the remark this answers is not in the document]
jeongarmy: 1. I adjust the values to user these for return type of binary_manager_set_bootparam, `binary_setbp_result_t`.
This function requires three result values because of the number of type is 3 (kernel, common, userapp).
So I adjust these values.

2. 'binmgr_result_type_e binary_manager_set_bootparam(uint8_t type, binary_setbp_result_t *update_result)`
`type` is parameter value user set the binaries that are need to update together.
`binary_group_type_e` is used for setting of binary type with `BM_XXXX_GROUP`.

os/kernel/Kconfig
```diff
@@ -966,7 +966,6 @@ config IDLETHREAD_STACKSIZE
 config USERMAIN_STACKSIZE
 	int "Main thread stack size"
 	default 2048
-	depends on !BINARY_MANAGER
```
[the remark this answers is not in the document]
jeongarmy: With binary manager support on binary separation, We don't need to execute user app because binary manager loads user binary application.
But now binary manager works in flat build also, user app should in flat build.
Please refer to line 318 in os_bringup.c.

sunghan-chang (approved):

PR #5327, opened 2021-11-29
arch/arm/src/armv7-m,8-m/up_assert.c : Add log for kernel heap corrup…

jeongchanKim:
…tion checking

To clarify, add log for kernel heap corruption checking result.


Taejun-Kwon (approved):

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -502,7 +502,7 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	lldbg("Checking kernel heap for corruption...\n");
 	if (mm_check_heap_corruption(g_kmmheap) == OK) {
-		lldbg("No heap corruption detected\n");
+		lldbg("No kernel heap corruption detected\n");
```
sunghan-chang, line 505: then how about app? line 518?

sunghan-chang (approved):

PR #5329, opened 2021-11-29
kernel/binary_manager: Add tcb to a binary thread list after task activation in pthread_create

jeongarmy:
If a caller thread is blocked by pthread_sem_take, binary_manager_remove_binlist is called first by executing the new created thread before adding the tcb to a list by binary_manager_add_binlist.
So it should be added to a binary thread list after task_activate directly before pthread_sem_take.

### DIFF
--- a/os/kernel/pthread/pthread_create.c
+++ b/os/kernel/pthread/pthread_create.c
@@ -451,6 +451,11 @@ int pthread_create(FAR pthread_t *thread, FAR const pthread_attr_t *attr, pthrea
 
 	sched_lock();
 	if (ret == OK) {
+#if defined(CONFIG_BINARY_MANAGER) && defined(CONFIG_APP_BINARY_SEPARATION)
+		/* Add tcb to binary thread list */
+
+		binary_manager_add_binlist(&ptcb->cmn);
+#endif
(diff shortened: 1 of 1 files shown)


os/kernel/pthread/pthread_create.c
```diff
@@ -455,6 +455,11 @@ int pthread_create(FAR pthread_t *thread, FAR const pthread_attr_t *attr, pthrea
 	}
 
 	if (ret == OK) {
+#if defined(CONFIG_BINARY_MANAGER) && defined(CONFIG_APP_BINARY_SEPARATION)
+		/* Add tcb to binary thread list */
+
+		binary_manager_add_binlist(&ptcb->cmn);
```
jeongchanKim, line 461: In task_create, add list is called before `task_activate`.
What is the difference?
Is it possible to move this before `task_activate` or move from `task_create`?

jeongarmy: modified

sunghan-chang (approved):

sunghan-chang:
after -> before in commit title

jeongchanKim (approved):
LGTM

sunghan-chang (approved):

PR #5331, opened 2021-11-30
fixup! mm/*mm_heap : Return NULL if requested size is 0

sunghan-chang:
When realloc is executed with size 0, it should try to free previous
allocated memory. But commit fad7c48 returns NULL without freeing.
This commit fixes wrong modification of commit fad7c48.


jsdosa (approved):
looks nice to me.

PR #5332, opened 2021-11-30
net/netmgr: add wifi deinit in kernel space

pillip8282:
add wifi deinit in kernel space for binary manager to deinit wifi

### DIFF
--- a/os/net/netmgr/netdev_wifi.c
+++ b/os/net/netmgr/netdev_wifi.c
@@ -145,6 +145,20 @@ int _trwifi_handle_command(struct netdev *dev, lwnl_req cmd)
 	return 0;
 }
 
+/* it's for reloading operation in binary manager.
+ * it must be used to binary manager only.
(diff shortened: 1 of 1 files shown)


os/net/netmgr/netdev_wifi.c
```diff
@@ -145,6 +145,20 @@ int _trwifi_handle_command(struct netdev *dev, lwnl_req cmd)
 	return 0;
 }
 
+/* it's for reloading operation in binary manager.
+ * it must be used to binary manager only.
+ */
+trwifi_result_e netdev_deinit_wifi(void)
```
an4967, line 151: Is it okay not to define this in header file?

pillip8282: It's function only for binary manager. So it's better not to define it to a header file.

an4967 (approved):
LGTM

PR #5336, opened 2021-12-01
kernel/binary_manager: Add only user threads to a thread list of binary

jeongarmy:
Each binary's thread list is used for  binary reloading.
When binary reloading, binary manager terminates all user threads and loads user binary again.
So add only user threads to a thread list of binary in task_create and pthread_create.


jeongarmy:
Close this PR because binary manager should terminate all threads created by user binary as we discussed.

PR #5105, opened 2021-08-05
[Under Powercut Testing] os/fs/smartfs : Add buffering of data to be written to smartfs file

amogh-samsung:
[Improve SmartFs Overwrite Method] These changes reduce the overhead in overwriting consecutive chunks of data of a file.
The case of appending data to the file is not affected as it is handled by a separate part of code, the modified code only deals with overwriting to the files.

- Hold the buffer from being committed to flash if the buffer is not completely full.
- The buffer will get synced when the file is closed or if another write request is
  issued which fills up the left over space in the buffer.

Signed-off-by: Taejun Kwon <tj80.kwon@samsung.com>, Amogh Hassija <a.hassija@samsung.com>

PR #5338, opened 2021-12-01
os/fs/smartfs: Fix chaining issue in jSmartFs overwrite

amogh-samsung:
- The last sector chained to the file may be overwritten during sync at close time.
- In some cases, the proper sector header may not be present in sf->buffer for sync.
- Hence, read the next sector always after filling up current sector to maintain
  proper contents in buffer.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -686,11 +686,18 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 				smartfs_setbuffer(&readwrite, sf->currsector, sf->curroffset,\
 					 fs->fs_llformat.availbytes - sf->curroffset, (uint8_t *)&buffer[byteswritten]);
 #endif
+				/* Only the data bytes of the sector are written,
+				 * we do not re-write/overwrite the header bytes.
+				 */
 				ret = FS_IOCTL(fs, BIOC_WRITESECT, (unsigned long)&readwrite);
 				if (ret < 0) {
 					fdbg("Error writing sector %d data, ret : %d\n", readwrite.logsector, ret);
 					goto errout_with_semaphore;
 				}
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+				/* Clear the flags, no longer dirty if the buffer content is written to flash */
+				sf->bflags = SMARTFS_BFLAG_UNMOD;
+#endif
 			}
 
 			/* Update our control variables */
@@ -712,22 +719,26 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t 
 				goto errout_with_semaphore;
 			}
 
-			/* If file is modified and more data remains to be appended to file, but no next sector is available,
-			 * do not update sf->currsector and curroffset. Will be handled when buffer data is synced.
+			/* If file is modified and more data remains to be appended to file,
+			 * but no next sector is available, do not update sf->currsector and curroffset.
+			 * This will be handled when buffer data is synced.
 			 */
 			if (SMARTFS_NEXTSECTOR(header) != 0xFFFF) {
-				sf->currsector = SMARTFS_NEXTSECTOR(header);
-				
 				/* Now get the chained sector info and reset the offset */
+				sf->currsector = SMARTFS_NEXTSECTOR(header);
 				sf->curroffset = size;
+
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
-				/* Read the next sector if more bytes remain in the write buffer */
+				/* Read the next sector's data if we have reached the end of the current sector
+				 * AND more data is left to be written.
+				 * The header of the current sector will be read later if required.
+				 */
 				if (buflen > 0) {
-					smartfs_setbuffer(&readwrite, sf->currsector, 0,\
-						 fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
+					smartfs_setbuffer(&readwrite, sf->currsector, size,\
+						 fs->fs_llformat.availbytes - size, (uint8_t *)&sf->buffer[size]);
 					ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
 					if (ret < 0) {
-						fdbg("Error %d reading sector %d header\n", ret, sf->currsector);
+						fdbg("Error %d reading sector %d\n", ret, sf->currsector);
 						goto errout_with_semaphore;
 					}
 				}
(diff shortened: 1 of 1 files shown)


davidfather:
Is this modification backward compatible? I mean, is updating user appliances with this modification possible or not?

amogh-samsung:
> Is this modification backward compatible? I mean, is updating user appliances with this modification possible or not?

This modification is coupled with #5105 
Both applied together would be backward compatible.
Using this modification in appliance code should not cause any problems

Taejun-Kwon:
> * The last sector chained to the file may be overwritten during sync at close time.
> * In some cases, the proper sector header may not be present in sf->buffer for sync.
> * Hence, read the next sector always after filling up current sector to maintain
>   proper contents in buffer.
> 
> Signed-off-by: Amogh Hassija [a.hassija@samsung.com](mailto:a.hassija@samsung.com)

Please verify 'some cases' first, and then I will check again.

amogh-samsung:
> > * The last sector chained to the file may be overwritten during sync at close time.
> > * In some cases, the proper sector header may not be present in sf->buffer for sync.
> > * Hence, read the next sector always after filling up current sector to maintain
> >   proper contents in buffer.
> > 
> > Signed-off-by: Amogh Hassija [a.hassija@samsung.com](mailto:a.hassija@samsung.com)
> 
> Please verify 'some cases' first, and then I will check again.

We encountered the bug, reproduced it, checked and then fixed.
It happens when we use the optimal buffer size for the smartfs sector size.

```
while (file_size >= buf_size) {
    write(buf_size);
    file_size -= buf_size;
}
if (file_size != 0) {
    write(file_size);
}
```

sf->buffer is only read when we open the file.
After that when we over write any sector completely, sf->buffer is not populated with next sector data.
So, when file is closed and final remaining bytes go for sync, sf->buffer does not have correct header data and chain gets corrupted.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -712,24 +716,26 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
 			if (SMARTFS_NEXTSECTOR(header) != 0xFFFF) {
-				sf->currsector = SMARTFS_NEXTSECTOR(header);
-				
 				/* Now get the chained sector info and reset the offset */
+				sf->currsector = SMARTFS_NEXTSECTOR(header);
 				sf->curroffset = size;
+
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
-				/* Read the next sector if more bytes remain in the write buffer */
-				if (buflen > 0) {
```
amogh-samsung, line 725: @Taejun-Kwon 
buflen passed to smartfs_write = fs->fs_llformat.availbytes
Hence, every time we write optimal buffer size, after that buflen = 0
So this part of code does not get executed and next sector is not read

It is fine when we are not waiting for smartfs_sync_internal, because Line 683-684 above do not write header
They just write file data

But when sync_internal writes last part of data, it needs the header which is not there in sf->buffer

os/fs/smartfs/smartfs_smart.c
```diff
@@ -712,24 +716,26 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
-						fdbg("Error %d reading sector %d header\n", ret, sf->currsector);
-						goto errout_with_semaphore;
-					}
+				/* Read the next sector if we have reached the end of the current sector.
+				 * Future file ops need this data to be present since buffer updation does not happen
+				 * automatically/otherwise
+				 */
+				smartfs_setbuffer(&readwrite, sf->currsector, 0,\
+					 fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
+				ret = FS_IOCTL(fs, BIOC_READSECT, (unsigned long)&readwrite);
```
sangam-swami, line 735: will it not be an overhead in case no next sector available case?

os/fs/smartfs/smartfs_smart.c
Taejun-Kwon: I think rollback #5105 is better option, we don't have to read whole of sector if buflen from high layer is not exist. only header is necessary

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Updated with an alternate solution to avoid overhead.
Please review.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: Updated with an alternate solution, please take a look.

davidfather:
@amogh-samsung Let's do some aging and stress tests along with power-cut tests.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -736,6 +747,20 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
 		}
 	}
 
+#ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
+	/* If there is unsynced data in the sf->buffer and the control is now leaving the
+	 * scope of this function, the correct header should also be present in sf->buffer
+	 * for smartfs_append_data or smartfs_sync_internal. Hence we read the header now.
+	 */
+	if (sf->bflags & SMARTFS_BFLAG_DIRTY && byteswritten > 0) {
```
Taejun-Kwon, line 755: Instead of this option, I think we can use sf->curroffset simply

os/fs/smartfs/smartfs_smart.c
```diff
@@ -712,22 +719,26 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
 #ifdef CONFIG_SMARTFS_USE_SECTOR_BUFFER
-				/* Read the next sector if more bytes remain in the write buffer */
+				/* Read the next sector's data if we have reached the end of the current sector
+				 * AND more data is left to be written.
+				 * The header of the current sector will be read later if required.
+				 */
 				if (buflen > 0) {
-					smartfs_setbuffer(&readwrite, sf->currsector, 0,\
-						 fs->fs_llformat.availbytes, (uint8_t *)sf->buffer);
+					smartfs_setbuffer(&readwrite, sf->currsector, size,\
```
Taejun-Kwon, line 737: you can improve here.
Let's think about buflen is greater than fs->fs_llformat.availbytes - size or smaller than it.

os/fs/smartfs/smartfs_smart.c
amogh-samsung: You mean
` if (sf->curroffset > size)`
?

Taejun-Kwon: yes

Taejun-Kwon (approved):

PR #5339, opened 2021-12-01
arch/armv7-m, armv8-m: Print stack contents in out of range stack poi…

sangam-swami:
…nter case

Stack contents are not printed if sp is not within allocated stack memory range.
Printing process stack and current stack pointer helps in debugging.

Signed-off-by: sangamanatha <sangam.swami@samsung.com>

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -364,6 +364,11 @@ static void up_dumpstate(void)
 
 	if (sp > ustackbase || sp <= ustackbase - ustacksize) {
 		lldbg("ERROR: Stack pointer is not within the allocated stack\n");
+		lldbg("Proper task stack dump:\n");
+		up_stackdump(ustackbase - ustacksize + 1, ustackbase);
+		lldbg("Wrong Stack pointer %08x: %08x %08x %08x %08x %08x %08x %08x %08x\n",
+		sp, *((uint32_t *)sp + 0), *((uint32_t *)sp + 1), *((uint32_t *)sp + 2), ((uint32_t *)sp + 3),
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -364,6 +364,10 @@ static void up_dumpstate(void)
 
 	if (sp > ustackbase || sp <= ustackbase - ustacksize) {
 		lldbg("ERROR: Stack pointer is not within the allocated stack\n");
+		up_stackdump(ustackbase - ustacksize + 1, ustackbase);
```
sunghan-chang, line 367: How about adding the log "Original? Proper Stack Dump" to say what they are clearly?

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -364,6 +364,10 @@ static void up_dumpstate(void)
...
+		uint32_t *ptr = (uint32_t *)sp;
```
jeongchanKim, line 368: Should we use this variable? I think we can use `sp` directly.

os/arch/arm/src/armv7-m/up_assert.c
sangam-swami: ok

davidfather (approved):
lgtm.

deepaksrma (commented):
LGTM

sunghan-chang (approved):

sunghan-chang:
@sangam-swami @thapav Could you check TRAP with this case?

thapav:
> @sangam-swami @thapav Could you check TRAP with this case?

Please find trap tool update for this PR below:
https://github.com/Samsung/TizenRT/pull/5352

PR #5344, opened 2021-12-07

ziyik:
Change time slot for BT Coex Case "AP + BT Scan" to 40,10,40,10 slot time


an4967:
This will be used after test previous libs

PR #5345, opened 2021-12-07

ziyik:
Change time slot for BT Coex Case "AP + BT Scan" to 38,12,38,12 slot time


jeongchanKim:
There are same PR(#5344). Please check it.

ziyik:
> There are same PR(#5344). Please check it.

The library is different, in binary form.
The time slot is different for these PR.

an4967:
This will be used after test previous libs

PR #5337, opened 2021-12-01
os/board/rtl8721csm: Tune BT Coex Case for AP + BT Scan

ziyik:
Change time slot for BT Coex Case "AP + BT Scan" to 37,13,37,13 slot time

PR #5346, opened 2021-12-08

ziyik:
Change time slot for BT Coex Case "AP + BT Scan" to 39,11,39,11 slot time


pillip8282:
If it's duplicated PR to https://github.com/Samsung/TizenRT/pull/5337 please remove it.

PR #5347, opened 2021-12-08
rtl8721csm: fix heap curruption when scan result is 0

pillip8282:
rtl8721 calls wifi_scan_networks_with_ssid() when scan with ssid was
requested. wifi_scan_networks_with_ssid() calls result_handler even though
scan_cnt which is result of wifi_scan is 0. In result_handler it
convert scan_buf which has invalid data to trwifi_scan_result. So it
does unexpected behavior.

### DIFF
--- a/os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
+++ b/os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
@@ -1936,25 +1936,24 @@ int wifi_scan_networks_with_ssid(int (results_handler)(char*buf, int buflen, cha
 	int ret;
 
 	scan_buf.buf_len = scan_buflen;
-	scan_buf.buf = (char*)rtw_malloc(scan_buf.buf_len);
-	if(!scan_buf.buf){
+	scan_buf.buf = (char *)rtw_malloc(scan_buf.buf_len);
+	if (!scan_buf.buf) {
 		RTW_API_INFO("\n\rERROR: Can't malloc memory(%d)", scan_buf.buf_len);
 		return RTW_NOMEM;
 	}
 	//set ssid
 	memset(scan_buf.buf, 0, scan_buf.buf_len);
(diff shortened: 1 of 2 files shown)


jeongchanKim (approved):
LGTM

os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
```diff
@@ -1936,25 +1936,24 @@ int wifi_scan_networks_with_ssid(int (results_handler)(char*buf, int buflen, cha
...
 	//set ssid
 	memset(scan_buf.buf, 0, scan_buf.buf_len);
 	memcpy(scan_buf.buf, &ssid_len, sizeof(int));
-	memcpy(scan_buf.buf+sizeof(int), ssid, ssid_len);
+	memcpy(scan_buf.buf + sizeof(int), ssid, ssid_len);
 
-	//Scan channel	
-	if((scan_cnt = wifi_scan(RTW_SCAN_TYPE_ACTIVE, RTW_BSS_TYPE_ANY, &scan_buf)) < 0){
+	//Scan channel
+	if ((scan_cnt = wifi_scan(RTW_SCAN_TYPE_ACTIVE, RTW_BSS_TYPE_ANY, &scan_buf)) <= 0) {
```
an4967, line 1950: I think that it is not fail case when the 'scan_cnt' is zero.
'scan_cnt==0' means that scan process was success but the result is empty.
How do you think about this?

pillip8282: From the user's point of view, it may be considered to have failed if the scan result is 0. So It's ok to return fail in this case.

an4967 (approved):
LGTM

os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
an4967: Okay, I see

PR #5351, opened 2021-12-10
kernel/mqueue : Set errno when fail to get/allocate the msg

jeongchanKim:
If fail to get the preallocated msg from interrupt handler, set errno as EBUSY.
If fail to allocat the msg from normal thread, set errno as ENOMEM.

### DIFF
--- a/os/kernel/mqueue/mq_sndinternal.c
+++ b/os/kernel/mqueue/mq_sndinternal.c
@@ -171,8 +171,10 @@ int mq_verifysend(mqd_t mqdes, FAR const char *msg, size_t msglen, int prio)
  *   None
  *
  * Return Value:
- *   A reference to the allocated msg structure.  On a failure to allocate,
- *   this function PANICs.
+ *   A reference to the allocated msg structure.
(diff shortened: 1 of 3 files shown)


os/kernel/mqueue/mq_sndinternal.c
```diff
@@ -214,9 +217,11 @@ FAR struct mqueue_msg_s *mq_msgalloc(void)
 			mqmsg = (FAR struct mqueue_msg_s *)kmm_malloc((sizeof(struct mqueue_msg_s)));
 
 			/* Check if we got an allocated message */
-
-			ASSERT(mqmsg);
```
sunghan-chang, line 218: Could you check line 175? Description of this function says assert on failure. Could you modify that according to our new policy?

jeongchanKim: Modified. Thanks.

sunghan-chang (dismissed):

sunghan-chang (approved):

PR #5353, opened 2021-12-10
Fix crash caused by procfs related 'cat' commands

amogh-samsung:
- 'cat /proc/partitions' command causes a device crash since the buffer
 supplied for reading is exceeded by the number of returned data bytes.
- Proper checks are missing from the related read functions and current
 buffer size set is too small and obsolete.
- Do not use obsolete streaming I/O buffer size and instead use larger
 buffer to avoid crash, operations related code will be checked in future.


amogh-samsung:
> * 'cat /proc/partitions' command causes a device crash since the buffer
>   supplied for reading is exceeded by the number of returned data bytes.
> * Proper checks are missing from the related read functions and current
>   buffer size set is too small and obsolete.
> * Do not use obsolete streaming I/O buffer size and instead use larger
>   buffer to avoid crash, operations related code will be checked in future.
> 
> Signed-off-by: Amogh Hassija [a.hassija@samsung.com](mailto:a.hassija@samsung.com)

Crash Example: -
```
TASH>>
TASH>>cat proc/partitions
part_procfs_read: Entry
part_procfs_read: buffer=0x2031a88 buflen=31
part_procfs_read: Start Reading Now!!!

part_procfs_read: File position = 0
part_procfs_read: Added 44 characters to the buffer

part_procfs_read: Read 44 bytes, buflen = 31, available = -13

part_procfs_read: Reading the partitions
part_procfs_read: Added 18 bytes to the buffer, total = 62
part_procfs_read: Adding nam of the MTD partition to the buffer
part_procfs_read: Now added 18 bytes to the buffer, total = 62
part_procfs_read: Returning 44

part_procfs_read: Entry
part_procfs_read: buffer=0x2031a88 buflen=31
part_procfs_read: Start Reading Now!!!

part_procfs_read: Read 0 bytes, buflen = 31, available = 31

part_procfs_read: Reading the partitions
part_procfs_read: Added 18 bytes to the buffer, total = 18
part_procfs_read: Adding nam of the MTD partition to the buffer
part_procfs_read: Now added 29 bytes to the buffer, total = 29
part_procfs_read: Returning 29

Name             Start    Sizebl1      0       4   progmem
part_procfs_read: Entry
part_procfs_read: buffer=0x2031a88 buflen=31
part_procfs_read: Start Reading Now!!!

part_procfs_read: Read 0 bytes, buflen = 31, available = 31

part_procfs_read: Reading arm_undefinedinsn:
Undefined instruction at 0x4003424
up_assert: Assertion failed at file:armv7-r/arm_undefinedinsn.c line: 107 task:h
```

amogh-samsung:
@Taejun-Kwon please let me know your opinion on the default buffer size.
The code, I believe, can possibly produce a crash till the time watertight checks are put into place.

apps/system/utils/Kconfig
```diff
@@ -23,7 +23,7 @@ config FS_CMDS
 if FS_CMDS
 config FSCMD_BUFFER_LEN
        int "Buffer length of FSCMD"
-       default 32
+       default 256
```
jeongchanKim, line 26: Could you change the defconfigs also?

os/fs/driver/mtd/mtd_partition.c
```diff
@@ -215,6 +215,10 @@ static bool part_bytecheck(FAR struct mtd_partition_s *priv, off_t byoff)
  * Private Functions
  ****************************************************************************/
 
+//TODO: Check mtd part functions for memory leaks such as unchecked buffer overflows, etc.
```
jeongchanKim, line 218: Could you separate this as different commit?

apps/system/utils/Kconfig
jeongchanKim: This config is used with local variable, especially `tash_cat`.
How it makes overflow?
`tash_cat` reads and writes based on this value.

amogh-samsung: The old small buffer size causes data to overflow and since there aren't many proper checks in place, the overflow causes a crash.

jeongchanKim: Yes, I know it is from overflow.
What I want to check, we should fix the point which makes overflow.
This is just increasing the buffer size, then it can be overflowed after some time.

amogh-samsung: Yes, this is very much a temporary fix as mentioned.
Proper fixing of multiple ProcFs operations might be required which requires a pre-requisite of checking them as well.
We will be taking that up based on the need of those procfs features and the severity of the issues, just a bit later.

os/fs/driver/mtd/mtd_partition.c
amogh-samsung: I understand and agree with your logic regarding separation of the commits.
However, this PR deals just with the avoiding of the crash issue instead of fixing it completely. Future changes might lead to this PR being reverted due to logical changes.
Hence I would like to keep these changes together since only the "TODO:" statement offers an explanation for the changes in the buffer size.

Taejun-Kwon: I agree with jeongchan's opinion. please separate commit.

apps/system/utils/Kconfig
sunghan-chang: In my opinion, this fix is not urgent so that we could make a change with root cause fix.

amogh-samsung: > Could you change the defconfigs also?

Done

os/fs/driver/mtd/mtd_partition.c
amogh-samsung: Done

Taejun-Kwon (approved):

PR #5354, opened 2021-12-13
mbedtls/alt: fix cfb128 test cases

pillip8282:
cfb128 only requires encryption key. So it doesn't need to use decrypt key.

### DIFF
--- a/external/mbedtls/alt/aes_alt.c
+++ b/external/mbedtls/alt/aes_alt.c
@@ -204,9 +204,8 @@ int mbedtls_aes_crypt_cfb128(mbedtls_aes_context *ctx,
 	if (mode == MBEDTLS_AES_ENCRYPT) {
 		ret = sl_aes_encrypt(ctx->shnd, &aes_in, &aes_param, ctx->enc_key_idx, &aes_out);
 	} else if (mode == MBEDTLS_AES_DECRYPT) {
-		ret = sl_aes_decrypt(ctx->shnd, &aes_in, &aes_param, ctx->dec_key_idx, &aes_out);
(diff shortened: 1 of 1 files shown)


jeongchanKim (approved):
LGTM

external/mbedtls/alt/aes_alt.c
```diff
@@ -201,12 +201,7 @@ int mbedtls_aes_crypt_cfb128(mbedtls_aes_context *ctx,
 	hal_aes_param aes_param = {HAL_AES_CFB128, iv, 16, iv_off, NULL, NULL, NULL};
 	int ret = 0;
 
-	if (mode == MBEDTLS_AES_ENCRYPT) {
-		ret = sl_aes_encrypt(ctx->shnd, &aes_in, &aes_param, ctx->enc_key_idx, &aes_out);
-	} else if (mode == MBEDTLS_AES_DECRYPT) {
-		ret = sl_aes_decrypt(ctx->shnd, &aes_in, &aes_param, ctx->dec_key_idx, &aes_out);
-	}
-
+	ret = sl_aes_encrypt(ctx->shnd, &aes_in, &aes_param, ctx->enc_key_idx, &aes_out);
```
an4967, line 204: Why don't you remove 'mode' param if this is not necessary anymore?

pillip8282: Thanks for review. I found that software method of mbedtls_aes_cyprt_cfb128 manipulate input and output depends on mode. I'll modify it.

an4967 (approved):
I tested this, and It works clearly

PR #5356, opened 2021-12-13
os/board/rtl8721csm: Change 'DiagPrintf()'/'DBG_8195A' to  'vddbg'

lu04-wang:


jeongchanKim (commented):
In commit titile, change 'DiagPrintf()'/'DBG_8195A' to 'vddbg'.
But in the commit, changed to `rtw_printf`.
Is it the same?

jeongchanKim (approved):
Oh I found that there is `vddbg` in `rtw_printf`.
Thanks.

lu04-wang:
> 

Hi jeongchanKim
Thanks for the review. The ticket was raised to change change 'DiagPrintf'/'DBG_8195A' to 'vddbg'. in os\drivers\wireless\realtek\os\os_dep\osdep_service.c, we define
int rtw_printf(const char *format,...)
{
	_tizenrt_printf(format);
} 
and in os\board\rtl8721csm\src\component\os\tizenrt\tizenrt_service.c. we define
int _tizenrt_printf(const char *format)
{
	vddbg(format, ##__VA_ARGS__);
}

Hence in the end, we are calling vddbg. 
Best Regards
Wang Lu

sunghan-chang (approved):

PR #5360, opened 2021-12-13
os/syscall: remove unnecessary and fix typo

sunghan-chang:
1. remove get_errno_ptr in syscall.csv
  The get_errno_ptr function is not allowed to access from user.
  That's why there is no get_errno_ptr in syscall_lookup.h.
  Let's remove it.
2. remove commented lines
  There are valid (uncommented) task_create and statfs.
  Let's remove unused duplicated commented lines.
3. fix typo of fs_fdopen
  The fdopen is in libc so that it does not need to be registered in syscall.
  But the fdopen uses fs_fdopen funtion internally, it's os API so that it should
  be a syscall API. The syscall.csv has a right name, fs_fdopen but syscall_lookup.h
  has a wrong name, fdopen.
  This commit fixes typo of fs_fdopen.


jeongchanKim (approved):
LGTM

thapav (approved):
LGTM

Taejun-Kwon (approved):

PR #5362, opened 2021-12-13
arch/armv8-m/up_blocktask.c : Fix wrong tcb for saving secure context

jeongchanKim:
To make context switching, current running tcb's secure context should be saved, not to be blocked one.
So change tcb to rtcb.

### DIFF
--- a/os/arch/arm/src/armv8-m/up_blocktask.c
+++ b/os/arch/arm/src/armv8-m/up_blocktask.c
@@ -144,7 +144,7 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 	if (switch_needed) {
 #ifdef CONFIG_ARMV8M_TRUSTZONE
 		if (tcb->tz_context) {
-			TZ_StoreContext_S(tcb->tz_context);
+			TZ_StoreContext_S(rtcb->tz_context);
 		}
 #endif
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/armv8-m/up_blocktask.c
```diff
@@ -144,7 +144,7 @@ void up_block_task(struct tcb_s *tcb, tstate_t task_state)
 	if (switch_needed) {
 #ifdef CONFIG_ARMV8M_TRUSTZONE
 		if (tcb->tz_context) {
-			TZ_StoreContext_S(tcb->tz_context);
+			TZ_StoreContext_S(rtcb->tz_context);
```
abhishek-samsung, line 147: Context switch will be needed only if the task that is being blocked is at the head of the ready-to-run list  (otherwise the head won't change and no context switch is needed). So, both rtcb and tcb will be same in this case.

abhishek-samsung (commented):
LGTM

sunghan-chang:
As far as I know, it's destination, not source. Your description could give wrong information. Could you check this?

jeongchanKim:
> As far as I know, it's destination, not source. Your description could give wrong information. Could you check this?

Modified. Thanks.

sunghan-chang (approved):

PR #5364, opened 2021-12-14
framework/ble_manager: implement BLE priority queue.

an4967:
- implement BLE priority queue.
- add ble_evt_handler

### DIFF
--- a/apps/examples/ble_rmc/ble_rmc_main.c
+++ b/apps/examples/ble_rmc/ble_rmc_main.c
@@ -68,19 +68,26 @@ static void ble_scan_state_changed_cb(ble_scan_state_e scan_state)
 
 /* These values can be modified as a developer wants. */
 static uint8_t ble_filter[] = { 0x02, 0x01, 0x05, 0x03, 0x19, 0x80, 0x01, 0x05, 0x03, 0x12, 0x18, 0x0f, 0x18 };
-static uint8_t ble_mac_filter[2][BLE_BD_ADDR_MAX_LEN] = { 
-	{ 0x10, 0x00, 0x00, 0x00, 0x00, 0x15 }, 
-	{ 0x28, 0x6d, 0x97, 0x86, 0x60, 0xc0 }, 
-};
 
-static void ble_device_scanned_cb_with_filter(ble_scanned_device *scanned_device)
+static void ble_device_scanned_cb_for_test(ble_scanned_device *scanned_device)
+{
+	RMC_LOG(RMC_CLIENT_TAG, "scanned mac : %02x:%02x:%02x:%02x:%02x:%02x\n", 
+		scanned_device->addr.mac[0],
+		scanned_device->addr.mac[1],
+		scanned_device->addr.mac[2],
+		scanned_device->addr.mac[3],
+		scanned_device->addr.mac[4],
+		scanned_device->addr.mac[5]
+	);
+}
+
+static void ble_device_scanned_cb_for_connect(ble_scanned_device *scanned_device)
 {
 	if (g_scan_done == 1) {
 		return;
 	}
 
-	RMC_LOG(RMC_CLIENT_TAG, "Find RMC with scan filter!!!\n");
-	printf("scanned mac : %02x:%02x:%02x:%02x:%02x:%02x\n", 
+	RMC_LOG(RMC_CLIENT_TAG, "Found mac : %02x:%02x:%02x:%02x:%02x:%02x\n", 
 		scanned_device->addr.mac[0],
 		scanned_device->addr.mac[1],
 		scanned_device->addr.mac[2],
@@ -94,26 +101,6 @@ static void ble_device_scanned_cb_with_filter(ble_scanned_device *scanned_device
 		g_target.type = scanned_device->addr.type;
 		g_scan_done = 1;
 	}
-}
-
-static void ble_device_scanned_cb_without_filter(ble_scanned_device *scanned_device)
-{
-	/*
-	In without filter callback, you should not do heavy works such as printing every result of scan result.
-	*/
-	if (memcmp(scanned_device->raw_data, ble_filter, sizeof(ble_filter)) == 0) {
-		RMC_LOG(RMC_CLIENT_TAG,"Find RMC with raw data !!!\n");
-		printf("scanned mac : %02x:%02x:%02x:%02x:%02x:%02x\n", 
-			scanned_device->addr.mac[0],
-			scanned_device->addr.mac[1],
-			scanned_device->addr.mac[2],
-			scanned_device->addr.mac[3],
-			scanned_device->addr.mac[4],
-			scanned_device->addr.mac[5]
-		);
-	}
-	
-	return;
 }
 
 static void ble_device_disconnected_cb(ble_client_ctx *ctx)
@@ -309,6 +296,28 @@ static int ble_connect_common(ble_client_ctx *ctx, ble_addr *addr, bool is_auto)
 	}
 
 	return 0;
+}
+
+static void set_scan_timer(uint32_t *scan_time, char *data)
+{
+	int temp = atoi(data);
+	if (temp < 0) {
+		RMC_LOG(RMC_CLIENT_TAG, "Fail to set timer\n");
+	} else {
+		*scan_time = (uint32_t)temp;
+	}
+}
+
+static void set_scan_filter(ble_scan_filter *filter, uint8_t *raw_data, uint8_t len, bool whitelist_enable, uint32_t scan_duration)
+{
+	memset(filter, 0, sizeof(ble_scan_filter));
+	if (raw_data != NULL && len > 0) {
+		memcpy(filter->raw_data, raw_data, len);
+		filter->raw_data_length = len;
+	}
+
+	filter->scan_duration = scan_duration;
+	filter->whitelist_enable = whitelist_enable;
 }
 
 /****************************************************************************
@@ -557,40 +566,56 @@ int ble_rmc_main(int argc, char *argv[])
 	/* 
 	* [ Scan ] Usage :
 	* 1. Normal Scan with MAX Scan Timeout
-	* TASH>> scan 1
-	* 2. Filter Scan
-	* TASH>> scan 2 [timer_value]
-	* ( timer_value : optional. this should be in seconds )
+	* TASH>> ble_rmc scan 1
+	* 2. Whitelist Scan
+	* TASH>> ble_rmc scan 2 [timer_value]
+	* ( timer_value : optional. this should be in seconds, default : 5s )
+	* 3. Filter Scan
+	* TASH>> ble_rmc scan 3 [timer_value]
+	* ( timer_value : optional. this should be in seconds, default : 5s )
+	* 4. Stop Scan
+	* TASH>> ble_rmc scan
 	*/
 	if (strncmp(argv[1], "scan", 5) == 0) {
-		if (argc >= 3 && argv[2][0] == '1') {
-			RMC_LOG(RMC_CLIENT_TAG, "Scan Start !\n");
-			scan_config.ble_client_device_scanned_cb = ble_device_scanned_cb_without_filter;
+		if (argc >= 3 && strncmp(argv[2], "1", 2) == 0) {
+			RMC_LOG(RMC_CLIENT_TAG, "Scan Start without filter !\n");
+			scan_config.device_scanned_cb = ble_device_scanned_cb_for_test;
 			ret = ble_client_start_scan(NULL, &scan_config);
 
 			if (ret != BLE_MANAGER_SUCCESS) {
 				RMC_LOG(RMC_CLIENT_TAG, "scan start fail[%d]\n", ret);
 				goto ble_rmc_done;
 			}
-		} else if (argc >= 3 && argv[2][0] == '2') {
+		} else if (argc >= 3 && strncmp(argv[2], "2", 2) == 0) {
 			RMC_LOG(RMC_CLIENT_TAG, "Scan Start with WhiteList!\n");
 
 			uint32_t scan_time = 5; // Seconds
 			if (argc == 4) {
-				int temp = atoi(argv[3]);
-				if (temp < 0) {
-					RMC_LOG(RMC_CLIENT_TAG, "Fail to set timer\n");
-				} else {
-					scan_time = (uint32_t)temp;
-					RMC_LOG(RMC_CLIENT_TAG, "Timer : %u s\n", scan_time);
-				}
+				set_scan_timer(&scan_time, argv[3]);
 			}
+			RMC_LOG(RMC_CLIENT_TAG, "Timer : %us\n", scan_time);
 
 			ble_scan_filter filter = { 0, };
-			filter.raw_data_length = 0;
-			filter.scan_duration = scan_time * 1000; // scan_duration should be in milliseconds.
-			filter.whitelist_enable = false;
-			scan_config.ble_client_device_scanned_cb = ble_device_scanned_cb_with_filter;
+			set_scan_filter(&filter, NULL, 0, true, scan_time * 1000);
+			scan_config.device_scanned_cb = ble_device_scanned_cb_for_test;
+			ret = ble_client_start_scan(&filter, &scan_config);
+
+			if (ret != BLE_MANAGER_SUCCESS) {
+				RMC_LOG(RMC_CLIENT_TAG, "scan start fail[%d]\n", ret);
+				goto ble_rmc_done;
+			}
+		} else if (argc >= 3 && strncmp(argv[2], "3", 2) == 0) {
+			RMC_LOG(RMC_CLIENT_TAG, "Scan Start with Packet Filter!\n");
+
+			uint32_t scan_time = 5; // Seconds
+			if (argc == 4) {
+				set_scan_timer(&scan_time, argv[3]);
+			}
+			RMC_LOG(RMC_CLIENT_TAG, "Timer : %us\n", scan_time);
+
+			ble_scan_filter filter = { 0, };
+			set_scan_filter(&filter, ble_filter, sizeof(ble_filter), false, scan_time * 1000);
+			scan_config.device_scanned_cb = ble_device_scanned_cb_for_test;
 			ret = ble_client_start_scan(&filter, &scan_config);
 
 			if (ret != BLE_MANAGER_SUCCESS) {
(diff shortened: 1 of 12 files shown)


framework/include/ble_manager/ble_queue.h
```diff
@@ -0,0 +1,68 @@
...
+	BLE_QUEUE_EVT_PRI_LOW,
+	BLE_QUEUE_EVT_PRI_MAX,
+} ble_event_priority_e;
+
+typedef struct {
+	int size;
+	volatile int write_index;
+	volatile int read_index;
+	sem_t countsem;
+    sem_t *grp_count;
```
JeongJunSik, line 51: tab or 4 space ?

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -109,32 +110,55 @@ static ble_result_e _convert_ret(trble_result_e val)
...
+		} break;
+		default:
+			break;
 		}
-		if (scan_queue->read_index == scan_queue->write_index) {
-			BLE_LOG_INFO("[BLEMGR] Scan Info Queue is Empty\n");
-			continue;
+
+		if (msg->param[2] != NULL) {
+			free(msg->param[2]);
```
jeongchanKim, line 150: I think it is better to set NULL after `free`.
Because `free` will only release the memory, but not set NULL.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,198 @@
...
+static int g_run_process = 0;
+
+static int _ble_evt_handler(int argc, char *argv[])
+{
+	int priority;
+	int err_no;
+	ble_queue *q;
+
+	while (g_run_process) {
+		if (sem_wait(&q_grp->countsem) < 0) {
```
jeongchanKim, line 34: Please include `semaphore.h`.

an4967: @jeongchanKim 
It is included in ble_queue.h

framework/include/ble_manager/ble_queue.h
an4967: @JeongJunSik 
Thanks. I will fix it

framework/src/ble_manager/ble_manager_state.c
an4967: "msg->param[2]" is just temporary variable. That would never be used after "_event_caller" is done.
Hence, I think making that be NULL is not necessary.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -155,28 +179,26 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
 			break;
 		}
 
-		sem_init(&scan_queue->countsem, 0, 0);
-		scan_queue->size = SCAN_INFO_BUFFER_SIZE;
-		scan_queue->queue = (trble_scanned_device *)malloc(sizeof(trble_scanned_device) * scan_queue->size);
-		if (scan_queue->queue == NULL) {
-			ret = TRBLE_FAIL;
+		if (ble_queue_init(_event_caller) != BLE_QUEUE_SUCCESS) {
+			ret= TRBLE_FAIL;
```
jeongchanKim, line 183: `ret` is initialized with `TRBLE_FAIL`, and not changed before this `break`.
(Changed from line 178, but there is `break`.)
So I think we don't need to set again.
We can reduce the code then.

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -155,28 +179,26 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
 			break;
 		}
-
-		int tid = task_create("ble_evt_handler", 100, 2048, (main_t)_bledev_handler, NULL);
-		if (tid < 0) {
-			ret = TRBLE_FAIL;
+		if (ble_queue_pri_set(BLE_QUEUE_EVT_PRI_HIGH, BLE_EVT_HIGH_BUFFER_SIZE, sizeof(blemgr_msg_params)) != BLE_QUEUE_SUCCESS) {
+			ble_queue_deinit();
+			ret= TRBLE_FAIL;
```
jeongchanKim, line 188: Same as above.

framework/src/ble_manager/ble_queue.c
jeongchanKim: It is better to include the header file from its own file.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+
+	return 0;
+}
+
+ble_queue_ret_e ble_queue_init(queue_event_caller caller)
+{
+	if (q_grp == NULL) {
+		q_grp = (ble_queue_group *)malloc(sizeof(ble_queue_group));
```
jeongchanKim, line 67: How about using `zalloc` instead of `malloc`?
Then we don't need to call `memset` from line 71.

framework/src/ble_manager/ble_manager_state.c
an4967: Okay, I will fix it

framework/src/ble_manager/ble_queue.c
an4967: Okay, I will fix it

framework/src/ble_manager/ble_manager_state.c
an4967: Okay, I will fix it

jeongchanKim (approved):
LGTM

apps/examples/ble_rmc/ble_rmc_main.c
```diff
@@ -581,15 +573,41 @@ int ble_rmc_main(int argc, char *argv[])
...
 			filter.scan_duration = scan_time * 1000; // scan_duration should be in milliseconds.
+			filter.whitelist_enable = true;
+			scan_config.device_scanned_cb = ble_device_scanned_cb_for_test;
+			ret = ble_client_start_scan(&filter, &scan_config);
+
+			if (ret != BLE_MANAGER_SUCCESS) {
+				RMC_LOG(RMC_CLIENT_TAG, "scan start fail[%d]\n", ret);
+				goto ble_rmc_done;
+			}
+		} else if (argc >= 3 && argv[2][0] == '3') {
```
pillip8282, line 591: with this code argument 30 3 300 execute same code. Is it acceptable?

apps/examples/ble_rmc/ble_rmc_main.c
```diff
@@ -581,15 +573,41 @@ int ble_rmc_main(int argc, char *argv[])
...
+			RMC_LOG(RMC_CLIENT_TAG, "Scan Start with Packet Filter!\n");
+
+			uint32_t scan_time = 5; // Seconds
+			if (argc == 4) {
+				int temp = atoi(argv[3]);
+				if (temp < 0) {
+					RMC_LOG(RMC_CLIENT_TAG, "Fail to set timer\n");
+				} else {
```
pillip8282, line 599: This case code in switch is long. how about to make it as function?

apps/examples/ble_rmc/ble_rmc_main.c
```diff
@@ -581,15 +573,41 @@ int ble_rmc_main(int argc, char *argv[])
...
+					scan_time = (uint32_t)temp;
+				}
+			}
+			RMC_LOG(RMC_CLIENT_TAG, "Timer : %us\n", scan_time);
+
+			ble_scan_filter filter = { 0, };
+			memcpy(filter.raw_data, ble_filter, sizeof(ble_filter));
```
pillip8282, line 606: It seems repeated code how about to make it function?

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+#include <tinyara/config.h>
+#include <errno.h>
+#include <ble_manager/ble_queue.h>
+
+#define BLE_QUEUE_CHECK do { if (q_grp == NULL) {return BLE_QUEUE_NOT_INIT;}} while(0)
+
+static ble_queue_group *q_grp = NULL;
```
pillip8282, line 24: how about to add g_ prefix for consistency?

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+	int priority;
+	int err_no;
+	ble_queue *q;
+
+	while (g_run_process) {
+		if (sem_wait(&q_grp->countsem) < 0) {
+			if (errno == EINTR) {
+				continue;
+			} else {
+				g_run_process = 0;
```
pillip8282, line 38: how about to print errno for debugging.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+		for (i = 0; i < BLE_QUEUE_EVT_PRI_MAX; i++) {
+			ret = ble_queue_pri_clear(i);
+			if (ret != BLE_QUEUE_SUCCESS) {
+				return BLE_QUEUE_FAIL;
+			}
+		}
+		q_grp->caller = NULL;
+		g_run_process = 0;
+		sem_post(&q_grp->countsem);
+		sem_destroy(&q_grp->countsem);
```
pillip8282, line 110: is it ok to sem_destroy in here right after sem_post? I mean task which in sem_wait or right before call sem_wait might be dangerous.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+	ble_queue *q = &q_grp->q[priority];
+
+	if (q->queue == NULL) {
+		q->queue = (void *)malloc(data_size * queue_size);
+		if (q->queue == NULL) {
+			printf("[BLEQUEUE] fail to set priority queue[%d]\n", priority);
+			return BLE_QUEUE_MEM_ALLOC_FAIL;
+		}
+		sem_init(&q->countsem, 0, 0);
+		q->grp_count = &q_grp->countsem;
```
pillip8282, line 133: Could you tell me why named a value countsem? I'll helpful if you explained the reason.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+	if (priority < 0 || priority >= BLE_QUEUE_EVT_PRI_MAX) {
+		return BLE_QUEUE_INVALID_ARGS;
+	}
+
+	ble_queue *q = &q_grp->q[priority];
+
+	if (q->queue != NULL) {
+		free(q->queue);
+		q->queue = NULL;
+		sem_destroy(&q->countsem);
```
pillip8282, line 157: Is it guaranteed that there is no task which use q->countsem? It should be checked. And if you comment the reason then it will help other developers.

framework/src/ble_manager/ble_queue.c
```diff
@@ -0,0 +1,204 @@
...
+	BLE_QUEUE_CHECK;
+
+	ble_queue *q = &q_grp->q[priority];
+
+	if (q->queue == NULL) {
+		return BLE_QUEUE_NOT_INIT;
+	}
+
+	int write_index_next = (q->write_index + 1) % q->size;
+	if (write_index_next == q->read_index) {
```
pillip8282, line 179: mutex is not needed to protect q->queue? If only one thread or task access the queue, it won't to be protected. if not it must be preserved.

pillip8282 (commented):
This PR will solve BLE responsibility. Adding priority queue is good approach to solve it.
Next time I have 2 suggestions
1) divide the commit (the commit size is too long so it's difficult to understand)
2) TDD (how about to add testcases to verify priority queue)

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -925,10 +943,14 @@ ble_result_e blemgr_handle_request(blemgr_msg_s *msg)
...
 
-		if (ctx && ctx->callbacks.ble_client_device_connected_cb) {
-			ctx->callbacks.ble_client_device_connected_cb((ble_client_ctx *)ctx, data);
+		if (ctx && ctx->callbacks.connected_cb) {
+			blemgr_msg_params param = {
+				.count = 3,
+				.param = {ctx->callbacks.connected_cb, ctx, msg->param},
+				.evt = msg->event
+			};
+			ble_queue_enque(BLE_QUEUE_EVT_PRI_HIGH, &param);
```
Taejun-Kwon, line 952: It is almost duplicated code, you can reduce it.

os/net/blemgr/bledev.c
```diff
@@ -60,13 +60,14 @@ int trble_scan_data_enque(trble_scanned_device *info)
 
 	int i;
 	uint32_t *u1 = (uint32_t *)info;
-	uint32_t *u2 = (uint32_t *)&g_scan_queue->queue[g_scan_queue->write_index];
-	for (i = 0; i < sizeof(trble_scanned_device) / sizeof(uint32_t); i++) {
+	uint32_t *u2 = (uint32_t *)(g_scan_queue->queue + (g_scan_queue->data_size * g_scan_queue->write_index));
+	for (i = 0; i < g_scan_queue->data_size / sizeof(uint32_t); i++) {
```
Taejun-Kwon, line 64: I think value of data_size can be changed in multi-thread environment, is it intended?
To make loop more stable you had better use below style to prevent error by changed condition. It is very basic style.
```
int a = a.size;  // int a = funcA();
for (i = 0; i < a; .....)
```

os/net/blemgr/bledev.c
```diff
@@ -92,7 +93,7 @@ int bledev_handle(struct bledev *dev, lwnl_req cmd, void *data, uint32_t data_le
 		if (t_server == NULL) {
 			t_server = bledrv_server_get_null_config();
 		}
-		g_scan_queue = (trble_scan_queue *)param.param[1];
+		g_scan_queue = (trble_queue *)param.param[1];
```
Taejun-Kwon, line 96: you should null check data  first. Since you develop middle layer, you should follow that coding style to prevent error in future.

apps/examples/ble_rmc/ble_rmc_main.c
an4967: I will fix it to use 'strncmp'

apps/examples/ble_rmc/ble_rmc_main.c
an4967: Okay, I will make it as a function

apps/examples/ble_rmc/ble_rmc_main.c
an4967: Okay, I will make it as a function

framework/src/ble_manager/ble_queue.c
an4967: Okay, I will fix it as 'g_q_grp'

framework/src/ble_manager/ble_queue.c
an4967: I will add the log

framework/src/ble_manager/ble_queue.c
an4967: Yes, you're right.
I will fix this to handle semaphores in the event_handler which waits them.

framework/src/ble_manager/ble_queue.c
an4967: The countsem of queue is for counting queue data.
Furthermore, the queue_group counts the total quantity of priority queues.

framework/src/ble_manager/ble_queue.c
an4967: Yes. it is guaranteed. Because a semaphore of a queue is waited by sem_tyrwait which is non-blocking function.

framework/src/ble_manager/ble_queue.c
an4967: Mutex is not necessary, because each queue has a single producer & single consumer.

framework/include/ble_manager/ble_queue.h
```diff
@@ -0,0 +1,68 @@
+/****************************************************************************
```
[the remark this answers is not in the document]
an4967: Yes, you are right. I will move it to _framework/src_

framework/src/ble_manager/ble_manager_state.c
```diff
@@ -109,32 +110,55 @@ static ble_result_e _convert_ret(trble_result_e val)
...
-		if (sem_wait(&scan_queue->countsem) < 0) {
-			if (errno == EINTR) {
-				continue;
-			} else {
-				g_run_process = 0;
-				return -1;
+static void _event_caller(int evt_pri, void *data) {
+	switch (evt_pri) {
+	case BLE_QUEUE_EVT_PRI_HIGH: {
+		blemgr_msg_params *msg = (blemgr_msg_params *)data;
```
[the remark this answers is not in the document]
an4967: Okay, I will add null check statement

framework/src/ble_manager/ble_manager_state.c
an4967: Okay, I will remove duplicated code

os/net/blemgr/bledev.c
an4967: 'data_size' of queue is not changed after initialized first time.

framework/src/ble_manager/ble_queue.c
pillip8282: I think sem_destroy can run during other task is running sem_trywait without protection. please check.

framework/src/ble_manager/ble_queue.c
pillip8282: enqueue function refers q->read_index then it can cause synchronization problem even though it's single producer & single consumer. For example please check following case.
```
ble_queue_ret_e ble_queue_enque(int priority, void *data) // 1) producer enqueue data R: 1 W: 0 (read index, write index)
if (q->read_index != q->write_index) { // 2) consumer check index R:1 W:0
q->write_index = write_index_next; // 3) context switching. a producer enqueue data again
// 4) R:1 W:1
if (q->read_index != q->write_index) { 5) context switching. packet drop because both indexes are same.
```

an4967: In 3) case, enque cannot process because it checks (W + 1 == R) for judging Q full when trying to enque.

framework/src/ble_manager/ble_queue.c
an4967: I checked all cases and I realized that 'ble_queue_pri_clear' is not needed.
The queue should be removed only when BLE group being de-initialized.
So, I removed ''ble_queue_pri_clear' function

pillip8282 (approved):
LGTM

PR #5365, opened 2021-12-14
fs/driver/mtd: Fix wrong output from '/proc/partitions'

jeongarmy:
- Fix wrong entry function name to 'procfs_test_main'
- Set the name of mtd devices in when device initialization
- fs/driver/mtd: Fix wrong output from '/proc/partitions'
  - Ex) cat /proc/partitions
 Name             Start    Size   MTD
            bl1       0       4    ameba_flash
            bl2       4       2    ameba_flash
       reserved       6       4    ameba_flash
             ss      10     118    ameba_flash
    system_data     128       2    ameba_flash
         kernel     130     448    ameba_flash
         kernel     578     448    ameba_flash
         userfs    1026    1020    ameba_flash
      bootparam    2046       2    ameba_flash

### DIFF
--- a/os/arch/arm/src/stm32l4/stm32l4_ospi.c
+++ b/os/arch/arm/src/stm32l4/stm32l4_ospi.c
@@ -1920,6 +1920,9 @@ FAR struct mtd_dev_s *up_flashinitialize(void)
     priv->bwrite = this_bwrite;
     priv->read = this_read;
     priv->ioctl = this_ioctl;
+#ifdef CONFIG_MTD_REGISTRATION
+    priv->mtd.name = "stm32l4_flash";
+#endif
   }
 
   return priv;
--- a/os/board/common/partitions.c
+++ b/os/board/common/partitions.c
@@ -100,6 +100,7 @@ FAR struct mtd_dev_s *mtd_initialize(void)
 
 static int type_specific_initialize(int minor, FAR struct mtd_dev_s *mtd_part, int partno, const char *types, partition_info_t *partinfo)
 {
+	int tagno = MTD_NONE;
 #ifdef CONFIG_FS_ROMFS
 	bool save_romfs_partno = false;
 #endif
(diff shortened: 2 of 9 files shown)


Taejun-Kwon:
We use only MTD driver for flash, so I think we can remove MTD column.
Instead of MTD, we can put some type of block driver, FTL, smart, N/A(c.f kernel)  something.

os/include/tinyara/fs/mtd.h
```diff
@@ -170,11 +170,13 @@ struct mtd_dev_s {
...
 
 enum mtd_partition_tag_s {
-	MTD_MASTER = 0,
+	MTD_NONE = 0,
 	MTD_FS = 1,
 	MTD_NV = 2,
 	MTD_OTA = 3,
-	MTD_ROMFS = 4
+	MTD_ROMFS = 4,
+	MTD_FTL = 5,
```
Taejun-Kwon, line 178: could you please add description here?
```
MTD_FTL = 5, /* Directly use FTL as a block driver(ex. BCH)

File: os/arch/arm/src/stm32l4/stm32l4_ospi.c [hunk 2]
```diff
@@ -1920,6 +1920,9 @@ FAR struct mtd_dev_s *up_flashinitialize(void)
     priv->bwrite = this_bwrite;
     priv->read = this_read;
     priv->ioctl = this_ioctl;
+#ifdef CONFIG_MTD_REGISTRATION
+	priv->mtd.name = "stm32l4_flash";
```
JeongJunSik, line 1924: please check indentation(tab size : 4 space)

os/include/tinyara/fs/mtd.h
```diff
@@ -237,6 +239,16 @@ FAR struct mtd_dev_s *mtd_partition(FAR struct mtd_dev_s *mtd, off_t firstblock,
  ****************************************************************************/
 FAR struct mtd_dev_s *get_mtd_partition(uint16_t tagno);
```
jeongchanKim, line 240: Currently, we can get only `MTD_FS`, `MTD_ROMFS` and `MTD_FTL` type.
Please add description about this.

os/board/common/partitions.c
```diff
@@ -114,26 +115,31 @@ static int type_specific_initialize(int minor, FAR struct mtd_dev_s *mtd_part, i
...
-	if (!strncmp(types, "ftl,", 4)
+	if (!strncmp(types, "ftl,", 4)) {
+		tagno = MTD_FTL;
+		do_ftlinit = true;
+	}
 #ifdef CONFIG_BINARY_MANAGER
-	|| !strncmp(types, "kernel,", 7)
+	else if (!strncmp(types, "kernel,", 7)
 	|| !strncmp(types, "bootparam,", 10)
 #ifdef CONFIG_APP_BINARY_SEPARATION
```
jeongchanKim, line 125: If enabled `CONFIG_BINARY_MANAGER`, but not enabled `CONFIG_APP_BINARY_SEPARATION`, it makes build error.
Please check this.

os/include/tinyara/fs/mtd.h
jeongarmy: I think I don't need to add description about that because it is not that this function doesn't support other values.

sunghan-chang (approved):

PR #5367, opened 2021-12-14
external/webserver: Increase timeout for keep-alive connections

vibhor-m:
Current timeout for read operation is 50 seconds.
Increase timeout to 300 seconds for keep-alive connection requests.

### DIFF
--- a/external/include/protocols/webserver/http_server.h
+++ b/external/include/protocols/webserver/http_server.h
@@ -93,6 +93,7 @@
 #define HTTP_CONF_CLIENT_STACKSIZE              8192
 #define HTTP_CONF_MIN_TLS_MEMORY                80000
 #define HTTP_CONF_SOCKET_TIMEOUT_MSEC           50000
+#define HTTP_CONF_KEEP_ALIVE_MSEC               300000
 #define HTTP_CONF_SERVER_MQ_MAX_MSG             10
 #define HTTP_CONF_SERVER_MQ_PRIO                50
 #define HTTP_CONF_SERVER_SIGWAKEUP              18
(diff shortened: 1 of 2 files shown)


external/webserver/http_client.c
```diff
@@ -582,6 +582,15 @@ int http_recv_and_handle_request(struct http_client_t *client, struct http_keyva
 	conn_type = http_keyvalue_list_find(request_params, "Connection");
 	if (!strncasecmp(conn_type, "Keep-Alive", strlen("Keep-Alive")+1)) {
 		client->keep_alive = 1;
+		struct timeval tv;
+		tv.tv_sec = HTTP_CONF_KEEP_ALIVE_CONN_SOCKET_TIMEOUT_MSEC / 1000;
+		tv.tv_usec = (HTTP_CONF_KEEP_ALIVE_CONN_SOCKET_TIMEOUT_MSEC % 1000) * 1000;
+		HTTP_LOGD("Keep-alive case, change timeout to (%d.%d)sec\n", tv.tv_sec, tv.tv_usec);
```
jeongchanKim, line 588: `tv_sec` is unsigned type, so I think it is better to use `%u` instead of `%d`.

vibhor-m: change done

external/include/protocols/webserver/http_server.h
```diff
@@ -93,6 +93,7 @@
 #define HTTP_CONF_CLIENT_STACKSIZE              8192
 #define HTTP_CONF_MIN_TLS_MEMORY                80000
 #define HTTP_CONF_SOCKET_TIMEOUT_MSEC           50000
+#define HTTP_CONF_KEEP_ALIVE_CONN_SOCKET_TIMEOUT_MSEC 300000
```
Taejun-Kwon, line 96: I think HTTP_CONF_KEEP_ALIVE_MSEC is enough name, current is too long.

deepaksrma: yes, try to keep it short and can provide a comment .

pillip8282 (approved):
lgtm

PR #5368, opened 2021-12-15
os/board/rtl8721csm: Update BT Coex table

lu04-wang:


an4967:
@lu04-wang 
Can you comment how did you change the Coex Table?

lu04-wang:
Hi @an4967 
I have change from case 2 to case 8 when Station is connected and BT  on.

an4967 (approved):
LGTM

PR #5369, opened 2021-12-15
framework/ble_manager: adjust BLE configs

an4967:
- relocate realtek BLE driver
- set BLE Kconfig for board specific configs


jeongchanKim (approved):
LGTM

os/drivers/bluetooth/realtek_ble/Kconfig
```diff
@@ -0,0 +1,34 @@
...
+	int "BLE Peripheral Connection"
+	default 1
+	---help---
+		Set the Maximum number of BLE Peripheral Connection
+
+config BLE_MAX_BONDED_DEVICE
+	int "BLE Max Bonded Device"
+	default 5
+	---help---
+		Set the Maximum number of BLE Bond Data
```
thapav, line 22: Does BLE Bond data mean BLE Bonded device only ?

an4967: Yes. That's right.
BLE driver only stores a mac address of bonded device.

framework/include/ble_manager/ble_common.h
```diff
@@ -38,8 +38,17 @@
 #define BLE_ADV_RESP_DATA_MAX_LEN 31
 
 /* These are board specific values */
-#define BLE_MAX_CONNECTION_COUNT 3
-#define BLE_MAX_BONDED_DEVICE 10
+#ifdef CONFIG_BLE_CENTRAL_CONN_NUM
+#define BLE_MAX_CENTRAL_CONN_NUM CONFIG_BLE_CENTRAL_CONN_NUM
```
Taejun-Kwon, line 42: I am sorry but This is VERY BAD approach, framework NEVER have dependency with H/W
Let's assume that we have a USB port and each of dongle has different number of connection.
If user changed dongle , what happened?????
So every init time of framework, you should ask h/w spec to driver.
In this time, you can use ioctl to get that info.

an4967: Yes, you are right.
Hence 'CONFIG_BLE_CENTRAL_CONN_NUM' is set by realtek driver layer.
All options which has dependency with H/W are located in _os/drivers/bluetooth/realtek_ble/Kconfig_,
As following your example, if an user changes dongle, the driver code should be changed too including Kconfig.

Taejun-Kwon: I mean various of h/w can be changed in real time, with one binary.
That's why most of platform has routine what checking h/w params when it initialized. F/W should not use config of h/w directly.

framework/include/ble_manager/ble_scan.h
```diff
@@ -64,8 +64,18 @@ typedef struct {
 
 /* This is a heuristic value. It can be customized following a system env */
 #define SCAN_INFO_BUFFER_SIZE 100
+
+#ifdef CONFIG_BLE_SCAN_MAX_WHITELIST
+#define SCAN_WHITELIST_SIZE CONFIG_BLE_SCAN_MAX_WHITELIST
+#else
```
pillip8282, line 70: How about to gather all BLE configuration in one file(ex. ble_config.h).
Then it's more easy to handle ble configurations.

os/drivers/bluetooth/realtek_ble/Make.defs
```diff
@@ -19,11 +19,11 @@
 ifeq ($(CONFIG_RTK_BLE),y)
 
 #Driver API
-CSRCS += 
+CSRCS += rtk_ble_driver.c
```
pillip8282, line 22: it seems right to move ble driver file to here.

framework/include/ble_manager/ble_common.h
pillip8282: I agree to Taejon's opinion. If new ble driver need to be ported then same configurations(ex. BLE_MAX_CENTRAL_CONN_NUM) should be defined to it too. So how about to define these configurations in BLE framework?

PR #5370, opened 2021-12-15
tools/mkbootparam: Change a range of CRC checking to the whole bootparam region with 4Kbytes

jeongarmy:
Change a range of CRC checking to the whole bootparam region with 4Kbytes to remove dependency of build type.

### DIFF
--- a/os/tools/mkbootparam.py
+++ b/os/tools/mkbootparam.py
@@ -24,24 +24,31 @@ config_file_path = os.path.dirname(__file__) + '/../.config'
 bootparam_file_path = os.path.dirname(__file__) + '/../../build/output/bin/bootparam.bin'
 user_file_dir = os.path.dirname(__file__) + '/../../build/output/bin/user'
 
-SIZE_OF_BOOTPARAM_PART = 8192
+SIZE_OF_BPx = 4096
+
+# The partition size of dual BPs
+SIZE_OF_BP_PARTITION = SIZE_OF_BPx * 2
 
 ###########################################################################################
 #
-# This script generates Boot Parameters for kernel.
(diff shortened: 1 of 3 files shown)


os/tools/mkbootparam.py
```diff
@@ -24,24 +24,31 @@
...
+SIZE_OF_BP_PARTITION = SIZE_OF_BPx * 2
 
 ###########################################################################################
 #
-# This script generates Boot Parameters for kernel.
+# This script generates Boot Parameters.
+#
+# Boot Parameter works based on dual partition, BP1 and BP2.
+# It puts BP data to first BP, BP1 and clears second BP, BP2.
+# Then it makes a output with the name 'bootparam.bin' (8K) by concatenating BP1 and BP2.
```
jeongchanKim, line 38: it makes an output

os/tools/mkbootparam.py
```diff
@@ -58,105 +65,108 @@
...
+	user_app_count = len(user_file_list)
+	if user_app_count > 0:
+		SIZE_OF_BINCOUNT = 1
+		SIZE_OF_BINNAME = 16
+		SIZE_OF_BINIDX = 1
+		app_data_size = SIZE_OF_BINCOUNT + (user_app_count * (SIZE_OF_BINNAME + SIZE_OF_BINIDX))
+
+		with open(bootparam_file_path, 'a') as fp:
+			# User Data
+			user_active_idx = 0
```
jeongchanKim, line 152: Most of not changed variables have upper case name. 
This variable is also not changed.
How about changing?

sunghan-chang (approved):

PR #5373, opened 2021-12-16
external/webserver: Support of Trailer header on webserver

amitpurwar:
Support of Trailer header on webserver

### DIFF
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -167,6 +167,49 @@ int http_client_release(struct http_client_t *client)
 	return HTTP_OK;
 }
 
+static int process_trailer_header(char *buf, unsigned int buf_len, struct http_req_message *req,
+				struct http_keyvalue_list_t *params,
+				struct http_client_t *client, unsigned int sentence_start)
+{
+	char *header_field_value = NULL;
+	unsigned int sentence_end;
+	char *entity;
+	unsigned int trailter_len;
+
+	/* Process the Trailer */
+	header_field_value = http_keyvalue_list_find(params, "Trailer");
+	if (strncmp(header_field_value, "(null)", 7) == 0) {
+		return -1;
+	}
+
+	/* match the Trailer field */
+	if (strncmp(buf + sentence_start, header_field_value, strlen(header_field_value))
+		 || *(buf + sentence_start + strlen(header_field_value)) != ':' ) {
+		return -1;
+	}
+
+	HTTP_LOGD("Trailer header: %s\n", header_field_value);
+	sentence_end = http_find_first_crlf(buf, buf_len, sentence_start);
+	if (sentence_end < 0 || sentence_end - sentence_start == 0) {
+		HTTP_LOGE("Error: Not accord with chunked encoding\n");
+		return -1;
+	}
+
+	/* copy Trailer data */
+	trailter_len = sentence_end - sentence_start;
+	entity = (char *)zalloc(trailter_len + 1);
+	if (entity == NULL) {
+		HTTP_LOGE("Error: Fail to alloc memory\n");
+		return -1;
+	}
+	memcpy(entity, buf + sentence_start, trailter_len);
+	req->entity = entity;
+	req->entity_len = trailter_len;
+	http_dispatch_url(client, req);
+
+	return 1;
+}
+
 int http_parse_message(char *buf, int buf_len, int *method, char *url,
 					   char **body, int *enc, int *state,
 					   struct http_message_len_t *len,
@@ -184,6 +227,7 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 	int read_finish = false;
 	char *entity = *body;
 	bool is_chunked = false;
+	int ret = 0;
 
 	while (!process_finish) {
 		/* At this point, we read a line of http request */
@@ -476,6 +520,13 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 						http_dispatch_url(client, req);
 						process_finish = true;
 						read_finish = true;
+
+						/* sentence start is next statement after \r\n */
+						len->sentence_start = sentence_end + 2;
+						ret = process_trailer_header(buf, buf_len, req, params, client, len->sentence_start);
+						if (ret < 0) {
+							HTTP_LOGD("Trailer header is not present or failed to process it \n");
+						}
 					}
 				}
 			}
@@ -542,7 +593,7 @@ int http_recv_and_handle_request(struct http_client_t *client, struct http_keyva
 			len = recv(client->client_fd, buf + buf_len, HTTP_CONF_MAX_REQUEST_LENGTH - buf_len, 0);
 		}
 		if (len < 0) {
-			HTTP_LOGE("Error: Receive Fail %d\n", len);
+			HTTP_LOGE("Error: Receive Fail %d errno:[%s-%d] \n", len, strerror(errno), errno);
 			goto errout;
 		} else if (len == 0) {
 			HTTP_LOGD("Finish read\n");


external/webserver/http_client.c
```diff
@@ -476,6 +477,30 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 						http_dispatch_url(client, req);
 						process_finish = true;
 						read_finish = true;
+
+						/* Process the Trailer */
+						header_field_value = http_keyvalue_list_find(params, "Trailer");
```
pillip8282, line 482: how about to make this code to function?

external/webserver/http_client.c
```diff
@@ -476,6 +477,30 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
...
+						if (strncmp(header_field_value, "(null)", 6)) {
```
pillip8282, line 483: it should be 7 strlen("(null)") + 1

external/webserver/http_client.c
```diff
@@ -476,6 +477,30 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
...
+							len->sentence_start = sentence_end + 2;
+
+							/* match the Trailer field */
+							if (strncmp(buf + len->sentence_start, header_field_value, strlen(header_field_value)) == 0) {
```
pillip8282, line 487: compare length strlen doesn't check null character. So you need to modify it.

external/webserver/http_client.c
```diff
@@ -476,6 +477,30 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
...
+								HTTP_LOGE("Trailer header: %s\n", header_field_value);
+								sentence_end = http_find_first_crlf(buf, buf_len, len->sentence_start);
+								if (sentence_end < 0) {
+									HTTP_LOGE("Error: Not accord with chunked encoding\n");
+									read_finish = true;
+									return read_finish;
+								}
+
+								/* copy Trailer data */
+								memset(entity, 0, HTTP_CONF_MAX_ENTITY_LENGTH);
```
pillip8282, line 497: is it guaranteed that the size of entity is less than len->sentence_end - len->sentence_start?

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
```diff
@@ -167,6 +167,45 @@ int http_client_release(struct http_client_t *client)
...
+	char *header_field_value = NULL;
+	int sentence_end;
+	char *entity;
+	int trailter_len;
+
+	/* Process the Trailer */
+	header_field_value = http_keyvalue_list_find(params, "Trailer");
+	if (strncmp(header_field_value, "(null)", 7)) {
+		/* match the Trailer field */
+		if (strncmp(buf + sentence_start, header_field_value, strlen(header_field_value)) == 0) {
```
jeongchanKim, line 183: I think it is better to check with `strlen() + 1` for null termination.

external/webserver/http_client.c
```diff
@@ -167,6 +167,45 @@ int http_client_release(struct http_client_t *client)
 	return HTTP_OK;
 }
 
+int process_trailer_header(char *buf, int buf_len, struct http_req_message *req,
```
pillip8282, line 170: how about to make this function static?

external/webserver/http_client.c
```diff
@@ -167,6 +167,45 @@ int http_client_release(struct http_client_t *client)
...
+			}
+			memset(entity, 0, trailter_len + 1);
+			memcpy(entity, buf + sentence_start, trailter_len);
+			req->entity = entity;
+			req->entity_len = trailter_len;
+			http_dispatch_url(client, req);
+		}
+	}
+
+	return 1;
```
Taejun-Kwon, line 206: so it should return 1 even if result of first strncmp is not true? otherwise, you should apply below style
```
if (!strncmp((header_field_value, "(null)", 7)) {
     return -1;
}

if (strncmp(buf + sentence_start, header_field_value, strlen(header_field_value)) == 0) {
....
}
```

external/webserver/http_client.c
```diff
@@ -167,6 +167,45 @@ int http_client_release(struct http_client_t *client)
...
+	int trailter_len;
+
+	/* Process the Trailer */
+	header_field_value = http_keyvalue_list_find(params, "Trailer");
+	if (strncmp(header_field_value, "(null)", 7)) {
+		/* match the Trailer field */
+		if (strncmp(buf + sentence_start, header_field_value, strlen(header_field_value)) == 0) {
+			HTTP_LOGE("Trailer header: %s\n", header_field_value);
+			sentence_end = http_find_first_crlf(buf, buf_len, sentence_start);
+			if (sentence_end < 0) {
```
Taejun-Kwon, line 186: I think below is correct, but 0 is acceptable for sentence_end ?
```
if (sentence_end < 0 || sentence_end - sentence_start <= 0) {
}
```

external/webserver/http_client.c
```diff
@@ -167,6 +167,45 @@ int http_client_release(struct http_client_t *client)
 	return HTTP_OK;
 }
 
+int process_trailer_header(char *buf, int buf_len, struct http_req_message *req,
+				struct http_keyvalue_list_t *params,
+				struct http_client_t *client, int sentence_start)
+{
+	char *header_field_value = NULL;
+	int sentence_end;
```
Taejun-Kwon, line 175: I think size_t or unsigned int is good for kind of size, offset,...

external/webserver/http_client.c
pillip8282: Then how about to check '\r' if it's end of line? It's more safe

amitpurwar: I have added the additional validation of collon (:) after trailer header (Expires)

external/webserver/http_client.c
```diff
@@ -476,6 +520,9 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 						http_dispatch_url(client, req);
 						process_finish = true;
 						read_finish = true;
+
+						len->sentence_start = sentence_end + 2;
+						process_trailer_header(buf, buf_len, req, params, client, len->sentence_start);
```
pillip8282, line 525: it doesn't check return value from function. It can be dangerous. please handle error cases.

external/webserver/http_client.c
```diff
@@ -167,6 +167,50 @@ int http_client_release(struct http_client_t *client)
...
+	HTTP_LOGD("Trailer header: %s\n", header_field_value);
+	sentence_end = http_find_first_crlf(buf, buf_len, sentence_start);
+	if (sentence_end < 0 || sentence_end - sentence_start == 0) {
+		HTTP_LOGE("Error: Not accord with chunked encoding\n");
+		return -1;
+	}
+
+	/* copy Trailer data */
+	trailter_len = sentence_end - sentence_start;
+	entity = HTTP_MALLOC(trailter_len + 1);
```
jeongarmy, line 200: How about using zalloc instead of malloc and memset?

external/webserver/http_client.c
```diff
@@ -476,6 +520,9 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 						http_dispatch_url(client, req);
 						process_finish = true;
 						read_finish = true;
+
+						len->sentence_start = sentence_end + 2;
```
jeongarmy, line 524: What does `2` mean? Please add the comment about that for better understanding.

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
```diff
@@ -476,6 +520,13 @@ int http_parse_message(char *buf, int buf_len, int *method, char *url,
 						http_dispatch_url(client, req);
 						process_finish = true;
 						read_finish = true;
+
+						/* sentence start is next statement after \r\n */
+						len->sentence_start = sentence_end + 2;
+						ret = process_trailer_header(buf, buf_len, req, params, client, len->sentence_start);
+						if(ret < 0) {
```
sunghan-chang, line 527: if (

external/webserver/http_client.c
```diff
@@ -167,6 +167,49 @@ int http_client_release(struct http_client_t *client)
...
+	HTTP_LOGD("Trailer header: %s\n", header_field_value);
+	sentence_end = http_find_first_crlf(buf, buf_len, sentence_start);
+	if (sentence_end < 0 || sentence_end - sentence_start == 0) {
+		HTTP_LOGE("Error: Not accord with chunked encoding\n");
+		return -1;
+	}
+
+	/* copy Trailer data */
+	trailter_len = sentence_end - sentence_start;
+	entity = zalloc(trailter_len + 1);
```
sunghan-chang, line 200: I think this causes build warning for type mismatch.
```
entity = (char *)zalloc(....);
```

amitpurwar: done

external/webserver/http_client.c
amitpurwar: done

pillip8282 (approved):
LGTM

PR #5377, opened 2021-12-20
net/lwip: support window scale option

pillip8282:
support window scale option

### DIFF
--- a/os/net/lwip/configs/tcp/Kconfig
+++ b/os/net/lwip/configs/tcp/Kconfig
@@ -18,6 +18,24 @@ config NET_TCP_WND
 		The size of a TCP window.  This must be at least (2 * TCP_MSS)
 		for things to work well
 
+config NET_WND_SCALE
+	bool "Window scale support"
+	range 0 14
+	default n
+	---help---
+		Set LWIP_WND_SCALE to 1 to enable window scaling.
+
+
+if NET_WND_SCALE
+config NET_TCP_RCV_SCALE
+	int "Receive window scale support"
+	default 0
+	---help---
+		Set TCP_RCV_SCALE to the desired scaling factor (shift count in the range of [0..14]).
+		When LWIP_WND_SCALE is enabled but TCP_RCV_SCALE is 0, we can use a large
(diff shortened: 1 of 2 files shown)


os/net/lwip/configs/tcp/Kconfig
```diff
@@ -18,6 +18,23 @@ config NET_TCP_WND
...
+	---help---
+		Set LWIP_WND_SCALE to 1 to enable window scaling.
+
+
+if NET_WND_SCALE
+config NET_TCP_RCV_SCALE
+	int "Receive window scale support"
+	default 0
+	---help---
+		Set TCP_RCV_SCALE to the desired scaling factor (shift count in the range of [0..14]).
```
jeongchanKim, line 33: If this config has value range, how about adding `range a b`?
Then we can prevent invalid value.

os/net/lwip/src/include/lwip/lwipopts.h
```diff
@@ -440,6 +440,14 @@
 #define TCP_WND_UPDATE_THRESHOLD	CONFIG_NET_TCP_WND_UPDATE_THRESHOLD
 #endif
 
+#ifdef CONFIG_NET_WND_SCALE
+#define LWIP_WND_SCALE 1
+#endif
+
+#ifdef CONFIG_NET_TCP_RCV_SCALE
+#define TCP_RCV_SCALE CONFIG_NET_TCP_RCV_SCALE
```
davidfather, line 448: If CONFIG_NET_TCP_RCV_SCALE is not set, what is TCP_RCV_SCALE?

os/net/lwip/configs/tcp/Kconfig
pillip8282: Thanks. It's better to reduce mistake.

os/net/lwip/src/include/lwip/lwipopts.h
pillip8282: it's defined to opt.h. default value is 0

an4967 (approved):
LGTM

PR #5384, opened 2021-12-23
os/board/rtl8721csm: LPWORK to HPWORK in osif timer functions

yeetee179:
Osif timer functions are only used in RTK BLE code. LP work has a lower priority, it will be preempted by others and this will cause the timer to be inaccurate.

### DIFF
--- a/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
+++ b/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
@@ -673,6 +673,9 @@ struct osif_timer_entry {
 _list osif_timer_table;
 bool osif_timer_table_init = 0;
 
+/****************************************************************************/
+/* To be only used in BLE related functions                                 */
+/****************************************************************************/
 void osif_timer_wrapper(void *timer)
 {
 	uint32_t lock;
@@ -698,7 +701,7 @@ void osif_timer_wrapper(void *timer)
 	}
 
 	if (timer_entry->timer->reload) {
-		if (work_queue(LPWORK, timer_entry->timer->work_hdl, osif_timer_wrapper, (void *) (timer_entry->timer), (timer_entry->timer->timeout * TICK_PER_SEC / 1000L)) != OK) {
+		if (work_queue(HPWORK, timer_entry->timer->work_hdl, osif_timer_wrapper, (void *) (timer_entry->timer), (timer_entry->timer->timeout * TICK_PER_SEC / 1000L)) != OK) {
 			dbg("work queue fail \n");
 			timer_entry->timer->live = 0;
 		}
@@ -803,7 +806,7 @@ bool osif_timer_create(void **pp_handle, const char *p_timer_name, uint32_t time
 }
 
 /****************************************************************************/
-/* Start software timer                                                     */
+/* Start software timer, to be only used in BLE related functions           */
 /****************************************************************************/
 bool osif_timer_start(void **pp_handle)
 {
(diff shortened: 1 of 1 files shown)


os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
```diff
@@ -815,7 +815,7 @@ bool osif_timer_start(void **pp_handle)
 	}
 
 	timer = *pp_handle;
-	if (work_queue(LPWORK, timer->work_hdl, osif_timer_wrapper, (void *) (timer), (timer->timeout * TICK_PER_SEC / 1000L)) != OK) {
+	if (work_queue(HPWORK, timer->work_hdl, osif_timer_wrapper, (void *) (timer), (timer->timeout * TICK_PER_SEC / 1000L)) != OK) {
```
an4967, line 818: Is it okay not to change other points?
Such as, _osif_timer_restart_, _osif_timer_wrapper_, _osif_timer_delete_, _osif_timer_stop_

sunghan-chang:
@yeetee179 Could you let us know usages of this timer? If you change it to HPWORK, it will be highest prioritized work. So we should know whether it's most important work or not.

jeongchanKim (commented):
Could you elaborate why you change the workqueue from LPWORK to HPWORK in commit msg?

os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
yeetee179: I have recommited and these places are included.

yeetee179:
> LP work has a lower priority, it will be preempted by others and this will cause the timer to be inaccurate.

yeetee179:
> 

LP work has a lower priority, it will be preempted by others and this will cause the timer to be inaccurate.

yeetee179:
> Could you elaborate why you change the workqueue from LPWORK to HPWORK in commit msg?

LP work has a lower priority, it will be preempted by others and this will cause the timer to be inaccurate.

yeetee179:
> @yeetee179 Could you let us know usages of this timer? If you change it to HPWORK, it will be highest prioritized work. So we should know whether it's most important work or not.

HPWORK need to be used, or else for example in BLE scan if LPWORK is used, the duration of the scan will be inaccurate because the LPWORK timer will be preempted by other tasks.

sunghan-chang:
>> @yeetee179 Could you let us know usages of this timer? If you change it to HPWORK, it will be highest prioritized work. So we should know whether it's most important work or not.
>
> HPWORK need to be used, or else for example in BLE scan if LPWORK is used, the duration of the scan will be inaccurate because the LPWORK timer will be preempted by other tasks.

You mean this will be used only for BLE scan?

yeetee179:
> > > @yeetee179 Could you let us know usages of this timer? If you change it to HPWORK, it will be highest prioritized work. So we should know whether it's most important work or not.
> > 
> > 
> > HPWORK need to be used, or else for example in BLE scan if LPWORK is used, the duration of the scan will be inaccurate because the LPWORK timer will be preempted by other tasks.
> 
> You mean this will be used only for BLE scan?

Yes it is only used in  rtw_ble_client_start_scan_with_filter().

sunghan-chang:
@yeetee179 @an4967 In my opinion, no issue with this PR. But we should use it only for BLE scan and should leave some comment for restrict usage of the timer.

yeetee179:
> @yeetee179 @an4967 In my opinion, no issue with this PR. But we should use it only for BLE scan and should leave some comment for restrict usage of the timer.

I have added the comments required

sangwon03 (approved):
LGTM

PR #5390, opened 2021-12-28
external/stress_tool: add assert_neq

pillip8282:
add assert_neq

### DIFF
--- a/external/include/stress_tool/st_perf.h
+++ b/external/include/stress_tool/st_perf.h
@@ -200,6 +200,15 @@ void perf_set_keeprunning(int enable);
 		}                                                               \
 	} while (0)
 
+#define ST_ASSERT_NEQ(val, exp)                                          \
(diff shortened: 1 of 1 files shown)


external/include/stress_tool/st_perf.h
```diff
@@ -200,6 +200,15 @@ void perf_set_keeprunning(int enable);
 		}                                                               \
 	} while (0)
 
+#define ST_ASSERT_NEQ(val, exp)                                          \
+	do {                                                                \
+		if (exp == val) {                                               \
```
jeongchanKim, line 205: Macro name is NEQ, not equal, but if `exp` and `val` is equal, it makes assert.
Is this intended?

jeongchanKim (approved):
LGTM

an4967 (approved):
LGTM

PR #5393, opened 2021-12-31
security: add new result in security hal

pillip8282:
if user request the key with wrong type then SE returns
HAL_BAD_KEY_TYPE.


os/include/tinyara/security_hal.h
```diff
@@ -36,6 +36,7 @@ typedef enum {
 	HAL_EMPTY_SLOT,
 	HAL_BAD_KEY,	         // only public key can be returned
 	HAL_BAD_KEY_PAIR,      // public and private keys do not match
+  HAL_BAD_KEY_TYPE, // the type when key was set and the type to get key are not matched.
```
sunghan-chang, line 39: Could you check an indentation?

pillip8282: Thanks I'll modify it

an4967 (approved):
LGTM

PR #5397, opened 2022-01-04
arch/arm : Fix Usage fault due to wrong handling of nested interrupts

kishore-sn:
**This patch has been tested several times on development board. 
We have also tested Easysetup, Control test and OTN test on product with following version of ecode repo:
os : v4.0.0-rc.15+220110 (reverted f23d085, c98b6f3 & with #5397 patch)
common : v0.7.1+220110
product : b937ed1 (branch : srid_dev_subham (latest))**

Usage fault occurs during wifi stress test and other tests due to
usage of wrong stack pointer. This is because msp or psp are having wrong
value when returning from exception handler to thread mode. Following
scenarios are causing the issue:

1. Nested irq before executing last instruction of exception handler
2. Nested irq before executing first instruction of exception handler
3. Process stack getting over-written by nested irq hw context saved by HW

This commit makes following changes to exception handler to fix the issues:
1. Disable irq at the start of exception handler (this does not actually fix
   the issue because irq can still occur at the first instruction of exception
   handler. However, this change would prevent any issues of nested irq that
   could happen before the SW context is saved on the stack)
2. If we enter exception.S in handler mode (nested case), save the previous
   PSP value on stack.
3. If we return from exception.S in handler mode (nested case), restore the
   PSP value from stack.
4. While returning from exception.S to thread mode with psp, set the msp to
   point to a backup irq stack. (If not the msp would contain a stale value
   which might cause the process stack to be overwritten by HW context at a
   later time)


os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -420,4 +451,13 @@ g_nestlevel:
 	.size   g_nestlevel, 4
 #endif
 
+	.bss
+	.global	g_bakirqstkalloc
+	.global	g_bakirqstkbase
+	.align	4
+g_bakirqstkalloc:
+	.skip	1024
```
jeongchanKim, line 459: I think the size of nested irq stack can be configurable.
Please make config for this.

os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -420,4 +451,13 @@ g_nestlevel:
 	.size   g_nestlevel, 4
 #endif
 
+	.bss
+	.global	g_bakirqstkalloc
```
jeongchanKim, line 455: This is my guess, `bak` means backup?
How about renaming it as g_nestedintstk?

jeongchanKim, line 455: This is not required for this PR, but how about printing the contents of this stack when asserted?
User stack and irq stack are printed from `up_assert`.

os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -165,23 +196,30 @@ exception_common:
...
+	msrne		psplim, r0
 	stmdb		sp!, {r1}
 #endif
-#ifdef CONFIG_ARMV8M_USEBASEPRI
-	mrs		r3, basepri				/* R3=Current BASEPRI setting */
-#else
-	mrs		r3, primask				/* R3=Current PRIMASK setting */
-#endif
+
+	/* If we are entering in handler mode, and pervious mode was thread mode with psp,
```
jeongchanKim, line 213: Please fix the typo : `and the previous`

jeongchanKim:
Could you check the armv7-m also? 
We have almost same code in armv7-m, so this issue can happen from there.
So if we should fix the issue on v7m also, please apply the changes.

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: Done

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: Done

kishore-sn: Yes I need to add the new stack in few places like stackcheck, etc.. I will do this soon.

os/arch/arm/src/armv8-m/up_exception.S
kishore-sn: Done

os/arch/arm/src/armv8-m/up_exception.S
```diff
@@ -165,23 +200,30 @@ exception_common:
...
+	msrne		psplim, r0
 	stmdb		sp!, {r1}
 #endif
-#ifdef CONFIG_ARMV8M_USEBASEPRI
-	mrs		r3, basepri				/* R3=Current BASEPRI setting */
-#else
-	mrs		r3, primask				/* R3=Current PRIMASK setting */
-#endif
+
+	/* If we are entering in handler mode, and the pervious mode was thread mode with psp,
```
amogh-samsung, line 217: could you please correct "pervious" -> "previous"

sunghan-chang (approved):

PR #5413, opened 2022-01-11
kernel/task/task_prctl.c : Change to get the TCB associated with the …

jeongchanKim:
…PID for all cases

In previous PR_GET/SET_NAME, for pid 0, it gets the current running tcb.
In this case, we cannot get the Idle task's tcb.
So change to always get based on the pid.

### DIFF
--- a/os/kernel/task/task_prctl.c
+++ b/os/kernel/task/task_prctl.c
@@ -124,7 +124,9 @@ int prctl(int option, ...)
 	va_start(ap, option);
 	switch (option) {
 	case PR_SET_NAME:
+	case PR_SET_NAME_BYPID:
 	case PR_GET_NAME:
+	case PR_GET_NAME_BYPID:
 #if CONFIG_TASK_NAME_SIZE > 0
 	{
 		/* Get the prctl arguments */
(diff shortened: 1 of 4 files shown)


os/kernel/task/task_prctl.c
```diff
@@ -133,15 +133,9 @@ int prctl(int option, ...)
...
-		 * pid==0 meaning "this thread")
-		 */
+		/* Get the TCB associated with the PID */
 
-		if (!pid) {
-			tcb = this_task();
-		} else {
-			tcb = sched_gettcb(pid);
-		}
+		tcb = sched_gettcb(pid);
```
jeongarmy, line 138: I think we'd better keep original to get/set name of calling thread at once without pid by calling getpid().

sunghan-chang: Then, how about using `MAX_PID_MASK`? If arg3 is MAX_PID_MASK, then it uses `this_task` to get current running task name.

jeongchanKim: @jeongarmy , As I mentioned in the commit msg, it is better to use its own pid.
@sunghan-chang , using `MAX_PID_MASK` is not good for understanding, I think.

jeongarmy (approved):
Looks good

sunghan-chang (commented):
Could you add and modify os/include/sys/prctl.h for description?

os/kernel/task/task_prctl.c
```diff
@@ -133,14 +135,12 @@ int prctl(int option, ...)
...
 		FAR struct tcb_s *tcb;
 
-		/* Get the TCB associated with the PID (handling the special case of
-		 * pid==0 meaning "this thread")
-		 */
-
-		if (!pid) {
-			tcb = this_task();
-		} else {
+		if (option == PR_SET_NAME_BYPID || option == PR_GET_NAME_BYPID) {
```
sunghan-chang, line 138: Need to add `(option == PR_SET_NAME_BYPID)` in line 166

jeongchanKim: Added. Thanks.

sunghan-chang (approved):

PR #5414, opened 2022-01-12
arch/arm/v7,8-m,r/up_assert.c : Show stack alloc address when asserted

jeongchanKim:
With stack alloc addr, we can get more information if there is an stack overflow.
For example, if there is a heap node corruption, but there is no overflowed prev node, then we can check the next heap node is stack or not.
If next heap node is stack node, then we can make suspicion about stack overflow.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -227,13 +227,13 @@ static void up_taskdump(FAR struct tcb_s *tcb, FAR void *arg)
 	/* Dump interesting properties of this task */
 
 #if CONFIG_TASK_NAME_SIZE > 0
-	lldbg("%*s | %5d | %4d | %7lu / %7lu\n", CONFIG_TASK_NAME_SIZE,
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -227,13 +227,13 @@ static void up_taskdump(FAR struct tcb_s *tcb, FAR void *arg)
 	/* Dump interesting properties of this task */
 
 #if CONFIG_TASK_NAME_SIZE > 0
-	lldbg("%*s | %5d | %4d | %7lu / %7lu\n", CONFIG_TASK_NAME_SIZE,
+	lldbg("%*s | %5d | %4d | %7lu / %7lu | %p\n", CONFIG_TASK_NAME_SIZE,
```
sunghan-chang, line 230: To align with title `STACK ALLOC ADDR`, let's add number like `%10p`.

jeongchanKim: Modified. Thanks.

sunghan-chang (approved):

PR #5422, opened 2022-01-12
os/LibTargets.mk : modify call to Makefile for mm, wqueue and lib

abhishek-samsung:
Currently, the following mm, wqueue and lib are not being recompiled
when changes are made. Modified their respective calls to the Makefile
so that they are recompiled.

### DIFF
--- a/os/LibTargets.mk
+++ b/os/LibTargets.mk
@@ -58,19 +58,19 @@
 # Possible kernel-mode builds
 
 libc$(DELIM)libkc$(LIBEXT): context
-	$(Q) $(MAKE) -C $(LIB_DIR)$(DELIM)libc TOPDIR="$(TOPDIR)" libkc$(LIBEXT) KERNEL=y EXTRADEFINES=$(KDEFINE)
+	$(Q) $(MAKE) -C $(LIB_DIR)$(DELIM)libc TOPDIR="$(TOPDIR)" libkc$(LIBEXT) KERNEL=y EXTRADEFINES=$(KDEFINE) BIN=libkc$(LIBEXT) BINDIR=kbin
 
 $(LIBRARIES_DIR)$(DELIM)libkc$(LIBEXT): libc$(DELIM)libkc$(LIBEXT)
 	$(Q) install $(LIB_DIR)$(DELIM)libc$(DELIM)libkc$(LIBEXT) $(LIBRARIES_DIR)$(DELIM)libkc$(LIBEXT)
 
 mm$(DELIM)libkmm$(LIBEXT): context
-	$(Q) $(MAKE) -C mm TOPDIR="$(TOPDIR)" libkmm$(LIBEXT) KERNEL=y EXTRADEFINES=$(KDEFINE)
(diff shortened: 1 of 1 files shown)


jeongchanKim (commented):
Only those three libs needed? Any other libs are okay?

abhishek-samsung:
> Only those three libs needed? Any other libs are okay?
This was happening for libs that had two pass build. libuarch and libkarch were being compiled without any problem. I found the problem with mm, lib, and wqueue only.

jeongchanKim:
> > Only those three libs needed? Any other libs are okay?
> 
> This was happening for libs that had two pass build. libuarch and libkarch were being compiled without any problem. I found the problem with mm, lib, and wqueue only.

Okay. But arch is also two pass build, I think. libkarch, libuarch.
They are also okay?

abhishek-samsung:
> > > Only those three libs needed? Any other libs are okay?
> > 
> > 
> > This was happening for libs that had two pass build. libuarch and libkarch were being compiled without any problem. I found the problem with mm, lib, and wqueue only.
> 
> Okay. But arch is also two pass build, I think. libkarch, libuarch. They are also okay?

Yes, arch is getting build fine as the rules of libkarch and libuarch are directly available in the Makefile
https://github.com/Samsung/TizenRT/blob/67cc2cb1d43b58c9eac89dca71ee6c2d57f3f994/os/arch/arm/src/Makefile#L189-L193
In the case of mm (wqueue/lib), during 2 pass build, the Makefile is calling itself recursively by passing a couple of more variables. This recursive call can be skipped by providing the additional variables during the first call (LibTargets.mk) itself.
https://github.com/Samsung/TizenRT/blob/67cc2cb1d43b58c9eac89dca71ee6c2d57f3f994/os/mm/Makefile#L110-L122

jeongchanKim:
> > > > Only those three libs needed? Any other libs are okay?
> > > 
> > > 
> > > This was happening for libs that had two pass build. libuarch and libkarch were being compiled without any problem. I found the problem with mm, lib, and wqueue only.
> > 
> > 
> > Okay. But arch is also two pass build, I think. libkarch, libuarch. They are also okay?
> 
> Yes, arch is getting build fine as the rules of libkarch and libuarch are directly available in the Makefile
> 
> https://github.com/Samsung/TizenRT/blob/67cc2cb1d43b58c9eac89dca71ee6c2d57f3f994/os/arch/arm/src/Makefile#L189-L193
> 
> 
> In the case of mm (wqueue/lib), during 2 pass build, the Makefile is calling itself recursively by passing a couple of more variables. This recursive call can be skipped by providing the additional variables during the first call (LibTargets.mk) itself.
> https://github.com/Samsung/TizenRT/blob/67cc2cb1d43b58c9eac89dca71ee6c2d57f3f994/os/mm/Makefile#L110-L122

Thanks for reply.

jeongchanKim (approved):
Looks good to me.

thapav (approved):
lgtm

sunghan-chang (approved):

PR #5426, opened 2022-01-14
fixup! net/lwip: support window scale option

sunghan-chang:
CONFIG_NET_WND_SCALE is a bool type so that it can't have a range syntax.
This causes menuconfig warnings as below:

net/lwip/configs/tcp/Kconfig:23:warning: range is only allowed for int or hex symbols
net/lwip/configs/tcp/Kconfig:23:warning: range is invalid

PR #5377 shows that range was for CONFIG_NET_TCP_RCV_SCALE. And description of that config
also says the same range.
This commit moves it to proper config to resolve the warning.

### DIFF
--- a/os/net/lwip/configs/tcp/Kconfig
+++ b/os/net/lwip/configs/tcp/Kconfig
@@ -20,7 +20,6 @@ config NET_TCP_WND
 
 config NET_WND_SCALE
 	bool "Window scale support"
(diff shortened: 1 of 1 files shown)


jeongchanKim (approved):
LGTM

jsdosa (approved):
looks good to me.

PR #5441, opened 2022-01-19
mm/check_heap_corruption,manage_allocfail : Change dbg to lldbg

jeongchanKim:
In mm_check_heap_corruption,mm_manage_allocfail, information is printed through dbg. So if called from up_assert, logs can be mixed.

### DIFF
--- a/os/drivers/memory/mminfo.c
+++ b/os/drivers/memory/mminfo.c
@@ -141,6 +141,12 @@ static int mminfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		ret = OK;
 		break;
 #endif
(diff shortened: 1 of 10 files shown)


29TB (commented):
LGTM. >.>

29TB (dismissed):
Seems to be properly redefined, as long as no other files use that variable (or whatever), I believe it should be fine. LGTM.

kishore-sn:
Looks good

sangam-swami (dismissed):
Looks fine

sunghan-chang:
@jeongchanKim As we discussed offline, this could not be a solution to prevent log mixing. Let's check flushing and sched lock.

PR #5443, opened 2022-01-20
kernel/binary_manager: Add missing BP recovery update and version update

jeongarmy:
- Add a flag for BP recovery
- Set updated BP data in g_bp_info when BP data is updated
- Update version after header verification is done

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -205,6 +205,7 @@ static int binary_manager_load(int bin_idx)
 				printf("[Binary Loading] Invalid Signature, name : %s, address : %p\n", BIN_NAME(bin_idx), BIN_PARTADDR(bin_idx, (BIN_USEIDX(bin_idx))));
 				if (--bin_count > 0) {
 					BIN_USEIDX(bin_idx) ^= 1;
+					need_update_bp = true;
 					printf("[Binary Loading] Try to read another partition %s\n", GET_PARTNAME(BIN_USEIDX(bin_idx)));
 					continue;
 				} else {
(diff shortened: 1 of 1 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -219,19 +220,22 @@ static int binary_manager_load(int bin_idx)
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 			if (bin_idx == BM_CMNLIB_IDX) {
 				ret = binary_manager_read_header(BINARY_COMMON, devpath, &common_header_data, true);
+				BIN_VER(bin_idx, BIN_USEIDX(bin_idx)) = common_header_data.version;
 			} else
 #endif
 			{
 				ret = binary_manager_read_header(BINARY_USERAPP, devpath, &user_header_data, true);
+				BIN_VER(bin_idx, BIN_USEIDX(bin_idx)) = user_header_data.version;
```
sunghan-chang, line 228: version -> bin_ver

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -219,19 +220,22 @@ static int binary_manager_load(int bin_idx)
...
 			}
 			if (ret == BINMGR_OK) {
 				printf("[Binary Loading] %s Header Checking Success\n", BIN_NAME(bin_idx));
 			} else {
+				BIN_VER(bin_idx, BIN_USEIDX(bin_idx)) = 0;
```
jeongchanKim, line 233: Could you leave the comment why binary version will be set to 0 for invalid case?

os/kernel/binary_manager/binary_manager_load.c
jeongchanKim, line 228: `common_header_data` has a member of `version`, but `user_header_data` has a `bin_ver`.
How about unifying the member variable name?

jeongarmy: I did to distinguish  2 version of kernel and user, `bin_ver` and `kernel_ver` in user binary header.

sunghan-chang (approved):

PR #5444, opened 2022-01-21
os/board/rtl8721csm: add wlan security modes in wifi hdl

gexuyan:


gexuyan:
checkout code fail, will retry

PR #5445, opened 2022-01-21
os/board/rtl8721csm: add wlan security modes in wifi hdl

gexuyan:

PR #5446, opened 2022-01-21
os/board/rtl8721csm: add wlan security modes in wifi hdl [DO NOT MERGE]

jeongarmy:
This is same with #5445. It is temporary PR to resolve CI issue.


j11591 (approved):
LGTM

PR #5448, opened 2022-01-21
os/board/rtl8721csm: Correct Usage of Timer Active Flag

ziyik:
- Correct Usage of Timer Active Flag
- Reduce Error log
- Remove backup variables flagcnt, up_tizen_flag and initial_tizen_flags.
- save_and_cli() and restore_flags() will direct call irqsave() and irqrestore().


os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -94,6 +103,40 @@ static unsigned int __div64_32(u64 *n, unsigned int base)
...
+void _tizenrt_timer_wrapper(void *timer)
+{
+	_list *plist;
+	struct _tizenrt_timer_entry *timer_entry = NULL;
+
+	save_and_cli();
+	plist = get_next(&_tizenrt_timer_table);
+	while ((_tizenrt_end_of_queue_search(&_tizenrt_timer_table, plist)) == _FALSE) {
+		timer_entry = LIST_CONTAINOR(plist, struct _tizenrt_timer_entry, list);
+		if (timer_entry->timer == timer) {
```
jeongchanKim, line 123: Is it okay not to check that `timer_entry` is NULL or not?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -851,13 +894,54 @@ _timerHandle _tizenrt_timerCreate(const signed char *pcTimerName, osdepTickType
...
+
+	struct _tizenrt_timer_entry *timer_entry;
+	timer_entry = _tizenrt_zmalloc(sizeof(struct _tizenrt_timer_entry));
+	if (timer_entry == NULL) {
+		kmm_free(timer->work_hdl);
+		kmm_free(timer);
+		return NULL;
+	}
+	memset(timer_entry, 0, sizeof(struct _tizenrt_timer_entry));
+	timer_entry->timer = timer;
```
jeongchanKim, line 911: `timer_entry` is set to 0 from line 910.
But in here, you used its pointer, `timer_entry->timer`. It looks `NULL->xxx`.
I think it can make assert.
Could you check this?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: It is okay, because timer_entry will be compare with timer.
Which is a pass in pointer by Timer handler, it will not be NULL.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: There is no assert, Line 910 is just initialize the whole structure to 0.
In 911, timer_entry->timer is been assign to timer, which is a pointer.

struct _tizenrt_timer_entry {
	struct list_head list;
	struct timer_list_priv *timer;
};

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -210,7 +253,7 @@ static void _tizenrt_mutex_init(_mutex *pmutex)
 	if (*pmutex == NULL) {
 		*pmutex = _tizenrt_zmalloc(sizeof(pthread_mutex_t));
 		if (*pmutex == NULL) {
-			ndbg("\r\n _tizenrt_mutex_init failed \n");
+			ndbg(" failed \n");
```
pillip8282, line 256: Now ndbg doesn't print file and line information. So please add those information in here. or you can print those information by calling vddbg.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -913,36 +997,38 @@ u32 _tizenrt_timerStop(_timerHandle xTimer, osdepTickType xBlockTime)
...
 
 u32 _tizenrt_timerChangePeriod(_timerHandle xTimer, osdepTickType xNewPeriod, osdepTickType xBlockTime)
 {
 	int ret;
 	struct timer_list_priv *timer = (struct timer_list_priv *)xTimer;
-	ret = work_queue(LPWORK, timer->work_hdl, timer->function, (void *)(timer), xNewPeriod);
+	ret = work_queue(LPWORK, timer->work_hdl, _tizenrt_timer_wrapper, (void *)(timer), xNewPeriod);
 	if (ret == -EALREADY) {
 		if (work_cancel(LPWORK, timer->work_hdl) != OK) {
 			goto cleanup;
```
pillip8282, line 1016: If work_cancel failed then is it ok to free timer->work_hdl which might be still in lpwork?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -284,7 +327,7 @@ static void _tizenrt_mutex_get(_mutex *plock)
 	int temp;
 	temp = sem_wait(*plock);
 	if (temp != 0) {
-		DBG_ERR("_tizenrt_spinlock failed!\n");
+		DBG_ERR(" failed!\n");
```
pillip8282, line 330: How about to use only one DBG_ERR or ndbg? It's better to manage code.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -851,13 +894,54 @@ _timerHandle _tizenrt_timerCreate(const signed char *pcTimerName, osdepTickType
 	timer->timevalue = xTimerPeriodInTicks;
 	timer->data = pvTimerID;
 	timer->function = pxCallbackFunction;
+
+	if (!_tizenrt_timer_table_init) {
```
pillip8282, line 898: _tizenrt_timer_table_init is shared variable. it seems it needs protection.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -851,13 +894,54 @@ _timerHandle _tizenrt_timerCreate(const signed char *pcTimerName, osdepTickType
...
+	if (timer_entry == NULL) {
+		kmm_free(timer->work_hdl);
+		kmm_free(timer);
+		return NULL;
+	}
+	memset(timer_entry, 0, sizeof(struct _tizenrt_timer_entry));
+	timer_entry->timer = timer;
+
+	save_and_cli();
+	list_add(&(timer_entry->list), &_tizenrt_timer_table);
```
pillip8282, line 914: Is it ok not ot protect list_add()?
Following scenario might cause crash please check
```
queue status:  tizenrt_timer_table -> 1 -> 2
1) task A call _tizenrt_timer_Create()
2) task B executes _tizenrt_timerDelete and remove 1 item in the list by list_del_init()
=> context switching pameters of __list_del are (tizenrt_timer_table, 2)
3) task A executes list_add to add 0 item to tizenrt_timer_table 
3-1) 1.prev = 0
3-2) 0.next = tizenrt_timer_table .next (1)
3-3) 0.prev = tizenrt_timer_table 
3-4) tizenrt_timer_table.next = 0
4) context switching task B runs __list_del
4-1) 2.prev = tizenrt_timer_table
4-2) tizenrt_timer_table.next = 2
```
Then item 0 might be missing. please check.

ziyik: save_and_cli() is calling irqsave(), so it will prevent context switching.
So list_add() is actually protected.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: _tizenrt_timer_table_init will only be used once, no other place will alter/modify it.
Does it need protection? if yes, is save_and_cli()(which is irqsave()) good enough?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
pillip8282: Sorry I was confused. It's ok because list_add() is protected

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
pillip8282: For example two tasks can call INIT_LIST_HEAD(). If you think it's ok then you don't need to protect it.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: _tizenrt_timerChangePeriod() will try to change or start the timer, if work_queue() is fail mean timer is already running.
So work_cancel() failed should not happen.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: Will change to DBG_ERR()
#define DBG_ERR(fmt, args...) ndbg("\n\r[%s] " fmt, __FUNCTION__, ##args)

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: Will unifi all to DBG_ERR(), and DBG_INFO()

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -1148,7 +1147,7 @@ void irq_enable(IRQn_Type IrqNum)
 void irq_disable(IRQn_Type IrqNum)
 {
 	if (IrqNum < 0) {
-		rtw_printf("INT %d should not come here\r\n", IrqNum);
+		DBG_INFO("INT %d should not come here \n", IrqNum);
```
sunghan-chang, line 1150: Could you let me know why you need a space at the end of the sentence?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -266,7 +309,7 @@ static void _tizenrt_mutex_free(_mutex *pmutex)
 		if (i == 0) {
 			kmm_free(*pmutex);
 		} else {
-			ndbg("\r\n _tizenrt_mutex_free fail!!! \n");
+			ndbg(" fail!!! \n");
```
sunghan-chang: When do you use `ndbg` and `DBG_ERR`?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -851,13 +894,54 @@ _timerHandle _tizenrt_timerCreate(const signed char *pcTimerName, osdepTickType
...
+	timer_entry = _tizenrt_zmalloc(sizeof(struct _tizenrt_timer_entry));
+	if (timer_entry == NULL) {
+		kmm_free(timer->work_hdl);
+		kmm_free(timer);
+		return NULL;
+	}
+	memset(timer_entry, 0, sizeof(struct _tizenrt_timer_entry));
+	timer_entry->timer = timer;
+
+	save_and_cli();
```
sunghan-chang, line 912: To manage timer list, do you really need interrupt lock?? not like sem or mutex?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: Just want to separate sentence and the next line "\n", if it is not necessary I can remove it.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: All ndbg have been unify to DBG_ERR().

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: We have been using it all along, we dont have intention to change it currently. Any issue?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: It is better to have the protection, it is added now.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -34,26 +34,34 @@
  ******************************************************************************/
 extern void timer_wrapper(_timerHandle timer_hdl);
 
+struct _tizenrt_timer_entry {
+	struct list_head list;
+	struct timer_list_priv *timer;
+};
+
+static _list _tizenrt_timer_table;
+static bool _tizenrt_timer_table_init = 0;
```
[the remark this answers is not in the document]
ziyik: _tizenrt_timer_table_init  is change to u8.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
pillip8282: Thanks for explanation.

Please share your opinion of following scenario
```
taskA executed work_queue() at line 985.
work_queue returned -EALREADY
context switching
lpwork executed timer
context switching
taskA executed work_cancel() and it returned ENOENT because it already had been run
taskA cleanup timer->workhdl => access to invalid memory
```
If  above scenario is wrong, please let me know.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -34,26 +34,34 @@
...
-       }else{
-               initial_tizen_flags = irqsave();
-       }
-       flagcnt++;
+	if(flagcnt){
+		up_tizen_flag = irqsave();
+	}else{
+		initial_tizen_flags = irqsave();
+	}
+	flagcnt++;
```
pillip8282, line 54: Could you explain why did you use both up_tizen_flag and initial_tizen_flag? what is purpose of those value. 
please let me know to understand code better. Thanks in advance.

ziyik: It is for nested situation, first call of save_and_cli() will back up with up_tizen_flag.
The rest will be initial_tizen_flags, when restore_flags() is call only the last will restore with up_tizen_flag.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
sunghan-chang: As you know, irq lock will cause irq missing. Because of irq missing, Realtek optimized irq usages like flash operation. I think this is not more important than flash operation. And there is no reason to use irqlock instead of sem. Could you let me know the reason why this is necessary?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
sunghan-chang: Let's remove it to reduce unnecessary data in uart send.

ziyik: Sure, will remove it on next push.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: I have change irqlock to sem/mutex to avoid irq missing.
By using sem/mutex it will have some drop for uplink t-put about 1Mbits/sec drops.
Because sem/mutex take more processing time, in lib_wlan all timer are using it.
Please take note about his.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: Yes, situation you mention it can happen, the free of timer handler will be remove, it will direct return Fail.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
pillip8282: Thanks for explanation. 
I checked tizenRT options and CONFIG_ARCH_HAVE_NESTED_INTERRUPT is related to this. 
Is there no possibility that CONFIG_ARCH_HAVE_NESTED_INTERRUPT  is disabled in tp1x? 
If so save_and_cli() code seems to be modified depend on the config. 

And second question is protection of flagcnt. 
I think two threads or interrupts can use up_tizen_flag. 
If IRQ B callled save_and_cli() before IRQ A increase flagcnt at line 55 then IRQ B uses up_tizen_flag which IRQ A used. 
Please let me know if I were wrong.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -34,26 +34,35 @@
...
+	struct timer_list_priv *timer;
+};
+
+static _list _tizenrt_timer_table;
+static u8 _tizenrt_timer_table_init = 0;
+static _mutex _tizenrt_timer_mutex = NULL;
+
 static irqstate_t initial_tizen_flags, up_tizen_flag;
 static int flagcnt = 0;
 void save_and_cli()
```
sunghan-chang, line 48: Are they HAL APIs, save_and_cli and restore_flags? I mean, do you need to keep those functions?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
sunghan-chang: @pillip8282 Could you check @ziyik 's comment? Originally it uses irq lock / unlock to manage common resource. But it will causes irq missing so that I asked to use sem or mutex. But he said it will cause wifi throughput performance drop. Please let us know your opinion.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: It is not related to nested interrupt. Nested situation I mention is as below example. 
When Function A call save_and_cli() then after some execution is call Function B, however in Function B it call save_and_cli() again.

For second question, do you suggest that we use mute to protect?

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: Yes, we still need save_and_cli and restore_flags APIs

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
jeongchanKim: @ziyik, I cannot understand why you use two flags.
Could you explain in more detailed?
And even if we need two flags, there is a problem.
Let's assume the case like below.
Task A calls `save_and_cli()`, so tries to do `initial_tizen_flags = irqsave();`.
But before calling `irqsave()`, there is a context switching, so task B calls `save_and_cli()`.
In this case, also `initial_tizen_flag` will be used. (`flagcnt` will be increased to 1 from task B)
After returning to task A, it will use `initial_tizen_flag` again, and increases the `flagcnt` as 2.
And when calling `restore_flags()` from task A, it can successfully restore the irq. (decrease the `flagcnt` to 1)
But if task B tries to call `restore_flags()` again, `flagcnt` is still 1, so it tries to restore the irq with `up_tizen_flag`, not `initial_tizen_flags`.

Could you check this situation and elaborate in detail why we need two flags?

jeongchanKim (commented):
Please elaborate the details in commit msg also.

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
ziyik: The reason of 2 flag is for same Task calling save_and_cli() twice before restore_flags().
When Function A call save_and_cli() then after some execution is call Function B, however in Function B it call save_and_cli() again before restore_flags() is call, the the first flags will keep the initial setting, the second call of save_and_cli() setting will be store in up_tizen_flag.
When execute of Function B is done it will call restore_flags(), which restore the up_tizen_flag setting.
Then follow by Function A call restore_flags(), this time initial_tizen_flags will be restore.

The problem you mention only happen if nested interrupt can be happen, because when Function A call save_and_cli() interrupt have been disable and flagcnt will be increment and decrement after interrupt have been disable.

pillip8282: CONFIG_ARCH_HAVE_NESTED_INTERRUPT=y is enabled in rtl8721csm. If the configuration enable nested interrupt then It means the issue joeengchan mentioned can happen.  Please check.

ziyik: @jeongchanKim @pillip8282 , the main purpose of this function is to prevent other task to interrupt.
Can I use sched_lock() and sched_unlock()?

ziyik: I used mutex for protection.
if sched_lock() and sched_unlock() can be use i will replace it with them.

pillip8282: Sorry for late resposne. I've considered the issue cases that I menntioned. however I couldn't find appropriate solution to resolve it. 
@jeongchanKim please give your opinion about ziyik's solution.

sunghan-chang: @ziyik I think you shouldn't use those HAL APIs. Then there is no problem as above. TizenRT irqsave and irqrestore support nested calling like
1. 1st irqsave
2. 2nd irqsave
3. 1st irqrestore for 2nd irqsave
4. 2nd irqrestore for 1st irqsave

Even if there is context swtching in the middle of steps above, there is no corruption for each other.

sunghan-chang: @pillip8282 CONFIG_ARCH_HAVE_NESTED_INTERRUPT means this architecture can use nested interrupt, but it does not mean it's enabled. CONFIG_ARCH_NESTED_INTERRUPT means it. 
And this code is not related CONFIG_ARCH_NESTED_INTERRUPT. This is interrupt lock in SW. ARCH_NESTED_INTERRUPT is HW interrupt handling.

ziyik: @sunghan-chang , So I just need to take out the last commit will be fine is it?

sunghan-chang: @ziyik No. If you keep your OSAL APIs, save_and_cli and restore_flags, it will have the issue as @jeongchanKim said. I suggested using TizenRT API for them. 
```
save_and_cli -> irqsave
restore_flags -> irqrestore
```
With that, you don't need to worry about nested case.

ziyik: How about the return value of irqsave() and pass in value of irqrestore()?
Caller of save_and_cli() and restore_flags() expect no return value and no pass in value.

sunghan-chang: You need to use TizenRT way without your API. That's why I asked you before like below.
```
[sunghan-chang] [7 days ago] 
Are they OSAL APIs, save_and_cli and restore_flags? I mean, do you need to keep those functions?
```
If not, you need to save the flag for each thread separately. But I don't know how.

sunghan-chang:
@ziyik Any update?

pillip8282 (approved):
LGTM

os/board/rtl8721csm/src/component/os/tizenrt/tizenrt_service.c
```diff
@@ -838,55 +871,81 @@ _timerHandle _tizenrt_timerCreate(const signed char *pcTimerName, osdepTickType
...
+	}
+
+	struct _tizenrt_timer_entry *timer_entry;
+	timer_entry = _tizenrt_zmalloc(sizeof(struct _tizenrt_timer_entry));
+	if (timer_entry == NULL) {
+		kmm_free(timer->work_hdl);
+		kmm_free(timer);
+		return NULL;
+	}
+	memset(timer_entry, 0, sizeof(struct _tizenrt_timer_entry));
```
sunghan-chang, line 904: This is duplicated zero initialization for zmalloc and here. Let's remove this line or use malloc instead of zmalloc.

ziyik: Done modify, thanks

sunghan-chang (approved):

PR #5457, opened 2022-01-25
os/tools: add build_test path in .gitignore

sunghan-chang:
When build test script meets build error, it leaves build logs in build_test folder.
Because it should not be a modification in git, this commit adds build_test path
in .gitignore.
This will remove untracked files in git status like below.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	tools/build_test/


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #5463, opened 2022-01-28
os/arch/arm: Fix build errors due to nested irq stack

kishore-sn:
Nested irq stack is defined only in ARMv8. So, it causes build errors
for other chipsets when used in common code. Fix by adding check for
definition of nested irq stack config.

### DIFF
--- a/os/arch/Kconfig
+++ b/os/arch/Kconfig
@@ -587,7 +587,7 @@ config ARCH_INTERRUPTSTACK
 
 config ARCH_NESTED_IRQ_STACK_SIZE
 	int "Nested IRQ secondary stack size"
-	depends on ARCH_HAVE_INTERRUPTSTACK
+	depends on ARCH_HAVE_INTERRUPTSTACK && ARCH_ARMV8M_FAMILY
 	default 512
 	---help---
 		This is an additional stack which will be used only when nested irq
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -266,19 +266,22 @@ static void up_dumpstate(void)
 	uint32_t istackbase;
 	uint32_t istacksize;
 #endif
+#ifdef CONFIG_ARCH_NESTED_IRQ_STACK_SIZE
 	uint32_t nestirqstkbase;
 	uint32_t nestirqstksize;
+#endif
 
 	/* Get the limits on the user stack memory */
 
 	ustackbase = (uint32_t)rtcb->adj_stack_ptr;
 	ustacksize = (uint32_t)rtcb->adj_stack_size;
+	lldbg("sp:     %08x\n", sp);
 
+#ifdef CONFIG_ARCH_NESTED_IRQ_STACK_SIZE
 	/* Get the limits on the nested irq stack */
 	nestirqstkbase = (uint32_t)&g_nestedirqstkbase;
 	nestirqstksize = (CONFIG_ARCH_NESTED_IRQ_STACK_SIZE & ~3);
 
-	lldbg("sp:     %08x\n", sp);
 	lldbg("Nested IRQ stack:\n");
 	lldbg("  base: %08x\n", nestirqstkbase);
 	lldbg("  size: %08x\n", nestirqstksize);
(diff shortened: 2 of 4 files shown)


abhishek-samsung (commented):
Hi @kishore-sn, if menuconfig is used for other boards, we still get CONFIG_ARCH_NESTED_IRQ_STACK_SIZE in the config. How about adding dependency of the config on armv8 in the kconfig?

kishore-sn:
> Hi @kishore-sn, if menuconfig is used for other boards, we still get CONFIG_ARCH_NESTED_IRQ_STACK_SIZE in the config. How about adding dependency of the config on armv8 in the kconfig?

Done

os/arch/Kconfig
```diff
@@ -587,7 +587,7 @@ config ARCH_INTERRUPTSTACK
 
 config ARCH_NESTED_IRQ_STACK_SIZE
 	int "Nested IRQ secondary stack size"
-	depends on ARCH_HAVE_INTERRUPTSTACK
+	depends on ARCH_HAVE_INTERRUPTSTACK && ARCH_ARMV8M_FAMILY
```
abhishek-samsung, line 590: Hi @kishore-sn, if menuconfig is used for other boards, we still get CONFIG_ARCH_NESTED_IRQ_STACK_SIZE in the config. How about adding dependency of the config on armv8 in the kconfig?

kishore-sn: done

jeongchanKim (approved):
LGTM

sunghan-chang:
@kishore-sn Could you let me know build_test script result?

kishore-sn:
> @kishore-sn Could you let me know build_test script result?

All the configs which were failing previously have passed in build_test. There was a failure in couple of configs, but not related to our changes. 

artik055s.audio 
===========
audio/lib_buffer.c: In function 'apb_reference':
audio/lib_buffer.c:121:33: error: taking address of packed member of 'struct ap_buffer_s' may result in an unaligned pointer value [-Werror=address-of-packed-member]
  121 | #define apb_semgive(b) sem_post(&b->sem)

esp_wrover_kit.hello_with_tash
=====================
sh: 1: xtensa-esp32-elf-gcc: not found
ERROR: xtensa-esp32-elf-gcc failed: 127

sunghan-chang:
@kishore-sn When I tested it last week, there was no error. Could you confirm that was done on dbuild?
You can find `t` option for build_test. I think 2nd failure comes from this.

kishore-sn:
> @kishore-sn When I tested it last week, there was no error. Could you confirm that was done on dbuild? You can find `t` option for build_test. I think 2nd failure comes from this.

Ah.. No.  I had not used dbuild.. I will try again now

kishore-sn:
All passed with dbuild.

==========================================
=========== Build Test Result ===========
==========================================
               artik055s/audio ------ SUCCESS
            artik053/st_things ------ SUCCESS
                   artik053/tc ------ SUCCESS
               qemu/build_test ------ SUCCESS
esp_wrover_kit/hello_with_tash ------ SUCCESS
imxrt1020-evk/loadable_elf_apps ------ SUCCESS
              rtl8721csm/hello ------ SUCCESS
      rtl8721csm/loadable_apps ------ SUCCESS
==========================================

os/arch/Kconfig
sunghan-chang: As far as I know, armv7-m and armv8-m are almost the same. Is it ok to apply only for v8?

sunghan-chang:
@kishore-sn Could you find my comment?
```
As far as I know, armv7-m and armv8-m are almost the same. Is it ok to apply only for v8?
```

kishore-sn:
> @kishore-sn Could you find my comment?
> 
> ```
> As far as I know, armv7-m and armv8-m are almost the same. Is it ok to apply only for v8?
> ```

We have not  been able to test this on armv7m. So, currently only applied on v8.
We will raise a new PR for v7 once we are able to test it.

sunghan-chang (approved):

sunghan-chang:
@kishore-sn i am going to merge this now. But could you enable it by default on defconfig?

kishore-sn:
> @kishore-sn i am going to merge this now. But could you enable it by default on defconfig? In another PR

Currently CONFIG_ARCH_NESTED_IRQ_STACK_SIZE is present in the defconfig for rtk. Once we verify for v7, I will add it in other defconfig also.

PR #5471, opened 2022-02-10
Fix mm_alloc_fail mixed logs

kishore-sn:
- Add ioctl call for mm_alloc_fail in app separation
    - Change some logs from dbg to lldbg
    - Disable irq during mm_alloc_fail
    - Disable irq during assert handler
    Disable irq during lldbg to avoid mixed logs. When irq is disabled,
    then printf and dbg logs will not get printed.

### DIFF
--- a/os/include/debug.h
+++ b/os/include/debug.h
@@ -499,6 +499,12 @@ int get_errno(void);
 #define mllvdbg(...)
 #endif
 
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
+#define mfdbg(format, ...) lldbg(format, ##__VA_ARGS__)
+#else
+#define mfdbg(format, ...) dbg(format, ##__VA_ARGS__)
+#endif
+
 #ifdef CONFIG_DEBUG_NET_ERROR
 #define ndbg(format, ...)    dbg_noarg(format, ##__VA_ARGS__)
 #define nlldbg(format, ...)  lldbg_noarg(format, ##__VA_ARGS__)
@@ -1307,6 +1313,12 @@ int get_errno(void);
 #else
 #define mvdbg       (void)
 #define mllvdbg     (void)
+#endif
+
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
+#define mfdbg lldbg
+#else
+#define mfdbg dbg
 #endif
 
 #ifdef CONFIG_DEBUG_NET_ERROR
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -713,7 +713,18 @@ int mm_check_heap_corruption(struct mm_heap_s *heap);
 #define HEAP_END_IDX   (CONFIG_KMM_NHEAPS - 1)
 
 /* Function to manage the memory allocation failure case. */
+#if defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
+void mm_ioctl_alloc_fail(size_t size);
+#define mm_manage_alloc_fail(h, b, e, s, t) 	do { \
+							(void)h; \
+							(void)b; \
+							(void)e; \
+							(void)t; \
+							mm_ioctl_alloc_fail(s); \
+						} while (0)
+#else
 void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size_t size, int heap_type);
+#endif
 
 #if CONFIG_KMM_NHEAPS > 1
 /**
--- a/os/mm/mm_heap/mm_check_heap_corruption.c
+++ b/os/mm/mm_heap/mm_check_heap_corruption.c
@@ -96,9 +96,9 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 #endif
 
 	if (type == TYPE_CORRUPTED) {
-		dbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
+		mfdbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
 	} else if (type == TYPE_OVERFLOWED) {
-		dbg("OVERFLOWED NODE: addr = 0x%08x size = %u type = %c\n", node, node->size, IS_ALLOCATED_NODE(node) ? 'A' : 'F');
+		mfdbg("OVERFLOWED NODE: addr = 0x%08x size = %u type = %c\n", node, node->size, IS_ALLOCATED_NODE(node) ? 'A' : 'F');
 	}
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	pid_t pid = node->pid;
@@ -112,12 +112,12 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 	}
 #if CONFIG_TASK_NAME_SIZE > 0
 	if (prctl(PR_GET_NAME_BYPID, myname, pid) == OK) {
-		dbg("Node owner pid = %d (%s%s), allocated by code at addr = 0x%08x\n", node->pid, myname, is_stack, node->alloc_call_addr);
+		mfdbg("Node owner pid = %d (%s%s), allocated by code at addr = 0x%08x\n", node->pid, myname, is_stack, node->alloc_call_addr);
 	} else {
-		dbg("Node owner pid = %d (Exited Task%s), allocated by code at addr = 0x%08x\n", node->pid, is_stack, node->alloc_call_addr);
+		mfdbg("Node owner pid = %d (Exited Task%s), allocated by code at addr = 0x%08x\n", node->pid, is_stack, node->alloc_call_addr);
 	}
 #else
-	dbg("Node owner pid = %d%s, allocated by code at addr = 0x%08x\n", node->pid, is_stack, node->alloc_call_addr);
+	mfdbg("Node owner pid = %d%s, allocated by code at addr = 0x%08x\n", node->pid, is_stack, node->alloc_call_addr);
 #endif
 #endif
 }
@@ -173,11 +173,11 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
 
 		for (; node < heap->mm_heapend[region]; prev = node, node = next, next = (struct mm_allocnode_s *)((char *)next + next->size)) {
 			if (prev && prev->size != MM_PREV_NODE_SIZE(node)) {
-				dbg("#########################################################################################\n");
-				dbg("ERROR: Heap node corruption detected\n");
+				mfdbg("#########################################################################################\n");
+				mfdbg("ERROR: Heap node corruption detected\n");
 				dump_node(prev, TYPE_OVERFLOWED);
 				dump_node(node, TYPE_CORRUPTED);
-				dbg("#########################################################################################\n");
+				mfdbg("#########################################################################################\n");
 				if (!aborted
 #if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
 					&& !up_interrupt_context()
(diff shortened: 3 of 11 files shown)


os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -68,6 +83,10 @@
 
 void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size_t size, int heap_type)
 {
+#ifdef __KERNEL__
```
jeongchanKim, line 86: As my check, `__KERNEL__` is defined only when protected build.

kishore-sn: I think it is defined for flat also. 
`KDEFINE = ${shell $(TOPDIR)/tools/define.sh "$(CC)" __KERNEL__}`

It is called in pass2 which is executed for both flat and protected.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -91,7 +110,23 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	}
 #endif
 
+}
+
+#ifdef CONFIG_APP_BINARY_SEPARATION
+void mm_ioctl_alloc_fail(size_t size)
```
sunghan-chang, line 116: mminfo is under CONFIG_MMINFO so that this will not work with disabled config. But this function must be work when CONFIG_MM_ASSERT_ON_FAIL is enabled. Please make it work always with CONFIG_MM_ASSERT_ON_FAIL enabled.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -91,7 +110,23 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+	} else {
+		int res = ioctl(mmfd, MMINFOIOC_MNG_ALLOCFAIL, (int)size);
+		if (res == ERROR) {
+			mdbg("Fail to call mm_manage_allocfail, errno %d\n", get_errno());
+		}
+		close(mmfd);
+	}
 #ifdef CONFIG_MM_ASSERT_ON_FAIL
 	PANIC();
 #endif
```
sunghan-chang, line 130: This should be in mm_manage_alloc_fail function, not here.

os/mm/umm_heap/umm_calloc.c
```diff
@@ -138,7 +138,11 @@ static void *heap_calloc(size_t n, size_t elem_size, int s, int e, size_t caller
 			return ret;
 		}
 	}
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	mm_ioctl_alloc_fail(n * elem_size);
+#else
 	mm_manage_alloc_fail(BASE_HEAP, s, e, n * elem_size, USER_HEAP);
+#endif
```
sunghan-chang, line 145: This looks not good usage. How about redirecting like below?
```
mm.h

#if defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
#define mm_manage_alloc_fail(h, b, e, s, t) do { \
      (void)h; \
      (void)b; \
      (void)e; \
      (void)t; \
      mm_ioctl_alloc_fail(s); \
  } while (0)
```

os/mm/mm_heap/mm_manage_allocfail.c
sunghan-chang: I don't think so.
Flatlibs.mk
```
TINYARALIBS += ... $(LIBRARIES_DIR)$(DELIM)libmm$(LIBEXT)
```
LibTargets.mk
```
mm$(DELIM)libmm$(LIBEXT): context
	$(Q) $(MAKE) -C mm TOPDIR="$(TOPDIR)" libmm$(LIBEXT)

$(LIBRARIES_DIR)$(DELIM)libmm$(LIBEXT): mm$(DELIM)libmm$(LIBEXT)
	$(Q) install mm$(DELIM)libmm$(LIBEXT) $(LIBRARIES_DIR)$(DELIM)libmm$(LIBEXT)
```
There is no `KERNEL=y EXTRADEFINES=$(KDEFINE)`. When `KERNEL=y EXTRADEFINES=$(KDEFINE)` is given, CFLAGS has -D__KERNEL__.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -21,6 +21,19 @@
...
+#undef  CONFIG_DEBUG_ERROR
+#undef  CONFIG_DEBUG_WARN
+#undef  CONFIG_DEBUG_VERBOSE
+#undef  CONFIG_LOGM
+#undef  CONFIG_DEBUG_MM
+#define CONFIG_DEBUG 1
+#define CONFIG_DEBUG_ERROR 1
+#define CONFIG_DEBUG_WARN 1
+#define CONFIG_DEBUG_VERBOSE 1
+#define CONFIG_DEBUG_MM 1
```
sunghan-chang, line 35: To print mdbg, CONFIG_DEBUG_MM_ERROR is also necessary.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,6 +64,12 @@
 #define HEAPINFO_INT INT16_MAX
 #define HEAPINFO_NONSCHED (INT16_MAX - 1)
 
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined (__KERNEL__)
+#define mfdbg lldbg
+#else  /* CONFIG_MM_ASSERT_ON_FAIL */
+#define mfdbg printf
+#endif
```
sunghan-chang, line 71: I don't know why this is necessary. Could you let me know the reason why this is added?

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -96,9 +96,9 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 #endif
 
 	if (type == TYPE_CORRUPTED) {
-		dbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
+		lldbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
```
sunghan-chang, line 99: Let's use mfdbg like others.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -64,6 +64,12 @@
 #define HEAPINFO_INT INT16_MAX
 #define HEAPINFO_NONSCHED (INT16_MAX - 1)
 
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
```
sunghan-chang, line 67: When app separation is enabled and we execute heapinfo command on TASH, this will cause log mix. How about deliver the function from caller? When TASH command is executed, give printf. When assert calls, give lldbg. Is it possible?

sunghan-chang: To printf lldbg like assert, `#define` is necessary.
```
#undef  CONFIG_DEBUG
#undef  CONFIG_DEBUG_ERROR
#undef  CONFIG_DEBUG_WARN
#undef  CONFIG_DEBUG_VERBOSE
#undef  CONFIG_LOGM
#undef  CONFIG_DEBUG_MM
#undef  CONFIG_DEBUG_MM_ERROR
#define CONFIG_DEBUG 1
#define CONFIG_DEBUG_ERROR 1
#define CONFIG_DEBUG_WARN 1
#define CONFIG_DEBUG_VERBOSE 1
#define CONFIG_DEBUG_MM 1
```

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -21,8 +21,27 @@
...
+#undef  CONFIG_DEBUG_ERROR
+#undef  CONFIG_DEBUG_WARN
+#undef  CONFIG_DEBUG_VERBOSE
+#undef  CONFIG_LOGM
+#undef  CONFIG_DEBUG_MM
+#undef  CONFIG_DEBUG_MM_ERROR
+#define CONFIG_DEBUG 1
+#define CONFIG_DEBUG_ERROR 1
+#define CONFIG_DEBUG_WARN 1
+#define CONFIG_DEBUG_VERBOSE 1
```
sunghan-chang, line 35: Because we only use lldbg here, no need to define WARN and VERBOSE. No problem to keep them, but unnecessary.

os/mm/mm_heap/mm_check_heap_corruption.c
sunghan-chang: I want to keep the same policy. Sometimes we use mfdbg, but others lldbg. What I suggest is that `mfdbg` is memory fault debug message. This function is for that. That's why I suggested. How do you think?

os/include/tinyara/mm/mm.h
```diff
@@ -713,7 +713,18 @@ int mm_check_heap_corruption(struct mm_heap_s *heap);
...
+#if defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
+void mm_ioctl_alloc_fail(size_t size);
+#define mm_manage_alloc_fail(h, b, e, s, t) 	do { \
+							(void)h; \
+							(void)b; \
+							(void)e; \
+							(void)t; \
+							mm_ioctl_alloc_fail(s); \
+						} while (0)
+#else
```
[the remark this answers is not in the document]
kishore-sn: For app separation user binary, only mm_ioctl_alloc_fail is required. 
For app separation kernel binary, only mm_manage_alloc_fail is required.
For flat build, only mm_manage_alloc_fail is required.
So, I think this is correct.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
sunghan-chang: Heapinfo command on TASH calls kernel function through ioctl so that it's kernel context.
https://github.com/Samsung/TizenRT/blob/master/apps/system/utils/utils_heapinfo.c#L286
https://github.com/Samsung/TizenRT/blob/master/os/drivers/memory/mminfo.c#L134
I mean this function will print logs using lldbg. Then it will cause log mix.

os/include/debug.h
```diff
@@ -499,6 +499,13 @@ int get_errno(void);
 #define mllvdbg(...)
 #endif
 
+#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
+#define mfdbg lldbg
+#else
+#define mfdbg mdbg
+#endif
```
sunghan-chang, line 506: Could you check format? It's different with above and below. I think you need to add this near line 1311.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -66,8 +60,27 @@
...
+		}
+		close(mmfd);
+	}
+}
+
+#else
+
 void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size_t size, int heap_type)
 {
+	(void)irqsave();
```
sunghan-chang, line 82: There is no irqrestore which is matched with this line. When we disable CONFIG_MM_ASSERT_ON_FAIL, this function won't cause device reset, it's just mm failure. So I think irqrestore is necessary. Could you check?

os/include/debug.h
kishore-sn: Moved.

os/mm/mm_heap/mm_manage_allocfail.c
kishore-sn: Added irqrestore at the the end of function.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: Ok. That's true. 
However, there is no way to avoid this situation. Since it is in kernel side, it will always use lldbg. 
Even the lldbg logs can get mixed with other printf lines. But, if we want to avoid this, we need to disable irqs. I dont think that is a good approach. Please let me know your opinion.

os/include/debug.h
sunghan-chang: You need to keep this here as well with proper format. (also need to add this at line 1311.)

os/mm/mm_heap/mm_heapinfo_parse_heap.c
sunghan-chang: How about my previous suggestion?

> How about deliver the function from caller? When TASH command is executed, give printf. When assert calls, give lldbg. Is it possible?

os/include/debug.h
kishore-sn: Yes Done.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
kishore-sn: It will be difficult. And I dont think it will prevent mixed logs. 
If printf is used, then it will get mixed with logs from other threads.
If lldbg is used, then each line will come clean. But in between the lines, the logs might get mixed with printf.

sunghan-chang: I don't know why printf causes log mixed. Could you let me know the reason?

sunghan-chang: When we use printf / dbg and lldbg at the same time, we can get log mixed. But if we use printf for heapinfo on TASH, there is no lldbg so that I think no log mixed.

kishore-sn:
I have modified heapinfo_parse to use lldbg for assert case and dbg for all other cases.

sunghan-chang:
@kishore-sn Could you confirm test result? reproduce with pr and resolve with pr?

kishore-sn:
> @kishore-sn Could you confirm operation? reproduce without pr and resolve with pr?

Hi
I have checked before and after this change. It is working well now.

sunghan-chang (approved):

PR #5472, opened 2022-02-11
st_perf: fix duplicate execution

pillip8282:
If val is function then val will be called twice because val is
executed for log in printf. If function uses shared variables inside
then return value from it might be differed. So stored function value
to tmp_val and reuse it's value when print log.


an4967 (approved):
LGTM

PR #5474, opened 2022-02-11
apps/examples/webserver_test: Fix keep-alive

hdotkashyap:
Fixed keep-alive to process all keep-alive responses by resetting read_finish

### DIFF
--- a/apps/examples/webserver_test/webserver_test_main.c
+++ b/apps/examples/webserver_test/webserver_test_main.c
@@ -528,7 +528,8 @@ static int ws_test_base(void *arg)
 	}
 
 	send_len = sndlen;
-	for (int i = 0; i <= param->keep_alive_count; i++) {
+	for (int i = 0; i < param->keep_alive_count; i++) {
+		read_finish = false;
 		if (!param->encoding) {
 			buf_len = 0;
 			sndlen = send_len;
@@ -721,6 +722,7 @@ static int ws_test_prepare_request(int argc, char** argv, struct http_client_req
 	FUNC_EN;
 	char *p, *q;
 	memset(request, 0, sizeof(struct http_client_request_t));
+	request->keep_alive_count = 1;
 
 	if (!strncmp(argv[2], "GET", 4)) {
 		request->method = WS_TEST_MODE_GET;


pillip8282 (dismissed):
Could you describe issue scenario more in detail. I seems that read_finish is not updated after parse function. It’s easier to understand the PR if you describe the problem and how to reproduce the problem.

hdotkashyap:
> Could you describe issue scenario more in detail. I seems that read_finish is not updated after parse function. It’s easier to understand the PR if you describe the problem and how to reproduce the problem.

Problem : In webserver_test, the variable **read_finish** becomes _True_ to signify parsing has finished. But when using webserver_test with **keep-alive** it becomes a problem as it does not parse the subsequent responses after the first keep-alive response since **read_finish** is already fixed to _True_.

How to reproduce : `webserver_test 1 GET http://127.0.0.1/ keep-alive=5`
This command will send 5 request in one keep-alive session but will parse only the first response.

 Fix : Inside the keep-alive for loop **read_finish** variable is reset to _False_ for every iteration.

sunghan-chang:
@hdotkashyap Please check our [commit rules](https://github.com/Samsung/TizenRT/wiki/Documentations#commit-rules). Current commit is not acceptable.

sunghan-chang:
> Problem : In webserver_test, the variable read_finish becomes True to signify parsing has finished. But when using webserver_test with keep-alive it becomes a problem as it does not parse the subsequent responses after the first keep-alive response since read_finish is already True.
>
> How to reproduce : webserver_test 1 GET http://127.0.0.1/ keep-alive=5
This command will send 5 GET requests in one keep-alive session but will parse only the first response and skip remaining 4 responses.
>
> Fix : Inside the keep-alive for loop read_finish variable is reset to False for every iteration.

Let's add above in commit descrition.

hdotkashyap:
> > Problem : In webserver_test, the variable read_finish becomes True to signify parsing has finished. But when using webserver_test with keep-alive it becomes a problem as it does not parse the subsequent responses after the first keep-alive response since read_finish is already True.
> > How to reproduce : webserver_test 1 GET http://127.0.0.1/ keep-alive=5
> > This command will send 5 GET requests in one keep-alive session but will parse only the first response and skip remaining 4 responses.
> > Fix : Inside the keep-alive for loop read_finish variable is reset to False for every iteration.
> 
> Let's add above in commit descrition.

Done

pillip8282 (approved):
LGTM
Please check sunghan's comment. This commit violates commit rule of commit title.

hdotkashyap:
> LGTM Please check sunghan's comment. This commit violates commit rule of commit title.

Done

thapav (approved):
May be title of commit could be something liek below:
apps/examples/webserver_test: Fix web server keep-alive test scenario.

vibhor-m:
Code changes looks fine. Please change commit title as suggested by other dev.

amitpurwar:
LGTM

hdotkashyap:
> May be title of commit could be something liek below: apps/examples/webserver_test: Fix web server keep-alive test scenario.

Done

PritamV (commented):
for (int i = 0; i < param->keep_alive_count; i++) 
Just to cross check, variable i should get started from 0 or 1.

hdotkashyap:
> for (int i = 0; i < param->keep_alive_count; i++) Just to cross check, variable i should get started from 0 or 1.

default value for keep_alive_count is 1, so even if keep-alive is OFF loop needs to run 1 iteration..... for that i starts from 0

lkasana:
Looks good to me

pillip8282 (approved):
LGTM

sangwon03 (approved):
LGTM

PR #5475, opened 2022-02-11
os/se/ameba: Link non-secure print to Secure

ziyik:
Allow to print secure log using non-secure print to prevent mixed log

### DIFF
--- a/os/se/ameba/rtl_security_api_wrapper_tz.c
+++ b/os/se/ameba/rtl_security_api_wrapper_tz.c
@@ -17,6 +17,7 @@
 ***************************************************************************************************/
 
 #include <stdint.h>
+#include <stdio.h>
 
 #include "flash_api.h"
 #include "device_lock.h"
@@ -29,6 +30,7 @@ typedef struct {
 	void (*device_unlock)(uint32_t);
 	void (*setstatusbits)(uint32_t);
 	int (*get_random_bytes)(void *, uint32_t);
+	void (*info_printf)(const char *);
 } nsfunc_ops_s;
 
 /* Flash Status Bit */
@@ -79,6 +81,11 @@ static void ns_setstatusbits(u32 NewState)
 	device_mutex_unlock(RT_DEV_LOCK_FLASH);
 }
 
+static void ns_printf(const char *input)
+{
+	printf("%s\n", input);
+}
+
 extern int rtw_get_random_bytes(void *dst, u32 size);
 void *rtl_set_ns_func(void)
 {
@@ -89,6 +96,7 @@ void *rtl_set_ns_func(void)
 	ns_func.device_unlock = device_mutex_unlock;
 	ns_func.setstatusbits = ns_setstatusbits;
 	ns_func.get_random_bytes = rtw_get_random_bytes;
+	ns_func.info_printf = ns_printf;
 
 	return (void *)&ns_func;
 }


sunghan-chang:
@ziyik There is a build break.
```
CC:  ameba/rtl_security_api_wrapper_tz.c
ameba/rtl_security_api_wrapper_tz.c: In function 'ns_printf':
ameba/rtl_security_api_wrapper_tz.c:86:29: error: expected expression before ';' token
  ret = sedbg("%s \n", input);
                             ^
Makefile:73: recipe for target 'ameba/rtl_security_api_wrapper_tz.o' failed
```
When SE debug config is disabled, it becomes `ret = ;`. How about changing ns_printf to `void` type?

sunghan-chang:
@ziyik Any update?

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -79,6 +80,13 @@ static void ns_setstatusbits(u32 NewState)
 	device_mutex_unlock(RT_DEV_LOCK_FLASH);
 }
 
+static int ns_printf(const char *input)
+{
+	int ret;
+	ret = sedbg("%s \n", input);
```
jeongarmy, line 86: I think `return sedbg("%s \n", input);` is enough.

jeongarmy, line 86: please add `#include <debug.h>` for `sedbg`

sunghan-chang: As I left before, when se debug config is disabled, sedbg will be nothing.
```
#ifdef CONFIG_DEBUG_SECURE_ELEMENT_ERROR
#define sedbg(format, ...)     dbg(format, ##__VA_ARGS__)
#define selldbg(format, ...)   lldbg(format, ##__VA_ARGS__)
#else
#define sedbg(...)
#define selldbg(...)
#endif
```
Then `return sedbg(...);` will be `return ;`. I think it also prints build error.

sunghan-chang:
@jeongarmy Please let @ziyik know which API should be used for this. I think sedbg is not enough to print because of the case of debug message config disabled.

sunghan-chang:
@ziyik Could you let us know which message uses this API to be printed?

os/se/ameba/rtl_security_api_wrapper_tz.c
jeongarmy, line 86: `Xdbg` functions have dependency on configuration like CONFIG_DEBUG_ERROR, CONFIG_DEBUG_XXX_ERROR.
I think you'd better to use printf instead of sedbg to print them always.

ziyik: Done

ziyik: I have change it to return void.

ziyik: printf is for application, SE layer is Kernel layer.
I think we should not use printf.

ziyik:
> @ziyik Could you let us know which message uses this API to be printed?

For Samsung to print message is TrustZone, for now it is KEK file

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -79,6 +81,11 @@ static void ns_setstatusbits(u32 NewState)
 	device_mutex_unlock(RT_DEV_LOCK_FLASH);
 }
 
+static void ns_printf(const char *input)
+{
+	sedbg("%s \n", input);
```
sunghan-chang, line 86: Could you remove a space to reduce unnecessary data?
```
sedbg("%s\n", input);
```

os/se/ameba/rtl_security_api_wrapper_tz.c
sunghan-chang: What we print for kek is necessary regardless of debug config so that @jeongarmy asked you to change it to printf. For the same reason, we use some printf in kernel.

sunghan-chang: As of now, we don't have any plan to add debug message more. If you add debug message additionally, (make current one as mandatory message), then we can choose proper way for each message. (But no plan to add another message)

jsdosa (approved):
looks nice to me.

os/se/ameba/rtl_security_api_wrapper_tz.c
ziyik: Understand, I have modify it to use printf.

sunghan-chang: Thank you

os/se/ameba/rtl_security_api_wrapper_tz.c
ziyik: Done modify

sunghan-chang:
@ziyik Once @jeongarmy verifies, I will merge. Thank you for the PR.

ziyik:
> @ziyik Once @jeongarmy verifies, I will merge. Thank you for the PR.

Yes, thanks
@jeongarmy please apply ecode PR15 to verify.

jeongchanKim (approved):
LGTM

jeongarmy (approved):
LGTM

sunghan-chang (approved):

PR #5482, opened 2022-03-01
os/board/rtl8721csm: Fix BLE msg send fail issue

ziyik:
Change msg receiver priority to be same as sender to prevent msg send fail in high load condition
Add prevention check for timer overflow

### DIFF
--- a/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
+++ b/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
@@ -37,8 +37,8 @@ uint32_t osif_sys_time_get(void)
 /****************************************************************************/
 bool osif_sched_start(void)
 {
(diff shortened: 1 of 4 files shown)


os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/board/common/os/tizenrt/osif_tizenrt.c
```diff
@@ -544,11 +544,15 @@ bool osif_msg_send(void *p_handle, void *p_msg, uint32_t wait_ms)
 		return _FAIL;
 	}
 
-	if (osif_task_context_check() == true && wait_ms != 0xFFFFFFFF) {
+	if (osif_task_context_check() == true && wait_ms != 0xFFFFFFFF && wait_ms != 0) {
```
jeongchanKim, line 547: How about checking `wait_ms` first?
This condition should check three cases, but if we check `wait_ms` first, then we can reduce the time if `wait_ms` is 0xFFFFFFFF or 0.

an4967 (approved):
LGTM

PR #5484, opened 2022-03-07
Add API to get largest free node size

jeongchanKim:


os/include/tinyara/mm/mm.h
```diff
@@ -694,6 +694,8 @@ struct mm_heap_s *mm_get_heap_with_index(int index);
 
 int mm_get_heapindex(void *mem);
 
+size_t mm_get_largest_freesize(void *address);
```
sunghan-chang, line 697: how about 'mm_get_largest_freenode_size`?

os/mm/mm_heap/mm_getheap.c
```diff
@@ -259,3 +259,48 @@ int mm_get_heapindex(void *mem)
...
+	struct mm_freenode_s *fnode;
+	int i = 0;
+
+	for(i = 0; i < MM_NNODES; ++i) {
+		for (fnode = heap->mm_nodelist[i].flink; fnode && fnode->size; fnode = fnode->flink) {
+			if (largest_size < fnode->size) {
+				largest_size = fnode->size;
+			}
+		}
+	}
```
sunghan-chang, line 275: This way looks full searching of free node. Is't it possible to check first meet node with descending order?

os/mm/mm_heap/mm_getheap.c
```diff
@@ -259,3 +259,48 @@ int mm_get_heapindex(void *mem)
...
+	return largest_size;
+}
+
+size_t mm_get_largest_freesize(void *address)
```
sunghan-chang, line 279: This file is getheap. How about making a new file for this new functionality?

os/mm/mm_heap/mm_getheap.c
```diff
@@ -259,3 +259,48 @@ int mm_get_heapindex(void *mem)
...
+#else
+	int heap_idx;
+	heap = g_kmmheap;
+	struct mm_heap_s *target_heap;
+	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
+		target_heap = &heap[heap_idx];
+		ret = mm_find_largest_freesize(target_heap);
+		if (ret > largest_size) {
+			largest_size = ret;
+		}
```
sunghan-chang, line 301: Could you let me know the meaning of this if conditional?

sunghan-chang (commented):
Do you have any plan to add getting free size?

os/include/tinyara/mm/mm.h
jeongchanKim: Modified. Thanks.

os/mm/mm_heap/mm_getheap.c
jeongchanKim: Create new file and move them. Thanks.

os/mm/mm_heap/mm_getheap.c
jeongchanKim: Kernel can have multi-heap, so we should check whole heaps to find the largest.
So we can get the largest node size after comparing N-heaps.

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,85 @@
...
+
+#ifdef __KERNEL__
+	size_t ret;
+	int heap_idx;
+	heap = g_kmmheap;
+	struct mm_heap_s *target_heap;
+
+	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
+		target_heap = &heap[heap_idx];
+		ret = mm_find_largest_freesize(target_heap);
```
sunghan-chang, line 70: Let's rename the name from `ret` to `cheap_fsize`.

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,85 @@
...
+	size_t largest_size = 0;
+	struct mm_freenode_s *fnode;
+	int i = 0;
+
+	for (i = 0; i < MM_NNODES; ++i) {
+		fnode = heap->mm_nodelist[i].flink;
+		if (fnode && largest_size < fnode->size) {
+			largest_size = fnode->size;
+		}
+	}
```
sunghan-chang, line 46: Could you check my previous comment? Do we need to search entire node?

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,85 @@
...
+#ifdef __KERNEL__
+	size_t ret;
+	int heap_idx;
+	heap = g_kmmheap;
+	struct mm_heap_s *target_heap;
+
+	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
+		target_heap = &heap[heap_idx];
+		ret = mm_find_largest_freesize(target_heap);
+		if (ret > largest_size) {
```
sunghan-chang, line 71: Could you leave comments why this is different to app heap?

os/mm/mm_heap/mm_getsize.c
sunghan-chang: If you change the code as I said, please leave comments for that.

jeongchanKim: This is not to traverse the entire node, it only traverses the head of each nodelist.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Modifed.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Left the comment. Thanks.

os/mm/mm_heap/mm_getheap.c
jeongchanKim: Modified. Thanks.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Oh if I start with end index, and found the node, then I can decide the largest.
I will fix this.
Thanks.

sunghan-chang: > This way looks full searching of free node. Is't it possible to check first meet node with descending order?

Please find `descending order`. I think if you check it with descending order, you don't need to check next node. Because first meet is largest. Could you check it?

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/****************************************************************************
+ * Name: mm_get_largest_freenode_size
+ *
+ * Description:
+ *   returns a largest node size in the heap
+ ****************************************************************************/
+
+size_t mm_get_largest_freenode_size(void *address)
```
sunghan-chang, line 60: `mm_get_largest_...` and `mm_find_largest_...` looks almost the same operation. I feel `mm_find` returns node and `mm_get` returns the size. But it doesn't.
How about `mm_get_heap_largest_free_...` and `mm_get_space_largest_free...`?
I name that `space` is kernel or user, `heap` is each heap.

os/include/tinyara/mm/mm.h
```diff
@@ -694,6 +694,8 @@ struct mm_heap_s *mm_get_heap_with_index(int index);
 
 int mm_get_heapindex(void *mem);
 
+size_t mm_get_largest_freenode_size(void *address);
```
kishore-sn, line 697: I think its not a good idea to pass address to this API. We just want to know the largest free node in the heap. So, why do we need address? Also, if we want to pass an address, we need to have a previous malloc pointer. So, I think this is not a convenient way to call this API.

sunghan-chang:
Usually umm and kmm have public API and mm has mm internal operation according to the call path. Is it possible to do like that?

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,90 @@
...
+	struct mm_freenode_s *fnode;
+	int i = 0;
+
+	/* Free nodes are sorted in a descending order,
+	 * so the first node in each nodelist is the largest within its nodelist.
+	 */
+	for (i = 0; i < MM_NNODES; ++i) {
+		fnode = heap->mm_nodelist[i].flink;
+		if (fnode && largest_size < fnode->size) {
+			largest_size = fnode->size;
```
kishore-sn, line 47: The node list are also arranged in descending order right?
For example first list contains all freenodes upto 4KB, second list from 4KB to 8KB and so on.
So, if we search the nodelists in the proper order, then we dont need to search all nodelists. We can break as soon as we find the largest one.

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,90 @@
...
+ *   returns a largest node size in the heap
+ ****************************************************************************/
+
+size_t mm_get_largest_freenode_size(void *address)
+{
+	struct mm_heap_s *heap;
+	size_t largest_size = 0;
+
+#ifdef __KERNEL__
+	size_t cheap_fsize;
```
kishore-sn, line 66: The name of this variable is not clear. Why do we call it cheap_fsize? Can we change this to something easily understandable?

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,90 @@
...
+	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
+		target_heap = &heap[heap_idx];
+		cheap_fsize = mm_find_largest_freesize(target_heap);
+		if (cheap_fsize > largest_size) {
+			largest_size = cheap_fsize;
+		}
+	}
+#else
+	/* User can only have single heap. */
+	heap = mm_get_heap(address);
```
kishore-sn, line 81: Instead of using the address, we can directly use BASE_HEAP. I will provide the current app's heap.

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	return largest_size;
+}
+
+/****************************************************************************
+ * Name: mm_get_largest_freenode_size
+ *
+ * Description:
+ *   returns a largest node size in the heap
```
abhishek-samsung, line 57: Please change "returns a largest node size" to "returns the largest free node size"

jeongchanKim:
> Usually umm and kmm have public API and mm has mm internal operation according to the call path. Is it possible to do like that?

I will change another mm call path with another commit. Thanks :)

os/include/tinyara/mm/mm.h
jeongchanKim: Yes, we can get kheap without address, but we don't have any way to get the user heap without address.
Do you have any idea?

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Yes, they are arranged in descending order. So I changed the way how to find.
Thanks.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Modified. Thanks.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Changed to curheap_freesize.
Thanks.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Thanks. Modified.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: Modified. Thanks.

os/mm/mm_heap/mm_getsize.c
jeongchanKim: `mm_find` also returns the size. 
What I meant, `mm_get` returns the size to user(because it is public API), `mm_find` is internal function to find the size.
Many other APIs are `mm_get_xxx` like `mm_get_heap`, so I named like this.
I understand what you are worried, but it is not to get space or heap.

os/mm/mm_heap/mm_getsize.c
```diff
@@ -0,0 +1,92 @@
...
+static size_t mm_find_largest_freesize(struct mm_heap_s *heap)
+{
+	size_t largest_size = 0;
+	struct mm_freenode_s *fnode;
+	int i = 0;
+
+	/* Free nodes are sorted in a descending order,
+	 * so the first node in each nodelist is the largest within its nodelist.
+	 */
+	for (i = MM_NNODES; i > 0; --i) {
```
jeongarmy, line 44: How about `node_idx`?

os/mm/mm_heap/mm_getsize.c
sunghan-chang: > mm_find also returns the size.

This is what I think a problem. `find` returns the free size, but `get` also returns the free size. Then both are the same in naming point of view.

> What I meant, mm_get returns the size to user(because it is public API), mm_find is internal function to find the size.

Then, please represent it in the name. ie. _mm_get_.... and mm_get_...
Because both returns free size (same operation).

> I understand what you are worried, but it is not to get space or heap.

`mm_find` only returns the value of one heap even that space has several heap.
But `mm_get` checks entire heap and returns largest.
So I think `mm_find` could be individual heap checking, and `mm_get` could be entire heap checking.

jeongchanKim: I changed `mm_find_largest_freesize` to `mm_find_largest_freesize_from_specific_heap`.

sunghan-chang: It returns size as well. Why is this `find`? not `get`?  In the same module. naming policy should be the same.

jeongchanKim: Modified. Thanks.

sunghan-chang: I think both find or both get is correct.

jeongchanKim: Changed to get for both.

sunghan-chang (approved):

PR #5490, opened 2022-03-08
os/board/rtl8721csm : fix SRAM heap size to maintain Wi-Fi performance

keerthanalex:
- The \"CONFIG_SRAM_HEAP_SIZE\" is added to all rtk defconfigs.
- Default is fixed to be 109(in KB) for all configs except for \"security_hal_test_tz(28[KB])\" and \"tc(38[KB])\", the value above available heap gives compile error(overflow) for latest TizenRT master[f8bc15b].
- The defconfig value also reflected on \"menuconfig -> Kernel Features -> SRAM Heap size information -> (109) SRAM heap size\" and it depends on \"ARCH_BOARD_RTL8721CSM\".
- The user can modify the macro \"CONFIG_SRAM_HEAP_SIZE\" on defconfig to change SRAM heap size.


sunghan-chang (commented):
This way looks strange. Even you define global variable - srHeap - for that region, you set the same region as first heap.
How about defining new region like BD_RAM_NS_HEAP?
```
BD_RAM_NS (rwx)  : 		ORIGIN = 0x10005000, LENGTH = 0x100xxxxx - 0x10005000	/* MAIN RAM NS: 456K - 109K */
BD_RAM_NS_HEAP (rwx)  : 	ORIGIN = 0x100xxxxx, LENGTH = 0x1007B000 - 0x100xxxxx	/* MAIN HEAP S: 109K */
BD_RAM_S (rwx)  : 		ORIGIN = 0x10077000, LENGTH = 0x1007B000 - 0x10077000	/* MAIN RAM S: 16K */
```
And using start and end address of new region for first heap region setting.

sunghan-chang:
@sangam-swami @kishore-sn Could you review this? This PR is to guarantee kernel heap on SRAM for network performance.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -52,9 +52,14 @@ extern unsigned int __PsramStackLimit;
...
 #define PSRAM_HEAP_LIMIT ((uintptr_t)&__PsramStackLimit)
+#define SRAM_EBSS_SECTION SECTION("._ebss.data")
+
+SRAM_EBSS_SECTION
+static unsigned char srHeap[(CONFIG_SRAM_HEAP_SIZE*1024)+CONFIG_IDLETHREAD_STACKSIZE];
+
 void os_heap_init(void){
-	kregionx_start[0] = (void *)HEAP_BASE;
-	kregionx_size[0] = (size_t)(HEAP_LIMIT - HEAP_BASE);
+	kregionx_start[0] = (void *)(srHeap + CONFIG_IDLETHREAD_STACKSIZE);
```
kishore-sn, line 61: I think you dont need to change kregionx_start, because it is the same value in above case while using HEAP_BASE.

os/board/rtl8721csm/src/component/soc/realtek/amebad/fwlib/ram_hp/rtl8721dhp_app_start.c
```diff
@@ -52,9 +52,14 @@ extern unsigned int __PsramStackLimit;
...
+	kregionx_size[0] = (size_t)(CONFIG_SRAM_HEAP_SIZE*1024);
```
kishore-sn, line 62: It is sufficient to set the required size using CONFIG_SRAM_HEAP_SIZE*1024. This will guarantee that the kernel heap is 109KB.
However, what if 109KB space is not available in the SRAM ? We need to check using HEAP_LIMIT this and if less than 109KB is available, then the size needs to be set accordingly.

build/configs/rtl8721csm/hello/defconfig
```diff
@@ -394,6 +394,11 @@ CONFIG_USERMAIN_STACKSIZE=2048
 CONFIG_PTHREAD_STACK_MIN=256
 CONFIG_PTHREAD_STACK_DEFAULT=2048
 
+#
+#SRAM Heap size information
+#
+CONFIG_SRAM_HEAP_SIZE=109
```
kishore-sn, line 400: Please add this to all rtk defconfigs.

keerthanalex:
> This way looks strange. Even you define global variable - srHeap - for that region, you set the same region as first heap. How about defining new region like BD_RAM_NS_HEAP?
> 
> ```
> BD_RAM_NS (rwx)  : 		ORIGIN = 0x10005000, LENGTH = 0x100xxxxx - 0x10005000	/* MAIN RAM NS: 456K - 109K */
> BD_RAM_NS_HEAP (rwx)  : 	ORIGIN = 0x100xxxxx, LENGTH = 0x1007B000 - 0x100xxxxx	/* MAIN HEAP S: 109K */
> BD_RAM_S (rwx)  : 		ORIGIN = 0x10077000, LENGTH = 0x1007B000 - 0x10077000	/* MAIN RAM S: 16K */
> ```
> 
> And using start and end address of new region for first heap region setting.

@sunghan-chang, we are not planning to define a new region. The "kregionx_start" remains the same as before (HEAP_BASE) with this patch and the size is defined by "CONFIG_SRAM_HEAP_SIZE" which might be adjusted later.

build/configs/rtl8721csm/hello/defconfig
keerthanalex: We will add to other rtk defconfigs once we confirm the configs that needs to be maintained.

build/configs/rtl8721csm/security_hal_test_tz/defconfig
```diff
@@ -388,6 +388,11 @@ CONFIG_USERMAIN_STACKSIZE=2048
 CONFIG_PTHREAD_STACK_MIN=256
 CONFIG_PTHREAD_STACK_DEFAULT=2048
 
+#
+#SRAM Heap size information
+#
+CONFIG_SRAM_HEAP_SIZE=28
```
keerthanalex, line 394: The available heap for this config before patch is "29984". So the "CONFIG_SRAM_HEAP_SIZE" value can only be maximum of (available heap/1024) or else we will get a overflow compile error. Then may need to adjust accordingly.

build/configs/rtl8721csm/tc/defconfig
```diff
@@ -387,6 +387,11 @@ CONFIG_USERMAIN_STACKSIZE=2048
 CONFIG_PTHREAD_STACK_MIN=256
 CONFIG_PTHREAD_STACK_DEFAULT=2048
 
+#
+#SRAM Heap size information
+#
+CONFIG_SRAM_HEAP_SIZE=38
```
keerthanalex, line 393: The available heap for this config before patch is "40192". So the "CONFIG_SRAM_HEAP_SIZE" value can only be maximum of (available heap/1024) or else we will get a overflow compile error. Then may need to adjust accordingly.

sangam-swami:
> > This way looks strange. Even you define global variable - srHeap - for that region, you set the same region as first heap. How about defining new region like BD_RAM_NS_HEAP?
> > ```
> > BD_RAM_NS (rwx)  : 		ORIGIN = 0x10005000, LENGTH = 0x100xxxxx - 0x10005000	/* MAIN RAM NS: 456K - 109K */
> > BD_RAM_NS_HEAP (rwx)  : 	ORIGIN = 0x100xxxxx, LENGTH = 0x1007B000 - 0x100xxxxx	/* MAIN HEAP S: 109K */
> > BD_RAM_S (rwx)  : 		ORIGIN = 0x10077000, LENGTH = 0x1007B000 - 0x10077000	/* MAIN RAM S: 16K */
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And using start and end address of new region for first heap region setting.
> 
> @sunghan-chang, we are not planning to define a new region. The "kregionx_start" remains the same as before (HEAP_BASE) with this patch and the size is defined by "CONFIG_SRAM_HEAP_SIZE" which might be adjusted later.
> 
> Since "CONFIG_SRAM_HEAP_SIZE" may be modified, then we cannot fix the start and end address and define a new region.

I hope to support multiple configs you are using Kconfig instead of adding new region anyway the HEAP size is going to be fixed value for each config

sangam-swami:
Is this PR for maintaining Wi-Fi performance or for improvement?

keerthanalex:
> Is this PR for maintaining Wi-Fi performance or for improvement?

This PR is for maintaining Wi-Fi performance.

PR #5496, opened 2022-03-10
os/kernel: Prevent task and pthread creation from kernel threads

kishore-sn:
If app separation is enabled, then only apps must be allowed to create
task and pthreads. Kernel can only create kthreads. If kernel side code
tries to create task or pthread, it will result in an ASSERT.

### DIFF
--- a/os/binfmt/binfmt_execmodule.c
+++ b/os/binfmt/binfmt_execmodule.c
@@ -150,7 +150,6 @@ static void exec_ctors(FAR void *arg)
 int exec_module(FAR struct binary_s *binp)
 {
 	FAR struct task_tcb_s *newtcb;
-	FAR struct tcb_s *rtcb;
 	FAR uint32_t *stack;
 	pid_t pid;
 	int ret;
@@ -182,8 +181,6 @@ int exec_module(FAR struct binary_s *binp)
 	pointer to the application's mm_heap object. Here we will store the mm_heap
 	pointer to the start of the text section */
 	*(uint32_t *)(binp->sections[BIN_DATA]) = (uint32_t)binp->uheap;
-	rtcb = (struct tcb_s *)sched_self();
-	rtcb->uheap = (uint32_t)binp->uheap;
 
 	/* Allocate a TCB for the new task. */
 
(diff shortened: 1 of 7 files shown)


thapav (dismissed):
LGM

os/kernel/pthread/pthread_create.c
```diff
@@ -260,6 +260,10 @@ int pthread_create(FAR pthread_t *thread, FAR const pthread_attr_t *attr, pthrea
 
 	trace_begin(TTRACE_TAG_TASK, "pthread_create");
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	ASSERT((sched_self()->flags & TCB_FLAG_TTYPE_MASK) != TCB_FLAG_TTYPE_KERNEL);
```
jeongchanKim, line 264: How about moving this before `trace_begin`?
Because we don't need to start `trace` if we will not create the pthread.

kishore-sn: Done

kishore-sn:
I added a new commit for refactoring some old code that was using uheap. Please check it.

os/binfmt/binfmt_execmodule.c
```diff
@@ -272,20 +269,23 @@ int exec_module(FAR struct binary_s *binp)
...
 	newtcb->cmn.uspace = binp->sections[BIN_TEXT] + 4;
 	newtcb->cmn.uheap = (uint32_t)binp->uheap;
+	/* Copy the MPU register values from parent to child task */
+#ifdef CONFIG_ARM_MPU
+	int i = 0;
+#ifdef CONFIG_OPTIMIZE_APP_RELOAD_TIME
+	for (; i < MPU_REG_NUMBER * MPU_NUM_REGIONS; i += MPU_REG_NUMBER)
+#endif
+	{
+		newtcb->mpu_regs[i + MPU_REG_RNR] = binp->mpu_regs[i + MPU_REG_RNR];
```
jeongchanKim, line 283: `newtcb` is `struct task_tcb_s` type, so you should use `newtcb.cmn.mpu_regs` instead of `newtcb.mpu_regs`.

kishore-sn: Fixed

jeongchanKim (approved):
LGTM

jsdosa (approved):
looks nice to me.

PR #5497, opened 2022-03-11
os/tools/*.py : Integrate common util functions

jeongchanKim:
1. There are many duplicated util functions. So integrated.
 1) get_value_from_file() checks that passed argument(target) is in the file or not, and returns the value if exist.
 2) check_config_existence() checks that passed argument(config) is in the file and not starts with '#'.
 3) get_binname_from_bininfo() returns the binary name read from .bininfo

2. Change indentation from tab to 4 spaces
 It is recommended to use Python Coding rule based on PEP 8.
 (PEP 8 : python.org/dev/peps/pep-0008)
 So change the indentations from tab to 4 spaces.

### DIFF
--- a/os/tools/config_util.py
+++ b/os/tools/config_util.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+###########################################################################
+#
+# Copyright 2022 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


os/tools/config_util.py
```diff
@@ -0,0 +1,43 @@
...
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+
+import os
+import sys
+
+os_folder = os.path.dirname(__file__) + '/..'
+
+def get_value_from_file(file_name, target):
```
sunghan-chang, line 25: Because they will be provided as common APIs, we need to explain what they are. Could you add some comments at each API?

jeongchanKim: Added the description, Thanks.

sunghan-chang (approved):

PR #5499, opened 2022-03-15
os/binfmt: Modify caching method for symtab and reltab

kishore-sn:
- If memory not available for allocating complete symtab or reltab,
then use caching instead.
- If memory not available for allocating the configured count of cache
blocks, allocate lesser number of blocks and proceed with loading.
- If cache allocation completely failed, still proceed with loading.
In this case, the symbols are read directly from flash or file system
and load time might be high.
- Use similar approach as above for string table also.


os/binfmt/libelf/libelf_symbols.c
```diff
@@ -213,20 +208,28 @@ int elf_findsymtab(FAR struct elf_loadinfo_s *loadinfo)
...
 {
+	int ret = OK;
 	FAR Elf32_Shdr *symtab = &loadinfo->shdr[loadinfo->symtabidx];
 
 	loadinfo->symtab = (uintptr_t)kmm_malloc(symtab->sh_size);
 
 	if (!loadinfo->symtab) {
 		berr("ERROR: Failed to allocate space for sym table. Size = %u\n", symtab->sh_size);
-		return;
+		ret = -ENOMEM;
```
jeongchanKim, line 220: We can just `return -ENOMEM` instead of setting the local variable.

os/binfmt/libelf/libelf_bind.c
```diff
@@ -96,25 +96,55 @@ static int g_num_lib_syms;
  * Private Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: elf_enable_caching
+ *
+ * Description:
+ *   Enable caching for large tables during relocation.
```
jeongchanKim, line 103: As I know, using cache is only when we cannot allocate whole sym/reloc tables into the RAM.
If correct, could you leave the description that when we use the cache?
Then it is much better to understand.

os/binfmt/libelf/libelf_bind.c
```diff
@@ -170,7 +200,9 @@ static int elf_relocate(FAR struct elf_loadinfo_s *loadinfo, int relidx, FAR con
 	int i;
 
 	/* Read the relocation table into memory */
-	elf_readreltab(loadinfo, relsec);
+	if (elf_readreltab(loadinfo, relsec) == -ENOMEM) {
+		elf_enable_caching(loadinfo);
```
jeongchanKim, line 204: Enabling cache cannot be guaranteed, I think.
How it will be if fail to initialize the cache?

os/binfmt/libelf/libelf_bind.c
```diff
@@ -96,25 +96,57 @@ static int g_num_lib_syms;
...
-static inline void elf_readreltab(FAR struct elf_loadinfo_s *loadinfo, FAR const Elf32_Shdr *relsec)
+static inline int elf_readreltab(FAR struct elf_loadinfo_s *loadinfo, FAR const Elf32_Shdr *relsec)
 {
+	int ret = OK;
 	loadinfo->reltab = (uintptr_t)kmm_malloc(relsec->sh_size);
 
 	if (!loadinfo->reltab) {
 		berr("ERROR: Failed to allocate space for relocation table. Size = %u\n", relsec->sh_size);
-		return;
+		ret = -ENOMEM;
```
jeongchanKim, line 137: Please change `return -ENOMEM` here also :)

os/binfmt/libelf/libelf_cache.c
```diff
@@ -480,8 +485,14 @@ int elf_cache_init(int filfd, uint16_t offset, off_t filelen)
 
 		if (!blockcache[i].out_buffer) {
 			berr("Failed kmm_malloc for blockcache's out_buffer\n");
-			elf_cache_uninit();
-			return -ENOMEM;
+			if (i < 2) {
+				/* We need to have minimum 2 cache blocks for caching to work */
+				elf_cache_uninit();
+				return -ENOMEM;
```
jeongchanKim, line 491: Is it okay not to free the `blockcache`?

os/binfmt/libelf/libelf_bind.c
```diff
@@ -96,25 +96,57 @@ static int g_num_lib_syms;
...
+ *
+ *   Even if there is a failure to allocate cache, we can still try
+ *   to load directly from the flash or file system. This might take
+ *   too much time depending on binary size and number of relocations.
+ *
+ *
+ ****************************************************************************/
+static inline void elf_enable_caching(FAR struct elf_loadinfo_s *loadinfo) {
+#if defined(CONFIG_ELF_CACHE_READ)
+	if (elf_cache_init(loadinfo->filfd, loadinfo->offset, loadinfo->filelen) != OK) {
```
jeongchanKim, line 115: How about checking `cached_read` first?
Because `elf_enable_caching` is called from many where, many times.
So if we check the `cached_read` first, we can skip to init again.
```
if (!loadinfo->cached_read && elf_cache_init(loadinfo->filfd, loadinfo->offset, loadinfo->filelen) != OK) {
```

os/binfmt/libelf/libelf_bind.c
```diff
@@ -378,11 +412,16 @@ int elf_bind(FAR struct elf_loadinfo_s *loadinfo, FAR const struct symtab_s *exp
...
-	elf_readsymtab(loadinfo);
+	if (elf_readsymtab(loadinfo) == -ENOMEM) {
+		elf_enable_caching(loadinfo);
+	}
 
-#ifdef CONFIG_SUPPORT_COMMON_BINARY
-	elf_readstrtab(loadinfo);
 
+	if (elf_readstrtab(loadinfo) == -ENOMEM) {
+		elf_enable_caching(loadinfo);
```
jeongchanKim, line 421: cache is already initialized from line 416.
Is it okay to initialize the cache again?

os/binfmt/libelf/libelf_bind.c
kishore-sn: Done.

os/binfmt/libelf/libelf_bind.c
kishore-sn: Done. Thanks.

jeongchanKim: You left `Done`, but there is no checking for `cache_read`.
Could you check again?

os/binfmt/libelf/libelf_bind.c
jeongchanKim: Okay. Thanks.

os/binfmt/libelf/libelf_bind.c
kishore-sn: I have added the check at the point of reading the tables. I feel this is better. 
Like... `if (!loadinfo->cached_read && elf_readstrtab(loadinfo) == -ENOMEM)`
For example.. if there is failure while reading symtab.. Then we allocate cache.
In this case, we dont need to try for reading strtab and reltab. We will directly use the cache.

jeongchanKim: Okay. Checked. Thanks.

jeongchanKim (approved):
Looks good to me.

PR #5501, opened 2022-03-21
fixup! os/tools/*.py: Integrate common util functions

sunghan-chang:
1. Fix wrong indentation at line 280 in mkbinheader.py
2. Fix wrong value of common binary version
   It should be converted to integer value before value verification.
3. Modify debug messages to know what problem is Exactly


jeongchanKim (approved):
LGTM

PR #5502, opened 2022-03-22
os/Makefile.unix: Refactor method for building app and common binary

kishore-sn:
- loadable_apps/loadable.mk will only build the app binary and copy it to output
- Move the steps for strip, compress and adding header to Makefile.unix
- Apps and commom binary will use same code in Makefile.unix for common steps
- Move the compression code for apps out of mkbinaryheader.py


os/Makefile.unix
```diff
@@ -495,7 +475,56 @@ ifeq ($(CONFIG_RAW_BINARY),y)
...
+
+ifeq ($(CONFIG_COMPRESSED_BINARY),y)
+define COMPRESS_BIN
+	$(Q) cp $(OUTBIN_DIR)/$1 $(OUTBIN_DIR)/$1.uncomp
+	$(Q) tools/compression/mkcompressimg $(CONFIG_COMPRESSION_BLOCK_SIZE) $(CONFIG_COMPRESSION_TYPE) $(OUTBIN_DIR)/$1 $(OUTBIN_DIR)/$1.comp
+	$(Q) mv $(OUTBIN_DIR)/$1.comp $(OUTBIN_DIR)/$1
+endef
+endif
+
+define INSTALL_APP
```
jeongchanKim, line 503: What is the meaning of install?
I think, this is for making binary as samsung format.
So `install` is not easy to understand.

os/Makefile.unix
```diff
@@ -495,7 +475,56 @@ ifeq ($(CONFIG_RAW_BINARY),y)
...
+	$(Q) $(call COMPRESS_BIN,$1)
+	$(Q) $(TOPDIR)/tools/mkbinheader.py $(OUTBIN_DIR)/$1 $2 $3 $1 $4 $5 $6 $7 $8
+	$(Q) $(TOPDIR)/tools/mkchecksum.py $(OUTBIN_DIR)/$1
+	$(Q) cp $(OUTBIN_DIR)/$1 $(OUTBIN_DIR)/user
+endef
+
 post:
+	$(Q) mkdir -p $(OUTBIN_DIR)/user
+ifneq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
+	$(Q) $(call VERIFY_APP,"app1")
```
jeongchanKim, line 516: Could you leave comment why we skip to verify the binary when `CONFIG_SUPPORT_COMMON_BINARY` is enabled?

os/Makefile.unix
kishore-sn: Modified.

jeongchanKim (dismissed):
Thanks for the changes.
All codes related to build are in Makefile.unix, so it is better to maintain and understand.
Approved.

os/Makefile.unix
```diff
@@ -495,7 +475,58 @@ ifeq ($(CONFIG_RAW_BINARY),y)
...
+
+ifeq ($(CONFIG_COMPRESSED_BINARY),y)
+define COMPRESS_BIN
+	$(Q) cp $(OUTBIN_DIR)/$1 $(OUTBIN_DIR)/$1.uncomp
+	$(Q) tools/compression/mkcompressimg $(CONFIG_COMPRESSION_BLOCK_SIZE) $(CONFIG_COMPRESSION_TYPE) $(OUTBIN_DIR)/$1 $(OUTBIN_DIR)/$1.comp
+	$(Q) mv $(OUTBIN_DIR)/$1.comp $(OUTBIN_DIR)/$1
+endef
+endif
+
+define PREPARE_APP
```
sunghan-chang, line 503: Could you add description of each define?

kishore-sn: Done.

sunghan-chang (approved):

PR #5504, opened 2022-03-24
external/webserver: Implement API for chunk transfer from webserver

amitpurwar:
API http_send_response_chunk sends response to client in chunk
format

Signed-off-by: Amit Purwar <amit.purwar@samsung.com>

### DIFF
--- a/apps/examples/webserver/webserver_main.c
+++ b/apps/examples/webserver/webserver_main.c
@@ -68,6 +68,9 @@
 #include <protocols/webserver/http_keyvalue_list.h>
 #include "mbedtls/certs.h"
 
+#define MAX_DATA_SIZE 50
+#define MAX_BUF_SIZE 100
+#define MAX_DATA_COUNT 5
 #define WEBSERVER_STACK_SIZE   (1024 * 8)
 #define WEBSERVER_SCHED_PRI    100
 #define WEBSERVER_SCHED_POLICY SCHED_RR
@@ -99,12 +102,43 @@ static const char g_httpcnlost[] = "close";
 struct http_server_t *http_server = NULL;
 struct http_server_t *https_server = NULL;
 
+static int http_send_chunk(struct http_client_t *client)
+{
+	char msg_body[MAX_DATA_SIZE + 1] = {0, };
+	int ret = 0;
+
+	//add data
+	memset(msg_body, 'a', MAX_DATA_SIZE);
+	msg_body[MAX_DATA_SIZE] = 0;
+
+	ret = http_send_response_chunk(client, 200, "OK", msg_body, NULL, FIRST_DATA);
+	if (ret < 0) {
+		printf("Error: Fail to send FIRST_DATA \n");
+		return ret;
+	}
+
+	for (int i = 0; i < MAX_DATA_COUNT; i++) {
+		ret = http_send_response_chunk(client, 200, "OK", msg_body, NULL, NEXT_DATA);
+		if (ret < 0) {
+			printf("Error: Fail to send NEXT_DATA \n");
+			return ret;
+		}
+	}
+
+	ret = http_send_response_chunk(client, 200, "OK", msg_body, NULL, LAST_DATA);
+	if (ret < 0) {
+		printf("Error: Fail to send LAST_DATA \n");
+		return ret;
+	}
+}
+
 /* GET callbacks */
 void http_get_root(struct http_client_t *client, struct http_req_message *req)
 {
 	struct http_keyvalue_list_t response_headers;
 	const char *msg = "This is a root page";
 	char contlen[6] = { 0, };
+	int resp_type = 1; // 0 - Normal response, 1 - Chunk response
 
 	http_keyvalue_list_init(&response_headers);
 
@@ -114,9 +148,17 @@ void http_get_root(struct http_client_t *client, struct http_req_message *req)
 	http_keyvalue_list_add(&response_headers, g_httpconnect, g_httpcnlost);
 
 	printf("===== GET_ROOT CALLBACK url : %s =====\n", req->url);
-	if (http_send_response(client, 200, msg, &response_headers) < 0) {
-		printf("Error: Fail to send response\n");
+
+	if (resp_type == 1) { // this is temporary change
+		if (http_send_chunk(client) < 0) {
+			printf("Error: Fail to send response\n");
+		}
+	} else {
+		if (http_send_response(client, 200, msg, &response_headers) < 0) {
+			printf("Error: Fail to send response\n");
+		}
 	}
+
 	http_keyvalue_list_release(&response_headers);
 }
 
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -871,6 +871,194 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
 	}
 }
 
+static int prepare_chunk_body(char *buf, int len, const char *body, bool is_last_msg)
+{
+	int chunk_size;
+	char httpcrnl[] = "\r\n";
+	int buflen = 0;
+
+	buflen = len;
+
+	if (is_last_msg) {
+		chunk_size = 0;
+	} else {
+		if (body == NULL) {
+			HTTP_LOGE("msg body is NULL \n");
+			return -1;
+		}
+		chunk_size = strlen(body);
+	}
+
+	//append length of message
+	buflen += snprintf(buf + buflen, HTTP_CONF_MAX_REQUEST_LENGTH - buflen, "%x", chunk_size);
+
+	//append httpcrnl
+	buflen += snprintf(buf + buflen, HTTP_CONF_MAX_REQUEST_LENGTH - buflen, "%s", httpcrnl);
+
+	if (is_last_msg) {
+		//append httpcrnl and no body
+		buflen += snprintf(buf + buflen, HTTP_CONF_MAX_REQUEST_LENGTH - buflen, "%s", httpcrnl);
+		return buflen;
+	}
+
+	if (HTTP_CONF_MAX_REQUEST_LENGTH - buflen - 2 <= chunk_size) {
+		HTTP_LOGE("msg body is larger than buffer can hold \n");
+		return -1;
+	}
+
+
+	//append body
+	buflen += snprintf(buf + buflen, HTTP_CONF_MAX_REQUEST_LENGTH - buflen, "%s", body);
+
+	//append httpcrnl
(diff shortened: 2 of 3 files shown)


external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+static int http_send_chunk_buffer(struct http_client_t *client, char *buf)
+{
+	int  ret = 0;
+	int sndlen = 0;
+	int buflen = 0;
+
+	sndlen = strlen(buf);
+	buflen = 0;
+	while (sndlen > 0) {
+#ifdef CONFIG_NET_SECURITY_TLS
```
PritamV, line 928: Just to cross check client is checked for NULL.
Similarly chk for other required places.

external/webserver/http_client.c
```diff
@@ -1008,6 +1189,8 @@ int http_send_response_helper(struct http_client_t *client, int status, const ch
 	HTTP_FREE(buf);
 	return HTTP_OK;
 }
+
+
```
s1-vivek, line 1193: Please remove this extra line

apps/examples/webserver/webserver_main.c
```diff
@@ -99,12 +102,30 @@ static const char g_httpcnlost[] = "close";
 struct http_server_t *http_server = NULL;
 struct http_server_t *https_server = NULL;
 
+static void http_send_chunk(struct http_client_t *client)
+{
+	char msg_body[MAX_DATA_SIZE + 1] = {0, };
+
+	//add data
+	memset(msg_body, 0, MAX_DATA_SIZE + 1 );
```
jeongchanKim, line 110: `msg_body` is already initialized as 0 from line 107. So we don't need to initialize again.
Even if you want to initialize again, we can do just like below.
Then we can save the time consumption.
```
memset(msg_body, 'a', MAX_DATA_SIZE);
msg_body[MAX_DATA_SIZE] = 0;
``

File: external/webserver/http_client.c [hunk 4]
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+	}
+
+	memset(buf, 0, HTTP_CONF_MAX_REQUEST_LENGTH);
+
+	if (data_type == FIRST_DATA) {
+		//add header and data
+		buflen = snprintf(buf, HTTP_CONF_MAX_REQUEST_LENGTH, "HTTP/1.1 %d %s\r\n",
+					  status, status_message);
+		if (headers) {
+			cur = headers->head->next;
```
jeongchanKim, line 991: Is it guaranteed that `headers->head` is not NULL?

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+			while (cur != headers->tail) {
+				if (strcmp(cur->key, "Content-Length") == 0 || strcmp(cur->key, "Content-Type") == 0
```
jeongchanKim, line 993: How about using `strncmp` instead of `strcmp` for enhanced security?

jeongchanKim (commented):
Could you split the commit as changes and examples?

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+	buflen = len;
+
+	if (is_last_msg) {
+		chunk_size = 0;
+	} else {
+		if (body == NULL) {
+			HTTP_LOGE("msg body is NULL \n");
+			return -1;
+		}
+		chunk_size = strlen(body);
```
jeongchanKim, line 889: Please include `string.h`.

apps/examples/webserver/webserver_main.c
```diff
@@ -114,9 +135,15 @@ void http_get_root(struct http_client_t *client, struct http_req_message *req)
 	http_keyvalue_list_add(&response_headers, g_httpconnect, g_httpcnlost);
 
 	printf("===== GET_ROOT CALLBACK url : %s =====\n", req->url);
-	if (http_send_response(client, 200, msg, &response_headers) < 0) {
-		printf("Error: Fail to send response\n");
+
+	if (resp_type == 1) {
```
sangam-swami, line 139: where is resp_type changing here? or it it for temporary change

apps/examples/webserver/webserver_main.c
```diff
@@ -114,9 +135,15 @@ void http_get_root(struct http_client_t *client, struct http_req_message *req)
...
+		http_send_chunk(client);
```
sangam-swami, line 140: can we handle return value of this function for errors

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+}
+
+static int http_send_chunk_buffer(struct http_client_t *client, char *buf)
+{
+	int  ret = 0;
+	int sndlen = 0;
+	int buflen = 0;
+
+	sndlen = strlen(buf);
+	buflen = 0;
```
sangam-swami, line 926: buflen is initialized, you can remove this line

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+
+	return 0;
+}
+
+static int http_send_chunk(struct http_client_t *client, char *buf, int buflen, const char *body, bool is_last_msg)
+{
+	int ret = 0;
+
+	ret = prepare_chunk_body(buf, buflen, body, is_last_msg);
+	if (ret < 0) {
```
sangam-swami, line 953: we can add error debug messages on failure cases

vibhor-m:
Please add commit message as per guidelines.
File / Folder changed :
Issue:
Solution:

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+		return -1;
+	}
+
+	ret = http_send_chunk_buffer(client, buf);
+	if (ret < 0) {
+		return -1;
```
Taejun-Kwon, line 959: return ret is correct, http_send_chunk_buffer returns defined value, and it seems to be expanded in future...(HTTP_ERROR and HTTP_OK)

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+			const char *body, struct http_keyvalue_list_t *headers, data_type_e data_type)
+{
+	char *buf;
+	int buflen = 0;
+	struct http_keyvalue_t *cur = NULL;
+	int ret = 0;
+
+	if (body == NULL) {
+		HTTP_LOGE("There is empty body \n");
+		return HTTP_ERROR;
```
Taejun-Kwon, line 975: I think you should expand type of error.
Here, you can use HTTP_ERROR_BODY_NULL, and HTTP_ERROR_NO_MEMORY for line 981.

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+		buflen += snprintf(buf + buflen,
+					HTTP_CONF_MAX_REQUEST_LENGTH - buflen, "\r\n");
+	}
+
+
+	// Include response body for chunk
+	ret = http_send_chunk(client, buf, buflen, body, false);
+	if (ret < 0) {
+		HTTP_FREE(buf);
+		return HTTP_ERROR;
```
Taejun-Kwon, line 1036: same as above, error value should be handled in fuction http_send_chunk, and just return ret here.

external/webserver/http_client.c
```diff
@@ -871,6 +871,187 @@ void http_handle_file(struct http_client_t *client, int method, const char *url,
...
+	}
+
+	if (data_type == LAST_DATA) {
```
Taejun-Kwon, line 1039: so, if data_type is LAST_DATA, it should call http_send_chunk twice?(line 1033 and here)

apps/examples/webserver/webserver_main.c
```diff
@@ -99,12 +102,30 @@ static const char g_httpcnlost[] = "close";
...
+{
+	char msg_body[MAX_DATA_SIZE + 1] = {0, };
+
+	//add data
+	memset(msg_body, 0, MAX_DATA_SIZE + 1 );
+	memset(msg_body, 'a', MAX_DATA_SIZE);
+
+	http_send_response_chunk(client, 200, "OK", msg_body, NULL, FIRST_DATA);
+
+	for(int i = 0; i < MAX_DATA_COUNT; i++) {
```
sunghan-chang, line 115: Please use the same policy for `( )` in the same file.
```
for (
```
or
```
for(
```

lkasana, line 115: As per  https://wiki.tizen.org/Native_Platform_Coding_Idiom_and_Style_Guide#C_Coding_Style

> [M03] Keywords have following space rules. 
> • Put a space after (most) keywords
> o example: if, switch, case, for, do, while
> • Do not put a space after function like keywords
> o example: sizeof, typeof, alignof, __attribute__

sunghan-chang: Because this is in the external, we don't need to follow TizenRT coding rules. But need to keep the same policy in the file.

apps/examples/webserver/webserver_main.c
amitpurwar: done

apps/examples/webserver/webserver_main.c
amitpurwar: done

apps/examples/webserver/webserver_main.c
amitpurwar: Temporary Change as per requirement. I will add the comment

external/webserver/http_client.c
amitpurwar: http_send_chunk() are internal functions and they can return different values as per its requirement

external/webserver/http_client.c
amitpurwar: yes, as it sends LAST_DATA and zero byte data

external/webserver/http_client.c
amitpurwar: done

external/webserver/http_client.c
jeongchanKim: It is recommended to include each header into each file which the function used.

amitpurwar: Is this coding guideline mentioned anywhere in TizenRT ?

an4967 (approved):
LGTM

PR #5505, opened 2022-03-25
os/kernel/log_dump.c : Add start and stop functionality

jeongchanKim:
To control the log saving, add start and stop functionality.
Log will be saved by default if log dump config is enabled.

Usage:
  #include <tinyara/log_dump/log_dump.h>
  START_LOGDUMP_SAVE(fd); //To start the log saving
  STOP_LOGDUMP_SAVE(fd);  //To stop the log saving

Example:
TASH>>log_dump

TASH>>This Log Should NOT be saved!!!

log_dump_main: 

*********************   LOG DUMP START  *********************

This Log Should be saved!!!

System Information:

	Version:

		Platform: 3.1	Binary: 200204

	Commit Hash: b008e5f

	Build User: root@jeongchanKim

	Build Time: 2022-03-28 09:50:21 KST

	System Time: 06 Dec 2011, 00:00:00 [s] UTC 

TASH>>Hello, World!!

TASH>>

TASH>>

TASH>>log_dump

TASH>>This Log Should be saved!!!

log_dump_main: 

*********************   LOG DUMP END  ***********************

### DIFF
--- a/apps/examples/log_dump/log_dump_main.c
+++ b/apps/examples/log_dump/log_dump_main.c
@@ -22,7 +22,9 @@
 
 #include <tinyara/config.h>
 #include <stdio.h>
+#include <errno.h>
 #include <fcntl.h>
+#include <tinyara/log_dump/log_dump.h>
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -42,24 +44,52 @@ int log_dump_main(int argc, char *argv[])
 {
 	int ret = 1;
 	char buf[READ_BUFFER_SIZE];
-	int fd = open("/proc/logsave", O_RDONLY);
+	int fd = OPEN_LOGDUMP();
 
 	if (fd < 0) {
 		printf("Please ensure that procfs is mounted\n");
-		return 0;
+		return -1;
 	}
+
+	/* Log dump starts automatically from boot up.
+	 * To test start, intentionally add to stop. */
+	if (STOP_LOGDUMP_SAVE(fd) < 0) {
+		printf("Failed to stop log dump, errno %d\n", get_errno());
+		goto errout;
+	}
+
+	printf("This Log Should NOT be saved!!!\n");
+	sleep(1);
 
 	lldbg("\n*********************   LOG DUMP START  *********************\n");
 
+	if (START_LOGDUMP_SAVE(fd) < 0)	{
+		printf("Failed to start log dump, errno %d\n", get_errno());
+		goto errout;
+	}
+
+	printf("This Log Should be saved!!!\n");
+	sleep(1);
+
+	if (STOP_LOGDUMP_SAVE(fd) < 0) {
+		printf("Failed to stop log dump, errno %d\n", get_errno());
+		goto errout;
+	}
+
 	while (ret > 0) {
-     		ret = read(fd, buf, sizeof(buf));
+		ret = READ_LOGDUMP(fd, buf, sizeof(buf));
 		for (int i = 0; i < ret; i++) {
 			lldbg_noarg("%c", buf[i]);
 		}
 	}
-	close(fd);
 
         lldbg("\n*********************   LOG DUMP END  ***********************\n");
 
+	CLOSE_LOGDUMP(fd);
+
 	return 0;
+
+errout:
+	CLOSE_LOGDUMP(fd);
+	return -1;
 }
--- a/os/fs/procfs/fs_procfslogsave.c
+++ b/os/fs/procfs/fs_procfslogsave.c
@@ -43,6 +43,8 @@
 #include <tinyara/fs/procfs.h>
 #include <tinyara/log_dump/log_dump.h>
 
+#include <tinyara/log_dump/log_dump_internal.h>
+
 #if !defined(CONFIG_DISABLE_MOUNTPOINT) && defined(CONFIG_FS_PROCFS)
 #if defined(CONFIG_LOG_DUMP)
 
@@ -74,6 +76,7 @@ struct logsave_file_s {
 static int logsave_open(FAR struct file *filep, FAR const char *relpath, int oflags, mode_t mode);
 static int logsave_close(FAR struct file *filep);
 static ssize_t logsave_read(FAR struct file *filep, FAR char *buffer, size_t buflen);
+static ssize_t logsave_write(FAR struct file *filep, FAR const char *buffer, size_t buflen);
 static int logsave_dup(FAR const struct file *oldp, FAR struct file *newp);
 static int logsave_stat(FAR const char *relpath, FAR struct stat *buf);
 
@@ -94,7 +97,7 @@ const struct procfs_operations logsave_operations = {
 	logsave_open,				/* open */
 	logsave_close,				/* close */
 	logsave_read,				/* read */
-	NULL,					/* write */
+	logsave_write,				/* write */
 
 	logsave_dup,				/* dup */
 
@@ -139,6 +142,15 @@ static int logsave_close(FAR struct file *filep)
 static ssize_t logsave_read(FAR struct file *filep, FAR char *buffer, size_t buflen)
 {
 	return log_dump_read(buffer, buflen);
+}
+
+/****************************************************************************
+ * Name: logsave_write
+ ****************************************************************************/
+
+static ssize_t logsave_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
+{
+	return log_dump_set(buffer, buflen);
 }
 
 /****************************************************************************
--- a/os/include/tinyara/log_dump/log_dump.h
+++ b/os/include/tinyara/log_dump/log_dump.h
@@ -19,10 +19,15 @@
 /****************************************************************************
  * Included Files
  ****************************************************************************/
+#include <stdio.h>
+#include <fcntl.h>
 
 /********************************************************************************
  * Definitions
  ********************************************************************************/
+
+ #define LOGDUMP_SAVE_START	"1"
+ #define LOGDUMP_SAVE_STOP	"2"
 
 /********************************************************************************
  * Public Types
@@ -32,10 +37,8 @@
  * Public Functions
  ********************************************************************************/
 
-int log_dump_init(void);
-
-void log_dump_read_init(void);
-
-int log_dump_save(char ch);
-
-size_t log_dump_read(FAR char *buffer, size_t buflen);
+#define OPEN_LOGDUMP()			open("/proc/logsave", O_RDWR)
+#define START_LOGDUMP_SAVE(fd)		write(fd, LOGDUMP_SAVE_START, strlen(LOGDUMP_SAVE_START) + 1)
+#define STOP_LOGDUMP_SAVE(fd)		write(fd, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1)
+#define READ_LOGDUMP(fd, buf, bufsize)	read(fd, buf, bufsize)
+#define CLOSE_LOGDUMP(fd)		close(fd)
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -33,6 +33,7 @@
 #include <queue.h>
 
 #include <tinyara/log_dump/log_dump.h>
+#include <tinyara/log_dump/log_dump_internal.h>
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -50,8 +51,11 @@
 #define LOG_DUMP_MEM_FAIL		-1
 #define LOG_DUMP_HEAP_FAIL		-2
 #define LOG_DUMP_SIZE_FAIL		-3
+#define LOG_DUMP_OPT_FAIL		-4
 
 #define LOG_CHUNK_SIZE			sizeof(struct log_dump_chunk_s)
+
+static bool is_started_to_save;
 
 /****************************************************************************
  * Private Type Declarations
@@ -99,6 +103,8 @@ int log_dump_init(void)
 	sq_addfirst((sq_entry_t *)node, &log_dump_chunks);
 	/* set the size of chunks to 1 chunk */
 	log_dump_size = LOG_CHUNK_SIZE;
+	/* Initialize to save the log by default */
+	is_started_to_save = true;
 
 	return LOG_DUMP_OK;
 }
@@ -125,8 +131,26 @@ static void log_dump_mem_check(size_t max_size)
 	}
 }
 
+int log_dump_set(FAR const char *buffer, size_t buflen)
+{
+	(void)buflen;
+
+	if (strncmp(buffer, LOGDUMP_SAVE_START, strlen(LOGDUMP_SAVE_START) + 1) == 0) {
+		is_started_to_save = true;
+	} else if (strncmp(buffer, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) == 0) {
+		is_started_to_save = false;
+	} else {
+		return LOG_DUMP_OPT_FAIL;
+	}
+	return LOG_DUMP_OK;
+}
+
 int log_dump_save(char ch)
 {
+	if (is_started_to_save == false) {
+		return LOG_DUMP_OK;
+	}
+
 	/* need to check if the current chunks size is over max_log_size or greater than x% of free heap */
 
 	size_t free_size;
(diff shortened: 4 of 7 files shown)


os/fs/procfs/fs_procfslogsave.c
```diff
@@ -141,6 +142,16 @@ static ssize_t logsave_read(FAR struct file *filep, FAR char *buffer, size_t buf
 	return log_dump_read(buffer, buflen);
 }
 
+/****************************************************************************
+ * Name: logsave_write
+ ****************************************************************************/
+
+static ssize_t logsave_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
+{
+	log_dump_set_option(buffer, buflen);
```
jeongarmy, line 151: I think `return log_dump_set_option(buffer, buflen);` is right to return result value.

os/include/tinyara/log_dump/log_dump.h
```diff
@@ -24,6 +24,9 @@
  * Definitions
  ********************************************************************************/
 
+ #define LOGDUMP_SAVE_START	"1"
```
jeongarmy, line 27: Is there any reason to define integer as string?
How about changing it to string like `#define LOGDUMP_SAVE_START "logdump_start"`.
Or you can use integer `#define LOGDUMP_SAVE_START 1` and change conditions `if (atoi(buffer) == LOGDUMP_SAVE_START)` in log_dump_set_option

jeongchanKim: `write` uses char arrays, so I want to use string.
comparing log string takes much time, so I don't want to use long string.
And two times of `atoi` also take much time.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +43,20 @@ int log_dump_main(int argc, char *argv[])
...
+	int fd = open("/proc/logsave", O_RDWR);
 
 	if (fd < 0) {
 		printf("Please ensure that procfs is mounted\n");
 		return 0;
 	}
 
+	if (write(fd, LOGDUMP_SAVE_START, strlen(LOGDUMP_SAVE_START) + 1) < 0)
+	{
+		printf("Fail to start log dump, %d\n", ret);
```
jeongarmy, line 55: The result of `write` is not assigned to `ret`. please change `ret` to `errno`

apps/examples/log_dump/log_dump_main.c
```diff
@@ -57,6 +65,14 @@ int log_dump_main(int argc, char *argv[])
 			lldbg_noarg("%c", buf[i]);
 		}
 	}
+
+	if (write(fd, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) < 0)
+	{
+		printf("Fail to stop log dump, %d\n", ret);
```
jeongarmy, line 71: please change `ret` to` errno`.

os/kernel/log_dump/log_dump.c
```diff
@@ -125,8 +130,28 @@ static void log_dump_mem_check(size_t max_size)
...
+	}
+	return LOG_DUMP_OK;
+}
+
+
+
 int log_dump_save(char ch)
 {
+	if (is_started_to_save == false) {
+		return LOG_DUMP_OK;
```
Taejun-Kwon, line 152: In my opinion, saved log should be cleaned here, because user can stop logging before read it.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,20 @@ int log_dump_main(int argc, char *argv[])
...
 	char buf[READ_BUFFER_SIZE];
-	int fd = open("/proc/logsave", O_RDONLY);
+	int fd = open("/proc/logsave", O_RDWR);
 
 	if (fd < 0) {
 		printf("Please ensure that procfs is mounted\n");
 		return 0;
 	}
 
+	if (write(fd, LOGDUMP_SAVE_START, strlen(LOGDUMP_SAVE_START) + 1) < 0)
```
Taejun-Kwon, line 54: below will be more faster.
```
enum SAVE_STATUS_E {
   SAVE_STATUS_START 0,
   SAVE_STATUS_STOP 1
};

...
uint8_t status = SAVE_STATUS_START;
write(fd, &status, sizeof(status));
...
read(fd, &status, sizeof(status));
if (status == SAVE_STATUS_START) {
   do_something;
}
```
And please make sure that procfs leads to read-only file system, so if possible you had better find another way.
But if it takes too many code changes, you can use write, in my opinion.

os/kernel/log_dump/log_dump.c
jeongchanKim: From the requirement, user wants to stop to save the log before read, and log will be saved by default after boot up.
So if we clean the saved log in here, we can miss some logs.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Yes, I'm also worried that procfs is recommended for read-only.
But for usage consistency, I implemented through write operation.
If it is not permitted to use write operation, then I will add new API for controlling the log saving functionality.
Please tell me your opinion @Taejun-Kwon .

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,20 @@ int log_dump_main(int argc, char *argv[])
...
+	{
```
abhishek-samsung, line 55: This example is for printing the saved logs. So if we enable logs just before printing and disable it right after, I don't think we can see any logs in the output. Is this intended?

os/fs/procfs/fs_procfslogsave.c
```diff
@@ -94,7 +95,7 @@ const struct procfs_operations logsave_operations = {
 	logsave_open,				/* open */
 	logsave_close,				/* close */
 	logsave_read,				/* read */
-	NULL,					/* write */
+	logsave_write,					/* write */
```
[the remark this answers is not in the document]
jeongchanKim: Modified. Thanks.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Frankly, the log would be saved by default after boot up.
You can see the [os/kernel/log_dump/log_dump.c](https://github.com/Samsung/TizenRT/pull/5505/files#diff-421a0070940db6407c73bb1b24551470aeb61ba3ddceeaf10e1b52a5782a2274):105.
But to show the usage, I intentionally added start in here.
Anyway, we can see the logs as you expected.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,20 @@ int log_dump_main(int argc, char *argv[])
...
+		printf("Fail to start log dump, errno %d\n", get_errno());
+		close(fd);
+		return 0;
```
sunghan-chang, line 58: How about -1 for failure?

apps/examples/log_dump/log_dump_main.c
sunghan-chang, line 54: The log dump starts automatically. The start is actually restart.
Then, don't it correct to be stop and start?

os/fs/procfs/fs_procfslogsave.c
```diff
@@ -141,6 +142,15 @@ static ssize_t logsave_read(FAR struct file *filep, FAR char *buffer, size_t buf
 	return log_dump_read(buffer, buflen);
 }
 
+/****************************************************************************
+ * Name: logsave_write
+ ****************************************************************************/
+
+static ssize_t logsave_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
+{
+	return log_dump_set_option(buffer, buflen);
```
sunghan-chang, line 151: It's difficult to say "option" for start and stop. How about just `log_dump_set`?

os/include/tinyara/log_dump/log_dump.h
sunghan-chang: Is it impossible to use as integer without string converting?

apps/examples/log_dump/log_dump_main.c
```diff
@@ -57,6 +66,14 @@ int log_dump_main(int argc, char *argv[])
 			lldbg_noarg("%c", buf[i]);
 		}
 	}
+
+	if (write(fd, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) < 0)
```
sunghan-chang, line 70: How about providing some wrapper APIs instead of raw usages of write function like reboot reason?

apps/examples/log_dump/log_dump_main.c
abhishek-samsung: Yes, I too agree. According to the current implementation, we get logs for the first time, but we get no logs afterward. So we need to stop log save, print logs, and then start log save again.

os/include/tinyara/log_dump/log_dump.h
jeongchanKim: Possible, but I think this is better, because no need to convert many times.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Modified. Thanks.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Change the app : Stop first and start again.
Thanks.

os/fs/procfs/fs_procfslogsave.c
jeongchanKim: Modified. Thanks.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Added API like below:
```
START_LOGDUMP_SAVE(fd);
STOP_LOGDUMP_SAVE(fd);
```

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,29 @@ int log_dump_main(int argc, char *argv[])
...
 	if (fd < 0) {
 		printf("Please ensure that procfs is mounted\n");
-		return 0;
+		return -1;
 	}
 
+	/* Log dump starts automatically from boot up.
+	 * To test start, intentionally add to stop. */
+	if (STOP_LOGDUMP_SAVE(fd) < 0)
+	{
```
sunghan-chang, line 57: if ( ) {

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,29 @@ int log_dump_main(int argc, char *argv[])
...
+		goto errout;
+	}
+
+	if (START_LOGDUMP_SAVE(fd) < 0)
+	{
+		printf("Fail to start log dump, errno %d\n", get_errno());
+		goto errout;
+	}
+
+	printf("This Log Should be saved!!!\n");
```
sunghan-chang, line 68: How about adding some log to show not saved log?
```
printf("This log should NOT be saved!!!\n");
if (START_LOGDUMP_SAVE(fd) < 0) {
   ...
}
printf("This log should be saved!!!\n");
...
```

apps/examples/log_dump/log_dump_main.c
```diff
@@ -57,9 +75,20 @@ int log_dump_main(int argc, char *argv[])
 			lldbg_noarg("%c", buf[i]);
 		}
 	}
+
+	if (STOP_LOGDUMP_SAVE(fd) < 0)
+	{
+		printf("Fail to stop log dump, errno %d\n", get_errno());
+		goto errout;
+	}
```
sunghan-chang, line 83: What I understand is that the stop is added to read saved log. But this example doesn't. Could you confirm the purpose of new functions?

os/include/tinyara/log_dump/log_dump.h
```diff
@@ -32,10 +35,15 @@
...
+
 int log_dump_init(void);
 
 void log_dump_read_init(void);
 
 int log_dump_save(char ch);
 
 size_t log_dump_read(FAR char *buffer, size_t buflen);
+
+size_t log_dump_set(FAR const char *buffer, size_t buflen);
```
sunghan-chang, line 49: How about splitting the area for user and for internal usage?

os/kernel/log_dump/log_dump.c
```diff
@@ -125,8 +130,28 @@ static void log_dump_mem_check(size_t max_size)
 	}
 }
 
+size_t log_dump_set(FAR const char *buffer, size_t buflen)
```
sunghan-chang, line 133: The return type looks strange. How about int?

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,29 @@ int log_dump_main(int argc, char *argv[])
 {
 	int ret = 1;
 	char buf[READ_BUFFER_SIZE];
-	int fd = open("/proc/logsave", O_RDONLY);
+	int fd = open("/proc/logsave", O_RDWR);
```
sunghan-chang, line 47: How about making this as definition as well?

os/kernel/log_dump/log_dump.c
```diff
@@ -125,8 +130,28 @@ static void log_dump_mem_check(size_t max_size)
...
+		is_started_to_save = true;
+	} else if (strncmp(buffer, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) == 0) {
+		is_started_to_save = false;
+	} else {
+		return LOG_DUMP_OPT_FAIL;
+	}
+	return LOG_DUMP_OK;
+}
+
+
```
amogh-samsung, line 147: Could we avoid these multiple extra empty lines please?

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,29 @@ int log_dump_main(int argc, char *argv[])
...
 		printf("Please ensure that procfs is mounted\n");
-		return 0;
+		return -1;
 	}
 
+	/* Log dump starts automatically from boot up.
+	 * To test start, intentionally add to stop. */
+	if (STOP_LOGDUMP_SAVE(fd) < 0)
+	{
+		printf("Fail to stop log dump, errno %d\n", get_errno());
```
amogh-samsung, line 58: Could you please change "Fail" to "Failed"

apps/examples/log_dump/log_dump_main.c
```diff
@@ -42,13 +44,29 @@ int log_dump_main(int argc, char *argv[])
...
+		goto errout;
+	}
+
+	if (START_LOGDUMP_SAVE(fd) < 0)
+	{
+		printf("Fail to start log dump, errno %d\n", get_errno());
```
amogh-samsung, line 64: Could you please change "Fail" to "Failed"

apps/examples/log_dump/log_dump_main.c
amogh-samsung, line 47: Suggest you to change the flag for opening to O_RDONLY

```
        /* PROCFS is read-only.  Any attempt to open with any kind of write
	 * access is not permitted.
	 *
	 * REVISIT:  Write-able proc files could be quite useful.
	 */
	if (((oflags & O_WRONLY) != 0 || (oflags & O_RDONLY) == 0) && (!filep->f_inode->u.i_mops->write)) {
		fdbg("ERROR: Only O_RDONLY supported\n");
		return -EACCES;
	}
```

O_RDWR = 0x0002
O_RDONLY = 0x0000
O_WRONLY = 0x0001

We have noticed that the flag O_RDWR often works instead of (O_RDONLY | O_WRONLY) but it is not the same.
If we try to fix this access check on the Procfs side in the future, it might cause problems here.
Please use O_RDONLY instead of you are only wanting to read the file.

jeongchanKim: As discussed in the below, we want to write now.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Modified. Thanks.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Modified.

apps/examples/log_dump/log_dump_main.c
jeongchanKim: Modified to stop before reading.

os/kernel/log_dump/log_dump.c
jeongchanKim: Removed. Thanks.

sunghan-chang (approved):

PR #5511, opened 2022-03-31
Add description about free size

jeongchanKim:


os/include/stdlib.h
```diff
@@ -119,7 +119,10 @@ struct mallinfo {
 	int uordblks;				/* This is the total size of memory occupied by
 								 * chunks handed out by malloc. */
 	int fordblks;				/* This is the total size of memory occupied
-								 * by free (not in use) chunks.*/
+								 * by free (not in use) chunks.
+								 *   NOTE : The available size to allocate is smaller than
+ 								 *    the return of this API, because of the size of alloc node.
+ 								 *   available size = free size - SIZEOF_MM_ALLOCNODE*/
```
sunghan-chang, line 125: Because it already has freenode, it should be `free size - (SIZEOF_MM_ALLOCNODE - SIZEOF_FREENODE)`. Could you check this?

jeongchanKim: > Because it already has freenode, it should be `free size - (SIZEOF_MM_ALLOCNODE - SIZEOF_FREENODE)`. Could you check this?

The first node in the heap is free node, it should be `free_size - (SIZEOF_MM_ALLOCNODE + SIZEOF_MM_FREENODE)`.
I will change like this.
Thanks.

jeongchanKim: For largest free node, the available size can be deducted to `SIZEOF_MM_ALLOCNODE` only.

os/include/stdlib.h
```diff
@@ -115,11 +115,15 @@ struct mallinfo {
 	int arena;					/* This is the total size of memory allocated
 								 * for use by malloc in bytes. */
 	int ordblks;				/* This is the number of free (not in use) chunks */
-	int mxordblk;				/* Size of the largest free (not in use) chunk */
+	int mxordblk;				/* Size of the largest free (not in use) chunk
+								 *   NOTE : The available size to allocate is smaller than
+ 								 *    the return of this API, because of the size of alloc node.
+ 								 *   available size = free size - SIZEOF_MM_ALLOCNODE*/
```
sunghan-chang, line 121: how about largest free size instead of free size? Because there is free size for another value.

jeongchanKim: Modified. Thanks.

sunghan-chang (approved):

PR #5512, opened 2022-04-01
netmgr/netdev_wifi.c: scan count is 0 with no AP

j11591:
-This PR fixes LWNL_EVT_SCAN_FAILED when there is no AP

### DIFF
--- a/os/net/netmgr/netdev_wifi.c
+++ b/os/net/netmgr/netdev_wifi.c
@@ -239,6 +239,10 @@ int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
 	int32_t cnt = 0;
 	int32_t total = 0;
 
+	if (item == NULL) {
+		return 0;
+	}
+
 	while (item) {
 		item = item->next;
 		cnt++;


jeongchanKim:
For failure case, it returns `-1` like line 257, but it returns 0 if `item` is NULL.
If it is failure case, then how about returning `-1` for this case also?

os/net/netmgr/netdev_wifi.c
```diff
@@ -239,6 +239,10 @@ int trwifi_serialize_scaninfo(uint8_t **buffer, trwifi_scan_list_s *scan_list)
 	int32_t cnt = 0;
 	int32_t total = 0;
 
+	if (item == NULL) {
+		return 0;
```
jeongchanKim, line 243: For failure case, it returns -1 like line 257, but it returns 0 if item is NULL.
If it is failure case, then how about returning -1 for this case also?

j11591: If there is no AP, scan result is zero. And it makes item is null.
This case is not fail. The scan is success but result is null.

jeongchanKim (approved):
Looks good to me :)

an4967 (approved):
LGTM

PR #5513, opened 2022-04-05
os/Makefile.unix: Make common changes for app build in all repos

kishore-sn:
Public and private repos use different number of apps. So, while
building the app binaries, check the corresponding config and
build only the required app binaries.

### DIFF
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -519,11 +519,19 @@ post:
 ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
 	$(Q) mkdir -p $(OUTBIN_DIR)/user
 ifneq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
+ifeq ($(CONFIG_APP1_INFO),y)
 	$(Q) $(call VERIFY_APP,"app1") #When common binary is used, undefined symbols of app
+endif
+ifeq ($(CONFIG_APP2_INFO),y)
 	$(Q) $(call VERIFY_APP,"app2") #will be defined in common binary. So dont perform "verify" for apps
+endif
 endif 
+ifeq ($(CONFIG_APP1_INFO),y)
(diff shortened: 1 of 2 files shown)


os/Makefile.unix
```diff
@@ -519,11 +519,25 @@ post:
 ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
 	$(Q) mkdir -p $(OUTBIN_DIR)/user
 ifneq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
+ifeq ($(CONFIG_APP_INFO),y)
+	$(Q) $(call VERIFY_APP,"app")
```
jeongchanKim, line 523: There are also differences between public repo and private repo for using CONFIG_APPX_XXX. For example, in `os/tools/set_bininfo.py`.
Could you integrate them also if you want to use common code for all repos?

os/Makefile.unix
```diff
@@ -519,11 +519,25 @@ post:
 ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
 	$(Q) mkdir -p $(OUTBIN_DIR)/user
 ifneq ($(CONFIG_SUPPORT_COMMON_BINARY),y)
+ifeq ($(CONFIG_APP_INFO),y)
```
sunghan-chang, line 522: There are no APP_INFO configs in any Kconfig. How about adding them?

os/Makefile.unix
kishore-sn: Done.

os/Makefile.unix
kishore-sn: In the ecode repo.. APP_INFO is added instead of APPx_INFO..

sunghan-chang: What I want is unifying the code.
Could you check options below?
1. remove app_info and use app1_info only for ecode
2. make new config to select single app or multi apps
   if single app is selected, app_info, if multi apps, app1_info and app2_info

I prefer 1st option, if impossible, how about 2nd?
I think if we apply 1st or 2nd, we can manage the same code for both.

jeongchanKim: I also think that 1st option is better.

kishore-sn: Yes, I agree with 1st option. 
In that case, is it ok to change ecode so that all places..
app -> app1
APP -> APP1 ?

jeongchanKim: Yes. app1 and APP1 looks okay.

kishore-sn: Ok.. Modified accordingly. 
I will make changes for app->app1 in ecode and raise a separate PR.

jeongchanKim (approved):
Looks good to me.

sunghan-chang (approved):

PR #5518, opened 2022-04-12
Remove unnecessary LZMA compression TC

jeongchanKim:


sunghan-chang (commented):
We don't have any responsibility for external modules like lzma and miniz.
So we don't need to have tc for them. We only need tc for compression layer.
I think
1. remove lmza and miniz main
2. keep compression api tc
3. if we miss tc for any compression public api, we need to add them in the tc.

sunghan-chang (approved):
We don't have any responsibility for external modules like lzma and miniz.
So we don't need to have tc for them. We only need tc for compression layer.
I think
1. remove lmza and miniz main
2. keep compression api tc
3. if we miss tc for any compression public api, we need to add them in the tc.

PR #5525, opened 2022-04-28
drivers/serial: Update log_dump to only store console logs

abhishek-samsung:
Currently log_dump stores all the data that is sent via the serial driver
interface. Updated log_dump to only store the console logs.

### DIFF
--- a/os/drivers/serial/serial.c
+++ b/os/drivers/serial/serial.c
@@ -269,7 +269,11 @@ static int uart_putxmitchar(FAR uart_dev_t *dev, int ch, bool oktoblock)
 		if (nexthead != dev->xmit.tail) {
 			dev->xmit.buffer[dev->xmit.head] = ch;
 #ifdef CONFIG_LOG_DUMP
-			log_dump_save(ch);
+			/* only save UART data that is being transmitted to the console */
+
+			if (dev->isconsole) {
+				log_dump_save(ch);
+			}
 #endif
 			dev->xmit.head = nexthead;
 			return OK;


os/drivers/serial/serial.c
```diff
@@ -269,7 +269,9 @@ static int uart_putxmitchar(FAR uart_dev_t *dev, int ch, bool oktoblock)
 		if (nexthead != dev->xmit.tail) {
 			dev->xmit.buffer[dev->xmit.head] = ch;
 #ifdef CONFIG_LOG_DUMP
-			log_dump_save(ch);
+			if (dev->isconsole) {
+				log_dump_save(ch);
```
jeongchanKim, line 273: Could you leave the comment why you added the if condition?

abhishek-samsung: drivers/serial is common for all the UART-related communication. As a result, earlier all the other data that was output using the UART interface was stored in the log_dump buffer (ex: console output, micom, etc.). Now we are restricting log_dump to only store console data with the condition.

sunghan-chang: > Currently log_dump stores all the logs that are sent via the serial driver
interface. Updated log_dump to only store the console logs.

small question,
can we say "log" for uart data which is not for console?

abhishek-samsung: Thanks, changed "logs" to "data"

sunghan-chang: > Could you leave the comment why you added the if condition?

I think he wants to leave code comment for the reason of if conditional, not pr comment.

sunghan-chang: `// save only logs, not UART data between chips` looks enough.

abhishek-samsung: Done.

sunghan-chang (approved):

PR #5526, opened 2022-04-28
apps: Restructure menuconfig options into sub-menus

amogh-samsung:
- Multiple applications of the same domain will be presented in a separate sub-menu
 under the "Application Configuration" menuconfig tab.
- Added MENUDESC option in Makefiles of subfolders to initiate presentation of
 applications as a separate sub-menu
- Added preconfig options in Makefiles for subfolders to include dependant
 applications in this separate sub-menu.

### DIFF
--- a/apps/examples/performance/Makefile
+++ b/apps/examples/performance/Makefile
@@ -51,4 +51,38 @@
 
 MENUDESC = "Performance"
 
-include $(APPDIR)/Directory.mk
+include $(APPDIR)/Make.defs
+
+SUBDIRS = $(dir $(wildcard */Makefile))
+
+all: nothing
+
+.PHONY: nothing context depend clean distclean
+
+define SDIR_template
+$(1)_$(2):
+	$(Q) $(MAKE) -C $(1) $(2) TOPDIR="$(TOPDIR)" APPDIR="$(APPDIR)" EXTDIR="$(EXTDIR)"
+endef
+
+$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),preconfig)))
+$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),context)))
+$(foreach SDIR, $(SUBDIRS), $(eval $(call SDIR_template,$(SDIR),depend)))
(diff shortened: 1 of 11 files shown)


apps/examples/testcase/Kconfig
```diff
@@ -1,49 +0,0 @@
...
-#
-
-menuconfig EXAMPLES_TESTCASE
-	bool "TestCase Example"
-	default n
-	---help---
-		Enable the TestCase Example
-
-if EXAMPLES_TESTCASE
-config DISABLE_MANUAL_TESTCASE
```
sunghan-chang, line 13: This is a necessary config. Could you add it at somewhere?

apps/examples/testcase/Kconfig_ENTRY
```diff
@@ -1,3 +0,0 @@
-config ENTRY_TESTCASE
-	bool "TestCase Example"
-	depends on EXAMPLES_TESTCASE
```
sunghan-chang, line 3: If you remove this, we can select the testcase using entry selection menu.

apps/examples/testcase/Kconfig
amogh-samsung: The right place to add it seems to be this folder's Kconfig only,
So, changes to this folder have been reverted

apps/examples/testcase/Kconfig_ENTRY
amogh-samsung: Changes to this folder have been reverted

sunghan-chang (dismissed):
Could you confirm menuconfig operation with current PR?

apps/examples/performance/Makefile
```diff
@@ -51,4 +51,38 @@
 
 MENUDESC = "Performance"
 
-include $(APPDIR)/Directory.mk
+include $(APPDIR)/Make.defs
```
amogh-samsung, line 54: Have this Makefile's own set of rues, even if dummy ones, instead of including the APPDIR Directory.mk file for rules

abhishek-samsung (commented):
LGTM

sangam-swami (commented):
Looks good.

sangam-swami:
we can review directories other than app/examples if needed later

sunghan-chang (approved):

PR #5528, opened 2022-05-02
kernel/log_dump : Update getting free heap size for protected build

sangam-swami:
Add umm_get_heap_free_size to support protected build.

Signed-off-by: sangamanatha <sangam.swami@samsung.com>

### DIFF
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -155,7 +155,11 @@ int log_dump_save(char ch)
 
 	size_t free_size;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	free_size = kmm_get_heap_free_size(void);
+#ifdef CONFIG_BUILD_PROTECTED
+	free_size = kmm_get_heap_free_size();
+#else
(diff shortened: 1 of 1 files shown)


os/kernel/log_dump/log_dump.c
```diff
@@ -155,7 +155,11 @@ int log_dump_save(char ch)
 
 	size_t free_size;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	free_size = kmm_get_heap_free_size(void);
+#ifdef CONFIG_BUILD_PROTECTED
+	free_size = kmm_get_heap_free_size();
```
jeongchanKim, line 159: `kmm_get_heap_free_size()` will be the same as `umm_get_heap_free_size()` for flat build.
So I think we don't need to separate this.

For `malloc` case, we can use `kmm_malloc()` from flat build, but `kmm_malloc()` will work properly.

sangam-swami: agree

sangam-swami:
As it is kernel code we use kmm_get_heap_free_size() for both flat and protected build.

PR #5530, opened 2022-05-10
arch/arm/src: Move the loading binary addresses before app heap corru…

thapav:
…ption check

This patch moves the elf_show_all_bin_addr() API call to before the app
heap corruption check in order to retrieve the loading binary addresses
for debug purposes if the device resets during heap corruption.

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -464,6 +464,10 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	up_dumpstate();
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	elf_show_all_bin_section_addr();
+#endif
+
 	lldbg("Checking kernel heap for corruption...\n");
 	if (mm_check_heap_corruption(g_kmmheap) == OK) {
 		lldbg("No kernel heap corruption detected\n");
@@ -478,7 +482,6 @@ void up_assert(const uint8_t *filename, int lineno)
 		if (mm_check_heap_corruption((struct mm_heap_s *)(fault_tcb->uheap)) == OK) {
 			lldbg("No app heap corruption detected\n");
 		}
-		elf_show_all_bin_section_addr();
 	}
 #endif
 	lldbg("Assert location (PC) : %08x\n", asserted_location);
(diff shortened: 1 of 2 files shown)


jeongchanKim (approved):
It is much better to debug even there is a heap corruption in the user heap.

sunghan-chang:
@thapav @jeongchanKim @jeongarmy When kernel heap is corrupted, don't we need app loading information to debug? I mean
don't we need to move it to front of kernel heap checking?

jeongchanKim:
> @thapav @jeongchanKim @jeongarmy When kernel heap is corrupted, don't we need app loading information to debug? I mean don't we need to call it before kernel heap checking?

ELF info is saved into the kernel heap, so if kernel heap is broken, we cannot guarantee the info.
So I think we don't need to show before checking kernel heap.

sunghan-chang:
@jeongchanKim Is it in the heap? Not static memory like bss or data? Could you confirm?

jeongchanKim:
> @jeongchanKim Is it in the heap? Not static memory like bss or data? Could you confirm?

Yes, it is in the kernel heap.

jeongchanKim:
> > @jeongchanKim Is it in the heap? Not static memory like bss or data? Could you confirm?
> 
> Yes, it is in the kernel heap.

The list is global variable, but each info is dynamically allocated and appended in the list.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -464,6 +464,8 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	up_dumpstate();
 
+	elf_show_all_bin_addr();
```
jeongchanKim, line 467: Could you check the build break?
You can see the CI build log.
Maybe it is related to config.

sunghan-chang:
@thapav Could you update this PR for rebasing with Ahreum's change and for build break?

thapav:
> @thapav Could you update this PR for rebasing with Ahreum's change and for build break?

Done

os/arch/arm/src/armv7-m/up_assert.c
thapav: Rectified

sunghan-chang (approved):

jeongchanKim (approved):
Looks good to me.

PR #5531, opened 2022-05-11
binfmt: save binary section addresses in static memory for debugging

jeongarmy:
save binary section address information, g_bin_addr_list in static memory for debugging when kernel heap corruption.

### DIFF
--- a/os/binfmt/libelf/libelf_sections.c
+++ b/os/binfmt/libelf/libelf_sections.c
@@ -65,13 +65,16 @@
 #include <tinyara/kmalloc.h>
 #include <tinyara/binfmt/elf.h>
 
-#include <queue.h>
 #include <string.h>
 #include <tinyara/binfmt/binfmt.h>
 #include <binary_manager/binary_manager.h>
 
 #include "libelf.h"
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
+/* The list for a common binary and user binaries(CONFIG_NUM_APPS) */
+static bin_addr_info_t g_bin_addr_list[CONFIG_NUM_APPS + 1];
+#endif
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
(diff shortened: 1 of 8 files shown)


os/binfmt/libelf/libelf_sections.c
```diff
@@ -208,64 +209,46 @@ void *elf_find_start_section_addr(struct binary_s *binp)
...
+	binfo("   bss_addr    : %x\n", g_bin_addr_list[bin_idx].bss_addr);
 #endif
 #endif
-		sq_addlast((sq_entry_t *)bin_info, &g_bin_addr_list);
-	} else {
-		berr("ERROR : Failed to allocate for bin symbol.\n");
-	}
 }
 
 void elf_delete_bin_section_addr(struct binary_s *bin)
```
jeongchanKim, line 238: In this case, we don't need to pass the whole `bin`.
It is better to pass the `bin->binary_idx` only.
Please check this.

os/binfmt/libelf/libelf_sections.c
```diff
@@ -65,13 +65,14 @@
...
 #include <tinyara/binfmt/elf.h>
 
-#include <queue.h>
 #include <string.h>
 #include <tinyara/binfmt/binfmt.h>
 #include <binary_manager/binary_manager.h>
 
 #include "libelf.h"
 
+static bin_addr_info_t g_bin_addr_list[CONFIG_NUM_APPS + 1];
```
sunghan-chang, line 74: Could you let me know why one more additional array size is necessary? `+ 1`?

os/binfmt/libelf/libelf_sections.c
```diff
@@ -208,64 +209,46 @@ void *elf_find_start_section_addr(struct binary_s *binp)
...
 
 void elf_save_bin_section_addr(struct binary_s *bin)
 {
-	bin_addr_info_t *bin_info;
-	bin_info = (bin_addr_info_t *)kmm_malloc(sizeof(bin_addr_info_t));
-	if (bin_info != NULL) {
-		bin_info->bin_idx = bin->binary_idx;
-		bin_info->text_addr = bin->sections[BIN_TEXT];
-		bin_info->text_size = bin->sizes[BIN_TEXT];
+	uint8_t bin_idx = bin->binary_idx;
```
jeongchanKim, line 220: It is better to check that `bin` is NULL or not.
Even though it never be NULL.

os/binfmt/libelf/libelf_sections.c
jeongchanKim, line 74: I think it is enough to have the array with `CONFIG_NUM_APPS` size.

jeongchanKim, line 74: How about defining this only when loadable?

jeongarmy: It needs additional space for common binary.

jeongarmy: It needs additional space for common binary.

jeongarmy: Right. I added dependency of CONFIG_APP_BINARY_SEPARATION.

os/binfmt/libelf/libelf_sections.c
```diff
@@ -206,68 +208,59 @@ void *elf_find_start_section_addr(struct binary_s *binp)
...
-		binfo("   bss_addr    : %x\n", bin_info->bss_addr);
+		binfo("   rodata_addr : %x\n", g_bin_addr_list[bin_idx].rodata_addr);
+		binfo("   data_addr   : %x\n", g_bin_addr_list[bin_idx].data_addr);
+		binfo("   bss_addr    : %x\n", g_bin_addr_list[bin_idx].bss_addr);
 #endif
 #endif
-		sq_addlast((sq_entry_t *)bin_info, &g_bin_addr_list);
 	} else {
-		berr("ERROR : Failed to allocate for bin symbol.\n");
+		berr("ERROR : Failed to save bin addresses\n");
```
jeongchanKim, line 242: I think it can be confusing, because we don't save the bin address. We try to save the each sections address.
How about changing the log about this?

os/binfmt/libelf/libelf_sections.c
sunghan-chang: Then, how about below?
```
#ifdef config_common_bin
#define NUM_LOADING_BIN  (CONFIG_NUM_APPS + 1)
#else
#define NUM_LOADING_BIN  (CONFIG_NUM_APPS)
#endif

static .... g_bin_addr_list[NUM_LOADING_BIN];
```

sunghan-chang:
```
#define NUM_LOADING_BIN (CONFIG_NUM_APPS \
#ifdef config_common_bin
               + 1 \
#endif
              )
```

os/binfmt/libelf/libelf_sections.c
```diff
@@ -206,68 +208,59 @@ void *elf_find_start_section_addr(struct binary_s *binp)
...
 }
 
 void elf_show_all_bin_addr(void)
 {
-	bin_addr_info_t *info;
-	info = (bin_addr_info_t *)sq_peek(&g_bin_addr_list);
-	while (info) {
-		lldbg("[%s] Text Addr : %p, Text Size : %u\n", BIN_NAME(info->bin_idx), info->text_addr, info->text_size);
-		info = (FAR void *)sq_next((FAR sq_entry_t *)info);
+	int bin_idx;
```
sangam-swami, line 255: function name can be changed to elf_show_all_bin_section_addr ?

os/binfmt/libelf/libelf_sections.c
jeongarmy: The index 0 in binary table is fixed for common binary to handle it differently from other binary.
```
/* Index of 'Common Library' data in binary table. */
#define BM_CMNLIB_IDX              0

/* Binary table, the first data [0] is for Common Library. */
static binmgr_uinfo_t bin_table[USER_BIN_COUNT + 1];
```

sunghan-chang (approved):

PR #5536, opened 2022-05-26
os/board/rtl8721csm: Disable WiFi CFO tracking when BT enable to fix BT frequency offset

yeetee179:
This change will not affect WiFi performance as the default crystal calibration value is used when BT is enabled.


os/board/rtl8721csm/src/libs/RELEASE_NOTE.txt
```diff
@@ -1,5 +1,8 @@
 /* == "version" + "Realtek git version" + "compile date" + "compile time" == */
 
+== version a16b340d3cd7a67dd5b27204b60af31557b0d0d8_2022/05/25-18:50:28 ==‎
+1.	Disable WiFi CFO tracking when BT enable to fix BT frequency offset in lib_wlan
```
thapav, line 4: May be we cam modify as below;
1. Disable WiFi CFO tracking when BT is enabled to fix BT frequency offset in lib_wlan

an4967 (approved):
LGTM

PR #5537, opened 2022-05-30
[UNDER POWERCUT TESTING] os/fs/driver/mtd: Remove additional CRC check of data read from flash

amogh-samsung:
- We do not need a CRC check for verification at read time if read back and verify happens at write time.
- In our focus area of jSmartFs (Journaling enabled), read back and verify always happens at write time,
 which gives us the option to avoid this unnecessarily time taking overhead.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -4399,8 +4399,9 @@ static int smart_readsector(FAR struct smart_struct_s *dev, unsigned long arg)
 	}
 #ifdef CONFIG_MTD_SMART_ENABLE_CRC
 
-	/* When CRC is enabled, we read the entire sector into RAM so we can
-	 * validate the CRC.
+	/* When CRC is enabled, we read the entire sector into RAM so we can validate the CRC.
+	 * However, if Journaling is enabled, written data is read back and verified with CRC,
+	 * so we can skip this step.
 	 */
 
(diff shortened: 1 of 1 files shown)


os/fs/driver/mtd/smart.c
```diff
@@ -4408,26 +4408,6 @@ static int smart_readsector(FAR struct smart_struct_s *dev, unsigned long arg)
 		fdbg("Error reading phys sector %d\n", physsector);
 		return -EIO;
 	}
-#if SMART_STATUS_VERSION == 1
```
thapav, line 4411: Can you please specify the improvement percentage in reading without the CRC check in the commit details ?

thapav, line 4411: What about the cases where journaling is not enabled and/or the readback wont happen while writing to flash?

amogh-samsung: I'll add a TODO comment for the same
We develop SmartFs with journaling primarily as a robust and thorough solution that guarantees FS integrity.
Hence cases where journaling is disabled are possible to build but do not majorly concern us

Taejun-Kwon: I think you can simply update it instead of TODO comment. 3~5 lines of code will be enough

Taejun-Kwon:
Major changes, powercut test should be done

Taejun-Kwon (approved):

PR #5538, opened 2022-06-02
external/json/cJSON.c : Include stdbool.h for true and false

jeongchanKim:
Without including stdbool.h, there can be build error from true and false.
 error: ‘false’ undeclared (first use in this function); did you mean ‘pause’?
         return false;
                ^~~~~
                pause


abhishek-samsung (commented):
LGTM

PR #5546, opened 2022-06-20
build/configs: Add missing board_metadata.txt

jeongarmy:
Add missing board_metadata.txt to set a name of kernel binary to kernel_<BOARD_NAME>.trpk.


build/configs/artik05x/board_metadata.txt
```diff
@@ -23,9 +23,9 @@ USBRULE_BOARD="artik"
 USBRULE_IDVENDOR="0403"
 USBRULE_IDPRODUCT="6010"
 
-BL1="bl1"
```
abhishek-samsung, line 26: minor change, please change  'dile' -> 'file' in the commit message.

build/configs/artik05x/board_metadata.txt
```diff
@@ -23,9 +23,9 @@ USBRULE_BOARD="artik"
...
-BL2="bl2"
-KERNEL="tinyara_head"
-OTA="ota"
+BL1="bl1.bin"
+BL2="bl2.bin"
+KERNEL="tinyara_head.bin"
+OTA="ota.bin"
```
sunghan-chang, line 29: Could you let me know why this should be moved? Please add the reason why this is necessary in the commit description.

jeongarmy: I added filename extension to a file name in board_metadata.txt to support various filename extensions (e.g. .elf, .hex).
For example, Esp32 board has a elf type, not bin type as TizenRT output file (tinyara.elf).

os/tools/set_bininfo.py
```diff
@@ -85,5 +84,7 @@ def save_bininfo(bin_name) :
             save_bininfo(COMMON_BIN_NAME + '.' + TARGET_EXT_NAME)
             continue
 else :
-    save_bininfo("tinyara.bin")
-
+    # All boards except QEMU should have board_metadata.txt.
```
jeongchanKim, line 87: Could you add the comment why QEMU doesn't need to have board metadata?

build/configs/artik05x/board_metadata.txt
jeongchanKim: You mean that we will flash the elf type for esp32 board?

jeongarmy: I checked we flash .elf file for esp32 in this link.
`https://github.com/Samsung/TizenRT/blob/master/build/configs/esp32_DevKitC/README.md`

build/configs/cy4390x/board_metadata.txt
```diff
@@ -0,0 +1,26 @@
+###########################################################################
```
abhishek-samsung, line 1: minor suggestion, could you please change : "to set a name of kernel binary" to "to set the name of kernel binary"

PR #5550, opened 2022-06-23
framework/src/arastorage: fix multiple definition of `g_storage_write_buffer' error

abhishek-samsung:
Multiple definition of `g_storage_write_buffer' error occurs when using docker 1.6.0
g_storage_write_buffer was only used in storage_interface.c file. So moved the
definition from the header into the file


framework/src/arastorage/storage_interface.c
```diff
@@ -84,9 +84,14 @@ struct index_record_s {
 };
 
 /****************************************************************************
-* Public Functions
+* Public Variables
 ****************************************************************************/
 #ifdef CONFIG_ARASTORAGE_ENABLE_WRITE_BUFFER
+struct insert_buffer_s g_storage_write_buffer;
```
jeongchanKim, line 90: If this variable `g_storage_write_buffer` uses only in this file, how about adding `static`?

framework/src/arastorage/storage_interface.c
```diff
@@ -84,9 +84,14 @@ struct index_record_s {
 };
 
 /****************************************************************************
-* Public Functions
+* Public Variables
 ****************************************************************************/
 #ifdef CONFIG_ARASTORAGE_ENABLE_WRITE_BUFFER
+static struct insert_buffer_s g_storage_write_buffer;
```
sunghan-chang, line 90: It's `static` global variable. Then, is it `public` or `private` variable?

thapav: I think this will come under private data

abhishek-samsung: Forgot to change the heading after adding static. Thanks for pointing it out Mr. Chang. Modified.

sunghan-chang (approved):

PR #5551, opened 2022-06-23
os/arch: Fix crash during boot up with Trustzone disabled

kishore-sn:
When TrustZone config is disabled, the EXC_RET values are not
set properly. Set the common value of EXC_RET_BASE for all cases.


os/arch/arm/src/armv8-m/exc_return.h
```diff
@@ -117,11 +119,29 @@
 
 #define EXC_RETURN_STD_CONTEXT   (1 << 4)
 
-/* EXC_RETURN_HANDLER: Return to handler mode. Exception return gets state from
- * the main stack. Execution uses MSP after return.
+/* EXC_RETURN_DEF_STACKING: Default callee register stacking (DCRS).
+ * Indicates whether the default stacking rules apply, or whether the callee
+ * registers are already on the stack.  The possible values of this bit are:
+ * 0 - Stacking of the callee saved registers skipped.  1 - Default rules for
```
jeongchanKim, line 125: How about moving the description of `1` at the next line?
It is much better to read.

os/arch/arm/src/armv8-m/exc_return.h
```diff
@@ -93,9 +92,12 @@
  */
 
 #define EXC_RETURN_BASE          0xffffffa0
-#else
-#define EXC_RETURN_BASE          0xffffffe1
```
thapav, line 97: Are these changes back ported from Nuttx ?

thapav: If Yes, in that case, we need to modify the EXC_RETURN_BASE accordingly ..

kishore-sn: I have checked nuttx, but those values are not working well for us. I have verified these values and they are fine.

os/arch/arm/src/armv8-m/exc_return.h
kishore-sn: Done.

sunghan-chang (approved):

PR #5553, opened 2022-06-24
os: Move board specific build rules into board specific Makefiles

kishore-sn:
There are some rules in the mk files of os folder which perform board
specific operation. Move these lines to board specific Makefiles in
arch folder.


os/arch/arm/src/amebad/Make.defs
```diff
@@ -199,3 +199,17 @@ else
 EXTRA_LIBS += ${TOPDIR}/board/rtl8721csm/src/libs/cmse_implib.a
 endif
 endif
+
```
sangam-swami, line 202: Can you confirm library inclusion in commit message

kishore-sn: Done. Thanks.

sunghan-chang (approved):

PR #5556, opened 2022-06-29
Update ITC & UTC with generic config's 

thapav:
This PR adds generic config's to ITC's & UTC's so that user can change values directly through menuconfig instead of hardcoding them.


apps/examples/testcase/ta_tc/systemio/itc/Kconfig
```diff
@@ -14,20 +14,79 @@ config SYSIO_ITC_UART
...
+	int "Sysio PWM test period"
+	default 1000
+endif
+
 config SYSIO_ITC_GPIO
 	bool "Sysio GPIO TC"
 	default n
+
+if SYSIO_ITC_GPIO
+config SYSIO_ITC_GPIO1_NUM
```
Taejun-Kwon, line 82: postfix 'NUM' is unnecessary I think, or 'PORT' is better option. How do you think?

sunghan-chang: @thapav Could you reply Taejun's comment?

thapav: Yes @Taejun-Kwon  we could remove it. Let me update it

thapav: Modified, thanks.

sunghan-chang (approved):

PR #5563, opened 2022-07-04
os/se/ameba: Fix hardfault caused by device lock calling from Secure

ziyik:
Remove calling device lock from Secure
Protect SE Secure API with Flash operation with device lock


sunghan-chang (approved):

sunghan-chang:
@ziyik It means this is the same operation as RTL8721CSM. Correct?

ziyik:
> @ziyik It means this is the same operation as RTL8721CSM. Correct?

Yes, correct, it does the same as RTL8721CSM

PR #5577, opened 2022-07-20
examples/log_dump: change lldbg to printf

abhishek-samsung:
After the addition of log_dump start and stop, we no longer
need to use lldbg to avoid saving log_dump command output to
the compressed logs.

### DIFF
--- a/apps/examples/log_dump/log_dump_main.c
+++ b/apps/examples/log_dump/log_dump_main.c
@@ -61,7 +61,7 @@ int log_dump_main(int argc, char *argv[])
 	printf("This Log Should NOT be saved!!!\n");
 	sleep(1);
 
-	lldbg("\n*********************   LOG DUMP START  *********************\n");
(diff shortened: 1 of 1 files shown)


apps/examples/log_dump/log_dump_main.c
```diff
@@ -61,7 +61,7 @@ int log_dump_main(int argc, char *argv[])
 	printf("This Log Should NOT be saved!!!\n");
 	sleep(1);
 
-	lldbg("\n*********************   LOG DUMP START  *********************\n");
+	printf("\n*********************   LOG DUMP START  *********************\n");
```
sangam-swami, line 64: if (STOP_LOGDUMP_SAVE(fd) < 0) {
		printf("Failed to stop log dump, errno %d\n", get_errno());
		goto errout;
	}

do we have to use lldbg in this failure to STOP case?

sangam-swami (approved):
looks good

PR #5581, opened 2022-07-21
os/board/rtl8720e/src/component/: Workaround for Flash lock/unlock hanging issue.

fengyunlai:
Workaround for flash lock/unlock issue, enable BLE in defconfig.


build/configs/rtl8720e/hello/defconfig
```diff
@@ -141,12 +141,12 @@ CONFIG_INIC_IPC_HIGH_TP=1
 #
 # Realtek RTL8720E BLE Support
 #
-#CONFIG_AMEBALITE_BLE=y
-#CONFIG_FTL_ENABLED=y
+CONFIG_AMEBALITE_BLE=y
```
Taejun-Kwon, line 144: hello is minimum set of TizenRT, so I think you should not enable BLE here.

fengyunlai: Hi Mr Kwon, I enable BLE here because I disabled it in my previous PR.
Actually for TP1x, I just checked, we also enabled BLE in hello config.
So do you want me to keep this PR's defconfig change, or disable BLE?
Thank you!

sunghan-chang (approved):

PR #5583, opened 2022-07-25
external/webserver: Fix svace WGID 230304 webserver memory leak

amitpurwar:
Signed-off-by: Amit Purwar <amit.purwar@samsung.com>

### DIFF
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -1150,6 +1150,7 @@ int http_send_response_helper(struct http_client_t *client, int status, const ch
                                                            body_len);
 				if (len < 0) {
 					HTTP_LOGE("Error: snprintf failed \n");
+					HTTP_FREE(buf);
 					return HTTP_ERROR;
 				}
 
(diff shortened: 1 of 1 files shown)


external/webserver/http_client.c
```diff
@@ -1150,6 +1150,7 @@ int http_send_response_helper(struct http_client_t *client, int status, const ch
                                                            body_len);
 				if (len < 0) {
 					HTTP_LOGE("Error: snprintf failed \n");
+					HTTP_FREE(buf);
```
abhishek-samsung, line 1153: how about adding goto for the error cases?
```
goto error;
```
```
error :
	HTTP_FREE(buf);
	return HTTP_ERROR;
```

amitpurwar: I prefer to avoid using "goto" statement.

j11591 (approved):
LGTM

vibhor-m (commented):
changes are fine

PR #5589, opened 2022-07-26
kernel/log_dump: reset writesize to blocksize before compression

abhishek-samsung:
The variable writesize provides the output buffer size to the compression
algorithm. So before compression, reset the writesize variable.

### DIFF
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -477,6 +477,7 @@ int log_dump(int argc, char *argv[])
 			compress_last_block = false;
 		} else {
 			/* compress the completely filled block */
+			writesize = CONFIG_LOG_DUMP_CHUNK_SIZE;
 			compress_ret = compress_block(out_buf, &writesize, uncomp_array, CONFIG_LOG_DUMP_CHUNK_SIZE);
 			if (compress_ret != LOG_DUMP_OK) {
 				lldbg("Fail to compress compress_ret = %d\n", compress_ret);


os/kernel/log_dump/log_dump.c
```diff
@@ -157,7 +157,8 @@ int log_dump_read_wake(void)
 		return ERROR;
 	}
 
-	int	msg_compress = true;
+	int msg_compress = true;
```
sunghan-chang, line 160: Let's split the commit for formatting.

os/kernel/log_dump/log_dump.c
```diff
@@ -157,7 +157,8 @@ int log_dump_read_wake(void)
...
+	writesize = CONFIG_LOG_DUMP_CHUNK_SIZE; /* reset writesize before compression */
```
sunghan-chang, line 161: I don't know why this is necessary. Is it changed by the result of the compress_read function even this is initialized as a global variable?

os/kernel/log_dump/log_dump.c
amogh-samsung, line 160: In the commit title, I think that 

> The variable writesize provides the output buffer size to the uncompression algorithm.

should be changed to 

> The variable writesize provides the output buffer size to the compression algorithm.

os/kernel/log_dump/log_dump.c
sunghan-chang: As we talked in the chat, we should re-organize the compression API. I think user think it's much difficult to use.

os/kernel/log_dump/log_dump.c
```diff
@@ -357,6 +358,7 @@ int log_dump_save(char ch)
 		}
 		int msg_compress = false;
 
+		writesize = CONFIG_LOG_DUMP_CHUNK_SIZE; /* reset writesize before compression */
```
sangam-swami, line 361: can we change the writesize before compress_block is called instead of here?

abhishek-samsung: Done

os/kernel/log_dump/log_dump.c
abhishek-samsung: Done

sunghan-chang (approved):

PR #5592, opened 2022-07-29
build/configs/rtl8720e: First release of Linux Download Image Tool to TP1e

SandyYeow:


build/configs/rtl8720e/.flashSpec.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2020 Samsung Electronics All Rights Reserved.
```
sunghan-chang, line 3: Could you modify 2020 to 2022?

PR #5597, opened 2022-08-10
build/configs: config change of SE selection for RTL8720E

ZhenBei-Sin:
Increased the CONFIG_SE_SECURE_CONTEXT_SIZE for support RDP - TP1E
Added board dependency for Secure Context Size in KConfig
Removed the RDP related config in KConfig
Modified the shell script to generate the image file
Removed ameba board checking condition for the ns_setstatusbits
Updated the trustzone image3 with enabled Read Efuse and Flash Write Protection


build/configs/rtl8720e/hello/defconfig
```diff
@@ -24,7 +24,6 @@ CONFIG_BUILD_FLAT=y
 # CONFIG_APP_BINARY_SEPARATION is not set
 # CONFIG_BUILD_2PASS is not set
 CONFIG_BOARD_BUILD_DATE="200204"
-# CONFIG_BINARY_SIGNING is not set
```
jeongchanKim, line 27: Could you revert this change?

build/configs/rtl8720e/rtl8720e_make_bin.sh
```diff
@@ -255,9 +230,6 @@ function copy_kr4_image()
 
 copy_bootloader;
 copy_flashloader;
+copy_km4_image3;
 copy_kr4_image;
-if [ "${CONFIG_BINARY_SIGNING}" == "y" ];then
-	concatenate_binary_with_signing;
```
sunghan-chang, line 260: Could you let me know why you remove the signing?

sunghan-chang (commented):
could you rollback the file permission to 644?

build/configs/rtl8720e/hello/defconfig
ziyik: The image signing mechanism is not available in TP1e TizenRT master, we have provided a signing script.
All image signing and encryption can be done via script.

build/configs/rtl8720e/rtl8720e_make_bin.sh
ziyik: The image signing mechanism is not available in TP1e TizenRT master, we have provided a signing script.
All image signing and encryption can be done via script.

build/configs/rtl8720e/hello/defconfig
sunghan-chang: @ziyik It's the same as TP1x. This config is necessary to decide whether your script should be executed or not internally.
Because it's impossible, we disable it by default. And no way to enable and even if it's enabled, nothing will be called.
Could you keep it as disabled? (revert the change)

build/configs/rtl8720e/rtl8720e_make_bin.sh
sunghan-chang: In the public, signing and non-signing eventually calls the same (nothing). We keep this for easy maintenance. Could you keep it as it is?

sunghan-chang:
> could you rollback the file permission to 644?

Could you check my previous comment?

build/configs/rtl8720e/hello/defconfig
ziyik: We will keep the config but it does nothing even if it's been called, is this the correct understanding?

build/configs/rtl8720e/rtl8720e_make_bin.sh
ZhenBei-Sin: Hi @sunghan-chang , wanna double confirm that as you have mentioned that "signing and non-signing eventually calls the same (nothing)"
may I know is that doing as below way:

if [ "${CONFIG_BINARY_SIGNING}" == "y" ];then
	**concatenate_binary_without_signing;**
else
	**concatenate_binary_without_signing;**
fi

Thank You.

build/configs/rtl8720e/hello/defconfig
sunghan-chang: correct.

build/configs/rtl8720e/rtl8720e_make_bin.sh
sunghan-chang: @ZhenBei-Sin Could you refer rtl8721csm_make_bin.sh and rtl8721csm_signing.sh?

ZhenBei-Sin: Copied rtl8721csm_signing.sh be rtl8720e_signing.sh

if [ "${CONFIG_BINARY_SIGNING}" == "y" ];then
  concatenate_binary_with_signing;
else
  concatenate_binary_without_signing;
fi

concatenate_binary_with_signing will call to rtl8720e_signing.sh

sunghan-chang:
@ZhenBei-Sin @ziyik One more question. RDP is supported in RTL8721CSM as well. Why is the context size changed to 9216?

ZhenBei-Sin:
> @ZhenBei-Sin @ziyik One more question. RDP is supported in RTL8721CSM as well. Why is the context size changed to 9216?

Hi @sunghan-chang , it was due to the compiler different. RTL8721CSM: IAR but RTL8720E: GCC.
The increased context size is not for enabling support RDP only, it was used for SE.

Thank You.

os/se/ameba/Kconfig
```diff
@@ -52,7 +52,6 @@ config HW_SE_STORAGE
 
 config SE_SECURE_CONTEXT_SIZE
 	int "Secure Context Size"
-	default 6144
+	default 9216
```
sunghan-chang, line 55: Then, how about making a dependency with the board?
```
default 6144 if AMEBAD_TRUSTZONE
default 9216 if AMEBALITE_TRUSTZONE
```
Because you use this config for both.

ZhenBei-Sin: Hi @sunghan-chang ,
We have made changes to it based on your suggestion.
Thank You.

sunghan-chang (approved):

PR #5616, opened 2022-08-24
os/tools: Validate packages after build

jeongarmy:
Call script for checking package validation after build.
It parses the package header information and show them.
And verify the package by calculating the crc and comparing it with the value of the header.

### DIFF
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -545,7 +545,7 @@ endif
 	$(Q) python ./tools/set_bininfo.py $(BIN_EXT)
 	$(Q) $(call MAKE_SAMSUNG_HEADER, $(BIN_EXE), $(BIN_EXT))
 ifeq ($(CONFIG_FLASH_PARTITION),y)
-	$(Q) python ./tools/check_output_size.py
+	$(Q) python ./tools/validate_output.py
 endif
 	$(Q) $(call MAKE_BOOTPARAM)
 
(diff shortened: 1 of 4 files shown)


sunghan-chang:
Which kind of validation is included? Could you leave what it is in the commit description?

os/tools/check_output.py
```diff
@@ -61,6 +62,16 @@ def check_binary_size(bin_type, part_size):
         os.remove(output_path)
         FAIL_TO_BUILD = True
 
+def check_binary_header(bin_type, part_size):
```
abhishek-samsung, line 65: Is part_size required in this function?

os/tools/check_output.py
```diff
@@ -133,4 +144,14 @@ def check_binary_size(bin_type, part_size):
...
+
+# Validate the header data of binaries
+check_binary_header("KERNEL", KERNEL_PARTITION_SIZE)
+if CONFIG_APP_BINARY_SEPARATION == "y" :
+    if APP1_IDX != 0 :
+        check_binary_header("APP1", APP1_PARTITION_SIZE)
+    if APP2_IDX != 0 :
+        check_binary_header("APP2", APP2_PARTITION_SIZE)
+    if CONFIG_SUPPORT_COMMON_BINARY == "y" :
+        check_binary_header("COMMON", COMMON_PARTITION_SIZE)
```
sunghan-chang, line 157: Could you check the line end?

sunghan-chang (commented):
I can't distinguish purpose of each, check_output.py and check_package.py.
`output` and `package` looks the same.
Could you rename them to know what it is from the name?

sunghan-chang (approved):

PR #5615, opened 2022-08-24
os/arch/arm/src/amebalite: Added serial flowcontrol warnings

kaikaikaime:
Warns the user to use UART0/3 when flowcontrol is used.

PR #5619, opened 2022-08-25
os/arch/arm/src/amebalite: Add serial flowcontrol warnings

kaikaikaime:
Warns the user to use UART0/3 when flowcontrol is used.

### DIFF
--- a/os/arch/arm/src/amebalite/amebalite_serial.c
+++ b/os/arch/arm/src/amebalite/amebalite_serial.c
@@ -199,7 +199,7 @@
  * Private Types
  ****************************************************************************/
 /*
- * Available UARTs in AmebaSmart
+ * Available UARTs in AmebaLite
  * UART0_DEV: hp uart0
  * UART1_DEV: hp uart1
  * UART2_DEV: hp uart2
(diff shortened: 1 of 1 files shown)


kaikaikaime:
Current status:
Used #if defined instead
Added uart port identifier in the loguart print
Included comment to uart1,2,4 that autoflow control were not implemented
loguart output is where warning should be placed

sunghan-chang (commented):
Do you need to modify something? Could you let me know current status regarding this?

kaikaikaime:
> Current status: Used #if defined instead Added uart port identifier in the loguart print Included comment to uart1,2,4 that autoflow control were not implemented loguart output is where warning should be placed
> 
> The printf is tied to loguart uart4 which will be init first and printf will head towards there , which is in (else condition) from UART0/1/2/3 for flow control !
> 
> [Capture](https://user-images.githubusercontent.com/61275451/186621522-2aa0fa50-6889-4776-840f-b7a3a5b8f860.PNG) l

This the updated version of the commit, and above are the changes that has been made.
It is a completed.

sunghan-chang:
@kaikaikaime You means below is done already in the PR?
> [kaikaikaime](https://github.com/kaikaikaime) commented [4 hours ago](https://github.com/Samsung/TizenRT/pull/5615#issuecomment-1226822009)
@sunghan-chang Ah yes, my ETA will be done with checks by next week

kaikaikaime:
Yes, I was planning to implement additional feature, which is not required at this point in time.

sunghan-chang (approved):

PR #5630, opened 2022-09-05
board/rtl8720e/ram_km4: Add Kernel OTA image signature verify API.

fengyunlai:
Add API OTA_Image2SignatureCheck() to verify Kernel OTA image's signature.


fengyunlai:
I tested in hello_main like this:
Input the FLASH address of OTA image to verify, output TRUE or FALSE.

#ifdef CONFIG_BUILD_KERNEL
int main(int argc, FAR char *argv[])
#else
int hello_main(int argc, char *argv[])
#endif
{
	u32 addr = 0x0837F000;
	printf("Hello, World!!\n");
	if (OTA_Image2SignatureCheck(addr) == TRUE)
		printf("OTA image check pass\n");
	else
		printf("OTA image check fail\n");
	return 0;
}

sunghan-chang:
This PR has too long title. Let's use like below.
```
board/rtl8720e/ram_km4: Add ....
```

sunghan-chang:
No need to give a size? Just start address is enough?

sunghan-chang:
We have the Samsung header at beginning of the package (binary). Is it ok with the start of the package? or should it be start of the realtek output (in other words, address of next of samsung header?)

jeongarmy:
@fengyunlai Which binary did you test for success case? Did you use the signing script with default keys sent by eric.kim by mail?

fengyunlai:
> No need to give a size? Just start address is enough?

No need, Mr Chang. The sub-image size is included in image header. And for TP1X, we also do not input size.

fengyunlai:
> We have the Samsung header at beginning of the package (binary). Is it ok with the start of the package? or should it be start of the realtek output (in other words, address of next of samsung header?)

Yes Mr Chang, I have considered the Samsung header size. So start of the package(includes Samsung Size) is correct input. You can check my test example case. for current FLASH layout, we input 0x0837F000 as OTA2 address, 0x0837F000~0x0837F010 is the Samsung size location.

fengyunlai:
> @fengyunlai Which binary did you test for success case? Did you use the signing script with default keys sent by eric.kim by mail?

I checked default keys yes. And actually it is okay to test with different keys. Because the success of verification should be suitable for any key sets.

sunghan-chang (approved):

PR #5631, opened 2022-09-05
os/board/rtl8720e: wifi fix AP compatibility issues

gexuyan:
Fix wifi connection issue with ASUS, Netgear, and Cisco routers.


sunghan-chang:
@gexuyan There is no code to be reviewed. Shall I merge this now? How about FPU enable or disable?

sunghan-chang:
@gexuyan Is this verified with Samsung HW?

os/net/netmgr/netmgr_ioctl_lwip.c
```diff
@@ -152,7 +152,7 @@ int _netdev_dhcpc_start(const char *intf)
 
 	NET_LOGKV(TAG, "DHCP client started successfully, waiting IP address (timeout %d msecs)\n", timeleft);
 	while (netifapi_dhcp_address_valid(cnif) != 0) {
-		usleep(100000);
+		usleep(200000);
 		timeleft -= 100;
```
kishore-sn, line 156: Is there any connection between `timeleft` and the `usleep` interval? The usleep is increased to 200ms, but timeleft is decremented by 100 only.

j11591:
@gexuyan What is problem with usleep(100000);?

gexuyan:
> @gexuyan Is this verified with Samsung HW?

Dear @sunghan-chang ,
Yes this has been verified with Samsung HW.

os/net/netmgr/netmgr_ioctl_lwip.c
gexuyan: Dear @kishore-sn ,
ya this seems a little bit tricky here, this modification's outcome is to double the timeout interval. Or do you suggest I change the CONFIG_LWIP_DHCPC_TIMEOUT in defconfig?

gexuyan:
> @gexuyan What is problem with usleep(100000);?

Dear @j11591 ,

Actually we just want to increase the timeout interval, because when connecting with ASUS ap, the ASUS ap will send fragmented dhcp frames leading to longer processing time. As long as increasing the timeout interval, the issue is resolved.

j11591:
> 

Does It happen only with rtl8720e??? or does same issue happen with rtl8721csm?

Dane-Kang:
@j11591 
Samsung HW team already confirmed TP1x is not a problem for this connection problem.
If you need double check, please contact HW team member.

j11591:
> @j11591 Samsung HW team already confirmed TP1x is not a problem for this connection problem. If you need double check, you can contact HW team member.

https://github.com/Samsung/TizenRT/blob/f8f776dd183246ad8890422c1ee5e8f33ab2aaaf/build/configs/rtl8721csm/loadable_apps/defconfig#L789
In TP1x config, timeout is 10sec. I am curious why those value are different between TP1x and TP1e.

gexuyan:
> > @j11591 Samsung HW team already confirmed TP1x is not a problem for this connection problem. If you need double check, you can contact HW team member.
> 
> https://github.com/Samsung/TizenRT/blob/f8f776dd183246ad8890422c1ee5e8f33ab2aaaf/build/configs/rtl8721csm/loadable_apps/defconfig#L789
> 
> 
> In TP1x config, timeout is 10sec. I am curious why those value are different between TP1x and TP1e.

Dear @j11591 , this should be due to tp1x and tp1e driver are different, and tp1e has an IPC(Inter Processor Communication) layer between kr4 and km4, so it may take longer time to processing the incoming dhcp frames.

j11591:
> 
I can't understand.... even though there is IPC layer... does it need 10 sec more?? I think it should be take under 1sec
With TP1x, I've never seen DHCP takes about 10sec....

gexuyan:
> 

Dear @j11591 , 
ya what you mentioned also make sense, I will check again for this timing issue

sunghan-chang:
@gexuyan @Dane-Kang RTL8720E spends more memory than RTL8721, takes more time to work wifi operation. But same throughput. I don't know what is better point than 8721.

gexuyan:
> @gexuyan @Dane-Kang RTL8720E spends more memory than RTL8721, takes more time to work wifi operation. But same throughput. I don't know what is better point than 8721.

Dear @sunghan-chang and @j11591 ,
For the memory size, I think this is due to some feature implementation is not stable at this time, the code size optimization is still ongoing. And for this ASUS timing issue, the dhcp timeout only happens sometimes, not always take 10+sec, but I found this phenomenon also happens with freertos image, but freertos image set the dhcp timeout interval longer to avoid the connection issue. I will keep checking this timing issue today. 
For the throughput, actually 8720E freertos image can hit higher TP(35M+) than 8721 freertos image, and 8720e tizenrt image on 8720 EVB board TP result(33M+) is also higher than 8721 tizenrt image, but with the 5th TPIE module, the TP will drop a little bit, that's why for the current stage the TP looks same. I think 8720E's new km4+kr4 with IPC structure actually provides more modularity than 8721, some current issues are due to the base sdk is not stable at this time, but in the future it should be easier to do the maintenance for wlan. For other detailed spec you may also check with our sales or PM.
Thank You.

sunghan-chang:
@gexuyan Thank you for explanation.

gexuyan:
> 

Dear @sunghan-chang ,

Actually jianwei's PR5634(already merged) also solved the ASUS ap connection dhcp slow problem. We updated the kr4 image with freertos sdk commits and make it reset the baseband, which can fix the DHCP hang issue. 
This matter I have discussed with Dane but he seems on leave today. I already updated the test image and he will help to check with other routers see whether still get any ap compatibility issues. If urgent, Samsung hw team can also test with RTL8720E branch latest image to check if these AP compatibility issues still happen.
Thank you!

Dane-Kang:
@sunghan-chang,
As xuyan mentioned, we noticed PR5634 was included this AP compatibility modification on wlan library.
And we are also confirmed ASUS AP can be connected even not modify the delay part. 
That's why xuyan close this PR.
Sorry to interrupt your understanding.

@gexuyan ,
Could you explain the modification on wifi library part for AP compatibility issue? 
Thank you.

sunghan-chang:
@gexuyan @Dane-Kang This is not good at co-work. We never know this was merged by PR 5634 merging. I already asked your team to leave any dependency on a PR like this case when we did previous project. This situation could cause any mis-understanding or any issue we can't guess. Please concern this. Thank you.

gexuyan:
> @gexuyan @Dane-Kang This is not good at co-work. We never know this was merged by PR 5634 merging. I already asked your team to leave any dependency on a PR like this case when we did previous project. This situation could cause any mis-understanding or any issue we can't guess. Please concern this. Thank you.

Sure please don't worry, we will indicate if there's any dependency on different PR. Actually this time it's the same modification in wlan driver fixed different issues. I checked with the original committer, this modification is to reset baseband conditions after RF calibration, if not doing this, RF hard to receive in some OFDM rate pkt and caused RX to hang. Thank You.

PR #5633, opened 2022-09-06
[ON HOLD] os/driver/mtd: Update smartfs garbage collection

amogh-samsung:
- Select a block which has the maximum released sectors for collection,
 and at the same time minimum committed sectors that would require relocation.
- We can exnd the search for a block if we fnd a block that we know is the
 best possible option for collection.

### DIFF
--- a/os/fs/driver/mtd/smart.c
+++ b/os/fs/driver/mtd/smart.c
@@ -4308,8 +4308,11 @@ static int smart_writesector(FAR struct smart_struct_s *dev, unsigned long arg)
 
 		/* Since we performed a relocation, do garbage collection to
 		 * ensure we don't fill up our flash with released blocks.
+		 * Check if the number of free sectors is less than a certain threshold.
 		 */
-		smart_garbagecollect(dev);
+		if (dev->freesectors < ((dev->availSectPerBlk * dev->neraseblocks) >> 4)) {
+			smart_garbagecollect(dev);
+		}
 	} else {				/* needsrelocate */
 
 #ifdef CONFIG_MTD_SMART_ENABLE_CRC
@@ -4609,13 +4612,16 @@ static int smart_allocsector(FAR struct smart_struct_s *dev, unsigned long reque
 		return -EIO;
 	}
 
-	/* Check if we need to do garbage collection.  We have to
-	 * ensure we keep enough reserved free sectors to perform garbage
-	 * collection as it involves moving sectors from blocks with
-	 * released sectors into blocks with free sectors, then
-	 * erasing the vacated block. */
+	/* Check if we need to do garbage collection. We have to ensure that we
+	 * keep enough reserved free sectors to perform garbage collection as it
+	 * involves moving active sectors from blocks with more released sectors
+	 * to blocks with more free sectors and then erasing the vacated block.
+	 * Check if the number of free sectors is less tha a certain threshold.
+	 */
+	if (dev->freesectors < ((dev->availSectPerBlk * dev->neraseblocks) >> 4)) {
+		smart_garbagecollect(dev);
+	}
 
-	smart_garbagecollect(dev);
 	/* Find a free physical sector. */
 
 	physicalsector = smart_findfreephyssector(dev, FALSE);


os/fs/driver/mtd/smart.c
```diff
@@ -3573,37 +3578,48 @@ static int smart_garbagecollect(FAR struct smart_struct_s *dev)
 {
 	uint16_t collectblock;
 	uint16_t releasemax;
+	uint16_t temp_score, max_score;
```
thapav, line 3581: Kindly declare in separate lines.

os/fs/driver/mtd/smart.c
```diff
@@ -3573,37 +3578,48 @@ static int smart_garbagecollect(FAR struct smart_struct_s *dev)
...
 	bool collect = TRUE;
 	int x;
 	int ret;
 #ifdef CONFIG_MTD_SMART_PACK_COUNTS
-	uint8_t count;
+	uint8_t count_released, count_free;
```
thapav, line 3586: same as above

amogh-samsung:
Given a fixed number of released sectors, the EraseBlock with the maximum number of free sectors (i.e. minimum number of committed sectors) will be chosen.

image

os/fs/driver/mtd/smart.c
```diff
@@ -3573,37 +3578,48 @@ static int smart_garbagecollect(FAR struct smart_struct_s *dev)
...
 		if (collect) {
 			/* Find the block with the most released sectors. */
-
 			collectblock = 0xFFFF;
-			releasemax = 0;
+
+			/* Maximum possible score of a block for GC */
+			releasemax = (dev->availSectPerBlk * SMART_RELEASED_SECTOR_GC_WEIGHTAGE);
+
+			/* The score for any block should be more than this score for it to be consdered for collecting garbage
```
thapav, line 3613: considered*

os/fs/driver/mtd/smart.c
```diff
@@ -3614,17 +3630,21 @@ static int smart_garbagecollect(FAR struct smart_struct_s *dev)
...
 #else
-				if (dev->releasecount[x] > releasemax) {
-					releasemax = dev->releasecount[x];
-					collectblock = x;
-				}
+				temp_score = (dev->releasecount[x] * SMART_RELEASED_SECTOR_GC_WEIGHTAGE) + (dev-freecount[x] * SMART_FREE_SECTOR_GC_WEIGHTAGE);
 #endif
+				if (temp_score > max_score) {
+					max_score = temp_score;
+					colectblock = x;
```
thapav, line 3641: collectblock*

Taejun-Kwon:
I can't see any meaningful benefit regarding this commit, Essentially, GC is necessary to get new free sector, so release count is important.
As I understand, your apporach is that reduce relocate block, am I right?
For example, 
Block 1 : release sector(3) free sector(1)
Block 2 : release sector(3) committed sector(1)
Then after GC, if we process block1, we can get new 3 sectors.
And block 2, we also get new 3 sectors (4 new sectors, but 1 relocated sector).
so if release count is same, then we gain same number of new sector.
And both of 2 case actually we lost chance to write 1 sector (block 1 has 1 freecount, but erase, block 2, committed sector to be relocated)
Most important thing is that smartfs have a wearlevel count, if we skip any block from gc by this logic, but it will be selected by wearlevel soon, and relocate block will be happened definately.
This is major point of smartfs, So you should do below.
1. Kind of Monkey test should be done for long long time.
2. During test, some of file never be erased to make gc condition, normal condition, clean condition. number of file should be increased step by step.
3. Count erase count for both of logic
4. Check wear level state.
5. Analysis result.

amogh-samsung:
@Taejun-Kwon I agree with your comment, the long term benefits of this change are insignificant.
PFB the data from the current algorithm and the suggested change in this PR.

Whatever Block we skip now will either be picked up later for Erasing or else it will be erased less and cause bad wear distribution.
The reduction in write counts is very small and insignificant as compared to the wear situation.

image

I will look at the bigger picture, modifying only the block selection algorithm may not provide benefits in the long run.

amogh-samsung:
Reducing unnecessary calls to smart_garbagecollect() function ->
image

Taejun-Kwon:
> @Taejun-Kwon I agree with your comment, the long term benefits of this change are insignificant. PFB the data from the current algorithm and the suggested change in this PR.
> 
> Whatever Block we skip now will either be picked up later for Erasing or else it will be erased less and cause bad wear distribution. The reduction in write counts is very small and insignificant as compared to the wear situation.
> 
> image
> 
> I will look at the bigger picture, modifying only the block selection algorithm may not provide benefits in the long run.

Test app will affect to result, you should test various type of test.

amogh-samsung:
> > @Taejun-Kwon I agree with your comment, the long term benefits of this change are insignificant. PFB the data from the current algorithm and the suggested change in this PR.
> > Whatever Block we skip now will either be picked up later for Erasing or else it will be erased less and cause bad wear distribution. The reduction in write counts is very small and insignificant as compared to the wear situation.
> > image
> > I will look at the bigger picture, modifying only the block selection algorithm may not provide benefits in the long run.
> 
> Test app will affect to result, you should test various type of test.

I ran different tests, all of them showed results on a similar trend.
The algorithm that I proposed first disturbs wear levelling, and does not improve write counts by much.
Hence, as per your suggestions, that trade-off is not suitable.

Now I added 2 new commits which do not change the block selection algorithm.

os/fs/driver/mtd/smart.c
```diff
@@ -3617,19 +3615,20 @@ static int smart_garbagecollect(FAR struct smart_struct_s *dev)
 				count = smart_get_count(dev, dev->releasecount, x);
 				if (count > releasemax) {
 					releasemax = count;
-					collectblock = x;
-				}
 #else
 				if (dev->releasecount[x] > releasemax) {
 					releasemax = dev->releasecount[x];
+#endif
```
Taejun-Kwon, line 3621: I prefer previous style it is more good readability
```
#ifdef A
    if (true) {
      do_something;
   }
#else
   if (true) {
     do_something;
   }
#endif

File: os/fs/driver/mtd/smart.c [hunk 6]
```diff
@@ -4308,8 +4306,11 @@ static int smart_writesector(FAR struct smart_struct_s *dev, unsigned long arg)
 
 		/* Since we performed a relocation, do garbage collection to
 		 * ensure we don't fill up our flash with released blocks.
+		 * Check if the number of free sectors is less than a certain threshold (4%)
 		 */
-		smart_garbagecollect(dev);
+		if (dev->freesectors < (dev->availSectPerBlk * dev->neraseblocks) / 25) {
```
Taejun-Kwon, line 4311: it's not good condition, we can adjust availSectPerBlk greater than 25.
I think making macro for condition will be better.
```
#define GC_CONDITION(dev) (dev->freesectors <= (dev->sectorsPerBlk + 4) || (dev->releasesectors > dev->freesectors && dev->freesectors < (dev->totalsectors >> 5)) ......
```

os/fs/driver/mtd/smart.c
amogh-samsung: The only difference between this suggestion and the old version is that with this suggestion we avoid duplicating 1 line of code.


```
#ifdef A
    if () {
        do something first based on A;
        then do something more
    }
#else
    if () {
        do something first based on B;
        then do something more;
    }
#endif
```

Has been changed to : -
 
```
#ifdef A
    if () {
        do something first base on A;
    }
#else
    if () {
        do something first based on B;
    }
#endif
    then do something more;
```

os/fs/driver/mtd/smart.c
amogh-samsung: The factor of 25 is not based on dev->availSectPerBlk
It represents 4% of the total sectors.

GC Threshold conditions are: -
1. (freesectors < (totalsectors / 32)) && (freesectors < releasedsectors)
This means freesectors < ~3.2% of totalsectors
2. freesectors < (dev->availSectPerBlk + 4)

A condition check of 4% is a superset of both of these conditions.
Hence, it limits the number of calls to the garbagecollect function without inhibiting any of the necessary calls.

Do you suggest that we should check both the threshold conditions itself with a MACRO before calling smart_garbagecollect()?

amogh-samsung: I think 5% is a better threshold to use, it covers our cases better

davidfather: @amogh-samsung What is the purpose of this modification? Is it to reduce the number of call to garbage collection? In the garbage collection fuction, there is already 4% check logic, so this modification can't change any behavior to garbage collection logic and wear leveling performance. Am I right?

amogh-samsung: Yes, changing GC logic risks our feature and stability.
This proposition only limits the calls to GC function, does not change GC logic itself.

https://github.com/Samsung/TizenRT/pull/5633#issuecomment-1243535626

os/fs/driver/mtd/smart.c
```diff
@@ -4613,9 +4614,13 @@ static int smart_allocsector(FAR struct smart_struct_s *dev, unsigned long reque
 	 * ensure we keep enough reserved free sectors to perform garbage
 	 * collection as it involves moving sectors from blocks with
 	 * released sectors into blocks with free sectors, then
-	 * erasing the vacated block. */
+	 * erasing the vacated block.
+	 * Check if the number of free sectors is less tha a certain threshold (4%)
+	 */
+	if (dev->freesectors < (dev->availSectPerBlk * dev->neraseblocks) / 25) {
```
davidfather, line 4620: Regarding the computational overhead of floating division, how about 6.3% threshold, i.e., dev->availSectPerBlk * dev->neraseblocks) >> 4?

amogh-samsung: I agree, using this will be advantageous
And 6.3% works as our limit well enough

amogh-samsung: Updated this PR
Sorry for the delay.

os/fs/driver/mtd/smart.c
```diff
@@ -4609,13 +4612,16 @@ static int smart_allocsector(FAR struct smart_struct_s *dev, unsigned long reque
 		return -EIO;
 	}
 
-	/* Check if we need to do garbage collection.  We have to
-	 * ensure we keep enough reserved free sectors to perform garbage
-	 * collection as it involves moving sectors from blocks with
-	 * released sectors into blocks with free sectors, then
-	 * erasing the vacated block. */
+	/* Check if we need to do garbage collection.  We have to ensure the  we
```
swasthik22, line 4615: we have to ensure the -> we have to ensure that ...

pillip8282 (commented):
LGTM
How about to use label that github provides instead of adding it on a title.

PR #5634, opened 2022-09-07
build/tools/amebalite/gnu_utility: Fix UDP issue during Easy Setup

jwei5:
* Update KR4 image to fix UDP issue during Easy Setup


sunghan-chang (commented):
Could you revert a file permission to 644?

sunghan-chang:
@jwei5 Could you let us know the reason and the solution of the issue?

jwei5:
@sunghan-chang The issue is due to our driver not resetting the baseband after setting channel. This causes the RX to hang, resulting in this issue. We have fixed this by resetting the baseband after setting channel.

sunghan-chang (approved):

sunghan-chang:
@j11591 Could you review and merge this?
```
The issue is due to our driver not resetting the baseband after setting channel(when starting SoftAP). This causes the RX to hang, resulting in this issue. We have fixed this by resetting the baseband after setting channel.
```

j11591 (approved):
LGTM

PR #5640, opened 2022-09-14
os/se/ameba: Added TP1E Samsung Key Address

ZhenBei-Sin:
Added TP1E Samsung Key Address with condition checking for the board type

### DIFF
--- a/os/se/ameba/security_ameba_wrapper_tz.c
+++ b/os/se/ameba/security_ameba_wrapper_tz.c
@@ -99,7 +99,12 @@
 #define SECTOR_SIZE 4096
 
 /* Secure efuse key location */
+#ifdef CONFIG_AMEBAD_TRUSTZONE
 #define SAMSUNG_KEY_ADDR 0x150
+#endif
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@ZhenBei-Sin Could you let us know what it is? what for?

ZhenBei-Sin:
> @ZhenBei-Sin Could you let us know what it is? what for?

Hi @sunghan-chang , as previous require have the Test API to inject the secure key and factory key/cert.
This API will be used temporarily until MP SDK is ready, so image3 is not committed in this PR.
While we are doing the test API for factory key/cert we do need to change the key address passing from SE layer.
The injection of the key is required the TP1E Samsung Key Address. This address is different for TP1x and TP1e.
So the changes to the address is needed.

Thank You

sunghan-chang:
@ZhenBei-Sin Sorry, I couldn't fully understand what you said. Let me ask it to Eric and Dane.

sunghan-chang (approved):

PR #5641, opened 2022-09-14
examples/ble_tester: Implement BLE example between 2 devices 

an4967:
- implement BLE example between 2 devices
- set BLE event handler stack size from 2048 to 4096
- set BLE default MTU from 240 to 247
- set interrupt message number from 8 to 32


ydlee-sec (approved):
LGTM~

PR #5643, opened 2022-09-14
os\board\rtl8720e\src\component\filesystem\ftl: Add conditional checking to verify pointer value returned by rtw_zalloc is not null.

yeetee179:


os/board/rtl8720e/src/component/filesystem/ftl/ftl.c
```diff
@@ -1440,6 +1440,10 @@ void ftl_mapping_table_init(void)
 		//                                  MAPPING_TABLE_SIZE);//table is initialised as 0
 
 		ftl_mapping_table = rtw_zmalloc(MAPPING_TABLE_SIZE);
+		if (NULL == ftl_mapping_table) {
+			FTL_PRINTF(FTL_LEVEL_ERROR, "[ftl] ftl_mapping_table rtw_zmalloc fail!");
+			return;
```
sunghan-chang, line 1445: This will cause uncomplete init. But ftl layer will execute next functions without init. Is it ok?

sunghan-chang:
@yeetee179 For SVACE, this is enough. But for the operation, this is not enough I think. Could you confirm?

os/board/rtl8720e/src/component/filesystem/ftl/ftl.c
yeetee179: Is it ok if FTL_ASSERT is added after FTL_PRINTF at line 1444, such that if init is incomplete it will stop execution.

sunghan-chang: If it's mandatory, then assert will be ok. If ftl is working without init, then, current is ok. It depends on the ftl structure.

sunghan-chang: @yeetee179 Could you let me know how will you do for this? Sorry but this is urgent for me.

yeetee179: I will add the assertion here, I will commit and push again.

sunghan-chang:
@yeetee179 Is this final? Can I merge it now?

yeetee179:
@sunghan-chang Yes you can merge it for now.

PR #5648, opened 2022-09-14
os/board/rtl8720e: Resolve SVACE warnings with WGID 244051

keerthanalex:
Resolve 244051 by removing retry_count part on wifi_eap_config.c


sunghan-chang:
@keerthanalex So, the function does not support a retry logic. Is this correct?

keerthanalex:
> @keerthanalex So, the function does not support a retry logic. Is this correct?

@sunghan-chang, we can also modify the code like below to support the retry. Please let us know on which version you want to support.

 _int retry_count = 3, ret;_

we can discuss further to fix the retry count number.

sunghan-chang:
@j11591 Could you give us your opinion for retry logic?

sunghan-chang (approved):

PR #5653, opened 2022-09-15
os/board/rtl8720e: Correct the internal pull up/pull down option for PA_19, LogUART RX, to pull up.

fengyunlai:
Change PA_19 from GPIO_PuPd_DOWN to GPIO_PuPd_UP, further changes might be made once the final defaul pulling table been settle done.


fengyunlai:
Sorry the branch has some problem, I will push another soon.

sunghan-chang:
@fengyunlai Could you let us know why this is necessary?

fengyunlai:
> @fengyunlai Could you let us know why this is necessary?

Because for UART, the internal pull up/ down should be set to pullup by default. Otherwise, we will face an issue if we reboot the board while disconnecting PA_19. Floating PA_19 will trigger LOG UART RX irq wrongly, because it is not pulled up.

But anyway, that is only for disconnecting LOG UART RX pin and reboot case. When connection is correct, no issue will happen.

sunghan-chang:
@fengyunlai We use two UARTs for console log and data with external module. Is it ok with PA_19 only?

fengyunlai:
> @fengyunlai We use two UARTs for console log and data with external module. Is it ok with PA_19 only?

Yes Mr Chang, PA_19 only is enough. Because when TizneRT use the two UARTs, in serial_init(), we will set the pins to pullup.

The issue fixed in this PR is for the LogUART pin status at the stage after boot, and before initialing console/data UART in TizenRT code.
So at that time, only LogUART is used by Realtek code.

sunghan-chang (approved):

PR #5654, opened 2022-09-15
Fix svace issues

jeongarmy:


sunghan-chang:
@sangam-swami @abhishek-samsung She make changes on compression. Could you review?

sunghan-chang:
@taehwan-hwang Could you review changes on ble manager for SVACE?

os/tools/mksyscall.c
```diff
@@ -543,10 +543,12 @@ int main(int argc, char **argv, char **envp)
 		case '?':
 			fprintf(stderr, "Unrecognized option: %c\n", optopt);
 			show_usage(argv[0]);
+			break;
 
 		case ':':
 			fprintf(stderr, "Missing option argument, option: %c\n", optopt);
 			show_usage(argv[0]);
+			break;
```
sunghan-chang, line 551: how about default?

os/tools/mksyscall.c
```diff
@@ -543,10 +543,12 @@ int main(int argc, char **argv, char **envp)
...
 
 		default:
```
abhishek-samsung, line 553: I think we will not reach default at all as the above cases cover all possibilities, if so can we remove the default case?

jeongarmy: I think we need to keep this because of unexpected inputs.

sunghan-chang:
@sangam-swami @abhishek-samsung Could you confirm the changes on compress is ok?

sunghan-chang (approved):

PR #5658, opened 2022-09-15
os/board/rtl8720e: Resolve SVACE warnings with WGID 244038/244039/244061

keerthanalex:
Resolve 244038 and 244039 by null terminating the string 
Resolve 244061 by using strncpy instead of strcpy and null terminating the string


os/board/rtl8720e/src/component/wifi/wifi_interactive_mode.c
```diff
@@ -793,6 +795,7 @@ static int _netlib_setmacaddr(const char *ifname, const uint8_t *macaddr)
 
 			/* Put the driver name into the request */
 			strncpy(req.ifr_name, ifname, 6);
```
sunghan-chang, line 797: To reduce duplicated operation, how about `5` instead of `6`? Because last index is set as `\0` at line 798 always.

keerthanalex: @sunghan-chang , how about modifying like below

_strncpy(req.ifr_name, ifname, IFNAMSIZ - 1);
req.ifr_name[IFNAMSIZ - 1] = '\0';_

sunghan-chang (approved):

PR #5660, opened 2022-09-16
os/board/rtl8720e: Fix SVACE warning WGID 244043

ziyik:


os/board/rtl8720e/src/component/bluetooth/driver/hci/hci_transport/hci_h4.c
```diff
@@ -149,6 +149,8 @@ static void h4_rx_thread(void *context)
             {
                 if (discardable)
                 {
+                    if (body_len - 1 > CONFIG_HCI_RX_BUF_LEN)
```
sunghan-chang, line 152: `body_len` is assinged to `hdr.evt.len` at line 145. But `hdr.evt.len` has uint8_t. So in this line, `body_len` has 0~255. 
`body_len -1` is never higher than `CONFIG_HCI_RX_BUF_LEN`. That's why we reverted previous change. (This is shown as another SVACE issue)

ziyik: Okay, understood.

PR #5662, opened 2022-09-16
os/binfmt: Add NULL checking to avoid dereference

jeongarmy:
Add NULL checking to avoid dereference


os/binfmt/binfmt_loadbinary.c
```diff
@@ -87,7 +87,11 @@ int load_binary(int binary_idx, FAR const char *filename, load_attr_t *load_attr
 	int ret;
 
 	/* Sanity check */
-	if (load_attr && load_attr->bin_size <= 0) {
+	if (load_attr) {
```
sunghan-chang, line 90: Maybe `!load_attr`, not `load_attr`

sunghan-chang (approved):

PR #5674, opened 2022-09-30
[Draft under review] New master

Tirtha-Samsung:

PR #5675, opened 2022-10-06
Add missing files/functions for Tp1e support to Public master

amogh-samsung:
This PR follows #5674


sunghan-chang:
@amogh-samsung Let's add `.gitignore` in 2nd commit. And could you add the location in 1st commit?
```
build/tools/amebalite/gnu_utility: First release of JLink Debug script
```

os/board/rtl8720e/src/libs/.gitignore
```diff
@@ -1,5 +0,0 @@
-!./lib_enc.a
-!./lib_enc_fpu.a
-!./lib_wps.a
-!./lib_wps_fpu.a
-!./cmse_implib.a
```
sunghan-chang, line 5: They have `!` so that original is correct to keep .a in the git.
I think we should revert this file.

sunghan-chang:
> @amogh-samsung Let's add .gitignore in 2nd commit. And could you add the location in 1st commit?
build/tools/amebalite/gnu_utility: First release of JLink Debug script

How about this?

os/board/rtl8720e/src/libs/.gitignore
amogh-samsung: Sorry, could you explain this comment please.

sunghan-chang: `!` means NOT. Because the .gitignore at the root has `*.a`, we can't push .a file in the git. So, this `!` allows to merged even it is .a file.

PR #5676, opened 2022-10-06
Sync STM32H745 support files to Public Master

amogh-samsung:


os/arch/arm/src/armv7-m/up_hardfault.c
```diff
@@ -85,7 +85,9 @@
```
thapav, line 85: Kindly mention start and end commit ID's in the commit message as well as the PR description

thapav: Along with the branch names*

os/arch/arm/src/armv7-m/up_hardfault.c
```diff
@@ -113,10 +115,17 @@ uint32_t system_exception_location;
...
 	uint32_t *regs = (uint32_t *)context;
-	system_exception_location = regs[REG_R15];
+#endif
+
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	g_assertpc = regs[REG_R15];
+#endif
 	/* Get the value of the program counter where the fault occurred */
 
+	stm32h745_flash_ecc_handler();
```
sunghan-chang, line 127: 1. Except this line, all of other changes should be reverted.
2. This file is a common for all of armv7-m architecture. But this is a ST specific function. Let's cover it with the ST chipset config.

sunghan-chang (approved):

PR #5678, opened 2022-10-12
os/board/rtl/8720e: add FTL write protection disable/enable

yeetee179:
- Added ftl write protection disable/enable to flash erase and write when flash is protected.

### DIFF
--- a/os/board/rtl8720e/src/component/filesystem/ftl/ftl.c
+++ b/os/board/rtl8720e/src/component/filesystem/ftl/ftl.c
@@ -122,6 +122,32 @@ extern bool ftl_page_erase(struct Page_T *p);
 void ftl_mapping_table_init(void);
 uint16_t read_mapping_table(uint16_t logical_addr);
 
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@yeetee179 Is this a write protection for ftl partition only? As far as I know, first 512KB (including bl, ss and so on) is the region of write protection. Could you let me know exact area for write protection?

yeetee179:
> 

Yes the flash protection is for the first 512Kb.

sunghan-chang:
@yeetee179 So, this pr is necessary to disable write protection to write something in ftl partition which is protected?

yeetee179:
@sunghan-chang Yes, the disabling of the flash protection for ftl is temporary. Meaning when erasing or writing is done to the FTL, it will disable its flash protection first, then write/erase it, followed by enabling the flash protection again. This procedure is done every write/erasure cycle.

sunghan-chang (approved):

PR #5679, opened 2022-10-12
os/board/rtl/8720e: Fix BLE scan stop delay issue.

yeetee179:
Issue: BLE scan does not stop immediately, instead with a long delay.
Cause: Time is spent polling for register values when controller reports scan results to host.
Fix: Register reading is not necessary, thus removed.


sunghan-chang:
@ydlee-sec Could you verify BLE scan operation with this PR and merge?

ydlee-sec (approved):
LGTM~

PR #5681, opened 2022-10-12
os/board/rtl8721csm: Add timer to check REG_TCR BIT_CRC(bit 8) every 1 min, if the value is wrong, set reboot reason code and do system reset

gexuyan:
Add timer to check REG_TCR BIT_CRC(bit 8) every 1 min, if the value is wrong, set reboot reason code and do system reset


thapav:
@gexuyan Is the timer change added in the library file ?

gexuyan:
> @gexuyan Is the timer change added in the library file ?

hi @thapav , sorry for the late reply, yes it's added in the lib

j11591 (approved):
LGTM

PR #5682, opened 2022-10-13
libc/math : Add float_t data type

kishore-sn:
Include type definition of float math.h doesn't include type definition of float_t, so add it.
It is required as per posix standard.


jeongchanKim (approved):
LGTM

Taejun-Kwon (approved):

PR #5683, opened 2022-10-21
arch/arm/src: Update assert code to accomodate extra arguments for

thapav:
debugging

THis patch updates the up_assert() API with an extra argument to print information useful for debugging.

Sample assert logs are as below:

TASH>>binary_manager_load_binary: Load success! [Name: app2] [Version: 190412] [Partition: A] [Compressed Binary] Test will be finished
nup_assert: Assertion failed at file:wifiapp.c line: 128 task: app2 assertInfo = TESTSTRING
up_dumpstate: sp:     021837b4
up_dumpstate: Nested IRQ stack:
up_dumpstate:   base: 1002a9b0
up_dumpstate:   size: 00000200
up_dumpstate:   used: 00000000
up_dumpstate: stack base: 02183860
up_dumpstate: stack size: 00001fe4
up_dumpstate: stack used: 00000278
up_stackdump: 021837a0: xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx xxxxxxxx 00000080 100718ac 02257a88
up_stackdump: 021837c0: 02200343 022579cc 022579e4 02257a04 10011e4b 02183828 000000e0 02200343
up_stackdump: 021837e0: 0218380c 02200343 022579cc 022579e4 02257a04 00000000 00000000 00000000
up_stackdump: 02183800: 00000000 0e041b73 00000080 00000080 02200325 02200325 10011ecf 02200343
up_stackdump: 02183820: 022579cc 022579e4 021ff9bf 00000080 02257a88 02200325 02183864 021ff959
up_stackdump: 02183840: 00000001 02183864 00000000 02200471 02183864 0e001c83 00000000 00000000

### DIFF
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -137,6 +137,11 @@ extern int g_irq_nums[3];
 #define IS_FAULT_IN_USER_SPACE(asserted_location)   (is_kernel_space((void *)asserted_location) == false)
 
 /****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
+
+/****************************************************************************
  * Private Data
  ****************************************************************************/
 
@@ -477,6 +482,11 @@ void up_assert(const uint8_t *filename, int lineno)
 #else
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
+
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (assert_info_str[0]) {
+		lldbg("%s\n", assert_info_str);
+	}
 
 #if defined(CONFIG_DEBUG_WORKQUEUE)
 #if defined(CONFIG_BUILD_FLAT) || (defined(CONFIG_BUILD_PROTECTED) && defined(__KERNEL__))
--- a/os/include/assert.h
+++ b/os/include/assert.h
@@ -68,8 +68,10 @@
  ****************************************************************************/
 
 #include <tinyara/compiler.h>
+#include <tinyara/config.h>
 #include <stdint.h>
 
+extern char assert_info_str[CONFIG_STDIO_BUFFER_SIZE];
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -83,6 +85,12 @@
 #undef PANIC					/* Unconditional abort */
 
 #ifdef CONFIG_HAVE_FILENAME
+#define ASSERT_INFO(f, fmt, ...) \
+	{ if (!(f))	{ \
+			  snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
+			  up_assert((const uint8_t *)__FILE__, (int)__LINE__); \
+		} \
+	}
 
 #define ASSERT(f) \
 	{ if (!(f)) up_assert((const uint8_t *)__FILE__, (int)__LINE__); }
@@ -114,6 +122,12 @@
 
 #else
 
+#define ASSERT_INFO(f, fmt, ...) \
+	{ if (!(f))	{ \
+			  snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
+			  up_assert(); \
+			} \
+	}
 /**
  * @brief Assert if the condition is not true
  *
--- a/os/kernel/semaphore/sem_post.c
+++ b/os/kernel/semaphore/sem_post.c
@@ -62,6 +62,7 @@
 #include <sched.h>
 #include <tinyara/arch.h>
 #include <tinyara/sched.h>
+#include <tinyara/mm/mm.h>
 
 #include "sched/sched.h"
 #include "semaphore/semaphore.h"
@@ -199,6 +200,8 @@ int sem_post(FAR sem_t *sem)
 {
 	irqstate_t saved_state;
 	int ret = ERROR;
+	size_t caller_retaddr = 0;
+	ARCH_GET_RET_ADDRESS(caller_retaddr);
 
 	/* Make sure we were supplied with a valid semaphore. */
 
@@ -211,6 +214,7 @@ int sem_post(FAR sem_t *sem)
 		saved_state = irqsave();
 
 		/* Perform the semaphore unlock operation. */
+		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x, caller address = 0x%x", sem, caller_retaddr);
 		ASSERT(sem->semcount < SEM_VALUE_MAX);
 		sem_releaseholder(sem, this_task());
 		sem->semcount++;
--- a/os/arch/arm/src/armv7-r/arm_assert.c
+++ b/os/arch/arm/src/armv7-r/arm_assert.c
@@ -123,6 +123,11 @@ static bool recursive_abort = false;
 bool g_upassert = false;
 
 /****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
+
+/****************************************************************************
  * Private Data
  ****************************************************************************/
 
@@ -920,6 +925,11 @@ void up_assert(const uint8_t *filename, int lineno)
 #else
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
+
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (assert_info_str[0]) {
+		lldbg("%s\n", assert_info_str);
+	}
 
 #if defined(CONFIG_DEBUG_WORKQUEUE)
 #if defined(CONFIG_BUILD_FLAT) || (defined(CONFIG_BUILD_PROTECTED) && defined(__KERNEL__))
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -148,6 +148,11 @@ extern int g_irq_nums[3];
 #define IS_FAULT_IN_USER_SPACE(asserted_location)   (is_kernel_space((void *)asserted_location) == false)
 
 /****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
+
+/****************************************************************************
  * Private Data
  ****************************************************************************/
 
@@ -511,6 +516,10 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (assert_info_str[0]) {
+		lldbg("%s\n", assert_info_str);
+	}
 #if defined(CONFIG_DEBUG_WORKQUEUE)
 #if defined(CONFIG_BUILD_FLAT) || (defined(CONFIG_BUILD_PROTECTED) && defined(__KERNEL__))
 	if (IS_HPWORK || IS_LPWORK) {
--- a/os/arch/xtensa/src/xtensa/xtensa_assert.c
+++ b/os/arch/xtensa/src/xtensa/xtensa_assert.c
@@ -82,6 +82,11 @@
 #endif
 
 /****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
+
+/****************************************************************************
  * Private Functions
  ****************************************************************************/
 
@@ -176,6 +181,11 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 #endif
+
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (assert_info_str[0]) {
+		lldbg("%s\n", assert_info_str);
+	}
 
 	xtensa_assert(EXIT_FAILURE);
 }


os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -431,6 +435,18 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	board_led_on(LED_ASSERTION);
 
+#ifdef CONFIG_CPP_HAVE_VARARGS
+	va_list valist;
+	const char *assertInfo;
+
+	/* Extract the extra debug arguments if any */
+	va_start(valist, lineno);
+	for (int i = 0; i < 10; i++) {
```
jeongchanKim, line 444: Is there any special reason to check 10 times?
If it is, how about declaring the value with proper name?

sunghan-chang:
Is this possible to print only character? Is it possible to do some value?

thapav:
> Is this possible to print only character? Is it possible to do some value?

I will check and update you

os/arch/arm/src/armv7-m/up_assert.c
thapav: Not needed. Updated the code.

thapav:
> Is this possible to print only character? Is it possible to do some value?

Yes, we can print any of these

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -431,6 +435,16 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	board_led_on(LED_ASSERTION);
 
+#ifdef CONFIG_CPP_HAVE_VARARGS
+	va_list valist;
+	const char *assertInfo;
+
+	/* Extract the extra debug arguments if any */
+	va_start(valist, lineno);
+	assertInfo = va_arg(valist, const char *);
```
abhishek-samsung, line 444: defining ```assertInfo``` as ```const``` and then re assigning it might not work. I feel like its better to assign it during declaration.

sunghan-chang:
For example of this change, could you make a new commit to check sem post assert?
https://github.com/Samsung/TizenRT/blob/master/os/kernel/semaphore/sem_post.c#L214
I think we should add two more information, return address to know who calls sem_post function and sem pointer.

thapav:
> For example of this change, could you make a new commit to check sem post assert? https://github.com/Samsung/TizenRT/blob/master/os/kernel/semaphore/sem_post.c#L214 I think we should add two more information, return address to know who calls sem_post function and sem pointer.

To identify that the assert is in sem_post , will need to compare either the filename or the new variable assert_info. If it is okay to use strncmp in up_assert(), then we can go ahead with this commit. I have verified the changes at my end.
Or do you suggest any other method to identify sem_post assertion ?

sunghan-chang:
@thapav I don't get your point fully. We can know the assert which comes from sem_post using assert log as shown below.
```
lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
```
What I want is just printing some more information at `ASSERT`. Currently it just compares some conditionals and makes a crash without why it happens. It makes difficult to debug root cause. 
New commit I asked needs to show printing two more information, return address and sem pointer, at assert.

thapav:
> @thapav I don't get your point fully. We can know the assert which comes from sem_post using assert log as shown below.
> 
> ```
> lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
> ```
> 
> What I want is just printing some more information at `ASSERT`. Currently it just compares some conditionals and makes a crash without why it happens. It makes difficult to debug root cause. New commit I asked needs to show printing two more information, return address and sem pointer, at assert.

Please see if below changes are what is required 
https://github.com/thapav/TizenRT/commit/b06b611e50c505ff41efc482ebf5e60330a4c69d

sunghan-chang:
> Please see if below changes are what is required
https://github.com/thapav/TizenRT/commit/b06b611e50c505ff41efc482ebf5e60330a4c69d

@thapav Not for up_assert.c. I want only the changes in sem_post.c. Because there are so many ASSERT at every module, we can't add some logs in up_assert like you added. The up_assert is enough to print additional values from ASSERT. And then, TRAP needs to convert them as readable sentence as we convert other assert info.
And, I think we need the return address who calls sem_post at ASSERT, sem pointer. But you add tcb info. I think it will give wrong info when irq calls sem_post.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -431,6 +435,15 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	board_led_on(LED_ASSERTION);
 
+#ifdef CONFIG_CPP_HAVE_VARARGS
+	va_list valist;
+
+	/* Extract the extra debug arguments if any */
+	va_start(valist, lineno);
+	const char *assertInfo = va_arg(valist, const char *);
```
kishore-sn, line 443: Hi Vidisha, 
There is an issue with this approach. As of now, the `ASSERT` macro is used many places in our code and we pass only a single condition as a parameter to it. So, the variable arguments list in `up_assert()` will be null. In this situation, if we call `va_arg()`, the behaviour is `undefined`.

So, it is better to create a new assert macro to handle the variable arguments.
Please try to change as follows:

Change existing assert macros as below
`#define ASSERT(f)      { if (!(f)) up_assert((const uint8_t *)__FILE__, (int)__LINE__, ""); }`

Create new macro as :
`#define ASSERT_INFO(f, fmt, ...)     { if (!(f)) up_assert((const uint8_t *)__FILE__, (int)__LINE__, fmt,  ##__VA_ARGS__); }`

Change the `up_assert()` as follows:
```
#ifdef CONFIG_CPP_HAVE_VARARGS
void up_assert(const uint8_t *filename, int lineno, char *fmt, ...)
#else
void up_assert(const uint8_t *filename, int lineno, char *fmt)
#endif
{

    lldbg("Assertion failed at file:%s line: %d task: %s\n", filename, lineno, fault_tcb->name);
    if (fmt) {
        va_list valist;
	/* Extract the extra debug arguments if any */
	va_start(valist, fmt);
        lldbg(fmt, valist);
        va_end(valist);
    }
```

Then, from now on, we can use the new `ASSERT_INFO` macro where ever we need it as follows:
`ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "Sem = 0x%0x count = %d MAX = %d\n", sem, sem->semcount, SEM_VALUE_MAX)`

thapav: @sunghan-chang Does this approach sound fine ?

thapav: Hi Kishore,
As discussed over knox, since this approach of using variable arguments has issues with syscall, let's find another suitable approach for this requirement. Till then we have kept the PR as *do not merge*.

lib/libc/misc/lib_dumpbuffer.c
```diff
@@ -67,6 +67,10 @@
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
+/****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
```
kishore-sn, line 73: I think it would be better to move this declaration to some file related to debug. 
Best place would be `lib/libc/misc/lib_dbg.c`. But, this file is not built always. It depends on `CONFIG_DEBUG` macro. 
So, maybe we can move it to `lib/libc/syslog/lib_syslog.c`

os/include/assert.h
```diff
@@ -83,26 +84,30 @@
 #undef PANIC					/* Unconditional abort */
 
 #ifdef CONFIG_HAVE_FILENAME
-
+#define ASSERT_INFO(f, fmt, ...) \
```
kishore-sn, line 87: Let us add this for `!CONFIG_HAVE_FILENAME` case also.

os/include/assert.h
```diff
@@ -207,8 +212,8 @@ extern "C" {
  * Public Function Prototypes
  ****************************************************************************/
 
-#ifdef CONFIG_HAVE_FILENAME
-void up_assert(FAR const uint8_t *filename, int linenum);
+#if CONFIG_HAVE_FILENAME
+void up_assert(FAR const uint8_t *filename, int linenum, const char *info);
 #else
 void up_assert(void);
```
kishore-sn, line 218: Let us add info argument for this also.

os/include/assert.h
```diff
@@ -207,8 +212,8 @@ extern "C" {
  * Public Function Prototypes
  ****************************************************************************/
 
-#ifdef CONFIG_HAVE_FILENAME
-void up_assert(FAR const uint8_t *filename, int linenum);
+#if CONFIG_HAVE_FILENAME
```
kishore-sn, line 215: Why do we need to change `#ifdef` to `#if`?

os/kernel/semaphore/sem_post.c
```diff
@@ -211,7 +211,7 @@ int sem_post(FAR sem_t *sem)
 		saved_state = irqsave();
 
 		/* Perform the semaphore unlock operation. */
-		ASSERT(sem->semcount < SEM_VALUE_MAX);
+		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x", sem);
```
kishore-sn, line 214: Please modify this to print sem, semcount and MAX value.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -478,6 +478,11 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (*info) {
```
kishore-sn, line 482: I think this needs to be `if (info)`

thapav: Correct, have updated the code

swasthik22 (commented):
LGTM.

os/kernel/semaphore/sem_post.c
```diff
@@ -211,7 +211,7 @@ int sem_post(FAR sem_t *sem)
 		saved_state = irqsave();
 
 		/* Perform the semaphore unlock operation. */
-		ASSERT(sem->semcount < SEM_VALUE_MAX);
+		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x, sem->semcount = %d, SEM_VALUE_MAX = %d", sem);
```
Tirtha-Samsung, line 214: LGTM

Tirtha-Samsung (approved):

kishore-sn (approved):
Looks good.

os/include/assert.h
```diff
@@ -83,26 +84,30 @@
...
+			    up_assert((const uint8_t *)__FILE__, (int)__LINE__, assert_info_str); \
+			} \
+	}
 #define ASSERT(f) \
-	{ if (!(f)) up_assert((const uint8_t *)__FILE__, (int)__LINE__); }
-
+	{ if (!(f)) up_assert((const uint8_t *)__FILE__, (int)__LINE__, ""); }
 #define VERIFY(f) \
-	{ if ((f) < 0) up_assert((const uint8_t *)__FILE__, (int)__LINE__); }
+	{ if ((f) < 0) up_assert((const uint8_t *)__FILE__, (int)__LINE__, ""); }
```
sunghan-chang, line 96: Is this correct with `""`? not NULL?

os/include/assert.h
```diff
@@ -83,26 +84,30 @@
 #undef PANIC					/* Unconditional abort */
 
 #ifdef CONFIG_HAVE_FILENAME
-
+#define ASSERT_INFO(f, fmt, ...) \
+	{ if (!(f))	{ \
+			    snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
+			    up_assert((const uint8_t *)__FILE__, (int)__LINE__, assert_info_str); \
```
sunghan-chang, line 90: Do we need to deliver the string as an argument?
How about printing it at the assert function directly?
Because this buffer is dedicated to the assert.

sunghan-chang: If we do as I suggested, we don't need to modify the up_assert format (arguments).

thapav: Do you mean to directly print semaphore or other required information directly ?

lib/libc/syslog/lib_syslog.c
```diff
@@ -70,6 +70,11 @@
  * Pre-processor Definitions
  ****************************************************************************/
 
+/****************************************************************************
+ * Public Variables
+ ****************************************************************************/
+char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = {'\0', };
```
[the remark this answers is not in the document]
thapav: How about adding it to any arch specific assert file then ?

os/include/assert.h
sunghan-chang: That's enough with snprintf. What I said is about up_assert. I want to keep the original up_assert.
```
#define ASSERT_INFO(f, fmt, ...) \
	{ if (!(f))	{ \
			    snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
			    up_assert((const uint8_t *)__FILE__, (int)__LINE__); \

int up_assert(....)
{
  // printing assert_info_str by accessing global variable directly
}
```

lib/libc/syslog/lib_syslog.c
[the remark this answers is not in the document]
sunghan-chang: Could you let me know where it is?

os/kernel/semaphore/sem_post.c
```diff
@@ -211,6 +211,7 @@ int sem_post(FAR sem_t *sem)
 		saved_state = irqsave();
 
 		/* Perform the semaphore unlock operation. */
+		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x, sem->semcount = %d, SEM_VALUE_MAX = %d", sem);
```
sunghan-chang, line 214: `semcount` and `SEM_VALUE_MAX` are unnecessary to debug this case. The sem address and caller are necessary. In this case, we should know what sem is and who calls like this.

os/arch/arm/src/armv7-m/up_assert.c
```diff
@@ -478,6 +483,11 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg("Assertion failed at file:%s line: %d\n", filename, lineno);
 #endif
 
+	/* Print the extra arguments (if any) from ASSERT_INFO macro */
+	if (assert_info_str) {
```
sunghan-chang, line 487: `assert_info_str` is a global variable so that this if case is always true. You should check whether first character is `\0` or not.

os/include/assert.h
```diff
@@ -70,6 +70,7 @@
 #include <tinyara/compiler.h>
 #include <stdint.h>
 
+extern char assert_info_str[CONFIG_STDIO_BUFFER_SIZE];
```
sunghan-chang, line 73: To use CONFIG_XX, you should include `<tinyara/config.h>`.  Let's include.

os/include/assert.h
```diff
@@ -83,6 +84,8 @@
 #undef PANIC					/* Unconditional abort */
 
 #ifdef CONFIG_HAVE_FILENAME
+#define ASSERT_INFO(f, fmt, ...) \
+	{ if (!(f)) snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); }
```
sunghan-chang, line 88: You miss calling `up_assert`.

os/kernel/semaphore/sem_post.c
thapav: updated

sunghan-chang (approved):

PR #5687, opened 2022-10-31
os/mm: Add caller address to mm alloc fail API

kishore-sn:
When we print information about memory alloc failure, it is good to print the caller address so that during debugging we can immediately find out where the memory request was made. Add the caller address information to all memory failure cases.

Also, use lldbg instead of dbg to print mm fail logs in flat build.

### DIFF
--- a/os/mm/kmm_heap/kmm_memalign.c
+++ b/os/mm/kmm_heap/kmm_memalign.c
@@ -106,13 +106,17 @@ void *kmm_memalign_at(int heap_index, size_t alignment, size_t size)
 	}
 
 	kheap = kmm_get_baseheap();
+	ret = mm_memalign(&kheap[heap_index], alignment, size
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	ret = mm_memalign(&kheap[heap_index], alignment, size, caller_retaddr);
-#else
-	ret = mm_memalign(&kheap[heap_index], alignment, size);
+			, caller_retaddr
 #endif
+			);
 	if (ret == NULL) {
-		mm_manage_alloc_fail(&kheap[heap_index], heap_index, heap_index, size, KERNEL_HEAP);
+		mm_manage_alloc_fail(&kheap[heap_index], heap_index, heap_index, size, KERNEL_HEAP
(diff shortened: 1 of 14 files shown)


os/mm/umm_heap/umm_calloc.c
```diff
@@ -102,7 +102,11 @@ void *calloc_at(int heap_index, size_t n, size_t elem_size)
 	ret = mm_calloc(&BASE_HEAP[heap_index], n, elem_size);
 #endif
 	if (ret == NULL) {
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+		mm_manage_alloc_fail(&BASE_HEAP[heap_index], heap_index, heap_index, n * elem_size, USER_HEAP, caller_retaddr);
```
sunghan-chang, line 106: How about just printing `caller_retaddr` here without giving as an argu?

os/mm/kmm_heap/kmm_memalign.c
```diff
@@ -112,7 +112,11 @@ void *kmm_memalign_at(int heap_index, size_t alignment, size_t size)
 	ret = mm_memalign(&kheap[heap_index], alignment, size);
 #endif
 	if (ret == NULL) {
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
```
sunghan-chang, line 115:
```
#ifdef CONFIG_...
   ret = mm_memalign(...);
   if (ret == NULL) {
      mm_manage_alloc_fail(...);
#else
...
#endif
```
or
```
ret = mm_memalign(.....
#ifdef CONFIG_..
         , caller_retaddr
#endif
        );
if (ret == NULL) {
    mm_manage_alloc_fail(.....
#ifdef CONFIG_...
                        , caller_retaddr
#endif
                         );
```

Just suggestion.

sunghan-chang:
What I want to say in two comments left is how to show the code better in readability point of view.

os/mm/kmm_heap/kmm_memalign.c
kishore-sn: I think .. let us use the second method.

sunghan-chang (approved):

PR #5689, opened 2022-11-01
os/board/rtl8721csm: Add blocking mode and callback API for BLE indication feature

jiangxiaoyunjoy:
Add blocking mode and callback API for BLE indication feature


ydlee-sec (approved):
LGTM.

PR #5692, opened 2022-11-04
os/board/rtl8721csm: Modify do-while condition check in BLE ADV start API

jiangxiaoyunjoy:
Modify do-while condition check in BLE ADV start API

### DIFF
--- a/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/example/ble_peripheral/ble_tizenrt_server.c
+++ b/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/example/ble_peripheral/ble_tizenrt_server.c
@@ -590,7 +590,7 @@ trble_result_e rtw_ble_server_start_adv(void)
         debug_print("Waiting for adv start \n");
         os_delay(100);
         le_get_gap_param(GAP_PARAM_DEV_STATE , &new_state);
-    } while(new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING);
+    } while((new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING) && ((new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN)&&(new_state.gap_adv_state != GAP_ADV_STATE_IDLE)));
 
     return TRBLE_SUCCESS;
 }


ydlee-sec:
Shouldn't it be changed as below?
```} while((new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING) && ((new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN)||(new_state.gap_adv_state != GAP_ADV_STATE_IDLE)));```

Current commit returns ```TRBLE_SUCCESS``` when ```ble_tizenrt_server_send_msg(BLE_TIZENRT_MSG_START_ADV, NULL)``` doesn't work.

**Comment by jiangxiaoyunjoy** [comment]
> @jiangxiaoyunjoy
> 
> Shouldn't it be changed as below?
> 
> `} while((new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING) && ((new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN)||(new_state.gap_adv_state != GAP_ADV_STATE_IDLE)));`
> 
> Please check the condition below
> 
> ```
> new_state.gap_adv_state == GAP_ADV_STATE_IDLE
> new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN
> ```
> 
> It returns `TRBLE_SUCCESS`, is it correct ?

1. If there is no external operation over 100ms, adv should start success.
new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING -> false, adv state should be advertising
new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN -> true, as no connection built 
new_state.gap_adv_state != GAP_ADV_STATE_IDLE -> true
do-while check false, break and return ret=0.(success)

2. If a master tries to connect TP1x within 100ms, connection built and new_state.gap_adv_state changed to idle.
new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING -> true
new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN -> false, as connection built
new_state.gap_adv_state != GAP_ADV_STATE_IDLE -> false
do-while check false, break and return ret=0. (success)
adv start success and build connection with other devices, so we think this adv_start operation success under such circumstance.

**Comment by ydlee-sec** [comment]
@jiangxiaoyunjoy 
I am concerned that the original source code operation will be disabled.
Isn't the code below WAITING for ACTUAL advertise to work?
```
    do
    {   
        debug_print("Waiting for adv start \n");
        os_delay(100);
        le_get_gap_param(GAP_PARAM_DEV_STATE , &new_state);
    } while(new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING);
```

After applying the commit, the WAITING  effect will be disabled. 
Before the actual operation is performed, each variable will be as follows.
- new_state.gap_adv_state = GAP_ADV_STATE_IDLE
- new_state.gap_adv_sub_state = GAP_ADV_TO_IDLE_CAUSE_STOP

And the conditions for while loop will be like below
new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING -> true
new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN -> true
new_state.gap_adv_state != GAP_ADV_STATE_IDLE -> false

It will returns success without waiting, is it correct?

**Comment by jiangxiaoyunjoy** [comment]
> @jiangxiaoyunjoy I am concerned that the original source code operation will be disabled. Isn't the code below WAITING for ACTUAL advertise to work?
> 
> ```
>     do
>     {   
>         debug_print("Waiting for adv start \n");
>         os_delay(100);
>         le_get_gap_param(GAP_PARAM_DEV_STATE , &new_state);
>     } while(new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING);
> ```
> 
> After applying the commit, the WAITING effect will be disabled. Before the actual operation is performed, each variable will be as follows.
> 
> * new_state.gap_adv_state = GAP_ADV_STATE_IDLE
> * new_state.gap_adv_sub_state = GAP_ADV_TO_IDLE_CAUSE_STOP
> 
> And the conditions for while loop will be like below new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING -> true new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN -> true new_state.gap_adv_state != GAP_ADV_STATE_IDLE -> false
> 
> It will returns success without waiting, is it correct?

Hi, sorry for the late reply.

Base on my test results, both solutions achieve the same result.
When connection built within the delay, start adv return success.
When adv start successfully, do-while check false, start adv return success.

For my commit, I focus on the case that connection built within the adv delay. 
For your solution, it points out that when adv status is idle or start, do-while still check false and continue without delay.

How about change to 
while((new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING) && (new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN))

BTW, it's better to add rtw_ble_server_stop_adv(...) before changing the adv setting, and then start adv again. Although we have status check in rtw_ble_server_start_adv(...), but still recommend using stop before start in application layer.

Thank you~

**Comment by ydlee-sec** [comment]
@jiangxiaoyunjoy 
Thank you for your comments.
```
while((new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING) 
   && (new_state.gap_adv_sub_state != GAP_ADV_TO_IDLE_CAUSE_CONN))
```
Regarding to the new suggestion above, I will follow your decision as the source code owner. 

please push a new commit that contains new changes 
 or 
Add a comment to use the current PR commit only.

Thank you.

jiangxiaoyunjoy:
will push a new commit

PR #5693, opened 2022-11-08
os/board/rtl8721csm: Modify do-while check in rtw_ble_server_start_adv(...) API

jiangxiaoyunjoy:
The modification can handle the scenarios listed as below:

1. If there is no external operation over 100ms, adv start success.
nRF is able to scan the device.
2. If a master tries to connect TP1x within 100ms, connection built and new_state.gap_adv_state changed to idle. 
adv start success and build connection with other devices, so we think this adv_start operation success under such case. nRF will unable to see device after return success.
3. If a master tries to connect TP1x within 100ms, connection built but disconnect within 100ms as well. 
we think this adv_start operation success under such case. But nRF will unable to see TP1x in this case. Customers need to check log to debug.

Non-connectable cases:
1. start non-connectable adv as the first adv -> return success
2. start connectable adv as the first adv, then start non-connectable adv as the second adv -> return success
3. start connectable adv as the first adv, build the connection, then start non-connectable adv as the second adv -> return success
4. start connectable adv as the first adv, build the connection, then start connectable adv as the second adv -> return fail

### DIFF
--- a/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/example/ble_peripheral/ble_tizenrt_server.c
+++ b/os/board/rtl8721csm/src/component/common/bluetooth/realtek/sdk/example/ble_peripheral/ble_tizenrt_server.c
@@ -24,6 +24,7 @@ uint16_t server_profile_count = 0;
 trble_server_init_config server_init_parm;
 bool (*ble_tizenrt_server_send_msg)(uint16_t sub_type, void *arg) = NULL;
 T_SEND_DATA_RESULT *send_indication_result = NULL;
+extern T_GAP_DEV_STATE ble_tizenrt_scatternet_gap_dev_state;
 
 trble_result_e rtw_ble_server_init(trble_server_init_config* init_parm)
 {
@@ -597,6 +598,11 @@ trble_result_e rtw_ble_server_start_adv(void)
     {   
         debug_print("Waiting for adv start \n");
         os_delay(100);
+        if (ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_CONN)
+        {
+            ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state = GAP_ADV_TO_IDLE_CAUSE_STOP;
+            break;
+        }
         le_get_gap_param(GAP_PARAM_DEV_STATE , &new_state);
     } while(new_state.gap_adv_state != GAP_ADV_STATE_ADVERTISING);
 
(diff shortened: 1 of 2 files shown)


ydlee-sec:
@jiangxiaoyunjoy 

What is this ? (==)
```
ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_STOP;
```

jiangxiaoyunjoy:
> @jiangxiaoyunjoy
> 
> What is this ? (==)
> 
> ```
> ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_STOP;
> ```

We decide to use a global variable ble_tizenrt_scatternet_gap_dev_state to trace the connection within the delay case but not the previous solution.
 
gap_adv_sub_state has two states: GAP_ADV_TO_IDLE_CAUSE_STOP and GAP_ADV_TO_IDLE_CAUSE_CONN.
gap_adv_sub_state will be mark as GAP_ADV_TO_IDLE_CAUSE_CONN when advertising is stopped for link establishment, 
and it will be turned to GAP_ADV_TO_IDLE_CAUSE_STOP when adv stopped by user stop or high duty cycle directed advertising timeout. 

However, gap_adv_sub_state will only be reset to 0 by lower layer with adv_stop cmd or adv timeout adv. This property may cause do-while check fail in the non-connectable case. Thus, we put a manual reset here to indicate the current adv is stopped, so that the non-connectable adv can be started smoothly.

Thank you~

ydlee-sec:
@jiangxiaoyunjoy  I mean..
```
ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_STOP;
```
Vs.
```
ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state = GAP_ADV_TO_IDLE_CAUSE_STOP;
```

jiangxiaoyunjoy:
> @jiangxiaoyunjoy I mean..
> 
> ```
> ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_STOP;
> ```
> 
> Vs.
> 
> ```
> ble_tizenrt_scatternet_gap_dev_state.gap_adv_sub_state = GAP_ADV_TO_IDLE_CAUSE_STOP;
> ```

Thank you! The error has been fixed.

Please note that for the connectable adv cases 2 and 3:
2. If a master tries to connect TP1x within 100ms, connection built and new_state.gap_adv_state changed to idle.
adv start success and build connection with other devices, so we think this adv_start operation success under such case. nRF will unable to see device after return success.
3. If a master tries to connect TP1x within 100ms, connection built but disconnect within 100ms as well.
we think this adv_start operation success under such case. But nRF will unable to see TP1x in this case. Customers need to check log to debug.

The design idea is, as long as adv status has been changed to advertising and has adv packets in the air, the return result will be success. And the following connect or disconnect behaviors won't change the return result.
Customers need to check connected/ disconnected callback log to debug. 

Thank you~

ydlee-sec (approved):
LGTM

PR #5702, opened 2022-11-24
apps/system/utils/fscmd.c: improve tash format command

swasthik22:
This patch is to update on how the format command works.

-  Current format logic is to format all the partitions till and including 9th partition.
- The userfs partition is found by parsing the FLASH_PART_NAME / SECOND_FLASH_PART_NAME from the config and only that partition is formatted. Renaming the function format_filesystem() to format_userfs()
- With the current change, only the userfs partition of the filesystem will be formatted when the tash format command is invoked.

### DIFF
--- a/apps/system/utils/fscmd.c
+++ b/apps/system/utils/fscmd.c
@@ -103,11 +103,12 @@
 #define CMD_FAILED_ERRNO  "%s : %s failed: %d\n"
 
 /** Define FS Type **/
-#define NONEFS_TYPE     "None FS"
-#define SMARTFS_TYPE    "smartfs"
-#define PROCFS_TYPE     "procfs"
-#define ROMFS_TYPE      "romfs"
-#define TMPFS_TYPE      "tmpfs"
+#define NONEFS_TYPE        "None FS"
+#define SMARTFS_TYPE       "smartfs"
+#define PROCFS_TYPE        "procfs"
+#define ROMFS_TYPE         "romfs"
+#define TMPFS_TYPE         "tmpfs"
+#define USERFS_PART_NAME   "userfs"
 
 /****************************************************************************
  * Private Types
@@ -136,6 +137,26 @@ static void fscmd_free(FAR char *path)
 		free(path);
 	}
 }
+
+
+#if defined(CONFIG_SECOND_FLASH_PART_NAME) || defined(CONFIG_FLASH_PART_NAME)
+static int get_partition_index(char *partname, char *part)
+{
+	int partno = 0;
+	int name_len = strlen(part);
+
+	while (*partname) {
+		if (!strncmp(partname, part, name_len)) {
+			if (*(partname + name_len) == ',') {
+				return partno;
+			}
+		}
+		partno++;
+		while (*(partname)++ != ',');
+	}
+	return ERROR;
+}
+#endif
 
 #ifndef CONFIG_DISABLE_ENVIRON
 /****************************************************************************
@@ -1347,7 +1368,7 @@ enum fs_minor_num_e {
 };
 typedef enum fs_minor_num_e fs_minor_t;
 
-static int format_filesystem(fs_minor_t minor)
+static int format_userfs(fs_minor_t minor)
 {
 	char name[CONFIG_PATH_MAX];
 	int fd;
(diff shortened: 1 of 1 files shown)


apps/system/utils/fscmd.c
```diff
@@ -103,11 +103,12 @@
 #define CMD_FAILED_ERRNO  "%s : %s failed: %d\n"
```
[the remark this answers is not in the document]
amogh-samsung: As a description, please add, in points, what was the need of the change, what change was made and what is the effect of the change.

apps/system/utils/fscmd.c
```diff
@@ -137,6 +138,26 @@ static void fscmd_free(FAR char *path)
 	}
 }
 
+
+#if defined(CONFIG_SECOND_FLASH_PART_NAME) || defined(CONFIG_FLASH_PART_NAME)
+static int get_partition_index(char *partname,char *part)
```
Taejun-Kwon, line 143: This will not work in TizenRT4.0 that's why I implemented as a current code.

ewoodev:
You should check code rule. Put a space after "char *partname," 143 line.

sunghan-chang:
> imporove -> improve

@swasthik22 You modified only PR title. But you should do commit title.

apps/system/utils/fscmd.c
```diff
@@ -137,6 +138,26 @@ static void fscmd_free(FAR char *path)
...
 }
 
+
+#if defined(CONFIG_SECOND_FLASH_PART_NAME) || defined(CONFIG_FLASH_PART_NAME)
+static int get_partition_index(char *partname, char *part)
+{
+	int partno = 0;
+	int name_len = strlen(part);
+
+	while (*partname) {
```
hdotkashyap, line 148: Please validate *partname before use

Taejun-Kwon:
App should not have any of dependency regarding config of system, we access it if there's no other way.
Especially, FLASH_PART_NAME / SECOND_FLASH_PART_NAME is not part file system, it is config values of board.
And logic is very similar essentially, previous code try open in loop, and this approach strcmp in loop. 
I don't think there's significant improvement, but dependency newly added.

apps/system/utils/fscmd.c
```diff
@@ -1357,20 +1378,36 @@ static int format_filesystem(fs_minor_t minor)
...
+#endif
+	}
+	if (index == ERROR) {
+		printf("Userfs partition not found, index : %d, minor : %d", index, minor);
+		return ERROR;
+	}
+	snprintf(name, sizeof(name), "/dev/smart%dp%d", minor, index);
+	fd = open(name, O_RDWR);
+	if (fd < 0) {
+		return ret;
```
jeongchanKim, line 1404: I think it is better to print the log about failure.

apps/system/utils/fscmd.c
```diff
@@ -1357,20 +1378,36 @@ static int format_filesystem(fs_minor_t minor)
...
+	if (minor == FS_BLOCK_MINOR_SECONDARY) {
+#ifdef CONFIG_SECOND_FLASH_PART_NAME
+			index = get_partition_index(CONFIG_SECOND_FLASH_PART_NAME, USERFS_PART_NAME);
+#else
+			printf("config CONFIG_SECOND_FLASH_PARTITION_NAME must be enabled to format\n");
+			return ERROR;
+#endif
+	} else {
+#ifdef CONFIG_FLASH_PART_NAME
+		index = get_partition_index(CONFIG_FLASH_PART_NAME, USERFS_PART_NAME);
```
amogh-samsung, line 1391: TJ's concern here is that using a kernel config (flash partition list) to complete a fs related TASH command is not desirable.
However, the current implementation loops over all partitions and tries to format them.
One additional problem is that the loop is limited to 10 while in some cases the userfs partition's index might be greater than 9.
So we want to find a middle ground for this, what is your idea @swasthik22

apps/system/utils/fscmd.c
```diff
@@ -1357,20 +1378,36 @@ static int format_filesystem(fs_minor_t minor)
...
-		ret = ioctl(fd, BIOC_BULKERASE, 0);
-		close(fd);
-		if (ret != OK) {
-			printf("Low level format failed ret : %d errno : %d", ret, errno);
-			return ret;
-		}
-		break;
+	int index = ERROR;
+	if (minor == FS_BLOCK_MINOR_SECONDARY) {
+#ifdef CONFIG_SECOND_FLASH_PART_NAME
```
amogh-samsung, line 1383: This config dependency, similar to the comment on line 1391

apps/system/utils/fscmd.c
amogh-samsung, line 103: @swasthik22 we can close this PR for now.
Let's not keep it hung up for long.

PR #5703, opened 2022-11-25
os/board/rtl8720e/src/component/wifi/: Update Wificsi API

zhongnuo-tang:
RTL8720E WIFI CSI API updates:
1. Insert CSI header information before CSI buffer, so no more reports of CSI header information via API
2. Add CSI buffer layout and CSI header information format
3. Add new parameters: csi_signature=0xABCD, hdr_len, csi_sequence in CSI header information
4. Modify variable name: from csi_raw_data_len to csi_data_len in CSI example code


sunghan-chang:
@zhongnuo-tang Thank you for the PR. But this is the public GitHub and we can't analyse the change always. Even we do, it's difficult to know what it is. Please leave what the PR is in the commit description. bug fix? api change? additional api support?

zhongnuo-tang:
> @zhongnuo-tang Thank you for the PR. But this is the public GitHub and we can't analyse the change always. Even we do, it's difficult to know what it is. Please leave what the PR is in the commit description. bug fix? api change? additional api support?

Dear @sunghan-chang,
I have made the relevant changes, added commit description. Thank you.

sunghan-chang:
@zhongnuo-tang Thank you very much.

os/include/tinyara/wifi/rtk/wifi_constants.h
```diff
@@ -633,10 +633,12 @@ enum _WIFI_EVENT_INDICATE{
 	WIFI_EVENT_IP_CHANGED = 16,
 	WIFI_EVENT_ICV_ERROR = 17,
 	WIFI_EVENT_CHALLENGE_FAIL = 18,
-	
+
+#ifdef CONFIG_ARCH_CHIP_RTL8720E
 		/* csi rx done event */
-	WIFI_EVENT_CSI_DONE = 160,
-	
+	WIFI_EVENT_CSI_DONE = 15,
```
sunghan-chang, line 639: 15 is duplicated. Is it ok?

zhongnuo-tang: Dear @sunghan-chang,

Yes, as RTL8720E does not have WIFI_EVENT_BEACON_AFTER_DHCP  event anymore, and tp1x don't have WIFICSI. Will it be better if i change it to this 

#ifdef CONFIG_ARCH_CHIP_RTL8720E
  		/* csi rx done event */
  	WIFI_EVENT_CSI_DONE = 15,
#else
	WIFI_EVENT_BEACON_AFTER_DHCP = 15,
#endif

sunghan-chang: Suggestion is better I think. And please leave your explanation there.

sunghan-chang: One question, if some other chip needs both, how will you do? No need to concern this? Is there any problem with previous value 160? or other no-duplicated value?

zhongnuo-tang: Changing it from 160 to 15 is for codesize cost down, and also synchronization with our internal sdk. Choosing other no-duplicated value might cause conflicts in the future when there are more events added in. Do you have any better suggestions for this?

Taejun-Kwon: > Dear @sunghan-chang,
> 
> Yes, as RTL8720E does not have WIFI_EVENT_BEACON_AFTER_DHCP event anymore, and tp1x don't have WIFICSI. Will it be better if i change it to this
> 
> #ifdef CONFIG_ARCH_CHIP_RTL8720E /* csi rx done event */ WIFI_EVENT_CSI_DONE = 15, #else WIFI_EVENT_BEACON_AFTER_DHCP = 15, #endif

In my opinion, even if it working exclusively, you should define different value.
Please assume that there's framework for both of tp1x & tp1e, if you define same value, then framework also should have a kind of configuration..below is more good approach
```
1. framework
void do_something_with_csi()
{
    ret = send_message(WIFI_EVENT_CSI_DONE);
    if (ret != -ENOSYS) { //board doesn't provide it
           return ret;
    }
    do_something;
}
void do_something_with_dhcp()
{
    ret = send_message(WIFI_EVENT_BEACON_AFTER_DHCP);
    if (ret != -ENOSYS) { //board doesn't provide it
           return ret;
    }
    do_something;
}
2. BSP
void send_message(WIFI_EVENT_INDICATE event)
{
#ifdef CONFIG_ARCH_CHIP_RTL8720E
     if (event == WIFI_EVENT_BEACON_AFTER_DHCP) {
          return -ENOSYS;
     }
#endif

#ifdef CONFIG_ARCH_CHIP_RTL8721
    if (event == WIFI_EVENT_CSI_DONE) {
          return -ENOSYS;
    }
#endif
   do_something;
}

3.enum
enum _WIFI_EVENT_INDICATE {
	WIFI_EVENT_BEACON_AFTER_DHCP = 15,
         ....
        WIFI_EVENT_CSI_DONE = 19,
         ....
}

File: os/include/tinyara/wifi/rtk/wifi_constants.h [hunk 2]
```diff
@@ -633,10 +633,10 @@ enum _WIFI_EVENT_INDICATE{
 	WIFI_EVENT_IP_CHANGED = 16,
 	WIFI_EVENT_ICV_ERROR = 17,
 	WIFI_EVENT_CHALLENGE_FAIL = 18,
-	
-		/* csi rx done event */
-	WIFI_EVENT_CSI_DONE = 160,
-	
+  /* csi rx done event */
+  WIFI_EVENT_CSI_DONE = 19,
```
sunghan-chang, line 637: Could you align this as others? We use a tab for indentation. Because this is not in the board (or chip) specific folders, we should follow TizenRT coding rules.

os/board/rtl8720e/src/component/wifi/driver/include/wifi_constants.h
```diff
@@ -698,7 +698,7 @@ typedef enum {
 	WIFI_EVENT_WPA_EAPOL_RECVD,
 
 	/* csi rx done event */
-	WIFI_EVENT_CSI_DONE,
+	WIFI_EVENT_CSI_DONE = 19,
```
sunghan-chang, line 701: Does this not show build warning about duplicated definition? Could you let me know why this is necessary even this is defined in os/include/tinyara/wifi/rtk/wifi_constants.h?

zhongnuo-tang: in example_wifi_csi.c, the function wifi_reg_event_handler(WIFI_EVENT_CSI_DONE, example_wifi_csi_report_cb, NULL); use WIFI_EVENT_MAX in this file as the max event, so if i don't define it here, the WIFI_EVENT_MAX will be treated as 16, and it will not register this event.

zhongnuo-tang: Dear @sunghan-chang ,
in example_wifi_csi.c, the function wifi_reg_event_handler(WIFI_EVENT_CSI_DONE, example_wifi_csi_report_cb, NULL); use WIFI_EVENT_MAX in this file as the max event, so if i don't define it here, the WIFI_EVENT_MAX will be treated as 16, and it will not register this event.

os/include/tinyara/wifi/rtk/wifi_constants.h
zhongnuo-tang: Dear @sunghan-chang,
Made the relevant changes.

os/board/rtl8720e/src/component/wifi/driver/include/wifi_constants.h
sunghan-chang: It means you don't use `os/include/tinyara/wifi/rtk/wifi_constants.h` in the app. That uses this header. Why does that example use internal header, not public header?

zhongnuo-tang: Dear @sunghan-chang,
Sorry it was my mistake, i should not include **os/include/tinyara/wifi/rtk/wifi_constants.h** in this example. This example is a demonstration on how to use wificsi related API and it is calling our internal driver directly so it uses internal header. I will revert the changes in **os/include/tinyara/wifi/rtk/wifi_constants.h,** and define a hardcoded value **WIFI_EVENT_CSI_DONE**   in **example_wifi_csi.c** to pass into **wifi_reg_event_handler,** .  In this case, i will use the **WIFI_EVENT_CSI_DONE** value defined in our internal header, is this ok?

sunghan-chang: @zhongnuo-tang It depends on your API implementation. When we use CSI APIs, do we need to use that definition? If yes, it should be defined in public header, not in internal, and the example should use the public. If upper layer is no need to use it, then it should be in internal.

sunghan-chang: @zhongnuo-tang Anyway duplicate two same header could cause some missing. I suggest removing one and using the same header at all of codes.

zhongnuo-tang: Dear @sunghan-chang,  
From my understanding, when registering CSI event, it will use api in 8720e/src/component/wifi/api, which refers to internal header. In order to refer to public header, will need Samsung to come up with the framework first. So is it possible to merge this PR first, then when the framework is ready, then modify the example to use public header, is this okay? Thank you.

sunghan-chang: @zhongnuo-tang Because a framework is a different module and a layer to the Realtek CSI module (BSP), we need any some header to call them. What should we use for that?

sunghan-chang: @zhongnuo-tang Once I check a proper header, I will merge this as it is. But later we could ask you to tidy-up the headers.

zhongnuo-tang: @sunghan-chang Okay thank you!

sunghan-chang (approved):

PR #5705, opened 2022-11-29
os/include/debug.h: Change the conditionals for mfdbg from dbg to lldbg

swasthik22:
This patch is for change in mfdbg to use lldbg instead of dbg in flat build
The logs get stuck for flat build, when there is heap corruption.
Change the mfdbg to use lldbg.
With the change the logs are getting displayed correctly.

LOGS BEFORE 

System Information:
Version:
Platform: 3.1 up_assert: Assertion failed at file:hello_main.c line: 83 task: appmn
up_dumpstate: Code asserted in normal thread!
up_dumpstate: Current SP is User Thread SP: 1006e6c8
........
TASH>>
TASH>>Hello, World!!
mm_check_heap_corruption: ###########################################################################
mm_check_heap_corruption: ERROR: Heap n

LOGS AFTER CHANGE

up_assert: Assertion failed at file:hello_main.c line: 83 task: hello
.......
mm_check_heap_corruption: #########################################################################################
mm_check_heap_corruption: ERROR: Heap node corruption detected.
mm_check_heap_corruption: ==================================================================================
mm_check_heap_corruption: Possible corruption scenario 1:
mm_check_heap_corruption: ...........

### DIFF
--- a/os/include/debug.h
+++ b/os/include/debug.h
@@ -499,8 +499,11 @@ int get_errno(void);
 #define mllvdbg(...)
 #endif
 
-#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
-#define mfdbg(format, ...) lldbg(format, ##__VA_ARGS__)
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
+#define mfdbg(format, ...) if (abort_mode || up_interrupt_context()) \
+                                                             lldbg(format, ##__VA_ARGS__); \
+                                                        else \
+                                                             dbg(format, ##__VA_ARGS__);
 #else
 #define mfdbg(format, ...) dbg(format, ##__VA_ARGS__)
 #endif
--- a/os/mm/mm_heap/mm_check_heap_corruption.c
+++ b/os/mm/mm_heap/mm_check_heap_corruption.c
@@ -77,7 +77,7 @@ enum node_type_e {
 };
 typedef enum node_type_e node_type_t;
 
-#if defined(__KERNEL__)
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
 extern bool abort_mode;
 #endif
 /****************************************************************************
--- a/os/arch/xtensa/src/xtensa/xtensa_assert.c
+++ b/os/arch/xtensa/src/xtensa/xtensa_assert.c
@@ -72,6 +72,8 @@
 #include "sched/sched.h"
 #include "xtensa.h"
 
+bool abort_mode = false;
+
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
@@ -167,6 +169,8 @@ static void xtensa_assert(int errorcode)
 void up_assert(const uint8_t *filename, int lineno)
 {
 	board_autoled_on(LED_ASSERTION);
+
+	abort_mode = true;
 
 #if defined(CONFIG_DEBUG)
 #if CONFIG_TASK_NAME_SIZE > 0 && defined(CONFIG_DEBUG_ERROR)
--- a/os/mm/mm_heap/mm_manage_allocfail.c
+++ b/os/mm/mm_heap/mm_manage_allocfail.c
@@ -94,6 +94,11 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 {
 	irqstate_t flags = irqsave();
 
+	extern bool abort_mode;
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+	abort_mode = true;
+#endif
+
 	mfdbg("Allocation failed from %s heap.\n", (heap_type == KERNEL_HEAP) ? KERNEL_STR : USER_STR);
 	mfdbg(" - requested size %u\n", size);
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
@@ -113,8 +118,6 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	WRITE_REBOOT_REASON(REBOOT_SYSTEM_MEMORYALLOCFAIL);
 #endif
 
-	extern bool abort_mode;
-	abort_mode = true;
 #endif /* CONFIG_MM_ASSERT_ON_FAIL */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO


sunghan-chang:
1. I don't want what the problem is. Where is the log stucked?
2. Even this is necessary for flat build, could do you remove existed conditionals?

swasthik22:
> 1. I don't know what the problem is. Where is the log stucked?
> 2. Even this is necessary for flat build, why do you remove existed conditionals?

`

 {

        printf("Hello, World!!\n");

+       int a = 5 / 0;

`
The problem is that, when there is heap curruption, the board crashes, however due to the

swasthik22:
> 1. I don't know what the problem is. Where is the log stucked?
> 2. Even this is necessary for flat build, why do you remove existed conditionals?

When there is heap corruption and the board crashes, in the up_assert we were using the lldbg, but for mfdbg we were using dbg,
and the logs were getting stuck on crash. So using lldbg for mfdbg as well so that there is no loss of logs.

mm_check_heap_corruption: ###########################################################################
mm_check_heap_corruption: ERROR: Heap n

--- there are more logs here but the logs are stuck, which are visible with lldbg.

sunghan-chang:
@swasthik22 Could you let me know a reply of 2nd comment?
```
2. Even this is necessary for flat build, why do you remove existed conditionals?
```

sunghan-chang:
> When there is heap corruption and the board crashes, in the up_assert we were using the lldbg, but for mfdbg we were using dbg,
and the logs were getting stuck on crash. So using lldbg for mfdbg as well so that there is no loss of logs.
> 
> mm_check_heap_corruption: ###########################################################################
mm_check_heap_corruption: ERROR: Heap n
> 
> --- there are more logs here but the logs are stuck, which are visible with lldbg.

Could you leave where the log is stucked in the commit description like below?
```
mm_check_heap_corruption: ###########################################################################
mm_check_heap_corruption: ERROR: Heap n (no more logs from here)
```

sunghan-chang:
> @swasthik22 Could you let me know a reply of 2nd comment?
> 
> 2. Even this is necessary for flat build, why do you remove existed conditionals?

Please verify this with 4 cases
1. flat build with CONFIG_MM_ASSERT_ON_FAIL
2. flat build without CONFIG_MM_ASSERT_ON_FAIL
3. loadable build with CONFIG_MM_ASSERT_ON_FAIL
4. loadable build without CONFIG_MM_ASSERT_ON_FAIL

ewoodev:
If CONFIG_MM_ASSERT_ON_FAIL is turned on, the board crashes. 
so we should use lldbg.

However, If It is turned off, It don't crash. so use dbg.

abhishek-samsung:
> If CONFIG_MM_ASSERT_ON_FAIL is turned on, the board crashes. so we should use lldbg.
> 
> However, If It is turned off, It don't crash. so use dbg.

In up_assert, we are checking heap corruption which in turn calls mfdbg,

https://github.com/Samsung/TizenRT/blob/24d2156157e99ad78d279df8f00239a3cd90dedb/os/arch/arm/src/armv8-m/up_assert.c#L530

https://github.com/Samsung/TizenRT/blob/24d2156157e99ad78d279df8f00239a3cd90dedb/os/mm/mm_heap/mm_check_heap_corruption.c#L181-L185

so if we call PANIC, this causes the issue as up_assert uses lldbg but mfdbg uses dbg currently. This will cause the logs to get stuck. This issue wont happen during crashes because dbg behaves as lldbg during irq handling. So this only occurs when we call PANIC in a running task

os/include/debug.h
```diff
@@ -499,7 +499,7 @@ int get_errno(void);
 #define mllvdbg(...)
 #endif
 
-#if defined(CONFIG_MM_ASSERT_ON_FAIL) && defined(CONFIG_APP_BINARY_SEPARATION) && defined (__KERNEL__)
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
 #define mfdbg(format, ...) lldbg(format, ##__VA_ARGS__)
```
kishore-sn, line 503: `lldbg` must be used only in low level code and irq handler or assert handler. We would like to avoid using lldbg in all other places. So, how about we change this as below?

```
#define mfdbg(format, ...)            if (abort_mode || up_interrupt_context()) \
                                                             lldbg(format, ##__VA_ARGS__); \
                                                        else \
                                                             dbg(format, ##__VA_ARGS__);
```

sunghan-chang: When CONFIG_MM_ASSERT_ON_FAIL is enabled, mfdbg can be called at allocation failure before assert. Then, is it possible to print all of logs using above?

abhishek-samsung: https://github.com/Samsung/TizenRT/blob/589e026d2f9467362b346fd490d240e9d9a420a1/os/mm/mm_heap/mm_manage_allocfail.c#L116-L118

we are setting abort_mode to true in mm_manage_allocfail, we can move the assignment to the start of the function. Then the above definition of mfdbg works.

sunghan-chang: ok. then, it will be the same as `heapinfo_dbg`.

sunghan-chang: But need to check the conditionals. The heapinfo_dbg has `#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)`. I don't remember the reason. When you check done, could you leave the reason why the conditionals are necessary using comment for future?

kishore-sn: Yes. We still need to check conditionals `#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)`.
It is because `abort_mode` variable is only defined in kernel binary. 
If it is app side binary (with binary separation) , we will always use `dbg`.

os/include/debug.h
```diff
@@ -499,7 +499,7 @@ int get_errno(void);
 #define mllvdbg(...)
```
Tirtha-Samsung, line 499: Please change "This patch is for change in mfdbg to use lldbg instead of dbg in flat build" to "This patch is to change mfdbg to use lldbg instead of dbg in flat build"

sunghan-chang:
@swasthik22 Could you let me know ETA for this?

os/include/debug.h
```diff
@@ -358,7 +358,7 @@ int get_errno(void);
 #else
 #  define bwarn(...)
 #endif
-
+:
```
abhishek-samsung, line 361: could you remove this?

swasthik22: yeah, doing it.

swasthik22:
**flat build without CONFIG_MM_ASSERT_ON_FAIL**

> Printing from thread 420 
> Printing from thread 421 
> Printing from thread 4mm_manage_alloc_fail: Allocation failed from user heap.
> mm_manage_alloc_fail:  - requested size 4000000000
> 22 
> Printing from thread 423 
> Printing from thread 424 

**flat build with CONFIG_MM_ASSERT_ON_FAIL**

> Printing from thread 420 
> Printing from thread 421 
> Printimm_manage_alloc_fail: Allocation failed from user heap.
> mm_manage_alloc_fail:  - requested size 4000000000
> mm_manage_alloc_fail:  - largest free size : 3924064
> mm_manage_alloc_fail:  - total free size   : 3957424
> up_assert: Assertion failed at file:mm_heap/mm_manage_allocfail.c line: 128 task: appmain
> up_dumpstate: Code asserted in normal thread!
> up_dumpstate: Current SP is User Thread SP: 1006e670
> up_dumpstate: User stack:
> up_dumpstate:   base: 1006e728
> up_dumpstate:   size: 000007ec
> (more logs here skipped due to size)

swasthik22:
**Loadable with CONFIG_MM_ASSERT_ON_FAIL**

> Printing from thread 428 
> Printing from thread 429 
> Printing from thread 430 
> mm_manage_alloc_fail: Allocation failed from user heap.
> mm_manage_alloc_fail:  - requested size 4000000000
> mm_manage_alloc_fail:  - caller address = 0x02238c7d
> mm_manage_alloc_fail:  - largest free size : 481584
> mm_manage_alloc_fail:  - total free size   : 481712
> ****************************************************************
> REGION #0 Start=0x0x2181840, End=0x0x21fe4b0, Size=511104
> ****************************************************************
>   MemAddr |   Size   | Status |    Owner   |  Pid  |
> ----------|----------|--------|------------|-------|
> 0x2181840 |       16 |   A    | 0x    dead |   9   |
> 0x2181850 |     8208 |   A    | 0x e009241 |  10(S)|
> 0x2183860 |      448 |   A    | 0x e001709 |   9   |
(more logs on memory, but the printf statements are not there)


**Loadable with CONFIG_MM_ASSERT_ON_FAIL**

> Printing from thread 429 
> Printmm_manage_alloc_fail: Allocation failed from user heap.
> mm_manage_alloc_fail:  - requested size 4000000000
> mm_manage_alloc_fail:  - caller address = 0x02238cf5
> ****************************************************************
> REGION #0 Start=0x0x2181860, End=0x0x21fe4d0, Size=511104
> ....(more logs)
> ****************************************************************
>      Details of Heap Usages (Size in Bytes)
> ****************************************************************
> < Free >
>   - Number of Free Node               : 2
>   - Largest Free Node Size            : 481584
> < Allocation >
>   - Current Size (Alive Allocation) = (1) + (2) + (3)
>      . by Dead Threads (*) (1)        : 704
>      . by Alive Threads
>         - Sum of "STACK"(**) (2)      : 24672
>         - Sum of "CURR_HEAP" (3)      : 3984
> ** NOTE **
> (*)  Alive allocation by dead threads might be used by others or might be a leakage.
> (**) Only Idle task has a separate stack region,
>   rest are all allocated on the heap region.
> < by Dead Threads >
>  Pid | Size 
> -----|------
>    9 |   704
> ing from thread 430 
> Printing from thread 431

swasthik22:
@sunghan-chang I have verified the stated scenarios, with malloc failures and heap corruption, the logs are getting displayed correctly for all the scenarios.

sunghan-chang:
@swasthik22 In flat build without the config, even it uses dbg, it prints the log at middle of some log,
```
Printing from thread 4mm_manage_alloc_fail: Allocation failed from user heap.
mm_manage_alloc_fail: - requested size 4000000000
22
``
"Printing from thread 4...22" can't print full log. Could you check that?

**Comment by swasthik22** [comment]
I have been running a background pthread for continuously printing logs, and heap corruption and/or memory allocation in different thread. Since both the threads are trying to print logs at some time there is log mix-up. 
I've tried printing logs with printf in one thread and dbg in other thread, even in this case the logs are getting mixed. So, this seems to be normal behavior in TizenRT.

**Comment by sunghan-chang** [comment]
@swasthik22 I don't think so. Could you share the case which is log mixed?

**Comment by swasthik22** [comment]
> @swasthik22 I don't think so. Could you share the case which is log mixed?

Hi, the log mixup was identified using the change - 

https://github.com/swasthik22/TizenRT/blob/c846ff7c8f903a8daa3a3220b2ed7a95926032eb/apps/examples/hello/hello_main.c#L65-L87

output - 
```

hPRINTING FROM THREAD                                                           
ePRINTING FROM THREAD                                                           
lPRINTING FROM THREAD                                                           
lPRINTING FROM THREAD                                                           
oPRINTING FROM THREAD                                                           
_PRINTING FROM THREAD                                                           
mPRINTING FROM THREAD                                                           
aPRINTING FROM THREAD                                                           
iPRINTING FROM THREAD                                                           
nPRINTING FROM THREAD                                                           
:PRINTING FROM THREAD                                                           
 PRINTING FROM THREAD                                                           
pPRINTING FROM THREAD                                                           
rPRINTING FROM THREAD                                                           
iPRINTING FROM THREAD                                                           
nPRINTING FROM THREAD                                                           
tPRINTING FROM THREAD                                                           
iPRINTING FROM THREAD                                                           
nPRINTING FROM THREAD                                                           
gPRINTING FROM THREAD
```

File: os/mm/mm_heap/mm_check_heap_corruption.c [hunk 4]
```diff
@@ -77,7 +77,7 @@ enum node_type_e {
 };
 typedef enum node_type_e node_type_t;
 
-#if defined(__KERNEL__)
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
```
thapav, line 80: Please add this at line 146 as well

os/include/debug.h
sunghan-chang: @swasthik22 @kishore-sn I still don't understand why you remove existed conditionals. Could you let me know?

sunghan-chang: @swasthik22 @kishore-sn Any reply for my query?
CONFIG_MM_ASSERT_ON_FAIL will call an assert in this case, but without the config, it just prints logs and goes next.
I think if it goes to an assert, it should be lldbg, but if not, it should be dbg. 
And why should it be dbg in loadable apps?

sunghan-chang: @swasthik22 Thank you for explanation. Then why do you add flat build config? It will be the same as loadable config.

sunghan-chang: > mfdbg always uses dbg if CONFIG_MM_ASSERT_ON_FAIL is false, which is wrong
consider the following case with CONFIG_MM_ASSERT_ON_FAIL disabled,
whenever heap corruption is followed by a board crash (ex : heap corruption by task A, followed by usage fault in task B), mfdbg uses dbg to print the heap corruption logs in assert, which results in incomplete logs.

Regarding what you said,
we use the mfdbg at two functions. 
You are right with heap corruption function, it prints corrupted heap info before assert or in assert.
But allocation failure function does not call the assert with disabled CONFIG_MM_ASSERT_ON_FAIL, then no need to check the flag.

sunghan-chang (approved):

os/mm/mm_heap/mm_check_heap_corruption.c
sunghan-chang: @thapav Oops sorry, I merged already.
@swasthik22 Could you check @thapav 's comment?

swasthik22: @sunghan-chang @thapav sorry I had forgot to push the changes for this, will be raising new PR addressing this comment.

PR #5712, opened 2022-12-08
os/board/rtl8721csm : fix error_flag for wifi_get_ap_bssid API

keerthanalex:
Update error_flag for successful wifi connection to get correct output of wifi_get_ap_bssid API


os/board/rtl8721csm/src/component/common/api/wifi/wifi_conf.c
```diff
@@ -805,6 +805,7 @@ int wifi_connect(
 	}
 
 	result = RTW_SUCCESS;
+	error_flag = RTW_NO_ERROR;
```
j11591, line 808: How about moving it to 795 line after "reason.reason_code = RTK_STATUS_SUCCESS;"

In upper layer, app can check "get info" right after link up.

If error_flag is after link up, there can be timing issue.

keerthanalex: Updated as suggested.

keerthanalex:
https://github.com/Samsung/TizenRT/pull/5719 includes this change. So closing this PR.

PR #5715, opened 2022-12-09
os/board/rtl8721csm: fix scan wpa2_3 mixed mode issue

gexuyan:
Add wpa2_3 mixed mode in scan result


gexuyan:
> 

Dear @j11591 ,
Sorry that the original committed lib has been found by Eric that it missed some specific AP's wpa2/3 mixed mode ciphers, so I added in the new lib and updated the libs. Now should be ok. Thank you.

j11591 (approved):
LGTM

PR #5718, opened 2022-12-14
os/board/rtl8721csm: Add ap mac address in get_info

j11591:


j11591:
moved error flag position in order to prevent timing issue.

PR #5723, opened 2022-12-19
[EXPERIMENTAL] os/fs/smartfs: Guaranteeing jSmartFs File Data Consistency during partial Overwrite

amogh-samsung:
- Currently, jSmartFs guarantees only file system consistency, an overwrite transaction is committed one sector at a time.
- This design does not ensure the presence of an entire file version, either from before or after the modification.
- The new logical design commits the transaction for the entire file at once, and hence ensures the presence of either the complete old or new version.


os/fs/smartfs/smartfs_smart.c
```diff
@@ -648,127 +650,224 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
+				fs->fs_llformat.availbytes, (uint8_t *)&fs->fs_rwbuffer[0]);
+
+		/* Will we be overwriting more bytes to a following sector?
+		 * If yes, pre-allocate a sector in the parallel branch and chain
+		 * it before writing the current sector.
+		 * However, before this, ssave the next chained sector in the main branch
+		 * for use later.
+		 */
+		if ((buflen - bytes) > 0 && (sf->filepos + bytes) < sf->entry.datalen) {
+			ret = FS_IOCTL(fs, BIOC_ALLOCSECT, 0xFFFF);
```
Taejun-Kwon, line 778: 1. Do not apply this to master
2. Here all sector should be released.

amogh-samsung: While we are allocating new sectors for creating a parallel branch with modified data,
We free the old chain after successfully committing the new chain in line 863

If we fail to commit the new chain because of a power outage, it is collected/released during recovery.

Taejun-Kwon: No, it's totally wrong. I mean -ENOSPC, not power cut. Here all newly changed sector should be released. then they can be consumed for next operation, you should not wait until recovery.

amogh-samsung: I understood your point now. Thankyou
I will add a fix here

os/fs/smartfs/smartfs_smart.c
```diff
@@ -648,127 +650,224 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
 		}
 
-		/* Limit the write based on current file length */
 		if (bytes > sf->entry.datalen - sf->filepos) {
 			bytes = sf->entry.datalen - sf->filepos;
 		}
 
-		/* Now perform the write. */
-		if (bytes > 0) {
+		//There should not be a case where bytes = 0
```
swasthik22, line 687: Can you explain why is this case.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -648,127 +650,224 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
 	}
 
-	header = (struct smartfs_chain_header_s *)fs->fs_rwbuffer;
+	if (buflen <= 0) {
+		/* Nothing to write */
+		ret = 0;
+		goto errout_with_semaphore;
+	}
+
+	// Permission and sanity checks completed
```
swasthik22, line 659: Use block comments for consistency.

os/fs/smartfs/smartfs_smart.c
```diff
@@ -648,127 +650,224 @@ static ssize_t smartfs_write(FAR struct file *filep, const char *buffer, size_t
...
+			/* Set this sector as the next chained sector for the parallel branch
+			 * But before that, we save the next chained sector number of the current branch
+			 * for our reference use
 			 */
-			memcpy(&sf->buffer[sf->curroffset], &buffer[byteswritten], bytes);
-			sf->bflags |= SMARTFS_BFLAG_DIRTY;
+			m_branch_sector = SMARTFS_NEXTSECTOR(header);
+			*((uint16_t *)header->nextsector) = (uint16_t)ret;
+			p_branch_sector = (uint16_t)ret;
+			save_sector = m_branch_sector; //Will be used to free the main chain later after the parallel chain is committed
```
swasthik22, line 711: Move the comment to next/previous line, use block comments.

PR #5724, opened 2022-12-21
Improved Heap corruption check 

kishore-sn:
Some changes to improve the heap corruption check and provide better debug information.

### DIFF
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -264,13 +264,13 @@ typedef size_t mmaddress_t;             /* 32 bit address space */
  */
 
 struct mm_allocnode_s {
-	mmsize_t size;					/* Size of this chunk */
 	mmsize_t preceding;				/* Size of the preceding chunk */
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	mmaddress_t alloc_call_addr;			/* malloc call address */
 	pid_t pid;					/* PID info */
 	uint16_t reserved;				/* Reserved for future use. */
 #endif
+	mmsize_t size;					/* Size of this chunk */
 
 };
 
@@ -302,8 +302,13 @@ struct mm_allocnode_s {
 /* This describes a free chunk */
 
 struct mm_freenode_s {
-	mmsize_t size;				/* Size of this chunk */
 	mmsize_t preceding;			/* Size of the preceding chunk */
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	mmaddress_t alloc_call_addr;			/* malloc call address */
+	pid_t pid;					/* PID info */
+	uint16_t reserved;				/* Reserved for future use. */
+#endif
+	mmsize_t size;				/* Size of this chunk */
 	FAR struct mm_freenode_s *flink;	/* Supports a doubly linked list */
 	FAR struct mm_freenode_s *blink;
 };
@@ -311,12 +316,7 @@ struct mm_freenode_s {
 /* What is the size of the freenode? */
 
 #define MM_PTR_SIZE sizeof(FAR struct mm_freenode_s *)
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-#define SIZEOF_MM_FREENODE \
-	(SIZEOF_MM_ALLOCNODE - SIZEOF_MM_MALLOC_DEBUG_INFO + 2 * MM_PTR_SIZE)
-#else
 #define SIZEOF_MM_FREENODE (SIZEOF_MM_ALLOCNODE + 2 * MM_PTR_SIZE)
-#endif
 
 #define CHECK_FREENODE_SIZE \
 	DEBUGASSERT(sizeof(struct mm_freenode_s) == SIZEOF_MM_FREENODE)
--- a/os/mm/Kconfig
+++ b/os/mm/Kconfig
@@ -154,3 +154,11 @@ config MM_ASSERT_ON_FAIL
 	---help---
 		When memory allocation fails, make assertion.
 		From the application binary with loadable build, it will be reloaded.
+
+config MM_DUMP_CORRPUTED_HEAP
+	bool "Dump the complete corrupted heap"
+	default n
+	---help---
+		When we find a heap corruption, we can dump the entire heap, so that
+		it can be manually analyzed to find more information about the corruption.
+
--- a/os/mm/mm_heap/mm_check_heap_corruption.c
+++ b/os/mm/mm_heap/mm_check_heap_corruption.c
@@ -65,18 +65,12 @@
  * Pre-processor Definitions
  ****************************************************************************/
 #define MM_PREV_NODE_SIZE(x)		((x)->preceding & ~MM_ALLOC_BIT)
-#define IS_ALLOCATED_NODE(x)		((x)->preceding & MM_ALLOC_BIT)
+#define IS_ALLOCATED_NODE(x)		(((x)->preceding & MM_ALLOC_BIT) != 0)
 #define IS_FREE_NODE(x)			(!IS_ALLOCATED_NODE(x))
 
 /****************************************************************************
  * Public data
  ****************************************************************************/
-enum node_type_e {
-	TYPE_CORRUPTED,
-	TYPE_OVERFLOWED,
-};
-typedef enum node_type_e node_type_t;
-
 #if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
 extern bool abort_mode;
 #endif
@@ -88,7 +82,7 @@ extern bool abort_mode;
  * Private Functions
  ****************************************************************************/
 
-static void dump_node(struct mm_allocnode_s *node, node_type_t type)
+static void dump_node(struct mm_allocnode_s *node, char *node_type)
 {
 #if defined(CONFIG_DEBUG_MM_HEAPINFO)  && (CONFIG_TASK_NAME_SIZE > 0)
 	char myname[CONFIG_TASK_NAME_SIZE + 1];
@@ -100,11 +94,8 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 		return;
 	}
 
-	if (type == TYPE_CORRUPTED) {
-		mfdbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
-	} else if (type == TYPE_OVERFLOWED) {
-		mfdbg("OVERFLOWED NODE: addr = 0x%08x size = %u type = %c\n", node, node->size, IS_ALLOCATED_NODE(node) ? 'A' : 'F');
-	}
+	mfdbg("%s: addr = 0x%08x, type = %c, size = %u, preceding size = %u\n", node_type, node, IS_ALLOCATED_NODE(node) ? 'A' : 'F', node->size, MM_PREV_NODE_SIZE(node));
+
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	pid_t pid = node->pid;
 	if (pid < 0) {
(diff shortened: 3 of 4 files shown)


os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -164,100 +174,107 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
 		 * Retake the semaphore for each region to reduce latencies
 		 */
 
-		if (!aborted
 #if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
-			&& !up_interrupt_context()
+		if (!abort_mode && !up_interrupt_context())
```
thapav, line 178: If we are using abort_mode, then may be we should add CONFIG_BUILD_FLAT to the abort_mode definition as well

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -141,19 +151,19 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
...
-	aborted = abort_mode;
-#endif
 #if CONFIG_KMM_REGIONS > 1
 	int region;
 #else
 #define region 0
 #endif
+	uint32_t start_corrupt = 0;
+	uint32_t end_corrupt = 0;
+	bool iscorrupt_f = false;
```
thapav, line 161: Please add description for these variables

os/mm/mm_heap/mm_initialize.c
```diff
@@ -173,6 +173,9 @@ int mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsiz
 	node            = (FAR struct mm_freenode_s *)(heapbase + SIZEOF_MM_ALLOCNODE);
 	node->size      = heapsize - 2 * SIZEOF_MM_ALLOCNODE;
 	node->preceding = SIZEOF_MM_ALLOCNODE;
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	heapinfo_update_node((FAR struct mm_allocnode_s *)node, 0xDEADDEAD);
```
thapav, line 177: Please add appropriate description as to where exactly is this magic number filled

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -227,33 +227,33 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
...
 			dump_node(heap->mm_heapend[region], "HEAP END NODE");
 			iscorrupt_r = true;
-			end_corrupt = heap->mm_heapend[region] + SIZEOF_MM_ALLOCNODE;
+			end_corrupt = (uint32_t)heap->mm_heapend[region] + SIZEOF_MM_ALLOCNODE;
 			mfdbg("#########################################################################################\n");
 		} else {
 			/* Backward traversal of heap */
 			for (prev = heap->mm_heapend[region], node = (struct mm_allocnode_s *)((char *)prev - MM_PREV_NODE_SIZE(prev));
-					node >= heap->mm_heapstart[region];
+					prev > heap->mm_heapstart[region];
```
thapav, line 238: I think this change belongs to the commit:
os/mm: Fix wrong assertion issue during heap corruption check

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -100,11 +101,9 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 		return;
 	}
 
-	if (type == TYPE_CORRUPTED) {
-		mfdbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
-	} else if (type == TYPE_OVERFLOWED) {
-		mfdbg("OVERFLOWED NODE: addr = 0x%08x size = %u type = %c\n", node, node->size, IS_ALLOCATED_NODE(node) ? 'A' : 'F');
-	}
+	bool is_allocated = IS_ALLOCATED_NODE(node);
+	mfdbg("%s: addr = 0x%08x type = %c size = %u preceding size = %u\n", node_type, node, is_allocated ? 'A' : 'F', node->size, MM_PREV_NODE_SIZE(node));
```
jeongchanKim, line 105: `is_allocated` is used only once from here, so we don't need to use local variable.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -60,12 +60,13 @@
...
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 #define MM_PREV_NODE_SIZE(x)		((x)->preceding & ~MM_ALLOC_BIT)
-#define IS_ALLOCATED_NODE(x)		((x)->preceding & MM_ALLOC_BIT)
+#define IS_ALLOCATED_NODE(x)		(((x)->preceding & MM_ALLOC_BIT) != 0)
 #define IS_FREE_NODE(x)			(!IS_ALLOCATED_NODE(x))
 
 /****************************************************************************
```
jeongchanKim, line 72: `node_type_t` is not used anymore, please remove it.

os/include/tinyara/mm/mm.h
```diff
@@ -302,21 +302,21 @@ struct mm_allocnode_s {
 /* This describes a free chunk */
 
 struct mm_freenode_s {
-	mmsize_t size;				/* Size of this chunk */
 	mmsize_t preceding;			/* Size of the preceding chunk */
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
```
jeongchanKim, line 306: Please check that other tools which related to heap like heapinfo works fine or not.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -164,100 +168,117 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
...
+						start_corrupt = (uint32_t)prev + prev->size - MM_LAST_CORRUPT_SIZE;
+					}
+					mfdbg("=========================================================================================\n");
+					break;
 				}
-				return -1;
-			} else if (node->size != MM_PREV_NODE_SIZE(next)) {
+			}
+
+			if (!iscorrupt_f && prev != heap->mm_heapend[region]) {
```
abhishek-samsung, line 212: Is this statement ever true? if we break the previous for loop, the iscorrupt_f is set to true so this will be false. If we complete the loop then I think prev = heap->mm_heapend[region], so I think this if statement is never executed. Could you please let me know the case where this if statement executes?

kishore-sn: Consider node N->size is corrupt. But node N->preceding is not corrupt. 
So, when prev = N-1 and node = N, then iscorrupt_f is still false.
But, when prev = N and node = N+N->size, then loop condition might fail because N->size is corrupt. 

At this point, iscorrupt_f is false and node != heapend.

abhishek-samsung: Okay, thanks

os/mm/Kconfig
```diff
@@ -154,3 +154,11 @@ config MM_ASSERT_ON_FAIL
 	---help---
 		When memory allocation fails, make assertion.
 		From the application binary with loadable build, it will be reloaded.
+
+config MM_DUMP_CORRPUTED_HEAP
+	bool "Dump the complete corrupted heap"
+	default n
+	---help---
+		When we find a heap corruption, we can dump the entire heap, so that
```
swasthik22, line 162: I think moving "so that" to next line gives better readability.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -164,100 +168,117 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
...
+			mfdbg("=========================================================================================\n");
+		} else {
+			/* Forward traversal of heap */
+			for (prev = heap->mm_heapstart[region], node = (struct mm_allocnode_s *)((char *)prev + prev->size);
+					node <= heap->mm_heapend[region];
+					prev = node, node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+				if (prev->size != MM_PREV_NODE_SIZE(node)) {
+					mfdbg("#########################################################################################\n");
+					mfdbg("ERROR: Heap corruption detected. Check below nodes for possible corruption.\n");
+					mfdbg("ERROR: Forward traversal of heap\n");
```
swasthik22, line 197: Better to keep the wordings similar to the line 183, helps in consistency.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -164,100 +168,117 @@ int mm_check_heap_corruption(struct mm_heap_s *heap)
...
+			mfdbg("=========================================================================================\n");
+			dump_node(heap->mm_heapstart[region], "HEAP START NODE");
+			iscorrupt_f = true;
+			start_corrupt = (uint32_t)(heap->mm_heapstart[region]);
+			mfdbg("=========================================================================================\n");
+		} else {
+			/* Forward traversal of heap */
+			for (prev = heap->mm_heapstart[region], node = (struct mm_allocnode_s *)((char *)prev + prev->size);
+					node <= heap->mm_heapend[region];
+					prev = node, node = (struct mm_allocnode_s *)((char *)node + node->size)) {
```
abhishek-samsung, line 193: what if node->size becomes 0 due to corruption? I think this loop runs forever then. Could you please add a check for that?

os/mm/Kconfig
kishore-sn: Thanks. But, I have adjusted the line length according to the kconfig file. It might be displayed differently in the menuconfig based on the screen size. So, let us keep the current format.

os/mm/mm_heap/mm_check_heap_corruption.c
kishore-sn: I have added these particular logs to give the required information about the issue. It will not be ok to make the log same as line 183.

os/mm/mm_heap/mm_check_heap_corruption.c
kishore-sn: Oh.. you are right!!
Fixed. Thank you.

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -117,16 +109,29 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 	}
 #if CONFIG_TASK_NAME_SIZE > 0
 	if (prctl(PR_GET_NAME_BYPID, myname, pid) == OK) {
-		mfdbg("Node owner pid = %d (%s%s), allocated by code at addr = 0x%08x\n", node->pid, myname, is_stack, node->alloc_call_addr);
+		mfdbg("%s owner pid = %d (%s%s), allocated by code at addr = 0x%08x\n", node_type, node->pid, myname, is_stack, node->alloc_call_addr);
 	} else {
-		mfdbg("Node owner pid = %d (Exited Task%s), allocated by code at addr = 0x%08x\n", node->pid, is_stack, node->alloc_call_addr);
+		mfdbg("%s owner pid = %d (Exited Task%s), allocated by code at addr = 0x%08x\n", node_type, node->pid, is_stack, node->alloc_call_addr);
```
amogh-samsung, line 114: `"Exited Task%s"` -> I understand that this would be a Task Name (%s)
Maybe a space after `"Task"` and before `"%s"` would help enhance readability?

os/mm/mm_heap/mm_check_heap_corruption.c
```diff
@@ -100,11 +95,8 @@ static void dump_node(struct mm_allocnode_s *node, node_type_t type)
 		return;
 	}
 
-	if (type == TYPE_CORRUPTED) {
-		mfdbg("CORRUPTED NODE: addr = 0x%08x size = %u preceding size = %u\n", node, node->size, MM_PREV_NODE_SIZE(node));
-	} else if (type == TYPE_OVERFLOWED) {
-		mfdbg("OVERFLOWED NODE: addr = 0x%08x size = %u type = %c\n", node, node->size, IS_ALLOCATED_NODE(node) ? 'A' : 'F');
-	}
+	mfdbg("%s: addr = 0x%08x type = %c size = %u preceding size = %u\n", node_type, node, IS_ALLOCATED_NODE(node) ? 'A' : 'F', node->size, MM_PREV_NODE_SIZE(node));
```
amogh-samsung, line 98: Minor - would you consider adding a comma ',' after every printed value?

os/mm/mm_heap/mm_check_heap_corruption.c
kishore-sn: Actually it prints either as `Exited task's stack` or `Exited task`.
So, it is fine.

os/mm/mm_heap/mm_check_heap_corruption.c
kishore-sn: Done. Thanks.

os/include/tinyara/mm/mm.h
sunghan-chang, line 306: 1. Could you let me know why this is necessary for free node?
2. When we free some node, the mm merges neighbor free nodes. Then this could give unwanted information.

kishore-sn: If a user calls mm_free, and then continues to use the same node, he might cause heap corruption. In such case, this might provide us the information about the process which had allocated the node. This scenario might not be true in most cases, but in some rare case, it will help us for debugging.

sunghan-chang (approved):

os/include/tinyara/mm/mm.h
```diff
@@ -264,13 +264,13 @@ typedef size_t mmaddress_t;             /* 32 bit address space */
  */
 
 struct mm_allocnode_s {
-	mmsize_t size;					/* Size of this chunk */
 	mmsize_t preceding;				/* Size of the preceding chunk */
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
 	mmaddress_t alloc_call_addr;			/* malloc call address */
 	pid_t pid;					/* PID info */
 	uint16_t reserved;				/* Reserved for future use. */
```
sunghan-chang, line 271: How about using this to get corrupted or not like adding 0xDEAD?

PR #5730, opened 2022-12-27
os/board/rtl8721csm/src: Add ble ADV Tx Power API

jiangxiaoyunjoy:
Add ble ADV Tx Power API:
0x6(-10 dBm)
0x1A(0 dBm)
0x26(-10 dBm)


ydlee-sec:
squashed to PR5731

PR #5737, opened 2023-01-11
dutils.sh: Add option to store output to a file

thapav:
This patch adds an option to store the output of the commands to a file with the name of command and timestamp as below:
    >> Output is stored in dutils_output_addr2line_2023.01.12-18.21.24.txt
    >> Output is stored in dutils_output_trap_2023.01.12-18.20.20.txt
    >> Output is stored in dutils_output_nm_2023.01.12-18.22.39.txt

### DIFF
--- a/os/dutils.sh
+++ b/os/dutils.sh
@@ -24,7 +24,7 @@ TOPDIR="${OSDIR}/.."
 BINDIR="${TOPDIR}/build/output/bin"
 TRAPDIR="${TOPDIR}/tools/trap"
 DOCKER_VERSION="1.5.5"
-
+OUTPUTFILE="dutils_output_"
 
 # Checking docker is installed
 nodocker() {
@@ -207,6 +207,7 @@ function MENU()
 	unset SELECTED_START
 
 	while [ 1 ]; do
+		OUTPUTFILE="dutils_output_"
 		if [ ! -z "$1" ];then
 			SELECTED_START=$1
 		else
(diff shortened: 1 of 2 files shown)


os/dutils.sh
```diff
@@ -243,12 +244,21 @@ function MENU()
...
+
 	while test $# -gt 0; do
 		ARGS+=" "$3
 		shift
 	done
+
+
 	echo $ARGS
-	docker run --rm -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged tizenrt/tizenrt:${DOCKER_VERSION} $ARGS
+	docker run --rm -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged tizenrt/tizenrt:${DOCKER_VERSION} $ARGS | tee $OUTPUTFILE
```
abhishek-samsung, line 259: does the command ```(ex : arm-none-eabi-addr2line -e tinyara 0x0...)``` also get added to the log file? I think it would be helpful as the file name doesn't contain the binary on which the command is run.

os/dutils.sh
```diff
@@ -243,12 +244,21 @@ function MENU()
 function TOOLCHAIN()
 {
 	ARGS=arm-none-eabi-$2
+
+	OUTPUTFILE+="$2.txt"
+	TIMESTAMP=$(date "+%Y.%m.%d-%H.%M.%S")
+	OUTPUTFILE=$OUTPUTFILE.$TIMESTAMP
```
sunghan-chang, line 250: How about formatting like this?
```
COMMAND_$TIMESTAMP.txt
```
The COMMAND could be `nm`, `addr2line` or manual input. (similar as Abhishek suggested)

os/dutils.sh
```diff
@@ -272,7 +286,9 @@ function TRAP_RUN()
 
 	echo "Executing: $TRAPCMD"
 	# execute TRAP script
-	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -it -w /root/tizenrt/tools/trap --privileged tizenrt/tizenrt:1.5.6 python3.7 $TRAPCMD
+	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -it -w /root/tizenrt/tools/trap --privileged tizenrt/tizenrt:1.5.6 python3.7 $TRAPCMD | tee $OUTPUTFILE
+
+	echo ">> Output is stored in $OUTPUTFILE"
```
sunghan-chang, line 291: The output should not be a changed in git. Could you add prefix in the .gitignore?
To do this, the output file should have a prefix like `dutils_output_xxxxxx.txt`.
i.e. dutils_output_addr2line_<timestamp>.txt

thapav: Updated

sunghan-chang (approved):

PR #5738, opened 2023-01-11
TRAP: Add functionality to dump a file from userfs partition on crash

amogh-samsung:
- This change allows dumping of a particular file by name/path from a crashed TizenRT device

USAGE: - After TizenRT device crashes

> 
> ************************************************************************************************
>                 Disconnect this serial terminal and run Dump Tool
> ************************************************************************************************
> 
> 

> /TizenRT/tools/trap$ sudo ./trap.sh 
> Please enter serial port adapter:
> For example: /dev/ttyUSB1 or /dev/ttyACM0
> Enter:
> /dev/ttyUSB1
> Target device locked and ready to DUMP!!!
> Choose from the following options:-
> 1. RAM Dump
> 2. Userfs Dump
> 3. Both RAM and Userfs dumps
> 4. External Userfs Partition Dump
> 5. Dump specific file from target device
> r. Reboot TARGET Device
> x. Exit TRAP tool
> You may chose mutiple dump options together by entering the option numbers consectuively
> For Example - '14' to dump both RAM and External USERFS partition contents OR '35' to print RAM, USERFS and FILE contents
> 5
> This utility will allow you to enter a file name that you want to be dumped from the device
> It is the responsibility of the user to ensure that the correct file name/path is provided
> do_handshake: Target Handshake successful
> Enter the absolute path of the file that you want to be dumped from the Filesystem: -/mnt/f1.txt
> Attempting to Dump File /mnt/f1.txt from Target Device
> 
> Receiving file /mnt/f1.txt of size 5522 from Target Device
> 
> [>]
> 
> Dumped file /mnt/f1.txt from filesystem successfully
> Choose from the following options:-
> 1. RAM Dump
> 2. Userfs Dump
> 3. Both RAM and Userfs dumps
> 4. External Userfs Partition Dump
> 5. Dump specific file from target device
> r. Reboot TARGET Device
> x. Exit TRAP tool
> You may chose mutiple dump options together by entering the option numbers consectuively
> For Example - '14' to dump both RAM and External USERFS partition contents OR '35' to print RAM, USERFS and FILE contents
> r
> NOTE: - CONFIG_BOARD_ASSERT_AUTORESET needs to be enabled to reboot TARGET Device after a crash
> Handsake for rebooting device may fail since device reboots before sending acknowledgement
> Wrong Target Handshake, ack = H
> Target Handshake Failed
> Dump tool exits after successful operation

### DIFF
--- a/os/board/common/crashdump.c
+++ b/os/board/common/crashdump.c
@@ -32,6 +32,7 @@
 #include "up_internal.h"
 #include "up_watchdog.h"
 #include <sys/boardctl.h>
+#include <sys/stat.h>
 #include <limits.h>
 
 #include <tinyara/mm/heap_regioninfo.h>
@@ -45,6 +46,7 @@
 #define FSDUMP_HANDSHAKE_STRING	"TIZENRTFSDUMP"
 #define EXTFSDUMP_HANDSHAKE_STRING	"TIZENRTEXTFSD"
 #define TARGET_REBOOT_STRING		"TIZENRTREBOOT"
+#define USERFS_FDUMP_HANDSHAKE_STRING	"TIZENRTFILDUM"
 #define HANDSHAKE_STR_LEN_MAX		(13)
 #define USERFS_PART_NAME		"user"
 #define USERFS_PART_NAME_LEN		(4)
@@ -205,6 +207,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
 	*index = -1;
 #endif
 
+	return 0;
+}
+
+static int send_file(void)
+{
+	int i;
+	int fd;
+	int ret;
+	struct stat st;
+	off_t size;
+	char buf;
+	char filename[CONFIG_SMARTFS_MAXNAMLEN + 1] = {'\0'};
+	char char_filesize;
+
+	/* Send acknowledgement for USERFS FILE DUMP Handshake */
+	up_lowputc('A');
+
+	/* Receive the file name to be sent to the host */
+	for (i = 0; i < CONFIG_SMARTFS_MAXNAMLEN + 1; i++) {
+		filename[i] = up_getc();
+	}
+	if (filename[i - 1] != '\0') {
+		return -ENAMETOOLONG;
+	}
+
+	ret = stat(filename, &st);
+	if (ret != OK) {
+		/* We are not able to stat the file for some reason */
+		return ERROR;
+	}
+	size = st.st_size;
+
+	/* Send the size of the file we are dumping to the host so that it knows how much data is coming in
+	 * It is a 64 bit data type so has to be transmitted accordingly
+	 * This also acts as a confirmation of the presence of the file in itself.
+	 */
+	for (i = 0; i < 8; i++) {
+		char_filesize = (size >> ((8 - 1 - i) * 8));
+		up_lowputc(char_filesize);
+	}
+
+	fd = open(filename, O_RDOK);
+	if (fd < 0) {
+		//Unable to open file for transmitting data
+		return ERROR;
+	}
+
+	while (size) {
+		ret = read(fd, &buf, 1);
+		if (ret != 1) {
+			//Unable to read from file
+			close(fd);
+			return ERROR;
+		}
+		up_lowputc(buf);
+		size--;
+	}
+
+	close(fd);
 	return 0;
 }
 
@@ -450,6 +511,12 @@ static int dump_via_uart(void)
 			lldbg("Unable to reset device\n");
 #endif
 			break;
+		/* Check handshake string for USERFS FILE DUMP signal */
+		} else if (strncmp(host_buf, USERFS_FDUMP_HANDSHAKE_STRING, strlen(USERFS_FDUMP_HANDSHAKE_STRING)) == 0) {
+			i = send_file();
+			if (i != 0) {
+				break;
+			}
 		} else {
 			/* Send NAK */
 			up_lowputc('N');
--- a/tools/trap/trap.c
+++ b/tools/trap/trap.c
@@ -367,10 +367,11 @@ int main(int argc, char *argv[])
 	char *dev_file;
 	char tty_path[TTYPATH_LEN] = {0, };
 	char tty_type[TTYTYPE_LEN] = {0, };
-	int choice;
+	char choice[4] = {'\0'};
 	char c;
 	char handshake_string[HANDSHAKE_STR_LEN_MAX];
 	int DUMP_FLAGS;
+	int choice_index = 0;
 
 	ret = 1;
 
@@ -384,12 +385,10 @@ int main(int argc, char *argv[])
 	dev_file = argv[1];
 
 	/* Get the tty type  */
-	if (!strcmp(dev_file, "/dev/ttyUSB1")) {
-		strncpy(tty_type, "ttyUSB1", TTYTYPE_LEN);
-	} else if (!strcmp(dev_file, "/dev/ttyUSB0")) {
-		strncpy(tty_type, "ttyUSB0", TTYTYPE_LEN);
-	} else if (!strcmp(dev_file, "/dev/ttyACM0")) {
-		strncpy(tty_type, "ttyACM0", TTYTYPE_LEN);
+	if (!strncmp(dev_file, "/dev/ttyUSB", 11)) {
+		strncpy(tty_type, &dev_file[5], TTYTYPE_LEN);
+	} else if (!strncmp(dev_file, "/dev/ttyACM", 11)) {
+		strncpy(tty_type, &dev_file[5], TTYTYPE_LEN);
 	} else {
 		printf("Undefined tty %s\n", dev_file);
 		return -1;
@@ -421,38 +420,43 @@ int main(int argc, char *argv[])
 	printf("Target device locked and ready to DUMP!!!\n");
 
 	while (1) {
-		printf("Choose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. External Userfs Partition Dump\n5. Exit TRAP Tool\n6. Reboot TARGET Device\n7. Dump specific file from Target Device\n");
+		printf("\nChoose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. External Userfs Partition Dump\n5. Dump specific file from target device\nr. Reboot TARGET Device\nx. Exit TRAP tool\n");
+		printf("\nYou may chose mutiple dump options together by entering the option numbers consectuively\nFor Example - \'14\' to dump both RAM and External USERFS partition contents OR \'35\' to print RAM, USERFS and FILE contents\n");
 		fflush(stdout);
-		scanf("%d", &choice);
+		scanf("%s", choice);
 		getchar();
 
 		DUMP_FLAGS = CLEAR_DUMP_FLAGS;
 
-		switch (choice) {
-		case 1:
-			DUMP_FLAGS |= RAMDUMP_FLAG;
-			break;
-		case 2:
-			DUMP_FLAGS |= USERFSDUMP_FLAG;
-			break;
-		case 3:
-			DUMP_FLAGS |= RAMDUMP_FLAG;
-			DUMP_FLAGS |= USERFSDUMP_FLAG;
-			break;
-		case 4:
-			DUMP_FLAGS |= EXTUSERFS_DUMP_FLAG;
-			break;
-		case 5:
-			break;
-		case 6:
-			DUMP_FLAGS |= REBOOT_DEVICE_FLAG;
-			break;
-		case 7:
-			DUMP_FLAGS |= FILE_DUMP_FLAG;
-			break;
-		default:
-			printf("Invalid Input\n");
-			continue;
+		while (choice_index < 4 && choice[choice_index] != '\0') {
+			switch (choice[choice_index]) {
+			case '1':
+				DUMP_FLAGS |= RAMDUMP_FLAG;
+				break;
+			case '2':
+				DUMP_FLAGS |= USERFSDUMP_FLAG;
+				break;
+			case '3':
+				DUMP_FLAGS |= RAMDUMP_FLAG;
+				DUMP_FLAGS |= USERFSDUMP_FLAG;
+				break;
+			case '4':
+				DUMP_FLAGS |= EXTUSERFS_DUMP_FLAG;
+				break;
+			case '5':
+				DUMP_FLAGS |= FILE_DUMP_FLAG;
+				break;
+			case 'r':
+				DUMP_FLAGS |= REBOOT_DEVICE_FLAG;
+				break;
+			case 'x':
+				break;
+			default:
+				printf("Invalid Input\n");
+				continue;
+			}
+
+			choice_index++;
 		}
 
 		if (DUMP_FLAGS & REBOOT_DEVICE_FLAG) {
--- a/tools/trap/trap.c
+++ b/tools/trap/trap.c
@@ -24,6 +24,14 @@
  ****************************************************************************/
 
 #include "trap.h"
+#include "../../os/include/tinyara/config.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+#ifndef CONFIG_SMARTFS_MAXNAMLEN
+#define USERFS_MAXNAMLEN 32
+#endif
 
 /****************************************************************************
  * Private Variables
@@ -288,6 +296,69 @@ static int extfsdump_recv(int dev_fd)
         return 0;
 }
 
+int dump_file_from_fs(int dev_fd, char *filename, int maxnamelen)
+{
+	int ret;
+	int fd;
+	int i, last_subdir;
+	char buf;
+	FILE *fp;
+	u_int64_t file_size;
+	char char_filesize[8];
+	char local_filename[maxnamelen + 9];
+
+	printf("Attempting to Dump File %s from Target Device\n", filename);
+
+	ret = write(dev_fd, filename, maxnamelen);
+	if (ret != maxnamelen) {
+		printf("Unable to send Dump File name to Target Device, ret = %d\n", ret);
+		return -1;
+	}
+
+	file_size = 0;
+	for (i = 0; i < 8; i++) {
+		ret = read(dev_fd, &char_filesize[i], sizeof(char_filesize[i]));
+		if (ret != sizeof(char_filesize[i])) {
+			printf("Unable to get file size from Target Device, ret = %d\n", ret);
+			return -1;
+		}
+	}
+
+	for (i = 7; i >= 0; i--) {
+		file_size += ((u_int64_t)char_filesize[8 - i - 1] << (i * 8)) & ((u_int64_t)0xFF << (i * 8));
+	}
+
+	printf("\nReceiving file %s of size %lu from Target Device\n\n", filename, file_size);
+
+	i = 0;
+	last_subdir = -1;
+	while (filename[i] != '\0') {
+		/* We dont want the file to be saved in the root folder, we want it in the current folder */
+		if (filename[i] == '/') {
+			last_subdir = i;
+		}
+		i++;
+	}
+
+	filename = filename + last_subdir + 1;
+	snprintf(local_filename, maxnamelen + 9, "%s%s", "DumpFile_", filename);
+
+	fp = fopen(local_filename, "w");
+	if (fp == NULL) {
+		printf("Unable to open file %s to save logs chain dump\n", filename);
+		return -1;
+	}
+
+	ret = get_dump(dev_fd, fp, file_size);
+	fclose(fp);
+	if (ret != 0) {
+		printf("Unable to receive logs chain dump file from target device, ret = %d\n", ret);
+		return -1;
+	}
+
+	return 0;
+}
+
 int main(int argc, char *argv[])
 {
 	int ret;
@@ -350,7 +421,7 @@ int main(int argc, char *argv[])
 	printf("Target device locked and ready to DUMP!!!\n");
 
 	while (1) {
-		printf("Choose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. External Userfs Partition Dump\n5. Exit TRAP Tool\n6. Reboot TARGET Device\n");
+		printf("Choose from the following options:-\n1. RAM Dump\n2. Userfs Dump\n3. Both RAM and Userfs dumps\n4. External Userfs Partition Dump\n5. Exit TRAP Tool\n6. Reboot TARGET Device\n7. Dump specific file from Target Device\n");
 		fflush(stdout);
 		scanf("%d", &choice);
 		getchar();
@@ -375,6 +446,9 @@ int main(int argc, char *argv[])
 			break;
 		case 6:
 			DUMP_FLAGS |= REBOOT_DEVICE_FLAG;
+			break;
+		case 7:
+			DUMP_FLAGS |= FILE_DUMP_FLAG;
 			break;
 		default:
 			printf("Invalid Input\n");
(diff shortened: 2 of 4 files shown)


tools/trap/trap.c
```diff
@@ -288,6 +291,67 @@ static int extfsdump_recv(int dev_fd)
...
 }
 
+int dump_file_from_fs(int dev_fd, char *filename)
+{
+	int ret;
+	int fd;
+	int i, last_subdir;
+	char buf;
+	FILE *fp;
+	u_int64_t file_size;
```
thapav, line 301: uint64_t*

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+
+	ret = stat(filename, &st);
+	if (ret != OK) {
+		/* We are not able to stat the file for some reason */
+		return -1;
+	}
+	size = st.st_size;
+
+	/* This Ack signal confirms the presence of the requested file */
+	//up_lowputc('A');
```
thapav, line 240: Is this needed ? If not please remove

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+	 * This also acts as a confirmation of the presence of the file in itself.
+	 */
+	for (i = 0; i < 8; i++) {
+		char_filesize = (size >> ((8 - 1 - i) * 8));
+		up_lowputc(char_filesize);
+	}
+
+	fd = open(filename, O_RDOK);
+	if (fd < 0) {
+		//Unable to open file for transmitting data
```
thapav, line 253: Please maintain consistent comment style
/* */

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+		return -EIO;
+	}
+
+	while (size) {
+		ret = read(fd, &buf, 1);
+		if (ret != 1) {
+			//Unable to read from file
```
thapav, line 260: Same as above

os/board/common/crashdump.c
```diff
@@ -450,6 +511,12 @@ static int dump_via_uart(void)
 			lldbg("Unable to reset device\n");
 #endif
 			break;
+		/* Check handshake string for USERFS FILE DUMP signal */
+		} else if (strncmp(host_buf, USERFS_FDUMP_HANDSHAKE_STRING, strlen(USERFS_FDUMP_HANDSHAKE_STRING)) == 0) {
```
thapav, line 515: Kinldy use strnlen()

sunghan-chang:
@amogh-samsung Could you add some example how to use in the commit description?

tools/trap/trap.c
```diff
@@ -24,6 +24,9 @@
  ****************************************************************************/
 
 #include "trap.h"
+#include "../../os/include/tinyara/config.h"
```
sunghan-chang, line 27: Could you let me know why includes this?

amogh-samsung: It is used for `CONFIG_SMARTFS_MAXNAMLEN`
The maximum possible size of the file path string

sunghan-chang: 1. But there is a definition for that at line 29.
[CONFIG_SMARTFS_MAXNAMLEN](https://github.com/Samsung/TizenRT/pull/5738/files#diff-d50c5afb6cd42629b54e5c4ead10ed2e4057df258a2e1f16d7a8f09a326679e9R29)
2. If user does not have built source, then this will be failed. Do we need to build TizenRT to use TRAP?

amogh-samsung: Sorry I need to remove the one in line 29
And I agree with point 2 given by you

Would you suggest I continue with a manually defined dummy config for the purpose of this tool?

sunghan-chang: Thapa also reads some config value on the TRAP. Could you check it with her about how to get a config value? Let's use the same way.

amogh-samsung: I checked tools/trap/ramdumpParser.py
This also uses the TizenRT .config file (Line 1075)

How about adding a comment in HowToUseTrap.md file about the need for building source code for making the config file available?

sunghan-chang: I think 
1. trying to read .config file from default location
2. If no .config there, feedback an error mesage about no .config and wait an input of configuration file 

Many developer uses the tool without source code so that I think getting necessary file separately is better.

os/board/common/crashdump.c
amogh-samsung: All handshake checks are currently using `strlen`, so sticking with the same.

tools/trap/trap.c
amogh-samsung: Pending, will take care of this

tools/trap/trap.c
amogh-samsung: Done

os/board/common/crashdump.c
amogh-samsung: Done

os/board/common/crashdump.c
amogh-samsung: Done

os/board/common/crashdump.c
amogh-samsung: Done

sunghan-chang:
> RAM Dump
Userfs Dump
Both RAM and Userfs dumps
External Userfs Partition Dump
Exit TRAP Tool
Reboot TARGET Device
Dump specific file from Target Device

The option after exit looks strange. How about adding that before exit and reboot?
To do this, we had better modify the numbers for exit and reboot not to get affected from new adding like
```
r. Reboot TARGET device
x. Exit TRAP
```

amogh-samsung:
> > 1. RAM Dump
> > 2. Userfs Dump
> > 3. Both RAM and Userfs dumps
> > 4. External Userfs Partition Dump
> > 5. Exit TRAP Tool
> > 6. Reboot TARGET Device
> > 7. Dump specific file from Target Device
> 
> The option after exit looks strange. How about adding that before exit and reboot? To do this, we had better modify the numbers for exit and reboot not to get affected from new adding like
> 
> ```
> r. Reboot TARGET device
> x. Exit TRAP
> ```

I agree, this makes sense
But perhaps as a separate commit maybe in the same or different PR?
Would that be alright?

sunghan-chang:
> I agree, this makes sense
But perhaps as a separate commit maybe in the same or different PR?
Would that be alright?

Both. Anything is ok for me.

tools/trap/trap.c
amogh-samsung: I pushed a slightly different version.
My estimate is that if the config.h file is not present because the user did not build the code, it is unlikely that they would have the config.h or config file present separately.
Even if they do, I feel it is better to ask the user for the config value directly instead of the file path.
The value is this config is always (so far in all defconfigs) = 32

Also, dynamic inclusion of the provided file path as a header file in C would not be easy.

tools/trap/trap.c
```diff
@@ -288,6 +289,67 @@ static int extfsdump_recv(int dev_fd)
...
+	}
+
+	file_size = (((u_int64_t)char_filesize[0] << 56) & 0xFF00000000000000U) | (((u_int64_t)char_filesize[1] << 48) & 0x00FF000000000000U) | (((u_int64_t)char_filesize[2] << 40) & 0x0000FF0000000000U) | (((u_int64_t)char_filesize[3] << 32) & 0x000000FF00000000U) | ((char_filesize[4] << 24) & 0x00000000FF000000U) | ((char_filesize[5] << 16) & 0x0000000000FF0000U) | ((char_filesize[6] <<  8) & 0x000000000000FF00U) | (char_filesize[7] & 0x00000000000000FFU);
+
+	printf("\nReceiving file %s of size %lu from Target Device\n\n", filename, file_size);
+
+	i = 0;
+	last_subdir = -1;
+	while (filename[i] != '\0') {
+		/* We dont want the file to saved in the root folder, we want it in the current folder */
```
abhishek-samsung, line 327:
```suggestion
		/* We dont want the file to be saved in the root folder, we want it in the current folder */
```

tools/trap/trap.c
```diff
@@ -463,6 +528,40 @@ int main(int argc, char *argv[])
...
+			printf("This utility will allow you to enter a file name that you want to be dumped from the device\n");
+			printf("It is the responsibility of the user to ensure that the correct file name/path is provided\n");
+
+			strncpy(handshake_string, USERFS_FDUMP_HANDSHAKE_STRING, HANDSHAKE_STR_LEN_MAX);
+			if (do_handshake(dev_fd, handshake_string) != 0) {
+				printf("Target handshake failed\n");
+				goto handshake_err;
+			}
+
+			printf("Enter the path of the file that you want to be dumped from the Filesystem: -");
```
abhishek-samsung, line 551: Is it **absolute path** or path relative to **/mnt**  ? I think its better to mention here to be clear.

amogh-samsung: Let's mention absolute

tools/trap/trap.c
amogh-samsung: Let me do this, thanks

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+	 */
+	for (i = 0; i < 8; i++) {
+		char_filesize = (size >> ((8 - 1 - i) * 8));
+		up_lowputc(char_filesize);
+	}
+
+	fd = open(filename, O_RDOK);
+	if (fd < 0) {
+		//Unable to open file for transmitting data
+		return -EIO;
```
Taejun-Kwon, line 254: Line 235 it returns -1, but here it returns -EIO.
Actually -1 is EPERM, so you should make clear this.
1. ERROR or OK
2. errno or OK

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+	}
+
+	while (size) {
```
Taejun-Kwon, line 257: You already know less I/O always improve performance.
```
char *buf = malloc(size...);
ret = read(fd, buf, size);
....
while (*size) {
   up_lowputc(*size++);
}
```

tools/trap/trap.c
```diff
@@ -288,6 +289,67 @@ static int extfsdump_recv(int dev_fd)
...
+	file_size = 0;
+	for (int i = 0; i < 8; i++) {
+		ret = read(dev_fd, &char_filesize[i], sizeof(char_filesize[i]));
+		if (ret != sizeof(char_filesize[i])) {
+			printf("Unable to get file size from Target Device\n");
+			return -1;
+		}
+	}
+
+	file_size = (((u_int64_t)char_filesize[0] << 56) & 0xFF00000000000000U) | (((u_int64_t)char_filesize[1] << 48) & 0x00FF000000000000U) | (((u_int64_t)char_filesize[2] << 40) & 0x0000FF0000000000U) | (((u_int64_t)char_filesize[3] << 32) & 0x000000FF00000000U) | ((char_filesize[4] << 24) & 0x00000000FF000000U) | ((char_filesize[5] << 16) & 0x0000000000FF0000U) | ((char_filesize[6] <<  8) & 0x000000000000FF00U) | (char_filesize[7] & 0x00000000000000FFU);
```
Taejun-Kwon, line 320: Instead of this tricky code, you can use uint64_t buf when you write & read.
write file size : write(fd, &buf, sizeof(uint64_t));
get file size : read(fd, &buf, sizeof(uint64_t));
anyway, is file_size written in header of file?

tools/trap/trap.c
```diff
@@ -288,6 +289,67 @@ static int extfsdump_recv(int dev_fd)
...
+	filename = filename + last_subdir + 1;
+	snprintf(local_filename, CONFIG_SMARTFS_MAXNAMLEN + 9, "%s%s", "DumpFile_", filename);
+
+	fp = fopen(local_filename, "w");
+	if (fp == NULL) {
+		printf("Unable to open file %s to save logs chain dump\n", filename);
+		return -1;
+	}
+
+	ret = get_dump(dev_fd, fp, file_size);
```
Taejun-Kwon, line 343:
```
get_dump(FILE *fp)
{
    fopen();
    ...do_something
    fclose();
}
```
get_dump seems that get data from somewhere, so it should handle fp alone

tools/trap/trap.c
```diff
@@ -463,6 +528,40 @@ int main(int argc, char *argv[])
 			printf("\nExternal Userfs partition dump received successfully\n");
 			fflush(stdout);
 		}
+
+		if (DUMP_FLAGS & FILE_DUMP_FLAG) {
+#ifdef CONFIG_SMARTFS_MAXNAMLEN
```
Taejun-Kwon, line 533: If I were you,

at the top of code,
```
#ifndef CONFIG_SMARTFS_MAXNAMLEN
#define CONFIG_SMARTFS_MAXNAMELEN 32
#endif
```
and if user input greater than this config, print out error message.

os/board/common/crashdump.c
amogh-samsung: I see crashdump.c does not use malloc at any place.
Also, are we really concerned about I/O time/performance when device is crashed? Our primary aim should be that we are able to extract the file/data correctly from the crashed device for analysis.

tools/trap/trap.c
```diff
@@ -463,6 +528,40 @@ int main(int argc, char *argv[])
...
+			int smartfs_maxnamelength = CONFIG_SMARTFS_MAXNAMLEN;
+#else
+			int smartfs_maxnamelength;
+			printf("No Config found for maximum possible file name length, please enter manually (usually = 32)\n");
```
swasthik22, line 537:
```suggestion
			printf("No Config found for maximum possible length of file name, please enter manually (usually = 32)\n");
```

tools/trap/trap.c
```diff
@@ -288,6 +289,67 @@ static int extfsdump_recv(int dev_fd)
...
+		printf("Unable to send Dump File name to Target Device\n");
+		return -1;
+	}
+
+	file_size = 0;
+	for (int i = 0; i < 8; i++) {
+		ret = read(dev_fd, &char_filesize[i], sizeof(char_filesize[i]));
+		if (ret != sizeof(char_filesize[i])) {
+			printf("Unable to get file size from Target Device\n");
+			return -1;
```
swasthik22, line 316: please check for consistency of return statements in all places. Some places there is flag based return, and in some there is numeric returns.

amogh-samsung: This is already under discussion with regards to line 254 in one of the above conversations.
Please check that

tools/trap/trap.c
amogh-samsung: Will do

tools/trap/trap.c
Taejun-Kwon: It is better that handle open & close in one function, it is always better option. you may know it enough, why don't you modify it?

tools/trap/trap.c
Taejun-Kwon: Because you write file name and readback, then calculate file size.
That's why asking you is there information regarding file size at the start of file.
I can't understand why this code as added because we can use stat to get size of file easily.

tools/trap/trap.c
Taejun-Kwon: Yes, maybe user who using dump may know well about our system & platform I agree with you. So,
1. Then it should ask size of name length to user all the time if you consider such a user as you mentioned above.
2. We can mount another file system, so here you should not refer to config in specific file system, smartfs.
3. But In my opinion, it should not check length of user input here, and fs will return ERROR & ENAMETOOLONG or EEXIST as a errno, then print out information regarding error when it fail to created file. Because they know well about our system as we expected.

tools/trap/trap.c
amogh-samsung: Yes, it can be done.
I will check and create a different commit for that if confirmed

os/board/common/crashdump.c
amogh-samsung: @kishore-sn what is your opinion on this please?

tools/trap/trap.c
amogh-samsung: We need file_size so that we know how many characters of file data we are supposed to receive from the crashed device.
We get this file size from the device itself after we pass the name of the file that we want.
The device side performs a stat for us to get the file size.
However, the device sends back one character at a time using up_lowputc, but we need 8 characters (64 bytes) in total for file_size.
Hence, we receive one character at a time and combine them using this logic into a 64 bit file_size.

We cannot perform a stat to get the userfs file size here (in trap.c) because this file is run on the Linux host machine

tools/trap/trap.c
amogh-samsung: 1. I don't think we should ask the user for the size of name length if we assume that the user is not familiar with our system. They might not know about the concept of a MAX NAME LENGTH in this case.
2. I agree with this point, we should use a common config, something maybe like CONFIG_FS_MAXNAMLEN. The maximum file path length should be a common restriction on any file system that passes through vfs.
As of now, we only have the one file system i.e. jSmartFs
3. I get your point. We can use an arbitrary limit on file name length here, and when the file name is passed to the device for dumping, the device checks the length of the file name and returns an error if it is too long (ENAMETOOLONG)
Let me check this third possibility

os/board/common/crashdump.c
```diff
@@ -208,6 +210,68 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
 
+static int send_file(void)
+{
+	int i;
+	int fd;
+	int ret;
+	struct stat st;
+	off_t size;
+	char buf;
+	char filename[CONFIG_SMARTFS_MAXNAMLEN + 1];
```
kishore-sn, line 221: I think its better to initialize the filename with null characters.

os/board/common/crashdump.c
```diff
@@ -208,6 +210,68 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+
+	ret = stat(filename, &st);
+	if (ret != OK) {
+		/* We are not able to stat the file for some reason */
+		return ERROR;
+	}
+	size = st.st_size;
+
+	/* This Ack signal confirms the presence of the requested file */
+	//up_lowputc('A');
```
kishore-sn, line 243: If not required, lets delete these lines.

os/board/common/crashdump.c
kishore-sn: In the crashed state, we dont know what is the state of kernel. So, all kernel operations must be avoided as much as possible. So, let us avoid malloc. However, if reading one character at a time from the file causes too much overhead, then we could maintain a static buffer of sectorsize or blocksize. It is left to you to experiment with this and decide if it is required.

tools/trap/trap.c
```diff
@@ -288,6 +294,67 @@ static int extfsdump_recv(int dev_fd)
...
+	file_size = 0;
+	for (int i = 0; i < 8; i++) {
+		ret = read(dev_fd, &char_filesize[i], sizeof(char_filesize[i]));
+		if (ret != sizeof(char_filesize[i])) {
+			printf("Unable to get file size from Target Device, ret = %d\n", ret);
+			return -1;
+		}
+	}
+
+	file_size = (((u_int64_t)char_filesize[0] << 56) & 0xFF00000000000000U) | (((u_int64_t)char_filesize[1] << 48) & 0x00FF000000000000U) | (((u_int64_t)char_filesize[2] << 40) & 0x0000FF0000000000U) | (((u_int64_t)char_filesize[3] << 32) & 0x000000FF00000000U) | ((char_filesize[4] << 24) & 0x00000000FF000000U) | ((char_filesize[5] << 16) & 0x0000000000FF0000U) | ((char_filesize[6] <<  8) & 0x000000000000FF00U) | (char_filesize[7] & 0x00000000000000FFU);
```
kishore-sn, line 325: This must work. Instead of having such long statement.
```
for (int i = 0; i < 8; i++)
    filesize |= (u_int64_t)char_filesize[8 - i - 1] << (i * 8);
```

os/board/common/crashdump.c
amogh-samsung: Understood, will do

tools/trap/trap.c
amogh-samsung: Yes will consider this, thankyou

tools/trap/trap.c
```diff
@@ -288,6 +296,70 @@ static int extfsdump_recv(int dev_fd)
...
+	for (i = 0; i < 8; i++) {
+		ret = read(dev_fd, &char_filesize[i], sizeof(char_filesize[i]));
+		if (ret != sizeof(char_filesize[i])) {
+			printf("Unable to get file size from Target Device, ret = %d\n", ret);
+			return -1;
+		}
+	}
+
+//	file_size = (((u_int64_t)char_filesize[0] << 56) & 0xFF00000000000000U) | (((u_int64_t)char_filesize[1] << 48) & 0x00FF000000000000U) | (((u_int64_t)char_filesize[2] << 40) & 0x0000FF0000000000U) | (((u_int64_t)char_filesize[3] << 32) & 0x000000FF00000000U) | ((char_filesize[4] << 24) & 0x00000000FF000000U) | ((char_filesize[5] << 16) & 0x0000000000FF0000U) | ((char_filesize[6] <<  8) & 0x000000000000FF00U) | (char_filesize[7] & 0x00000000000000FFU);
+	for (i = 7; i >= 0; i--) {
```
amogh-samsung, line 328: Pending resolution

tools/trap/trap.c
```diff
@@ -367,7 +367,7 @@ int main(int argc, char *argv[])
 	char *dev_file;
 	char tty_path[TTYPATH_LEN] = {0, };
 	char tty_type[TTYTYPE_LEN] = {0, };
-	int choice;
+	char choice;
```
Tirtha-Samsung, line 370: Why did you change the int to char? is it needed?

amogh-samsung: Minor change to facilitate accepting 'r' and 'x' as inputs for resetting and exiting.

os/board/common/crashdump.c
```diff
@@ -208,6 +210,68 @@ static int get_ext_userfs_part(size_t *size, int *index)
...
+	char filename[CONFIG_SMARTFS_MAXNAMLEN + 1];
+	char char_filesize;
+
+	/* Send acknowledgement for USERFS FILE DUMP Handshake */
+	up_lowputc('A');
+
+	/* Receive the file name to be sent to the host */
+	for (i = 0; i < CONFIG_SMARTFS_MAXNAMLEN + 1; i++) {
+		filename[i] = '\0';
+	}
```
sunghan-chang, line 230: Is this necessary?
If necessary, how about doing at definition?
```
char filename[CONFIG_... + 1] = { 0, };
```

tools/trap/trap.c
```diff
@@ -24,6 +24,14 @@
...
 
 #include "trap.h"
+#include "../../os/include/tinyara/config.h"
+
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+#ifndef CONFIG_SMARTFS_MAXNAMLEN
+#define USERFS_MAXNAMLEN 32
+#endif
```
sunghan-chang, line 34: Even you include the header, is this necessary?
If you think some config which does not have the config in the .config, it could be fs disabled.

tools/trap/trap.c
```diff
@@ -350,32 +421,35 @@ int main(int argc, char *argv[])
...
 			DUMP_FLAGS |= RAMDUMP_FLAG;
 			DUMP_FLAGS |= USERFSDUMP_FLAG;
 			break;
-		case 4:
+		case '4':
 			DUMP_FLAGS |= EXTUSERFS_DUMP_FLAG;
 			break;
-		case 5:
+		case '5':
+			DUMP_FLAGS |= FILE_DUMP_FLAG;
```
sunghan-chang, line 446: Does this support multiple command at once? If not, we can just set it without bit comparision.

os/board/common/crashdump.c
amogh-samsung: Agreed

tools/trap/trap.c
amogh-samsung: In case fs is not disabled
The application will return an error while looking for the file that the user has asked to dump.

Also, I believe that the user may be using userfs with SmartFS enabled but may not build the code themselves and may not have the .config file available.
Hence, there are other ways to provide the parameter of maximum file name length.

tools/trap/trap.c
amogh-samsung: Does not support currently,
We can do it
I will add this in a separate commit.

amogh-samsung:
> Sorry for late comment. Is it necessary at assert case to get file dump? File data is in the flash so that it will not disappear after reset.

We do not prefer having an assert case for dumping FS contents. Ideally, we should be able to dump them without crashing the device.
However, when the device is active, we might need a separate UART-console communication channel to transfer the dump along with the usual console logs.
@kishore-sn could you help me explain this point please?

Additionally, @sunghan-chang , I have wanted to make FS dump available without crashing and I will push commits to make it work if and when I do find a way.

sunghan-chang:
@amogh-samsung I think we can use a new TASH command for that. Impossible?

amogh-samsung:
> @amogh-samsung I think we can use a new TASH command for that. Impossible?

I would like to check that possibility, but would also like to have this feature included here in the TRAP tool.
What do you suggest?

sunghan-chang:
It becomes much huge file (over 500 lines!). Big file is much difficult to analyse and to modify later so that I want to confirm whether this is really necessary or not. I see no one use this tool even though we added several dump feature..

amogh-samsung:
> It becomes much huge file (over 500 lines!). Big file is much difficult to analyse and to modify later so that I want to confirm whether this is really necessary or not. I see no one use this tool even though we added several dump feature..

The intent of this change is a mixture of 'good-to-have' and 'debugging-utility'
We have used this feature internally in the past to test and debug certain other implementations.
@r-prabu could probably help explain further

os/board/common/crashdump.c
```diff
@@ -208,6 +210,65 @@ static int get_ext_userfs_part(size_t *size, int *index)
 	return 0;
 }
 
+static int send_file(void)
+{
+	int i;
+	int fd;
+	int ret;
+	struct stat st;
```
pillip8282, line 218: How about to initialize values to remove initialization warning in code review tool.

r-prabu:
> tool even though we added several dump feature

@sunghan-chang , I think Users are not aware of the options available. even in SRIB, a team from DA (WAC) mentioned this is useful during development. Let us know your opinion on this. Also this PR is kept open for a long time. If you want to change the command order we can take it in separate PR immediately after merging this.

sunghan-chang (approved):

r-prabu (approved):

PR #5740, opened 2023-01-12
mm/mm_heap: add DEBUG_SET_CALLER_ADDR() macro.

ewoodev:
Malloc API caller address is saved in alloc_call_addr of chunk(alloc node). If wrapping malloc API, The caller address becames warpped function address.

This commit adds macro(DEBUG_SET_CALLER_ADDR), that recives allocated memory address and sets caller address of current funtion to alloc_call_addr of chunk.

When warpping malloc API, have to update caller address with this macro.

Signed-off-by: eunwoo.nam <eunwoo.nam@samsung.com>

### DIFF
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -249,6 +249,15 @@ typedef size_t mmsize_t;
 #error Unknown CONFIG_ARCH option, malloc debug feature wont work.
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+/* This macro updates a caller return address in the 'mem' node.
+ * This will show the real owner of the 'mem' even it's allocated through wrapping APIs of malloc.
+ */
+#define DEBUG_SET_CALLER_ADDR(mem) heapinfo_set_caller_addr(mem, __builtin_return_address(0))
+#else
+#define DEBUG_SET_CALLER_ADDR(mem)
+#endif
+
 /* typedef is used for defining size of address space */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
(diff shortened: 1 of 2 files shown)


os/include/tinyara/mm/mm.h
```diff
@@ -249,6 +249,13 @@ typedef size_t mmsize_t;
 #error Unknown CONFIG_ARCH option, malloc debug feature wont work.
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+/* Macro sets caller address of malloc API to mem chunk.
+ * If rapping malloc APT, have to called after calling the malloc API.
+ */
+#define DEBUG_SET_CALLER_ADDR(mem) heapinfo_set_caller_addr(mem, __builtin_return_address(0))
+#endif
```
sunghan-chang, line 257: How about below?
```
#ifdef CONFIG_DEBUG_MM....
#define AA bb()
#else
#define AA
#endif
```
Then, user just calls the macro without the conditinal.

ewoodev: Yes, I think so. Thanks, Fixed.

os/include/tinyara/mm/mm.h
```diff
@@ -249,6 +249,15 @@ typedef size_t mmsize_t;
 #error Unknown CONFIG_ARCH option, malloc debug feature wont work.
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+/* Macro sets caller address of malloc API to mem chunk.
+ * If rapping malloc APT, have to called after calling the malloc API.
```
sunghan-chang, line 254: How about below?

This macro updates a caller return address in the 'mem' node.
This will show the real owner of the `mem` even it's allocated through wrapping APIs of malloc.

ewoodev: Thank you. I modified comment.

sunghan-chang (approved):

PR #5746, opened 2023-01-16
fs/select : Prevent assert when select called after close

Taejun-Kwon:


os/fs/vfs/fs_poll.c
```diff
@@ -273,36 +273,41 @@ int file_poll(FAR struct file *filep, FAR struct pollfd *fds, bool setup)
 	FAR struct inode *inode;
 	int ret = -ENOSYS;
 
-	DEBUGASSERT(filep != NULL && filep->f_inode != NULL);
+	DEBUGASSERT(filep != NULL);
```
amogh-samsung, line 276: I understand the idea
Calling select on a closed file may be a mistake from the app side, we do not need to have an assert failure.
We can just report the error.

But can you help me understand how and why the need to make this change came up?

amogh-samsung (commented):
LGTM

os/fs/vfs/fs_poll.c
Taejun-Kwon: To prevent unexpected error in product level.

bobo2108 (approved):
LGTM!

sunghan-chang (approved):

ewoodev (commented):
LGTM

jylee9613 (commented):
lgtm

PR #5748, opened 2023-01-19
os/mm: Add heap dump option to heapinfo cmd

kishore-sn:
Allow user to dump entire heap regions using heapinfo command. Add new option -d for dumping heap. The -d option may be followed by option "kernel" or the app name to indicate the heap to be dumped.


sunghan-chang:
@kishore-sn Almost build configuration fails to build. Could you check that?

sunghan-chang (approved):

PR #5751, opened 2023-01-30
log_dump.c: Fix hang issue in log dump module

kishore-sn:
If log dump is enabled, and we run heapinfo command, then there is a hang issue. This is because, the heapinfo command locks the kernel heap. But, at the same time heapinfo logs are sent to log dump module which tries to allocate buffer. Since, the heap is already locked by the heapinfo command, it results in a lock-up.

To overcome this issue, upon entering logdump api we will check if the kernel heap is locked before performing the log dump operations which require memory to be allocated or freed. The log dump operation will not be performed as long as the heap is locked. So, any logs printed during this time will not get included in the log dump.

### DIFF
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -86,13 +86,16 @@ struct log_dump_chunk_s {
 /****************************************************************************
  * Private Variables
  ****************************************************************************/
-
+static mqd_t mq_fd;
 static sq_queue_t log_dump_chunks;	/* queue for holding the logs */
 static int compress_curbytes = 0;	/* number of bytes in the current chunk that were filled */
 static int compress_rdptr = 0;		/* current pointer for compressed data read */
 static size_t log_dump_size = 0;
 static struct log_dump_chunk_s *read_node;
-static signed char *uncomp_array;	/* initial uncompressed log dump data */
+static signed char uncomp_buf[CONFIG_LOG_DUMP_NUMBUFS][CONFIG_LOG_DUMP_CHUNK_SIZE];	/* initial uncompressed log dump data */
+static bool uncomp_buf_full[CONFIG_LOG_DUMP_NUMBUFS];
+static int uncomp_idx = 0;
+static int comp_idx = 0;
 static int uncomp_curbytes = 0;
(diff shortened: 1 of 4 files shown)


os/kernel/log_dump/log_dump.c
```diff
@@ -135,13 +159,16 @@ int log_dump_init(void)
...
 
 	return LOG_DUMP_OK;
-}
 
-/* Structures used to compress last partially filled block */
-unsigned char *last_comp_block;
-size_t last_comp_block_size;
-int last_comp_block_ptr;
-bool compress_last_block;
+exit_with_outbuf:
```
jeongchanKim, line 163: All failure cases print the same message.
How about printing the log from here?
Then we can reduce the code also.

os/kernel/log_dump/log_dump.c
```diff
@@ -118,12 +124,30 @@ int log_dump_init(void)
...
+		lldbg("memory allocation failure\n");
+		goto exit_with_uncomp;
+	}
+
+	last_comp_block = allocate_compress_buffer(4, CONFIG_LOG_DUMP_CHUNK_SIZE);
+	if (last_comp_block == NULL) {
+		lldbg("memory allocation failure\n");
+		goto exit_with_outbuf;
 	}
+
```
amogh-samsung, line 149: Minor, please avoid extra line spaces

os/kernel/log_dump/log_dump.c
```diff
@@ -118,12 +124,30 @@ int log_dump_init(void)
...
+	out_buf = allocate_compress_buffer(0, CONFIG_LOG_DUMP_CHUNK_SIZE);
+	if (out_buf == NULL) {
+		lldbg("memory allocation failure\n");
+		goto exit_with_uncomp;
+	}
+
+	last_comp_block = allocate_compress_buffer(4, CONFIG_LOG_DUMP_CHUNK_SIZE);
+	if (last_comp_block == NULL) {
+		lldbg("memory allocation failure\n");
+		goto exit_with_outbuf;
```
amogh-samsung, line 147: We are making allocations at init time itself to avoid semaphore contentions later?

swasthik22 (commented):
Looks good.

sunghan-chang (approved):

PR #5756, opened 2023-02-10
os/board/rtl8721csm : fix softap fail case

keerthanalex:
Fix the softap return for fail case
After modification, get info no longer returns incorrect value for fail case


j11591 (approved):
LGTM

keerthanalex:
> LGTM

@j11591 , any update on this PR?

PR #5744, opened 2023-01-13
os/board/rtl8721csm: add DEBUG_SET_CALLER_ADDR() for malloc

ziyik:

PR #5762, opened 2023-02-17
board/rtl8720e/src/component/: Fix_wifi_disconnect_hang_issue

kaikaikaime:
Other issues:
amebalite_timerisr.c - Revert define to extern variable due to incorrect value
osif_tizenrt.c - Fix missing variable

### DIFF
--- a/os/arch/arm/src/amebalite/amebalite_timerisr.c
+++ b/os/arch/arm/src/amebalite/amebalite_timerisr.c
@@ -71,7 +71,7 @@
 /*----------------------------------------------------------------------------
   Clock Variable definitions
  *----------------------------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


sunghan-chang:
@kaikaikaime How is RTL8721CSM? Because many codes are the same with RTL8721CSM, I leave this query. No need to modify it?

kaikaikaime:
> @kaikaikaime How is RTL8721CSM? Because many codes are the same with RTL8721CSM, I leave this query. No need to modify it?

The mistake was made with new version osif_tizenrt.c for rtl8720e while bringing in modifications from #5744 , which also modifies RTL8721CSM

sunghan-chang (approved):

PR #5767, opened 2023-02-24
os/board/rtl8720e: Fix wlan idx selection

jwei5:
Fix check for wlan idx by using flag, previous implementation affected throughput

In 10.1a SDK update, STA is wlan idx 0 and SOFTAP is wlan idx 1 in Wi-Fi driver
Because application layer is using wlan idx 0 regardless of STA or SOFTAP, it needs to know whether to send to STA (idx 0) or SOFTAP (idx 1) when application is sending data to Wi-Fi driver. This is done by checking whether application layer is in SOFTAP mode
Previously, this was done using wifi_is_running(), but it affected the Wi-Fi throughput and caused a bottleneck
This PR will resolve the bottleneck by using a flag "softap_flag" to check whether application is in SOFTAP mode and send data to the Wi-Fi driver with the correct idx

### DIFF
--- a/os/board/rtl8720e/src/component/os/tizenrt/ethernetif_tizenrt.c
+++ b/os/board/rtl8720e/src/component/os/tizenrt/ethernetif_tizenrt.c
@@ -97,6 +97,9 @@
 #define IF2NAME1 '2'
 #endif
 
+/* Check if current mode is softap */
+extern int softap_flag;
+
 static void arp_timer(void *arg);
 
 #if 0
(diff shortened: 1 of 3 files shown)


sunghan-chang:
@jwei5 This is not acceptable. Could you follow TizenRT commit rules?
```
board/rtl8720e: fix wlan idx selection

blah blah <--- need explanation, what issue is, how to resolve.
```

jwei5:
@sunghan-chang Sorry about that, I have updated the initial comment with more details about the issue

sunghan-chang:
@jwei5 Thank you for update. But you modified the PR, not the commit. Could you update it on the commit?
```
$ git commit --amend
modify the title and description
$git push -f
```

sunghan-chang:
@jwei5 We don't have PR rule because it's a temporary step to merge a commit. What we think important is a commit. Because it's left in a commit history forever. When we find some history, we use a commit history.

jwei5:
@sunghan-chang I have amended the commit. Understood, I'll take note of this for future commits

sunghan-chang (approved):

PR #5769, opened 2023-02-28
external/webserver/http_client.c: Add NULL validation

aadotverma:
NULL check for body param added in http_send_response & http_send_response_with_status_msg to avoid error in strlen function.

### DIFF
--- a/external/webserver/http_client.c
+++ b/external/webserver/http_client.c
@@ -1257,13 +1257,19 @@ int http_send_response(struct http_client_t *client, int status, const char *bod
 {
 	const char* status_message = (status == 200) ? "OK" : body;
 
-	int body_len = strlen(body);
+	int body_len = 0;
+	if (body) {
+		body_len = strlen(body);
+	}
 	return http_send_response_helper(client, status, status_message, body, body_len, headers);
 }
 
 int http_send_response_with_status_msg(struct http_client_t *client, int status, const char* status_message, const char* body, struct http_keyvalue_list_t *headers)
 {
-	int body_len = strlen(body);
+	int body_len = 0;
+	if (body) {
+		body_len = strlen(body);
+	}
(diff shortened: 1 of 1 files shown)


external/webserver/http_client.c
```diff
@@ -1263,6 +1263,11 @@ int http_send_response(struct http_client_t *client, int status, const char *bod
 
 int http_send_response_with_status_msg(struct http_client_t *client, int status, const char* status_message, const char* body, struct http_keyvalue_list_t *headers)
 {
+	if (body == NULL) {
```
vibhor-m, line 1266: please check helper function called at end is accepting null body. 
You may also skip strlen and proceed. check if possible

external/webserver/http_client.c
```diff
@@ -1257,13 +1257,19 @@ int http_send_response(struct http_client_t *client, int status, const char *bod
 {
 	const char* status_message = (status == 200) ? "OK" : body;
 
-	int body_len = strlen(body);
+	int body_len = 0;
+	if (body) {
+		body_len = strlen(body);
+	}
```
sunghan-chang, line 1263: `body` and `body_len` are given to `http_send_response_helper`. But it checks the `body` with null always and `body_len` is used in body conditional. Is this really necessary?

sunghan-chang:
@aadotverma We have some commit rules and they require commit description. Could you add some explanation there?

external/webserver/http_client.c
vibhor-m: Use of strlen function with body param value as NULL will give error. So Null check is added to avoid the issue.

vibhor-m:
> @aadotverma We have some commit rules and they require commit description. Could you add some explanation there?

@aadotverma Please update your commit also. Currently you just updated the PR description

j11591 (approved):
LGTM

PR #5775, opened 2023-03-03
build/rtl8721csm/cpp_loadable : add support for cpp on loadable config

abhishek-samsung:
Here, we are supporting cpp on loadable using libsupc++_nano instead of
libsupc++. The nano version is a stripped down version of the libsupc++
and does not have the exceptions support. If we want to use cpp on loadable,
we will have to use the tizenrt docker version 1.6.0 as c++ support for
armv8m was present on the older docker versions.
```
System Information:
        Version:
                Platform: 3.1   Binary: 200204
        Commit Hash: 000ee1a
        Build User: root@=
        Build Time: 2023-03-03 15:26:22 IST
        System Time: 06 Dec 2011, 00:00:01 [s] UTC
This is WIFI App

Select Test Scenario.
        -Press R or r : Recovery Test
        -Press U or u : Binary Update Test
        -Press X or x : Terminate Tests.
TASH>>binary_manager_load_binary: Load success! [Name: app2] [Version: 190412] [Partition: A] [Compressed Binary]
Test will be finished.

TASH>>helloxx
TASH>>helloxx_main: Saying hello from the dynamically constructed instance
CHelloWorld::HelloWorld: Hello, World!!
helloxx_main: Saying hello from the instance constructed on the stack
CHelloWorld::HelloWorld: Hello, World!!
helloxx_main: Saying hello from the statically constructed instance
CHelloWorld::HelloWorld: Hello, World!!

TASH>>
```


build/configs/rtl8721csm/cpp_loadable/Make.defs
```diff
@@ -0,0 +1,252 @@
+###########################################################################
+#
+# Copyright 2020 Samsung Electronics All Rights Reserved.
```
jeongchanKim, line 3: Please change the year to 2023.

apps/examples/helloxx/Kconfig
```diff
@@ -6,7 +6,7 @@
 config EXAMPLES_HELLOXX
```
amogh-samsung, line 6: In the PR description, did you mean "since c++ support was **NOT** available in previous docker versions"?

os/binfmt/libelf/libelf_ctors.c
```diff
@@ -175,14 +175,22 @@ int elf_loadctors(FAR struct elf_loadinfo_s *loadinfo)
 		if ((shdr->sh_flags & SHF_ALLOC) == 0) {
 			/* Allocate memory to hold a copy of the .ctor section */
 
+			loadinfo->ctoralloc = (uint32_t)kmm_malloc(ctorsize);
+			if (!loadinfo->ctoralloc) {
+				berr("Failed to allocate memory for .ctors\n");
```
amogh-samsung, line 180: Is ".ctors" a typo here?

apps/examples/helloxx/Kconfig
```diff
@@ -6,7 +6,7 @@
...
 	bool "\"Hello, World!\" C++ example"
 	default n
-	depends on HAVE_CXXINITIALIZE
+	depends on HAVE_CXXINITIALIZE || BINFMT_CONSTRUCTORS
```
sunghan-chang, line 9: When BINFMT is enabled and HAVE_CXXINIT is disabled, is this working?

os/binfmt/libelf/libelf_ctors.c
abhishek-samsung: We are allocating memory for .ctors section. So it's fine.

apps/examples/helloxx/Kconfig
abhishek-samsung: In the case of flat build, CONFIG_HAVE_CXXINITIALIZE is for calling the static object constructors for kernel

https://github.com/Samsung/TizenRT/blob/8a30134393295de817a8a41de90749489ab44a0f/apps/system/init/init.c#L124-L129

CONFIG_BINFMT_CONSTRUCTORS also does the same thing while loading the application. So they both need not be enabled for loadable as we have cpp only in the app side.

https://github.com/Samsung/TizenRT/blob/3fc8caed54420a3dfa6e002e8807feee20407142/os/binfmt/binfmt_execmodule.c#L116-L132

os/include/tinyara/binfmt/elf.h
```diff
@@ -117,6 +117,8 @@ struct elf_loadinfo_s {
 	uint8_t *iobuffer;			/* File I/O buffer */
 
 #ifdef CONFIG_BINFMT_CONSTRUCTORS
+	FAR void *ctoralloc;			/* Memory allocated for ctors */
+	FAR void *dtoralloc;			/* Memory allocated dtors */
```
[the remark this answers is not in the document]
abhishek-samsung: Done

sunghan-chang (approved):

PR #5777, opened 2023-03-08
os/board/rtl8721csm : reset scan with ssid setting to default

keerthanalex:
Reset the  "rltk_wlan_enable_scan_with_ssid_by_extended_security" setting back to default 0 
Modified for both success and fail case of scan with ssid


j11591 (approved):
LGTM

PR #5785, opened 2023-03-21
os/board/rtl8721csm: Mask partial connected AP information log

lu04-wang:
Only show first 4 character of AP name, the rest is shown as * during wifi connection process. only show last 4 digit of MAC address, the rest is shown as * during wifi connection process


j11591 (approved):
LGTM

PR #5788, opened 2023-03-24
examples/lw_aifw_sine_test_app: Includes stdlib.h for free API

vibhor-m:
examples/lw_aifw_sine_test_app: Includes stdlib.h for free API

PR #5789, opened 2023-03-24
src/lw_aifw: Public classes of AI Framework moved to include/lw_aifw

vibhor-m:
framework/src/lw_aifw:
Public classes of AIFramework moved to framework/include/lw_aifw

apps/examples/lw_aifw_sine_test_app:
DataSource moved to application scope
Modify sample application to sync with AI Framework interfaces


framework/include/lw_aifw/AIServiceCore.h
```diff
@@ -21,8 +21,8 @@
 #include <signal.h>
 #include "lw_aifw/lw_aifw.h"
 #include "lw_aifw/AIDataSource.h"
-#include "AIDataBuffer.h"
-#include "AIModel.h"
+#include "../../src/lw_aifw/include/AIDataBuffer.h"
+#include "../../src/lw_aifw/include/AIModel.h"
```
sunghan-chang, line 25: Public header needs them, then are not they public headers?

vibhor-m: These are not public headers, they are used by AI Framework public headers but not by applications. Public header file declares their objects for its internal implementation.

framework/include/lw_aifw/AIServiceCore.h
```diff
@@ -21,8 +21,8 @@
 #include <signal.h>
 #include "lw_aifw/lw_aifw.h"
 #include "lw_aifw/AIDataSource.h"
-#include "AIDataBuffer.h"
-#include "AIModel.h"
+#include "../../src/lw_aifw/include/AIDataBuffer.h"
```
Taejun-Kwon, line 24: It should be synced with prod branch, currently they're different

vibhor-m: AI Framework has public interfaces at framework/include/lw_aifw.
Public interfaces use lw_aifw implementation specific header files at location framework/src/lw_aifw/include.
Sample application is made in CPP & includes AIServiceCore.h.
Referring to suggestion https://github.com/Samsung/TizenRT/pull/5788#discussion_r1147397427, I also think application should not directly include non public header files.

apps/examples/lw_aifw_sine_test_app/AICSVDataSource.h
```diff
@@ -19,7 +19,7 @@
 #pragma once
 #include "lw_aifw/lw_aifw_result.h"
 #include "lw_aifw/AIDataSource.h"
-#include "AICSVReader.h"
+#include "lw_aifw/AICSVReader.h"
```
Taejun-Kwon, line 22: why it is moved to app folder?
This is basic class which aifw provide

vibhor-m: As discussed, I will change

framework/include/lw_aifw/AIServiceCore.h
Taejun-Kwon: I think we should clean header files.
Public interface should not refer to private interface.
In other words, header in fw/include should not refer to fw/src/include, on the contrary, it acceptable.

Taejun-Kwon (approved):

PR #5793, opened 2023-03-30
kernel/task: Add sched_lock() to task_terminate_unloaded() API to avoid

thapav:
race condition during binary unloading & reloading scenarios.


ewoodev (approved):
LGTM~

While disabling the MPU region, if context switching and unloding TCB gets CPU, could became race condition.

sunghan-chang (approved):

PR #5796, opened 2023-03-31
kernel/semaphore: Remove duplicate ASSERT() in sem_post.c

ewoodev:
ASSERT_INFO() and ASSERT() are checking same condition. (sem->semcount < SEM_VALUE_MAX)
So, This commit removes ASSERT().

Signed-off-by: eunwoo.nam <eunwoo.nam@samsung.com>


thapav (approved):
LGTM

sunghan-chang:
Let's re-title like below
```
kernel/semaphore: Remove duplicate ASSERT() in sem_post.c
```

sunghan-chang (approved):

PR #5799, opened 2023-04-03
os/Makefile.unix: remove the CONFIG_RAW_BINARY dependency 

swasthik22:
Remove dependency on CONFIG_RAW_BINARY from defconfig and always generate .bin format. Move the .hex creation logic after the addition of samsung headers.

### DIFF
--- a/os/Kconfig
+++ b/os/Kconfig
@@ -315,12 +315,12 @@ config MOTOROLA_SREC
 		and should not be selected if you are not using the GNU toolchain.
 
 config RAW_BINARY
-	bool "Raw binary format"
-	default n
+	bool
+	default y
 	---help---
-		Create the tinyara.bin in the raw binary format that is used with many
-		different loaders using the GNU objcopy program.  This option
-		should not be selected if you are not using the GNU toolchain.
+		Create the tinyara.bin in the raw binary format that is used with many 
+		different loaders using the GNU objcopy program. 
+		This configuration should always be enabled mandatorily in TizenRT.	
 
 menuconfig UBOOT_UIMAGE
 	bool "U-Boot uImage"
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -222,9 +222,9 @@ endif
 ifeq ($(CONFIG_RAW_BINARY),y)
 BIN_EXT = bin
 endif
+KERNEL_START=$(CONFIG_FLASH_START_ADDR)
 
 # This is the name of the final target (relative to the top level directorty)
-
 BIN_EXE = tinyara$(EXEEXT)
 BIN = $(OUTBIN_DIR)/$(BIN_EXE)
 
@@ -462,14 +462,6 @@ pass2: pass2deps
 		cp -f $(BIN) /tftpboot/$(BIN_EXE).${CONFIG_ARCH}; \
 	fi
 
-ifeq ($(CONFIG_INTELHEX_BINARY),y)
-	$(Q) echo "CP: $(BIN_EXE).hex"
-	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O ihex $(BIN) $(BIN).hex
-endif
-ifeq ($(CONFIG_MOTOROLA_SREC),y)
-	$(Q) echo "CP: $(BIN_EXE).srec"
-	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O srec $(BIN) $(BIN).srec
-endif
 ifeq ($(CONFIG_RAW_BINARY),y)
 	$(Q) echo "CP: $(BIN_EXE).bin"
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O binary $(BIN) $(BIN).bin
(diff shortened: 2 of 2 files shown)


os/Makefile.unix
```diff
@@ -544,6 +527,17 @@ endif
...
 	$(Q) python ./tools/set_bininfo.py $(BIN_EXT)
 	$(Q) $(call MAKE_SAMSUNG_HEADER, $(BIN_EXE), $(BIN_EXT))
+
+ifeq ($(CONFIG_INTELHEX_BINARY),y)
+	$(Q) echo "CP: $(BIN_EXE).hex"
+	$(Q) $(OBJCOPY) $(OBJCOPYARGS) --change-addresses $(KERNEL_START) -I binary -O ihex $(BIN).bin $(BIN).hex
+endif
+ifeq ($(CONFIG_MOTOROLA_SREC),y)
+	$(Q) echo "CP: $(BIN_EXE).srec"
+	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O srec $(BIN) $(BIN).srec
```
sunghan-chang, line 537: Does not SREC need the `--change-addresses` option?

build/configs/artik053/STDK/defconfig
```diff
@@ -29,7 +29,6 @@ CONFIG_BUILD_FLAT=y
 #
 # CONFIG_INTELHEX_BINARY is not set
 # CONFIG_MOTOROLA_SREC is not set
-CONFIG_RAW_BINARY=y
```
sunghan-chang, line 32: You removed this in all of defconfigs. But I can't find Kconfig change related this. Could you check it?

sunghan-chang:
@swasthik22 Please check belows
1. TizenRT does not allow no commit description. Please write some description why and what you modified.
2. Let's split the commit as
    1. remove bin format in Kconfig and Makefile
    2. Apply the config removal in all of defconfig

sunghan-chang:
@swasthik22 Could you let me know plan for this?

build/configs/artik053/STDK/defconfig
swasthik22: Yes. I had missed the kconfig. fixing it.

os/Makefile.unix
swasthik22: it needs as per my opinion. But no concrete proof that I found online. Since hex needs this option and is using the same command It should be needed.

os/Kconfig
```diff
@@ -314,14 +314,6 @@ config MOTOROLA_SREC
...
 		and should not be selected if you are not using the GNU toolchain.
 
-config RAW_BINARY
-	bool "Raw binary format"
-	default n
-	---help---
-		Create the tinyara.bin in the raw binary format that is used with many
-		different loaders using the GNU objcopy program.  This option
-		should not be selected if you are not using the GNU toolchain.
-
```
Tirtha-Samsung, line 324: LGTM~

build/configs/artik053/STDK/defconfig
```diff
@@ -29,7 +29,6 @@ CONFIG_BUILD_FLAT=y
 #
```
Tirtha-Samsung, line 29: Please add commit description

swasthik22: The rebase was updated forgot to continue the rebase.

build/configs/artik053/STDK/defconfig
sunghan-chang: Sorry for late suggestion.
RAW binary is mandatory on TizenRT but in output extension on menuconfig does not have it. And there is no content for this. To know it easily, how about keeping this config, hidden it from menuconfig and set it default y?
How do you think? Then we don't need to modify all of defconfigs.

swasthik22: By default all of the defconfig has the binary as y. However it is still configurable. So unknowingly if someone changes it. It will cause build failures. So i think it is better to remove the configs itself so no configuration is possible by menuconfig or by force by updating the defconfig. Just to be on safe side.

sunghan-chang: That's why I suggest hiding it form menuconfig.
```
config AAAAA
    bool
```
If you remove "description" after type, menuconfig does not show it up. So user can't disable it.

sunghan-chang:
```
config AAA
   bool "description"
```
User can control this on the menuconfig (on / off).
```
config AAA
   bool
   default y
```
This is enabled always and user can't disable it.

os/Kconfig
```diff
@@ -315,12 +315,8 @@ config MOTOROLA_SREC
 		and should not be selected if you are not using the GNU toolchain.
 
 config RAW_BINARY
-	bool "Raw binary format"
-	default n
-	---help---
-		Create the tinyara.bin in the raw binary format that is used with many
-		different loaders using the GNU objcopy program.  This option
-		should not be selected if you are not using the GNU toolchain.
```
sunghan-chang, line 323: I think we can keep the description except last sentence. And we can add some description that this is mandatory on TizenRT.

os/dbuild.sh
```diff
@@ -54,7 +54,7 @@ BINFILE=
 function FIND_BINFILE()
 {
 	source ${CONFIGFILE}
-	BOARDNAME=${CONFIG_ARCH_BOARD}
+	BOARDNAME=${CONFIG_ARCH_BOAD}}
```
sunghan-chang, line 57: This looks wrong. Previous one is correct. Could you check again?

sunghan-chang:
Let's split the commit as two changes.
1.
```
os/Kconfig: set RAW_BINARY by default

The bin output is mandatory on TizenRT. blah blah
```

2. 
```
os/Makefile.unix: convert hex and srec from bin

blah blah
```

os/dbuild.sh
```diff
@@ -54,7 +54,7 @@ BINFILE=
 function FIND_BINFILE()
 {
 	source ${CONFIGFILE}
-	BOARDNAME=${CONFIG_ARCH_BOARD}
+	BOARDNAME=${CONFIG_ARCH_BOARD}}
```
ewoodev: Could you check this `}`

os/dbuild.sh
```diff
@@ -54,7 +54,7 @@ BINFILE=
 function FIND_BINFILE()
 {
 	source ${CONFIGFILE}
-	BOARDNAME=${CONFIG_ARCH_BOARD}}
```
ewoodev: It could have been added '}', during commit split

sunghan-chang:
`Therefor -> Therefore` in the first commit description

sunghan-chang (approved):

PR #5792, opened 2023-03-30
build/tools/amebad: Modify random seed accuracy

ziyik:
extend initial random seed to 32 bits for more accuracy

PR #5803, opened 2023-04-06
build/tools/amebad: Fix the secure boot error caused by commit 54fde196895da0f8d67aeca6914d8fdf8d8b70f5

fengyunlai:
km0_image2.bin in 54fde196895da0f8d67aeca6914d8fdf8d8b70f5 missed CRC attached at the end of binary before padding. This will affect secure boot modules' integrity check for km0_image2.bin.
PR related: https://github.com/Samsung/TizenRT/pull/5792


sunghan-chang (approved):

sunghan-chang:
@jylee9613 Could you verify this with signing and without signing?

jylee9613 (approved):
I tested in both 3.1 and 4.0 (with/without binary signing), and both of them work well.
LGTM!

PR #5807, opened 2023-04-10
os/board/rtl8721csm: Add support for client side connection parameter update

yeetee179:


framework/include/ble_manager/ble_common.h
```diff
@@ -69,11 +69,17 @@ typedef struct {
 	ble_addr_type_e type;
 } ble_addr;
 
+typedef enum {
+	SLAVE_CONN_PARAM_UPDATE,
```
thapav, line 73: Are these enums used anywhere ?
If not these can be removed

yeetee179: @thapav The enum name is changed to match the naming convention.

ydlee-sec (approved):
LGTM

PR #5809, opened 2023-04-12
apps/shell : Apply PBKDF2 for TASH password

jylee9613:
Change TASH password encryption from SHA256 to PBKDF2 which already applied in previous version.
All of values used in this PR are sample.
- key length : 32 bytes (64 characters)
- salt : hard-coded with encrypted value (length = 16 bytes)
- iteration : 777
- default password : tizenrt

### DIFF
--- a/apps/shell/Kconfig
+++ b/apps/shell/Kconfig
@@ -59,18 +59,35 @@ config TASH_SCRIPT
 
 config SECURED_TASH
 	bool "Enable Secure shell"
-	depends on NET_SECURITY_TLS
+	select NET_SECURITY_TLS
+	select MBEDTLS_PKCS5_C
 	default n
 
 if SECURED_TASH
 
-config TASH_PASSWORD_SHA256
-	string "SHA256 encrypted password (32 Bytes)"
-	default "436890f31fcd6b154a0e49df9f190bd8b1dcd4b917a7a80b2d928452c22901f7"
+config TASH_PASSWORD_HASH
+	string "PBKDF2 encrypted password (32 Bytes)"
+	default "3D9740229177EBE6B11D0371342308881AE1463D1142E9FF932AB180AF4E8BD6"
 	---help---
-		Enable password to enter TASH. The password is hashed by SHA256 so that
+		Enable password to access TASH command line. The password is hashed by PBKDF2 so that
 		the length of this config should be 32 bytes (64 characters).
-		default : hash value of "TizenRT" (Password : TizenRT)
+		default : hash value of "tizenrt" (Password : tizenrt)
+
+config TASH_SALT_ENCRYPTED
+	string "AES encrypted salt (16 Bytes)"
+	default "00112233445566778899AABBCCDDEEFF"
+	---help---
+		Enable salt value for TASH password encryption. The salt is encrypted by AES128 so that
+		the length of this config should be 16 bytes (32 characters).
+		default : encrypted value of "00112233445566778899AABBCCDDEEFF"
+		Please encrypt this value with AES key and set this value.
+
+config TASH_HASH_ITERATION
+	int "Number of PBKDF2 iteration"
+	default 777
+	---help---
+		The number of hash iteration used in PBKDF2.
+
 endif
 
 endif
--- a/apps/shell/tash_security.c
+++ b/apps/shell/tash_security.c
@@ -23,15 +23,134 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
-#include <mbedtls/sha256.h>
 #ifdef CONFIG_TASH_REBOOT
 #include <sys/boardctl.h>
 #endif
+#include <security/security_crypto.h>
+#include <mbedtls/pkcs5.h>
 #include "tash_internal.h"
 
-#define SHA256_HASH_nCHAR  (64) // 32 bytes
-#define SHA256_HASH_nBYTE  (32) // 32 bytes
-#define CONVERT_nCHAR      (sizeof(unsigned long) * 2) // unsigned long for strtoul
+#define CONVERT_nCHAR           (sizeof(unsigned long) * 2) // unsigned long for strtoul
+
+#define PBKDF2_KEY_nCHAR        (64) // 32 bytes
+#define PBKDF2_KEY_LEN          (32) // 32 bytes
+#define PBKDF2_SALT_LEN         (16) // 16 bytes
+#define PBKDF2_SALT_nCHAR       (32) // 16 bytes
+#define AES_BLOCK_SIZE          (16)
+#define AES128_KEY_SLOT         "ss/38" // slot where AES key is stored
+
+
+static int salt_aes_decrypt(char *input, unsigned int input_size, char *output, unsigned int output_size) 
+{
+	if ((input_size <= 0) || (input_size % AES_BLOCK_SIZE)) {
+		printf("Invalid AES key input size\n");
+		printf("input size : %d\n", input_size);
+		return -1;
+	}
+
+	if (input == NULL) {
+		printf("AES input is null\n");
+		return -1;
+	} 
+
+	if (output_size < input_size) {
+		printf("Invalid AES key output size\n");
+		printf("output size : %d\n", output_size);
+		return -1;
+	}
+
+	if (output == NULL) {
+		printf("AES output is null\n");
+		return -1;
+	}
+
+	security_handle hnd;
+	security_data encrypt_data;
+	security_data decrypt_data;
+	security_aes_param aes_param;
+	security_error ret;
+	char *tmp = NULL;
+
+	ret = security_init(&hnd);
+	if (ret != SECURITY_OK) {
+		printf("security_crypto : security_init failed. ret : %d\n", ret);
+		return -1;
+	}
+
+	encrypt_data.data = input;
+	encrypt_data.length = input_size;
+
+	decrypt_data.length = output_size;
+
+	aes_param.mode = AES_CBC_NOPAD;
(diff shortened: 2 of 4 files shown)


apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+		printf("Failed to setup hash function");
+		return -1;
+	}
+
+	ret = mbedtls_pkcs5_pbkdf2_hmac(&sha_ctx, (const unsigned char *)input, input_len, (const unsigned char *)salt, salt_len, iteration, PBKDF2_KEY_LEN, pbkdf2_output);
+	if (ret != 0) {
+		printf("Call to mbedtls PBKDF2 function failed\n");
+		return -1;
+	}
+	mbedtls_md_free(&sha_ctx);
```
jeongchanKim, line 106: Is it okay not to free the `sha_ctx` for every failure case above?

apps/shell/Kconfig
```diff
@@ -64,13 +64,13 @@ config SECURED_TASH
...
 if SECURED_TASH
 
-config TASH_PASSWORD_SHA256
-	string "SHA256 encrypted password (32 Bytes)"
-	default "436890f31fcd6b154a0e49df9f190bd8b1dcd4b917a7a80b2d928452c22901f7"
+config TASH_PASSWORD_HASH
+	string "PBKDF2 encrypted password (32 Bytes)"
+	default "3d9740229177ebe6b11d0371342308881ae1463d1142e9ff932ab180af4e8bd6"
 	---help---
 		Enable password to enter TASH. The password is hashed by SHA256 so that
```
sunghan-chang, line 71: Could you check below?
`hashed by SHA256`

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+#include <mbedtls/pkcs5.h>
+#include <security/security_crypto.h>
 #include "tash_internal.h"
 
-#define SHA256_HASH_nCHAR  (64) // 32 bytes
-#define SHA256_HASH_nBYTE  (32) // 32 bytes
-#define CONVERT_nCHAR      (sizeof(unsigned long) * 2) // unsigned long for strtoul
+#define CONVERT_nCHAR			(sizeof(unsigned long) * 2) // unsigned long for strtoul
+
+#define PBKDF2_SALT_ENCRYPTED	"\x50\x3B\x3C\x3E\xBF\x0D\x3D\x2C\x9D\x6E\xA2\xF3\x62\xF5\x0F\x9F"
```
sunghan-chang, line 32: This and the iteration value could be changed by user. Let's provide them as configs.

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+#define PBKDF2_SALT_LEN			(16)
+#define PBKDF2_KEY_nCHAR		(64) // 32 bytes
+#define PBKDF2_KEY_LEN			(32) // 32 bytes
+#define PBKDF2_ITERATION		(777)
+
+#define AES_BLOCK_SIZE			(16)
+#define AES128_KEY				"ss/1"
```
sunghan-chang, line 39: Could you align the value with other lines?

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+
+
+security_error salt_aes_decrypt(char *input, unsigned int input_size, char *output, unsigned int output_size)
```
Taejun-Kwon, line 42: I think there's no reason for security_error type because it defined in lower layer.
tash should have it's own type instead of lower's

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+	encrypt_data.data = input;
+	encrypt_data.length = input_size;
+
+	decrypt_data.length = output_size;
+
+	aes_param.mode = AES_CBC_NOPAD;
+	aes_param.iv = NULL;
+	aes_param.iv_len = 0;
+
+	if (0 != crypto_aes_decryption(hnd, &aes_param, AES128_KEY, &encrypt_data, &decrypt_data)) {
```
Taejun-Kwon, line 68: crypto_aes_decryption will return various type of error.
so it should print out type of error, it will be helpful when you do debugging
```
security_error ret = crypto_aes_decryption(...);
if (ret != SECURITY_OK) {
    printf("blahblah ret : %d\n", ret);
    return ERROR;
}
```

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+		return SECURITY_INVALID_INPUT_PARAMS;
+	}
+
+	security_handle hnd;
+	security_data encrypt_data;
+	security_data decrypt_data;
+	security_aes_param aes_param;
+	char *tmp = NULL;
+
+	if (0 != security_init(&hnd)) {
```
Taejun-Kwon, line 54: same as below comment

apps/shell/tash_security.c
```diff
@@ -23,43 +23,129 @@
...
+#define PBKDF2_KEY_LEN			(32) // 32 bytes
+#define PBKDF2_ITERATION		(777)
+
+#define AES_BLOCK_SIZE			(16)
+#define AES128_KEY				"ss/1"
+
+
+security_error salt_aes_decrypt(char *input, unsigned int input_size, char *output, unsigned int output_size) 
+{
+	if ((input_size <= 0) || (input == NULL) || (output_size < input_size) || (output == NULL) || (input_size % AES_BLOCK_SIZE)) {
```
Taejun-Kwon, line 44: Below style is more good for readability. 
```
if ((input_size <=0 || (input_size % AES_BLOCK_SIZE != 0) {
    printf("blahblah input_size : %d\n", input_size);
    return SECURITY_INVALID_INPUT_PARAMS;
}
if (input == NULL) {
   printf("input is null\n");
   return SECURITY_INVALID_INPUT_PARAMS;
}
....

File: apps/shell/tash_security.c [hunk 9]
```diff
@@ -23,43 +23,167 @@
...
+		mbedtls_md_free(&sha_ctx);
+		return -1;
+	}
+
+	mbedtls_md_free(&sha_ctx);
+
+	return 1;
+}
+
+void convert_str_to_hex(char *input, unsigned char *output, int char_len)
```
sunghan-chang, line 134: Only this file uses this internally. Let's add `static`.

KEITHNN: Hi what happen to the page? Straight line please...

apps/shell/Kconfig
```diff
@@ -64,13 +64,28 @@ config SECURED_TASH
...
 
-config TASH_PASSWORD_SHA256
-	string "SHA256 encrypted password (32 Bytes)"
-	default "436890f31fcd6b154a0e49df9f190bd8b1dcd4b917a7a80b2d928452c22901f7"
+config TASH_PASSWORD_HASH
+	string "PBKDF2 encrypted password (32 Bytes)"
+	default "3D9740229177EBE6B11D0371342308881AE1463D1142E9FF932AB180AF4E8BD6"
 	---help---
-		Enable password to enter TASH. The password is hashed by SHA256 so that
+		Enable password to enter TASH. The password is hashed by PBKDF2 so that
```
amogh-samsung, line 71: How about phrasing this as "Enable password to access TASH command line ..."

apps/shell/tash_security.c
```diff
@@ -23,43 +23,167 @@
...
+
+	ret = mbedtls_pkcs5_pbkdf2_hmac(&sha_ctx, (const unsigned char *)input, input_len, (const unsigned char *)salt, salt_len, iteration, PBKDF2_KEY_LEN, pbkdf2_output);
+	if (ret != 0) {
+		printf("Call to mbedtls PBKDF2 function failed\n");
+		mbedtls_md_free(&sha_ctx);
+		return -1;
+	}
+
+	mbedtls_md_free(&sha_ctx);
+
```
amogh-samsung, line 130: Would you consider using "goto" for all failure cases above to unify the mbedtls_md_free() calls?

sunghan-chang (approved):

apps/shell/Kconfig
```diff
@@ -59,18 +59,35 @@ config TASH_SCRIPT
 
 config SECURED_TASH
 	bool "Enable Secure shell"
-	depends on NET_SECURITY_TLS
+	depends on NET_SECURITY_TLS && ENABLE_IOTIVITY && ENABLE_IOTIVITY_SECURED
```
sunghan-chang, line 62: I understand to have a dependency with TLS, but don't know why it has iotivity dependencies. Could you check it?

jylee9613: `MBEDTLS_PKCS5_C` should be enabled to use `mbedtls_pkcs5_pbkdf2_hmac`.
But `MBEDTLS_PKCS5_C` is related to `MBEDTLS_OCF_PATCH`. 

```
#undef MBEDTLS_PKCS5_C
#if defined(MBEDTLS_OCF_PATCH)
#define MBEDTLS_PKCS5_C
#endif
```

MBEDTLS_OCF_PATCH is enabled when enables IOTIVITY_SECURED. That's why I add dependenciy in Iotivity.

apps/shell/Kconfig
```diff
@@ -59,7 +59,8 @@ config TASH_SCRIPT
 
 config SECURED_TASH
 	bool "Enable Secure shell"
-	depends on NET_SECURITY_TLS
+	select NET_SECURITY_TLS
+	select MBEDTLS_PKCS5_C
```
sunghan-chang, line 63: Let's include this change in another commit, not here

sunghan-chang (approved):

PR #5810, opened 2023-04-12
up_assert: fix used stack log of current task

ewoodev:
When assert, There is problem that used stack and stack size of current task are same even through stack is not full.

The reason is below.
1. The stack size is (stack top address - stack allocated address + 4)
2. Check used stack of current task. start : (stack top address - stack size)
	end   : (start + stack size)
So, checking range : (stack allocated address - 4) ~ (stack top address)
This means is checking address not range of stack.

Also, stack_color is set to (stack allocate address) to (stack top address + 4) and This check uses stack color.

Therefore, This commit changes start address of check to (stack top addresss - stack size + 4)


ewoodev:
need to check setting stack size

PR #5811, opened 2023-04-14
correct the range to check stack_color

ewoodev:
1. 
The range of setting stack_color or checking stack_color is not the same.

In set stack_color:
	start: (stackbase + 3) & ~3  //align up
	end  : (stackbase + stacksize) & ~3  //align down
In check stack_color:
	start: (stackbase) & ~3  //align down
	end  : (stackbase + stacksize + 3) & ~3  //align up

So, It can check the range that is not set to stack_color. and it can make incorrect the stack usage log.

Therefore, This commit correct range of check stack color.
And it modifies to use STACK_ALIGN_UP, STACK_ALIGN_DOWN in setting stack_color and checking stack_color.

2. 
when using CONFIG_MPU_STACK_GUARD_SIZE, we should check if CONFIG_MPU_STACK_OVERFLOW_PROTECTION is enabled
This should be matched with create stack function.

This PR adds to check CONFIG_MPU_STACK_OVERFLOW_PROTECTION in arch/arm/common


os/arch/arm/src/common/up_checkstack.c
```diff
@@ -72,6 +72,25 @@
...
+#ifdef __ARM_EABI__
+#define CONFIG_STACK_ALIGNMENT 8
+#else
+#define CONFIG_STACK_ALIGNMENT 4
+#endif
+#endif
+
+/* Stack alignment macros */
+
+#define STACK_ALIGN_MASK    (CONFIG_STACK_ALIGNMENT-1)
```
sunghan-chang, line 90: (CONFIG_XX - 1)
please add spaces around `-`.

ewoodev: Thank you, Fixed!~

os/arch/arm/src/common/up_checkstack.c
```diff
@@ -72,6 +72,25 @@
...
+#define CONFIG_STACK_ALIGNMENT 4
+#endif
+#endif
+
+/* Stack alignment macros */
+
+#define STACK_ALIGN_MASK    (CONFIG_STACK_ALIGNMENT - 1)
+#define STACK_ALIGN_DOWN(a) ((a) & ~STACK_ALIGN_MASK)
+#define STACK_ALIGN_UP(a)   (((a) + STACK_ALIGN_MASK) & ~STACK_ALIGN_MASK)
+
```
abhishek-samsung, line 93: I think it's better to define them in a common header instead of duplicating the code. 

https://github.com/Samsung/TizenRT/blob/29151cd53f5fe15e1a788689d1a881435f0ad5ae/os/arch/arm/src/common/up_createstack.c#L92-L109

os/arch/arm/src/common/up_createstack.c
```diff
@@ -297,9 +297,9 @@ void up_stack_color(FAR void *stackbase, size_t nbytes)
 {
 	/* Take extra care that we do not write outsize the stack boundaries */
```
amogh-samsung, line 298: Could you correct "outsize" to "outside" please?

os/arch/arm/src/common/up_initialize.c
```diff
@@ -131,26 +131,16 @@ static void up_calibratedelay(void)
 #if defined(CONFIG_STACK_COLORATION) && CONFIG_ARCH_INTERRUPTSTACK > 3
```
amogh-samsung, line 131: Hi, requesting you to consider re-phrasing the commit description a little: -
```
irq stack sets INTSTACK_COLOR when initialized, but while checking, STACK_COLOR is used in up_assert().
The setting logic and check logic must be consistent.

This commit changes irq_stack color to same STACK_COLOR during initialization.
Since we use stack color, we should check CONFIG_STACK_COLORATION.
This commit adds to check CONFIG_STACK_COLORATION in up_color_nestirqstack().
```

os/arch/arm/src/common/up_internal.h
```diff
@@ -223,6 +222,31 @@ extern "C" {
...
+/* The symbol  __ARM_EABI__ is defined by GCC if EABI is being used.  If you
+ * are not using GCC, make sure that CONFIG_STACK_ALIGNMENT is set correctly!
+ */
+
+#ifdef __ARM_EABI__
+#define CONFIG_STACK_ALIGNMENT 8
+#else
+#define CONFIG_STACK_ALIGNMENT 4
+#endif
+#endif
```
kishore-sn, line 241: These changes look good. However, to make it more accurate and complete, we could change the declaration of interrupt stacks in the exception.S file. Currently, the interrupt stack is always aligned to 4 bytes as follows. But it must be either 4 or 8 bytes. 

```
        .align  4
g_nestedirqstkalloc:
        .skip   (CONFIG_ARCH_NESTED_IRQ_STACK_SIZE & ~3)
g_nestedirqstkbase:

```

os/arch/arm/src/common/up_internal.h
```diff
@@ -193,7 +193,6 @@
  */
 
 #define STACK_COLOR    0xdeadbeef
-#define INTSTACK_COLOR 0xdeadbeef
 #define HEAP_COLOR     'h'
```
jylee9613, line 197: Could you explain why erase INTSTACK_COLOR?

PR #5815, opened 2023-04-21
[ble manager] add a omitted api declaration to header file

ydlee-sec:
add a omitted api declaration to header file


hs36-kim (approved):
LGTM

PR #5817, opened 2023-04-24
dbuild.sh, dutils.sh: add GET_SPECIFIC_DOCKER_IMAGE functiuon

ewoodev:
docker image checking is done only build command.
so It adds CHECK_DOCKER_IMAGE function that check docker image and pull docker image. 
And apply in all command.

On the other hand, This commit is for support other docker repository. 
In CHECK_DOCKER_IMAGE function, It can add logic that for use other docker reposiotry image.

### DIFF
--- a/os/dbuild.sh
+++ b/os/dbuild.sh
@@ -24,6 +24,8 @@ TOPDIR="${OSDIR}/.."
 BUILDDIR="${TOPDIR}/build"
 BINDIR="${BUILDDIR}/output/bin"
 CONFIGDIR="${BUILDDIR}/configs"
+DOCKER_IMAGE=
+DOCKER_PUBLIC_IMAGE="tizenrt/tizenrt"
 DOCKER_VERSION="1.5.6"
 
 STATUS_LIST="NOT_CONFIGURED BOARD_CONFIGURED CONFIGURED BUILT PREPARE_DL DOWNLOAD_READY"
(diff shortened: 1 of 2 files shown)


os/dutils.sh
```diff
@@ -287,7 +313,8 @@ function TRAP_RUN()
 	echo "Executing: $TRAPCMD" | tee $OUTPUTFILE
 	echo "" >> $OUTPUTFILE
 	# execute TRAP script
-	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -it -w /root/tizenrt/tools/trap --privileged tizenrt/tizenrt:1.5.6 python3.7 $TRAPCMD | tee -a $OUTPUTFILE
+	CHECK_DOCKER_IMAGE 1.5.6
```
sunghan-chang, line 316: Is this intended?

os/dutils.sh
```diff
@@ -287,7 +313,8 @@ function TRAP_RUN()
...
+	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -it -w /root/tizenrt/tools/trap --privileged ${DOCKER_IMAGE}:1.5.6 python3.7 $TRAPCMD | tee -a $OUTPUTFILE
```
sunghan-chang, line 317: `${DOCKER_IMAGE}:1.5.6`
Why do you use hard-coded version instead of DOCKER_VERSION?

os/dbuild.sh
```diff
@@ -46,6 +47,30 @@ if ! which docker > /dev/null; then
 	exit 1
 fi
 
+# check docker image and pull docker image
+function CHECK_DOCKER_IMAGE()
```
sunghan-chang, line 51: I think this function's main operation is getting a specific version.
How about GET_SPECIFIC_DOCKER_IMAGE?

os/dutils.sh
ewoodev: I need to check why it is only 1.5.6
I'll check version.

jylee9613: I think you should check this.
https://github.com/Samsung/TizenRT/pull/5819

os/dbuild.sh
```diff
@@ -168,7 +193,8 @@ function BUILD_TEST()
 {
 	# execute a shell script for build test
 	pushd ${OSDIR} > /dev/null
-	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged tizenrt/tizenrt:${DOCKER_VERSION} bash -c "./tools/build_test.sh"
+	CHECK_DOCKER_IMAGE ${DOCKER_VERSION}
```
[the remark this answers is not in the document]
ewoodev: Thanks, if will docker version became same, I'll modify this.
and This could feel that late running dbuild.sh, because the menu does not appear immediately
 so I'll add to print "check docker image"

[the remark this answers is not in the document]
ewoodev: Done

os/dbuild.sh
ewoodev: Thank you for review, I changed function name.

os/dutils.sh
ewoodev: Thank you, It applied in #5819

sunghan-chang (approved):

jylee9613 (approved):
LGTM

PR #5818, opened 2023-04-26
lib/libc/misc : Add new crc16 API 

jylee9613:
Current crc16 calculation is incorrect, but it cannot be modified because it is already used in many area.
So I add new CRC16 API (crc_ccitt) using CCITT polynomial (same as the table used in nuttx).


Taejun-Kwon:
Uh oh, actually file system uses crc16, so header in all sector already have a crc16  value.
so Let's assume that below case. 
1. Apply this PR
2. Header of each sector in Flash of all product include crc value
3. OTN already deployed product
4. Reboot & check validation of  each header.
5. It calculate different value with already written in header
6. so Mount will be failed forever.
So you should provide new API, should not overwrite current function for our product.

sunghan-chang (commented):
Let's add testcases in apps/examples/testcase for new apis.

sunghan-chang (approved):

PR #5820, opened 2023-04-27
mm/mm_heap: disable watchdog in mm_manage_alloc_fail function

ewoodev:
There are printing many logs in heapinfo_parse_heap() 
So during print log, Its reboots from the watchdog used by wifi, etc.

Therefore, This commit add to disable watchdog in mm_manage_alloc_fail()
that only when CONFIG_DEBUG_MM_HEAPINFO is enabled.

### DIFF
--- a/os/mm/mm_heap/mm_manage_allocfail.c
+++ b/os/mm/mm_heap/mm_manage_allocfail.c
@@ -23,10 +23,16 @@
 #include <tinyara/config.h>
 #if defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
 #include <stdio.h>
-#include <errno.h>
 #include <fcntl.h>
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/mminfo.h>
+#endif
+#if defined(CONFIG_DEBUG_MM_HEAPINFO) && defined(CONFIG_WATCHDOG)
+#include <tinyara/watchdog.h>
+#include <fcntl.h>
+#endif
+#if defined(CONFIG_DEBUG_MM_HEAPINFO) && defined(CONFIG_WATCHDOG) || defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
+#include <errno.h>
 #endif
 #include <debug.h>
 #include <string.h>
@@ -44,6 +50,55 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
+
+
+/************************************************************************
+ * Name: mm_heapinfo_stop_watchdog
+ *
+ * Description:
+ *   Stop the watchdog.
+ *   Forced stop or pause of the watchdog breaks the time guaranteed by the watchdog.
+ *   Therefore, if only watchdog is running and CONFIG_MM_ASSERT_ON_FAIL is enabled,
+ *   try to stop the watchdog.
+ *
+ * Return Value:
+ *   OK    - stopped watichdog or watchdog is not running.
+ *   ERROR - fail to stop watchdog
+ *
+ ************************************************************************/
+#if defined(CONFIG_DEBUG_MM_HEAPINFO) && defined(CONFIG_WATCHDOG)
+static inline int mm_heapinfo_stop_watchdog(void)
+{
+	struct watchdog_status_s wd_status;
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+		return ERROR;
+	}
+
+	if (ioctl(fd, WDIOC_GETSTATUS, (unsigned long)&wd_status) != OK) {
+		mfdbg("Fail to get watchdog state, errno %d\n", get_errno());
+		return ERROR;
+	}
+
+	if (!(wd_status.flags & WDFLAGS_ACTIVE)) {
+		/* watchdog is not running */
+		return OK;
+	}
+
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+	/* watchdog can be stopped only when CONFIG_MM_ASSERT_ON_FAIL is enabled */
+	if (ioctl(fd, WDIOC_STOP, 0) == OK) {
+		return OK;
+	}
+	mfdbg("Fail to stop watchdog, errno %d\n", get_errno());
+#endif
+
+	return ERROR;
+}
+#else
+#define mm_heapinfo_stop_watchdog() OK
+#endif
 
 /****************************************************************************
  * Public Functions
@@ -121,10 +176,24 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 #endif /* CONFIG_MM_ASSERT_ON_FAIL */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	for (int idx = startidx; idx <= endidx; idx++) {
-		heapinfo_parse_heap(&heap[idx], HEAPINFO_DETAIL_ALL, HEAPINFO_PID_ALL);
-	}
+	/* heapinfo_parse_heap() operations can print logs for more than 3 minutes,
+	 * which can be rebooted by watchdog. So try to stop the watchdog. */
+
+	int ret = mm_heapinfo_stop_watchdog();
+
+#ifndef CONFIG_MM_ASSERT_ON_FAIL
+	/* If CONFIG_MM_ASSERT_ON_FAIL is enabled, will be rebooted anyway, So even if
+	 * fail to stop watchdog, print heapinfo_parse_heap log. */
+
+	if (ret == OK)
 #endif
+	{
+		for (int idx = startidx; idx <= endidx; idx++) {
+			heapinfo_parse_heap(&heap[idx], HEAPINFO_DETAIL_ALL, HEAPINFO_PID_ALL);
+		}
+	}
+
+#endif /* CONFIG_DEBUG_MM_HEAPINFO */
 
 #ifdef CONFIG_MM_ASSERT_ON_FAIL
 	PANIC();


os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,30 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+			mfdbg("Fail to get watchdoc state, errno %d\n", get_errno());
+
+		/* if watchdoc is running, pause the watchdoc */
+		} else if (status.flags & WDFLAGS_ACTIVE && !(status.flags & WDFLAGS_WAIT)) {
+			if (ioctl(fd, WDIOC_PAUSE, 0) != OK) {
+				mfdbg("Fail to pause watchdoc, errno %d\n", get_errno());
+			} else {
+				/* only if watchdoc paused, the watchdoc must be resumed */
+				isPaused = true;
+				mfdbg("pause watchdoc\n");
```
kishore-sn, line 126: watchdoc -> watchdog

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,30 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s status;
+	bool isPaused = false;
```
kishore-sn, line 110: Before checking watchdog status, let us print a message like
`Temporarily pause active watchdog`
Otherwise the next messages do not make sense to user.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,30 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s status;
```
kishore-sn, line 109: We can rename all our variables like status -> wd_status, so that it indicates relation to wdog.

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Oops.... Thank you Done

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Thanks! added

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Thank you for review. applied

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+	struct watchdog_status_s wd_status;
+	bool is_wd_paused = false;
+	mfdbg("Temporarily pause active watchdog\n");
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+	} else {
+		/* success open watchdog, try to get watchdog state */
+		if (ioctl(fd, WDIOC_GETSTATUS, (unsigned long)&wd_status) != OK) {
+			mfdbg("Fail to get watchdog state, errno %d\n", get_errno());
```
jylee9613, line 118: It's minor, how about modifying the comment more clearly like this?
// If watchdog is successfully opened, try to get watchdog state.

ewoodev: thanks!.. but 'watchdog is successfully opened' could  be wrong...., because It is watchdoc driver.
It can be understood as a watchdoc function.
so Could i change `if successfully open`.

jylee9613: I understand. Then change to "if watchdog driver successfully opened,~"

ewoodev: yes, Thank you for review!~ chaged

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
```
abhishek-samsung, line 108: How about adding a check on ```CONFIG_WATCHDOG``` also? I think ```CONFIG_WATCHDOG_DEVPATH``` is not defined when watchdog is not enabled and this might cause some errors.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+	} else {
+		/* if watchdoc driver is successfully opened, try to get watchdog state */
+		if (ioctl(fd, WDIOC_GETSTATUS, (unsigned long)&wd_status) != OK) {
+			mfdbg("Fail to get watchdog state, errno %d\n", get_errno());
+
+		/* if watchdog is running, pause the watchdog */
+		} else if (wd_status.flags & WDFLAGS_ACTIVE && !(wd_status.flags & WDFLAGS_WAIT)) {
+			if (ioctl(fd, WDIOC_PAUSE, 0) != OK) {
+				mfdbg("Fail to pause watchdog, errno %d\n", get_errno());
```
abhishek-samsung, line 123: Is just printing an error message sufficient here? We might run into watchdog reset again in this case, right?

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Thank you for review. I think so. added

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s wd_status;
+	bool is_wd_paused = false;
+	mfdbg("Temporarily pause active watchdog\n");
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+	} else {
+		/* if watchdoc driver is successfully opened, try to get watchdog state */
```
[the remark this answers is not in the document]
ewoodev: wow... thank you!

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: yes, this case can watchdog reset.
but I think, in fail case is not a normal operation.

even if pause fail, we need to print heapinfo log as many as possible that is next code.
so, I didn't add any action

if must be paused watchdog in error operation, I can find way without watchdog driver.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s wd_status;
+	bool is_wd_paused = false;
+	mfdbg("Temporarily pause active watchdog\n");
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
```
Taejun-Kwon, line 113: please use below style to reduce if-ese state for more good readability.
```
if (!a) {
   mfdbg(...., get_errno());
   return;
}
if (!b) {
   mfdbg();
   return;
}

File: os/mm/mm_heap/mm_manage_allocfail.c [hunk 9]
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+			mfdbg("Fail to get watchdog state, errno %d\n", get_errno());
+
+		/* if watchdog is running, pause the watchdog */
+		} else if (wd_status.flags & WDFLAGS_ACTIVE && !(wd_status.flags & WDFLAGS_WAIT)) {
+			if (ioctl(fd, WDIOC_PAUSE, 0) != OK) {
+				mfdbg("Fail to pause watchdog, errno %d\n", get_errno());
+			} else {
+				/* only if watchdog paused, the watchdog must be resumed */
+				is_wd_paused = true;
+				mfdbg("pause watchdog\n");
```
Taejun-Kwon, line 127: Isn't there verbose message?
It seems that type of DEBUG_ERROR, it will be displayed almost all the time

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -23,11 +23,17 @@
...
-#include <errno.h>
 #include <fcntl.h>
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/mminfo.h>
 #endif
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+#include <tinyara/watchdog.h>
+#include <fcntl.h>
+#endif
+#if defined(CONFIG_DEBUG_MM_HEAPINFO) || defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)
```
sunghan-chang, line 34: I don't know why you check `defined(CONFIG_APP_BINARY_SEPARATION) && !defined(__KERNEL__)`. Could you let me know reason?

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s wd_status;
+	bool is_wd_paused = false;
+	mfdbg("Temporarily pause active watchdog\n");
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+	} else {
+		/* if watchdoc driver is successfully opened, try to get watchdog state */
+		if (ioctl(fd, WDIOC_GETSTATUS, (unsigned long)&wd_status) != OK) {
```
sunghan-chang, line 117: Do we need to check a status first? What happen if we call the pause without this?

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,31 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	struct watchdog_status_s wd_status;
+	bool is_wd_paused = false;
+	mfdbg("Temporarily pause active watchdog\n");
```
sunghan-chang, line 111: Not important but anyway it's not `temporarily`.

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: This is just to avoid duplicate include.

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: thanks, I changed to stop watchdog only when CONFIG_MM_ASSERT_ON_FAIL is enabled.

ewoodev:
Some changed code.
Forced stop or pause of the watchdog breaks the time guaranteed by the watchdog.

So,I changed to stop watchdog only when CONFIG_MM_ASSERT_ON_FAIL is enabled.
and if disable CONFIG_MM_ASSERT_ON_FAIL, do not print heapinfo_parse_heap() log.

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: yes, I changed to do not check.

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Thank you for review.
yes It is verbose message. I think so too.

but This time is after irqsave(), verbose message can't printed. becase block uart interrupt that used in uart driver.
so, I removed meesage.
Thank you!!~

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -99,6 +105,25 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
 #endif
 
+#if defined(CONFIG_DEBUG_MM_HEAPINFO)
+#if defined(CONFIG_WATCHDOG) && defined(CONFIG_MM_ASSERT_ON_FAIL)
```
sunghan-chang, line 109: I don't know those conditionals are necessary. Let's discuss it again.

| watchdog enabled | watchdog started | reboot enabled | disable watchdog | print heapinfo |
|---|---|---|---|---|
| X | X | X | X | O |
| O | X | X | X | O |
| O | O | X | X | X |
| O | O | O | O | O |
| X | X | O | X | O |

ewoodev: Thanks, I organized your table in more detail.
| watchdog enabled(config)	| reboot enabled(config) | watchdog started| watchdog stoped  | = print heapinfo|
|---|---|---|---|---|
|X|-|-|-|=O|
|O|X|X|-(must don't stop)|=O|
|O|X|O|-(must don't stop)|=X|
|O|O|X|-|=O|
|O|O|O|O|=O|

1. if watchdog is disabled, always print heapinfo. 
2. if reboot config is disabled, must not stop watchdog, but if watchdog is started, do not print heapinfo.
3. if reboot config is enabled, able to stop watchdog, print heapinfo
4. if watchdog is not started, always print heapinfo.

---
This is error case.
| watchdog enabled(config)| reboot enabled(config) | watchdog started| watchdog stoped  | = print heapinfo|
|---|---|---|---|---|
|O|X|?(error)|-(must don't stop)|=X|
|O|O|?(error)|O|=O|
|O|O|?(error)|X|=X|
|O|O|O|?(error)|=O|

1. if reboot config is disabled and we can't get watchdog state, don't print heapinfo.
2. if reboot conifg is enabled and wa can't get wtchdog state, try to stop watchdog and follow the result
3. if fail to stop watchdog, print heapinfo. below reason

* Some chipset(STM32) don't have function of stop watchdog.
try to stop watchdog, get a fail.
this case must be printed heapinfo.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -121,10 +146,16 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 #endif /* CONFIG_MM_ASSERT_ON_FAIL */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
+#if !defined(CONFIG_WATCHDOG) || defined(CONFIG_WATCHDOG) && defined(CONFIG_MM_ASSERT_ON_FAIL)
+	/* This can print logs for more than 3 minutes, which can be rebooted by watchdog.
+	 * To do this, there is an action to stop the watchdog, but it is only when CONFIG_MM_ASSERT_ON_FAIL
+	 * is enabled. Otherwise don't print heapinfo_parse_heap() log.
+	 */
 	for (int idx = startidx; idx <= endidx; idx++) {
 		heapinfo_parse_heap(&heap[idx], HEAPINFO_DETAIL_ALL, HEAPINFO_PID_ALL);
```
kishore-sn, line 155: heapinfo_parse_heap can be called from the application also. Even in that case, we can have a watchdog reset. 
So, how about we move the watchdog stop logic inside the heapinfo_parse_heap() ?

```
heapinfo_parse_heap
{
    # if watchdog
         pause watchdogs
    # endif

    print logs


    # if watchdog
        resume watchdog
    # endif
}
```

ewoodev: yes Thank you, I think this is good idea, 
but, if can stop the watch, we should know able to reboot. (if force stop watchdog, we must reboot.)
because forced stop or pause of the watchdog breaks the time guaranteed by the watchdog.
if logic is in heapinfo_parse_heap. we don't know able to reboot 

I think pause and resume is not right
Thank you for your review~

abhishek-samsung: > heapinfo_parse_heap can be called from the application also. Even in that case, we can have a watchdog reset.

Isn't the watchdog reset here happening because of disabling irq in mm_manage_alloc_fail? I think it might not be needed in this case.

abhishek-samsung:
Could you please mention in the commit message that we are also disabling irq during this time? Watchdog reboot happens because we are disabling irqs and printing logs for 3 min.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -121,10 +176,24 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
...
+	/* heapinfo_parse_heap() operations can print logs for more than 3 minutes,
+	 * which can be rebooted by watchdog. So try to stop the watchdog. */
+
+	int ret = mm_heapinfo_stop_watchdog();
+
+#ifndef CONFIG_MM_ASSERT_ON_FAIL
+	/* If CONFIG_MM_ASSERT_ON_FAIL is enabled, will be rebooted anyway, So even if
+	 * fail to stop watchdog, print heapinfo_parse_heap log. */
+
+	if (!ret)
```
sunghan-chang, line 188: If we use OK and ERROR, then this could be changed like below
```
int ret = mm_heapinfo_stop_watchdog();
if (ret == OK) {
   print heap info
}
```
We can understand like this
1. try to stop
2. if stopped, print heapinfo

I think `!` mean negative. I could understand like below
1. try to stop
2. if it's wrong, print heapinfo

abhishek-samsung:
As per my understanding, the current issue is not because of the watchdog but because of disabling irq in mm_alloc_fail function. So disabling watchdog might not be a good option. Instead, it's better to think about why the irqs are disabled. I think to prevent mixed logs, irqs are disabled here. So, instead of that, we can flush the serial driver first so that the mixed logs issue doesn't occur. I think TCFLSH + Schedlock might be able to do the job. Please let me know what you think.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -45,6 +51,55 @@
...
+		return 0;
+	}
+
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+	/* watchdog can be stopped only when CONFIG_MM_ASSERT_ON_FAIL is enabled */
+	if (ioctl(fd, WDIOC_STOP, 0) == OK) {
+		return 0;
+	}
+	mfdbg("Fail to stop watchdog, errno %d\n", get_errno());
+#endif
```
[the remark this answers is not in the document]
ewoodev: if CONFIG_MM_ASSERT_ON_FAIL is disabled, It is fail when try to stop watchdog..
So.. if CONFIG_MM_ASSERT_ON_FAIL is disabled, retrun fail(ERROR)..

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -45,6 +51,55 @@
...
+ *   Therefore, if only watchdog is running and CONFIG_MM_ASSERT_ON_FAIL is enabled,
+ *   try to stop the watchdog.
+ *
+ * Return Value:
+ *   0  - stopped watichdog or watchdog is not running.
+ *   -1 - fail to stop watchdog
+ *
+ ************************************************************************/
+#if defined(CONFIG_DEBUG_MM_HEAPINFO) && defined(CONFIG_WATCHDOG)
+static inline int mm_heapinfo_stop_watchdog(void) {
```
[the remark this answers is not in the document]
ewoodev: Thank you, fixed!

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -45,6 +51,55 @@
...
+
+	struct watchdog_status_s wd_status;
+	int fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDWR);
+	if (fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+		return -1;
```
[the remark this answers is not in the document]
ewoodev: Thanks fixed!

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: Thank you for detail review, changed

sunghan-chang (approved):

PR #5821, opened 2023-05-02
apps/shell : Allow 'reboot' command before TASH authentication

jylee9613:
Modify the reboot to operate regardless of TASH authentication. 
(To improve the accuracy of Hotfix download)

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -100,9 +100,6 @@ struct tash_cmd_info_s {
 static int tash_help(int argc, char **args);
 static int tash_clear(int argc, char **args);
 static int tash_exit(int argc, char **args);
-#ifdef CONFIG_TASH_REBOOT
-static int tash_reboot(int argc, char **argv);
-#endif
 #if TASH_MAX_STORE   > 0
 static int tash_history(int argc, char **argv);
 #endif
@@ -500,27 +497,6 @@ int check_exclam_cmd(char *buff)
 }
 #endif
 
-#ifdef CONFIG_TASH_REBOOT
-static int tash_reboot(int argc, char **argv)
-{
-	/*
-	 * Invoke the BOARDIOC_RESET board control to reset the board. If
-	 * the board_reset() function returns, then it was not possible to
-	 * reset the board due to some constraints.
-	 */
-#ifdef CONFIG_SYSTEM_REBOOT_REASON
-	WRITE_REBOOT_REASON(REBOOT_SYSTEM_USER_INTENDED);
-#endif
-	boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
-
-	/*
-	 * boarctl() will not return in this case.  It if does, it means that
-	 * there was a problem with the reset operaion.
-	 */
-	return ERROR;
-}
-#endif /* CONFIG_TASH_REBOOT */
-
 /** @brief Launch a task to run tash cmd asynchronously
  *  @ingroup tash
  */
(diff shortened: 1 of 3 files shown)


apps/shell/tash_security.c
```diff
@@ -77,6 +80,23 @@ void tash_check_security(int fd)
...
 		}
 
+		#ifdef CONFIG_TASH_REBOOT
+		if (!strncmp((unsigned char *)input, "reboot", 6)) {
+			/*
+			* Invoke the BOARDIOC_RESET board control to reset the board. If
+			* the board_reset() function returns, then it was not possible to
+			* reset the board due to some constraints.
+			*/
+			boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
```
sunghan-chang, line 90: Let's call the `tash_reboot`.

apps/shell/tash_security.c
```diff
@@ -77,6 +80,23 @@ void tash_check_security(int fd)
 			continue;
 		}
 
+		#ifdef CONFIG_TASH_REBOOT
+		if (!strncmp((unsigned char *)input, "reboot", 6)) {
```
sunghan-chang, line 84: Is `6` correct? or `7` correct?

apps/shell/tash_security.c
```diff
@@ -77,6 +80,23 @@ void tash_check_security(int fd)
...
+			* the board_reset() function returns, then it was not possible to
+			* reset the board due to some constraints.
+			*/
+			boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+
+			/*
+			* boarctl() will not return in this case.  It if does, it means that
+			* there was a problem with the reset operaion.
+			*/
+			return ERROR;
```
sunghan-chang, line 96: This function has void for return so that this should be `return;`, not `return ERROR`. or `break`.

sunghan-chang:
`implement` looks strange. How about `app/shell: allow the reboot command before TASH Password`?

apps/shell/tash_security.c
jylee9613: We don't need to compare the end of the string(\0), so 6 is correct.

apps/shell/tash_command.c
```diff
@@ -804,3 +780,24 @@ int tash_get_cmdpair(char *str, TASH_CMD_CALLBACK *cb, int index)
...
+#endif
+	boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+
+	/*
+	 * boarctl() will not return in this case.  It if does, it means that
+	 * there was a problem with the reset operaion.
+	 */
+	return ERROR;
+}
+#endif /* CONFIG_TASH_REBOOT */
```
sunghan-chang, line 803: Please check the warning

apps/shell/tash_security.c
sunghan-chang: But then, it will work with `rebootaa`. Is this ok?

jylee9613: oh, I didn't think of that part... I'll fix it.

apps/shell/tash_security.c
```diff
@@ -77,6 +80,14 @@ void tash_check_security(int fd)
 			continue;
 		}
 
+		/* Allow reboot command regardless of TASH authentication */
+		#ifdef CONFIG_TASH_REBOOT
+		if (!strncmp((unsigned char *)input, "reboot", 7)) {
+			int ret = tash_reboot(0, NULL);
```
sunghan-chang, line 86: `ret` is not used. How about using void instead of it?
```
(void)tash_reboot(0, NULL);
return ;
```

sunghan-chang (approved):

PR #5822, opened 2023-05-04
os/board/rtl8720e: Add buffer print IPC to avoid mixed log

ziyik:
-Added buffer print IPC to avoid KR4 log mixed with KM4
-KR4 log will be send to KM4 through IPC and print in KM4


sunghan-chang:
@ziyik Could you let us know belows?
1. Is that working with interrupt?
2. Is that working on an interrupt for data send and receive?
    1. If yes, maximum time?
3. The buffer is in KR4? or KM4? Does KM4 re-direct a log with entire sentense? I mean when KM4 gets line end, it sends the log into the uart?

ziyik:
> 

1. Yes, it used an IPC to transfer data from KR4. It's an interrupt.
2. What do you mean by maximum time?
3. the buffer is in KR4, each time KR4 is called a print it will convert the message used vsnprintf(0 to a buffer and send it to KM4 through IPC. So the whole print log is sent to KM4. If the function is enabled, it's default disabled.

sunghan-chang:
>> 2. What do you mean by maximum time?

If data send / receive is done in an ISR, it will block other interrupt at that time. So I want to know maximum time you block (elasped time of ISR).

sunghan-chang:
@ziyik We have no plan with this chip. Could you provide ETA for maximum elasped time of ISR?

ziyik:
> @ziyik We have no plan with this chip. Could you provide ETA for maximum elasped time of ISR? (Any time is ok for ETA)

I will measure the speed and maximum time needed for the IPC by 16th June.

ziyik:
> > > 2. What do you mean by maximum time?
> 
> If data send / receive is done in an ISR, it will block other interrupt at that time. So I want to know maximum time you block (elasped time of ISR).

The data size does not really affect the processing timer of IRQ.
Based on my test results, the maximum processing time of buffer print IPC IRQ is less than 15us.
The majority of the time needed is 5~8us.
The current implementation of buffer print will only transfer 255 characters per print, the exceeded characters will be dropped.
Each trigger of print in KR4 will trigger 1 time of IRQ in KM4.

sunghan-chang:
@ziyik Thank you for testing. Could you add what you left in the commit description? (not pr description)

ziyik:
> @ziyik Thank you for testing. Could you add what you left in the commit description? (not pr description)

Details information of IPC is added to the commit log.

sunghan-chang (approved):

PR #5823, opened 2023-05-08
apps/examples/testcase : Add kernel_tc test cases for CRC16-CCITT

jylee9613:
#commit 1.
apps/examples/testcase : Add kernel_tc test cases for CRC16-CCITT
Add test cases for CRC16-CCITT API

#commit 2.
lib/libc/misc : Change type of crc16 table
Change type of crc16 table from static to const


apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
```diff
@@ -168,6 +178,182 @@ static void tc_libc_misc_crc16part(void)
...
+	TC_ASSERT_EQ("crc16_ccitt_xmodem", ret_chk, VAL_11298);
+
+	/* Return value should be 5649 as calculated by crc16_ccitt_xmodem */
+
+	src_arr[0] = VAL_50;
+	ret_chk = crc16_ccitt_xmodem(src_arr, length);
+	TC_ASSERT_EQ("crc16_ccitt_xmodem", ret_chk, VAL_5649);
+
+	TC_SUCCESS_RESULT();
+}
```
sunghan-chang, line 208: How about below?
```
#define ARR_NUM 2
#define IN_VAL_100 100
#define IN_VAL_50 50
#define CRC_RES_OF_100 11298  // result value of crc16_ccitt_xmodem for input 100
#define CRC_RES_OF_50 5649 // result value of crc16_ccitt_xmodem for input 50

static void tc_libc_misc_crc16_ccitt_xmodem(void)
{
	uint16_t ret_chk;
	uint8_t src_arr[2] = { IN_VAL_100, IN_VAL_50 };
	uint8_t res_arr[2] = { CRC_RES_OF_100, CRC_RES_OF_50 };
	size_t length = 1;

	/* check crc16_ccitt_xmodem results according to input values */

	for (int arr_idx = 0; arr_idx < ARR_NUM; arr_idx++) {
		ret_chk = crc16_ccitt_xmodem(&src_arr[arr_idx], length);
		TC_ASSERT_EQ("crc16_ccitt_xmodem", ret_chk, res_arr[arr_idx]);
	}

	TC_SUCCESS_RESULT();
}
```

lib/libc/misc/lib_crc16.c
```diff
@@ -80,7 +80,7 @@
 
 /* crc16_tab calculated by Mark G. Mendel, Network Systems Corporation */
 
-static uint16_t crc16_tab[256] = {
+static const uint16_t crc16_tab[256] = {
```
sunghan-chang, line 83: This is not `add kernel tc for crc16-ccitt`. Please split the commit.

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
jylee9613: I think it's good idea to make array for src and res.
But it seems better to leave the definition variable as it is, since there are overlapping numbers among the result values.

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
```diff
@@ -168,6 +179,164 @@ static void tc_libc_misc_crc16part(void)
...
+static void tc_libc_misc_crc16_ccitt_xmodem(void)
+{
+	uint16_t ret_chk;
+	uint8_t src_arr[ARR_NUM] = { VAL_50, VAL_100 };
+	uint16_t res_arr[ARR_NUM] = { VAL_5649, VAL_11298 };
+	size_t length = 1;
+
+	/* Check crc16_ccitt_xmodem results according to input values */
+
+	for(int arr_idx = 0; arr_idx < ARR_NUM; arr_idx++) {
```
sunghan-chang, line 200: coding rule violation
```
for (
```

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
```diff
@@ -168,6 +179,164 @@ static void tc_libc_misc_crc16part(void)
...
+ * @Preconditions       :None
+ * @Postconditions      :None
+ * @Return              :void
+ */
+static void tc_libc_misc_crc16_ccitt_xmodem(void)
+{
+	uint16_t ret_chk;
+	uint8_t src_arr[ARR_NUM] = { VAL_50, VAL_100 };
+	uint16_t res_arr[ARR_NUM] = { VAL_5649, VAL_11298 };
+	size_t length = 1;
```
sunghan-chang, line 196: This is unchanged. It looks we don't need to use a variable for this. How about using just 1 at line 201 directly?

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
jylee9613: Oh I missed it.. Thank you for review.

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
jylee9613: It doesn't matter, but I think it would be better to use length as #define for readablility. Is it okay?

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
```diff
@@ -37,12 +37,24 @@
 
 #define BUFF_SIZE 256
 #define USEC_100 100
+#define ARR_NUM 2
+#define LENGTH 1
```
sunghan-chang, line 41: `LENGTH` is too common. Could you add which length at the definition name?

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
```diff
@@ -152,22 +161,173 @@ static void tc_libc_misc_crc16part(void)
...
+ * @API's covered       :crc16_ccitt_xmodem
+ * @Preconditions       :None
+ * @Postconditions      :None
+ * @Return              :void
+ */
+static void tc_libc_misc_crc16_ccitt_xmodem(void)
+{
+	uint16_t ret_chk;
+	uint8_t src_arr[ARR_NUM] = { VAL_50, VAL_100 };
+	uint16_t res_arr[ARR_NUM] = { VAL_5649, VAL_11298 };
```
sunghan-chang, line 192: The `res_arr` has VAL_XXX. But is this readable? Is 5649 value itself important? or Is result value of 50 hashed important?

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
amogh-samsung: Please also consider adding a short description of both macros in the comments please.

apps/examples/testcase/le_tc/kernel/tc_libc_misc.c
jylee9613: As you mentioned, VAL_XXX is not readable and the value is not important itself. I just followed the existing code, but if I have to change it, I think it's better to use raw number, not defined value.

sunghan-chang (approved):

PR #5825, opened 2023-05-10
rtl8721csm_download, rtl8720e_download: fix to download fail

ewoodev:
when download without bootparam, we can have below error. because It is not check, during file copy and try download.

so, This commit adds to check for bootparam.bin file before try copy or download.

Error log:
	cp: cannot stat '/root/tizenrt/build/configs/../../build/output/bin/bootparam.bin': No such file or directory
	...
	Makefile.unix:630: recipe for target 'download' failed
	make: *** [download] Error 1

Also This commit removes unnecessary checking.

### DIFF
--- a/build/configs/rtl8720e/rtl8720e_download.sh
+++ b/build/configs/rtl8720e/rtl8720e_download.sh
@@ -52,7 +52,9 @@ function pre_download()
 	if test -f "${SMARTFS_BIN_PATH}"; then
 		cp -p ${SMARTFS_BIN_PATH} ${IMG_TOOL_PATH}/${CONFIG_ARCH_BOARD}_smartfs.bin
 	fi
-	cp -p ${BIN_PATH}/bootparam.bin ${IMG_TOOL_PATH}/bootparam.bin
+	if test -f "${BIN_PATH}/${BOOTPARAM}.bin"; then
+		cp -p ${BIN_PATH}/${BOOTPARAM}.bin ${IMG_TOOL_PATH}/${BOOTPARAM}.bin
+	fi
 }
 
 function board_download()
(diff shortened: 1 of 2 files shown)


build/configs/rtl8720e/rtl8720e_download.sh
```diff
@@ -52,7 +52,9 @@ function pre_download()
 	if test -f "${SMARTFS_BIN_PATH}"; then
 		cp -p ${SMARTFS_BIN_PATH} ${IMG_TOOL_PATH}/${CONFIG_ARCH_BOARD}_smartfs.bin
 	fi
-	cp -p ${BIN_PATH}/bootparam.bin ${IMG_TOOL_PATH}/bootparam.bin
+	if test -f "${BIN_PATH}/${BOOTPARAM}.bin"; then
```
abhishek-samsung, line 55: How about checking if CONFIG_BP is enabled?
https://github.com/Samsung/TizenRT/blob/a93cd711656fecb66f98b34f9f251c53d1894a49/build/configs/rtl8721csm/hello/defconfig#L1308
If it is enabled, then bootparam.bin must be downloaded, so in that case, showing the error might be better.

ewoodev: Thanks!
if we don't have bootparam.bin or other file, printing error log, in below line.
so I think, don't need to add error log.
 
```
function board_download()
{
	cd ${IMG_TOOL_PATH}
	if [ ! -f ${IMG_TOOL_PATH}/$3 ];then
		echo "$3 not present"
	else
		./upload_image_tool_linux "download" $1 1 $2 $3
	fi
}
```


in error log
```
==========================
Downloading bootparam binary
==========================
bootparam.bin not present

Download COMPLETE!
Makefile.unix:614: recipe for target 'download' failed
make: *** [download] Error 1
```

abhishek-samsung: Oh okay.

sunghan-chang (approved):

PR #5828, opened 2023-05-19
apps/system/utils : Add TASH command to turn on/off the assert log

jylee9613:
Add new TASH command to turn on/off the assert log. 
The on/off value is stored in secure storage, and you can read the value through 'read_seclog'.

seclog_on : enable assert log
seclog_off : disable assert log


apps/system/utils/secure_log_cmd.c
```diff
@@ -0,0 +1,109 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <apps/shell/tash.h>
+#include <security/security_ss.h>
+
+/****************************************************************************
+ * Preprocessor Definitions
+ ****************************************************************************/
+#define LOG_SS_SLOT_PATH        "ss/32"
```
sunghan-chang, line 28: If we define this here, other module could overwrite it with another purpose. This should be located at se public header.

apps/system/utils/secure_log_cmd.c
```diff
@@ -0,0 +1,109 @@
...
+/********************************************************************************
+ * Private Function Prototypes
+ ********************************************************************************/
+
+static void secure_log_on(int argc, char **args);
+static void secure_log_off(int argc, char **args);
+
+const static tash_cmdlist_t secure_log_cmds[] = {
+	{"seclog_on",      secure_log_on,    TASH_EXECMD_SYNC},
+	{"seclog_off",     secure_log_off,   TASH_EXECMD_SYNC},
```
sunghan-chang, line 39: Is there any reason that the command should be separated for on and off?
How about providing a single command and giving an argument for on and off?
```
seclog on
or 
seclog off
```

os/kernel/debug/secure_log.c
```diff
@@ -0,0 +1,127 @@
...
+        output.priv_len = 0;
+        output.data_len = 2;
+        output.data = (char *)malloc(output.data_len);
+
+        if (SECLINK_OK != sl_read_storage(hnd, ss_idx, &output)) {
+                printf("[write_seclog] Failed to sl_read_storage\n");
+                return;
+        }
+
+        if (!strncmp(output.data, "1", 2)) {
```
sunghan-chang, line 99: Let's use readable definition instead of `"1"`.

os/kernel/debug/secure_log.c
```diff
@@ -0,0 +1,127 @@
...
+ *   None
+ *
+ * Return Value:
+ *   0(disable) or 1(enable)
+ *
+ * Assumptions:
+ *   None
+ *
+ ****************************************************************************/
+int read_seclog(void)
```
sunghan-chang, line 124: We will use this to determin log on / off. But I don't think we should name this as `seclog`.
How about `seclevel` or something else? If we use another name, then we can use this at other module when it needs.

os/kernel/init/os_bringup.c
```diff
@@ -262,6 +265,7 @@ static inline void os_do_appstart(void)
 
 #ifdef CONFIG_SE
 	se_initialize();
+	write_seclog();
```
sunghan-chang, line 268: Is this write? It looks strange.

os/kernel/debug/secure_log.h
```diff
@@ -0,0 +1,65 @@
...
+ *   This function write the value of secure log variable (which is saved in 
+ *   secure storage) in seclog_enable.
+ ****************************************************************************/
+int write_seclog(void);
+
+/****************************************************************************
+ * Name: read_seclog
+ *
+ * Description:
+ *   This function returns the seclog_enable.
```
sunghan-chang, line 61: I think this is not what user wants to know. 
```
Description:
   returns the secure level
       1 : secure level high? low?. Log will incude or exclude secure datas.
       0 : .... low? high? ....
```

os/kernel/debug/secure_log.h
```diff
@@ -0,0 +1,65 @@
...
+ ****************************************************************************/
+int read_seclog(void);
```
ewoodev, line 63: It is local header.
To use in other module, It should be loacted public header.

ewoodev, line 63: How about below?
```
#if CONFIG_XXXXXXXX
int read_seclog(void);
#else
#define read_seclog 1
#endif
```

in this, if check only CONFIG_SE, we have to alway check CONFIG_SE in apps to use this.

os/kernel/debug/Make.defs
```diff
@@ -33,5 +33,9 @@ ifeq ($(CONFIG_ENABLE_STACKMONITOR)$(CONFIG_DEBUG),yy)
 CSRCS += stackinfo_save_terminated.c
 endif
 
+ifeq ($(CONFIG_SE),y)
+CSRCS += secure_log.c
```
ewoodev, line 37: is it secure element's default operation?
I think, It is operation using secure element.
so How about add new config?

os/kernel/debug/secure_log.c
```diff
@@ -0,0 +1,128 @@
...
+
+        if (SECLINK_OK != sl_read_storage(hnd, ss_idx, &output)) {
+                printf("[set_security_level] Failed to sl_read_storage\n");
+                return;
+        }
+
+        if (!strncmp(output.data, SECLOG_ON, 2)) {
+                security_level = 1;
+        } else {
+                security_level = 0;
```
Taejun-Kwon, line 103: so you set value of security_level here, but how will you use this variable?

apps/system/utils/secure_log_cmd.c
```diff
@@ -0,0 +1,83 @@
...
+/****************************************************************************
+ * Preprocessor Definitions
+ ****************************************************************************/
+#define LOG_SS_SLOT_PATH        "ss/32"
+
+/********************************************************************************
+ * Private Function Prototypes
+ ********************************************************************************/
+
+static void secure_log(int argc, char **args)
```
Taejun-Kwon, line 34: set_secure_log is good for function name

apps/system/utils/secure_log_cmd.c
```diff
@@ -0,0 +1,83 @@
...
+	input.length = 1;
+
+        /* init */
+        if (0 != security_init(&hnd)) {
+		printf("Fail to init security handler\n");
+		return;
+	}
+
+        /* Write 'log on/off' variable to secure storage  */
+        if (0 != ss_write_secure_storage(hnd, LOG_SS_SLOT_PATH, 0, &input)) {
```
Taejun-Kwon, line 57: 1. in my opinion there's no reason divide security data to input & output
2. It should compare user input & changed value in ss level.
3. Below is my suggestion
```
int log_level;
security_data data = {NULL, 0};

if (!strncmp(args[1], "on", 3)) {
	data.data = "1";
	log_level = 1;
} else if (!strncmp(args[1], "off", 4)) {
	data.data = "0";
	log_level = 0;
} else {
	printf("Invalid argument\n");
	return;
}

...
ret = ss_write_secure_storage(... &data);
...
ret = ss_read_secure_storage(... &data);

if (log_level == 1) {
	if (!strncmp(output.data, "1", 2)) {
		printf("Assert log is enabled\n");
	} else {
		printf("error, set failed");
	}
} else {
	if (!strncmp(output.data, "0", 2)) {
		printf("Assert log is disabled\n");
	} else {
		printf("error, set failed");
	}
}
```

os/kernel/debug/secure_log.c
```diff
@@ -0,0 +1,128 @@
...
+ *   None
+ *
+ * Return Value:
+ *   0(disable) or 1(enable)
+ *
+ * Assumptions:
+ *   None
+ *
+ ****************************************************************************/
+int check_security_level(void)
```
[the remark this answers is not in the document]
jylee9613: It will be used to determine whether to print the assert log or not.
Before printing the assert log, we will check the security level with this function and decide whether to print or not.

os/kernel/debug/secure_log.c
```diff
@@ -0,0 +1,128 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private data
+ ****************************************************************************/
+static int security_level = 0;
```
ewoodev, line 51: how about name 'security_log_level' ?
It can change from only tash command, but tash command name is seclog. so it is for only security log.

apps/system/utils/secure_log_cmd.c
```diff
@@ -0,0 +1,83 @@
...
+ ****************************************************************************/
+#define LOG_SS_SLOT_PATH        "ss/32"
+
+/********************************************************************************
+ * Private Function Prototypes
+ ********************************************************************************/
+
+static void secure_log(int argc, char **args)
+{
+        security_handle hnd;
```
ewoodev, line 36: How about move setting ss storege code to os/kernel/debug/secure_log.c and just call setting api here?

ss read is called here and os/kernel/debug/secure_log.c.
so how about  call ss function in only one directory?

apps/system/utils/Kconfig
```diff
@@ -12,6 +12,20 @@ config SYSTEM_CMDS
...
 		Let's enable system commands at each configuration.
 
+config SECURITY_LEVEL_CMDS
+	bool "Security level commands"
+	default y
+        depends on TASH
+	select SECURITY_API
+	select SECURITY_SS
+	---help---
+		Disable setting security level command in TASH
```
sunghan-chang, line 22: Disable -> Enable

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,97 @@
...
+	/* Change security_level according to user input */
+	ss_data.length = 2;
+	if (!strncmp(args[1], "high", 5)) {
+		ss_data.data = "1";
+		security_level = 1;
+	} else if (!strncmp(args[1], "low", 4)) {
+		ss_data.data = "0";
+		security_level = 0;
+	} else {
+		printf("Invalid argument\n");
```
sunghan-chang, line 51: Please reference other commands. All commands provide the `help` option to show how to use. User of the command could be a non-developer so that Kconfig is not enough to check how to use. Even he or she is a developer, he or she could miss the Kconfig description.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,97 @@
...
+/********************************************************************************
+ * Private Function Prototypes
+ ********************************************************************************/
+
+static void tash_set_security_level(int argc, char **args)
+{
+	int security_level;
+	char ss_path[7] = { 0, };
+        security_handle hnd;
+	security_data ss_data = {NULL, 0};
```
sunghan-chang, line 39: Let's use the same policy for the coding style.
`{ NULL, 0 }` like line 37
or 
`{0,}` like line 39

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,97 @@
...
+			printf("security level : low\n");
+		} else {
+			printf("Failed to set security_level\n");
+		}
+	}
+}
+
+void security_level_register_utilcmds(void)
+{
+	tash_cmd_install("security_level", tash_set_security_level, TASH_EXECMD_SYNC);
```
sunghan-chang, line 96: This only supports `set` option, but I think `read` is necessary.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,97 @@
...
+		if (!strncmp(ss_data.data, "1", 2)) {
+			printf("security level : high\n");
+		} else {
+			printf("Failed to set security_level\n");
+		}
+	} else {
+		if (!strncmp(ss_data.data, "0", 2)) {
+			printf("security level : low\n");
+		} else {
+			printf("Failed to set security_level\n");
```
sunghan-chang, line 89: I don't know why this is necessary. You already check the return value of the write function. Even it returns OK, it could be unwritten?? That's why you read again to confirm? If yes, it's a serious problem. We never trust the API.

os/include/tinyara/ss_info.h
```diff
@@ -0,0 +1,23 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @brief ss_info contains the slot information used in secure storage.
+
+#include <stdio.h>
```
sunghan-chang, line 21: It looks unnecessary. In this file, nothing needs this header. Let's remove.

os/include/tinyara/ss_info.h
```diff
@@ -0,0 +1,23 @@
...
+
+#define SECURITY_LEVEL_SS_SLOT_INDEX     32
```
sunghan-chang, line 23: This is a security file, not security level file.
Let's re-name it with module name for prefix.
```
#define SS_SLOT_INDEX_SECURITY_LEVEL  (32)

File: os/include/tinyara/ss_info.h [hunk 22]
```diff
@@ -0,0 +1,23 @@
+/****************************************************************************
```
sunghan-chang, line 1: File name `ss_info` is too common.
I suggest `ss_slot_index.h`.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,128 @@
...
+
+        /* init */
+	ret = security_init(&hnd);
+        if (ret != SECURITY_OK) {
+		printf("[security_init] error ret : %d\n", ret);
+		return;
+	}
+
+	/* Convert secure storage slot index to string path */
+	snprintf(ss_path, 7, "ss/%d", SECURITY_LEVEL_SS_SLOT_INDEX);
```
sunghan-chang, line 77: The ss_path has fixed value always, "ss/32". Do you need to get it at runtime?
I think we can use a definition and it will reduce unnecessary operation.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,128 @@
...
+
+/********************************************************************************
+ * Private Functions
+ ********************************************************************************/
+static void tash_set_security_level(int argc, char **args);
+static void tash_get_security_level(int argc, char **args);
+
+const static tash_cmdlist_t security_level_cmds[] = {
+	{"set_sec_level", tash_set_security_level, TASH_EXECMD_SYNC},
+	{"get_sec_level", tash_get_security_level, TASH_EXECMD_SYNC},
```
sunghan-chang, line 34: Do you need to provide two separate command for get and set?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,128 @@
...
+		printf("[ss_read_secure_storage] error ret : %d\n", ret);
+		return;
+	}
+
+	if (!strncmp(ss_data.data, "1", 2)) {
+		printf("security level : high\n");
+	} else if (!strncmp(ss_data.data, "0", 2)) {
+		printf("security level : low\n");
+	} else {
+		printf("Failed to get security_level\n");
```
sunghan-chang, line 118: Let's add read value at error sentence. If you print just failure, we can't debug the issue.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,128 @@
...
+	{NULL, NULL, 0}
+};
+
+static void show_usage(void) 
+{
+	printf("\nUsage : set_sec_level [options]\n");
+	printf("Set your security level high:\n");
+	printf("	set_sec_level high\n");
+	printf("Set your security level low:\n");
+	printf("	set_sec_level low\n");
```
sunghan-chang, line 44: Let's reference Linux's command help sentence.

apps/system/utils/security_level_cmd.c
jylee9613: I add --help option in set_sec_level. Thank you for your review.

os/kernel/security_level/security_level.c
```diff
@@ -0,0 +1,127 @@
...
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: set_security_level
+ *
+ * Description:
+ *   This function read the level of security from secure storage and write it
+ *   in 'security_level' variable.
```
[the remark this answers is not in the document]
jylee9613: Okay. I added comment.

os/kernel/security_level/security_level.c
```diff
@@ -0,0 +1,127 @@
...
+
+        output.priv_len = 0;
+        output.priv = NULL;
+        output.data_len = 2;
+        output.data = (char *)kmm_malloc(output.data_len);
+
+        ret = g_se->ops->read_storage(SECURITY_LEVEL_SS_SLOT_INDEX, &output);
+        if (ret != HAL_SUCCESS) {
+                printf("[set_security_level] Error in read_storage. ret : %d\n", ret);
+                return;
```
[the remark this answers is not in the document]
jylee9613: Do you mean set_security_level should be return error type, instead of printing logs?
I think it is enough printing logs when it is failed.

os/include/tinyara/security_level.h
```diff
@@ -0,0 +1,69 @@
...
+ ****************************************************************************/
+void set_security_level(void);
+
+/****************************************************************************
+ * Name: get_security_level
+ *
+ * Description:
+ *   This function returns the security level
+ *      1 : High level of security. You can see all secure informations
+ *      0 : Low level of security. You cannot see some secure informations.
```
sunghan-chang, line 65: Usually high security level blocks secure logs and low secure level shows secure data.
How about exchanging them?
```
1: high level of security. You cannot see some secure information.
2: low level...... You can see all ....
```

os/include/tinyara/security_level.h
```diff
@@ -0,0 +1,69 @@
...
+
+/********************************************************************************
+ * Public Function Prototypes
+ ********************************************************************************/
+#if defined(CONFIG_SECURITY_LEVEL)
+/****************************************************************************
+ * Name: set_security_level
+ *
+ * Description:
+ *   This function write the level of security (which is saved in secure storage)
```
sunghan-chang, line 54: 1. typo, writes
2. You wrote `the level of security` and the `security_level` for different usages.
   But in `get_security_level`, it says `the level of security`. What is correct?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+        if (ret != SECURITY_OK) {
+		printf("[ss_write_secure_storage] error ret : %d\n", ret);
+		return;
+	} 
+	
+	printf("Security level is set\n");
+
+#ifdef CONFIG_SECURITY_LEVEL
+	/* Apply changed value to the kerenl variable */
+	ret = prctl(PR_SET_SECURITY_LEVEL, getpid(), NULL);
```
sunghan-chang, line 109: Why do you need `getpid`?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+	if (ret != OK) {
+		printf("Fail to launch set_security_level. ret : %d\n", ret);
```
sunghan-chang, line 111: 1. launch? call?
2. As I left before, let's use the same format for the logs. You add `[ ]` at every fail logs. But not here.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+
+	/* Change security_level according to user input */
+	ss_data.length = 2;
+	if (!strncmp(args[1], "high", 5)) {
+		ss_data.data = "1";
+	} else if (!strncmp(args[1], "low", 4)) {
+		ss_data.data = "0";
+	} else {
+		printf("Invalid argument.\n");
+		printf("Use [--help] option if you want to know how to use.\n");
```
sunghan-chang, line 94: Just call `show_usage` here instead of saying --help option.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	/* Change security_level according to user input */
+	ss_data.length = 2;
+	if (!strncmp(args[1], "high", 5)) {
+		ss_data.data = "1";
+	} else if (!strncmp(args[1], "low", 4)) {
+		ss_data.data = "0";
+	} else {
+		printf("Invalid argument.\n");
```
sunghan-chang, line 93: Let's print `args[1]` at the logs.
```
printf("Invalid argument (%s).\n" args[1]);
```

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+	/* Apply changed value to the kerenl variable */
+	ret = prctl(PR_SET_SECURITY_LEVEL, getpid(), NULL);
+	if (ret != OK) {
+		printf("Fail to launch set_security_level. ret : %d\n", ret);
+		return;
+	}
+
+	/* Check the kerenl variable to confirm whether the value is correct */
+	ret = prctl(PR_GET_SECURITY_LEVEL, getpid(), NULL);
+	printf("[kerenl][get_security_level] ret : %d\n", ret);
```
sunghan-chang, line 117: Is this necessary message for user?

os/include/tinyara/ss_slot_index.h
```diff
@@ -0,0 +1,22 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/// @brief ss_info contains the slot information used in secure storage.
+
+#define SS_SLOT_INDEX_SECURITY_LEVEL    (32)
+#define SS_SLOT_PATH_SECURITY_LEVEL     "ss/32"
```
sunghan-chang, line 22: 1. This is used in the specific file. Do you need to add this here? I think you can move this in .c file.
2. Oh, if we don't call sprintf to get `"ss/32"`, we can't use `SS_SLOT_INDEX_SECURITY_LEVEL`. Then, I think we could use sprintf. Sorry for opposite opinion.

os/kernel/security_level/security_level.c
```diff
@@ -0,0 +1,130 @@
...
+ * Return Value:
+ *   None
+ *
+ * Assumptions:
+ *   None
+ *
+ ****************************************************************************/
+void set_security_level(void)
+{
+        struct sec_lowerhalf_s *g_se = se_get_device();
```
sunghan-chang, line 84: It's not a global variable. Let's remove prefix, `g_`.

os/kernel/Kconfig
```diff
@@ -1031,3 +1031,8 @@ config LOG_DUMP_NUMBUFS
 	default 2
 
 endif # LOG_DUMP
+
+config SECURITY_LEVEL
+	bool "Enable setting security level"
+	default y
+	select SE
```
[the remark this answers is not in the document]
sunghan-chang: This is not modified and no reply. Could you let me know your opinion?

apps/system/utils/security_level_cmd.c
sunghan-chang: Currently there is no `read` or `get` option. Please let me know your opinion.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+
+#ifdef CONFIG_SECURITY_LEVEL
+	/* Apply changed value to the kerenl variable */
+	ret = prctl(PR_SET_SECURITY_LEVEL, getpid(), NULL);
+	if (ret != OK) {
+		printf("Fail to launch set_security_level. ret : %d\n", ret);
+		return;
+	}
+
+	/* Check the kerenl variable to confirm whether the value is correct */
```
amogh-samsung, line 115: Could you please change "kerenl" to "kernel"

apps/system/utils/security_level_cmd.c
amogh-samsung, line 117: Similar grammatical error here in "kerenl"

os/kernel/security_level/security_level.c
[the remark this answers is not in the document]
amogh-samsung: The caller of get/set_security level may benefit from knowing why these functions have failed.
Also, this is kind of standard practice where the lower level functions return the ERROR code to the caller to report the cause of failure.
So you could consider having these functions return integer types.

apps/system/utils/Kconfig
```diff
@@ -12,6 +12,20 @@ config SYSTEM_CMDS
 		system commands. Because they can be used without TASH by function calling.
 		Let's enable system commands at each configuration.
 
+config SECURITY_LEVEL_CMDS
+	bool "Security level commands"
+	default y
+        depends on TASH
```
ewoodev, line 18: This is being turned on by CONFIG_TASH. It is not TASH default operation.
So If we should set dependency about TASH, How about `depends on TASH && SECURITY_LEVEL`?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	/* Show usage */
+	if (!strncmp(args[1], "--help", 7)) {
+		show_usage();
+		return;
+	}
+
+	/* Init security handler */
+	ret = security_init(&hnd);
```
ewoodev, line 60: The codes of about seurity_level setting and reading is better to be gathered in one place.

How about move the this code to os/kernel/security_level and just all here?

os/include/tinyara/security_level.h
jylee9613: I was also thinking about it... I agree that we have to follow the standard. I'll exchange it.

os/include/tinyara/security_level.h
jylee9613: I missed the `_` in get_security_level. I'll add it.

apps/system/utils/security_level_cmd.c
jylee9613: I left this for me because of debugging... I'll erase this.

os/kernel/security_level/security_level.c
```diff
@@ -0,0 +1,130 @@
...
+                printf("[set_security_level] Error in read_storage. ret : %d\n", ret);
+                return;
+        }
+        
+        if (!strncmp(output.data, HIGH_SECURITY_LEVEL, 2)) {
+                security_level = 1;
+        } else if (!strncmp(output.data, LOW_SECURITY_LEVEL, 2)) {
+                security_level = 0;
+        } else {
+                printf("[set_security_level] : Wrong value is written in security_level. ret : %s\n", output.data);
```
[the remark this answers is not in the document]
jylee9613: If it is corrupted, we cannot read secure storage, and that case is checked in line 94~98.
Is there any other corrupted case besides this one?

apps/system/utils/Kconfig
jylee9613: I'll add it. Thanks.

apps/system/utils/security_level_cmd.c
jylee9613: Please confirm this. Is that your opinion?

1. When user change value in TASH, security_level_cmd.c change the `security_level` variable, which is defined in security_level.c
2. In security_level.c, write `security_level` variable in secure storage.

apps/system/utils/security_level_cmd.c
jylee9613: I think we need to read the variable for users, so I added 'status' option to read the variable.

apps/system/utils/Kconfig
```diff
@@ -12,6 +12,21 @@ config SYSTEM_CMDS
 		system commands. Because they can be used without TASH by function calling.
 		Let's enable system commands at each configuration.
 
+config SECURITY_LEVEL_CMDS
+	bool "Security level commands"
+	default y
+        depends on TASH && SECURITY_LEVEL
+	select SECURITY_API
+	select SECURITY_SS
```
sunghan-chang, line 20: 1. Let's use the same way for indentation.  You use spaces in line 18, but use a tab for other lines.
2. When do you use `depends on` and `select`?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,131 @@
...
+
+static void tash_security_level(int argc, char **args)
+{
+        security_handle hnd;
+	security_data ss_data = {NULL, 0};
+	char ss_path[7];
+	security_error ret;
+
+	if (argc < 2) {
+		printf("Use [--help] option if you want to know how to use.\n");
```
sunghan-chang, line 50: As I left before, let's call the `show_usage` function here instead of this log.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,131 @@
...
+	/* Show usage */
+	if (!strncmp(args[1], "--help", 7)) {
+		show_usage();
+		return;
+	}
+
+	/* Init security handler */
+	ret = security_init(&hnd);
+        if (ret != SECURITY_OK) {
+		printf("[security_init] error ret : %d\n", ret);
```
sunghan-chang, line 63: This function is operation of user input. So, if we meet some failure case, we should reply failure of command execution first. And then we should add error type for development.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,131 @@
...
+			return;
+		}
+
+		if (!strncmp(ss_data.data, "1", 2)) {
+			printf("security level : high\n");
+		} else if (!strncmp(ss_data.data, "0", 2)) {
+			printf("security level : low\n");
+		} else {
+			printf("Wrong value is written in security_level.\n");
+			printf("[ss_read_secure_storage] security_level data : %s\n", ss_data.data);
```
sunghan-chang, line 84: how about below for the message?
```
Security Level : Unexpected value!
Please set the security level using \"sec_level high\" or \"sec_level low\".
```

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,131 @@
...
+	/* Apply changed value to the kernel variable */
+	ret = prctl(PR_SET_SECURITY_LEVEL);
+	if (ret != 0) {
+		printf("[prctl] Fail to call set_security_level. ret : %d\n", ret);
+		return;
+	}
+
+	/* Check the kernel variable to confirm whether the value is correct */
+	//ret = prctl(PR_GET_SECURITY_LEVEL);
+	//printf("[kernel][get_security_level] ret : %d\n", ret);
```
sunghan-chang, line 121: Do you need commented codes?

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,131 @@
...
+        /* Write 'security_level' variable to secure storage  */
+	ret = ss_write_secure_storage(hnd, ss_path, 0, &ss_data);
+        if (ret != SECURITY_OK) {
+		printf("[ss_write_secure_storage] error ret : %d\n", ret);
+		return;
+	} 
+	
+	printf("Security level is set\n");
+
+#ifdef CONFIG_SECURITY_LEVEL
```
sunghan-chang, line 111: You already make a dependency with it. In other words, there is no case not defined CONFIG_SECURITY_LEVEL. This `#ifdef` is unnecessary.

os/include/tinyara/security_level.h
```diff
@@ -0,0 +1,74 @@
...
+#define LOW_SECURITY_LEVEL      "0"
+
+/********************************************************************************
+ * Global Type Declarations
+ ********************************************************************************/
+typedef enum {
+        SECURITY_LEVEL_OK,
+        SECURITY_LEVEL_FAIL_READ,
+        SECURITY_LEVEL_INVALID_VALUE
+} sec_level_result;
```
sunghan-chang, line 37: This is not our style.
```
enum security_level_result_e {
    ....
};
typedef enum security_level_result_e  security_level_result_t;
```

os/kernel/Kconfig
```diff
@@ -1031,3 +1031,9 @@ config LOG_DUMP_NUMBUFS
 	default 2
 
 endif # LOG_DUMP
+
+config SECURITY_LEVEL
```
sunghan-chang, line 1035: You still don't reply my  comment.
We have the se module. Do we need to add this in kernel module??

os/kernel/init/os_bringup.c
```diff
@@ -264,6 +267,13 @@ static inline void os_do_appstart(void)
 	se_initialize();
 #endif
 
+#ifdef CONFIG_SECURITY_LEVEL
+	int ret = set_security_level();
+	if (ret != SECURITY_LEVEL_OK) {
+		dbg("[set_security_level] Failed to call set_security_level. ret : %d\n", ret);
```
sunghan-chang, line 273: There are two `set_security_level` messages. Let's use once to save unnecessary flash usages and to optimize performance.

os/kernel/security_level/security_level.c
```diff
@@ -0,0 +1,135 @@
...
+        if (hal_ret != HAL_SUCCESS) {
+                printf("[read_storage] Error in read_storage. hal_ret : %d\n", hal_ret);
+                ret = SECURITY_LEVEL_FAIL_READ;
+                return ret;
+        }
+        
+        if (!strncmp(output.data, HIGH_SECURITY_LEVEL, 2)) {
+                security_level = 1;
+        } else if (!strncmp(output.data, LOW_SECURITY_LEVEL, 2)) {
+                security_level = 0;
```
sunghan-chang, line 104: How about below?
```
1. run atoi
2. check whether it's 1 or 0
3. 2nd is right, save
```
Then we don't need to do strncmp twice.

apps/system/utils/security_level_cmd.c
ewoodev: yes I think this

os/kernel/Kconfig
jylee9613: Sorry for the late reply.
At first, I added this in kerenl module because I thought this was about assert log on/off.
However, since the variable has been changed to security_level now, it would be right to move to SE module as you said.
Thank you for review.

apps/system/utils/security_level_cmd.c
jylee9613: To change the code according to your opinion, we must create another API to access `security_level` or define `security_level` as a global variable. It means that other apps can also change this variable.
However, `security_level` will only be used in kernel, and there is no reason to access it from other apps.
So I think the current condition is a better way.

apps/system/utils/security_level_cmd.c
```diff
@@ -0,0 +1,128 @@
...
+			return;
+		}
+
+		if (!strncmp(ss_data.data, "1", 2)) {
+			printf("security level : high\n");
+		} else if (!strncmp(ss_data.data, "0", 2)) {
+			printf("security level : low\n");
+		} else {
+			printf("security level : Unexpected value!!!\n");
+			printf("Please set the security level using \"sec_level high\" or \"sec_level high\".\n");
```
sunghan-chang, line 86: `\"sec_level high\" or \"sec_level high\"`
Please check typo.

apps/system/utils/security_level_cmd.c
ewoodev: I agree that a new API should be added. However, security_level does not need to be a global variable.

Is security_level.c an API for reading and writing only security_level variables?

The security_level variable is changed to the secure_storage value.
However, the code that actually write secure_storage value does not exist in security_level.c.
We do not know who set the secure_storage to set the security_level to on/off in security_level.c.

but It is just my opinion.

PR #5829, opened 2023-05-22
apps/mem_leak_checker, kernel/debug: change usage for mem leak checker

abhishek-samsung:
Consolidate app and kernel leak check under mem_leak

```
TASH>>mem_leak

Kernel :
*** NO MEMORY LEAK.

Below are text addresses of loadable apps (and common binary if enabled) :
The pc value of the allocation can be obtained by subtracting the text start address of the appropriate binary

[app1] Text Addr : 0x217d040, Text Size : 8064
[app2] Text Addr : 0x21fe700, Text Size : 366624

app1 :
*** NO MEMORY LEAK.
app2 :
*** NO MEMORY LEAK.
TASH>>
```

### DIFF
--- a/os/kernel/debug/mem_leak_checker.c
+++ b/os/kernel/debug/mem_leak_checker.c
@@ -304,16 +304,6 @@ static void print_info(char *bin_name, int leak_cnt, int broken_cnt, uint32_t bi
 	volatile struct mm_allocnode_s *node;
 	uint32_t owner_addr;	
 
-#ifdef CONFIG_APP_BINARY_SEPARATION
-        bin_addr_info_t *bin_addr_info = (bin_addr_info_t *)get_bin_addr_list();
-	int bin_idx;
-	for (bin_idx = 0; bin_idx <= CONFIG_NUM_APPS; bin_idx++) {
-                if (bin_addr_info[bin_idx].text_addr != 0) {
-                        printf("[%s] Text Addr : %p, Text Size : %u\n", BIN_NAME(bin_idx), bin_addr_info[bin_idx].text_addr, bin_addr_info[bin_idx].text_size);
-                }
-        }
-#endif
-
 	if (leak_cnt > 0 || broken_cnt > 0) {
 		printf("Type   |    Addr    | Size(byte) |    Owner   | PID \n");
 		printf("---------------------------------------------------\n");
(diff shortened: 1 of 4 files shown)


os/kernel/debug/mem_leak_checker.c
```diff
@@ -414,3 +414,19 @@ int run_mem_leak_checker(int checker_pid, char *bin_name)
 
 	return OK;
 }
+
+void run_all_mem_leak_checker(int checker_pid, char *bin_name)
```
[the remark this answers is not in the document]
sunghan-chang: @abhishek-samsung Could you check @ewoodev 's comment?

os/kernel/debug/mem_leak_checker.c
```diff
@@ -414,3 +414,19 @@ int run_mem_leak_checker(int checker_pid, char *bin_name)
...
+{
+	printf("Kernel :\n");
+	run_mem_leak_checker(checker_pid, "kernel");
+#ifdef CONFIG_APP_BINARY_SEPARATION
+	bin_addr_info_t *bin_addr_info = (bin_addr_info_t *)get_bin_addr_list();
+	int bin_idx;
+	for (bin_idx = 1; bin_idx <= CONFIG_NUM_APPS; bin_idx++) {
```
Taejun-Kwon, line 425: it started from 1, and you execute run_mem_leak_checker with "kernel" as a 2nd parameter.
so I guess 0 is always kernel.
Then can't we modify loop to start bin_idx from 0?

abhishek-samsung: 0 index is reserved for common binary. So we cant use it for kernel.

os/kernel/debug/mem_leak_checker.c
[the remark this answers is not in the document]
abhishek-samsung: Done. I have also modified the output slightly and added some descriptions. Could you please check it and let me know if it's okay?

```
TASH>>mem_leak

Kernel :
*** NO MEMORY LEAK.

Below are text addresses of loadable apps (and common binary if enabled) :
The pc value of the allocation can be obtained by subtracting the text start address of the appropriate binary

[app1] Text Addr : 0x217d040, Text Size : 8064
[app2] Text Addr : 0x21fe700, Text Size : 366624

app1 :
*** NO MEMORY LEAK.
app2 :
*** NO MEMORY LEAK.
TASH>>
```

os/kernel/debug/mem_leak_checker.c
Taejun-Kwon: One question, then it seems that it doesn't include kernel, because here it doesn't check kernel.
0 is for common binary, then it never check common binary? or is it unncessary?

abhishek-samsung: Yes, we will not check common binary. Common binary is code that will be used by apps. So any allocation done from app through common binary code will still come under the app's allocation. So, it's not required.

sunghan-chang: Could you leave some code comment why we check it from index 1?

abhishek-samsung: I have added a comment for that in the file. I am attaching the code for reference here.

https://github.com/Samsung/TizenRT/blob/2b9459d794b105c766350da270aa4e5b9e1820d6/os/kernel/binary_manager/binary_manager_data.c#L64-L66

os/kernel/debug/mem_leak_checker.c
```diff
@@ -414,3 +404,40 @@ int run_mem_leak_checker(int checker_pid, char *bin_name)
 
 	return OK;
 }
+
+int run_all_mem_leak_checker(int checker_pid)
```
sunghan-chang, line 408: Could you check whether the `checker_pid` is necessary or not?

jylee9613 (approved):
Looks good to me!

ewoodev (approved):
LGTM~

os/kernel/debug/mem_leak_checker.c
abhishek-samsung: Yes, we need to skip checking the stack region of the checker, as it may contain the address that we are checking. So we need the pid to get the stack limits and exclude it during the checking.

sunghan-chang (approved):

PR #5835, opened 2023-05-25
Add support and test cases for binary data type preference

amogh-samsung:
########## Preference UTC End [PASS : 58, FAIL : 0] ##########

Added 8 new testcases for (shared/private, get/set preference)

### DIFF
--- a/apps/examples/testcase/ta_tc/preference/utc/utc_preference_main.c
+++ b/apps/examples/testcase/ta_tc/preference/utc/utc_preference_main.c
@@ -27,26 +27,38 @@
 
 #include "tc_common.h"
 
-#define INT_KEY "INTEGER"
-#define DOUBLE_KEY "DOUBLE"
-#define BOOL_KEY "BOOL"
-#define STRING_KEY "STRING"
+#define INT_KEY		"INTEGER"
+#define DOUBLE_KEY	"DOUBLE"
+#define BOOL_KEY	"BOOL"
+#define STRING_KEY	"STRING"
+#define BINARY_KEY	"BINARY"
 
-#define SHARED_KEY_PATH "utc/preference"
-#define SHARED_INTKEY_PATH "utc/preference/INTEGER"
-#define SHARED_DOUBLEKEY_PATH "utc/preference/DOUBLE"
-#define SHARED_BOOLKEY_PATH "utc/preference/BOOL"
-#define SHARED_STRINGKEY_PATH "utc/preference/STRING"
+#define SHARED_KEY_PATH		"utc/preference"
+#define SHARED_INTKEY_PATH	"utc/preference/INTEGER"
+#define SHARED_DOUBLEKEY_PATH	"utc/preference/DOUBLE"
+#define SHARED_BOOLKEY_PATH	"utc/preference/BOOL"
+#define SHARED_STRINGKEY_PATH	"utc/preference/STRING"
+#define SHARED_BINARYKEY_PATH	"utc/preference/BINARY"
 
-#define INT_VALUE 300
-#define DOUBLE_VALUE 4.21
-#define BOOL_VALUE true
-#define STRING_VALUE "preference string value"
+#define INT_VALUE	300
+#define DOUBLE_VALUE	4.21
+#define BOOL_VALUE	true
+#define STRING_VALUE	"preference string value"
 
-#define INVALID_KEY "INVALIDKEY"
-#define CB_DATA "PREFERENCE_CBDATA"
+#define INVALID_KEY	"INVALIDKEY"
+#define CB_DATA		"PREFERENCE_CBDATA"
+
+struct bin_pref_val_s {
+	int val1;
+	char val2;
+} BINARY_VALUE;
 
 static bool cb_flag;
+
+bool check_binary_value(struct bin_pref_val_s *v, struct bin_pref_val_s b)
+{
+	return ((v->val1 == b.val1) && (v->val2 == b.val2));
+}
 
 static preference_changed_cb value_changed_cb(const char *key, void *user_data)
 {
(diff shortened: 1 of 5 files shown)


amogh-samsung:
> @amogh-samsung Could you let me know why this is necessary?

Hi @sunghan-chang ,
This was requested was made by a parallel-working team
@r-prabu could you please elaborate?

r-prabu:
> > @amogh-samsung Could you let me know why this is necessary?
> 
> Hi @sunghan-chang , This was requested was made by a parallel-working team @r-prabu could you please elaborate?

> Hi @sunghan-chang , Since Application team is integrating matter stack in which they need to store the binary type key-value pair.

framework/src/preference/shared_preference.c
```diff
@@ -198,6 +217,28 @@ int preference_shared_get_string(const char *key, char **value)
 	return ret;
 }
 
+int preference_shared_get_binary(const char *key, char **value)
```
r-prabu, line 220: It would be good to provide size for the binary type. So that API user's can do the operations accordingly.

apps/examples/testcase/ta_tc/preference/utc/utc_preference_main.c
```diff
@@ -269,6 +311,45 @@ static void utc_preference_get_string_n(void)
...
+	TC_ASSERT_EQ_CLEANUP("preference_get_binary", value, NULL, free(value));
+	TC_ASSERT_EQ("preference_get_binary", ret, PREFERENCE_INVALID_PARAMETER);
+
+	ret = preference_get_binary(INT_KEY, &value);
+	TC_ASSERT_EQ_CLEANUP("preference_get_binary", value, NULL, free(value));
+	TC_ASSERT_EQ("preference_get_binary", ret, PREFERENCE_INVALID_PARAMETER);
+
+	ret = preference_get_binary(INVALID_KEY, &value);
+	TC_ASSERT_EQ_CLEANUP("preference_get_binary", value, NULL, free(value));
+	TC_ASSERT_EQ("preference_get_binary", ret, PREFERENCE_KEY_NOT_EXIST);
```
jeongchanKim, line 348: I think it is better to check the line 348 first, and check the value as line 347 after.

framework/src/preference/private_preference.c
```diff
@@ -104,6 +104,25 @@ int preference_set_string(const char *key, char *value)
 	return prctl(PR_SET_PREFERENCE, &data);
 }
 
+//For setting preference as binary, we need the length of the binary data that is to be stored
```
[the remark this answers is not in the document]
amogh-samsung: Done

framework/src/preference/shared_preference.c
amogh-samsung: Done

apps/examples/testcase/ta_tc/preference/utc/utc_preference_main.c
amogh-samsung: Pending

framework/src/preference/private_preference.c
```diff
@@ -198,6 +217,29 @@ int preference_get_string(const char *key, char **value)
...
+	}
+
+	data.key = key;
+	data.type = PRIVATE_PREFERENCE;
+	data.attr.type = PREFERENCE_TYPE_BINARY;
+
+	/* Get preference with prctl */
+	ret = prctl(PR_GET_PREFERENCE, &data);
+	if (ret == OK) {
+		*value = (void *)data.value;
```
sunghan-chang, line 236: Don't you need to check `*value` is null?

framework/src/preference/private_preference.c
```diff
@@ -198,6 +217,29 @@ int preference_get_string(const char *key, char **value)
...
+		*len = (int *)data.attr.len;
```
sunghan-chang, line 237: How about len for null?

os/include/tinyara/preference.h
```diff
@@ -44,6 +44,7 @@ enum preference_type_e {
 	PREFERENCE_TYPE_DOUBLE,
 	PREFERENCE_TYPE_BOOL,
 	PREFERENCE_TYPE_STRING,
+	PREFERENCE_TYPE_BINARY,
```
sunghan-chang, line 47: It looks it's used in the framework only. Do we need to maintain here, not in the framework include?

amogh-samsung: I agree, the entire enumeration can be moved to framework specific include path.
However, it should be done in a separate commit.

framework/src/preference/private_preference.c
amogh-samsung: NULL returned values should be checked as general code practice, so I understand your point.
However, I have 2 considerations here: -
1. The writing of preference value is accompanied by checksum, so a successful write return is a good estimator of confirmation. Also, in `get_preference`, `read_fs_key` function involves CRC verification of the read preference. I think these 2 are enough for safeguarding us. Any additional logical checks of the key-value should be in the hands of the use-case developer.
2. No other `get_preference` API checks for NULL values. All should be updated together for consistency.

Hence, I feel current version is alright.

framework/src/preference/private_preference.c
amogh-samsung: I feel that NULL values need not be checked as per my reply to the previous comment.
Kindly let me know.

ritesh55555 (commented):
test cases for binary data type preference added looks good

framework/src/preference/private_preference.c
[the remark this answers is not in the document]
amogh-samsung: Fair point, understood, thankyou.
Will make appropriate changes.

framework/src/preference/shared_preference.c
```diff
@@ -104,6 +104,25 @@ int preference_shared_set_string(const char *key, char *value)
 	return prctl(PR_SET_PREFERENCE, &data);
 }
 
+// We need length of the binary data to be stored
```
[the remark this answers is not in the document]
amogh-samsung: In accordance with previous comment, will make appropriate changes.

framework/src/preference/private_preference.c
```diff
@@ -198,6 +217,29 @@ int preference_get_string(const char *key, char **value)
 	return ret;
 }
 
+int preference_get_binary(const char *key, void **value, int *len)
+{
+	int ret;
+	preference_data_t data;
+
+	if (key == NULL) {
```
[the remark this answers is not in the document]
amogh-samsung: Understood, thankyou.
@sunghan-chang seems to have suggested the same, I will make changes.

framework/src/preference/private_preference.c
amogh-samsung: Update: - while int, bool and double values are updated in the variable passed by reference and the space from the kernel side is freed,
In the case of string and binary preferences, only a pointer is passed from the application side to retrieve the address of the variable holding the preference data/value. It is initialized as NULL and updated with the address of the variable holding the preference data. In this case, the space from the kernel side is not freed and it is still pointed to by the pointer sent from the application call to preference API.
Hence I feel it is OK to initialize string and binary preference value pointers as NULL, and they should not be checked for NULL values, because it is intended.

the `len` pointer should however be checked for NULL values as it directly retrieves the value stored and does not simply point to it.

@kishore-sn @sunghan-chang

amogh-samsung: Updated as per suggestions from yu and @kishore-sn

sunghan-chang (approved):

PR #5836, opened 2023-05-26
mm_heap/mm_initialize.c: set total_alloc_size to include first and last node

abhishek-samsung:
total_alloc_size is being initialized to zero and is being updated only during mm related calls (malloc, free etc). So, it will not include the size of the gaurd nodes. If we check mallinfo implementation, the allocated size includes the size of guard nodes. So, initialize it to proper value.

before fix :
[tc_umm_heap_get_heap_free_size] FAIL [Line : 485] umm_get_heap_free_size : Values (free_size == 0x736c0) and (info.fordblks == 0x736a0) are not equal

after fix :
[tc_umm_heap_get_heap_free_size] PASS
.
.
.########## Kernel TC End [PASS : 415, FAIL : 0] ##########

### DIFF
--- a/os/mm/mm_heap/mm_initialize.c
+++ b/os/mm/mm_heap/mm_initialize.c
@@ -191,6 +191,14 @@ int mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsiz
 	heap->mm_nregions++;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	/* add guard nodes size to the newly added region as they are considered as allocated*/
+	heap->total_alloc_size += 2 * SIZEOF_MM_ALLOCNODE;
+	if (heap->peak_alloc_size < heap->total_alloc_size) {
+		heap->peak_alloc_size = heap->total_alloc_size;
+	}
+#endif
+
 	/* Add the single, large free node to the nodelist */
 
 	mm_addfreechunk(heap, node);
@@ -253,6 +261,10 @@ int mm_initialize(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
 
 	mm_seminitialize(heap);
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	heap->total_alloc_size = heap->peak_alloc_size = 0;
+#endif
+
 	/* Add the initial region of memory to the heap */
 
 	ret = mm_addregion(heap, heapstart, heapsize);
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@abhishek-samsung I understand the issue. But do we need to modify the heapinfo instead of the mallinfo? The guards are not user's heap usages.

sunghan-chang:
If the `mm_heapsize` in the `struct mm_heap_s` includes them, let's merge current.

os/mm/mm_heap/mm_initialize.c
```diff
@@ -263,7 +263,8 @@ int mm_initialize(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
 	for (i = 0; i < CONFIG_MAX_TASKS; i++) {
 		heap->alloc_list[i].pid = HEAPINFO_INIT_INFO;
 	}
-	heap->total_alloc_size = heap->peak_alloc_size = 0;
+	/* two gaurd nodes are always allocated for the start and end of the heap */
+	heap->total_alloc_size = heap->peak_alloc_size = 2 * SIZEOF_MM_ALLOCNODE;
```
ewoodev, line 267: we can have multiple heap regions
Could you check if it should be `heap->peak_alloc_size = 2 * SIZEOF_MM_ALLOCNODE * CONFIG_KMM_REGIONS`

```
struct mm_heap_s {
        ... 
	FAR struct mm_allocnode_s *mm_heapstart[CONFIG_KMM_REGIONS];
	FAR struct mm_allocnode_s *mm_heapend[CONFIG_KMM_REGIONS];
        ....
```

abhishek-samsung: Thanks for pointing it out. However also have USER heap along with KERNEL heap. So, to support them both, I have moved the logic to mm_addregion. Please review.

abhishek-samsung:
> If the `mm_heapsize` in the `struct mm_heap_s` includes them, let's merge current.

Yes, it includes them. In mallinfo we include the guard nodes size in uordblks.

https://github.com/Samsung/TizenRT/blob/d7343f3f49adb663c794436d3e619ff12e4dc6dd/os/mm/mm_heap/mm_mallinfo.c#L135

I have printed the values for user heap and verified them.
```
info->fordblks : 00074f60, info->uordblks : 00007c80, info->arena : 0007cbe0
fordblks + uordblks = 0x00074f60 + 0x00007c80 = 0x7CBE0 = mm_heapsize
```

os/mm/mm_heap/mm_initialize.c
```diff
@@ -191,6 +191,14 @@ int mm_addregion(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsiz
 	heap->mm_nregions++;
 #endif
 
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+	/* add garud nodes size to the newly added region as they are considered as allocated*/
```
sunghan-chang, line 195: typo, guard

sunghan-chang (approved):

ewoodev (approved):

Tirtha-Samsung (approved):

jylee9613 (approved):

PR #5850, opened 2023-06-15
arch/armv7-a, amebasmart: Use CONFIG_FLASH_START_ADDR

Tirtha-Samsung:
In rtl8730e and armv7-a architecture, we use CONFIG_FLASH_START instead of CONFIG_FLASH_START_ADDR


sunghan-chang (changes requested):
Sorry but I don't agree this change. I think we should use CONFIG_FLASH_START_ADDR instead of CONFIG_FLASH_START. There is no reason to use it for only armv7-a.

sunghan-chang:
@Tirtha-Samsung Please let me know reason of `In rtl8730e and armv7-a architecture, we use CONFIG_FLASH_START instead of CONFIG_FLASH_START_ADDR`.

os/Makefile.unix
```diff
@@ -222,7 +222,11 @@ endif
 ifeq ($(CONFIG_RAW_BINARY),y)
 BIN_EXT = bin
 endif
+ifeq ($(CONFIG_ARCH_ARMV7A_FAMILY),y)
+KERNEL_START=$(CONFIG_FLASH_START)
```
ewoodev, line 226: Why we should use two config `CONFIG_FLASH_START` and `CONFIG_FLASH_START_ADDR`?
Is the address value of the two confg different?

If the two address values are the same, I think we can do as below.
```
#define CONFIG_FLASH_START CONFIG_FLASH_START_ADDR
```
or change to use CONFIG_FLASH_START_ADDR in armv7-a

sunghan-chang (approved):

sunghan-chang:
1. `arch/arm/src/armv7-a, arch/arm/src/amebasmart: Use CONFIG_FLASH_START_ADDR as per TizenRT`
is too long. How about below?
`arch/armv7-a, amebasmart: user CONFIG_...._ADDR`
2. There is no commit description. This is not allowed. Please write add some description.

ewoodev (approved):
LGTM~

sunghan-chang (approved):

PR #5852, opened 2023-06-19
arch/amebad, board/rtl8721csm: set-up uart pins in up_serialinit

ewoodev:
This commit changes time of serial pin initialize.

### DIFF
--- a/os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
+++ b/os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
@@ -502,6 +502,14 @@ void serial_init(serial_t *obj, PinName tx, PinName rx)
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


os/board/rtl8721csm/src/component/common/mbed/targets/hal/rtl8721d/serial_api.c
```diff
@@ -502,6 +502,13 @@ void serial_init(serial_t *obj, PinName tx, PinName rx)
 #endif
 }
 
+void serial_pin_init(PinName tx, PinName rx) {
```
sunghan-chang, line 505: Please keep the same coding style in this file.
Other function does
```
void xxxx(yyyy)
{
```

ewoodev: Thanks Done!~

sunghan-chang (approved):

PR #5856, opened 2023-06-22
os/arch/amebasmart: Update peripherals

zhongnuo-tang:
Add watchdog, rtl8730e does not support watchdog stop. 
Add timer, tested ok with tc


build/configs/rtl8730e/Make.defs
```diff
@@ -219,7 +219,7 @@ endef
 SECURE_HEADER_SIZE=32
 
 define MAKE_SAMSUNG_HEADER
-  $(TOPDIR)/../os/tools/mksamsungheader.py kernel ${SECURE_HEADER_SIZE}
+  $(TOPDIR)/../os/tools/mksamsungheader.py $(1) kernel ${SECURE_HEADER_SIZE}
```
sunghan-chang, line 222: @ewoodev Could you confirm this is enough to resolve the build break?
@zhongnuo-tang This is not `update peripherals`. Could you split the commit for this?

os/arch/arm/src/amebasmart/amebasmart_watchdog_lowerhalf.c
```diff
@@ -0,0 +1,485 @@
...
+
+/****************************************************************************
+ * Name: amebasmart_wdg_initialize
+ *
+ * Description:
+ *   Initialize and register the amebasmart watchdog. Watchdog timer is registered
+ *   as 'devpath'.
+ *
+ ****************************************************************************/
+int amebasmart_wdg_initialize(const char *devpath, uint32_t timeout_ms)
```
sunghan-chang, line 433: Who calls this and timer init?

zhongnuo-tang: > Who calls this and timer init?

Dear Mr Chang,
wdg init and timer init will be called in rtl8730e_boot.c if CONFIG_WATCHDOG or CONFIG_TIMER is enable.

build/configs/rtl8730e/Make.defs
zhongnuo-tang: Dear @sunghan-chang,

Okay, I will split the commit after @ewoodev has confirmed that it can fix the build break.

zhongnuo-tang: Dear @ewoodev,
Should i commit this fix in this PR or Samsung will fix this?

Dear @sunghan-chang,
If I would need to commit this fix, should it be another commit in this same PR or raise another PR for this fix?

Thank you

sunghan-chang: @zhongnuo-tang It should be a different commit but both is ok for a PR.

sunghan-chang (approved):

PR #5857, opened 2023-06-22
build/tools/amebasmart: fix loguart hanging issue

fengyunlai:
Fix loguart hanging issue, in the case that copying pasting strings to tera term console will cause hang in previous version.
It is caused by 3 cores' loguart rx handling.
Disable km0 and km4's loguart rx irq, only tizenrt's uart serial console should handle the inputs from tera term.
This modification can fix the issue and it has been verified.


sunghan-chang (commented):
The title is difficult to know what board has the issue and is fixed.
How about `build/tools/amebasmart: fix loguart hanging issue`?
And add some description about the issue on the commit description.

fengyunlai:
> The title is difficult to know what board has the issue and is fixed. How about `build/tools/amebasmart: fix loguart hanging issue`? And add some description about the issue on the commit description.

Hi @sunghan-chang , thank you for your reminder, I changed the title and added more descriptions.

sunghan-chang:
https://github.com/Samsung/TizenRT/blob/master/tools/uart_hw_test/README.md

As you know, we have some specs for uart operation. Could you test amebasmart using above and let us know result?

fengyunlai:
> https://github.com/Samsung/TizenRT/blob/master/tools/uart_hw_test/README.md
> 
> As you know, we have some specs for uart operation. Could you test amebasmart using above and let us know result?

Hi @sunghan-chang , I commited another commit to fixing uart irq handler inputs not match issue.
Now I can pass all UART itc tests.
I will check the uart_hw_test and UART utc tests next week. 
But I suggest we merge this PR first, so that other developers can avoid these issues. Thank you!

sunghan-chang (approved):

PR #5860, opened 2023-06-26
Loadable Apps bring up for RTK8730e

kishore-sn:
Bring up code for loadable apps on RTK8730e chipset. This commit does not include memory protection for loadable apps. Since Armv7a uses MMU instead of MPU, we will make new implementation for memory protection using MMU.


lib/libc/machine/arm/armv7-a/arch_elf.c
```diff
@@ -0,0 +1,499 @@
...
+
+  return OK;
+}
+
+int up_relocateadd(const Elf32_Rela *rel, const Elf32_Sym *sym,
+                   uintptr_t addr)
+{
+  berr("ERROR: RELA relocation not supported\n");
+  return -ENOSYS;
+}
```
sunghan-chang, line 499: Let's follow our coding rules.

lib/libc/machine/arm/armv7-a/arch_elf.c
```diff
@@ -0,0 +1,499 @@
...
+      return -EINVAL;
+    }
+
+  return OK;
+}
+
+int up_relocateadd(const Elf32_Rela *rel, const Elf32_Sym *sym,
+                   uintptr_t addr)
+{
+  berr("ERROR: RELA relocation not supported\n");
```
sunghan-chang, line 497: Do we have `berr`, `binfo` and so on?

os/include/sys/syscall_lookup.h
```diff
@@ -0,0 +1,406 @@
+/****************************************************************************
```
abhishek-samsung, line 1: Is this file needed? We have syscall.csv which is used to generate similar mappings.

abhishek-samsung: Looks like we already have the file,
https://github.com/Samsung/TizenRT/blob/master/os/syscall/syscall_lookup.h

lib/libc/machine/arm/armv7-a/arch_elf.c
kishore-sn: Done.

os/syscall/syscall_names.c
```diff
@@ -0,0 +1,105 @@
...
+/****************************************************************************
+ * Public Data
+ ****************************************************************************/
+
+/* System call name tables.  This table is indexed by the system call numbers
+ * defined above.  Given the system call number, this table provides the
+ * name of the corresponding system function.
+ */
+
+const char *g_funcnames[SYS_nsyscalls] =
```
abhishek-samsung, line 97: I think this file is also present under a different name,

https://github.com/Samsung/TizenRT/blob/ebdd02ca6b5b34e535d696492a0f5ef6732c9c4c/os/syscall/syscall_funclookup.c#L115-L122

could you please check if the new file is necessary?

kishore-sn: `g_funclookup` returns the function address and `g_funcnames` returns the function name as a string. So, both are required. However, there was a mistake in the macro expansion to return the function name. I have fixed it. Thanks.

abhishek-samsung (approved):

build/configs/rtl8730e/Make.defs
```diff
@@ -121,6 +121,9 @@ ifeq ($(CONFIG_FRAME_POINTER),y)
 endif
 
 ARCHCFLAGS += -fno-builtin 
+ifeq ($(CONFIG_APP_BINARY_SEPARATION),y)
```
Tirtha-Samsung, line 124: The PR, Loadable Apps bring up for RTK8730e, looks good to me.

sunghan-chang (approved):

PR #5861, opened 2023-06-26
os/arch/amebasmart: add GPIO

zhongnuo-tang:


os/arch/arm/src/amebasmart/Make.defs
```diff
@@ -231,7 +231,7 @@ CHIP_CSRCS += amebasmart_serial.c
 CHIP_CSRCS += amebasmart_timerisr.c
 CHIP_CSRCS += amebasmart_enet.c
 CHIP_CSRCS += amebasmart_flash.c
-#CHIP_CSRCS += amebasmart_lowerhalf.c
+CHIP_CSRCS += amebasmart_lowerhalf.c
```
sunghan-chang, line 234: How about naming `amebasmart_gpio_lowerhalf.c`? Because there are several lowerhalf files.

os/arch/arm/src/amebasmart/amebasmart_lowerhalf.c
sunghan-chang, line 1: Unnecessary file. Let's remove.

sunghan-chang (approved):

PR #5865, opened 2023-07-04
tinyara/arch.h: Add dummy TZ APIs for ARMV7A

kishore-sn:
ARMV7A does not use NSC (Non Secure Callable) approach for calling secure APIs from non secure side. As a result APIs to allocate and free secure context are no longer required. However, RTK vendor code for SE tries to call these APIs, which is causing build error. This commit provides dummy defines for these APIs to prevent the errors.

### DIFF
--- a/os/include/tinyara/arch.h
+++ b/os/include/tinyara/arch.h
@@ -445,6 +445,16 @@ void up_allocate_secure_context(TZ_ModuleId_t size);
  ****************************************************************************/
 
 void up_free_secure_context();
+
+#else		//CONFIG_ARMV8M_TRUSTZONE
+
+// Adding below defines to prevent errors when vendor code tries to call
+// security API's which are not implemented as they are not required by the 
+// architecture. For example ARMV7A does not use NSC based TZ calls.
+
+#define up_allocate_secure_context(size)
+#define up_free_secure_context()
+
 #endif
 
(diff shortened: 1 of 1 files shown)


os/include/tinyara/arch.h
```diff
@@ -445,6 +445,16 @@ void up_allocate_secure_context(TZ_ModuleId_t size);
...
 void up_free_secure_context();
+
+#else		//CONFIG_ARMV8M_TRUSTZONE
+
+// Adding below defines to prevent errors when vendor code tries to call
+// security API's which are not implemented as they are not required by the 
+// architecture. For example ARMV7A does not use NSC based TZ calls.
+
+#define up_allocate_secure_context(size);
+#define up_free_secure_context();
```
sunghan-chang, line 456: Let's remove `;`.

os/include/tinyara/arch.h
```diff
@@ -445,6 +445,16 @@ void up_allocate_secure_context(TZ_ModuleId_t size);
  ****************************************************************************/
 
 void up_free_secure_context();
```
sunghan-chang, line 447: Could you add `void` at argument with different commit? It's the TizenRT coding rule.

kishore-sn: Done.. Thanks.

PR #5868, opened 2023-07-06
os/arch/arm/src/amebasmart: add sys_reset api

edwakuwaku:
Add amebasmart_systemreset.c for loadable_apps
Rename bsp file name into debug_printf.h, to prevent TizenRT debug.h being shadowed


sunghan-chang (commented):
This commit has three different kinds of changes.
1. add sys_reset api
2. change board specific header name
3. enable board reset by default

Could you split them as different commits?

sunghan-chang (approved):

PR #5875, opened 2023-07-10
preference: Move PREFERENCE_TYPE macros to framewrk/../preference.h

amogh-samsung:
- PREFERENCE_TYPE macros are only used by the preference framework code so they need not be defined in kernel/../preference.h header file

### DIFF
--- a/framework/include/preference/preference.h
+++ b/framework/include/preference/preference.h
@@ -27,12 +27,29 @@
 #ifndef __FRAMEWORK_INCLUDE_PREFERENCE_PREFERENCE_H__
 #define __FRAMEWORK_INCLUDE_PREFERENCE_PREFERENCE_H__
 
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
 #include <stdbool.h>
 #include <tinyara/preference.h>
 
-#ifdef __cplusplus
+/****************************************************************************
+ * Pre-processor Definitions
+ ****************************************************************************/
+#ifdef __cpluplus
 extern "C" {
 #endif
+
+/****************************************************************************
+ * Public Type Definitions
+ ****************************************************************************/
+enum preference_type_e {
+        PREFERENCE_TYPE_INT = 0,
+        PREFERENCE_TYPE_DOUBLE,
+        PREFERENCE_TYPE_BOOL,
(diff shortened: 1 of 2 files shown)


framework/include/preference/preference.h
```diff
@@ -27,10 +27,24 @@
...
 
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
 #include <stdbool.h>
 #include <tinyara/preference.h>
 
-#ifdef __cplusplus
+/****************************************************************************
+ * Pre-processor Definitions
```
kishore-sn, line 37: Enum does not come under preprocessor defines. 
Please change it to `Public Types `

os/include/tinyara/preference.h
```diff
@@ -39,14 +39,6 @@
 #define PREF_PRIVATE_PATH PREF_PATH"/private"
 #define PREF_SHARED_PATH PREF_PATH"/shared"
 
-enum preference_type_e {
-	PREFERENCE_TYPE_INT = 0,
-	PREFERENCE_TYPE_DOUBLE,
-	PREFERENCE_TYPE_BOOL,
-	PREFERENCE_TYPE_STRING,
-	PREFERENCE_TYPE_BINARY,
-};
```
sunghan-chang, line 48: Sorry.. I asked you for this change. But now, I think we should keep the same location for every definition. If we keep them at several headers, it could be complex.
(Originally I though we could move entire definitions into the framework. But it's not)

amogh-samsung: We cannot move all definitions to framework, that is true.
Should we close this PR then?

sunghan-chang: How do you think? We have two options.
1. Some definitions are used only in the framework so that they should be defined in the header of the framework.
2. Other almost definitions are in the header of the kernel so that keeping in the kernel header is better to know preference definitions.

amogh-samsung: I think it is OK to have the definition in the module/folder where it is used.
I mean, if we can shift PREFERENCE_TYPEs to framework, but the rest have to stay in kernel, that should also be OK.

sunghan-chang (approved):

PR #5877, opened 2023-07-12
os/board/rtl8730e: Update Flash Layout

zhongnuo-tang:
- New flash layout as follows:
- km4_boot_all.bin  0x08000000    0x08010000
- kernel_rtl8730e_200204.trpk 0x08080000    0x0837F000
- kernel_rtl8730e_200204.trpk 0x0837F000    0x0867E000
- bootparam.bin   0x087FE000    0x08800000


build/configs/rtl8730e/hello/defconfig
```diff
@@ -281,9 +281,9 @@ CONFIG_ARCH_BOARD_HAVE_FLASH=y
 #
 CONFIG_FLASH_PARTITION=y
 CONFIG_FLASH_MINOR=0
-CONFIG_FLASH_PART_SIZE="128,3260,3260,512,1024,8,"
-CONFIG_FLASH_PART_TYPE="none,kernel,kernel,smartfs,ftl,bootparam,"
-CONFIG_FLASH_PART_NAME="bl1,kernel,kernel,userfs,reserved,bootparam,"
+CONFIG_FLASH_PART_SIZE="48,52,12,400,3132,3132,512,896,8,"
```
sunghan-chang, line 284: I need 1024 for last reserved. It's impossible 896.

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -115,6 +115,5 @@ menu "Realtek RTL8730E TrustZone Support"
 config AMEBASMART_TRUSTZONE
 	bool "Enable Amebasmart TrustZone"
 	default n
-	select ARMV8M_TRUSTZONE
```
sunghan-chang, line 118: This is a different topic. Please split a commit.

os/se/ameba/rtl_se_crypto_function.h
```diff
@@ -65,7 +65,7 @@
 #define SS_BASE_ADDRESS 0x14000
 #endif
 #ifdef CONFIG_AMEBASMART_TRUSTZONE
-#define SS_BASE_ADDRESS 0x1F0000 /* ToDo: Temporary set need changes based flash addr */
+#define SS_BASE_ADDRESS 0x1C000
```
sunghan-chang, line 68: We should get this from CONFIG FLASH PART SIZE. Bcause we could move the location. (This is why we use the config.)

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -53,8 +53,8 @@ static uint32_t SLOT1_START_ADDR = 0x27000;
 #endif
 #ifdef CONFIG_AMEBASMART_TRUSTZONE
 /* ToDo: Temporary set need changes based flash addr */
-static uint32_t SLOT0_START_ADDR = 0x200000;
-static uint32_t SLOT1_START_ADDR = 0x203000;
+static uint32_t SLOT0_START_ADDR = 0x2C000;
```
sunghan-chang, line 56: Could you let me know what they are?

ZhenBei-Sin: Hi @sunghan-chang , 

This is the secure storage slot address used for the secure storage test code. 
May refer to the below image:

image

Thank You.
Best Regards,
Zhen Bei

sunghan-chang: For security, we don't use real address for this. Do you need to use real address? How about just using slot index with security api?

os/se/ameba/rtl_se_crypto_function.h
sunghan-chang: This is ok to keep current.

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
ZhenBei-Sin: Hi @sunghan-chang,

We can make it with 2 options 
1. Add a helper function in Non-Secure
2. Add an API in Secure call via NSC / SMCCC. 
Two options also need to change on TP1x, TP1e, and TPDual.
We need time to implement it and please let us know you prefer on which option.

We suggest that you may proceed with merging this PR 1st.
If have a concern about the above change, we may remove the changes 1st.
Please take note this SStorage example is unable to enable and cannot be tested once removed the SLOT address.

Thank You.

zhongnuo-tang:
> @zhongnuo-tang `- Remove AMEBASMART_TRUSTZONE tie with ARMV8M_TRUSTZONE in Kconfig` does not follow the TizenRT commit rules. How about below?
> 
> ```
> arch/amebasmart: remove ARMV8M_TRUSTZONE dependency
> 
> blah blah (description)
> ```

Dear @sunghan-chang,
Will make the relevant changes

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
sunghan-chang: @zhongnuo-tang There are read / write APIs for secure storage in the SE module (os/se). I don't know what you need to implement for them. Could you let us know?

ziyik: There are 2 types of read / write APIs for secure storage.
One type is NSC, which will pass the RAW data into write API, it will encrypt it and write encrypted data with a specific format into Flash. NSC read API will read the encrypted data, decrypts it and return RAW data.
One type is non-secure API, which is used to read RAW data from Flash.
NSC API required slot index.
Non-secure API required the exact address of the slot.

These are requested example, the purpose is to verify the data in Flash is not RAW data.

sunghan-chang: @ziyik Is this an example of non-secure API?
```
One type is non-secure API, which is used to read RAW data from Flash.
NSC API required slot index.
Non-secure API required the exact address of the slot.
```
That's why it has the exact address of the slot?
For encryption verfication?

ziyik: Yes, the exact address of the slot is used by a non-secure API.
rtl_ss_flash_read() is a non-secure API to read the RAW data from Flash, so it required the exact address.
ss_read_secure_storage() and ss_write_secure_storage() are NSC API, which links to secure API, they only required a slot number to read or write.

In this secure storage example, it first reads with NSC API to check the slot data.
Then, it reads the RAW data from the actual Flash location of the data (areas A and B).
Follow by writing data to secure storage with NSC API by slot number.
Next, read with ss_read_secure_storage() to check the slot data is correctly written.
Lastly, reads the RAW data from the actual Flash location of the data (areas A and B).

Compare the data read by both APIs, data read by a non-secure API should be encrypted, it should be different from NSC API. This proves that data in Flash is encrypted.

sunghan-chang (approved):

PR #5879, opened 2023-07-13
os/board/rtl8730e: Add checks for validity of specified wifi channels

jwei5:
- Add checks in rtk_netmgr to check if channel is a valid 2.4ghz/5ghz channel when scanning specific channel or starting softap


j11591 (dismissed):
Please check when channel is 0. It means whole channel scan

jwei5:
Hi, may I check which is the current case for whole scan with channel 0? For scanning with SSID and channel = 0, code will enter condition at line 457 in rtk_netmgr.c to scan whole channel for SSID

j11591 (approved):
LGTM

PR #5885, opened 2023-07-21
build/tools/amebasmart/bootloader: update flash ChipInfo for 16M

zhongnuo-tang:


sunghan-chang:
Is this available for both 8M and 16M?

sunghan-chang (approved):

PR #5886, opened 2023-07-25
build/tools/amebasmart/bootloader: update hotfix

zhongnuo-tang:
- add error log when handshake is not detected
- add reset after hotfix is done correctly
- add Samsung hotfix hash for PBKDF2


sunghan-chang:
@zhongnuo-tang Could you print the bootloader version like RTL8721CSM?

zhongnuo-tang:
> @zhongnuo-tang Could you print the bootloader version like RTL8721CSM?

Dear @sunghan-chang,
Yes I can implement it, can i raise a separate PR after adding bootloader version?

sunghan-chang:
@zhongnuo-tang For history, let's do that with new PR.

PR #5890, opened 2023-08-01
kernel/log_dump: refactor log_dump

zambbo:
log_dump still have a lot of issues.
So, I rewrite the log_dump module in order to increase the readability and memory efficiency.
The biggest difference is that the thread writing log should wait for the compression thread when a buffer is full, but the rewritten module did not wait until the entire buffer is full.
In addition, the spare buffer is used so that some of the logs can be saved when they have not been saved before.


ewoodev (commented):
Please add commit comment.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -69,6 +69,9 @@ int log_dump_main(int argc, char *argv[])
 	}
 
 	printf("This Log Should be saved!!!\n");
+	for(int i=0; i<1000; i++) {
+		printf("This is long log.\n");
+	}
```
ewoodev, line 74: It is not about log_dump refactoring.
Please separate commit.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -69,6 +69,9 @@ int log_dump_main(int argc, char *argv[])
 	}
 
 	printf("This Log Should be saved!!!\n");
+	for(int i=0; i<1000; i++) {
```
jeongchanKim, line 72: Based on TizenRT Coding Rule,
please add space properly.

os/kernel/log_dump/log_dump_initialize.c
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	comp_chunk_append(node);
+
+	comp_queue.size_of_chunks = LOG_CHUNK_SIZE;
+	SAVE_START();
+
+	return LOG_DUMP_OK;
+
+	exit_with_node:
```
jeongchanKim, line 140: Please remove the tab for `goto` statement.

os/kernel/log_dump/log_dump_manager.c
```diff
@@ -0,0 +1,89 @@
...
+		if (compressed_size == LOG_DUMP_COMPRESS_FAIL) {
+			lldbg("Fail to compress");
+		}
+		else {
+			log_dump_append((const char*)out_buf, compressed_size);
+		}
+
+		UNCOMP_HEAD.curbytes = 0;
+		
+		size_t free_size = log_dump_get_free_size();
```
jeongchanKim, line 81: In this case, we don't need to declare local variable, because it is only used one-time.
If you use like `UNCOMP_HEAD.free_size = log_dump_get_free_size();` directly, you can save the stack.

os/kernel/log_dump/log_dump_manager_start.c
```diff
@@ -0,0 +1,52 @@
...
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void log_dump_manager_start(void) {
+	int pid;
+	svdbg("Starting log_dump manager thread\n");
```
jeongchanKim, line 47: How about printing this log after checking the successful thread creation?
If the kernel thread is not created well, both `Starting` and `Failed` are showing, so can be confused.

os/kernel/log_dump/log_dump_save.c
```diff
@@ -0,0 +1,171 @@
...
+		return LOG_DUMP_FAIL;
+	return LOG_DUMP_OK;
+}
+
+static int __request_compress(const int index)
+{
+	struct log_dump_msg comp_msg;
+	comp_msg.uncomp_idx = index;
+
+	int ret = mq_send(mq_fd, (const char *)&comp_msg, sizeof(struct log_dump_msg), 100);
```
jeongchanKim, line 57: For many cases, to debug the failure of mq creation, it is better to check the return value of `mq_send`. 
How about printing if failed?

os/kernel/log_dump/log_dump.h
```diff
@@ -0,0 +1,178 @@
...
+#include <queue.h>
+#include <mqueue.h>
+#include <stdbool.h>
+
+/****************************************************************************
+ * Pre-Processor Definitions
+ ****************************************************************************/
+
+#define LOG_DUMP_OK_HEAPCHK		1	/* To check heap available after success return */
+#define LOG_DUMP_MEM_FAIL	-1
```
jeongarmy, line 37: It is same value as LOG_DUMP_COMPRESS_FAIL.
Please change and move it to line 43 for readability.

os/include/tinyara/log_dump/log_dump.h
```diff
@@ -19,21 +19,22 @@
...
-/* Log Dump Thread information */
-#define LOG_DUMP_NAME        "log_dump"		/* Log dump thread name */
-#define LOG_DUMP_STACKSIZE   16384		/* Log dump thread stack size */
-#define LOG_DUMP_PRIORITY    100		/* Log dump thread priority */
-
 #define LOGDUMP_SAVE_START	"1"
 #define LOGDUMP_SAVE_STOP	"2"
 
+#define LOG_DUMP_OK		0	/* Used for successful return check */
+#define LOG_DUMP_OPT_FAIL	-2
```
jeongarmy, line 35: There are definitions for failures like LOG_DUMP_XXX_FAIL in multiple files, os/include/tinyara/log_dump/log_dump.h and os/kernel/log_dump/log_dump.h
How about putting them in one place to avoid confusion?

zambbo: The header file in os/include/tinyara/log_dump/log_dump.h can be used any thread while the header file in os/kernel/log_dump/log_dump.h is only used internally.
So, I seperated the definitions.

PR #5895, opened 2023-08-04
os/arch/arm/src/amebasmart: Set AmebaSmart UART0, 1

kaikaikaime:
The Uart has been set similar to AmebaD ttyS2 is connected UART1


sunghan-chang:
Here is a public one. Noone knows what micom is. Let's title it as `UART 1 ? 2? 3?` like that.

sunghan-chang (approved):

PR #5898, opened 2023-08-10
build/configs/rtl87XX: Set -fno-common flag always

jeongarmy:
This flag specifies that the compiler places uninitialized global variables in the BSS section of the object file. It is ok to always have the fno-common flag, so I set the flag to avoid confusion.


sunghan-chang (commented):
Let's revert file permission to 644.

sunghan-chang (approved):

PR #5899, opened 2023-08-11
os/board/rtl8721csm: Add non-secure print into Secure fault prehandler

ziyik:
Add PANIC() to print non-secure register into Secure fault prehandler
Add Security level slot info into Secure during init


jylee9613 (approved):
looks good to me

jeongarmy (approved):
LGTM

PR #5901, opened 2023-08-16
framework/preference: Fix build issues with perefernce framework

amogh-samsung:
- Add missing header file for shared and private preference.
" CC:  src/preference/private_preference.c
src/preference/private_preference.c: In function 'preference_set_int':
src/preference/private_preference.c:45:19: error: 'PREFERENCE_TYPE_INT' undeclared (first use in this function)
  data.attr.type = PREFERENCE_TYPE_INT;
                   ^~~~~~~~~~~~~~~~~~~  "

- Correct input parameter type for private preference functions to match function declarations in header file
" CC:  src/preference/private_preference.c
src/preference/private_preference.c:281:5: error: conflicting types for 'preference_set_changed_cb'
 int preference_set_changed_cb(char *key, preference_changed_cb callback, void *user_data)
     ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/preference/private_preference.c:30:0:
/root/tizenrt/framework/include/preference/preference.h:195:5: note: previous declaration of 'preference_set_changed_cb' was here
 int preference_set_changed_cb(const char *key, preference_changed_cb callback, void *cb_data);
     ^~~~~~~~~~~~~~~~~~~~~~~~~
src/preference/private_preference.c:312:5: error: conflicting types for 'preference_unset_changed_cb'
 int preference_unset_changed_cb(char *key)
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/preference/private_preference.c:30:0:
/root/tizenrt/framework/include/preference/preference.h:204:5: note: previous declaration of 'preference_unset_changed_cb' was here
 int preference_unset_changed_cb(const char *key);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
Makefile:57: recipe for target 'private_preference.o' failed "

### DIFF
--- a/framework/src/preference/private_preference.c
+++ b/framework/src/preference/private_preference.c
@@ -26,6 +26,7 @@
 #include <sys/prctl.h>
 
 #include <tinyara/preference.h>
+#include <preference/preference.h>
 
 #include "preference_internal.h"
 
@@ -276,7 +277,7 @@ int preference_remove_all(void)
 /****************************************************************************
  * Callback Functions
  ****************************************************************************/
-int preference_set_changed_cb(char *key, preference_changed_cb callback, void *user_data)
+int preference_set_changed_cb(const char *key, preference_changed_cb callback, void *user_data)
 {
 	int ret;
 	struct sigaction act;
@@ -307,7 +308,7 @@ int preference_set_changed_cb(char *key, preference_changed_cb callback, void *u
 	return prctl(PR_SET_PREFERENCE_CB, &data);
 }
 
-int preference_unset_changed_cb(char *key)
+int preference_unset_changed_cb(const char *key)
 {
 	if (key == NULL) {
 		return PREFERENCE_INVALID_PARAMETER;
--- a/framework/src/preference/shared_preference.c
+++ b/framework/src/preference/shared_preference.c
@@ -26,6 +26,7 @@
 #include <sys/prctl.h>
 
 #include <tinyara/preference.h>
+#include <preference/preference.h>
 
 #include "preference_internal.h"
 


sunghan-chang:
1. typo on the title
perefernce -> preference
2. Please leave build error logs in the description.
3. Please leave why you change `input parameter` in the description.

framework/src/preference/private_preference.c
```diff
@@ -276,7 +277,7 @@ int preference_remove_all(void)
 /****************************************************************************
  * Callback Functions
  ****************************************************************************/
-int preference_set_changed_cb(char *key, preference_changed_cb callback, void *user_data)
+int preference_set_changed_cb(const char *key, preference_changed_cb callback, void *user_data)
```
ewoodev, line 280: Why do you change only `key`?
`Callback` and `user_data` do not need to be changed?

amogh-samsung: The function declaration for `preference_set_changed_cb()` and `preference_unset_changed_cb()` states key as a constant character pointer.
But the current function definitions use a non-constant character pointer.
Hence, I suggest this change.

amogh-samsung:
> 1. typo on the title
>    perefernce -> preference
> 2. Please leave build error logs in the description.
> 3. Please leave why you change `input parameter` in the description.

Done all

framework/src/preference/private_preference.c
sunghan-chang: If necessary, you should modify the header as well.

amogh-samsung: The header file `framework/include/preference/preference.h` has a function declaration which identifies the key paramter as a constant character pointer.
However, this header file was not included in `private_preference.c` earlier and the build error indicating conflicting definitions (as shown in description) was avoided.
At the same time, similar function definitions (set_changed_cb and unset_changed_cb) in `shared_preference.c` already use constant keyword for character key pointers.

Hence, I feel adding `const` is correct.

Additionally, why the need for adding the header file (preference/preference.h) now?
This is because since https://github.com/Samsung/TizenRT/pull/5875 we have a missing MACRO error during build, which is also described in the description.

amogh-samsung: I think this public API which is a callback should have constant pointer for key anyways,
but we did not see error because the header file for the API was not included. 
Hence the function definition worked fine as stand alone.

jeongarmy (approved):
Looks good to me.

framework/src/preference/private_preference.c
amogh-samsung: Yes,
and done

sunghan-chang (approved):

PR #5905, opened 2023-08-17
mm/mm_heap, mem_leak_checker: fix heap corruption hang

ewoodev:
The chunk of heap can know the address of the next chunk through `node = node + node->size`.
Using this, all chunks of the heap can be traversed in the forward direction.

If the heap is corrupted and the node->size becomes 0, It can't move to the next chunk, and the iteration turns infinite.
Also, It is Locked the heap during the heap traversing. As a result, it becomes deadlock.

Therefore, Add to check the node size and it to cause ASSERT if it is zero during traversing forward direction.

hang reproduce code:
```
	int *val = (int*)malloc(sizeof(int) * 100);
	val = val - 1;
	*val = 0;

	mallinfo();
```

### DIFF
--- a/os/kernel/debug/mem_leak_checker.c
+++ b/os/kernel/debug/mem_leak_checker.c
@@ -22,6 +22,7 @@
 #include <tinyara/config.h>
 #include <stdlib.h>
 #include <debug.h>
+#include <assert.h>
 #include <stdio.h>
 #include <string.h>
 #include <queue.h>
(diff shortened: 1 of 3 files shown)


sunghan-chang (dismissed):

os/kernel/debug/mem_leak_checker.c
```diff
@@ -168,6 +170,7 @@ static void fill_hash_table(int *leak_cnt, int *broken_cnt)
 	{
 		node_size = SIZEOF_MM_ALLOCNODE;
 		for (node = leak_checker.heap_start[region]; node < leak_checker.heap_end[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
+			DEBUGASSERT(node->size);
```
sunghan-chang, line 173: Is there any reason why you use `DEBUGASSERT` instead of `ASSERT`?

ewoodev: No, ASSERT() is correct because these ASSERT checks cause hang. I modified it.

sunghan-chang (approved):

jylee9613 (approved):
looks good!

PR #5906, opened 2023-08-18
build/tools/amebasmart/bootloader: fix booting log corruption

zhongnuo-tang:
1. fix corrupted booting log due to mix log between different cores.
2. include target_loader.asm and target_loader.map, help to debug in the future if there are any issue in bootloader


sunghan-chang (commented):
Description and title are not matched. Could you let me know what the issue is? And could you write details in the description?

zhongnuo-tang:
> Description and title are not matched. Could you let me know what the issue is? And could you write details in the description?

Dear Mr Chang,
This PR is to fix boot log corruption.. In the future, when updating bootloader, I will include the .map and .asm file generated so that it will be easier for us to debug if there are any fault/issue faced.. So is it ok if i change the details as follows:

build/tools/amebasmart/bootloader: fix booting log corruption

1. fix corrupted booting log due to mix log between different cores.
2. include target_loader.asm and target_loader.map, help to debug in the future if there are any issue in bootloader

sunghan-chang (approved):

PR #5915, opened 2023-08-30
Enable MMU based memory protection for loadable apps

kishore-sn:
Enable MMU based memory protection for loadable apps


os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -254,3 +281,190 @@ void mmu_invalidate_region(uint32_t vstart, size_t size)
...
+ * Input Parameters:
+ *
+ * Returned Value:
+ ****************************************************************************/
+void mmu_map_app_region(int app_id, uint32_t *l1_pgtbl, uint32_t start, uint32_t size, bool ro, bool exec, bool global)
+{
+	uint32_t idx;
+	uint32_t val;
+	uint32_t end = start + size;
+	int l2_idx = 0;
```
abhishek-samsung, line 373: ```l2_idx``` is reset for every call to mmu_map_app_region and in this case, we call it thrice for rox, ro, and rw. So I think we may need to store it for an app.

os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -254,3 +281,190 @@ void mmu_invalidate_region(uint32_t vstart, size_t size)
...
+			
+			if ((l1_pgtbl[idx] & PMD_TYPE_MASK) != PMD_TYPE_PTE) {
+				// Yes. Allocate L2 page table for app.
+				l2_pgtbl = mmu_allocate_app_l2_pgtbl(app_id, l2_idx++);
+
+				// Fill the newly allocated l2 page table with default kernel page entries
+				uint32_t l2_start = start & PMD_SECT_PADDR_MASK;
+                                for (int i = 0; i < L2_PGTBL_NENTRIES; i++) {
+                                        l2_pgtbl[i] = l2_start | MMU_L2_MEMFLAGS;
+                                        l2_start += 0x1000;
```
abhishek-samsung, line 401: It's better to use ```SMALL_PAGE_SIZE``` here instead of 0x1000.

os/binfmt/libelf/libelf_addrenv.c
```diff
@@ -229,7 +240,11 @@ int elf_addrenv_alloc(FAR struct elf_loadinfo_s *loadinfo)
 #elif defined(CONFIG_ARMV8M_MPU)
 	loadinfo->binp->ramsize = MPU_ALIGN_UP(loadinfo->binp->ramsize);
 	loadinfo->binp->ramstart = (uint32_t)kmm_memalign_at(CONFIG_HEAP_INDEX_LOADED_APP, MPU_ALIGNMENT_BYTES, loadinfo->binp->ramsize);
+#elif defined(CONFIG_ARCH_USE_MMU)
+	loadinfo->binp->ramsize = MMU_ALIGN_UP(loadinfo->binp->ramsize);
+	loadinfo->binp->ramstart = (uint32_t)kmm_memalign_at(CONFIG_HEAP_INDEX_LOADED_APP, MMU_ALIGNMENT_BYTES, loadinfo->binp->ramsize);
```
sunghan-chang, line 245: Is it possible to move architecture-specific codes in a new file?
I think we should keep common files and arch-specific files separately.
Mixed code is much difficult to maintain.

os/arch/arm/src/armv7-a/mmu.h
```diff
@@ -664,6 +664,7 @@
 #  define ALL_PGTABLE_SIZE PGTABLE_SIZE
 #endif
 
+
```
sunghan-chang, line 728: Let's revert unnecessary change

sunghan-chang:
@kishore-sn There are several build breaks. Please check them.

os/arch/arm/src/armv7-a/arm_mmu.c
kishore-sn: Fixed.

os/arch/arm/src/armv7-a/arm_mmu.c
kishore-sn: Done.

os/arch/arm/src/armv7-a/mmu.h
kishore-sn: Sorry, i dont understand. These changes are required.

os/binfmt/libelf/libelf_addrenv.c
kishore-sn: In this case, it will be difficult to separate arch code. The memory for apps needs to be allocated based on the version of mmu or mpu used. But also, the allocated memory needs to be stored in the various data structures related to binary loader.

os/arch/arm/src/armv7-a/mmu.h
sunghan-chang: If you check this in the first commit, there is no other change. You just add a line. But it's minor.

os/binfmt/libelf/libelf_addrenv.c
sunghan-chang: In common file
```
int abc(void)
{
    ....
    loadinfo->binp->ramstart = boardspecific_memory(loadinfo);
    ...
}
```
In board-specific file
```
void *boardspecific_memory(.....)
{
#ifdef ARMV7_MPU
     .....
#elif ARMV8M_MPU
    ....
#elif ARMV7A_MMU
    ...
#else
    ...
#endif
}
```

Impossible this kind of separation?

sunghan-chang: Both can be in binfmt, but I want to separate a file for common and board-specific code like ARMv7m, v8m and v7a or mmu and mpu.

os/arch/arm/src/armv7-a/mmu.h
kishore-sn: Fixed.

os/binfmt/libelf/libelf_addrenv.c
sunghan-chang: @kishore-sn Could you let me know your opinion for this?

kishore-sn: Done. Pls check.

sunghan-chang (approved):

PR #5919, opened 2023-09-05
os/se/security_level : Add null check at malloc

jylee9613:
Add null check at malloc to prevent memory leakage

### DIFF
--- a/os/se/security_level/security_level.c
+++ b/os/se/security_level/security_level.c
@@ -85,17 +85,20 @@ int set_security_level(void)
 	struct sec_lowerhalf_s *se = se_get_device();
 	hal_data output;
 	hal_result_e hal_ret;
-	security_level_result_t ret;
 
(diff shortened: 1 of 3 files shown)


os/se/security_level/security_level.c
```diff
@@ -85,17 +85,21 @@ int set_security_level(void)
...
 	hal_result_e hal_ret;
-	security_level_result_t ret;
 
 	output.priv_len = 0;
 	output.priv = NULL;
 	output.data_len = 2;
 	output.data = (char *)kmm_malloc(output.data_len);
 
+	if (output.data == NULL) {
+		kmm_free(output.data);
```
sunghan-chang, line 95: `if (output.data == NULL)` means fail to allocate. Then, there is no address to free.

sunghan-chang (approved):

PR #5924, opened 2023-09-07
os/arch: Fix prefetch abort when CONFIG_OPTIMIZE_APP_RELOAD_TIME=n

kishore-sn:
When CONFIG_OPTIMIZE_APP_RELOAD_TIME is disabled, then the memory permission is set to RW-X so that we can execute the code in this region. However, this case of RW-X was not handled in the mmu setting APIs. Hence the permission was set as RW-NX which resulted in prefetch abort. Add new RW-X setting in mmu to fix the issue.


os/arch/arm/src/armv7-a/mmu.h
```diff
@@ -651,16 +651,22 @@
 #ifdef CONFIG_SMP
 #define MMU_APP_L1_RW         	(PMD_TYPE_SECT | PMD_SECT_AP_RW01 | PMD_CACHEABLE | \
                               		PMD_SECT_S | PMD_SECT_DOM(0) | PMD_SECT_XN | PMD_SECT_NS)
+#define MMU_APP_L1_RW         	(PMD_TYPE_SECT | PMD_SECT_AP_RW01 | PMD_CACHEABLE | \
```
abhishek-samsung, line 654: I think this should be ```MMU_APP_L1_RWX``` instead of ```MMU_APP_L1_RW```

os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -65,6 +65,8 @@ static void mmu_set_flags(uint32_t *val, bool ro, bool exec, uint8_t isL1)
 			*val |= MMU_APP_L1_ROX;
 		} else if (ro) {
 			*val |= MMU_APP_L1_RO;
+		} else if (!ro && exec) {
```
jeongarmy, line 68: I think `!ro` condition is unnecessary because of above `else if (ro)` condition.

os/arch/arm/src/armv7-a/mmu.h
kishore-sn: Fixed.

os/arch/arm/src/armv7-a/arm_mmu.c
kishore-sn: Fixed.

abhishek-samsung (approved):

sunghan-chang (commented):
This change is valid in ARMv7-A only. How about `arch/armv7-a: Fix ....` as a title?

jeongarmy (approved):
Looks good to me.

sunghan-chang (approved):

PR #5930, opened 2023-09-13
os/arch/arm/src/amebasmart, os/board/rtl8730e: Amebasmart SPI,I2C,I2S…

kaikaikaime:
… import and verification

SPI, I2C, I2S wrapper has been implemented together with testcase def config Fix one of the amebalite spi in one semaphore error get.


sunghan-chang:
Could you split commits like below?
1. Amebalite spi issue fix
2. Amebasmart i2c support
3. ..... spi support
4. ..... i2s support

sunghan-chang (commented):
@kaikaikaime Could you let me know where does call the initialization function (start function) of each peripheral?

os/arch/arm/src/amebasmart/Make.defs
```diff
@@ -233,10 +233,11 @@ CHIP_CSRCS += amebasmart_timer_lowerhalf.c
 CHIP_CSRCS += amebasmart_rtc_lowerhalf.c
 CHIP_CSRCS += amebasmart_rtc.c
 CHIP_CSRCS += amebasmart_watchdog_lowerhalf.c
-#CHIP_CSRCS += amebasmart_i2c.c
-#CHIP_CSRCS += amebasmart_i2s.c
-#CHIP_CSRCS += amebasmart_spi.c
-
+CHIP_CSRCS += amebasmart_i2c.c
+CHIP_CSRCS += amebasmart_spi.c
```
sunghan-chang, line 237: How about adding conditionals for them like i2s?
```
ifeq ($(CONFIG_AMEBASMART_I2C),y)
CHIP.....+=...i2c.c
endif
ifeq(....SPI),y)
CHIP...+=...spi.c
endif
```

kaikaikaime:
> @kaikaikaime Could you let me know where does call the initialization function (start function) of each peripheral?

Hi, it will be in rtl8730e_boot.c

sunghan-chang:
> Hi, it will be in rtl8730e_boot.c

They are not mandatory so that we could disable them. Please check my previous comment and add those conditional there as well.

os/arch/arm/src/amebasmart/Make.defs
sunghan-chang: This is the my previous comment. I suggest the same conditionals in the boot.c for init function calls.

kaikaikaime:
Hi Chang, this commit I2S tx is been tested to run. As for RX, this will be done in the later date.

PR #5933, opened 2023-09-14
os/board/rtl8730e: fix SoftAP MAC addr, refine high tp macro

jwei5:
* Set concurrent_enabled to false as concurrent mode is not used
* Fix issue of SoftAP having different MAC address from STA interface
* Refine high throughput macro definition for CA32, set in platform_autoconf.h


sunghan-chang (commented):
Could you split commits?
1. fix softap mac addr
2. refine high tp macro

If they are not related.

jwei5:
Noted, I have split the commits

sunghan-chang (approved):

PR #5938, opened 2023-09-18
os/board/rtl8730e: fix BLE related SVACE and warnings

yeetee179:


os/board/rtl8730e/src/component/bluetooth/api/ipc/bt_ipc_parameter_conversion.c
```diff
@@ -2260,6 +2260,7 @@ uint32_t rtk_br_get_act_ret_param_size(uint16_t act)
 //	}
 //
 //	return max_size;
+	return RTK_BT_OK;
```
sunghan-chang, line 2263: Is this function necessary with always OK even it requires size get?

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
```diff
@@ -287,15 +287,14 @@ void general_client_notify_hdl(void *data)
 	gattc_dump(ntf_ind->len, ntf_ind->value, (uint8_t *)"notify event");
 	trble_data read_result;
 	read_result.length = ntf_ind->len;
-	read_result.data = malloc(read_result.length);
+	read_result.data = (uint8_t *)malloc(read_result.length);
```
sunghan-chang, line 290: Here is no checking for malloc. As you already make a change for this kind of defect, I think you should check this as well.

os/board/rtl8730e/src/component/bluetooth/api/ipc/bt_ipc_parameter_conversion.c
yeetee179: I have modified the return values to max size.

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
yeetee179: I have added checks for malloced pointers

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
```diff
@@ -308,7 +312,12 @@ void general_client_indicate_hdl(void *data)
 	trble_data read_result;
 	
 	read_result.length = indicate_ind->len;
-	read_result.data = malloc(read_result.length);
+	read_result.data = (uint8_t *)malloc(read_result.length);
+	if (!read_result.data)
+	{
+		printf("fail to malloc data \n");
```
sunghan-chang, line 318: This is the same as line 293. When you find this log, you can't distinguish where it fails to allocate. Is it ok?

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
```diff
@@ -317,7 +326,7 @@ void general_client_indicate_hdl(void *data)
 //	}
 	
 	read_result.length = indicate_ind->len;
-	read_result.data = malloc(read_result.length);
+	read_result.data = (uint8_t *)malloc(read_result.length);
```
sunghan-chang, line 329: 1. There are two malloc for `read_result.data` but is an one free for `read_result.data`. First malloc will be a leak.
2. Here is no checking for null return.

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
yeetee179: I have modified the error log so that it prints the function name.

os/board/rtl8730e/src/component/bluetooth/profile/client/rtk_gcs_client.c
yeetee179: The additional malloc for read_result.data is removed, so there is no need for a second free on read_result.data. 
Also 1 check for null return is enough

jeongarmy:
Could you update commit message about SVACE warnings?

yeetee179:
> Could you update commit message about SVACE warnings?

I have updated the commit messages.

sunghan-chang (approved):

PR #5941, opened 2023-09-21
os/board/rtl8730e: update wificfg for tx power limit

jwei5:
- disable tx power limit for wifi


sunghan-chang:
Could you leave why you disable?

Dane-Kang:
@sunghan-chang , It is for sync WiFi status with TP1x. 
TP1x already disabled WiFi limit function. and we found AIdual is enabled.
So, to match WiFi status, we modified to disable the WiFi limit.
Thank you.

sunghan-chang (approved):

ewoodev (approved):
LGTM

PR #5944, opened 2023-09-22
os/board/rtl8730e: Fix WiFi related Build Warning and SVACE

ZhenBei-Sin:
Build Warning
- Added macro checking for declare the unused variable as it depends on the macro on/off
- Changed the NULL checking to 0 
SVACE
- DEREF_OF_NULL.RET.STAT: Added the NULL pointer checking
- BUFFER_OVERFLOW.LIB.EX: Modified the API with correct pass in parameter for the size
- UNREACHABLE_CODE: Added macro to wrap it for cover the switch case break
- OVERFLOW_UNDER_CHECK: Declared the wlan_idx[NET_IF_NUM] with follow thw NET_IF_NUM value
- PROC_USE.VULNERABLE: Replaced strcpy with strncpy with pass in length
- TAINTED_INT.LOOP: Added the condition checking for the unsigned value

### DIFF
--- a/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
+++ b/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
@@ -125,7 +125,9 @@ static void inic_ipc_host_rx_tasklet(void)
 	_queue *recv_queue = NULL;
 	struct pbuf *p_buf = NULL;
 	int index = 0;
+#ifndef CONFIG_PLATFORM_TIZENRT_OS
 	err_enum_t error = ERR_OK;
+#endif
 	inic_ipc_ex_msg_t ipc_msg = {0};
 
 	recv_queue = &g_inic_host_priv.recv_queue;
@@ -217,6 +219,15 @@ void inic_ipc_host_init_skb(void)
 	host_skb_data = (struct skb_data *)rtw_zmalloc(wifi_user_config.skb_num_ap * sizeof(struct skb_data));
 	if (!host_skb_info || !host_skb_data) {
 		DBG_8195A("%s=>skb malloc fail!\n\r", __func__);
+		if (host_skb_info) {
+			rtw_mfree((u8 *)host_skb_info, 0);
+			host_skb_info = NULL;
+		}
+		if (host_skb_data) {
+			rtw_mfree((u8 *)host_skb_data, 0);
+			host_skb_data = NULL;
+		}
+		return;
 	}
 
 	/*make sure the real memory is set to zero, or DCache_Invalidate in inic_ipc_host_send will get wrong values*/
@@ -309,6 +320,12 @@ void inic_ipc_host_rx_handler(int idx_wlan, struct sk_buff *skb)
 
 	/* allocate host_recv_buf and associate to the p_buf */
 	precvbuf = (struct host_recv_buf *)rtw_zmalloc(sizeof(struct host_recv_buf));
+	if (!precvbuf) {
+		if (p_buf) {
+			pbuf_free(p_buf);
+		}
+		goto RSP;
+	}
 	precvbuf->p_buf = p_buf;
 	precvbuf->idx_wlan = idx_wlan;
 
--- a/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api.c
+++ b/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api.c
@@ -275,7 +275,9 @@ void inic_ipc_api_host_task(void)
 		p_ipc_msg->EVENT_ID = 0;
 		DCache_Clean((u32)p_ipc_msg, sizeof(inic_ipc_dev_request_message));
 	} while (1);
-	vTaskDelete(NULL);
+#ifndef CONFIG_PLATFORM_TIZENRT_OS
+	rtw_delete_task(NULL);
+#endif
 }
 
 /**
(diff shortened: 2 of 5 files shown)


ewoodev (approved):

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
```diff
@@ -309,6 +312,9 @@ void inic_ipc_host_rx_handler(int idx_wlan, struct sk_buff *skb)
 
 	/* allocate host_recv_buf and associate to the p_buf */
 	precvbuf = (struct host_recv_buf *)rtw_zmalloc(sizeof(struct host_recv_buf));
+	if (!precvbuf) {
+		goto RSP;
```
sunghan-chang, line 316: I can't find pbuf free in this case. Could you check it?

ZhenBei-Sin: Hi @sunghan-chang ,

Added the call to free the pbuf 
/* free p_buf for the case: SUCCESS allocate pbuf but FAILED allocate precvbuf */

https://github.com/Samsung/TizenRT/compare/1207b1c2d6345ee7a8c7b1caaead340f116c265f..463ddef16a14b779afd017d80cfe2474b2913e14

Thank You.

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
```diff
@@ -323,6 +329,11 @@ void inic_ipc_host_rx_handler(int idx_wlan, struct sk_buff *skb)
 	DCache_CleanInvalidate(((u32)skb - sizeof(struct list_head)), sizeof(struct skb_info));
 #endif /* CONFIG_ENABLE_CACHE */
 
+	/* free p_buf for the case: SUCCESS allocate pbuf but FAILED allocate precvbuf */
```
sunghan-chang, line 332: The `RSP` tab is executed at both cases, success and failure.
When everything is ok (success case), this will free the `pbuf`. 
I think this is wrong. Could you confirm it again?

ZhenBei-Sin: Your statement is correct.
RSP label will go in as no return in between the function.
I have removed the comment.

Added the p_buf checking before freeing the p_buf due to line298: goto RSP 
If p_buf is NULL and goto RSP without checking the p_buf and do free will be an issue.

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
```diff
@@ -323,6 +329,10 @@ void inic_ipc_host_rx_handler(int idx_wlan, struct sk_buff *skb)
 	DCache_CleanInvalidate(((u32)skb - sizeof(struct list_head)), sizeof(struct skb_info));
 #endif /* CONFIG_ENABLE_CACHE */
 
+	if (p_buf) {
+		pbuf_free(p_buf);
+	}
```
sunghan-chang, line 334: @ZhenBei-Sin You removed the comment only. But freeing `p_buf` at function success case is wrong. `p_buf` should be kept and should be used in recv buffer. Isn't it?

sunghan-chang: function operation failure case
1. fail to alloc p_buf, go to the the RSP label, there is no free memory 
2. fail to alloc precvbuf, go to the RSP label, p_buf should be freed not to get memory leakage
They are ok with this code.


function operation success case (no error at anywhere)
1. after in-queue, go in the RSP label, it should not free any memory (p_buf, precvbuf) to use them later
but this new code frees p_buf in this case. then, your some code will use freed memory, then it causes memory corruption.

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_trx.c
```diff
@@ -217,6 +219,7 @@ void inic_ipc_host_init_skb(void)
 	host_skb_data = (struct skb_data *)rtw_zmalloc(wifi_user_config.skb_num_ap * sizeof(struct skb_data));
 	if (!host_skb_info || !host_skb_data) {
 		DBG_8195A("%s=>skb malloc fail!\n\r", __func__);
+		return;
```
sunghan-chang, line 222: Don't you need to free any allocated memory? I mean 
1. both failed -> no need to free
2. alloc host_skb_info, but fail on host_skb_data -> should free host_skb_info
3. fail on info, allocated data -> should free host_skb_data

But this just ends the function. When you call this function again later, previous allocated memory will be a memory leak.

sunghan-chang: Oh beginning of this function checks allocated memory. But that's all. It will fail to do 2nd calling after 1st allocation failure. How about adding freeing allocated one and adding null for init?

ZhenBei-Sin: Thank you for your explanation and suggestion way.
Added the free based on the scenario given.

sunghan-chang: Need to do initialization with NULL.
```
if (info) {
   free(info);
   info = NULL;
}
....
```
Because you check that at beginning of the function.

ZhenBei-Sin: yes, correct... forgot to add the assign to NULL.
Let me add on now.

sunghan-chang (approved):

PR #5947, opened 2023-09-27
os/board/rtl8721csm: fix Assert on MAC filtering(in black list) with WPA3 issue

gexuyan:


sunghan-chang:
1. Could you add details in the description?
2. How about other chipsets like rtl8730e?

gexuyan:
> 1. Could you add details in the description?
> 2. How about other chipsets like rtl8730e?

Sorry Mr. Chang this PR will be closed first. Thank you.

PR #5948, opened 2023-09-27
Armv7a Arch code refactoring 

kishore-sn:
When we ported armv7a code from nuttx, we have not picked some of the arch related changes done in tizenrt for armv8m and other architectures. his commit picks these changes to arnv7a.

    - Port tizenrt specific changes from v8m to v7a
    - Move duplicate files in v7a to common directory
    - Rename some arch files (example arm_internal.h to up_internal.h)
    - Refactor and enhance assert code to handle app reload, dumps, etc..
    - Add configs to simplify the dumps which will be printed during assert
    - Enable app reloading and reload test scenarios
    - Fix heap corruption during app reload


build/configs/rtl8730e/loadable_apps_16M/defconfig
```diff
@@ -1274,7 +1275,7 @@ CONFIG_EXAMPLES_HELLO=y
 # CONFIG_EXAMPLES_LIBTUV is not set
 # CONFIG_EXAMPLES_LWNL_SAMPLE is not set
 # CONFIG_EXAMPLES_MEMORY_FRAGMENTATION_TEST is not set
-# CONFIG_EXAMPLES_MPU_TEST is not set
+CONFIG_EXAMPLES_MPU_TEST=y
```
sunghan-chang, line 1278: There is no mpu on rtl8730e. This looks strange. How about renaming as memory protection test regardless of this PR? I mean new PR.

os/arch/Kconfig
```diff
@@ -454,6 +454,19 @@ config ARCH_IRQPRIO
 		interrupts with certain configuration options.  So this selection
```
amogh-samsung, line 454: Minor: commit title says - 
"This commit picks these changes to arnv7a" -> "armv7a" please

build/configs/rtl8730e/loadable_apps_16M/defconfig
kishore-sn: Done. I added a commit for this.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ap_core/ameba_app_start.c
```diff
@@ -21,6 +20,34 @@
...
+ * (since each L2 table can map 1 MB memory). However, we are currently reserving 
+ * space assuming each app and common binary require 4 L2 page tables
+ * 5. The start address of first kernel L1 page table must be 16KB aligned
+ *
+ * Based on the above considerations, we setup the page table size as 
+ * 1 Kernel L1 table + 2 app L1 tables + (2 apps + 1 common) * 4 L2 page tables
+ * 16 KB + 16 KB + 16 KB + 3 * 4 * 1 KB = 60 KB
+ * 60 KB = 0xF000 ==> Round it up to 16KB boundary ==> 0x10000
+ */
+#define PGTABLE_SIZE       0x00010000
```
abhishek-samsung, line 40: How about using the newly added config CONFIG_NUM_L2_PER_APP here so that user need not change this whenever the config is changed?

os/binfmt/binfmt_unloadmodule.c
```diff
@@ -189,11 +189,15 @@ int unload_module(FAR struct binary_s *binp)
...
 #endif
+			for (int section_idx = 0; section_idx < BIN_MAX; section_idx++) {
+				binp->sections[section_idx] = NULL;
+			}
 		}
 #else
 		/* Whole loading sections are in one memory block, so free the first allocated memory is enough. */
 		binfo("Freeing : %p\n", binp->sections[0]);
 		kmm_free((FAR void *)binp->sections[0]);
+		binp->sections[0] = NULL;
```
abhishek-samsung, line 200: Why is this assignment required? Isn't kmm_free enough?

kishore-sn: We seem to be freeing these section multiple times based on whether the section pointer is non zero. Even though this wont create any issue, its better to always set the pointer to null after freeing it to avoid any future possibility of dangling pointers.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ap_core/ameba_app_start.c
kishore-sn: I tried this, but this is causing several other issues. 
Even if we make this change, there are many other things like page size, etc which are defined in other header files. 
The main problem is that these values are defined in mmu.h and used in other assembly files of the target chipset. So, any attempt to use complex definition directives is resulting in multiple errors. So let us keep the current status.

os/kernel/task/task_dumptasks.c
```diff
@@ -139,14 +139,15 @@ static void task_taskdump(FAR struct tcb_s *tcb, FAR void *arg)
 
 void task_show_alivetask_list(void)
 {
-	lldbg("*******************************************\n");
-	lldbg("List of all tasks in the system:\n");
-	lldbg("*******************************************\n");
+	lldbg_noarg("---------------------------------------------------------------------------------------------------\n");
```
sunghan-chang, line 142: With this change, TRAP is working? Could you check this with @thapav ?

os/binfmt/binfmt_arch_apis.c
```diff
@@ -109,19 +109,19 @@ void binfmt_arch_init_mem_protect(struct binary_s *binp)
 
 void binfmt_arch_deinit_mem_protect(struct binary_s *binp)
 {
-#if (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU))
 #ifdef CONFIG_SUPPORT_COMMON_BINARY
 	if (binp->islibrary) {
+#if (defined(CONFIG_ARMV7M_MPU) || defined(CONFIG_ARMV8M_MPU))
```
sunghan-chang, line 114: os/kernel/task/task_terminate_unloaded.c
os/kernel/task/task_terminate.c

Both uses `CONFIG_ARM_MPU` instead of 7m_mpu and 8m_mpu configs.
https://github.com/Samsung/TizenRT/pull/5948/files#diff-de18ff5621c82ce73f2ccd2b0d9bbead6e92ae7836171469f03d4610b0cd6d34R191
Do we need to use them differently?

os/arch/arm/src/common/up_usestack.c
```diff
@@ -171,15 +152,20 @@ int up_use_stack(struct tcb_s *tcb, void *stack, size_t stack_size)
...
+#ifdef CONFIG_ARCH_ARMV7A_FAMILY
+	size_of_stack = top_of_stack - (uint32_t)tcb->stack_alloc_ptr;
+#else
 	size_of_stack = top_of_stack - (uint32_t)tcb->stack_alloc_ptr + 4;
+#endif
 
 	/* Save the adjusted stack values in the struct tcb_s */
 
 	tcb->adj_stack_ptr = (uint32_t *)top_of_stack;
+  tcb->stack_base_ptr  = tcb->stack_alloc_ptr;
```
sunghan-chang, line 164: Please check alignment

os/arch/arm/src/common/up_usestack.c
```diff
@@ -156,7 +133,11 @@ int up_use_stack(struct tcb_s *tcb, void *stack, size_t stack_size)
 	 * referenced as positive word offsets from sp.
 	 */
 
+#ifdef CONFIG_ARCH_ARMV7A_FAMILY
+	top_of_stack = (uint32_t)tcb->stack_alloc_ptr + stack_size;
+#else
 	top_of_stack = (uint32_t)tcb->stack_alloc_ptr + stack_size - 4;
+#endif
```
sunghan-chang, line 140: This is a common file. It looks not good to add archi specific code here.
How about defining new config for this?
1. new config to on / off for -4
```
in up_userstack.c
#ifdef CONFIG_NEW
   top_.._stack = ....;
#else
   top_.._stack = .... - 4;
#endif

in Kconfig
config ARMV7A
    select NEW
```
2. new config to give adj value
```
in c
    top_...stack = .... + stack_size - CONFIG_NEW;

in defconfig
CONFIG_NEW=4
```

sunghan-chang: And I can't find why -4 is unnecessary. Could you leave comment?

sunghan-chang (commented):
1. I can find mixed indentations, tab or spaces. We should use tabs instead of spaces. After this PR, please make a commit for them.
2. There are several modifications in common files. Are they verified on other chipsets?

sunghan-chang:
@kishore-sn As you know, I am a beginner in this topic. Could you get review from your team and let me know result?

os/kernel/task/task_dumptasks.c
thapav: Sure I will check it once merged.

os/arch/arm/src/common/up_usestack.c
kishore-sn: hi @sunghan-chang 
I have been looking for a reason for this. But, as of now I have not been able to find out the exact reason. 
However, it looks like all cortex m code requires the -4 and cortex a32 does not need the -4.
If I try to change this in anyway, it is resulting in various crashes. 
So, I would like to keep this code for now and we will look into the reasoning for this at a later time. I think this can be a low priority activity.

kishore-sn: > This is a common file. It looks not good to add archi specific code here. How about defining new config for this?
> 
> 1. new config to on / off for -4
> 
> ```
> in up_userstack.c
> #ifdef CONFIG_NEW
>    top_.._stack = ....;
> #else
>    top_.._stack = .... - 4;
> #endif
> 
> in Kconfig
> config ARMV7A
>     select NEW
> ```
> 
> 2. new config to give adj value
> 
> ```
> in c
>     top_...stack = .... + stack_size - CONFIG_NEW;
> 
> in defconfig
> CONFIG_NEW=4
> ```

I am not sure what to name the CONFIG_NEW because I am not able to find out the reason for this.

kishore-sn:
> 2. There are several modifications in common files. Are they verified on other chipsets?

@sunghan-chang 
I have verified build_test with these changes. 
Also, I have tested basic functionality on all rtk chipsets.

```
RTK8721
Flat bootup			: PASS
Loadable bootup		: PASS
app reload			: PASS
kernel tc 			: PASS

RTK 8720  (Same behavior without my commits also)
Flat bootup			: PASS
Loadable bootup		: FAIL (Hang)
app reload			: FAIL (Not enabled)
kernel tc 			: FAIL (Assertion)

RTK 8730
Flat bootup			: PASS
Loadable bootup		: PASS
app reload			: PASS
kernel tc 			: PASS
```

Please let me know if anything else needs to be tested

os/arch/arm/src/common/up_internal.h
```diff
@@ -332,6 +413,60 @@ EXTERN uint32_t _eramfuncs;		/* Copy destination end address in RAM */
...
+void modifyreg8(unsigned int addr, uint8_t clearbits, uint8_t setbits);
+void modifyreg16(unsigned int addr, uint16_t clearbits, uint16_t setbits);
+void modifyreg32(unsigned int addr, uint32_t clearbits, uint32_t setbits);
+void arm_boot(void);
+uint32_t *arm_decodeirq(uint32_t *regs);
+void arm_fullcontextrestore(uint32_t *restoreregs) noreturn_function;
+void arm_switchcontext(uint32_t **saveregs, uint32_t *restoreregs);
+void arm_sigdeliver(void);
+uint32_t *arm_doirq(int irq, uint32_t *regs);
+uint32_t *arm_dataabort(uint32_t *regs, uint32_t dfar, uint32_t dfsr);
```
ewoodev, line 426: It make increase code in up_internal.h

up_XXX and arm_XXX both exist.
I think, we need to discuss about this.

1. keep arm_xxx and up_xxx, and remove common/internal.h and seperate per armvX-X
2. unify both into one, only in arch/arm.

os/arch/arm/src/common/up_internal.h
```diff
@@ -122,12 +129,43 @@
...
 
+
+/* ARM requires at least a 4-byte stack alignment.  For use with EABI and
+ * floating point, the stack must be aligned to 8-byte addresses.
+ */
+
+#ifndef CONFIG_STACK_ALIGNMENT
+#ifdef __ARM_EABI__
+#define CONFIG_STACK_ALIGNMENT 8
+#else
```
ewoodev, line 140: This CONFIG_STACK_ALIGNMENT and STACK_ALIGNMENT also exist.

kishore-sn: Removed it as it is not required.

os/arch/arm/src/common/up_internal.h
kishore-sn: True. We need to discuss this. 
But let us merge this PR first, as it is required to verify many features on 8730.
We can take up this activity separately.

sunghan-chang (dismissed):

os/arch/Kconfig
```diff
@@ -454,6 +454,19 @@ config ARCH_IRQPRIO
...
 		should be used with caution.
 
+config ARCH_MMUDUMP
```
sunghan-chang, line 457: Please check first commit, it's another commit and it has your changes. Something wrong.

os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -508,27 +512,27 @@ void mmu_dump_pgtbl(void)
...
 	uint32_t *l1tbl = mmu_l1_pgtable();
 
-	lldbg("L1 page table base addr = 0x%08x\n", l1tbl);
+	lldbg_noarg("L1 page table base addr = 0x%08x\n", l1tbl);
 
-	lldbg("=========================================\n");
-	lldbg("ENTRY      TYPE    OUT             ACCESS\n");
-	lldbg("ADDR               ADDR                  \n");
-	lldbg("=========================================\n");
+	lldbg_noarg("=====================================================================\n");
```
sunghan-chang, line 517: Will you handle this for the TRAP later? @thapav

kishore-sn: Yes Chang.. We will make sure to raise a separate PR for TRAP.

sunghan-chang (approved):

PR #5949, opened 2023-09-29
os/arch/arm/src/ameba: GPIO pull up/down fix

kaikaikaime:
Fix initial status of the GPIO when it should be pulled up/down in either directions


sunghan-chang (commented):
1. Without this PR, amebad was working well. Is this necessary?
2. Is there no specific hw register for this? To do pull up / down, do you want to set value for high or low??

PR #5952, opened 2023-10-11
mem_leak test on rtl8730e

ritesh55555:
- Enables mem_leak_test config defualt to y 
- Adds memory_region.h file for rtl8730e board type 
- Initializes "tcb->adj_stack_ptr" in  armv7-a/arm_createstack.c as it was missing before . It is used to get user heap data structure using umm_get_heap(tcb->adj_stack_ptr) .   ( incase of flat build)
- initializing "tcb->adj_stack_ptr" is armv7-a/arm_createstack.c is already taken care in https://github.com/Samsung/TizenRT/pull/5948 

Result of mem_leak test is as follows:
```
TASH>>Hello, World!!
allocated address is 0x603a1f30
mem_leak

Kernel :
Type   |    Addr    | Size(byte) |    Owner   | PID 
---------------------------------------------------
LEAK   | 0x603a1f30 |        16  | 0x6030ff0b | 13
*** 1 LEAKS, 0 BROKENS.
TASH>>hello
TASH>>Hello, World!!
allocated address is 0x603a85f0
mem_leak

Kernel :
Type   |    Addr    | Size(byte) |    Owner   | PID 
---------------------------------------------------
LEAK   | 0x603a1f30 |        16  | 0x6030ff0b | 13
LEAK   | 0x603a85f0 |        16  | 0x6030ff0b | 15
*** 2 LEAKS, 0 BROKENS.

```


apps/system/mem_leak_checker/Kconfig
```diff
@@ -5,7 +5,7 @@
 
 config MEM_LEAK_CHECKER
 	bool "Memory Leak Checker"
-	default n
+	default y
```
sunghan-chang, line 8: Let's enable them in defconfig, not in Kconfig.

sunghan-chang:
1. Please add description why you modify this in the commit description.
2. Please add mem leak result with your pr on rtl8730e in commit description.

sunghan-chang:
Please confirm there is no mem leak on rtl8730e.

abhishek-samsung:
I think it's better to add a check in mem_leak_checker.c file to see if the heap pointer returned is valid,

https://github.com/Samsung/TizenRT/blob/f5bdafdab2781c843de93515e78eace35f173ae8/os/kernel/debug/mem_leak_checker.c#L366-L373

If its not, then we can terminate the task and display the appropriate error message, something like this

`if (!heap) {
        printf("No valid heap found for binary : %s\n", bin_name);
}`

build/configs/rtl8730e/cpp_loadable_8M/defconfig
```diff
@@ -1434,7 +1434,9 @@ CONFIG_SYSTEM_PREAPP_STACKSIZE=2048
 CONFIG_SYSTEM_IPERF=y
```
abhishek-samsung, line 1434: This commit doesn't follow the commit guidelines, please modify the title accordingly. And I think we can split this PR into two commits,

1. add missing initialization of adj_stack_ptr for armv7a
2. fix mem leak checker for rtl8730e and enable it by default

sunghan-chang:
@ritesh55555 Could you check the build break?

build/configs/rtl8730e/loadable_apps_16M/defconfig
```diff
@@ -1209,8 +1209,8 @@ CONFIG_ELF_CACHE_READ=y
 CONFIG_ELF_CACHE_BLOCK_SIZE=2048
 CONFIG_ELF_CACHE_BLOCKS_COUNT=60
 # CONFIG_SYMTAB_ORDEREDBYNAME is not set
-# CONFIG_OPTIMIZE_APP_RELOAD_TIME is not set
-# CONFIG_SAVE_BIN_SECTION_ADDR is not set
+CONFIG_OPTIMIZE_APP_RELOAD_TIME=y
+CONFIG_SAVE_BIN_SECTION_ADDR=y
```
sunghan-chang, line 1213: 1. To work mem leak, is this mandatory? If yes, we should check it on menuconfig using Kconfig dependency.
2. Let's enable them with new commit.

sunghan-chang:
@ritesh55555 Two commits have no description. Could you fill in the commit description with what issue is and why you modify?

sunghan-chang:
@ritesh55555 With this PR, does the mem leak of rtl8730e/hello work?

build/configs/rtl8730e/loadable_apps_16M/defconfig
ritesh55555: 1. yes , it is mandatory for loadable types . i didnot understand "we should check it on menuconfig using Kconfig dependecny"

sunghan-chang:
```
config MEM_LEAK  (I don't write exact name)
    depends on OPTIMIZE_APP_RELOAD_TIME  if LOADABLE
   or
   select OPTIMIZE_APP..._TIME if LOADABLE
```
You said they are mandatory to work memory leakage checker. Then, we should prevent them disabled with mem leak checker enabled.

sunghan-chang: If the config is standalone (no dependency with other config), we can use `select`.
But if the config has other dependencies, then we can't use `select`. We should use `depends on`

sunghan-chang: Mandatory is `CONFIG_SAVE_BIN_SECTION_ADDR` only? or both `CONFIG_SAVE_BIN_SECTION_ADDR` and `CONFIG_OPTIMIZE_APP_RELOAD_TIME`?

ritesh55555: > ```
> config MEM_LEAK  (I don't write exact name)
>     depends on OPTIMIZE_APP_RELOAD_TIME  if LOADABLE
>    or
>    select OPTIMIZE_APP..._TIME if LOADABLE
> ```
> 
> You said they are mandatory to work memory leakage checker. Then, we should prevent them disabled with mem leak checker enabled.

it is already implemented in Kconfig by using ```select```

ritesh55555: > Mandatory is `CONFIG_SAVE_BIN_SECTION_ADDR` only? or both `CONFIG_SAVE_BIN_SECTION_ADDR` and `CONFIG_OPTIMIZE_APP_RELOAD_TIME`?

Both

sunghan-chang (approved):

amogh-samsung:
@ritesh55555 request you please push the Kconfig changes with which these selections will be retained after menuconfig.

PR #5955, opened 2023-10-12
os/arch/arm/src/ameba: GPIO pull up/down fix

kaikaikaime:
Tested that by removing gpio int deinit** during gpio initialisation, pull up/ down is working.
The deinit interrupt function also clears out the previous parameters. To prevent doing so deinit, this line has been removed.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_lowerhalf.c
+++ b/os/arch/arm/src/amebad/amebad_lowerhalf.c
@@ -205,7 +205,6 @@ static int amebad_gpio_enable(FAR struct gpio_lowerhalf_s *lower, int falling,
 	} else {
 		gpio_irq_disable((gpio_irq_t *)&priv->obj);
 		gpio_irq_set((gpio_irq_t *)&priv->obj, event, 0);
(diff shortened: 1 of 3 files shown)


sunghan-chang:
@kaikaikaime Could you add some description why removing irq disable makes pull up/down working?

kaikaikaime:
I made an error, it was gpio interrupt deinit

sunghan-chang:
@kaikaikaime 1. Could you leave that in the commit description? not in PR description.
2. `The deinit interrupt function also clears out the previous parameters. To prevent doing so deinit, this line has been removed.`
  Does it mean deinit is unnecessary?? Don't you need to move location instead of removal?

sunghan-chang:
@kaikaikaime Any update?

kaikaikaime:
@sunghan-chang ,

I have added the extra line into the commit message. 
As for the deinit function, removing would be a better option since it resets the gpio back to default mode and deinit is common function used in both gpio irq and gpio api.

sunghan-chang (approved):

PR #5956, opened 2023-10-12
board/rtl8730e: Add choice to select flash size of board in Kconfig

jeongarmy:
There is 2 kinds of flash capacity of rtl8730e board: 8Mbytes, 16Mbytes.
The related config values should be set according to flash size user selected.


os/board/rtl8730e/Kconfig
```diff
@@ -20,9 +20,25 @@ config FLASH_START_ADDR
         A start address of flash (in hex).
         This is fixed value, so user doesn't need to change it.
 
+choice
+    prompt "Board Flash Size"
```
sunghan-chang, line 24: Let's add default value.
```
choice
   prompt "...."
   default BOARD_FLASH_16M

....
```

sunghan-chang (approved):

PR #5960, opened 2023-10-17
build/tools/amebasmart/bootloader: print Flash size in bootloader

zhongnuo-tang:


sunghan-chang:
Could you add details in commit description? (not in pr description)
1. new bootloader version you uploaded
2. example of new message for flash size

zhongnuo-tang:
> Could you add details in commit description? (not in pr description)
> 
> 1. new bootloader version you uploaded
> 2. example of new message for flash size

Dear @sunghan-chang,
I have added the above descriptions in commit message.

PR #5961, opened 2023-10-17
apps/examples/board_specific_demo : Apply secure world test for RTL8730E

jylee9613:
If non-secure world tries to access secure world, secure fault occurs. 
This commit applies secure world test for RTL8730E.


apps/examples/board_specific_demo/ameba_secure_world_demo/Kconfig
```diff
@@ -6,7 +6,7 @@
 config EXAMPLES_SECURE_WORLD
 	bool "Secure World example"
 	default n
-	depends on ARMV8M_TRUSTZONE && SE_AMEBA
+	depends on (ARMV8M_TRUSTZONE || AMEBASMART_TRUSTZONE) && SE_AMEBA
```
sunghan-chang, line 9: Original one is chip architecture config (ARMV8M), but your code is chipset code.
I think it should be ARMV7A_TRUSTZONE

jeongarmy, line 9: `ARMV7M_TRUSTZONE` has no relation with `CONFIG_AMEBASMART_TRUSTZONE`.
Please add `selects ARMV7M_TRUSTZONE` to AMEBASMART_TRUSTZONE like AMEBALITE_TRUSTZONE.

jylee9613: Unlike ARMV8M, ARMV7A_TRUSTZONE config is not defined in Kconfig.
When ARMV7A_TRUSTZONE is added, Menuconfig does not show secure_world example config.

jeongarmy: Sorry, I mean `ARMV7A`, not `ARMV7M`.

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -66,6 +66,16 @@ static int ns_flash_write(uint32_t address, uint32_t len, uint8_t *data)
 	return ret;
 }
 
+int rtl_ss_flash_write(uint32_t address, uint32_t len, uint8_t *data)
```
sunghan-chang, line 69: Let's split commits with single topic. This commit has too many different topics.

jylee9613: I splited it to secure_storage test commit. Thank you.

sunghan-chang: 1. This commit has changes on os and apps together. Let's split them as different commit.
2. Is this necessary in the se module? Is this a security operation?

jylee9613:
@sunghan-chang

2. Is this necessary in the se module? Is this a security operation?
-> Secure Storage is in secure world, and non-secure world cannot access to secure storage directly. So if trying to access secure storage address directly, assert occurs.

sunghan-chang:
@jylee9613 That's not what I asked.

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -214,12 +215,27 @@ int sstorage_main(int argc, char *argv[])
...
+	if (slot_index == 0) {
+		address = SLOT0_START_ADDR;
+	} else {
+		address = SLOT1_START_ADDR + (SLOT_SIZE * (slot_index - 1));
+	}
+
+	printf("Write Raw data in Slot: %d, Area A\n", slot_index);
+	printf("This operation will cause assert\n");
+	flash_address = address;
+	*flash_address = input_data;
```
sunghan-chang, line 237: This is wrong to write some data into flash. Please find proper way.

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -59,6 +61,47 @@ static uint32_t SLOT1_START_ADDR = 0x2F000;
...
+		printf("Failed to allocate buffer %d\n", MAX_SS_SIZE);
+		return -1;
+	}
+	memset(buffer, data, len);
+
+	#ifdef CONFIG_MTD_BYTE_WRITE
+	if (dev_mtd->write != NULL) {
+		ret = MTD_WRITE(dev_mtd, address, len, buffer);
+	}
+	#endif
```
sunghan-chang, line 91: As far as  I know, MTD function is not for user context.
@jeongarmy Could you share how to write some data at flash memory?

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -59,6 +61,47 @@ static uint32_t SLOT1_START_ADDR = 0x2F000;
 static uint32_t SLOT_SIZE = 0x1000;
 static uint32_t SLOT_AREA_OFFEST = 0x23000;
 
+static int flash_write(uint32_t address, uint32_t len, uint8_t *data) {
```
sunghan-chang, line 64: coding rule violation

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -59,6 +61,48 @@ static uint32_t SLOT1_START_ADDR = 0x2F000;
...
+		printf("Failed to allocate buffer %d\n", MAX_SS_SIZE);
+		return -1;
+	}
+	memset(buffer, data, len);
+
+	#ifdef CONFIG_MTD_BYTE_WRITE
+	if (dev_mtd->write != NULL) {
+		ret = MTD_WRITE(dev_mtd, address, len, buffer);
+	}
+	#endif
```
sunghan-chang, line 92: This function is `flash_write`. If the config `CONFIG_MTD_BYTE_WRITE` is disabled, then how does this function work?

apps/examples/board_specific_demo/ameba_secure_storage_demo/secure_storage.c
```diff
@@ -59,6 +61,48 @@ static uint32_t SLOT1_START_ADDR = 0x2F000;
...
+
+	if (ret <= 0) {
+		printf("Write Failed : %d\n", ret);
+		free(buffer);
+		return -1;
+	} else {
+		printf("Write Success\n");
+		printf("Length of written data : %d\n\n", ret);
+	}
```
sunghan-chang, line 101: You free the buffer at failure case, but you don't do in success case. Is it correct?

apps/examples/board_specific_demo/ameba_ss_write_protection_demo/ss_write_protection_test.c
```diff
@@ -214,12 +259,51 @@ int sstorage_main(int argc, char *argv[])
...
+	}
+
+	sleep(1);
+	address += SLOT_AREA_OFFEST;
+	printf("\nRead Raw data from Slot: %d, Area B\n", slot_index);
+	if (!rtl_ss_flash_read(address, length, (uint8_t *)rwbuf, 1)) {
+		printf("rtl_ss_flash_read() Fail, %d\n", __LINE__);
+	}
+
+	printf("Read Complete\n");
```
ewoodev, line 298: Both Fail and Read can be printed.

how about change reture -1?

jylee9613: Thank you for your review. I added function for read/write flash, and add "goto exit" if function failed.

os/arch/arm/src/armv7-a/Kconfig
```diff
@@ -169,3 +169,8 @@ config ARMV7A_DECODEFIQ
 		Select this option if your platform supports the function
 		arm_decodefiq().  This is used primarily to support secure TrustZone
 		interrupts received on the FIQ vector.
+
+config ARMV7A_TRUSTZONE
```
sunghan-chang, line 173: When do we need the config?
Basically RTL8730E uses the trustzone, then should we enable this by default? or this is only for the example?

jylee9613: This config is used in syscall.h, but it's not defined in Kconfig.
Because of this, when config is changed through menuconfig, build fails because CONFIG_SYS_RESERVED value is getting different.
```
#ifdef CONFIG_LIB_SYSCALL

#ifndef CONFIG_SYS_RESERVED
#error "CONFIG_SYS_RESERVED must be defined to have the value"
#endif //CONFIG_SYS_RESERVED

#ifdef CONFIG_BUILD_PROTECTED

#ifdef CONFIG_ARMV7A_TRUSTZONE
#if CONFIG_SYS_RESERVED != 11
#error "CONFIG_SYS_RESERVED must have the value 11"
#endif
#else
#if CONFIG_SYS_RESERVED != 8
#error "CONFIG_SYS_RESERVED must have the value 8"
#endif
#endif //CONFIG_ARMV7A_TRUSTZONE

#else //CONFIG_BUILD_PROTECTED

#ifdef CONFIG_ARMV7A_TRUSTZONE
#if CONFIG_SYS_RESERVED != 7
#error "CONFIG_SYS_RESERVED must have the value 7"
#endif
#else
#if CONFIG_SYS_RESERVED != 4
#error "CONFIG_SYS_RESERVED must have the value 4"
#endif
#endif //CONFIG_ARMV7A_TRUSTZONE

#endif //CONFIG_BUILD_PROTECTED

#endif //CONFIG_LIB_SYSCALL
```

jylee9613: @sunghan-chang @ewoodev I removed ARMv7-A tz config.

eunwoo-nam:
https://github.com/Samsung/TizenRT/pull/7388 is duplicated

PR #5962, opened 2023-10-18
build/tools/amebasmart: Modify Coex Case of Wifi Scan

ziyik:
-Use PTA instead of TDMA when doing Wifi Scan


sunghan-chang (commented):
What is the "PTA"?

sunghan-chang (approved):

PR #5968, opened 2023-10-23
add DEBUGASSERT_MM_FREE_NODE macro 

ewoodev:
### commit 1    
    mm/mm_heap: add mm_dump_heap_free_node_list function

    add mm_dump_heap_free_node_list function.
    This function prints the hex valuse contents of heap free nodes using heap free node list.
    
    example logs:
    
    ` #########################################################################################
      Dump heap free node list
      [ndx], [HEAD]: [FREE NODES(SIZE)]
    ` #########################################################################################
        0, 1003dc80:
        1, 1003dc98:
        2, 1003dcb0:
        3, 1003dcc8: 0200f220(240)
        4, 1003dce0:
        5, 1003dcf8: 02012dc0(592) 02016b70(576)
        6, 1003dd10: 0201eb70(1488)
        7, 1003dd28:
        8, 1003dd40:
        9, 1003dd58:
       10, 1003dd70:
       11, 1003dd88:
       12, 1003dda0:
       13, 1003ddb8:
       14, 1003ddd0:
       15, 1003dde8:
       16, 1003de00: 02201a10(1828320)
       17, 1003de18:
       18, 1003de30:
    `  #########################################################################################

### commit 2
    mm/mm_heap: move dump_node to mm_dump_node
    
    add the dump_node function, which was static, to the header for use in other codes and move it to mm_heap_dbg.c

### commit 3
    mm/mm_heap/mm_node.h: add DEBUGASSERT_MM_FREE_NODE macro
    
    The back link of heap free node cannot be null at any time, If null, it is an assert situation.
    Therefore, in the REMOVE_NODE_FROM_LIST macro, null is checked with DEBUGASSERT.
    But we can't known why the back link of the free node became null.
    
    Therefore, This commit adds DEBUGASSERT_MM_FREE_NODE macro, that print addresses of free nodes, head
    list of free node link and corrupted node nearby node for debugging.

### DIFF
--- a/os/mm/mm_heap/mm_node.h
+++ b/os/mm/mm_heap/mm_node.h
@@ -24,14 +24,29 @@
  ****************************************************************************/
 
 #include <assert.h>
+#include <debug.h>
+
+#include <tinyara/mm/mm.h>
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
+#define DEBUGASSERT_MM_FREE_NODE(heap, node)		\
+	do {		\
+		DEBUGASSERT(node);		\
+		if (!((node)->blink)) {		\
+			mfdbg("Corrupted free node %08x\n", node);		\
+			mm_dump_heap_free_node_list(heap);		\
+			mm_dump_node((struct mm_allocnode_s *)node, "CORRUPT NODE");		\
+			mm_dump_node((struct mm_allocnode_s *)((char *)node - (node->preceding & ~MM_ALLOC_BIT)), "PREV NODE");		\
+			mm_dump_node((struct mm_allocnode_s *)((char *)node + node->size), "NEXT NODE");		\
+			DEBUGPANIC();		\
+		}		\
+	} while (0)
+
 #define REMOVE_NODE_FROM_LIST(node)				\
 	do {							\
-		DEBUGASSERT((node)->blink);			\
 		(node)->blink->flink = (node)->flink;		\
 		if ((node)->flink) {				\
 			(node)->flink->blink = (node)->blink;	\
--- a/os/mm/mm_heap/mm_check_heap_corruption.c
+++ b/os/mm/mm_heap/mm_check_heap_corruption.c
@@ -80,41 +80,6 @@
  * Private Functions
  ****************************************************************************/
 
-static void dump_node(struct mm_allocnode_s *node, char *node_type)
-{
-#if defined(CONFIG_DEBUG_MM_HEAPINFO)  && (CONFIG_TASK_NAME_SIZE > 0)
-	char myname[CONFIG_TASK_NAME_SIZE + 1];
-	char is_stack[9] = "'s stack";
-#endif
-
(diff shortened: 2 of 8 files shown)


os/mm/mm_heap/mm_node.h
```diff
@@ -24,14 +24,48 @@
...
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
+#define DEBUGASSERT_MM_FREE_NODE(heap, node)		\
+	do {		\
+		DEBUGASSERT(node);		\
+		if (!((node)->blink)) {		\
+			abort_mode = true; \
```
jeongarmy, line 40: DEBUGPANIC() is a macro for `up_assert` function and up_assert sets `abort_mode` to true internally. Is it necessary?

os/mm/mm_heap/mm_node.h
```diff
@@ -24,14 +24,48 @@
...
+#include <debug.h>
+
+#include <tinyara/mm/mm.h>
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
+#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
+#define DEBUGASSERT_MM_FREE_NODE(heap, node)		\
```
jeongarmy, line 36: Do you have any reasons to separate DEBUGASSERT_MM_NODE according to build type? I think both are same except setting of `abort_mode`. If it is not necessary as I said in below comment, we don't need to separate them.

os/mm/mm_heap/mm_node.h
ewoodev:
```
#if defined(CONFIG_BUILD_FLAT) || defined(__KERNEL__)
#define mfdbg(format, ...)						\
	do {								\
		if (!IS_SECURE_STATE()) {				\
			if (abort_mode || up_interrupt_context()) {	\
				lldbg(format, ##__VA_ARGS__);		\
			} else {					\
				dbg(format, ##__VA_ARGS__);		\
			}						\
		}							\
	} while (0)
#else
#define mfdbg(format, ...) dbg(format, ##__VA_ARGS__)
#endif

```
https://github.com/Samsung/TizenRT/blob/4cced9a8944514cac4c28a2aab2717138b1e474b/os/include/debug.h#L504

mfdbg operation has dependency with abort_mode.

if after print mfdbg( using dbg), up_assert is called, dbg can't finsh printing.
becase uart driver and lib stdout buffer is stop and print only using lldbg.

So, To complete print just before up_assert, abort mode setting is needed.

ewoodev: Actullay, This issue is not solved by setting abort_mode early about loadable user heap.. .
So,  I'm checking to add lib stdout flush at up_assert.c

ewoodev: I raised flushing log uart PR #6660

PR #5742, opened 2023-01-12
lib/libc/syslog/lib_syslog.c : fix dbg and printf log mixup

swasthik22:
Use semaphore on stdout file stream, so that the entire dbg log is pushed to output thus stopping the log mixup due to context switching. 

PRINTING FROM THREAD
hello_main: printing from main
PRINTING FROM THREAD
hello_main: printing from main
PRINTING FROM THREAD
hello_main: printing from main
PRINTING FROM THREAD
hello_main: printing from main

PR #5969, opened 2023-10-25
mm/mm_heap: fix crash when CONFIG_DEBUG_MM_INFO is eanbled

abhishek-samsung:
When CONFIG_DEBUG_MM_INFO, dbg is used to print few logs during task clean up stage. This causes issue as during task clean up the stdout stream related semaphore is uninitialized in the func lib_stream_release , but the same sem is used later for dbg logs. Thus, this results in assert failure in lib_filesem.c as the error number we get when taking sem on an invalid sem is EINVAL

So, use mllvdbg instead of mvdbg

### DIFF
--- a/os/mm/mm_heap/mm_free.c
+++ b/os/mm/mm_heap/mm_free.c
@@ -92,7 +92,7 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
 	FAR struct mm_freenode_s *prev;
 	FAR struct mm_freenode_s *next;
 
-	mvdbg("Freeing %p\n", mem);
+	mllvdbg("Freeing %p\n", mem);
 
 	/* Protect against attempts to free a NULL reference */
 
--- a/os/mm/mm_heap/mm_mallinfo.c
+++ b/os/mm/mm_heap/mm_mallinfo.c
@@ -113,7 +113,7 @@ int mm_mallinfo(FAR struct mm_heap_s *heap, FAR struct mallinfo *info)
 		mm_takesemaphore(heap);
 
 		for (node = heap->mm_heapstart[region]; node < heap->mm_heapend[region]; node = (struct mm_allocnode_s *)((char *)node + node->size)) {
-			mvdbg("region=%d node=%p size=%u preceding=%u (%c)\n", region, node, node->size,
+			mllvdbg("region=%d node=%p size=%u preceding=%u (%c)\n", region, node, node->size,
 				(node->preceding & ~MM_ALLOC_BIT), (node->preceding & MM_ALLOC_BIT) ? 'A' : 'F');
 			ASSERT(node->size);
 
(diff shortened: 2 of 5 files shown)


sunghan-chang:
@abhishek-samsung How about log mixed with lldvg?

abhishek-samsung:
> @abhishek-samsung How about log mixed with lldvg?

Hello @sunghan-chang , mixed logs can happen because of this. But this will avoid the crash (caused by assert). Will discuss this with my team and will update you if we find a different approach.

abhishek-samsung:
A bit more context, this issue was introduced by the PR https://github.com/Samsung/TizenRT/pull/5742 , before this there was no dependency on stdout file sem for dbg.

ewoodev:
@abhishek-samsung How about remove dependency stdout sem and add syslog buffer for only dbg. 
That way #5742 this log mix issue can be solved by uart sem.

The uart write is getting argument buffer and buffer size and take sem unit buffer. but dbg has not buffer, so I think, occerd log mix issue.
Therefore, if add dbg buffer, we can seperate printf(stdout) and dbg.

abhishek-samsung:
> @abhishek-samsung How about remove dependency stdout sem and add syslog buffer for only dbg. That way #5742 this log mix issue can be solved by uart sem.
> 
> The uart write is getting argument buffer and buffer size and takes sem per-buffer. but dbg has not buffer, so I think, occerd log mix issue. Therefore, if add dbg buffer, we can seperate printf(stdout) and dbg.

I think the initial design for dbg was to not use any buffering. 

https://github.com/Samsung/TizenRT/blob/0f974de4254efaa7b986a3a6b7f9659c3a2b065e/os/include/debug.h#L95-L97

also, do you mean we will need to maintain two different buffers? Again we need some kind of synchronization between them when outputting right?

sunghan-chang:
@abhishek-samsung If dbg does not cause log mixed with printf log, then we don't need PR #5742 and any buffer. What we want is just one. No log mixed with printf and dbg. For this, we applied the same sem for both.

sunghan-chang:
As I remember, initially dbg and printf do not mix logs. `dbg` is just debugging message so that we don't need to use different path to print logs. Only `lldbg` is necessary for different path at specific state (assert or before uart console init)
I think we can change like this
```
#ifdef config_debug_error
#define dbg() printf()
#else
#define dbg()
#endif

#ifdef config_debug_verbose
#define vdbg() printf()
#else
#define vdbg()
#endif

#ifdef config_debug_info....
....
```

ewoodev:
This chage need prerepuisites as below

Do not use dbg case of after called sched_lock() and irqsave(). 
   lib_printf can be context changed. but dbg is used in scheduling code, so that this also must be chaged.

If we will change, please also refer to this

abhishek-samsung:
@sunghan-chang and @ewoodev, my team is looking into this. Please give us some time to do the analysis.

os/mm/mm_heap/mm_free.c
```diff
@@ -92,7 +92,7 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
 	FAR struct mm_freenode_s *prev;
 	FAR struct mm_freenode_s *next;
 
-	mvdbg("Freeing %p\n", mem);
+	mllvdbg("Freeing %p\n", mem);
```
gSahitya-samsung, line 95: Changes look fine

neel-samsung:
Please resolve the conflicts if needed.

PR #5974, opened 2023-11-01
Update libc->math using nuttx's commit to fix libc

sahil1dotgupta:
Issue fixed: Libc caused build issues in libcxx.
libcxx/math.h functions' definitions are not found due to issue in old libc/math.h, which has definitions of same functions in form of macros.

This commit by Nuttx changes the macros to function declarations and defines them in separate files. Nuttx commit id: https://github.com/apache/nuttx/commit/df812d09a219495a8ba1d0c30a2e9ebc05c37f3b Commit message by nuttx:

math: Make this friendly with libcxx
- Turn some macros into functions

- Implement some type-agnostic functions. (Just use __builtin_xxx)

- Add some missing function prototypes (Just prototypes, not actually implemented in this commit)

### DIFF
--- a/lib/libc/math/Make.defs
+++ b/lib/libc/math/Make.defs
@@ -81,6 +81,11 @@ CSRCS += lib_tanl.c lib_tanhl.c lib_asinhl.c lib_acoshl.c lib_atanhl.c lib_erfl.
 CSRCS += lib_copysignl.c lib_truncl.c lib_nextafterl.c lib_nexttowardl.c lib_remainderl.c
 CSRCS += lib_remquol.c
 
+CSRCS += lib_erfc.c lib_erfcf.c lib_erfcl.c
+CSRCS += lib_expm1.c lib_expm1f.c lib_expm1l.c
+CSRCS += lib_lround.c lib_lroundf.c lib_lroundl.c
+CSRCS += lib_llround.c lib_llroundf.c lib_llroundl.c
+CSRCS += lib_nan.c lib_nanf.c lib_nanl.c
 CSRCS += __cos.c __sin.c lib_gamma.c lib_lgamma.c
 
 CSRCS += lib_libexpi.c lib_libsqrtapprox.c
--- a/lib/libc/math/lib_erfc.c
+++ b/lib/libc/math/lib_erfc.c
@@ -1,4 +1,21 @@
 /****************************************************************************
+ *
+ * Copyright 2023 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
  * lib/math/lib_erfc.c
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more
--- a/lib/libc/math/lib_erfc.c
+++ b/lib/libc/math/lib_erfc.c
@@ -50,6 +50,6 @@
 #ifdef CONFIG_HAVE_DOUBLE
 double erfc(double x)
 {
-  return 1 - erf(x);
+	return 1 - erf(x);
 }
 #endif
--- a/lib/libc/math/lib_erfc.c
+++ b/lib/libc/math/lib_erfc.c
@@ -1,0 +1,38 @@
+/****************************************************************************
+ * lib/math/lib_erfc.c
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.  The
+ * ASF licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the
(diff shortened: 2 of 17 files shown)


jeongarmy:
Please revert changed file modes from 755 to 644.

lib/libc/math/lib_erfc.c
```diff
@@ -0,0 +1,39 @@
+/****************************************************************************
```
jeongarmy, line 1: We added samsung license to all files in lib directory. Could you add it with 2023 year for new files you added?

 *
 * Copyright 2023 Samsung Electronics All Rights Reserved.

lib/libc/math/Make.defs
```diff
@@ -81,9 +81,14 @@ CSRCS += lib_tanl.c lib_tanhl.c lib_asinhl.c lib_acoshl.c lib_atanhl.c lib_erfl.
 CSRCS += lib_copysignl.c lib_truncl.c lib_nextafterl.c lib_nexttowardl.c lib_remainderl.c
 CSRCS += lib_remquol.c
 
+CSRCS += lib_libexpi.c lib_libsqrtapprox.c
```
sunghan-chang, line 84: Actually this line is not modified, but because of location change, it looks it gets modified. If there is no specific reason to move, how about keep this as previous?

lib/libc/math/lib_erfc.c
```diff
@@ -0,0 +1,39 @@
...
+#include <math.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+#ifdef CONFIG_HAVE_DOUBLE
+double erfc(double x)
+{
+  return 1 - erf(x);
```
sunghan-chang, line 37: We should apply TizenRT coding rule for libc. Only external, arch and bsp can have own coding rules.

lib/libc/math/lib_erfcf.c
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <tinyara/compiler.h>
```
sunghan-chang, line 26: Both looks unnecessary. Could you check?

lib/libc/math/lib_erfc.c
sahil1dotgupta: Done, please check.

vibhor-m (commented):
Source commit has changes done in libcxx/math.h. It modified nuttx/lib/math.h 
Commit message mentioned libcxx/math.h functions cannot be found
math.h changes are all additions except signbit

Shouldn't these changes be done in libcxx/math.h instead of libc/math.h?

lib/libc/math/lib_erfcf.c
vibhor-m: Lets fix suggestions in newly added code.

lib/libc/math/lib_erfc.c
sahil1dotgupta: Done

lib/libc/math/lib_erfcf.c
sahil1dotgupta: Done

lib/libc/math/lib_erfc.c
```diff
@@ -0,0 +1,38 @@
...
+ * Included Files
+ ****************************************************************************/
+
+#include <math.h>
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+#ifdef CONFIG_HAVE_DOUBLE
```
sunghan-chang, line 33: This file needs to include <tinyara/config.h> for this line.
What I left before does not have `#ifdef CONFIG_XX`. That's why I said they are unnecessary.

lib/libc/math/lib_erfc.c
sunghan-chang: You changed Nuttx license to Samsung one. But if you get this from Nuttx, you must keep Nuttx license. And then you should add the Samsung license at the top.
````
samsung license
nuttx license
codes
```

lib/libc/math/lib_libexpif.c
```diff
@@ -0,0 +1,97 @@
...
+  (float)M_E64   /* e^64 */
+};
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+float lib_expif(size_t n)
+{
+  size_t i;
```
sunghan-chang, line 75: You don't apply TizenRT coding rules here. Let's do that here as well.

lib/libc/math/lib_erfc.c
sunghan-chang: CONFIG_HAVE_DOUBLE kinds of configs are in compiler.h. Let's keep original including headers.
I mean,
if there are `#ifdef config_have_double`, keep original including.
if there is no one, remove.

lib/libc/math/lib_erfc.c
sahil1dotgupta: Correction done.

lib/libc/math/lib_erfc.c
sahil1dotgupta: Okay, done.

sunghan-chang:
@sahil1dotgupta Let's follow TizenRT commit rules. 
```
lib/libc: add math APIs

Issue fixed: Libc caused build issues in libcxx.
libcxx/math.h functions' definitions are not found due to issue in old libc/math.h, which has definitions of same functions in form of macros.

This commit by Nuttx changes the macros to function declarations and defines them in separate files. Nuttx commit id: https://github.com/apache/nuttx/commit/df812d09a219495a8ba1d0c30a2e9ebc05c37f3b Commit message by nuttx:

math: Make this friendly with libcxx

Turn some macros into functions

Implement some type-agnostic functions. (Just use __builtin_xxx)

Add some missing function prototypes (Just prototypes, not actually implemented in this commit)
```

sunghan-chang (approved):

PR #5979, opened 2023-11-03
[do not merge]armv7-a/arm_assert: add to check kmm heap in case os is secure state

ewoodev:
If os secure state(security level high), assertion logs is not printed. 
But regardless of this, the kernel heap should be checked.

Therefore, Add kernel heap check, if os is secure state

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -571,7 +571,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	size_t kernel_assert_location = 0;
 	ARCH_GET_RET_ADDRESS(kernel_assert_location)
 	struct tcb_s *fault_tcb = this_task();
-
+	bool is_kmm_corruption = false;
 
 	board_autoled_on(LED_ASSERTION);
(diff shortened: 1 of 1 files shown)


ewoodev:
We need to have a common structure for up_assert in arm.
let me change in this PR.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -607,14 +607,18 @@ void up_assert(const uint8_t *filename, int lineno)
 	 */
 	if (!IS_SECURE_STATE()) {
 		print_assert_detail(filename, lineno, fault_tcb, asserted_location);
+	} else {
+		if (mm_check_heap_corruption(g_kmmheap) != OK) {
```
kishore-sn, line 611: If we call mm_check_heap_corruption() here, then kernel heap details might get printed even if SECURE_STATE is present. Is this ok?

ewoodev: Thank you for review.

I added checking secure_state in mfdbg and heap_dbg.
https://github.com/Samsung/TizenRT/commit/0f974de4254efaa7b986a3a6b7f9659c3a2b065e

So It is not printed.

PR #5980, opened 2023-11-06
os/arch/arm/src/ameba: GPIO pull up/down fix revert

kaikaikaime:
GPIO pull direction works but induce UART initialisation delay
Revert and will seek alternative solution


sunghan-chang:
@kaikaikaime This is what I asked in previous commit. At that time, you said, removal is correct. Is this really correct modification?
And what I asked is removing UART pins from gpio table. But this does not include it. Could you check this?

sunghan-chang:
@kaikaikaime minor comment.
When you fill in the commit, you should add one empty line between title and description. If not, description is following a title.
```
arch/arm/ameba: title

description is here after an empty line
```

kaikaikaime:
Hi @sunghan-chang,

I tried with your command ` git log --oneline ` , the title is separated from the description. I edited once more, let me know if it is still happening on your side.

kaikaikaime:
I will close this PR as it seems the PA12 init in rtl8721csm_boot.c is causing conflict with the UART. I will raise another PR with just PA12 init removed.

PR #5958, opened 2023-10-17
Enable SMP for ARMV7A

kishore-sn:
Enable SMP for ARMV7A

Verification :
Config: sabre-6quad/hello_smp
SMP test app ported from nuttx for verification.

```
qemu-system-arm -nographic -machine sabrelite -chardev stdio,id=con,mux=on -smp cores=4 -serial chardev:con -mon chardev=con,mode=readline -kernel tinyara.axf_smp 
up_allocate_kheap: start = 0x1081e400 size = 16653312
System Information:
	Version:
		Platform: 3.1	Binary: 200204
	Commit Hash: 6bcf11f
	Build User: root@kishore-VirtualBox
	Build Time: 2023-10-17 11:26:30 IST
	System Time: 06 Dec 2011, 00:00:00 [s] UTC 
Hello, World!!
TASH>>smp
TASH>>  Main[0]: Running on CPU1
  Main[0]: Initializing barrier
Thread[1]: Started
Thread[1]: Running on CPU0
  Main[0]: Thread 1 created
Thread[2]: Started
Thread[2]: Running on CPU1
  Main[0]: Thread 2 created
  Main[0]: Now running on CPU2
Thread[3]: Started
Thread[3]: Running on CPU2
  Main[0]: Thread 3 created
  Main[0]: Now running on CPU3
Thread[4]: Started
Thread[4]: Running on CPU3
  Main[0]: Thread 4 created
  Main[0]: Now running on CPU0
Thread[5]: Started
Thread[5]: Running on CPU3
Thread[4]: Now running on CPU1
  Main[0]: Thread 5 created
  Main[0]: Now running on CPU2
Thread[2]: Now running on CPU2
Thread[3]: Now running on CPU1
Thread[6]: Started
Thread[6]: Running on CPU3
Thread[4]: Now running on CPU0
Thread[5]: Now running on CPU2
  Main[0]: Thread 6 created
  Main[0]: Now running on CPU3
Thread[1]: Now running on CPU3
Thread[2]: Now running on CPU1
Thread[6]: Now running on CPU2
Thread[7]: Started
Thread[7]: Running on CPU0
Thread[3]: Now running on CPU3
Thread[5]: Now running on CPU1
Thread[4]: Now running on CPU2
  Main[0]: Thread 7 created
  Main[0]: Now running on CPU0
Thread[1]: Now running on CPU0
Thread[2]: Now running on CPU3
Thread[6]: Now running on CPU1
Thread[7]: Now running on CPU2
[...]

PR #5985, opened 2023-11-10
armv7-a/arm_assert: fix board reset in assertion

ewoodev:
If CONFIG_BOARD_ASSERT_AUTORESET is enabled and occur crash(assertion), board will be rebooted. 
But, rebooting does not work with armv7-a, because CONFIG_BOARD_ASSERT_AUTORESET is not applied in armv7-a/arm_assert.c yet.

Therefore, Fix so that it reboots correctly when an assertion occurs.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -415,12 +415,13 @@ static void up_dumpstate(struct tcb_s *fault_tcb, uint32_t asserted_location)
 
 static void arm_assert(void)
 {
+#ifdef CONFIG_BOARD_ASSERT_AUTORESET
+	boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+#else
+#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
 	/* Are we in an interrupt handler or the idle task? */
 
-	if (CURRENT_REGS || (this_task())->flink == NULL)
-	{
-#if CONFIG_BOARD_RESET_ON_ASSERT >= 1
-		board_reset(CONFIG_BOARD_ASSERT_RESET_VALUE);
+	if (CURRENT_REGS || (this_task())->flink == NULL) {
 #endif
 		/* Disable interrupts on this CPU */
 		irqsave();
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -415,12 +415,13 @@ static void up_dumpstate(struct tcb_s *fault_tcb, uint32_t asserted_location)
...
+	boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+#else
+#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
 	/* Are we in an interrupt handler or the idle task? */
 
-	if (CURRENT_REGS || (this_task())->flink == NULL)
-	{
-#if CONFIG_BOARD_RESET_ON_ASSERT >= 1
-		board_reset(CONFIG_BOARD_ASSERT_RESET_VALUE);
+	if (CURRENT_REGS || (this_task())->flink == NULL) {
```
jeongarmy, line 424: How about changing the condition for idle task to `(this_task())->pid == 0`?

ewoodev: Sure, I changed,

abhishek-samsung: according to the SMP design that is going to be added in https://github.com/Samsung/TizenRT/pull/5958, there will be an idle thread for each cpu. Therefore only checking ```pid == 0``` is not enough in the case of SMP. Incase that PR merges before this, this change needs to be reverted.

ewoodev: I understand, thank you, changed,

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -440,14 +441,12 @@ static void arm_assert(void)
 		up_mdelay(250);
 #endif
 		}
+#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
+	} else {
```
kishore-sn, line 445: This needs to be like below:
```
       if(xxx) {
       }
#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
	else {
		exit(errorcode);
	}
#endif
```

ewoodev: Thank you for review~

Please look below codes,
The start conditional statement is also disabled by config.
So, It don't need changes.

```
#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
	/* Are we in an interrupt handler or the idle task? */

	if (CURRENT_REGS || (this_task())->flink == NULL) {
#endif
```

```
#ifndef CONFIG_BOARD_ASSERT_SYSTEM_HALT
	} else {
		exit(errorcode);
	}
#endif
```

sunghan-chang: @kishore-sn If it's ok, could you leave an approval?

sunghan-chang: @kishore-sn Could you check this?

sunghan-chang (approved):

PR #5986, opened 2023-11-15
os/board/rtl8730e: Change alignment of malloc

jwei5:
- Use kmm_memalign instead of kmm_malloc in RTK malloc operations to align malloc to 64b to resolve cache operation issues on CA32
- Fix DCache_Clean for ssid array during wifi scan to ensure cache is cleaned properly for cache coherency

### DIFF
--- a/os/board/rtl8730e/src/component/os_dep/osdep_service_memory.c
+++ b/os/board/rtl8730e/src/component/os_dep/osdep_service_memory.c
@@ -6,15 +6,17 @@
 #include <osdep_service.h>
 #include <stdio.h>
 
+/* For Smart, need to align malloc to 64bytes (cache line size of AP) for cache operations */
+int align = 64;
+
 void *rtw_vmalloc(u32 sz)
 {
 	void *pbuf = NULL;
-	pbuf = kmm_malloc(sz);
+	pbuf = kmm_memalign(align, sz);
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-        if (pbuf)
-        {
-            DEBUG_SET_CALLER_ADDR(pbuf);
-        }
+	if (pbuf){
+		DEBUG_SET_CALLER_ADDR(pbuf);
+	}
 #endif
 	return pbuf;
 }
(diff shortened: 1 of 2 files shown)


os/board/rtl8730e/src/component/os_dep/osdep_service_memory.c
```diff
@@ -6,15 +6,21 @@
 #include <osdep_service.h>
 #include <stdio.h>
 
+/* For Smart, need to align malloc to 64bytes (cache line size of AP) for cache operations */
+#if CONFIG_AMEBAD2
```
sunghan-chang, line 10: Could you let me know where this config is defined?

jwei5: Hi, this config is defined in RTK's [platform_autoconf.h](https://github.com/Samsung/TizenRT/blob/master/os/board/rtl8730e/src/project/realtek_amebaD2_va0_example/inc/inc_ap/platform_autoconf.h)

sunghan-chang: 1. The title says "RTL8730E". Is that `AMEBAD2`? not `AMEBASMART`?
2. You define a proper config and don't define other chip. In other word,
````
#define CONFIG_AMEBAD2 1
// no define for CONFIG_AMEBALITE
```
Then, you should use `#ifdef` and `#elif defined` instead of `#if` and `#elif`.
If you use like below
```
#define CONFIG_AMEBAD2 1
#define CONFIG_AMEBALITE 0
```
Then, current is correct.

jwei5: Hi @sunghan-chang,  since the file is in the rtl8730e path, I will remove the condition for checking the chip as it would be unnecessary to check.

sunghan-chang (approved):

sunghan-chang:
@jwei5 Hello.
Is this necessary to merge before new SDK porting?

jwei5:
Hi @sunghan-chang ,
This change will also be needed in the new SDK porting. If you prefer to merge it in the new SDK porting, we can also do so

sunghan-chang:
@jwei5 If it's not conflict with new SDK porting, let me merge now.

jwei5:
Hi @sunghan-chang , this change will not have a conflict with SDK porting, thank you.

PR #5987, opened 2023-11-16
os/board/rtl8721csm: add feature for selecting wpa mode

jwei5:
- Add feature to select the wpa mode to use to connect to APs in mixed security mode based on security type indicated by user


hk-gwak (approved):
Looks good to me.

PR #5989, opened 2023-11-17
os/fs/smartfs: Skip smartfs_sector_recovery in Nxfuse image build

amogh-samsung:
- smartfs_sector_recovery() need not be invoked during Nxfuse image build because we are building a fresh image to flash to the partition and hence recovery is not required.
- Additionally, in the case of Nxfuse host build, we will always get the error message "CRITICAL BUG!! Root sector is gone!!" from smartfs_sector_recovery despite successful completion.

### DIFF
--- a/os/fs/smartfs/smartfs_smart.c
+++ b/os/fs/smartfs/smartfs_smart.c
@@ -1274,11 +1274,13 @@ static int smartfs_bind(FAR struct inode *blkdriver, const void *data, void **ha
 	}
 
 	*handle = (void *)fs;
-	
+
+#ifndef NXFUSE_HOST_BUILD
 	ret = smartfs_sector_recovery(fs);
 	if (ret != 0) {
 		goto error_with_semaphore;
 	}
+#endif
 
 	smartfs_semgive(fs);
 	return ret;


amogh-samsung:
```
Formatting /root/tizenrt/tools/nxfuse/../../build/output/bin/rtl8730e_smartfs.bin with smartfs format
SMART Scan
   Erase size:                 4096
   Erase block:                 128
   Sect/block:                    4
   MTD Blk/Sect:                  4
   avail sect/block:              4
   Data Erase block:            122
   Journal Erase block:           6
   Journal Data/Block:          315
   Journal total:              1890
   Journal usage:                 0
	    SMARTFS:
	    Sector size:     1024
	    Bytes/sector     1018
	    Num sectors:     488
	    Free sectors:    479
	    Max filename:    32
	    RootDirSector:   3
Format successful
mount binary
Copy files
cp: cannot stat '/root/tizenrt/tools/nxfuse/../../tools/fs/contents-smartfs/rtl8730e/base-files/*': No such file or directory
DONE
```

Taejun-Kwon (approved):

PR #5992, opened 2023-11-28
Revert "os/binfmt: Modify caching method for symtab and reltab"

abhishek-samsung:
This reverts commit 5ae1088fbf0250eda3e8bf18907707b1ded9d659. ELF caching doesnt not work with this commit applied. Hence reverthing this commit. Will recheck and apply it in a later patch


build/configs/rtl8721csm/loadable_apps/defconfig
```diff
@@ -1195,9 +1195,7 @@ CONFIG_ELF_STACKSIZE=2048
 CONFIG_ELF_BUFFERSIZE=32
 CONFIG_ELF_BUFFERINCR=32
 CONFIG_ELF_EXCLUDE_SYMBOLS=y
-CONFIG_ELF_CACHE_READ=y
-CONFIG_ELF_CACHE_BLOCK_SIZE=2048
-CONFIG_ELF_CACHE_BLOCKS_COUNT=60
+# CONFIG_ELF_CACHE_READ is not set
```
sunghan-chang, line 1198: We should keep this enabled. If this is a part of the commit reverted, then let's add a new commit to keep this enabled.

abhishek-samsung: Done.

sunghan-chang (approved):

PR #5993, opened 2023-11-29
tools/check_pacakage_size.py: Declare 'global' keyword once

jeongarmy:
In python, 'global' keyword is needed for usage of global variable. It is enough to declare 'global' keyword once in a function because 'if' context doesn't make new scope. With this change, the below build warning is fixed.

========== Verification of partition settings ========== => Partition verification SUCCESS!! The setting of all partitions is OK. ./tools/../tools/check_package_size.py:72: SyntaxWarning: name 'FAIL_TO_BUILD' is assigned to before global declaration
  global FAIL_TO_BUILD


ewoodev (approved):
lgtm~

sunghan-chang (approved):

PR #5996, opened 2023-11-30
os/arch/arm/src/amebasmart/Kconfig: Turn on SPI1 smart by default

kaikaikaime:
Set SPI1 Smart Kconfig to default yes

### DIFF
--- a/os/arch/arm/src/amebasmart/Kconfig
+++ b/os/arch/arm/src/amebasmart/Kconfig
@@ -76,7 +76,7 @@ config AMEBASMART_SPI0
 	default n
 config AMEBASMART_SPI1
 	bool "SPI1"
-	default n
+	default y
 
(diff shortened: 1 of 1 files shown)


sunghan-chang (commented):
@kaikaikaime Sorry for miscommunication. What I asked is defconfig modification. Because CONFIG_AMEBASMART_SPI1 is not enabled in current defconfig, we can't use SPI1 now.

Taejun-Kwon:
No we should use both of spi1 & spi0

kaikaikaime:
Hi @sunghan-chang ,

I would like to confirm when you mention defconfig, do you meant only hello, or all the defconfig in smart?

Kai

sunghan-chang:
@kaikaikaime all the defconfig in smart

PR #6002, opened 2023-12-05
os/arch/arm/amebasmart: Change peripheral config to board level config.

ewoodev:
amebasmart_i2c.c have dependency CONFIG_AMEBASMART_I2C_XXXX configuration
So, when deciding whether to include this file in build, we should check CONFIG_AMEBASMART_I2C, not CONFIG_I2C. 
The same goes for amebasmart_spi.c.

Therefore, Change CONFIG_I2C and CONFIG_SPI to each CONFIG_AMEBASMART_I2C and CONFIG_AMEBASMART_SPI.


jeongarmy (approved):
LGTM

sunghan-chang (approved):

sunghan-chang (approved):

PR #6005, opened 2023-12-14
os/board/rtl8730e: Port 2 adapters for I2S

kaikaikaime:
Fix the I2S tx handler


os/arch/arm/src/amebasmart/amebasmart_i2s.h
```diff
@@ -52,7 +52,9 @@ extern "C" {
 */
 typedef enum {
 	I2S_NUM_0 = 0x0,			/*!< I2S 0 */
-	//I2S_NUM_1 = 0x1,			/*!< I2S 1 */
+	I2S_NUM_1 = 0x1,			/*!< I2S 1 */
```
sunghan-chang, line 55: You removed I2S0 and I2S1 from the Kconfig. Is this necessary to leave them?

sunghan-chang (commented):
This single commit has two different topics, audio and i2s. Could you split them as different commits?

os/arch/arm/src/amebasmart/amebasmart_i2s.h
kaikaikaime: The current adaptor index is aligned with the MAX adaptors, this is done to prevent confusion in the future. Would this be fine with you?

sunghan-chang: Do you mean that I2S0 and I2S1 will be added later?

kaikaikaime: We have 4 adaptors, I2S0 and I2S1 are internal I2S without GPIO pins which currently will not used based on the current requirement. If there is a need, then we will add.  Do you prefer to remove instead?

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -391,9 +405,11 @@ static int amebasmart_i2s_tx(struct amebasmart_i2s_s *priv, struct amebasmart_bu
 		}
 		apb->curbyte += tx_size; /* No padding, ptx_buf is big enough to fill the whole tx_size */
 
+		i2s_enable(&priv->i2s_object);
+		lldbg("i2s send page called\n");
```
abhishek-samsung, line 409: Please remove this line, I think it was added for debugging.

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -130,26 +130,17 @@ config AMEBASMART_I2CTIMEOTICKS
...
 
 config AMEBASMART_I2S
-	bool "Enable Amebasmart I2S"
+	bool "Amebasmart I2S"
 	default n
 
 if AMEBASMART_I2S
-config AMEBASMART_I2S0
-	bool "I2S_0"
+config AMEBASMART_I2S2
```
jeongarmy, line 137: What does the number after AMEBASMART_I2S? 0 and 1 is not nessasary?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -271,65 +221,64 @@ void i2s_set_dma_buffer(i2s_t *obj, char *tx_buf, char *rx_buf,
 
 	if (obj->direction == I2S_DIR_TX) {
```
jeongarmy, line 223: I think `obj` and `i2s_idx` should be checked ahead because it is assigned by parameter.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -260,8 +210,8 @@ void i2s_init(i2s_t *obj, PinName sck, PinName ws, PinName sd_tx, PinName sd_rx,
 void i2s_set_dma_buffer(i2s_t *obj, char *tx_buf, char *rx_buf,
 						uint32_t page_num, uint32_t page_size)
 {
-	//uint8_t i2s_idx = obj->i2s_idx;
-	u32 i, j;
+	uint8_t i2s_idx = obj->i2s_idx;
```
jeongarmy, line 213: `i2c_idx` is used as index of array. I think `obj` and `i2s_idx` should be checked ahead to avoid invalid memory access because it is assigned by parameter.

os/arch/arm/src/amebasmart/Kconfig
sunghan-chang: Please refer his previous reply.
```
 [kaikaikaime]
We have 4 adaptors, I2S0 and I2S1 are internal I2S without GPIO pins which currently will not used based on the current requirement. If there is a need, then we will add. Do you prefer to remove instead?
```

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -345,13 +294,17 @@ void i2s_set_dma_buffer(i2s_t *obj, char *tx_buf, char *rx_buf,
   */
 void i2s_tx_irq_handler(i2s_t *obj, i2s_irq_handler handler, uint32_t id)
 {
-	(void) obj;
-	(void) handler;
-	(void) id;
+	uint8_t i2s_index = obj->i2s_idx;
```
jeongarmy, line 297: Please add null checking variables

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -391,9 +405,10 @@ static int amebasmart_i2s_tx(struct amebasmart_i2s_s *priv, struct amebasmart_bu
 		}
 		apb->curbyte += tx_size; /* No padding, ptx_buf is big enough to fill the whole tx_size */
 
+		i2s_enable(&priv->i2s_object);
 		i2s_send_page(&priv->i2s_object, (uint32_t *)ptx_buf);
 	}
-	return OK;
+	return ret;
```
jeongarmy, line 411: It always returns OK. It it ok to return OK even `bfcontainer` or `bfcontainer->apb` is NULL?

os/arch/arm/src/amebasmart/amebasmart_i2s.h
Taejun-Kwon: if there are 4 i2s ports, it should be defined. By KConfig or something, some of those should be selected.

kaikaikaime: There are reasons for this, the internal I2S will not be able to use TX/RX as I2S0 & I2S1 externally as there are no external pins and used internally for our audio Sport. Therefore I2S0 & I2S1 cannot be used, this is a limitation. We were hoping to directly link the porting layer of (i2s2) to our driver (i2s2). 
If this is confusing, we can change to porting layer (i2s0) (i2s1) to link our driver layer (i2s2) (i2s3) respectively.

os/arch/arm/src/amebasmart/Kconfig
kaikaikaime: Yes, i2s2 is tx and i2s3 is rx respectively.

os/arch/arm/src/amebasmart/amebasmart_i2s.h
kaikaikaime: Based on discussion, we will remove I2S_NUM_0 & I2S_NUM_1, since it does not impact the code flow in anyway.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
kaikaikaime: I will place another check here.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
kaikaikaime: Added -1 and i2serr message

abhishek-samsung (approved):

Taejun-Kwon (approved):

PR #6015, opened 2024-01-02
os/arch/arm/src, os/board/rtl8730e: Power Save for RTL8730E

edwakuwaku:
- Tested with running on XIP
- Verified interrupt source: WIFI keepalive (IPC), BT Adv, MICOM UART, Timer interrupt
- Verified with application control through procfs layer
- Next step: Prepare to integrate with product code


os/arch/arm/src/amebasmart/ameba_pmc.c
```diff
@@ -0,0 +1,242 @@
+#ifndef CONFIG_PLATFORM_TIZENRT_OS
```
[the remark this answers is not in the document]
amogh-samsung: And a copyright header would be required here too?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+	irqstate_t flags;
+	int ret;
+
+	/* Decide, which power saving level can be obtained */
+	newstate = pm_checkstate(PM_IDLE_DOMAIN);
+
+	/* Check for state changes */
+	if (newstate != oldstate)
+	{
+		//TODO: Critical section code needed for SMP case?
```
amogh-samsung, line 64: You can remove any such additional comments that I had added initially while pushing the first skeleton.
If we are pushing PM code without completely handling SMP then we can let this TODO be, but other comment lines below this should still be resolved and removed.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+		//Any additional implications of putting a core in critical section while trying to sleep?
```
amogh-samsung, line 65: These implications should be analysed, resolved and this comment removed from here.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+				break;
+			case PM_SLEEP:
+				ap_timer_helper();
+				if (up_cpu_index() == 0) {
+					/* mask sys tick interrupt*/
+					arm_arch_timer_int_mask(1);
+					up_timer_disable();
+					flags = irqsave();
+					if (tizenrt_ready_to_sleep()) {
+// Consider for dual core condition
```
amogh-samsung, line 95: This was another comment that I added as part of the skeleton, this should be resolved and removed from here.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+						} else {
+							if (!check_wfi_state(1)) {
+								goto EXIT;
+							}
+						}
+#endif
+						// Interrupt source from BT/UART will wake cpu up, just leave expected idle time as 0
+						// Enter sleep mode for AP
+						configPRE_SLEEP_PROCESSING(xModifiableIdleTime);
+						/* When wake from pg, arm timer has been reset, so a new compare value is necessary to
```
amogh-samsung, line 123: Please use 
/*
 *
 */
as the format for multi line comments for readability

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+					up_irq_enable();
+				}
+				// This case is consideration for secondary core
+				else if (up_cpu_index() == 1) {
+					if (pmu_get_sleep_type() == SLEEP_PG) {
+						if (tizenrt_ready_to_sleep()) {
+							/* CPU1 will enter hotplug state. Raise a task yield to migrate its task */
+							pmu_set_secondary_cpu_state(1, CPU1_HOTPLUG);
+							// Check portYIELD();
+							// portYIELD();
```
amogh-samsung, line 152: If these 2 lines are not needed can they be removed please?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+				if (ret < 0) {
+					oldstate = PM_NORMAL;
+				}
+				printf("Wakeup from Sleep!!\n");
+
+				break;
+			default:
+				break;
+		}
+			//TODO: Handle critical section access logic for SMP case in 8730E?
```
amogh-samsung, line 172: Similarly, for this comment, can we analyse, resolve and remove it now?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -47,15 +193,23 @@
...
 
-  asm("WFI");
+	up_idlepm();
 #endif
 }
+
+#ifdef CONFIG_PM
+void arm_pminitialize(void)
+{
+	/* Then initialize the TinyAra power management subsystem proper */
```
amogh-samsung, line 212: "proper" -> "properly" ?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -0,0 +1,79 @@
+/****************************************************************************
```
amogh-samsung, line 1: Samsung copyright header would be required here as well

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1131,6 +1166,178 @@ static bool rtl8730e_up_txempty(struct uart_dev_s *dev)
...
+	(void)expected_idle_time;
+	(void)param;
+	serial_change_clcksrc(sdrv[uart_index_get(g_uart1priv.tx)], 1);
+	return 1;
+}
+#endif
+
+
+/****************************************************************************
+ * Name: amebasmart_serial_pmnotify (DUMMY)
```
amogh-samsung, line 1216: Can remove the (DUMMY) from the description now

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1131,6 +1166,178 @@ static bool rtl8730e_up_txempty(struct uart_dev_s *dev)
...
+ * Returned Value:
+ *   None - The driver already agreed to transition to the low power
+ *   consumption state when when it returned OK to the prepare() call.
+ *
+ *
+ ****************************************************************************/
+
+#ifdef CONFIG_PM
+static void amebasmart_serial_pmnotify(FAR struct pm_callback_s *cb, int domain,
+                                   enum pm_state_e pmstate)
```
amogh-samsung, line 1239: Minor, we can have the function name in one line?

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1131,6 +1166,178 @@ static bool rtl8730e_up_txempty(struct uart_dev_s *dev)
...
+		default:
+			break;
+	}
+
+	return;
+}
+#endif
+
+/****************************************************************************
+ * Name: amebasmart_serial_pmprepare (DUMMY)
```
amogh-samsung, line 1262: Here also, we can remove "DUMMY" from the description

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1131,6 +1166,178 @@ static bool rtl8730e_up_txempty(struct uart_dev_s *dev)
...
+#ifdef CONFIG_RTL8730E_UART2
+			if ((g_uart2priv.txint_enable) && (!serial_writable(sdrv[2]))) {		/* If Tx init enable and FIFO not empty */
+					return ERROR;
+			}
+#endif
+#ifdef CONFIG_RTL8730E_UART4
+			// LOGUART_TypeDef *UARTLOG = LOGUART_DEV;
+			// if (!(UARTLOG->LOGUART_UART_LSR & LOG_UART_IDX_FLAG[2].empty)) {		/* log UART Tx Empty Check */
+			// 	return ERROR;
+			// }
```
amogh-samsung, line 1330: If these lines are not needed, can they be removed?

amogh-samsung: Also @edwakuwaku , I remember that during our testing, UART4 would always reject PM state change because of some reason, is that resolved in this PR

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -319,11 +319,6 @@ void amebasmart_mount_partitions(void)
 #endif /* CONFIG_FLASH_PARTITION */
 }
 
-void amebasmart_memory_initialize(void)
-{
-
-}
-
```
amogh-samsung, line 326: Please tell us why memory_initialize and all its references have been removed for amebasmart?

os/pm/pm_changestate.c
```diff
@@ -105,7 +105,7 @@ static void pm_timer(int domain)
 	static const int pmtick[3] = {
 		TIME_SLICE_TICKS * CONFIG_PM_IDLEENTER_COUNT,
 		TIME_SLICE_TICKS * CONFIG_PM_STANDBYENTER_COUNT,
-		TIME_SLICE_TICKS * CONFIG_PM_SLEEPENTER_COUNT
+		TIME_SLICE_TICKS *CONFIG_PM_SLEEPENTER_COUNT
```
amogh-samsung, line 108: Please undo this change

os/pm/pm_changestate.c
```diff
@@ -114,10 +114,12 @@ static void pm_timer(int domain)
...
 		int delay = pmtick[pdom->state] + pdom->btime - clock_systimer();
-		int left  = wd_gettime(pdom->wdog);
-
+		int left = wd_gettime(pdom->wdog);
+		if (delay <= 0) {
+			delay = 1;
+		}
 		if (!WDOG_ISACTIVE(pdom->wdog) || abs(delay - left) > PM_TIMER_GAP) {
-			wd_start(pdom->wdog, delay, (wdentry_t)pm_timer_cb, 0);
+			wd_start(pdom->wdog, delay, (wdentry_t) pm_timer_cb, 0);
```
amogh-samsung, line 122: Please undo this change

os/arch/arm/src/amebasmart/amebasmart_boot.c
```diff
@@ -359,16 +359,6 @@ void arm_boot(void)
```
amogh-samsung, line 359: I see this commit "os/arch/arm/src, os/board/rtl8730e: Minor fixes" is not related to PM.
Could you push this commit in a separate PR please?
It will help review PM changes better and this commit can be merged easier separately as well.

os/pm/pm_changestate.c
```diff
@@ -114,10 +114,12 @@ static void pm_timer(int domain)
 
 	if (pdom->state < PM_SLEEP && !pdom->stay[pdom->state] && pmtick[pdom->state]) {
 		int delay = pmtick[pdom->state] + pdom->btime - clock_systimer();
```
amogh-samsung, line 116: Delay is logically calculated as: -
Total waiting time in current state before moving to next state - (time already elapsed in current state)
The latter is calculated as (current system time - system time at which current state was started)

If this delay is 0, then that means that the time duration for waiting in current state has already elapsed.
This is possible if: -
-> The value of `TIME_SLICE_TICKS * CONFIG_PM_<STATE>ENTER_COUNT` is too less
In this case, I think we should add a comment to the code to inform the developer.
We will have a practical lower limit to the config mentioned above, which will be determined as we go ahead.
OR
-> We have already stayed in this state for a long time, which is not expected because pm_timer() is called as soon as the state is changed or a state change fails and the timer for the current state is restarted.


My reason for writing this is that when delay<=0, you are manually changing it to delay=1
This starts a watchdog timer for 1 system tick only.
But we should be more concerned about how we got delay<=0, and such cases should be avoided in the first place.

os/pm/pm_procfs.c
```diff
@@ -466,24 +479,58 @@ static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, s
 /****************************************************************************
  * Name: power_unlock_write
  ****************************************************************************/
+#ifdef CONFIG_ARCH_CORTEXA32
+static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer, size_t timer_interval)
+#else
```
amogh-samsung, line 484: I think when we eliminated the need to PM control from APP size, we also eliminated the need for having option to set wakeup timer.
We do not need to provide customizable timer option now. Core will wakeup from interrupt, or from fixed timer set for wifi alive signal.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+					__asm("	DSB");
+					__asm("	WFI");
+					__asm("	ISB");
+					up_irq_enable();
+				}
+				ret = pm_changestate(PM_IDLE_DOMAIN, PM_NORMAL);
+				if (ret < 0) {
+					oldstate = PM_NORMAL;
+				}
+				printf("Wakeup from Sleep!!\n");
```
amogh-samsung, line 166: Need to remove this printf statement

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+
+					flags = irqsave();
+					__asm("	DSB");
+					__asm("	WFI");
+					__asm("	ISB");
+					up_irq_enable();
+				}
+				ret = pm_changestate(PM_IDLE_DOMAIN, PM_NORMAL);
+				if (ret < 0) {
+					oldstate = PM_NORMAL;
```
amogh-samsung, line 164: lets add a TODO comment here please
//TODO: If we are unable to reset the state back to PM_NORMAL then what to do?

os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -106,22 +106,22 @@ static void mmu_set_flags(uint32_t *val, bool ro, bool exec, uint8_t isL1)
 #ifndef CONFIG_ARCH_ROMPGTABLE
 void mmu_l1_setentry(uint32_t paddr, uint32_t vaddr, uint32_t mmuflags)
 {
-  uint32_t *l1table = mmu_l1_pgtable();
-  uint32_t  index   = vaddr >> 20;
+	uint32_t *l1table = mmu_l1_pgtable();
```
ewoodev, line 109: This change is the same.

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -76,59 +76,56 @@
...
-    }
-
-  /* Registers with 8-bits per interrupt */
-
-  for (irq = GIC_IRQ_SPI; irq < nlines; irq += 4)
-    {
-      putreg32(0x80808080, GIC_ICDIPR(irq));   /* SPI priority */
-      putreg32(0x01010101, GIC_ICDIPTR(irq));  /* SPI on CPU0 */
-    }
+	unsigned int nlines = arm_gic_nlines();
```
ewoodev, line 79: This changs is seen to not related to PM.
Please leave this out of the commit

os/board/rtl8730e/src/rtl8730e_boot.c
edwakuwaku: Hi @amogh-samsung ,

That is referenced from imx6 during early bringup stage, but we do not need that, thus it should be removed.
I will filter out the cleanup changes to another PR.

os/pm/pm_procfs.c
edwakuwaku: As per the conversation we had before, SW timer can be an option to choose whether to apply or not, if yes, it will be an extra interrupt signal, if there is no other signal arrive during the time period given. Thus, it should have no conflict even though we don't allow direct state jump to PM_SLEEP.
Could you please elaborate more on "fixed timer set for wifi alive signal"?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+		/* MCU-specific power management logic */
+		switch (newstate) {
+			case PM_NORMAL:
+				break;
+			case PM_IDLE:
+				break;
+			case PM_STANDBY:
+				break;
+			case PM_SLEEP:
+				ap_timer_helper();
```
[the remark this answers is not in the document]
edwakuwaku: If the state is able to fall into PM_SLEEP, it is impossible that the sleep will be blocked again, as the remaining operation is to setup the pre-sleep logic flow (Meaning that the driver state rejection/ application usage protection should be robust, as there is no blocker left after reaching PM_SLEEP, THE ONLY blocker is in SMP, which if secondary core is not in hotplug mode, we will set it to hotplug mode first, and go to sleep at the next cycle, but since this feature is not fully completed yet, we should discuss it at later stage).
Thus the timer can be handled properly as for single core case, I will add in some consideration logic to setup timer for condition with or w/o SMP.

os/arch/arm/src/amebasmart/amebasmart_serial.c
edwakuwaku: Yes, when we checked and tested on this part previously, the logic is correct as there is log need to be printed, so the condition for checking the bit will cause it not able to enter sleep. Thus, these lines should be removed.

os/board/rtl8730e/src/rtl8730e_boot.c
amogh-samsung: Yes please, thankyou

os/arch/arm/src/amebasmart/amebasmart_boot.c
edwakuwaku: Yes, will do

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,160 @@
...
+		//Any additional implications of putting a core in critical section while trying to sleep?
+		/* Perform board-specific, state-dependent logic here */
+	  	pmvdbg("newstate= %d oldstate=%d\n", newstate, oldstate);
+
+		/* Then force the global state change */
+		ret = pm_changestate(PM_IDLE_DOMAIN, newstate);
+		if (ret < 0) {
+			/* The new state change failed, revert to the preceding state */
+			pmdbg("State change failed! Current state = %d, newstate = %d\n", oldstate, newstate);
+			newstate = oldstate;
```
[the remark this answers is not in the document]
edwakuwaku: Agree, exit condition added

os/arch/arm/src/armv7-a/arm_mmu.c
edwakuwaku: Reverted

os/arch/arm/src/armv7-a/arm_gicv2.c
edwakuwaku: Removed~

os/arch/arm/src/amebasmart/amebasmart_idle.c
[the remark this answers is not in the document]
amogh-samsung: Ohkay, I get your point.
Could you add a TODO comment here please to indicate that we have to check some things in the future as you mentioned?

os/pm/pm_procfs.c
edwakuwaku: Hi @amogh-samsung ,

Okay, thanks for reminding this part. But I will still leave the timer interrupt setup, just in case there is some scenario that might require the timer in the future. If the application dont need it, you can just simply provide 0 to the last argument when relaxing the state, so it will not trigger any timer interrupt.

os/arch/arm/src/amebasmart/amebasmart_idle.c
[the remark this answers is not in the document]
edwakuwaku: I have added some TODO comment at below section, I think that part should be relevant to this topic.

os/pm/pm_procfs.c
amogh-samsung: I get your point, it can be a good to have feature.
I have one question though, if multiple applications block and then unblock the same domain, then the timer interval will be set by the last unblocking application right?

edwakuwaku: Yes, you are right~ If there is some other requirement to choose the timer interval from multiple application in the future, we can extend that too.

os/pm/pm_changestate.c
```diff
@@ -276,34 +279,31 @@ int pm_changestate(int domain_indx, enum pm_state_e newstate)
...
+		 * disagreed and the state has been reverted).  Set the new state.
+		 */
+		pm_changeall(domain_indx, newstate);
+		pdom->state = newstate;
 
-		pm_timer(domain_indx);
 	}
-
+EXIT:
+	/* Start PM timer to decrease PM state */
```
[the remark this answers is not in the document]
edwakuwaku: Agree with your consideration here, modification done.

os/arch/arm/src/armv7-a/arm_gicv2.c
edwakuwaku: Hi @ewoodev ,

Please help to check this PR: https://github.com/Samsung/TizenRT/pull/6025

Thank you~

os/pm/pm_procfs.c
amogh-samsung: Ohkay, we'll also take @sunghan-chang 's opinion on this please

os/arch/arm/src/amebasmart/amebasmart_idle.c
amogh-samsung: @edwakuwaku whats your opinion on this?

os/pm/pm_changestate.c
amogh-samsung: @edwakuwaku what's your opinion on this?

os/arch/arm/src/amebasmart/amebasmart_idle.c
edwakuwaku: I couldn't think of any scenario that will cause fail in changing state here, I suggest that we should run with application first, and check whether your concern is going to happen.

os/pm/pm_changestate.c
edwakuwaku: Regarding the first point, I agree with you that we should add in some kind of lower bound for the state drop counter.
For the second point, could you please share the scenario that might cause staying a state for a long time? As far as I understand, this should not happen according to the PM driver realization.
This change is to prevent giving negative input to wdog timer (referenced from NUTTX), I'm not sure this is added because they encountered problem similar as to your concerned, or just a basic protection. Anyway, I am on the same page with you regarding this concern, and due to short timespan, I think we should proceed to conduct test on application first, and get back to review this point if we encounter any problem related to this logic.

amogh-samsung: For the second point, as I mentioned, it is not expected.
However, developers face errors only when something unexpected happens. We should have safeguards in place for such unexpected cases.
Please add a debug statement here (pmdbg) to print log if delay<=0 is observed
Then, we will continue to manually reset delay to 1 to pass to wdog function as you have implemented.

If we face such cases and they cannot be resolved by increasing wait duration, we will check what to do at that time.

os/arch/arm/src/amebasmart/amebasmart_idle.c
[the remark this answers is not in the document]
amogh-samsung: I saw the comment below for SMP case and timer.
However, I still request you to add a short TODO comment here for tracking purposes please.

os/arch/arm/src/amebasmart/amebasmart_idle.c
edwakuwaku: Yes, I agree with you that unexpected things happen where there exists some corner cases we might misconsider it. Lets follow your suggestion first to enhance the effectiveness when debugging.
May I add todo comment and the PM log from some other discussion thread together in a new commit?

amogh-samsung: I understand it can be very irritating to rebase and make changes to lower-stacked commits sometimes but modularity of commits is very important for us for easy look-back and reference in the future.
Hence we requested you to separate non-PM changes in a separate PR and for the same reason, I request you to add all debug statement and TODO comments in the same commits itself please.

edwakuwaku: Sure no problem~

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -23,14 +23,181 @@
...
+				}
+
+/* TODO: This exit is for secondary core
+   After 2nd core entered hotplug mode, TizenRT should remove the idle task for 2nd core
+   Revisit here to see if anything else need to be done */
+#ifdef CONFIG_SMP
+EXIT2:
+#endif
+/* If state transition is rejected, exit directly*/
+REJECTED:
```
amogh-samsung, line 190: I think this REJECTED label should be outside the switch case, at the end of the up_idlepm function

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,119 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
...
 {
 	int state = atoi(buffer);
+	(void)buflen;
 
 	if (state >= PM_NORMAL && state < PM_SLEEP) {
-		pm_stay(0, state);
+		pm_stay(PM_IDLE_DOMAIN, state);
 		fvdbg("power locked!\n");
 	} else {
+		return 1;
```
amogh-samsung, line 486: Is it right to return 1 when invalid state is passed as input to this function?

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,119 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
...
+*   buflen (NOT USED) - Length of the buffer
+*
+* 	Returned Value:
+*   0 (OK) means the state lock was successful.
+*	1 (Failed) means the input state was incorrect.
  ****************************************************************************/
 static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
 {
 	int state = atoi(buffer);
+	(void)buflen;
```
amogh-samsung, line 480: Do we still need this if were not returning buflen anymore?

os/pm/pm_procfs.c
edwakuwaku: It depends to our design, since our return value has 0/1, from the upper layer, we can add something like
`if (ret > 0) {....}` which means it has failed due to wrong input.

os/pm/pm_procfs.c
edwakuwaku: Since the input argument is not used, it should be voided to prevent from unused variable warning.
The function definition follows the structure to all of the syscall API, lets keep them coherent.

os/pm/pm_procfs.c
amogh-samsung: Let's follow the standard TizenRT practice of returning <0 for erroneous cases here please?

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,119 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
...
 
 	if (state >= PM_NORMAL && state < PM_SLEEP) {
-		pm_stay(0, state);
+		pm_stay(PM_IDLE_DOMAIN, state);
 		fvdbg("power locked!\n");
 	} else {
+		return 1;
 		fdbg("ERROR: Invalid state passed \n");
```
amogh-samsung, line 487: Also, the dbg statement after the return statement will never be reached

edwakuwaku: yes, this is wrong as mentioned, I will correct it.

amogh-samsung:
> 1. That pm_stay() in os_start is a must from my point of view, because without that, it might be dangerous because if some app run immediately after immediately after boot up, the state drop might cause some unexpected problem. So, application side should setup PM file descriptor and invoke syscall to unlock the state.
> 2. I assume you didn't enable BLE Adv mode (which will be one of the interrupt src), UART (another interrupt source), or timer(another interrupt source, but as mentioned in (1), you need to setup PM file descriptor, and unlock state by providing timer interval, you can check the discussion thread above). LOGUART is not an interrupt source, thus it is normal not waking up by sending command to tash.
>    After sleep has triggered, please long press the hard reset button until you see `[MAIN-I] LP INT`, then release it, it will do system reboot, it works the same for download mode (toggle the UART JIG stick).

Thankyou for the clarifications.
1. Regarding this point, I think we could add a pm_relax() statement after DEBUGVERIFY(os_bringup()) to unblock state transition after the bootup process is complete.
2. I understand. Earlier during testing we had LOGUART enabled as an interrupt source so I assumed the same now.
For the 1st test case that I described earlier, I will set up a timer interrupt that wakes up the cores and then lets them go to sleep again.

build/configs/rtl8730e/scripts/rlx8730e_img2.ld
```diff
@@ -135,6 +135,7 @@ SECTIONS
 
 		*(.vectors)
 		*(.start)
+		*(.ps_boot)
```
amogh-samsung, line 138: It would be more suitable to have this commit as a separate PR since it's not directly related to PM

edwakuwaku: It is related to PM, for the arch codes in a32_pm.S, it need to be placed in PSRAM.

os/board/rtl8730e/src/component/bluetooth/inc/rtk_bt_power_control.h
```diff
@@ -44,7 +44,11 @@ void rtk_bt_disable_power_save(void);
 * @param[in] p_resume_callback: Callback invoked after system waking from power save mode.
```
amogh-samsung, line 44: Is the file ownership mode change intended here, or pushed by mistake?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
```diff
@@ -204,10 +204,12 @@ int  flash_read_word(flash_t *obj, u32 address, u32 *data)
 	u32 temp;
```
amogh-samsung, line 204: Same as the file above

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/serial_api.c
```diff
@@ -542,7 +542,7 @@ void serial_baud(serial_t *obj, int baudrate)
 	UART_SetBaud(puart_adapter->UARTx, baudrate);
```
amogh-samsung, line 542: Same as the file above

os/board/rtl8730e/src/component/os_dep/osdep_service_critical.c
```diff
@@ -9,7 +9,8 @@
```
amogh-samsung, line 9: Same as the file above

os/board/rtl8730e/src/component/os_dep/osdep_service_misc.c
```diff
@@ -191,3 +191,19 @@ void irq_disable(IRQn_Type IrqNum)
 	}
```
amogh-samsung, line 191: Same as the file above

amogh-samsung (commented):
Could you confirm this for all files where the ownership mode is changed please?

build/configs/rtl8730e/scripts/rlx8730e_img2.ld
amogh-samsung: Okay, I misunderstood it as a general change for relocating multiple/all code sources. Is that not the case?

edwakuwaku: No, only some specific codes will have special marking to be placed to PSRAM section.

os/include/tinyara/pm/pm.h
```diff
@@ -297,13 +297,18 @@ enum pm_state_e {
 	PM_COUNT,
 };
 
+struct timer_s {
```
sunghan-chang, line 300: The name is too common, let me suggest `struct pm_wakeup_timer_s`

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,117 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
...
+	/* Buffer string need not to be considered, just div level is enough */
+	/*
+	   Only affect active mode power consumption
+	   0 -> 1.2GHz
+	   1 -> 600MHz
+	   2 -> 400Mhz
+	   3 -> 300Mhz
+	 */
+	(void)buffer;
+	SOCPS_Tune_APFreq(div_lvl);
```
sunghan-chang, line 565: @amogh-samsung @edwakuwaku This is the chip-specific API. We don't call it directly from the common module. Could you check it?

sunghan-chang: @amogh-samsung Even we add this like this now, we should clean up this. PM should not have any chip specific code. PM should have HAL or driver for this and chipset should enable the pm feature if it supports.

os/include/tinyara/pm/pm.h
edwakuwaku: Sure, will change accordingly~

os/pm/pm_procfs.c
edwakuwaku: Hi @sunghan-chang ,

Previously we have created another file named pm_ioctl.c, and using that as an interface to connect application and PM driver. But, I was told that that is not needed, which we should utilize the procfs layer for application control, which we then switch to verified that on test suite already.
As the requirement is keep on changing over time, I hope that Samsung side can have internal discussion regarding the application control design, and provide a unified and stable requirement.

Thank you.

amogh-samsung: Hi,
I think the issue here is that though we need to use procfs only for syscalls, but we cannot invoke chipset specific APIs from here.
So we need to add an indirect call from here.

amogh-samsung: https://github.com/amogh-samsung/TizenRT/commit/e2714de14628ddb41438d4dc6d720abde8433192

Please refer to this for a wrapped/mask call to chipset specific frequency tuning API

edwakuwaku:
Hello @sunghan-chang , can you merge this now~?

os/pm/pm_dvfs.c
```diff
@@ -0,0 +1,110 @@
...
+{
+	DEBUGASSERT(div_lvl >= 0 && div_lvl <= 3);
+	/* Only affects active mode power consumption
+	 * 0 -> 1.2GHz
+	 * 1 -> 600MHz
+	 * 2 -> 400Mhz
+	 * 3 -> 300Mhz
+	 */
+	/* Placed in os/arch/arm/src/amebasmart/amebasmart_pmc.c, can be invoked directly */
+	SOCPS_Tune_APFreq(div_lvl);
```
abhishek-samsung, line 106: This name is very specific to realtek, but this is a common file. ```up_set_dvfs``` or something similar so that other boards can also use it. @amogh-samsung could you please check this?

os/include/tinyara/pm/pm.h
abhishek-samsung: This change is still pending...

os/pm/pm_dvfs.c
edwakuwaku: This wrapper is provided by Amogh, if this API naming is too specific, I can just change it to up_set_dvfs(), is that okay?

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,117 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
...
+* 	Input Parameters:
+*   filep          - File path of the power domain
+*   buffer         - Desired state to be unlocked
+*   timer_interval - Timer interval for the timer interrupt
+*
+* 	Returned Value:
+*   0 (OK) means the state unlock was successful.
+*	1 (Failed) means the input state was incorrect.
  ****************************************************************************/
+#ifdef CONFIG_ARCH_CORTEXA32
```
thapav, line 513: Adding architecture specific check in common PM code is not the best method.

os/pm/Kconfig
```diff
@@ -82,6 +82,14 @@ config PM_MEMORY
 
 		Setting CONFIG_PM_MEMORY=1 disables all smoothing.
 
+if ARCH_CORTEXA32
+config PM_DVFS
+	bool "Dynamic Voltage Frequency Scaling"
+	default y
```
abhishek-samsung, line 88: This shouldnt be enabled by default and dependency on CORTEXA32 shouldnt be added in common code.  

This should be selected by CORTEXA32 (or I think should be selected by the board RTL8730E). @amogh-samsung could you please check this?

thapav:
Hi @edwakuwaku 
The dvfs implementation in PM looks tightly coupled with tp dual board. I think this is not the right method to implement dvfs. Rather dvfs support should be enabled at board level (arch/arm/src/amebasmart). PM should just call standard dvfs API's exposed by board specific code to tune the frequency. (Similar to how we call up_idle().

@sunghan-chang What is your opinion ?

edwakuwaku:
Hi @thapav ,

The dvfs is implemented in board level (ie. arch/arm/src/amebasmart). And PM is invoking BSP level API for the operation. As mentioned by @sunghan-chang , we should not invoke it in the pm_procfs.c, that's why we created another file named pm_dvfs.c as a wrapper for this usage. And the above comment from @abhishek-samsung requested to rename it to a more commonly used named. I don't think your question here is a concern, please check again on the comment above.

os/pm/pm_dvfs.c
amogh-samsung: The wrapper is ok
Changing the name as you mentioned is OK from my side @edwakuwaku

os/pm/pm_procfs.c
amogh-samsung: What is your suggestion on how to add DVFS specific calls to the PM control flow architecture?

os/pm/Kconfig
amogh-samsung: Do you suggest that selecting 8730E and PM should automatically select CONFIG_PM_DVFS?

abhishek-samsung: Yes, if possible that would be good.

os/pm/pm_procfs.c
thapav: As mentioned below:

> Hi @edwakuwaku The dvfs implementation in PM looks tightly coupled with tp dual board. I think this is not the right method to implement dvfs. Rather dvfs support should be enabled at board level (arch/arm/src/amebasmart). PM should just call standard dvfs API's exposed by board specific code to tune the frequency. (Similar to how we call up_idle().
> 
> @sunghan-chang What is your opinion ?

os/pm/Kconfig
edwakuwaku: As suggested, moved to definition under os/arch/arm/Kconfig ARCH_CHIP_AMEBASMART.

amogh-samsung (approved):
@thapav 's suggestion for DVFS call design change is noted and will be addressed later.

> The dvfs implementation in PM looks tightly coupled with tp dual board. I think this is not the right method to implement dvfs. Rather dvfs support should be enabled at board level (arch/arm/src/amebasmart). PM should just call standard dvfs API's exposed by board specific code to tune the frequency. (Similar to how we call up_idle().

sunghan-chang (approved):

PR #6018, opened 2024-01-04
os/board/rtl8730e: Add CONFIG_AMEBASMART_FTL in Kconfig and defconfig

yeetee179:


ewoodev (approved):

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -164,8 +164,13 @@ menu "Realtek RTL8730E BLE Support"
 config AMEBASMART_BLE
 	bool "Enable Amebasmart BLE"
 	default n
+	select FTL_ENABLED
 
 if AMEBASMART_BLE
+config FTL_ENABLED
```
sunghan-chang, line 170: How about `amebasmart_ftl_enabled`?

yeetee179: FTL_ENABLED is used in TP1x as well, I think it should be consistent with TP1x. If you want it to be amebasmart_ftl_enabled, I can change it too.

yeetee179: I have modified it as you have suggested in AIdual, for TP1x I will make the changes in a separate PR.

sunghan-chang (approved):

PR #6020, opened 2024-01-08
os/board/rtl8730e: Set alignment requirements in rtl8730e malloc

jwei5:
- Set global variable "align" used in osdep_service_memory to static to protect variable from being overwritten


ewoodev (commented):
Why use `static` and not `const` or `#define ~` ?

jwei5:
Hi @ewoodev , the original intention was to prevent the global "align" from being changed by other files.
We can use `#define ~` as well since the value will not change. I will make the changes to use `#define ~`

ewoodev (approved):

ewoodev (approved):

jeongarmy (approved):
LGTM

PR #6021, opened 2024-01-09
build/configs/rtl8730e: deletee 8M flash defconfigs 

ewoodev:


sunghan-chang:
@Dane-Kang We wanna clean up defconfigs. There are too many defconfigs so that we should modify them all together always. Could you let us know when you can get 16m boards?

sunghan-chang (approved):

PR #6028, opened 2024-01-11
os/arch/arm/src/armv7-a: Place MMU mapping related APIs to PSRAM

edwakuwaku:
- Changed file: os/arch/arm/src/armv7-a/arm_mmu.c
- After moving out all files to XIP, there is an unexpected delay in boot sequence (around 5 secs)
- Discovered that TLB invalidation operation is causing the delay
- Move several MMU related APIs to PSRAM, delay in boot sequence can be shorten (around 1 sec)
- Note: Commenting out mmu_invalidate_region() in mmu_l1_setentry() will not incur any delay in boot sequence, and able to boot up successfully. The operation flow should be checked in depth.

### DIFF
--- a/build/configs/rtl8730e/scripts/rlx8730e_img2.ld
+++ b/build/configs/rtl8730e/scripts/rlx8730e_img2.ld
@@ -128,6 +128,24 @@ __ca32_fip_dram_start__ = ORIGIN(CA32_FIP_DRAM_NS);
 
 SECTIONS
 {
+	.code :
+	{
+		. = ALIGN(4096);
+		__dram_text_start__ = .;
+		. = ALIGN(0x4);
+
+		*(.vectors)
+		*(.start)
+		*(.timesensitive.text*)
+		*(.sramdram.only.text*)		
+		/* Place arm_mmu.c in PSRAM to reduce delay in boot sequence */
+		*arm_mmu.o (.text*)
+
+		. = ALIGN(4096);
+		__dram_text_end__ = .;
+		_etext_ram = ABSOLUTE(.);
+	} > CA32_BL3_DRAM_NS
+
 	/* xip part */
 	.xip_image2.text ORIGIN(CA32_IMG2_XIP):
 	{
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -277,6 +290,9 @@ void mmu_l1_map_regions(const struct section_mapping_s *mappings,
  ****************************************************************************/
 
 #ifndef CONFIG_ARCH_ROMPGTABLE
+#ifdef CONFIG_ARCH_CHIP_AMEBASMART
+SRAMDRAM_ONLY_TEXT_SECTION
+#endif
```
kishore-sn, line 295: Can you pls add this macro for `mmu_map_app_region` and ` mmu_clear_app_pgtbl` apis also, because we are using invalidate api in both these places as well. The `mmu_map_app_region` will also be called during boot up in case of loadable config.

edwakuwaku: done~

os/arch/arm/src/armv7-a/arm_mmu.c
```diff
@@ -104,6 +108,9 @@ static void mmu_set_flags(uint32_t *val, bool ro, bool exec, uint8_t isL1)
  ****************************************************************************/
 
 #ifndef CONFIG_ARCH_ROMPGTABLE
```
sunghan-chang, line 110: This is the arm architecture code. It looks strange with chip specific code here. @kishore-sn how do you think?

kishore-sn: True.. We could remove the check for AMEBASMART chip and keep the macro for placing it in ram. 
If the macro is used in ld file, then it will get placed in the ram.

edwakuwaku: > True.. We could remove the check for AMEBASMART chip and keep the macro for placing it in ram. If the macro is used in ld file, then it will get placed in the ram.

I think the checking is still needed, as we will not expect all cortex-A chipset face similar issue, and the section config might varies between too,

kishore-sn: How about you just add the whole arm_mmu.o in your psram section in linker script?

edwakuwaku: Good idea, done changing~

kishore-sn (approved):
Looks good.

sunghan-chang (approved):

PR #6029, opened 2024-01-11
os/board/rtl8730e/src/component: fix wifi connection failure with wpa…

Wuth5:
- Changed file: os/board/rtl8730e/src/component/wifi/wifi_interactive_mode.c
- fix wifi connection failure when chose wpa23_aes, now driver will use wpa3 to connect under mix mode


j11591:
Is it needed for TP1x??? or only for AI dual??
If WPA2/3 mixed has to be added, is there no need to add WPA1/2 mixed?

Wuth5:
Hi @j11591 ,
this is for AIDual, i will check tp1x later
wpa/wpa2 mix already existed
image

j11591 (approved):
LGTM

PR #6030, opened 2024-01-12
 os/board/rtl8730e/src: extend watchdog timeout

zhongnuo-tang:
1. watchdog Timeout data type change from u16 to u32
2. adjust prescaler so that maximum timeout value increase from ~65s to ~188s

### DIFF
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/wdt_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/wdt_api.c
@@ -48,6 +48,42 @@ IRQn_Type WdgIrqNum = WDG4_IRQ;
   * @{
   */
 
+
+
+void WDG_Init(WDG_TypeDef *WDG, WDG_InitTypeDef *WDG_InitStruct)
+{
+	u32 prescaler = 0;
+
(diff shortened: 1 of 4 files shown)


os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/wdt_api.c
```diff
@@ -48,6 +48,42 @@ IRQn_Type WdgIrqNum = WDG4_IRQ;
   * @{
   */
 
+
+
+void WDG_Init(WDG_TypeDef *WDG, WDG_InitTypeDef *WDG_InitStruct)
```
sunghan-chang, line 53: The function is newly added. Could you let me know who calla this?

zhongnuo-tang: Dear @sunghan-chang,
This function is not newly added, it was originally at ameba_wdg.c where amebasmart_wdg_settimeout in porting layer will call this function to reset watchdog timer with new timeout value.. 
Thank you

hk-gwak (approved):
I checked that the system is rebooted after near 180s by watchdog.

PR #6031, opened 2024-01-15
os/arch/arm/Kconfig: Remove ARM_HAVE_MPCORE for Cortex-A32

ziyik:
-ARM_HAVE_MPCORE is not needed for Cortex-A32
-Remove default select ARM_HAVE_MPCORE for Cortex-A32


ewoodev (commented):
This config was force disabled by realtek in xip commit.
Could you please explain why disabled this config?

edwakuwaku:
Hi @ewoodev ,

This macro was set back in when we are working on system bringup, and it was referenced from imx6. That chipset is cortex-A9, which introduced the MPCore concept. But for our case, cortexA32 do not need that, and due to recently we have been checking a lot on the bringup codes from NUTTX and RTK (when moving code to XIP), we are cleaning up some redundant codes/config at the same time.
For your information, this macro affect the cache and TLB invalidation process, although there is no big difference between enable/disabling this macro, but we should execute the operations without this macro.

sunghan-chang (dismissed):

sunghan-chang:
@kishore-sn Please confirm this.

ewoodev (commented):
Sorry for delay in merging. 

Another PR was merged, creating a conflict.

Could you please update your branch base?

os/arch/arm/Kconfig
```diff
@@ -262,14 +262,13 @@ config ARCH_CORTEXR4
...
+	bool
+	default n
+	select ARCH_DCACHE
+	select ARCH_ICACHE
+	select ARCH_HAVE_MMU
+	select ARCH_USE_MMU
+	select ARCH_HAVE_THREAD_LOCAL
 	select ARCH_ARMV7A_FAMILY
 
 config ARCH_ARMV7M_FAMILY
```
amogh-samsung, line 274: LGTM

amogh-samsung (approved):

sunghan-chang (approved):

PR #6034, opened 2024-01-18
os/board/rtl8730e: Move Flash Related API to PSRAM

ZhenBei-Sin:
Issue:
When perform smartfs "format" will causing hang.

Root Cause:
After implement the XIP changes, the flash api was default located to the XIP. 
For the flash related API must be locate to PSRAM.

Changes Note:
Add the SRAMDRAM_ONLY_TEXT_SECTION for the flash related API move to PSRAM. 
Add the FLASH_Write_Lock/Unlock to make sure before perform the flash operation, CA32 must be disable the IRQ.


ewoodev (approved):
I checked operation. LGTM~

PR #6038, opened 2024-01-19
apps/examples/power: Modify Power Main example

amogh-samsung:
-Update power_main with new APP control functionality - blocking and unblocking
 power state transitions.
-We do not support triggering of PM_SLEEP from application side anymore.
-The application simply sleeps after blocking and unblocking state transition in
 an infinite loop.
-NOTE: This application does not test DVFS functionality yet.

### DIFF
--- a/apps/examples/power/power_main.c
+++ b/apps/examples/power/power_main.c
@@ -33,70 +33,75 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define POWER_THEAD_SIZE					2048
-#define POWER_THEAD_PRIORITY					255
-#define PM_DAEMON_INTERVAL					15
-#define PM_LOCK(x)						power_set_int(PM_LOCK_PATH, x)
-#define PM_UNLOCK(x)						power_set_int(PM_UNLOCK_PATH, x)
+#define POWER_THREAD_SIZE				2048
+#define POWER_THREAD_PRIORITY				255
+#define PM_DAEMON_SLEEP_INTERVAL			5000000	//Microseconds
+#define PM_WAKEUP_TIMER_DURATION			5000000	//Microseconds
 
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
-static int power_write_buf(char *file, char *buf)
-{
-	int fd;
-	int r;
-	int ret = 0;
 
-	fd = open(file, O_WRONLY);
-	if (fd == -1)
-		return -ENOENT;
-
-	r = write(fd, buf, strlen(buf));
-	if (r < 0)
-		ret = -EIO;
-
-	close(fd);
-
-	return ret;
-}
-
-int power_set_int(char *fname, int val)
-{
-	char buf[8];
-	int ret = 0;
-
-	snprintf(buf, sizeof(buf), "%d", val);
-
-	if (power_write_buf(fname, buf) != 0)
-		ret = -EIO;
-
-	return ret;
-}
-
+/* PM_UNLOCK request is always accompanied by a timer interval duration in microseconds.
+ * A timer interrupt is scheduled to wake up the board after this duration.
+ * Replace the timer interval duration with 0 to avoid the use of the timer interrupt.
+ */
 static pthread_addr_t power_daemon()
 {
-	char *state = "sleep";
+	int ret = 0;
+	int fd_lock, fd_unlock;
 
 	/* Release PM lock after bootup*/
-	PM_UNLOCK(PM_NORMAL);
+	fd_unlock = open(PM_UNLOCK_PATH, O_WRONLY);
+	if (fd_unlock < 0) {
+		printf("Failed to open PM_UNLOCK file, ret: %d\n", fd_unlock);
+		return NULL;
+	}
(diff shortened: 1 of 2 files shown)


amogh-samsung:
This PR is for testing reference and should only be merged after https://github.com/Samsung/TizenRT/pull/6015/files#diff-22a6c0ef6ce6ad31a503af87b85ec16b0bfe2af93949a4a2d66fda97244ad3ce

os/pm/pm_procfs.c
```diff
@@ -447,43 +460,113 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
 }
 
 /****************************************************************************
- * Name: power_lock_write
+* 	Name: power_lock_write
```
Taejun-Kwon, line 463: procfs is only for provide information, not for control system.
If you want control something through file system, you should register it's own driver.

apps/examples/power/power_main.c
```diff
@@ -33,54 +33,51 @@
...
+static int power_write_buf(char *file)
+#endif
 {
 	int fd;
-	int r;
 	int ret = 0;
 
 	fd = open(file, O_WRONLY);
 	if (fd == -1)
 		return -ENOENT;
```
edwakuwaku, line 56: Suggest to add pmdbg(...) to notice user that we have failed to open some file descriptors:
Possible reasons:
- did not enable procfs automount config

Similar as to read/write operation below~

r-prabu (approved):

apps/examples/power/power_main.c
```diff
@@ -33,54 +33,51 @@
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
-#define PM_LOCK_PATH						"/proc/power/domains/0/pm_lock"
-#define PM_UNLOCK_PATH					"/proc/power/domains/0/pm_unlock"
-#define PM_SLEEP_PATH						"/proc/power/domains/0/enter_sleep"
+#define POWER_THREAD_SIZE				2048
+#define POWER_THREAD_PRIORITY				255
+#define PM_DAEMON_INTERVAL				5
```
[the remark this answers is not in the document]
edwakuwaku: We are using microseconds precision. Yes, adding comment here is fine~
Also, this path is under general examples, let's try to keep it generally applicable too~
Else, we can consider to move it under board specific examples

apps/examples/power/power_main.c
amogh-samsung: Using printf as a debug statement here
Added

apps/examples/power/power_main.c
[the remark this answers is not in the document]
amogh-samsung: Agreed and added the units in the comments.

apps/examples/power/power_main.c
```diff
@@ -33,54 +33,51 @@
...
-	if (power_write_buf(fname, buf) != 0)
-		ret = -EIO;
-
-	return ret;
-}
-
 static pthread_addr_t power_daemon()
 {
-	char *state = "sleep";
+	int ret = 0;
```
edwakuwaku, line 80: We should initialize the value as int ret = -1? Setting to 0 assumes it is correct initially.

apps/examples/power/power_main.c
```diff
@@ -33,54 +33,51 @@
...
 
 	/* Release PM lock after bootup*/
 	PM_UNLOCK(PM_NORMAL);
```
edwakuwaku, line 83: This PM_UNLOCK input argument should be changed according to pm.h, please check again.

apps/examples/power/power_main.c
```diff
@@ -33,54 +33,51 @@
...
+#define POWER_THREAD_SIZE				2048
+#define POWER_THREAD_PRIORITY				255
+#define PM_DAEMON_INTERVAL				5
 
-#define POWER_THEAD_SIZE				2048
-#define POWER_THEAD_PRIORITY		255
-#define PM_DAEMON_INTERVAL			15
-#define PM_LOCK(x)								power_set_int(PM_LOCK_PATH, x)
-#define PM_UNLOCK(x)							power_set_int(PM_UNLOCK_PATH, x)
+#define PM_WAKEUP_TIMER_DURATION			5
```
edwakuwaku, line 40: Did you add the units in comment already? It is not existing here, also the timer duration should be in microseconds precision, you are setting sec as precision here, please revise accordingly/

amogh-samsung: Push failed, sorry I did not notice.
Thankyou for pointing out

apps/examples/power/power_main.c
amogh-samsung: The application has become heavily board dependent and that is not good.
Function calls from the APP layer are currently dependent on the board.
I will change this, but in some time.

apps/examples/power/power_main.c
amogh-samsung: Does not matter in this case as it will always be changed by the very first function call of PM_UNLOCK

edwakuwaku (approved):
LGTM.

abhishek-samsung (approved):

sunghan-chang (approved):

PR #6039, opened 2024-01-22
kernel/task: refactor task_terminate function

abhishek-samsung:
make chages to add critial section only when needed. Also, move the
task cleanup logic to be executed after removing the task from the
task_list.

### DIFF
--- a/os/kernel/task/task_terminate.c
+++ b/os/kernel/task/task_terminate.c
@@ -166,12 +166,6 @@ int task_terminate(pid_t pid, bool nonblocking)
 #endif
 	int ret = OK;
 
-	/* Make sure the task does not become ready-to-run while we are futzing with
-	 * its TCB by locking ourselves as the executing task.
-	 */
-
-	flags = enter_critical_section();
-
 	/* Find for the TCB associated with matching PID */
 
 	dtcb = sched_gettcb(pid);
@@ -181,6 +175,12 @@ int task_terminate(pid_t pid, bool nonblocking)
 		goto errout_with_lock;
 	}
 
+	/* Make sure the task does not become ready-to-run while we are futzing with
+	 * its TCB by locking ourselves as the executing task.
+	 */
+
+	flags = enter_critical_section();
+
 	/* Verify our internal sanity */
 
 #ifdef CONFIG_SMP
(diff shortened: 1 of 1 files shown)


os/kernel/task/task_terminate.c
```diff
@@ -170,7 +170,7 @@ int task_terminate(pid_t pid, bool nonblocking)
 	 * its TCB by locking ourselves as the executing task.
 	 */
 
-	flags = enter_critical_section();
+	sched_lock();
```
sunghan-chang, line 173: 1. Do we need to do sched_lock before getting tcb and disabling mpu?
2. I think that it's logically correct sequence as below.
    1.  remove a tcb from scheduling list
    2.  and then clean up resources 
I heard this causes mqueu assert. Is it possible to update mqueue together?

os/kernel/task/task_terminate.c
```diff
@@ -279,7 +278,7 @@ int task_terminate(pid_t pid, bool nonblocking)
 	
 
 errout_with_lock:
-	leave_critical_section(flags);
+	sched_unlock();
 	return ret;
 	
 }
```
amogh-samsung, line 284: This seems good

amogh-samsung (approved):

r-prabu (approved):

os/kernel/task/task_terminate.c
sunghan-chang: @abhishek-samsung Could you check my comments?

abhishek-samsung: Hello, I will update the PR by monday end. I need some time to go through the task exit/termination logic in order to make the changes.

abhishek-samsung: > Do we need to do sched_lock before getting tcb and disabling mpu?

No, sched lock is not needed for getting tcb and disabling mpu. It's only needed for removing the task from the task list.  Once we remove the task from the task_list, it can no longer come into context. So, the app related mpu/mmu settings can be modified without critical section.

>I think that it's logically correct sequence as below.
>remove a tcb from scheduling list
>and then clean up resources
>I heard this causes mqueu assert. Is it possible to update mqueue together?

1. I have updated the code. Now, we only have removing the task from tasklist in critical section and the clean up is done after that. Once its removed from tasklist, it no longer can come into context, so we can continue clean up without critical section.
2. mqueue assert was caused because tcb->task_state was wrongly set to INVALID in the SMP PR before task_exithook was called. This issue was fixed by my previous PR. I have made the tcb->task_state invalid only after the cleanup is done.

Also, regarding calling remove from the task_list first and then setting the task_state as invalid after calling task_exithook :

I have checked regarding this. As we are removing the task from the task_list first, it will not be able to come to context. As a result, we need not worry about setting the task_state. So, the current changes should not cause any issues.
Also, for further confirmation regarding this, I have checked nuttx code. They are not even setting the task_state to invalid in task_terminate, so the current and the previous PR for task_terminate should not cause any issues for single core.

https://github.com/apache/nuttx/blob/d79e5a49328f1c48f4465d543608f5cdee722555/sched/task/task_terminate.c#L99

sunghan-chang:
@abhishek-samsung Could you confirm the onboarding  with this on both chipsets?

abhishek-samsung:
> @abhishek-samsung Could you confirm the onboarding with this on both chipsets?

I have already tested it with loadable tp1x. Will test with aidual tomorrow and will let you know the result (I dont have the setup with me to test today).

abhishek-samsung:
> > @abhishek-samsung Could you confirm the onboarding with this on both chipsets?
> 
> I have already tested it with loadable tp1x. Will test with aidual tomorrow and will let you know the result (I dont have the setup with me to test today).

Tested it on aidual too and onboarding works fine.

sunghan-chang (approved):

PR #6044, opened 2024-01-26
framework/include/security : Increase secure storage data slot size

jylee9613:
There is a slot that stores AP information, but the required size has increased while storing multi-AP information. Since the number of slots is limited, it is more efficient to store the data by increasing the size of existing slots instead of dividing it into several slots. Therefore, this commit increase data slot size from 256 to 4064. Maximum size of a slot is 4KB,
but when data is stored in one slot, 32 bytes of tag is added. Therefore, user can store 4064 bytes of data.

### DIFF
--- a/framework/include/security/security_common.h
+++ b/framework/include/security/security_common.h
@@ -25,8 +25,8 @@
 extern "C" {
 #endif
 
-#define SECURITY_MAX_KEY_BUF          4096
-#define SECURITY_MAX_CERT_BUF         4096
+#define SECURITY_MAX_KEY_BUF          (4064) // (Physical MAX slot size) - (head & tail size) (4096 - 32)
+#define SECURITY_MAX_CERT_BUF         (4064) 
 #define SECURITY_MAX_SS_BUF           4096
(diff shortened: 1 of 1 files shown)


framework/include/security/security_common.h
```diff
@@ -25,8 +25,8 @@
 extern "C" {
 #endif
 
-#define SECURITY_MAX_KEY_BUF          4096
-#define SECURITY_MAX_CERT_BUF         4096
+#define SECURITY_MAX_KEY_BUF          (4064) // (Physical MAX slot size) - (head & tail size) (4096 - 32)
+#define SECURITY_MAX_CERT_BUF         (4064) 
 #define SECURITY_MAX_SS_BUF           4096
```
sunghan-chang, line 30: We can save 4064 byte as you modified, but ss buf is still 4096. Is it ok??

sunghan-chang:
You write `SECURITY_MAX_CERT_BUF is also changed to 4096 for the identity with trt_se.h.` in the description. But there is no trt_se.h file in the pr.

sunghan-chang (approved):

PR #6045, opened 2024-01-30
os/arch/arm: allocate fiq stack based on CONFIG_ARMV7A_DECODEFIQ

abhishek-samsung:
fiq stack is allocated in all cases as of now. Hence, allocate fiq stack only if CONFIG_ARMV7A_DECODEFIQ is enabled.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_irq.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_irq.c
@@ -54,7 +54,6 @@ volatile uint32_t *g_current_regs[CONFIG_SMP_NCPUS];
  */
 
 uint64_t g_irqstack_alloc[INTSTACK_ALLOC >> 3];
-uint64_t g_fiqstack_alloc[INTSTACK_ALLOC >> 3];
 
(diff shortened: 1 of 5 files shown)


abhishek-samsung:
@prabu-r @kishore-sn please review

os/arch/arm/src/amebasmart/amebasmart_irq.c
```diff
@@ -72,6 +71,10 @@ uintptr_t g_irqstack_top[CONFIG_SMP_NCPUS] =
 #endif
 };
 
+#if defined(CONFIG_ARMV7A_DECODEFIQ)
```
r-prabu, line 74: Is this config CONFIG_ARMV7A_DECODEFIQ disabled by default?

abhishek-samsung: Yes, Its disabled in RTL8730e. It will be enabled only if we use fiq.

r-prabu (approved):

sunghan-chang (approved):

PR #6046, opened 2024-01-31
os/board/rtl8730e: increase max BLE indication queue size to 60

yeetee179:


sunghan-chang:
@yeetee179 Please leave some description in the commit.

yeetee179:
> @yeetee179 Please leave some description in the commit.

I have added some descriptions in the commit

sunghan-chang:
@yeetee179 1. If it hits the max, what happen?
2. If we increase this to 60, it does not hit the max?
3. Is there any reason why 60 is proper? not 50? not 70? not any other value?

PR #6047, opened 2024-02-01
os/board/rtl8730e: Fix for tash console bootup stage input problem

edwakuwaku:
1) Early stage crash when input to tash:
- Solution: Remove customization for disabling LP core Loguart interrupt
2) Early stage hang when input to tash:
- Solution: Trigger IPC to LP core to release Loguart interrupt for AP core


jeongarmy (approved):
LGTM

PR #6048, opened 2024-02-02
os/arch/arm/src/amebasmart, os/pm: Refine PM DVFS and fix struct naming mismatch for RTL8730E

edwakuwaku:
- Update DVFS wrapper layer
- Fixed PM timer struct naming mismatch (For PR#6015)

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmc.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmc.c
@@ -58,6 +58,8 @@
 #include <task.h>
 #include "freertos_pmu.h"
 #endif
+/* Include header for DVFS implementation */
+#include <tinyara/pm/pm.h>
 #include "ameba_soc.h"
 #include "sys_io.h"
 #include "gic.h"
@@ -281,7 +283,8 @@ void SOCPS_SleepCG(void)
 	//pmu_acquire_wakelock(PMU_OS);
 }
 
-void up_set_dvfs(u32 val)
+#ifdef CONFIG_PM_DVFS
+void up_set_dvfs(int div_lvl)
 {
 	/* Please refer to sysreg_hsys.h */
 	// AP_CLK_DIV1		0
@@ -292,6 +295,7 @@ void up_set_dvfs(u32 val)
 	/* Div clock speed by input val */
 	reg_div = HAL_READ32(SYSTEM_CTRL_BASE_HP, REG_HSYS_HP_CKSL);
 	reg_div &= ~HSYS_MASK_CKD_AP;
-	reg_div |= HSYS_CKD_AP(val);
+	reg_div |= HSYS_CKD_AP(div_lvl);
 	HAL_WRITE32(SYSTEM_CTRL_BASE_HP, REG_HSYS_HP_CKSL, reg_div);
 }
+#endif
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -51,7 +51,7 @@
 #include "timer_api.h"
 
 gtimer_t g_timer1;
-extern struct timer_s g_timer_wakeup;
+extern struct pm_wakeup_timer_s g_timer_wakeup;
 
 void SOCPS_SetAPWakeEvent_MSK0(u32 Option, u32 NewStatus)
 {
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -616,12 +616,16 @@ int pm_changestate(int domain, enum pm_state_e newstate);
 
 enum pm_state_e pm_querystate(int domain);
 
+#ifdef CONFIG_PM_DVFS
 /****************************************************************************
  * Name: pm_dvfs
  *
  * Description:
- *   This internal function is called to reduce the frequency of operation
- *   of the secondary AP core of AI_Dual chipset.
+ *   This internal function is called to reduce the clock frequency
+ *   of CPU core. It can be applied on some scenario which when low
+ *	 loading activity is expected, we can invoke this API to wind down
+ *   CPU cores with high operating frequency, to enhance the effectiveness
+ *	 for power saving.
  *
  * Input Parameters:
  *   div_lvl - voltage frequency scaling level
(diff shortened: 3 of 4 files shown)


os/include/tinyara/pm/pm.h
```diff
@@ -635,6 +635,23 @@ enum pm_state_e pm_querystate(int domain);
...
+ * Input Parameters:
+ *   div_lvl - voltage frequency scaling level
+ *
+ * Returned Value:
+ *   None.
+ *
+ *
+ ****************************************************************************/
+void up_set_dvfs(int div_lvl);
+
```
amogh-samsung, line 654: Please take a look at the function description for pm_dvfs() in pm_dvfs.c and the header file and make any changes required to make the description accurate.

os/arch/arm/src/amebasmart/amebasmart_pmc.c
```diff
@@ -57,11 +57,15 @@
 #include "FreeRTOS.h"
 #include <task.h>
 #include "freertos_pmu.h"
+#else
+#ifdef CONFIG_PM
```
amogh-samsung, line 61: from Make.defs

```
ifeq ($(CONFIG_PM),y)
CMN_CSRCS += psci.c amebasmart_pmc.c
```

So do we need to check again?

os/arch/arm/src/amebasmart/amebasmart_pmc.c
```diff
@@ -281,7 +285,8 @@ void SOCPS_SleepCG(void)
 	//pmu_acquire_wakelock(PMU_OS);
 }
 
-void up_set_dvfs(u32 val)
+#ifdef CONFIG_PM_DVFS
```
amogh-samsung, line 288: From Make.defs

```
ifeq ($(CONFIG_PM),y)
CMN_CSRCS += psci.c amebasmart_pmc.c
```

So do we need to check again?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -51,7 +51,7 @@
 #include "timer_api.h"
 
 gtimer_t g_timer1;
-extern struct timer_s g_timer_wakeup;
+extern struct pm_wakeup_timer_s g_timer_wakeup;
```
amogh-samsung, line 54: Please separate this into a different commit as a fix

edwakuwaku: This fix is already separated on an independent commit

os/include/tinyara/pm/pm.h
```diff
@@ -635,6 +635,23 @@ enum pm_state_e pm_querystate(int domain);
  ****************************************************************************/
 void pm_dvfs(int div_lvl);
 
+/****************************************************************************
```
amogh-samsung, line 638: If CONFIG_PM is checked for compiling up_set_dvfs(), then it should be checked in the header file too. Otherwise up_set_dvfs gets declared in header file but will never be defined

edwakuwaku: Added as suggested~

os/arch/arm/src/amebasmart/amebasmart_pmc.c
```diff
@@ -57,11 +57,13 @@
 #include "FreeRTOS.h"
```
amogh-samsung, line 57: Please update the description of this commit to briefly explain the changes made in the commit

amogh-samsung (commented):
Could you please update the title of the PR according to the latest pushed commit titles and descriptions

abhishek-samsung (approved):

thapav (approved):
LGTM

os/include/tinyara/pm/pm.h
```diff
@@ -633,7 +633,28 @@ enum pm_state_e pm_querystate(int domain);
  *   Incorrect frequency scaling level given as input will lead to an assert.
  *
  ****************************************************************************/
+#ifdef CONFIG_PM_DVFS
```
r-prabu, line 636: It would be better if the conditional flags also includes the documentation.

amogh-samsung:
This fixes a naming mismatch of PM wakeup timer structure from #6015

os/include/tinyara/pm/pm.h
edwakuwaku: done~

os/arch/arm/src/amebasmart/amebasmart_pmc.c
edwakuwaku: done~

amogh-samsung (approved):

os/arch/arm/src/amebasmart/amebasmart_pmc.c
```diff
@@ -281,7 +283,8 @@ void SOCPS_SleepCG(void)
 	//pmu_acquire_wakelock(PMU_OS);
 }
 
-void up_set_dvfs(u32 val)
+#ifdef CONFIG_PM_DVFS
+void up_set_dvfs(int val)
```
sunghan-chang, line 287: You add the prototype in pm.h as below
```
void up_set_dvfs(int div_lvl);
```
But here is `(int val)`.

os/include/tinyara/pm/pm.h
```diff
@@ -616,12 +616,13 @@ int pm_changestate(int domain, enum pm_state_e newstate);
...
 
+#ifdef CONFIG_PM_DVFS
 /****************************************************************************
  * Name: pm_dvfs
  *
  * Description:
- *   This internal function is called to reduce the frequency of operation
- *   of the secondary AP core of AI_Dual chipset.
+ *   This internal function is called to reduce the clock frequency
+ *   of the AP core (CA32) of AI_Dual chipset.
```
sunghan-chang, line 625: 1. AI Dual is our internal name. It's not proper to use in public github.
2. I left the comment as it's not proper to add chip-specific API because pm is the common module. With the same reason, `AP core (CA32) of AI_Dual chipset` can be changed as general description.

amogh-samsung: I agree, this function is intended to be used in common for all chipsets later.
So lets not have the description specific to one chipset.

edwakuwaku: Modified to common description, please check~

sunghan-chang (approved):

PR #6040, opened 2024-01-23
os/arch/arm/src, os/board/rtl8730e: SMP for RTL8730E

edwakuwaku:
- SMP dual core feature porting
- Verified running ok with smp_main.c (Test on XIP)
- Wifi basic function ok
- BT init working fine
- Hotplug mode implementation will have an independent PR
- Ported customization for CPU gating when doing flash operation

PR #6049, opened 2024-02-05
os, lib: Fix build errors for SMP with loadable apps

kishore-sn:
When SMP is enabled for loadable app configs, we are getting errors:

- Missing sections in ld file
- Missing code for some syscalls
- Critical section apis cannot be used in user side library


os/arch/arm/Kconfig
```diff
@@ -275,6 +275,7 @@ config ARCH_CORTEXA32
 	select ARCH_USE_MMU
 	select ARCH_HAVE_THREAD_LOCAL
 	select ARCH_ARMV7A_FAMILY
+	select ARCH_HAVE_TESTSET
```
abhishek-samsung, line 278: How about adding dependency on smp here?
```suggestion
	select ARCH_HAVE_TESTSET if SMP
```

edwakuwaku: In https://github.com/Samsung/TizenRT/blob/master/os/kernel/Kconfig#L290-L294, this is one of the dependency for having SMP, which means ARCH_HAVE_TESTSET -> CONFIG_SMP, if adding as you suggested, it becomes CONFIG_SMP -> ARCH_HAVE_TESTSET

abhishek-samsung: Thanks for pointing it out.

edwakuwaku: Here is some explanation on why we need this macro: https://github.com/Samsung/TizenRT/pull/6040#discussion_r1476235756

sunghan-chang:
@kishore-sn @edwakuwaku Some files look the same as Edward's PR #6040 . Which one should I merge?

edwakuwaku:
Hi Mr.Chang, let's merge PR https://github.com/Samsung/TizenRT/pull/6040 first, and follow by this one to fix loadable_apps SMP compilation error.
Hi @kishore-sn , can you share the status once you verified bootup working fine with PR#6040? We can have another PR for the hotplug mode implementation.

lib/libc/syslog/lib_lowsyslog.c
```diff
@@ -150,9 +150,9 @@ int lowvsyslog(int priority, FAR const char *fmt, va_list ap)
 #if defined(CONFIG_LOGM) && defined(CONFIG_SYSLOG2LOGM)
 		ret = logm_internal(LOGM_LOWPUT, LOGM_UNKNOWN, priority, fmt, ap);
 #else
-		irqstate_t flags = enter_critical_section();
```
edwakuwaku, line 153: May I know the reason to change from enter/leave_critical_section to irqsave/irqrestore here? 
As for now, we are still not very sure which scenario to apply which. In RTK BSP level, there is also quite some places will have to enable/disable interrupts for certain operation. As far as I understood, enter/leave_critical_section will have consideration for multicore coherency (using spinlock), could you share any specific scenario we should apply those?

kishore-sn: enter_critical_section API is only available in kernel side code. 
So, if the library is built for app side, then it generates a build error. 
However, this particular API of lowsyslog will be used only in kernel side. So, I have modified the code to use enter_critical_section if we are building the library for kernel.


In a general case, we need to replace all irqsave / restore by enter / leave_critical_section for SMP. I will check this in case of TizenRT code. Please check rtk code and replace the irqsave api in all locations.

edwakuwaku: Understood, I will check and raise it in another PR.

sunghan-chang:
@kishore-sn Is this ready to review?
If yes, @r-prabu Could you review this?

os/include/sys/syscall.h
```diff
@@ -99,19 +99,28 @@
 #define SYS_sched_yield                (CONFIG_SYS_RESERVED + 12)
 #define SYS_set_errno                  (CONFIG_SYS_RESERVED + 13)
 
+#ifdef CONFIG_SMP
+#define SYS_sched_getaffinity	       (CONFIG_SYS_RESERVED + 14)
+#define SYS_sched_setaffinity	       (CONFIG_SYS_RESERVED + 15)
+#define SYS_sched_getcpu	       (CONFIG_SYS_RESERVED + 16)
+#define __SYS_sched_smp_end	       (CONFIG_SYS_RESERVED + 17)
+#else
+#define __SYS_sched_smp_end	       (CONFIG_SYS_RESERVED + 14)
```
sunghan-chang, line 108: Much minor comment.

All other `__SYS_xxx` represents start of next items.
For example,
```
#define SYS_sem_setprotocol            (__SYS_sched_smp_end + 5)
#define __SYS_named_sem                (__SYS_sched_smp_end + 6)
#else
#define __SYS_named_sem                (__SYS_sched_smp_end + 5)
#endif
```
`__SYS_named_sem` items are located next of it.
But only here, it shows end of previous item topic.
How about `#define __SYS_sem` instead of `__SYS_sched_smp_end`?

edwakuwaku (approved):
Looks fine~

sunghan-chang (approved):

PR #6054, opened 2024-02-06
os/pm: Modify procfs layer interface for PM control

amogh-samsung:
-Procfs layer of PM should not be board dependant.
 tunefreq depends on CONFIG_DVFS, which in turn depends on board support
-Unify function for unlocking PM state transition. We will always support
 option of timer interrupt for waking up from PM_SLEEP, independant of board.
-We will not support the option of pushing the board to go to PM_SLEEP state
 from the app itself, hence remove enter_sleep_write function.
-Correct alignment and spaces.

### DIFF
--- a/os/pm/pm_procfs.c
+++ b/os/pm/pm_procfs.c
@@ -190,19 +190,19 @@ static const uint8_t g_power_statescount = sizeof(g_power_states) / sizeof(g_pow
  */
 
 const struct procfs_operations power_procfsoperations = {
-	power_open,					/* open */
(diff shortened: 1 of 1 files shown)


edwakuwaku:
LGTM.

edwakuwaku (approved):
LGTM.

sunghan-chang (approved):

edwakuwaku (approved):
Good.

sunghan-chang (approved):

PR #6056, opened 2024-02-08
 os/board/rtl8730e: SPI multi-slave implementation

kaikaikaime:
Do note that the SPI0 cs1 pin is currently set as PB31, SPI1 set as PB29*

* Just randomly selected pin

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_spi.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_spi.c
@@ -100,11 +100,16 @@ struct amebasmart_spidev_s {
 	PinName spi_mosi;
 	PinName spi_miso;
 	PinName spi_sclk;
-	PinName spi_cs;
+	PinName spi_cs0;
 	int8_t nbits;               /* Width of word in bits */
 	uint8_t mode;               /* Mode 0,1,2,3 */
 	int role;
+	int8_t multi_cs:1;
 	gpio_t gpio_cs0;
+#ifdef CONFIG_SPI_CS
+	gpio_t gpio_cs1;
+	PinName spi_cs1;
+#endif
 };
 
 enum amebasmart_delay_e {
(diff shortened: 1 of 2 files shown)


abhishek-samsung (approved):

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -211,10 +216,13 @@ static struct amebasmart_spidev_s g_spi0dev = {
...
-	.spi_mosi = PB_29,
-	.spi_miso = PB_30,
-	.spi_sclk = PB_31,
-	.spi_cs = PC_0,
+	.spi_mosi = PB_4,
+	.spi_miso = PB_3,
+	.spi_sclk = PB_6,
+	.spi_cs0 = PB_5,
+#if defined(CONFIG_SPI_CS) && defined(CONFIG_AMEBASMART_SPI0_CS)
+	.spi_cs1 = PB_31,
```
sunghan-chang, line 224: This means that we can add just 1 more slave?

kaikaikaime: Yes, current PR is an addition of 1 slave.

sunghan-chang: Is there any reason to support only 1 additional slave? I mean even I merge this as it is now, can we add more slave later? Is there any hw or sw limitation on this?

kaikaikaime: Theoretically there is no HW limit, with more devices, capacitance and length in wires may start to affect signal integrity. My recommendation is a maximum of 3 slaves with testing for stability. As you can observed, to add new slaves, additional pins are necessary, a point to consider.
You most certainly add move slaves in SW similar to the modification done in the PR.

sunghan-chang (approved):

PR #6058, opened 2024-02-14
os: Add functionality for timed PM state locking

amogh-samsung:
os: Add functionality for timed PM state locking
-Add a timer API to pm layer that allows setting timed callbacks.
-This is currently used to relax PM state transition after a certain
 time of locking.
-It may be expanded to other timed callbacks in the future

apps/examples/power: Fix bad sleep duration macro for power daemon
-Power daemon sleep duration value should be in seconds instead
 of microseconds.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -51,7 +51,7 @@
 #include "timer_api.h"
 
 gtimer_t g_timer1;
-extern struct pm_wakeup_timer_s g_timer_wakeup;
+extern struct pm_timer_s g_pm_timer;
 
 void SOCPS_SetAPWakeEvent_MSK0(u32 Option, u32 NewStatus)
 {
@@ -115,19 +115,35 @@ int SOCPS_AONWakeReason(void)
 
 void pg_timer_int_handler(void *Data)
 {
-	pmvdbg("Timer wakeup interrupt handler!!\n");
+	pmvdbg("PM Timer interrupt handler!!\n");
+	switch (g_pm_timer.timer_type) {
+		case PM_LOCK_TIMER:
+			pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
+			break;
+		case PM_WAKEUP_TIMER:
+			pm_activity(PM_IDLE_DOMAIN, 9);
+			break;
+		default:
+			pmdbg("Timer callback triggered without setting timer, Unexpected Error!!!\n");
+			break;
+	}
 	// Reset the global struct
-    g_timer_wakeup.use_timer = 0;
-    g_timer_wakeup.timer_interval = 0;
+	g_pm_timer.timer_type = 0;
+	g_pm_timer.timer_interval = 0;
 	// Switch status back to normal mode after wake up from interrupt
-	pm_activity(PM_IDLE_DOMAIN, 9);
 }
 
-void ap_timer_helper(void) {
+void up_set_pm_timer(void) {
 	// Check whether timer interrupt need to be set
-	if (g_timer_wakeup.use_timer) {
-		gtimer_init(&g_timer1, TIMER1);
-		gtimer_start_one_shout(&g_timer1, g_timer_wakeup.timer_interval, (void *)pg_timer_int_handler, NULL);
+	if (g_pm_timer.timer_type) {
+		/* Open timer device, initialize and start timer if required
+		 * OR
+		 * Call unified timer setting BSP API
+		 */
+		//TODO: This is pending
+
+		//gtimer_init(&g_timer1, TIMER1);
+		//gtimer_start_one_shout(&g_timer1, g_pm_timer.timer_interval, (void *)pg_timer_int_handler, NULL);
 	}
 }
 
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -303,9 +303,15 @@ enum pm_state_e {
 	PM_COUNT,
 };
 
-struct pm_wakeup_timer_s {
-	bool use_timer;
-	uint32_t timer_interval;
+enum pm_timer_type_e {
+	PM_WAKEUP_TIMER = 1,
+	PM_LOCK_TIMER = 2,
+	/* Scope for future expansion, up to 8 timer types can be supported */
+};
+
+struct pm_timer_s {
+	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
+	uint32_t timer_interval;	/* The interval for whichever timer is to be used */
 };
 
 /* This structure contain pointers callback functions in the driver.  These
@@ -522,6 +528,25 @@ void pm_stay(int domain, enum pm_state_e state);
  ****************************************************************************/
 
 void pm_relax(int domain, enum pm_state_e state);
+
+/****************************************************************************
+ * Name: pm_set_timer
+ *
+ * Description:
+ *   This function is called to set a timed callback to an intended function.
+ *   It may be used to invoke pm_relax() after a fixed time duration of
+ *   locking using pm_stay().
+ *
+ * Input Parameters:
+ *   timer_type - the type of the timer which determines the callback
+ *   timer_interval - duration of the timer
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_set_timer(int timer_type, size_t timer_interval);
 
 /****************************************************************************
  * Name: pm_staycount
--- a/os/pm/pm_activity.c
+++ b/os/pm/pm_activity.c
@@ -203,6 +203,36 @@ void pm_stay(int domain, enum pm_state_e state)
 }
 
 /****************************************************************************
+ * Name: pm_set_timer
+ *
+ * Description:
+ *   This function is called to set a timed callback to an intended function.
+ *   It may be used to invoke pm_relax() after a fixed time duration of
+ *   locking using pm_stay().
+ *
+ * Input Parameters:
+ *   timer_type - the type of the timer which determines the callback
+ *   timer_interval - duration of the timer
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_set_timer(int pm_timer_type, size_t timer_interval)
+{
+	/* Set the global variables to trigger the appropriate timer */
+	g_pm_timer.timer_type = pm_timer_type;
+	g_pm_timer.timer_interval = timer_interval;
+
+	/* Wakeup timer will be triggered at the time of sleep, otherwise trigger now. */
+	if (pm_timer_type != PM_WAKEUP_TIMER) {
+		up_set_pm_timer();
+	}
+	return;
+}
+
+/****************************************************************************
  * Name: pm_relax
  *
  * Description:
--- a/os/pm/pm_procfs.c
+++ b/os/pm/pm_procfs.c
@@ -117,7 +117,7 @@ struct power_procfs_entry_s {
 };
 
 /* Added to record timer countdown interrupt */
-struct pm_wakeup_timer_s g_timer_wakeup = { 0, 0 };
+struct pm_timer_s g_pm_timer = { 0, 0 };
 
 /****************************************************************************
  * Private Function Prototypes
@@ -461,7 +461,7 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
  * Input Parameters:
  *   filep             - File path of the power domain
  *   buffer            - Not used
- *   buflen            - Not used
+ *   buflen            - Use of timer interrupt and its duration
  *
  * Returned Value:
  *  (OK) means the state lock was successful.
@@ -470,7 +470,6 @@ static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_
 static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
 {
 	(void)buffer;
-	(void)buflen;
 
 	/* Always lock PM_NORMAL state, check again whether this implementation
 	is enough to handle, if not, consider below methods...
@@ -481,7 +480,11 @@ static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, s
 	*/
 	pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
 	fvdbg("State locked!\n");
-
+	if (buflen > 0) {
+		pm_set_timer(PM_LOCK_TIMER, buflen);
+	} else {
+		g_pm_timer.timer_type = 0;
+	}
 	return OK;
 }
 
@@ -510,11 +513,9 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
 	pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
 	fvdbg("State unlocked!\n");
 	if (buflen > 0) {
-		g_timer_wakeup.use_timer = 1;
-		g_timer_wakeup.timer_interval = buflen;
+		pm_set_timer(PM_WAKEUP_TIMER, buflen);
 	} else {
-		g_timer_wakeup.use_timer = 0;
-		g_timer_wakeup.timer_interval = 0;	
+		g_pm_timer.timer_type = 0;
 	}
 	return OK;
 }
--- a/apps/examples/power/power_main.c
+++ b/apps/examples/power/power_main.c
@@ -35,7 +35,7 @@
  ****************************************************************************/
 #define POWER_THREAD_SIZE				2048
 #define POWER_THREAD_PRIORITY				255
-#define PM_DAEMON_SLEEP_INTERVAL			5000000	//Microseconds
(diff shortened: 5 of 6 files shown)


os/include/tinyara/pm/pm.h
```diff
@@ -303,9 +303,14 @@ enum pm_state_e {
...
 
-struct pm_wakeup_timer_s {
-	bool use_timer;
-	uint32_t timer_interval;
+enum pm_timer_type_e {
+	USE_PM_WAKEUP_TIMER = 1,
+	USE_PM_LOCK_TIMER = 2,
+};
+
+struct pm_timer_s {
```
edwakuwaku, line 311: Lets keep the struct name as pm_wakeup_timer_s, it is decided by Mr.Chang.

os/pm/pm_procfs.c
```diff
@@ -481,7 +480,13 @@ static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, s
 	*/
 	pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
 	fvdbg("State locked!\n");
-
+	if (buflen > 0) {
+		g_pm_timer.use_timer = USE_PM_LOCK_TIMER;
```
edwakuwaku, line 484: This method of assigning variable directly, can only support for current situation, if in the future, there is more request for timer usage, how do you plan to handle it? How about we use the bit masking method now?

os/include/tinyara/pm/pm.h
```diff
@@ -303,9 +303,14 @@ enum pm_state_e {
 	PM_COUNT,
 };
 
-struct pm_wakeup_timer_s {
-	bool use_timer;
-	uint32_t timer_interval;
+enum pm_timer_type_e {
```
edwakuwaku, line 306: Lets remove the USE
enum pm_timer_type_e {
  PM_COMMON_ONE_SHOT_TIMER = 1,
  PM_HOLD_STATE_TIMER                 = 2,
  /* More to add */
  PM_TIMER_MAX                              = 8, // Should never reach, if yes, then we have too much case to handle, consider to expand to u16
}

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -115,19 +115,29 @@ int SOCPS_AONWakeReason(void)
...
+	g_pm_timer.timer_interval = 0;
 	// Switch status back to normal mode after wake up from interrupt
-	pm_activity(PM_IDLE_DOMAIN, 9);
 }
 
-void ap_timer_helper(void) {
+void up_set_timer(void) {
 	// Check whether timer interrupt need to be set
-	if (g_timer_wakeup.use_timer) {
+	if (g_pm_timer.use_timer) {
```
edwakuwaku, line 138: Similarly here, we are having the condition of 2 timer use case which will not have conflict, but what if later we have one-shot and periodical together? Then we will need another g_timer_2, then this method cannot work, we need bit masking method for more general working.

os/pm/pm_procfs.c
edwakuwaku: g_pm_timer.use_timer |= (1 << PM_LOCK_TIMER);

os/pm/pm_procfs.c
```diff
@@ -510,11 +515,10 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
 	pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
 	fvdbg("State unlocked!\n");
 	if (buflen > 0) {
-		g_timer_wakeup.use_timer = 1;
-		g_timer_wakeup.timer_interval = buflen;
+		g_pm_timer.use_timer = USE_PM_WAKEUP_TIMER;
```
edwakuwaku, line 518: g_pm_timer.use_timer |= (1 << PM_COMMON_ONE_SHOT_TIMER);

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
edwakuwaku: Let's do a switch case here similar to the callback handler

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -115,19 +115,29 @@ int SOCPS_AONWakeReason(void)
 
 void pg_timer_int_handler(void *Data)
 {
-	pmvdbg("Timer wakeup interrupt handler!!\n");
+	pmvdbg("PM Timer interrupt handler!!\n");
+	switch (g_pm_timer.use_timer) {
```
edwakuwaku, line 119: For multiple timer case, we need to identify which source trigger the callback, thus we need to provide the callback data for the last argument, which indicate the trigger type.

os/include/tinyara/pm/pm.h
amogh-samsung: The timer usage is not restricted to waking up now so I generalized the name

os/pm/pm_procfs.c
amogh-samsung: even if we support multiple timers (more than 2), and we allow bit wise setting to indicate which timer to use,
The timer_interval member of the struct can still only accommodate only only value, for the latest timer setting that is called.
Hence, to avoid confusion with the bit wise masking, I am setting the timer type directly.

os/include/tinyara/pm/pm.h
edwakuwaku: Okay, make sense to me~

os/pm/pm_procfs.c
edwakuwaku: Since we are setting the variable to a flat value, I believe we are going to have another callback handler for simultaneous timer operation, else we cannot identify to execute which operation when timer expires (in multiple timer scenario). Or do you have any other opinion to achieve this?
We should list out all possible ways to achieve the target, and choose the most efficient way with least overhead.

edwakuwaku: > even if we support multiple timers (more than 2), and we allow bit wise setting to indicate which timer to use, The timer_interval member of the struct can still only accommodate only only value, for the latest timer setting that is called. Hence, to avoid confusion with the bit wise masking, I am setting the timer type directly.

We can add one more timer_interval to differentiate both one-shot and periodical, if both will run simultaneously.

amogh-samsung: Working according to the current requirement of 2 types of timer callbacks (lock release and wakeup),
1. power_lock_write() can reset old PM_WAKEUP_TIMER. It will add bits to indicate that lock timer needs to be set, but it also needs to undo bits of wakeup timer or otherwise, in pg_int_handler callback, both timers will be found as set.
2. We have confirmation that both these timers currently cannot be invoked simultaneously because they are strictly associated with their own PM_STATEs which are mutually exclusive.
3. I am not considering about periodical timer as an additional functionality right now. From our previous discussion, we found that it can be very useful for keep alive requirements, but it comes with multiple considerations for multi-timer synchronization. My design ideas currently do not consider the scope of periodic timer in addition to one shout.
We can take it up if and when we really feel the need of an additional timer.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
amogh-samsung: 1. Keeping in mind that we do not need to consider periodic timers right now, does your opinion change?
2. I observe that currently, in all cases, the callback has to go to the same pg_int_handler() function, so a switch cases is not needed while starting the one-shout timer. The checking can just be done in the callback itself.
3. Again, the scope for future expansion, but modifications for that can be done if and when required.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
amogh-samsung: Agree to your point
For multiple timers coming in concurrently, this and some other considerations need to be handled.
However, I suggest let's not handle those cases right now since we don't have the clear requirement. It could again become a guessing game.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
edwakuwaku: I think we can implement that now, despite of the future implementation grows, as the bit masking method can be generally fit into all kind of considerations.
I suggest as so is because this is a structure implementation, not a request-dependent implementation.

amogh-samsung:
@edwakuwaku 
Implementation is done here keeping in mind the possibility of expansion in the future.
However, support for simultaneously setting multiple timers is not taken of right now

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
amogh-samsung: Noted for future scope, in separate commits.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -115,19 +115,29 @@ int SOCPS_AONWakeReason(void)
...
+	g_pm_timer.timer_interval = 0;
 	// Switch status back to normal mode after wake up from interrupt
-	pm_activity(PM_IDLE_DOMAIN, 9);
 }
 
-void ap_timer_helper(void) {
+void up_set_timer(void) {
 	// Check whether timer interrupt need to be set
-	if (g_timer_wakeup.use_timer) {
+	if (g_pm_timer.timer_type) {
```
sunghan-chang, line 138: Even pm calls the `up_set_timer`, if `g_pm_timer.timer_type` is null, then it does nothing without any warning or error. Is it ok?

sunghan-chang:
@amogh-samsung 
I think we can use several timers in pm.
1. UART could use a timer to wake up for sending alive packet.
2. BLE could use a timer to keep pm normal.

Does it support multi-timer now?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -115,19 +115,29 @@ int SOCPS_AONWakeReason(void)
...
 		gtimer_init(&g_timer1, TIMER1);
-		gtimer_start_one_shout(&g_timer1, g_timer_wakeup.timer_interval, (void *)pg_timer_int_handler, NULL);
+		gtimer_start_one_shout(&g_timer1, g_pm_timer.timer_interval, (void *)pg_timer_int_handler, NULL);
```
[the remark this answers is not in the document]
edwakuwaku: In AIDual, timer1~13 is general purpose timer free for used, and it depends on your choice, I just simply picked TIMER1. In other words, there is no relation between functionality usage and timer index. It only depends on your use case and scenarios.
If needed, you can change the timer index according to your preferences.

edwakuwaku:
@amogh-samsung , seems Mr.Chang has the same concern, do we apply the bit masking method now?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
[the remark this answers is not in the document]
edwakuwaku: Additionally, we have a timer porting layer https://github.com/Samsung/TizenRT/blob/master/os/arch/arm/src/amebasmart/amebasmart_timer_lowerhalf.c, but I'm not sure whether Samsung side is utilizing this using FD from application layer for timer control? 
If yes, do we still invoke timer like this way (ie. invoking BSP code directly)?

os/include/tinyara/pm/pm.h
```diff
@@ -303,9 +303,16 @@ enum pm_state_e {
...
+	PM_WAKEUP_TIMER = 1,
+	PM_LOCK_TIMER = 2,
+	/* Scope for future expansion */
+	PM_MAX_TIMER = 8,
```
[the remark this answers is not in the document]
amogh-samsung: > Usually we use a last definition of an enum to get validation of input value range
Yes. However we have not used it as a formal check anywhere in code as you correctly mentioned.

> or just showing maximum value to be added for future?
So ultimately, because no check is added currently, it is serving this purpose.

os/pm/pm_activity.c
```diff
@@ -202,6 +202,28 @@ void pm_stay(int domain, enum pm_state_e state)
...
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_set_timer(void)
```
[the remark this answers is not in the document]
amogh-samsung: I would like to know your opinion on calling a function like `up_set_pm_timer()` which is in `amebasmart_pmhelpers.c`, directly from `os/pm/pm_*.c` files

We felt that it might be better to have another layer in between so we chose this design. Would like to know your ideas on this please.

Also, what would be your opinion about using a different name like `pm_timer_wrapper()` instead of `pm_set_timer()`?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
amogh-samsung: This is intended functionality, if `g_pm_timer.timer_type` is `0`, that means we do not need to invoke a timed callback along with `pm_stay()`

amogh-samsung:
> @amogh-samsung I think we can use several timers in pm.
> 
> 1. UART could use a timer to wake up for sending alive packet.
> 2. BLE could use a timer to keep pm normal.
> 
> Does it support multi-timer now?

Yes @sunghan-chang , the current implementation supports the use of multiple timers, with each having its own callback to be triggered. However, setting multiple timers simultaneously is not yet made possible.

The PM_LOCK_TIMER is intended for BLE keeping the state in PM_NORMAL for 10 mins

os/pm/pm_activity.c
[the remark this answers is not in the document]
sunghan-chang: pm function which manages a timer for pm operation can have a name with pm_xxxx.
But the `up_set_timer` can be a general timer API, I don't think that BSP should have pm specific timer API. Because it does not need to do something for pm only.
1. pm timer function should manage all timer usages for uart, ble and so on.
2. pm timer function should call timer BSP API or hw timer API with open, read, write as @edwakuwaku said.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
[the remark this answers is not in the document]
sunghan-chang: @amogh-samsung Currently, we can know which timer is used when we search entire codes. I mean user can't know TIMER1 is available or not. Actually it's unavailable because pm uses. 
In short-term, leave a comment on timer or somewhere (pm?) that pm uses TIMER1 on RTL8730E.
In long-term, we should make a timer resource manager to allocate free timer. (This is not related to PM)

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
sunghan-chang: Instead of timer stop, don't call the callback?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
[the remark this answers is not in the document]
amogh-samsung: Understood,
@edwakuwaku does the use of the porting layer help with this?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
amogh-samsung: Sometimes, the locking of PM state transition may not be time bound
i.e. we may want to lock the state transition indefinitely.
So in such cases, we don't want a timer for anything, so no timer is set at all.

os/include/tinyara/pm/pm.h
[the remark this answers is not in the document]
sunghan-chang: Then, just a comment is enough? (Why is maximum 8?)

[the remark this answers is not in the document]
amogh-samsung: Understood, I will modify.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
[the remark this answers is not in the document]
edwakuwaku: It should be similar as to UART registration, and we can set how many of them need to be utilised, using defconfig and macros. Since we have 13 timers for your use case, we can just assigned each as to specific use case as for now, which is the "short term" mentioned by @sunghan-chang . If in the future, we cannot cover all the use case, let's go for timer resource manager. 
I have one question, does the TP1x use this timer porting layer before?

os/pm/pm_activity.c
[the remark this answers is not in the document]
amogh-samsung: Modified the control flow as follows: -
1. APP tries to lock/unlock file with timer functionality
2. pm_procfs.c redirects timer setting call to pm_set_timer()
3. pm_set_timer() sets the timer attributes and calls BSP layer helper function for PM
4. BSP layer PM helper function calls unified timer setting API for AI Dual

**_Steps 3 and 4 may be clubbed, PENDING_**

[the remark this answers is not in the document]
amogh-samsung: @sunghan-chang requesting your review here please

sunghan-chang (approved):

PR #6064, opened 2024-02-19
[Pending for requirement scope] os: Refine timer driver to work with PM module

edwakuwaku:
- Refine timer implementation in RLT8730E
- Apply bit masking method to support multiple timer registration
- Revise the TizenRT timer driver to integrate with PM module
- Register each timer with specific use cases as for now
- Compilation OK
- Extended implementation for multi-timer, base PR: https://github.com/Samsung/TizenRT/pull/6067

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -41,6 +41,8 @@
 #include <tinyara/config.h>
 #include <tinyara/arch.h>
 #include <tinyara/irq.h>
+#include <fcntl.h>
+#include <tinyara/timer.h>
 
 #ifdef CONFIG_PM
 #include <tinyara/pm/pm.h>
@@ -113,10 +115,12 @@ int SOCPS_AONWakeReason(void)
 	return reason;
 }
 
-void pg_timer_int_handler(void *Data)
+/* Follow the callback structure in os/include/tinyara/timer.h */
+/* Update next_interval_us will only benefit timer with periodical mode */
+void timer_int_handler(uint32_t *next_interval_us, pm_timer_type_t timer_type)
 {
 	pmvdbg("PM Timer interrupt handler!!\n");
-	switch (g_pm_timer.timer_type) {
+	switch (timer_type) {
 		case PM_LOCK_TIMER:
 			// Relax the lock on the PM state transitions after the timer interrupt
 			pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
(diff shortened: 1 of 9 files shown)


os/pm/pm_procfs.c
```diff
@@ -118,6 +118,7 @@ struct power_procfs_entry_s {
 
 /* Added to record timer countdown interrupt */
 struct pm_timer_s g_pm_timer = { 0, 0 };
+struct pm_timer_s g_pm_timer = { 0, 0 };
```
ritesh55555, line 121: initialized two times

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -125,20 +129,53 @@ void pg_timer_int_handler(void *Data)
...
+	notify.pid   = (pid_t)getpid();
+	notify.timer_type = pm_timer_type;
+	int fd = 0;
+	switch (pm_timer_type) {
+		case PM_WAKEUP_TIMER:
+			fd = open("/dev/timer1", O_RDONLY);
+			ioctl(fd, TCIOC_SETMODE, MODE_ONESHOT);
+			break;
+		case PM_LOCK_TIMER:
+			fd = open("/dev/timer2", O_RDONLY);
```
ritesh55555, line 157: Should we assign a specific timer to a specific timer_type Or timer_type should take any timer_i if that is available . Suppose we want to run two periodical timer at the same time, then this code might work.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -125,20 +129,53 @@ void pg_timer_int_handler(void *Data)
...
+	ioctl(fd, TCIOC_SETTIMEOUT, g_pm_timer.timer_interval);
+	/* Set callback structure */
+	if (ioctl(fd, TCIOC_NOTIFICATION, (unsigned long)((uintptr_t)&notify)) < 0) {
+		fprintf(stderr, "ERROR: Failed to set the timer handler: %d\n", errno);
+		return;
+	}
+	/* Fire the timer */
+	if (ioctl(fd, TCIOC_START, 0) < 0) {
+		fprintf(stderr, "ERROR: Failed to start the timer: %d\n", errno);
+		return;
```
ritesh55555, line 178: we can do close(fd).

edwakuwaku: Agree, will make changes along with other comments

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
edwakuwaku: Please view the comment in the previous PR from Mr.Chang: https://github.com/Samsung/TizenRT/pull/6058#discussion_r1492352013. Ultimately, there should be a resource manager for timer, but currently in TizenRT there isn't. Until we have an idea or stable scope being introduced, we can only go with the specific timer usage.

samsung-singh (approved):
Refine timer driver to work with PM module, good to merge.

edwakuwaku:
Hi @ritesh55555  @amogh-samsung , this current implementation idea is to remove any further adjustment from app layer, which with just a simple control command for setting up timer. Then, it should go through syscall and reach pm_procfs, and finally, the timer driver control (ioctl) will be done in the BSP layer.
Please help to check whether in the future, app layer will give up other driver control (related to PM), (ie. GPIO, timer, more to come....), which means we just have to create a simple interface, and leave all the jobs to be done in BSP layer. 
OR, the app layer will take control to the drivers (such as UART, through ioctl).
As these implementation will affect both app and driver/kernel layer, thus, some element that need to be take into consideration are as (code maintenance, efficiency for feature implementation, benefits to PM, etc....)
I think this should be discussed, and we hope to see solid feedbacks. If there is any comparison on advantage/disadvantage, we should do that too.

edwakuwaku:
no requirement for triggering multi HW timer, close it for now

PR #6067, opened 2024-02-19
os: Add functionality for timed PM state locking

amogh-samsung:
os: Add functionality for timed PM state locking
-Add a timer API to pm layer that allows setting timed callbacks.
-This is currently used to relax PM state transition after a certain
time of locking.
-It may be expanded to other timed callbacks in the future

apps/examples/power: Fix bad sleep duration macro for power daemon
-Power daemon sleep duration value should be in seconds instead
of microseconds.

os/board/rtl8730e/src/component/bluetooth: Keep board in PM_NORMAL state when onboarding is triggered
-Once onboarding process is triggered by the user from the ST App, board
should stay in PM_NORMAL state for 10 minutes to allow successful completion

Signed-off-by: Amogh Hassija [a.hassija@samsung.com](mailto:a.hassija@samsung.com)

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -51,7 +51,7 @@
 #include "timer_api.h"
 
 gtimer_t g_timer1;
-extern struct pm_wakeup_timer_s g_timer_wakeup;
+extern struct pm_timer_s g_pm_timer;
 
 void SOCPS_SetAPWakeEvent_MSK0(u32 Option, u32 NewStatus)
 {
@@ -115,20 +115,32 @@ int SOCPS_AONWakeReason(void)
 
 void pg_timer_int_handler(void *Data)
 {
-	pmvdbg("Timer wakeup interrupt handler!!\n");
+	pmvdbg("PM Timer interrupt handler!!\n");
+	switch (g_pm_timer.timer_type) {
+		case PM_LOCK_TIMER:
+			// Relax the lock on the PM state transitions after the timer interrupt
+			pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
+			break;
+		case PM_WAKEUP_TIMER:
+			// Switch status back to normal mode after wake up from interrupt
+			pm_activity(PM_IDLE_DOMAIN, 9);
+			break;
+		default:
+			pmdbg("Timer callback triggered without setting timer, Unexpected Error!!!\n");
+			break;
+	}
 	// Reset the global struct
-    g_timer_wakeup.use_timer = 0;
-    g_timer_wakeup.timer_interval = 0;
-	// Switch status back to normal mode after wake up from interrupt
-	pm_activity(PM_IDLE_DOMAIN, 9);
+	g_pm_timer.timer_type = 0;
+	g_pm_timer.timer_interval = 0;
 }
 
-void ap_timer_helper(void) {
+void up_set_pm_timer(void) {
 	// Check whether timer interrupt need to be set
-	if (g_timer_wakeup.use_timer) {
+	if (g_pm_timer.timer_type) {
 		gtimer_init(&g_timer1, TIMER1);
-		gtimer_start_one_shout(&g_timer1, g_timer_wakeup.timer_interval, (void *)pg_timer_int_handler, NULL);
+		gtimer_start_one_shout(&g_timer1, g_pm_timer.timer_interval, (void *)pg_timer_int_handler, NULL);
 	}
+	return;
 }
 
 /* Interrupt callback from wifi-keepalive, which LP received designated packet*/
--- a/os/board/rtl8730e/src/component/bluetooth/example/ble_scatternet/scatternet.c
+++ b/os/board/rtl8730e/src/component/bluetooth/example/ble_scatternet/scatternet.c
@@ -28,6 +28,12 @@
 #include <tinyara/net/if/ble.h>
 #include <ble_tizenrt_service.h>
 #include <debug.h>
+#include <tinyara/config.h>
+#ifdef CONFIG_PM
+#include <tinyara/pm/pm.h>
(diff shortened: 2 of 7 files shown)


edwakuwaku:
This looks clean and straightforward, have you compiled it with PM enabled?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -115,20 +115,31 @@ int SOCPS_AONWakeReason(void)
...
+		default:
+			pmdbg("Timer callback triggered without setting timer, Unexpected Error!!!\n");
+			break;
+	}
 	// Reset the global struct
-    g_timer_wakeup.use_timer = 0;
-    g_timer_wakeup.timer_interval = 0;
+	g_pm_timer.timer_type = 0;
+	g_pm_timer.timer_interval = 0;
 	// Switch status back to normal mode after wake up from interrupt
```
edwakuwaku, line 133: Can we move this comment to type specific?

amogh-samsung: Thankyou for noticing, done

os/board/rtl8730e/src/component/bluetooth/example/ble_scatternet/scatternet.c
```diff
@@ -287,6 +293,21 @@ static rtk_bt_evt_cb_ret_t ble_tizenrt_scatternet_gap_app_callback(uint8_t evt_c
...
+            } else {
+		    /* We lock PM in PM_NORMAL state for 10 minutes to allow comfortable onboarding */
+                    ret = write(fd, NULL, 600000000);
+                    if (ret < 0) {
+                            pmdbg("Failed to write via procfs to lock PM, ret:  %d\n", ret);
+                    }
+                    pmvdbg("PM state transition is locked\n");
+            }
+            close(fd);
+#endif
```
edwakuwaku, line 310: Looks good, but one thing to confirm. In os_start, PM_NORMAL will be locked once, did we unlock once first before starting wifi and bt? If not, the counter will be 2 after locking here, hence even timer expires callback do PM_RELAX, it will still not do state drop.

amogh-samsung: The unlock has to be done on from the product code side.
It will be part of different code.

edwakuwaku (approved):
LGTM.

os/board/rtl8730e/src/component/bluetooth/example/ble_scatternet/scatternet.c
```diff
@@ -287,6 +293,21 @@ static rtk_bt_evt_cb_ret_t ble_tizenrt_scatternet_gap_app_callback(uint8_t evt_c
...
+            int fd = open(PM_LOCK_PATH, O_WRONLY);
+            if (fd < 0) {
+                    pmdbg("Failed to lock PM state transition, fd: %d\n", fd);
+            } else {
+		    /* We lock PM in PM_NORMAL state for 10 minutes to allow comfortable onboarding */
+                    ret = write(fd, NULL, 600000000);
+                    if (ret < 0) {
+                            pmdbg("Failed to write via procfs to lock PM, ret:  %d\n", ret);
+                    }
+                    pmvdbg("PM state transition is locked\n");
```
abhishek-samsung, line 307: this statement should be in else case, if not it gets printed along with ```pmdbg("Failed to write via procfs to lock PM, ret:  %d\n", ret);```

os/board/rtl8730e/src/component/bluetooth/example/ble_scatternet/scatternet.c
```diff
@@ -287,6 +293,21 @@ static rtk_bt_evt_cb_ret_t ble_tizenrt_scatternet_gap_app_callback(uint8_t evt_c
...
+            }
+            close(fd);
```
abhishek-samsung, line 309: this should be only called in else condition.

amogh-samsung: Thankyou for pointing out, done.

os/pm/pm_activity.c
```diff
@@ -202,6 +208,36 @@ void pm_stay(int domain, enum pm_state_e state)
...
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_set_timer(int pm_timer_type, size_t timer_interval)
+{
+	/* Set the global variables to trigger the appropriate timer */
+	g_pm_timer.timer_type = pm_timer_type;
+	g_pm_timer.timer_interval = timer_interval;
```
r-prabu, line 232: Is this required to do only one time set? If yes then how are we restricting? If no then what will be the side effects of updating in multiple times?

os/include/tinyara/pm/pm.h
```diff
@@ -303,9 +303,15 @@ enum pm_state_e {
...
-	bool use_timer;
-	uint32_t timer_interval;
+enum pm_timer_type_e {
+	PM_WAKEUP_TIMER = 1,
+	PM_LOCK_TIMER = 2,
+	/* Scope for future expansion, up to 8 timer types can be supported */
+};
+
+struct pm_timer_s {
+	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
```
r-prabu, line 313: I assume bit are updated according to  pm_timer_type_e . Documentation can be updated accordingly.

amogh-samsung: Bit masking is not supported currently as synchronous multi timer support is not a part of the scope right now.
That will come in next.

os/pm/pm_activity.c
amogh-samsung: Multiple synchronous timers are not supported right now.

sunghan-chang (approved):

r-prabu (approved):

PR #6069, opened 2024-02-20
arm/src/amebasmart,drivers/audio: align alc1019 buffer size with DMA page size

abhishek-samsung:
Set alc1019 apb buffer size to be equal to i2s DMA page size. Also increase i2s DMA page size to 2048

### DIFF
--- a/os/drivers/audio/alc1019.c
+++ b/os/drivers/audio/alc1019.c
@@ -47,8 +47,12 @@
 #define ALC1019_I2S_TIMEOUT_MS 200
 
 /* Default configuration values */
+
+/*
+ * It's good to match the buffer size with i2s DMA page size
+ */
 #ifndef CONFIG_ALC1019_BUFFER_SIZE
-#define CONFIG_ALC1019_BUFFER_SIZE       4096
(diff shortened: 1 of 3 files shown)


os/drivers/audio/alc1019.c
```diff
@@ -47,8 +47,12 @@
 #define ALC1019_I2S_TIMEOUT_MS 200
 
 /* Default configuration values */
+
+/*
+ * It's good to align the buffer size with i2s DMA page size
```
kishore-sn, line 52: I think you mean to say that the buffer size must match i2s size.
Can you pls change align --> match?

os/drivers/audio/alc1019.c
```diff
@@ -47,8 +47,12 @@
...
+ */
 #ifndef CONFIG_ALC1019_BUFFER_SIZE
-#define CONFIG_ALC1019_BUFFER_SIZE       4096
+#define CONFIG_ALC1019_BUFFER_SIZE       2048
```
Taejun-Kwon, line 55: dupliacted value defined in alc1019.h, please remove values in header file.

PR #6070, opened 2024-02-20
mem_leak_checker: Remove unnecessary static memory.

ewoodev:


os/kernel/debug/mem_leak_checker.c
```diff
@@ -246,7 +246,7 @@ static void heap_check(int *leak_cnt)
 					if ((leak_chk >= exclude_bottom && leak_chk <= exclude_top)) {
 						continue;
 					}
-					if (search_hash(*(unsigned long volatile *)leak_chk - (unsigned long)SIZEOF_MM_ALLOCNODE)) {
+					if (search_hash(*(unsigned long volatile *)leak_chk)) {
```
abhishek-samsung, line 263: I think this change is wrong. This logic is very similar to checking the data sections (like .data and .bss)  https://github.com/Samsung/TizenRT/blob/a67b9f7f062b443d4f778117f720af50ea557263/os/kernel/debug/mem_leak_checker.c#L215

Since, here, we are checking the contents of the node, we need to subtract SIZEOF_MM_ALLOCNODE to get the pointer to the alloc_node.

we are checking for cases like this,
```
int ** a;
a = malloc((sizeof(int *)) * 10);
for (int i = 0; i < 10; i++) a[i] = malloc(sizeof(int));
```

alloc node of a contains the pointers of a[0], a[1]..... , 
but the value of alloc node of a[0] which is present in the hash table  = a[0] - SIZEOF_MM_ALLOCNODE

ewoodev: yes, right.
I too found from my tests that this is the correct code
Thank you for reiew~

I am facing problem that detects wrong mem leak.
some memories that were previously confirmed as memory in the data and BSS areas are being detected as memory leaks.
As you reviewed, this code would not be the cause of the problem.
let me check more~ , Thank you~

os/kernel/debug/mem_leak_checker.c
```diff
@@ -52,27 +52,37 @@ struct alloc_node_info_s {
...
+static struct alloc_node_info_s **hash_table;
+static struct alloc_node_info_s *node_info;
 
 static void hash_init(void)
 {
 	int index;
+
+	DEBUGASSERT(hash_table == NULL && node_info == NULL);
+
+	hash_table = (struct alloc_node_info_s **)malloc(sizeof(struct alloc_node_info_s *) * HASH_SIZE);
```
Taejun-Kwon, line 64: It alloc properly for all the time?

os/kernel/debug/mem_leak_checker.c
```diff
@@ -89,9 +99,11 @@ static void add_hash(int index)
 static bool search_hash(unsigned long value)
 {
 	long key = value % HASH_SIZE;
-	struct alloc_node_info_s *cur = hash_table[key];
+	struct alloc_node_info_s *cur;
+
+	cur = hash_table[key];
```
Taejun-Kwon, line 104: you can asign directly as previous code. and now it use direct allocated memory, so null check for cur also need to be added in line 106

os/kernel/debug/mem_leak_checker.c
ewoodev: Thank you for review.
I added error handling

os/kernel/debug/mem_leak_checker.c
ewoodev: Thanks, changed

os/kernel/debug/mem_leak_checker.c
ewoodev: Thank you for review.
I added error handling

os/kernel/debug/mem_leak_checker.c
```diff
@@ -52,46 +52,66 @@ struct alloc_node_info_s {
...
-#ifdef CONFIG_MM_KERNEL_HEAP
-extern struct mm_heap_s g_kmmheap[CONFIG_KMM_NHEAPS];
-#endif
-
-static void hash_init(void)
+static int hash_init(void)
 {
 	int index;
+
+	DEBUGASSERT(g_hash_table == NULL && g_node_info == NULL);
```
sunghan-chang, line 62: Hmmmm. Is it necessary to be asserted? If it's not a null, then just return ok? Then, is there any problem?
Is there any chance to execute mem leak twice without end?

os/kernel/debug/mem_leak_checker.c
```diff
@@ -52,46 +52,66 @@ struct alloc_node_info_s {
...
+
+	g_hash_table = (struct alloc_node_info_s **)malloc(sizeof(struct alloc_node_info_s *) * HASH_SIZE);
+	if (!g_hash_table) {
+		return ERROR;
+	}
+
 	for (index = 0; index < HASH_SIZE; ++index) {
-		hash_table[index] = NULL;
+		g_hash_table[index] = NULL;
 	}
```
sunghan-chang, line 71: How about doing this after allocating the `g_node_info`? Because if the `g_node_info` has null, then no need to execute this.

os/kernel/debug/mem_leak_checker.c
```diff
@@ -52,46 +52,66 @@ struct alloc_node_info_s {
...
+	if (!g_node_info) {
+		free(g_hash_table);
+		return ERROR;
+	}
+	return OK;
+}
+
+static void hash_deinit(void)
+{
+	DEBUGASSERT(g_hash_table && g_node_info);
```
sunghan-chang, line 83: If we just return when they are null already, what happened?
There is no chance to call the `deinit` twice so that they should never be null?

os/kernel/debug/mem_leak_checker.c
ewoodev: yes It is checking already in run_mem_leak_checker function.
we don't need this.
```
int run_mem_leak_checker(int checker_pid, char *bin_name)
	if (g_hash_table || g_node_info) {
		printf("mem_leak_checker is already running.\n");
		return ERROR;
	}
```

os/kernel/debug/mem_leak_checker.c
ewoodev: thank you! I changed

os/kernel/debug/mem_leak_checker.c
ewoodev: oh right
You can't deinit twice or deinit without init.
I removed

sunghan-chang (approved):

PR #6072, opened 2024-02-22
os/board/rtl8730e: Fix build warning

ziyik:


os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -378,7 +378,7 @@ void i2s_set_direction(i2s_t *obj, int trx_type)
   */
 void i2s_set_param(i2s_t *obj, int channel_num, int rate, int word_len)
 {
-	uint32_t clock_mode;
+	uint32_t clock_mode = 0;
```
sunghan-chang, line 381: Is this a build warning fix? Not svace issue?

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_codec.c
```diff
@@ -2013,19 +2013,19 @@ void AUDIO_CODEC_SetMicBstChnMute(u32 amic_sel, u32 type, u32 newstate)
 		} else if (amic_sel == AMIC3) {
 			AUD->AUD_MICBST_CTL1 &= ~AUD_MASK_MICBST2_MUTE_L;
 			if (type == LINEIN) {
-				//RTK_LOGE(TAG, "AMIC3 NOT SUPPORT LINEIN MODE\n");
+				RTK_LOGE(TAG, "AMIC3 NOT SUPPORT LINEIN MODE\n");
```
sunghan-chang, line 2016: As far as I know, we don't use Realtek Audio. Because of that, we removed audio region in the RAM layout. Please confirm again whether this is necessary or not.

sunghan-chang:
Please leave warning logs and some explanation in the `commit` description.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
ziyik: Yes, this is a build warning.

CC:  component/mbed/targets/hal/rtl8730e/i2s_api.c
component/mbed/targets/hal/rtl8730e/i2s_api.c: In function 'i2s_set_param':
component/mbed/targets/hal/rtl8730e/i2s_api.c:421:40: warning: 'clock_mode' may be used uninitialized in this function [-Wmaybe-uninitialized]
  SP_InitStruct[obj->i2s_idx].SP_SelClk = clock_mode;
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_codec.c
ziyik: This is because some PINs are shared with Audio and are default configurated as Audio functions, so some Audio APIs are needed to disable the Audio PINs. (to enable to work as normal GPIO)
Only a few of the APIs are needed, compiler optimization will exclude the rest.

sunghan-chang (approved):

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_codec.c
sunghan-chang: That means this file is built? And audio is not used? Then how about removing the log you added? We need to optimize flash usage. Please check those logs are necessary or not.

ziyik: This file is built, but only a few APIs are included in build, the rest of the APIs are excluded. 
This is to be consistent with our default SDK, so those APIs are remain.

sunghan-chang: I think it's valid with the flat build. But It could be no in the loadable build. Could you verify that almost APIs are excluded in the loadable build?

ziyik: I have checked with laodable_apps_16M config, only 4 Audio APIs are included in the build, other unused APIs have been excluded from the build.

PR #6075, opened 2024-02-23
os/arch/arm/src/amebasmart: Remove Secondary core shutdown during cold boot phase

edwakuwaku:
- Remove core shutdown operation during cold boot


sunghan-chang:
Could you leave some explanation in the commit description? I think your comment in the code is enough.

sunghan-chang (commented):
And I suggest `......: remove cpu disable in coldboot`. Because I can't get what it is from the title. If we can't get it, actuallywe don't need a title.

edwakuwaku:
Thank you for your suggestion~ Can you please check again?

PR #6080, opened 2024-02-27
external/mbedtls/config.h: Enable MBEDTLS_AES_ROM_TABLES

jylee9613:
If MBEDTLS_AES_ROM_TABLES is disabled, aes_gen_tabels() calculate S-Box table and save it in the static buffers. 
However, S-Box table is unchanged unless the seed is changed, and there was no operation to change the seed. 
Therefore, this commit enables MBEDTLS_AES_ROM_TABLES to save S-Box tables as const (to remove unnecessary static memory)

### DIFF
--- a/external/include/mbedtls/config.h
+++ b/external/include/mbedtls/config.h
@@ -467,7 +467,7 @@
  *
  * Uncomment this macro to store the AES tables in ROM.
  */
-//#define MBEDTLS_AES_ROM_TABLES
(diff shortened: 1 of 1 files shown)


sunghan-chang (approved):
LGTM, but need to get confirm from Jinhwan.

sunghan-chang:
@j11591 Could you review this?

PR #6082, opened 2024-02-28
os/pm/Kconfig: Changes default values of state change threshold count

ritesh55555:
- lower the threshold from the current total transition time of 15 seconds which is too much to wait for
- Also board does not go to sleep if the total threshold time is more than 10 seconds
- This is a heuristic solution
- The threshold count for different states have been changed from (3,5,7) to (2,2,2)

### DIFF
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -200,7 +200,7 @@
 #endif
 
 #ifndef CONFIG_PM_IDLEENTER_COUNT
-#define CONFIG_PM_IDLEENTER_COUNT     30	/* Thirty IDLE slices to enter
+#define CONFIG_PM_IDLEENTER_COUNT     20	/* 20 IDLE slices to enter
 											 * IDLE mode from normal
 											 */
 #endif
@@ -218,7 +218,7 @@
 #endif
 
 #ifndef CONFIG_PM_STANDBYENTER_COUNT
-#define CONFIG_PM_STANDBYENTER_COUNT  50	/* Fifty IDLE slices to enter
+#define CONFIG_PM_STANDBYENTER_COUNT  20	/* 20 IDLE slices to enter
 											 * STANDBY mode from IDLE
 											 */
 #endif
@@ -236,7 +236,7 @@
 #endif
 
 #ifndef CONFIG_PM_SLEEPENTER_COUNT
-#define CONFIG_PM_SLEEPENTER_COUNT    70	/* 70 IDLE slices to enter SLEEP
+#define CONFIG_PM_SLEEPENTER_COUNT    20	/* 20 IDLE slices to enter SLEEP
 											 * mode from STANDBY
 											 */
 #endif
--- a/os/pm/Kconfig
+++ b/os/pm/Kconfig
@@ -136,7 +136,7 @@ config PM_IDLEENTER_THRESH
 	int "PM IDLE enter threshold"
 	default 1
 	---help---
-		State changes then occur when the weight activity account crosses
+		State changes occur when the weight activity account crosses
 		threshold values for certain periods of time (time slice count).
 
 			CONFIG_PM_xxxENTER_THRESH is the threshold value for entering state xxx.
@@ -151,7 +151,7 @@ config PM_IDLEEXIT_THRESH
 	int "PM IDLE exit threshold"
 	default 2
 	---help---
-		State changes then occur when the weight activity account crosses
+		State changes occur when the weight activity account crosses
 		threshold values for certain periods of time (time slice count).
 
 			CONFIG_PM_xxxENTER_THRESH is the threshold value for entering state xxx.
@@ -164,9 +164,9 @@ config PM_IDLEEXIT_THRESH
 
 config PM_IDLEENTER_COUNT
 	int "PM IDLE enter count"
-	default 30
+	default 20
 	---help---
-		State changes then occur when the weight activity account crosses
+		State changes occur when the weight activity account crosses
 		threshold values for certain periods of time (time slice count).
 
 			CONFIG_PM_xxxENTER_THRESH is the threshold value for entering state xxx.
@@ -181,7 +181,7 @@ config PM_STANDBYENTER_THRESH
 	int "PM STANDBY enter threshold"
 	default 1
 	---help---
-		State changes then occur when the weight activity account crosses
+		State changes occur when the weight activity account crosses
 		threshold values for certain periods of time (time slice count).
 
 			CONFIG_PM_xxxENTER_THRESH is the threshold value for entering state xxx.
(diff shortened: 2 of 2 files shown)


edwakuwaku:
Your current change is from 3+5+7 = 15secs -> 2+2+2 = 6secs
1) Do you mean more than 10sec?
2) May I know how do you get this results, or what kind of test have you conducted to come out with this observation?

ritesh55555:
> Your current change is from 3+5+7 = 15secs -> 2+2+2 = 6secs
> 
> 1. Do you mean more than 10sec?
> 2. May I know how do you get this results, or what kind of test have you conducted to come out with this observation?

1. yeah my bad, i mean more than 10sec
2. I have tested the board with real product refrigerator(taking latest code and pm configs enabled), i tried for several combination of threshold count and noticed whenever sum is more than 10 sec , board never goes to sleep . 
3. Furthermore, when i added extra logs , i could find board getting a periodic uart Tx callback every 10 sec before going to sleep, that calls the pm_activity() and changes board state to normal

edwakuwaku:
Understood, thanks for updating. This seems to be a product code behavior, which is not related to PM mechanism.
I guess 6 secs (2+2+2) is just a rough estimation, or is there any other reason you set this value?
Anyway, the change in value regarding the transition gap should be explained, even if it is just a temporary value.
At the final phase, those value need to be taken into account again, for power save optimization.

sunghan-chang:
@edwakuwaku Agree with you. It looks just temporary values so that I asked Ritesh and Amogh to do pm profiling for sleep and wakeup time. We should keep a sleep state as long as possible.
And, do we need the standby state? I think idle, normal, sleep are enough. And when we change normal -> idle, I think we don't need to get transit time. when idle -> sleep, then that time is necessary. Could you let me know your opinion?

os/pm/Kconfig
```diff
@@ -164,7 +164,7 @@ config PM_IDLEEXIT_THRESH
 
 config PM_IDLEENTER_COUNT
 	int "PM IDLE enter count"
-	default 30
+	default 20
 	---help---
 		State changes then occur when the weight activity account crosses
```
kishore-sn, line 169: `State changes then occur when` -> `PM state changes occur when`

sunghan-chang:
@ritesh55555 Let's leave `Your current change is from 3+5+7 = 15secs -> 2+2+2 = 6secs` this kind of explanation as well.

edwakuwaku:
Hi @sunghan-chang, indeed the pm profiling should be identified (together with the application/system scenario), which comes the first, then we will know how to design to the best extend for longer sleep period.
I think it is okay to have just 3 states (NORMAL, IDLE/STANDBY, SLEEP), we can minimize the transition gap, or merge both IDLE and STANDBY state into one directly. As this state will act as a last defender to check driver states before proceeding to sleep.
Other than that, as mentioned previously, when enabling PM, the PM loop will raise the power consumption around 2.5x higher, as compared to without enabling PM. This is not yet being addressed, there should be a counter-solution to handle this, or else we will face power wastage if there is some scenario that we expect to have a long period that we cannot go to sleep (ie. staying at PM_NORMAL), as far as I knew, we already have such scenario which is during onboarding, we will stay at PM_NORMAL for 10mins after BT init, before we going to sleep.

os/include/tinyara/pm/pm.h
```diff
@@ -200,7 +200,7 @@
 #endif
```
amogh-samsung, line 200: Lets prefer not to have a product based commit description
You can indicate via the description that the reason is to lower the threshold from the current total transition time of 15 seconds which is too much to wait for.

samsung-singh (approved):
changes default values, good to merge.

amogh-samsung:
> Understood, thanks for updating. This seems to be a product code behavior, which is not related to PM mechanism. I guess 6 secs (2+2+2) is just a rough estimation, or is there any other reason you set this value? Anyway, the change in value regarding the transition gap should be explained, even if it is just a temporary value. At the final phase, those value need to be taken into account again, for power save optimization.

The duration of 10 sec might be specific to product_refrigerator
Hence I suggest changing the commit description to indicate that 15 sec is too long to wait before sleeping, hence heuristically lowering it to 6 secs.

One other point I have, we expect that PM will mostly be beneficial in OOB state.
Lets say total time in OOB state is T, current drawn without PM is C
with PM, current becomes 2.5C, t1 is the time board is not sleeping, T-t1 is the time that board sleeps
(T x C) vs (t1 x C x 2.5) + ((T-t1) x C x 0.1)
We can estimate the ratio of t1/T if we know C, so that we know how much proportion of sleep time makes PM beneficial

edwakuwaku:
The current value different that I mentioned ~2.5x, is in idle state (bootup and do nothing), in onboarding cases, the ratio might be different. Additionally, when going to sleep, we need to do BLE advertisement, thus, the current ratio as compared to idle state is also not confirmed yet.
Anyway, please update the comments accordingly to indicate that this is a heuristic solution~

amogh-samsung:
> The current value different that I mentioned 2.5x, is in idle state (bootup and do nothing), in onboarding cases, the ratio might be different. Additionally, when going to sleep, we need to do BLE advertisement, thus, the current ratio as compared to idle state is also not confirmed yet. Anyway, please update the comments accordingly to indicate that this is a heuristic solution~

@ritesh55555 @edwakuwaku let's set up a test to determine the constants in the above equation for the OOB state so that we ca report a ratio of t1:T for PM

edwakuwaku:
> > The current value different that I mentioned 2.5x, is in idle state (bootup and do nothing), in onboarding cases, the ratio might be different. Additionally, when going to sleep, we need to do BLE advertisement, thus, the current ratio as compared to idle state is also not confirmed yet. Anyway, please update the comments accordingly to indicate that this is a heuristic solution~
> 
> @ritesh55555 @edwakuwaku let's set up a test to determine the constants in the above equation for the OOB state so that we ca report a ratio of t1:T for PM

I think we need input from @sunghan-chang here.

os/pm/Kconfig
ritesh55555: ok will update

ritesh55555:
> > The current value different that I mentioned 2.5x, is in idle state (bootup and do nothing), in onboarding cases, the ratio might be different. Additionally, when going to sleep, we need to do BLE advertisement, thus, the current ratio as compared to idle state is also not confirmed yet. Anyway, please update the comments accordingly to indicate that this is a heuristic solution~
> 
> @ritesh55555 @edwakuwaku let's set up a test to determine the constants in the above equation for the OOB state so that we ca report a ratio of t1:T for PM

Test setup to determine the constants are not present yet. Currently we can use logs only to infer knowledge about PM working on different threshold count and sleep time. We need to ensure that PM is working fine for different products and for different threshold counts.

edwakuwaku:
Hi @ritesh55555 , I think you forgot to update your commit description, also could you please add the change regarding the value of transition gap?

ritesh55555:
> Hi @ritesh55555 , I think you forgot to update your commit description, also could you please add the change regarding the value of transition gap?

Updated as required

amogh-samsung (approved):
LGTM
This can be the first heuristic update, more to come on the basis of profiling.

edwakuwaku (approved):
LGTM.

sunghan-chang (approved):

PR #6083, opened 2024-02-29
os/board/rtl8730e,os/arch/arm/src/amebasmart: Amebasmart USB host driver

kaikaikaime:
This includes USB host class CDC ACM
Changes in irq attachment, due to the replacement of M-cortex and A-cortex APcore, LSB 1 is not required for thumb architecture see comment in osdep_service_misc.c


os/board/rtl8730e/src/component/os_dep/osdep_service_queue.c
```diff
@@ -29,33 +32,139 @@ u32 rtw_end_of_queue_search(_list *head, _list *plist)
...
+	if (timeout_ms != 0xFFFFFFFF) {
+		struct timespec ts;
+		clock_gettime(CLOCK_REALTIME, &ts);
+		ts.tv_sec += timeout_ms / 1000;
+		ts.tv_nsec += (timeout_ms % 1000) * 1000 * 1000;
+		if (ts.tv_nsec >= NSEC_PER_SEC) {
+			ts.tv_sec ++;
+			ts.tv_nsec -= NSEC_PER_SEC;
+		}
+		if (mq_timedreceive((mqd_t) *queue, message, ((mqd_t) *queue)->msgq->maxmsgsize, &prio, &ts) == ERROR) {
```
Taejun-Kwon, line 116: Is there any specific reason why use mq among the kernel features? if you added this for yout testing, then please add some define or config only for your test

kaikaikaime: Hi @Taejun-Kwon ,

In USB operation, we used queue in our FreeRTOS OS, an alternative in Nuttx is mq.  Would there be issue using this?

Kai

Taejun-Kwon: because this will share some of result with Virtual driver of USB(dev/usb or something), I think callback mechanism is enough in kernel side.  Chracter Device, should share result with Application message queue because only IPC allowed between kernel & app

kaikaikaime: Hi Taejun,

I will try a work around.

Kai

PR #6089, opened 2024-03-07
build/tools/amebasmart: Fix ameba_hal_aes_decrypt() issue

ZhenBei-Sin:
Changes Note:
Issue:
1. Hang in 2nd time execute ameba_hal_aes_decrypt()
2. The AES encrypt and decrypt retrieved empty AES IV value

The information related to the factory key, including the IV, is loaded only during the SE initialization -> Load Factory Key.
At SE initialization, malloc is performed to create key storage, which includes for the IV.
This IV pointer is utilized for both encryption and decryption processes.

Root Cause: The IV pointer was malloc and free at the incorrect place (encryption and decryption processes).
Solution: The IV only be free when the SE is deinitialized or the key is removed.


aura1213 (approved):
LGTM

sunghan-chang (approved):

PR #6092, opened 2024-03-11
Fix SMP issues

kishore-sn:
Fix crash in sched_unlock for flat apps SMP
Fix assertion fail in irq_csection during smp repeat test
Fix crash in sched_setaffinity api for loadable apps smp


os/kernel/sched/sched_unlock.c
```diff
@@ -125,9 +125,14 @@ int sched_unlock(void)
...
 		/* Decrement the preemption lock counter */
-
 		if (rtcb->lockcount > 0) {
+			DEBUGASSERT(g_cpu_schedlock == SP_LOCKED && \
+					(g_cpu_lockset & (1 << cpu)) != 0);
+
 			rtcb->lockcount--;
+		} else {
+			DEBUGASSERT(g_cpu_schedlock == SP_UNLOCKED && \
+					(g_cpu_lockset & (1 << cpu)) == 0);
```
edwakuwaku, line 135: Hi @kishore-sn , how about moving this to line 146~148?

edwakuwaku: Additionally, the spin_clrbit in line 150~152 is redundant, since we will already face assert crash if the condition is violated.

os/kernel/task/task_exit.c
```diff
@@ -108,11 +108,15 @@ static void prepare_exit(struct tcb_s *tcb)
 #ifdef CONFIG_SMP
 	/* Make sure that the system knows about the locked state */
 
-	spin_setbit(&g_cpu_lockset, this_cpu(), &g_cpu_locksetlock, \
+	cpu_set_t lock = g_cpu_lockset;
+	lock &= (1 << this_cpu());
+	if (!lock) {
```
edwakuwaku, line 113: There will be 2 condition able to enter this condition,
1) No CPU hold the locks
2) Another CPU hold the locks
If we are in second condition, is it okay to just invoke spin_setbit directly?

edwakuwaku: I think some small changes can be made...?

`#ifdef CONFIG_SMP
	/* Make sure that the system knows about the locked state */
	cpu_set_t lock = g_cpu_lockset;
	lock &= (1 << this_cpu());
	if (!lock) {
		spin_setbit(&g_cpu_lockset, this_cpu(), &g_cpu_locksetlock, \
				&g_cpu_schedlock);
		lock = 1;
	}
#endif

	exit_tcb->lockcount++;

	mm_is_sem_available(exit_tcb);

	exit_tcb->lockcount--;
#ifdef CONFIG_SMP
	/* Make sure that the system knows about the locked state */
	if (exit_tcb->lockcount <= 0 && lock == 1) {
		spin_clrbit(&g_cpu_lockset, this_cpu(), &g_cpu_locksetlock, \
				&g_cpu_schedlock);
	}
#endif
}`

os/kernel/sched/sched_unlock.c
edwakuwaku: After thinking again, I think there should be no cases to enter sched_unlock with lockcount = 0, so it should be fine with current implementation~
Sorry for mistake, please ignore the comments above~

os/arch/arm/src/armv7-a/arm_cpuidlestack.c
```diff
@@ -141,6 +141,7 @@ int up_cpu_idlestack(int cpu, struct tcb_s *tcb, size_t stack_size)
   tcb->adj_stack_size  = SMP_STACK_SIZE;
   tcb->stack_alloc_ptr = (void *)stack_alloc;
   tcb->stack_base_ptr  = tcb->stack_alloc_ptr;
+  tcb->adj_stack_ptr = (void *)(stack_alloc + SMP_STACK_SIZE - 4);
```
neel-samsung, line 144: Don't we need to do this for other TCBs?

kishore-sn: For other tasks, it will be taken care when createstack() API is called. The issue is only with idle stack.

os/kernel/irq/irq_csection.c
```diff
@@ -644,4 +654,48 @@ bool irq_cpu_locked(int cpu)
 }
 #endif
 
+#ifdef CONFIG_SMP
+void restore_critical_section(void)
+{
+  /* NOTE: The following logic for adjusting global IRQ controls were
+   * derived from nxsched_add_readytorun() and sched_removedreadytorun()
```
edwakuwaku, line 661: Please check naming in comment

os/kernel/irq/irq_csection.c
```diff
@@ -644,4 +654,48 @@ bool irq_cpu_locked(int cpu)
...
+
+  /* Adjust global IRQ controls.  If irqcount is greater than zero,
+   * then this task/this CPU holds the IRQ lock
+   */
+
+  if (tcb->irqcount > 0)
+    {
+      /* Do notihing here
+       * NOTE: spin_setbit() is done in nxsched_add_readytorun()
+       * and nxsched_remove_readytorun()
```
edwakuwaku, line 677: similar as above

os/kernel/sched/sched_resumescheduler.c
```diff
@@ -74,46 +74,6 @@ void sched_resume_scheduler(FAR struct tcb_s *tcb)
...
-    {
-      /* Release our hold on the IRQ lock. */
-
-      if ((g_cpu_irqset & (1 << me)) != 0)
-        {
-          spin_clrbit(&g_cpu_irqset, me, &g_cpu_irqsetlock,
-                      &g_cpu_irqlock);
-        }
-    }
-#endif /* CONFIG_SMP */
```
edwakuwaku, line 116: You are moving this code piece out to form a new API --> restore_critical_section(). But here you are not invoking that API.
As I checked, there are quite some other places invoking sched_resume_scheduler(). After removing this code piece, the remaining operation left in this API is wrapped by macros, which mean we might be doing nothing here if all macros are disabled.
Could you please check whether a restore_critical_section() is needed here?

os/kernel/irq/irq_csection.c
```diff
@@ -245,6 +245,7 @@ irqstate_t enter_critical_section(void)
 			 * interrupt handler.
 			 */
 		} else {
+              int paused = false;
```
edwakuwaku, line 248: The modification in this file alignment seems incorrect, could you please check?

apps/examples/testcase/le_tc/kernel/itc_pthread.c
```diff
@@ -81,7 +82,8 @@ static void *thread_setgetname(void *param)
 */
 static void *thread_equal_check(void *param)
 {
-	if (pthread_equal(g_pthread, pthread_self()) == false) {
+	sleep(1);
```
sunghan-chang, line 85: It looks strange. Why do we need to sleep before execution without any conditional?

apps/examples/testcase/le_tc/kernel/tc_mqueue.c
```diff
@@ -596,6 +595,9 @@ static void tc_mqueue_mq_open_close_send_receive(void)
 	status = pthread_attr_setschedparam(&attr, &sparam);
 	TC_ASSERT_EQ("pthread_attr_setschedparam", status, OK);
 
+#ifdef CONFIG_SMP
+	CPU_SET(0, &(attr.affinity));
+#endif
```
sunghan-chang, line 600: Actually this is an app. Do we need to call this here?

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -100,6 +100,7 @@ static int delete_task(int argc, char *argv[])
 */
 static int restart_task(int argc, char *argv[])
 {
+	sleep(SEC_1); /* task will be restarted during this time, so counter will not be incremented */
```
sunghan-chang, line 103: same as previous. I don't know why this is necessary.

os/arch/arm/src/common/up_checkstack.c
```diff
@@ -174,7 +174,7 @@ static size_t do_stackcheck(uintptr_t start, uintptr_t end)
 
 size_t up_check_tcbstack(FAR struct tcb_s *tcb)
 {
-	return do_stackcheck((uintptr_t)tcb->stack_alloc_ptr, (uintptr_t)tcb->adj_stack_ptr);
+	return do_stackcheck((uintptr_t)tcb->stack_base_ptr, (uintptr_t)tcb->adj_stack_ptr);
```
sunghan-chang, line 177: How about stack coloration?
https://github.com/Samsung/TizenRT/blob/master/os/arch/arm/src/common/up_createstack.c#L266

os/kernel/sched/sched_resumescheduler.c
kishore-sn: I have few observations on this. 
First of all, I feel sched_resume_scheduler() is supposed to do only scheduler related work and the irqcount related code was placed wrongly in it. When we took the latest code from nuttx, it seems they have corrected this. 
Secondly, I see that resume_scheduler() API is called in following places. 

os/arch/arm/src/armv7-a/arm_releasepending.c
os/arch/arm/src/armv7-a/arm_cpustart.c
os/arch/arm/src/armv7-a/arm_exit.c
os/arch/arm/src/armv7-a/arm_unblocktask.c 	 	
os/arch/arm/src/armv7-a/arm_reprioritizertr.c
os/arch/arm/src/armv7-a/arm_cpupause.c  
os/arch/arm/src/armv7-a/arm_blocktask.c 

In each of the above cases, the API is called from IRQ context. But now we have moved the irqcount logic to doirq() and so, we would be still performing this check when we are returning from IRQ context. So, I think this is fine.

The only exception is arm_exit(). However, since this code is working well in nuttx and also all our tests seem to be fine, I think we can keep this change for now. If we face any issues in future, we can recheck this part.

os/kernel/task/task_exit.c
kishore-sn: Yes. You are right. I changed the whole logic to take the lock based on the current tcb lockcount.

os/kernel/sched/sched_resumescheduler.c
edwakuwaku: Thank you for detailed explanation, it make sense to me.
But, could you please add a remark or comment in the code piece here, since it might be a potential checkpoint?

os/kernel/task/task_exit.c
edwakuwaku: Looks good to me

apps/examples/testcase/le_tc/kernel/itc_pthread.c
kishore-sn: This change is no longer required. So, I am reverting it.

apps/examples/testcase/le_tc/kernel/tc_mqueue.c
kishore-sn: Previously we were facing issues in this testcase when smp is enabled. But, after the final smp changes in kernel, the testcase is passing without this change. So i will revert it.

apps/examples/testcase/le_tc/kernel/tc_task.c
kishore-sn: Reverted.

edwakuwaku:
Hi @kishore-sn , current loadable_apps enable CONFIG_PM by default. Thus, CONFIG_SMP shouldn't be enabled now, because hotplug mode implementation is pending. The logic for dual core + PM is also not verified and it requires some fix and changes I think.

kishore-sn:
> Hi @kishore-sn , current loadable_apps enable CONFIG_PM by default. Thus, CONFIG_SMP shouldn't be enabled now, because hotplug mode implementation is pending. The logic for dual core + PM is also not verified and it requires some fix and changes I think.

Ok. I will revert it for now. Let us enable it later.

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -100,6 +100,7 @@ static int delete_task(int argc, char *argv[])
 */
 static int restart_task(int argc, char *argv[])
 {
+	//sleep(SEC_1); /* task will be restarted during this time, so counter will not be incremented */
```
sunghan-chang, line 103: Need to keep this code with comment?

kishore-sn: Fixed.

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -97,43 +98,35 @@ static volatile spinlock_t g_cpu_paused[CONFIG_SMP_NCPUS];
...
-  return spin_islocked(&g_cpu_paused[cpu]);
+  return spin_is_locked(&g_cpu_paused[cpu]);
 }
 
 /****************************************************************************
- * Name: up_cpu_paused
+ * Name: up_cpu_paused_save
  *
  * Description:
  *   Handle a pause request from another CPU.  Normally, this logic is
```
edwakuwaku, line 108: The description here seems duplicated with up_cpu_paused, could you please check?

sunghan-chang:
@kishore-sn could you check Edward comment?
@r-prabu Once review is done, please merge this.

os/kernel/sched/sched_resumescheduler.c
kishore-sn: I have added a note in the commit msg.

edwakuwaku:
Hello @kishore-sn  @sunghan-chang , when we are merging this PR, please also review and merge https://github.com/Samsung/TizenRT/pull/6079 together. It contains fixes and changes required for SMP in BSP level.

os/arch/arm/src/armv7-a/arm_cpupause.c
kishore-sn: Done. Thanks.

edwakuwaku:
Hi @sunghan-chang , as discussed before, we might need some time until we finish integration for PM+SMP. In the meantime, will we create another defconfig for SMP first? So that we can verify it with onboarding and OTN process.
Hello @kishore-sn , could you please share your thoughts as well~?

kishore-sn:
> Hi @sunghan-chang , as discussed before, we might need some time until we finish integration for PM+SMP. In the meantime, will we create another defconfig for SMP first? So that we can verify it with onboarding and OTN process. Hello @kishore-sn , could you please share your thoughts as well~?

Yes. I agree. I have created a temporary defconfig with loadable apps SMP. However, currently there is a build error with this config which is related to PM. If we comment out "arm_gic_raise_softirq(1, 0);", then the build is successful.

edwakuwaku:
> > Hi @sunghan-chang , as discussed before, we might need some time until we finish integration for PM+SMP. In the meantime, will we create another defconfig for SMP first? So that we can verify it with onboarding and OTN process. Hello @kishore-sn , could you please share your thoughts as well~?
> 
> Yes. I agree. I have created a temporary defconfig with loadable apps SMP. However, currently there is a build error with this config which is related to PM. If we comment out "arm_gic_raise_softirq(1, 0);", then the build is successful.

Hi Kishore, currently you are enabling PM in the SMP config. It will have no problem because the state is locked in PM_NORMAL. the arm_gic_raise_softirq(1, 0) is rtk realization, in TizenRT, we should use arm_cpu_sgi(), both are serving the same purpose, which is to send signal (ie. SGI) to target cores under same cluster.
There will be 2 option:
1) Keep CONFIG_PM, but please don't allow state drop as of now
2) Disable CONFIG_PM
If we go for first option, you may comment out the code first, and I will make necessary changes in my PR. But still, the PM under SMP condition should not work as expected~

kishore-sn:
> > > Hi @sunghan-chang , as discussed before, we might need some time until we finish integration for PM+SMP. In the meantime, will we create another defconfig for SMP first? So that we can verify it with onboarding and OTN process. Hello @kishore-sn , could you please share your thoughts as well~?
> > 
> > 
> > Yes. I agree. I have created a temporary defconfig with loadable apps SMP. However, currently there is a build error with this config which is related to PM. If we comment out "arm_gic_raise_softirq(1, 0);", then the build is successful.
> 
> Hi Kishore, currently you are enabling PM in the SMP config. It will have no problem because the state is locked in PM_NORMAL. the arm_gic_raise_softirq(1, 0) is rtk realization, in TizenRT, we should use arm_cpu_sgi(), both are serving the same purpose, which is to send signal (ie. SGI) to target cores under same cluster. There will be 2 option:
> 
> 1. Keep CONFIG_PM, but please don't allow state drop as of now
> 2. Disable CONFIG_PM
>    If we go for first option, you may comment out the code first, and I will make necessary changes in my PR. But still, the PM under SMP condition should not work as expected~

@edwakuwaku 
We will use option 1. I have commented out the API call and created a temporary commit.

edwakuwaku:
Hi @kishore-sn , okay. Btw, don't you need to update .circleci/config.yml? We should include for server build as well.

edwakuwaku (approved):
Looking good.

sunghan-chang (approved):

PR #6094, opened 2024-03-12
os/board/rtl8730e: Add_BLE_indication_callback

yeetee179:
- added indication callback to inform the application layer of the current indication's outcome and the number of pending indications.
- fixed wrongly used parameter in p_func in TP1x that causes fault when CCCD is updated


hs36-kim (approved):
This is working on TP1x, AI-Dual .

PR #6095, opened 2024-03-12
os/board/rtl8730e,os/arch/arm/src/amebasmart: Smart SPI Master write …

kaikaikaime:
…to disable read

Kconfig allows disabling of read function for master write 
SPI Frequency prevents additional divider value


os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -77,12 +77,27 @@ config AMEBASMART_SPI0
 config AMEBASMART_SPI1
 	bool "SPI1"
 	default n
-
 config AMEBASMART_SPI_EXCHANGE
 	bool "Enable AMEBASMART SPI EXCHANGE"
 	default y
-endif
 
+config SPI_MASTER_WRITE_DISABLE_READ
```
sunghan-chang, line 84: 1. All others have `AMEBASMART_` prefix, but this is not. How about adding that here as well?
2. If I enable this, but disable AMEBASMART_SPI0_MASTER_WRITE_DISABLE_READ and AMEBASMART_SPI1_MASTER_WRITE_DISABLE_READ, what happen?
Do we need `SPI_MASTER_WRITE_DISABLE_READ` separately?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
```diff
@@ -622,16 +622,17 @@ void spi_frequency(spi_t *obj, int hz)
...
+	/* Adjust SCK Divider to an even number */
+	ClockDivider = (u32)(IpClk / hz / 2) * 2;
+
+	if ((IpClk / ClockDivider) > (u32)hz) {
+		ClockDivider += 2;
 	}
+
 	if (ClockDivider >= 0xFFFF) {
-		/*  devider is 16 bits */
+		/* devider is 16 bits */
```
sunghan-chang, line 633: Is this related the commit? If not, let's make a new commit for this.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
anjana348, line 1: This change restricts SPI frequency to < 50 MHz. Is it intended? Above 50MHz, it will create issue as ClockDivider will be 0 at line 626 and line 628 will cause division by 0.

kaikaikaime: Yes, this is correct, the maximum baudrate for SPI adapter is 50Mhz, 100Mhz is the base clock, therefore a basic division of 2 is a must. I can add an error message is frequency is above 50Mhz

kaikaikaime: I will create another PR for this change as recommendation from Chang

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
kaikaikaime: Will do

os/arch/arm/src/amebasmart/Kconfig
kaikaikaime: Modification done

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -857,8 +858,10 @@ static uint16_t amebasmart_spi_send(FAR struct spi_dev_s *dev, uint16_t wd)
 	DEBUGASSERT(priv);
 
 	if (priv->role == AMEBASMART_SPI_MASTER) {
-
-		ret = spi_master_write(&priv->spi_object, wd);
+		if (priv->mtr_wrrd_disable)
+			ret = spi_master_write(&priv->spi_object, wd, 1);
+		else
+			ret = spi_master_write(&priv->spi_object, wd, 0);
```
abhishek-samsung, line 864: I think this will affect other devices that are connected to the same spi port that use both tx and rx. So, this will not be a good approach. @Taejun-Kwon could you also please take a look at this change?

sunghan-chang: @abhishek-samsung @kaikaikaime @Taejun-Kwon
What I think is
1. The write function should do only write operation without read.
2. The exchange function should do both write and read.

So, LCD driver will use the write, not the exchange because the LCD does not support the read.
@abhishek-samsung @Taejun-Kwon Do you need the read at the write function? Can't you use the exchange?

@kaikaikaime If @abhishek-samsung and @Taejun-Kwon can use the exchange instead of the write, then, the write should do write only regardless of the config. (No need to add the config)

sunghan-chang: We should concern that some SPI device does not support read operation.

ewoodev:
```
static const struct spi_ops_s g_spi1ops = {
...
	.send         = amebalite_spi_send,
#ifdef CONFIG_SPI_EXCHANGE
	.exchange     = amebalite_spi_exchange,
#else
	.sndblock     = amebalite_spi_sndblock,
	.recvblock    = amebalite_spi_recvblock,
#endif
...
}
```
There is send api without exchange in SPI ops interface

How about support send of amebasmart_spi which not related to exchange.

abhishek-samsung: Currently when spi exchange config is enabled, snd and rcv are both use exchange internally. So, it needs to be split accordingly and we need to only modify the snd api of SPI without changing anything else.
https://github.com/Samsung/TizenRT/blob/1601a36b77dc7bdee54f3b72876213b6b56c2e8e/os/include/tinyara/spi/spi.h#L281-L285
So, I think instead of a config, it would be good if @kaikaikaime would provide us a snd API and plug it in ```amebalite_spi_sndblock``` like Eunwoo suggested. Then we will need to make the upper layer changes on our side.

sunghan-chang: @kaikaikaime Sorry but this is urgent. Could you make a change ASAP?

kaikaikaime: Hi all, based on the earlier discussion, this will be my following actions, do correct me if I get them wrong:
1. I will create another internal write function which does not run read function ()
2. Using this new internal write and place them in SPI_SNDBLOCK when CONFIG_SPI_EXCHANGE is disabled.

Kai

kaikaikaime: With a new creation of spi master send with read remove, I can now place under the condition when exchange is undefined. Change is minimal.

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -857,8 +857,11 @@ static uint16_t amebasmart_spi_send(FAR struct spi_dev_s *dev, uint16_t wd)
 	DEBUGASSERT(priv);
 
 	if (priv->role == AMEBASMART_SPI_MASTER) {
-
+#ifndef CONFIG_SPI_EXCHANGE
```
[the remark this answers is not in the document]
ewoodev:
```
static const struct spi_ops_s g_spi0ops = {
	.send         = amebasmart_spi_send,
```

amebasmart_spi_send is used by amebasmart_spi_exchange
We need a send operation that operates independently of the exchange.

Please separate the send API so that it is not related to exchange.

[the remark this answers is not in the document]
kaikaikaime: I understood that need to create a new function amebasmart_send_read_disabled() ...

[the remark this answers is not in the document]
sunghan-chang: @kaikaikaime Please refer other chip architecture.
imxrt
```
	.send         = imxrt_lpspi_send,
#ifdef CONFIG_SPI_EXCHANGE
	.exchange     = imxrt_lpspi_exchange,
#else
	.sndblock     = imxrt_lpspi_sndblock,
	.recvblock    = imxrt_lpspi_recvblock,
#endif
```
esp32
```
	.send = spi_send,
#ifdef CONFIG_SPI_EXCHANGE
	.exchange = spi_exchange,
#else
	.sndblock = spi_sndblock,
	.recvblock = spi_recvblock,
#endif
```

All support `send` separately from exchange.
We want the `send` regardless of the exchange.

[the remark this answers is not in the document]
sunghan-chang: The `send` should not have the receive internally.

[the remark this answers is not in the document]
kaikaikaime: amebasmart_spi_send will now only do internal function write and does not do return read 
internal function spi_master_exchange used in exchange will now do internal write and read 

Kai

kaikaikaime:
Hi, with the SPI DMA implementation in the PR, do you still plan to merge this master send only ?

PR #6096, opened 2024-03-12
os/pm: Adds PM functionalities for sending Wifi keep alive

ritesh55555:
- This commits add methods to help board wake up for sending wifi keep alive signal. It stores time ticks of last time when wifi keep alive signal was sent and accordingly sets the sleep duration of board. So that board can wake up when next signal is supposed to be sent
- power_set_next_wakeup_interval() is added . This can be called from app side just after "wifi keep alive signal was sent" . This will set the WAKE_UP_TIMER again with the required time interval. This function will also store the time tick of the Wifi keep alive signal sent time. This can be used to calculate actual required sleep duration just before sleep.
- power_get_last_wifi_alive_send_time and power_get_current_time can be used to find difference between the time ticks of last wifi keep alive signal sent and now .
- I have also changed the threshold values of entering each state as (10 + 10 + 10)  , which is total of 3 sec from Normal to sleep. It is still a heuristic solution . This has been done to let board able to sleep some time before waking up after 10 sec.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -130,7 +130,7 @@ void pg_timer_int_handler(void *Data)
 			break;
 	}
 	// Reset the global struct
-	g_pm_timer.timer_type = 0;
+	g_pm_timer.timer_type = PM_NO_TIMER;
 	g_pm_timer.timer_interval = 0;
 }
 
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -244,6 +244,9 @@
 /* Defines max length of device driver name for PM callback. */
 #define MAX_PM_CALLBACK_NAME    32
 
+/* Defines default sleep duration */
+#define DEFAULT_PM_SLEEP_DURATION    10000000 /* 10 sec in microsecond*/
+
 #define PM_LOCK_PATH					"/proc/power/domains/0/pm_lock"
 #define PM_UNLOCK_PATH					"/proc/power/domains/0/pm_unlock"
 #ifdef CONFIG_PM_DVFS
@@ -304,6 +307,7 @@ enum pm_state_e {
 };
 
 enum pm_timer_type_e {
+	PM_NO_TIMER = 0,
 	PM_WAKEUP_TIMER = 1,
 	PM_LOCK_TIMER = 2,
 	/* Scope for future expansion, up to 8 timer types can be supported */
@@ -312,6 +316,7 @@ enum pm_timer_type_e {
 struct pm_timer_s {
 	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
 	uint32_t timer_interval;	/* The interval for whichever timer is to be used */
+	clock_t last_wifi_alive_send_time;       /* Time tick of the last wifi keep alive signal sent time */
 };
 
 /* This structure contain pointers callback functions in the driver.  These
@@ -547,6 +552,24 @@ void pm_relax(int domain, enum pm_state_e state);
  ****************************************************************************/
 
 void pm_set_timer(int timer_type, size_t timer_interval);
+
+/****************************************************************************
+ * Name: pm_adjust_sleep_duration
+ *
+ * Description:
+ *   This function is called just before sleep to calculate exact required
+ *   sleep duration "if needed". This function is used to timely wake up board
+ *   so that we can sent the next wifi keep alive signal.
+ *
+ * Input Parameters:
+ *   None
+ *
+ * Returned Value:
+ *   None.
+ *
+ ****************************************************************************/
+
+void pm_adjust_sleep_duration(void);
 
 /****************************************************************************
  * Name: pm_staycount
--- a/os/pm/pm_changestate.c
+++ b/os/pm/pm_changestate.c
@@ -299,6 +299,13 @@ int pm_changestate(int domain_indx, enum pm_state_e newstate)
 		*/
 		pm_changeall(domain_indx, newstate);
 		pdom->state = newstate;
+		
+		/* Adjusting the sleep duration, so that system can revert to NORMAL
+		*  state when next Wifi keep alive signal is required to be sent. 
+		*/
+		if (newstate == PM_SLEEP) {
+			pm_adjust_sleep_duration();
+		}
 
 	}
 EXIT:
--- a/os/pm/pm_procfs.c
+++ b/os/pm/pm_procfs.c
@@ -74,6 +74,7 @@
 #include <tinyara/fs/fs.h>
 #include <tinyara/fs/procfs.h>
 #include <tinyara/fs/dirent.h>
+#include <tinyara/clock.h>
 
 #include <tinyara/pm/pm.h>
 
@@ -117,7 +118,7 @@ struct power_procfs_entry_s {
 };
 
 /* Added to record timer countdown interrupt */
-struct pm_timer_s g_pm_timer = { 0, 0 };
+struct pm_timer_s g_pm_timer = { 0, 0, 0};
 
 /****************************************************************************
  * Private Function Prototypes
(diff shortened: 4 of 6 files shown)


samsung-singh (approved):
Adding PM functionalities for sending Wifi keep alive, good to merge.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -132,9 +133,24 @@ void pg_timer_int_handler(void *Data)
 	// Reset the global struct
 	g_pm_timer.timer_type = 0;
 	g_pm_timer.timer_interval = 0;
+	g_pm_timer.last_wakeup_time = 0;
 }
 
 void up_set_pm_timer(void) {
+
```
sunghan-chang, line 140: This is a pm timer for BSP. But what you added is a pm general functionality. Are you able to move them into the pm? not bsp?

ritesh55555: Yes correct , I moved it to PM .

os/pm/pm_procfs.c
```diff
@@ -520,6 +527,85 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
...
+ *   filep          - File path of the power domain
+ *   buffer         - Not used
+ *   buflen         - Not used
+ *
+ * Returned Value:
+ *  Time tick value of the last wakeup time.
+ ****************************************************************************/
+
+static size_t power_get_last_wakeup_time(FAR struct file *filep, FAR char *buffer, size_t buflen)
+{
```
edwakuwaku, line 581: void the input argument if not used
```
(void)buffer;
(void)buflen;
```

os/pm/pm_procfs.c
```diff
@@ -520,6 +527,85 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
...
+ *   filep          - File path of the power domain
+ *   buffer         - Not used
+ *   buflen         - Not used
+ *
+ * Returned Value:
+ *  Time tick value of current time
+ ****************************************************************************/
+
+static size_t power_get_current_time(FAR struct file *filep, FAR char *buffer, size_t buflen)
+{
```
edwakuwaku, line 603: similar as above

edwakuwaku:
Is this verfied?

edwakuwaku:
The global struct is not updated:
`struct pm_timer_s g_pm_timer = { 0, 0 };` -> `struct pm_timer_s g_pm_timer = { 0, 0 , 0};`

ritesh55555:
> Is this verfied?

Yes, I have tested it on product. It is able to wake up and send Wifi keep alive every 10 sec and sleep inbetween. But sleep time is not constant . It usually varies from 3sec to 7sec as seen from cases of testing.
I have tested it on two products and it is working fine.

ritesh55555:
@amogh-samsung @edwakuwaku @sunghan-chang , We would need a good method to test average power consumption during OOB state because as board keeps waking up and sleep every 10 second, the power will fluctuate.

edwakuwaku (approved):
LGTM.

os/pm/pm_procfs.c
```diff
@@ -520,6 +527,94 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
...
+ * Returned Value:
+ *  0 (OK) means the state unlock was successful.
+ ****************************************************************************/
+ 
+static size_t power_set_next_wakeup_interval(FAR struct file *filep, FAR const char *buffer, size_t buflen)
+{
+	(void)buffer;
+
+	g_pm_timer.last_wakeup_time = clock_systimer();
+	if (g_pm_timer.timer_type == 0) {
```
edwakuwaku, line 558: `if (g_pm_timer.timer_type == 0 && buflen > 0)`

ritesh55555: done

os/pm/pm_changestate.c
```diff
@@ -299,6 +299,13 @@ int pm_changestate(int domain_indx, enum pm_state_e newstate)
 		*/
 		pm_changeall(domain_indx, newstate);
 		pdom->state = newstate;
+		
+		/* Adjusting the sleep duration, so that system can revert to NORMAL
+		*  state when next Wifi keep alive signal is required to be sent. 
+		*/
+		if (newstate == PM_SLEEP) {
+			pm_adjust_sleep_duration();
+		}
```
amogh-samsung, line 308: I assume that timer types are still as before
i.e. PM_WAKEUP_TIMER and PM_LOCK_TIMER

As per this logic, the changestate function does not differentiate between whether the next wakeup is for the wifi alive signal, or just any other wakeup requested by another APP/source.
It will always adjust the time according to the wifi signal wakeup requirement, but the wakeup may not necessarily be for sending the wifi alive signal only.

You could add a new timer type PM_WIFI_ALIVE_WAKEUP or something and adjust the sleep time in that case only.

os/pm/pm_procfs.c
```diff
@@ -520,6 +527,92 @@ static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer,
...
+ * Returned Value:
+ *  0 (OK) means the state unlock was successful.
+ ****************************************************************************/
+ 
+static size_t power_set_next_wakeup_interval(FAR struct file *filep, FAR const char *buffer, size_t buflen)
+{
+	(void)buffer;
+
+	g_pm_timer.last_wakeup_time = clock_systimer();
+	if (g_pm_timer.timer_type == 0 && buflen > 0) {
```
amogh-samsung, line 558: Since we have an enum for the time types
Use the enum name itself instead of 0 to directly indicate the timer type and make the code easier to understand,.

os/pm/pm_procfs.c
```diff
@@ -146,6 +147,9 @@ static size_t power_metrics_read(FAR struct file *filep, FAR char *buffer, size_
 static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_t buflen);
 static size_t power_lock_write(FAR struct file *filep, FAR const char *buffer, size_t buflen);
 static size_t power_unlock_write(FAR struct file *filep, FAR const char *buffer, size_t buflen);
+static size_t power_set_next_wakeup_interval(FAR struct file *filep, FAR const char *buffer, size_t buflen);
+static size_t power_get_last_wakeup_time(FAR struct file *filep, FAR char *buffer, size_t buflen);
+static size_t power_get_current_time(FAR struct file *filep, FAR char *buffer, size_t buflen);
```
amogh-samsung, line 152: From these 2 functions, it seems to me that they are doing very little, and just returning the global pm timer parameters to the APP side.
Then the APP will use these returned values to set the next wakeup interval after the current wifi alive signal has been sent.

If my understanding here is correct, then I suggest avoiding these 2 functions which just return a parameter to the APP and have the main timer setting function power_set_next_wakeup_interval itself do the work.
The logic for identifying the wifi signal wakeup time based on elapsed time can be brought into the PM layer, where the parameters are available anyways.

Also, yet again, since this function is intended for setting the sleep interval for wifi alive signal wakeups, the name of the function may be changed to indicate that
(power_set_wifi_alive_wakup_interval OR power_set_signal_wakeup_interval maybe since we expect more such alive signals in the future to be handled).

Lastly, procfs needs to be substituted by a driver dedicated to PM, so these comments may be considered there instead.

os/pm/pm_changestate.c
edwakuwaku: Hi @amogh-samsung , but I think the operation here is fine, as the wifi keep alive is a signal which have to be sent periodically. Thus, even if other app request for different timer duration, the baseline to wakeup for sending wifi keep alive need to be retained, which I think why we do consideration here so that we won't miss out sending keepalive signal.
Since wifi keep alive signal is a periodic signal, then we should set a periodic timer to handle that~

os/pm/pm_procfs.c
ritesh55555: Actually, 0 doesnot have any enum name. timer_type == 0 means timer has not set.

os/include/tinyara/pm/pm.h
```diff
@@ -312,6 +312,7 @@ enum pm_timer_type_e {
 struct pm_timer_s {
 	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
 	uint32_t timer_interval;	/* The interval for whichever timer is to be used */
+	clock_t last_wakeup_time;       /* Time tick of last wake up time */
```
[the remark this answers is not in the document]
ritesh55555: updated

os/pm/pm_procfs.c
ritesh55555: I should add these when driver dedicated to PM is created.

os/pm/pm_procfs.c
amogh-samsung: 0 then can be added as an enumerated value to avoid using the hardcoded value '0'?

ritesh55555: updated

os/pm/pm_changestate.c
ritesh55555: Hi, @edwakuwaku @amogh-samsung  , For now i am just resetting the wake up timer just before sleep every time. So it behaves as periodic .

sunghan-chang (approved):

PR #6100, opened 2024-03-18
os/arch/arm/src/amebasmart: Fix SPI clock divider to be even number

kaikaikaime:
Fix the divider to be even, as a requirement of the specs sheet

### DIFF
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
@@ -622,16 +622,17 @@ void spi_frequency(spi_t *obj, int hz)
 
 	IpClk = 100000000;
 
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@kaikaikaime I can't get which freq from the title. How about below?
```
arch/arm/amebasmart: Improve SPI frequency divider precision
```
And I can't get how it is better with this change. Could you add some example on the description?

kaikaikaime:
> @kaikaikaime I can't get which freq from the title. How about below?
> 
> ```
> arch/arm/amebasmart: Improve SPI frequency divider precision
> ```
> 
> And I can't get how it is better with this change. Could you add some example on the description?

Hi Chang, after checking, this is not an improvement but a fix, as divider are mandated to be divisible by even number

sunghan-chang (approved):

sunghan-chang (approved):

PR #6101, opened 2024-03-20
fix CONFIG_LCD_LANDSCAPE orientation

anjana348:
The resolution values were swapped twice in driver and app when CONFIG_LCD_LANDSCAPE=y 
Fixed by removing swapping from app.


sunghan-chang (commented):
Let's split the commit for lcd driver and app.
and each commit should have a location at start of a title.
```
1. 
os/driver/lcd: blah blah

need explanation of the commit as description

2.
example/lcd_test: blah blah

need explanation of the commit as description
```

sunghan-chang (approved):

PR #6102, opened 2024-03-20
os/board/rtl8730e/src, os/arch/arm/src/amebasmart: Fix Flash operation hang issue

edwakuwaku:
- Added check spic busy condition before entering user mode for spic
- Remove rom code from build
- Remove write lock/unlock for flash read
- Revise ipc semaphore type instead of void

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_flash.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_flash.c
@@ -270,6 +270,7 @@ static ssize_t amebasmart_bwrite(FAR struct mtd_dev_s *dev, off_t startblock, si
 static ssize_t amebasmart_read(FAR struct mtd_dev_s *dev, off_t offset, size_t nbytes, FAR uint8_t *buffer)
 {
 	ssize_t result;
+	DEBUGASSERT(nbytes > 0);
 	result = amebasmart_flash_read(CONFIG_AMEBASMART_FLASH_BASE + offset, buffer, nbytes);
 	return result < 0 ? result : nbytes;
 }
@@ -284,6 +285,7 @@ static ssize_t amebasmart_write(FAR struct mtd_dev_s *dev, off_t offset, size_t 
 {
 	size_t addr;
 	ssize_t result;
+	DEBUGASSERT(nbytes > 0);
 	addr = CONFIG_AMEBASMART_FLASH_BASE + offset;
 	result = amebasmart_flash_write(addr, buffer, nbytes);
 	return result < 0 ? result : nbytes;
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
@@ -130,6 +130,7 @@ int flash_get_status(flash_t *obj)
   * @retval none
   * @note LSB 12bits will be masked.
   */
+SRAMDRAM_ONLY_TEXT_SECTION
 void flash_erase_sector(flash_t *obj, u32 address)
 {
 	/* To avoid gcc warnings */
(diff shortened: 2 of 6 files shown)


edwakuwaku:
Only apply at irqsave :
    - OTN test : [OTN Success]
    - power cut test : [Success]
    - kernel binary OTN repeat test : [OTN Success]
   Apply at both irqsave and irqrestore : 
    - OTN test : [OTN Success]
    - power cut test : [Success]
    - kernel binary OTN repeat test : [OTN Success]

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
```diff
@@ -275,15 +273,13 @@ int  flash_write_word(flash_t *obj, u32 address, u32 data)
...
-  * @retval 1: Success.
-  * @retval Others: Error.
+  * @retval len: Total length of bytes read.
   * @note It is ok to read under auto mode due to flash cache.
   */
 int  flash_stream_read(flash_t *obj, u32 address, u32 len, u8 *data)
 {
 	/* To avoid gcc warnings */
 	(void) obj;
-	FLASH_Write_Lock();
```
r-prabu, line 286: can we add the zero length check at the beginning of the function?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
```diff
@@ -366,8 +361,7 @@ int  flash_stream_read(flash_t *obj, u32 address, u32 len, u8 *data)
   * @param  len: Length of the data to write.
   * @param  data: Pointer to a byte array that is to be written.
   * @return Write result.
-  * @retval 1: Success.
-  * @retval Others: Error.
+  * @retval size: Total length of bytes written.
   */
```
r-prabu, line 365: can we add the zero length check at the beginning of the function?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/flash_api.c
edwakuwaku: Yes sure, I have added, can you help to check again~?

os/arch/arm/src/amebasmart/amebasmart_flash.c
```diff
@@ -284,6 +285,7 @@ static ssize_t amebasmart_write(FAR struct mtd_dev_s *dev, off_t offset, size_t
 {
 	size_t addr;
 	ssize_t result;
+	DEBUGASSERT(nbytes > 0);
```
sunghan-chang, line 288: Do we need ASSERT in this case? return 0 is not enough?

edwakuwaku: Sorry I missed out that datatype, seems the only case to be considered here is 0.
If we are just checking for 0, then I think return warning message should be enough. In my opinion, it should be tolerable as we can just do nothing and return from there

sunghan-chang: So, final your opinion is?
I will follow yours because it's a DEBUGASSERT.

edwakuwaku: Okay, let's go with DEBUGASSERT. Since return error and adding error logs will also induce more code size.

sunghan-chang (approved):

sunghan-chang:
@edwakuwaku Please update this once you verify this with Dane.

edwakuwaku:
Hi @sunghan-chang, yes we will~

edwakuwaku:
Hello @sunghan-chang 님, sorry for late update. With this latest change, so far we tested with single core on 3 cases, which are
1) smartfs_powercut
2) onboarding
3) OTN
Above test cases are working fine so far
For SMP cases, as it is not yet merge all the fixes, so I manually added all the changes and test on smartfs_powercut example first, and I did not identify similar flash hang issue anymore (ie. bus fault causing hang). Once all thing is verified, we will proceed to test with onboarding and OTN for SMP case.
During the process, there is some deadlock issue identified, which the fix has been included in another PR: https://github.com/Samsung/TizenRT/pull/6079

sunghan-chang (dismissed):

sunghan-chang:
@edwakuwaku Could you resolve the conflict ASAP? When it's resolved, I will merge this with PR #6159 .

edwakuwaku:
Hi @sunghan-chang, I have rebased it~

sunghan-chang (approved):

PR #6105, opened 2024-03-20
os/driver/lcd: set orientation in st7789_lcdintialize

anjana348:
orientaton using config was set in init ioctl for lcd it is now set in st7789_lcdinitialize as well if init ioctl is not called

### DIFF
--- a/os/drivers/lcd/st7789.c
+++ b/os/drivers/lcd/st7789.c
@@ -783,17 +783,17 @@ static int st7789_init(FAR struct lcd_dev_s *dev)
 	st7789_sendcmd(priv, ST7789_DISPON);
 	lcd_unlock(priv->spi);
 
-	#if defined(CONFIG_LCD_PORTRAIT)
-		st7789_setorientation(priv, LCD_PORTRAIT);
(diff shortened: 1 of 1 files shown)


os/drivers/lcd/st7789.c
```diff
@@ -910,6 +910,17 @@ FAR struct lcd_dev_s *st7789_lcdinitialize(FAR struct spi_dev_s *spi, struct st7
 	st7789_init_cmd(priv);
 	lcd_unlock(priv->spi);
 
+	#if defined(CONFIG_LCD_PORTRAIT)
+		st7789_setorientation(priv, LCD_PORTRAIT);
```
sunghan-chang, line 914: For indentation,
```
	#if defined(CONFIG_LCD_PORTRAIT)
		st7789_setorientation(priv, LCD_PORTRAIT);

-> 

#if defined(CONFIG_LCD_PORTRAIT)
	st7789_setorientation(priv, LCD_PORTRAIT);
```

sunghan-chang (approved):

PR #6106, opened 2024-03-21
os/net/wifi: add an API for scanning multiple specified APs at once

Jw00:
Since multiple AP profiles can be stored and used, it must be possible to scan by specifying multiple APs.


jwei5:
Ok, we will use this struct for scanning multiple specified APs.
-RTK

jwei5:
Hi @Jw00, may I check if this PR will also include the API for scanning multiple APs in wifi_manager_api.c?
Thank you.

Jw00:
> Hi @Jw00, may I check if this PR will also include the API for scanning multiple APs in wifi_manager_api.c? Thank you.

As you can see, there is not any modification in `wifi_manager_api.c`.
Should it is necessary?

jwei5:
> > Hi @Jw00, may I check if this PR will also include the API for scanning multiple APs in wifi_manager_api.c? Thank you.
> 
> Hi @jwei5 As you can see, there is not any modification in `wifi_manager_api.c`. Should it is necessary?

Hi @Jw00,
From my understanding, wifi_manager is how the higher/application layers in TizenRT call into the lower layer APIs. 
As this multiple AP scan feature is a separate API from the normal AP scan and is also a new method in trwifi_ops, I think a multiple scan API is required in wifi_manager to call into the lower layers.

One such modification might be the enum lwnl_req_wifi in tinyara\net\if\wifi.h, I think there needs to be another LWNL_REQ_WIFI_ to match the multiple ap scan method in struct "trwifi_ops"

Is my understanding correct?

Jw00:
@jwei5 Thank you for your detail description, I will check it.

framework/src/wifi_manager/wifi_manager_state.c
```diff
@@ -320,6 +321,55 @@ wifi_manager_result_e _wifimgr_scan(wifi_manager_scan_config_s *config)
 	return WIFI_MANAGER_SUCCESS;
 }
 
+static wifi_manager_result_e _wifimgr_scan_multi_aps(wifi_manager_scan_multi_configs_s *configs)
+{
+	if (!configs) {
+		WIFIMGR_CHECK_UTILRESULT(wifi_utils_scan_multi_aps(NULL), TAG,
+								 "request scan multi aps to wifi utils is fail");
+		return WIFI_MANAGER_SUCCESS;
```
sunghan-chang, line 329: Debug message says `is fail` but return value is `SUCCESS`. Is it correct?

Jw00: > Debug message says `is fail` but return value is `SUCCESS`. Is it correct?

Yes.
That message is for when the `wifi_utils_scan_multi_aps` execution fails. You can see `WIFIMGR_CHECK_UTILRESULT` macro.

Jw00:
@jwei5 I made that you described, would you check it?

framework/src/wifi_manager/wifi_manager_api.c
```diff
@@ -210,6 +210,26 @@ wifi_manager_result_e wifi_manager_scan_ap(wifi_manager_scan_config_s *config)
 	RETURN_RESULT(wifimgr_post_message(&msg), msg);
 }
 
+wifi_manager_result_e wifi_manager_scan_multi_aps(wifi_manager_scan_multi_configs_s *configs)
+{
+	NET_LOGI(TAG, "--> %s %d\n", __FUNCTION__, __LINE__);
+	if (configs) {
```
jwei5, line 216: May I ask if we should do a check here on whether config is NULL?
RTK's implementation of scanning multiple AP API requires config to be passed in, if it is NULL the API will fail.

jwei5 (commented):
Hi @Jw00 , I have checked it, could you also check my comments?
I applied our multiple AP scan API on this PR and had to make the changes mentioned in the comments for it to work

framework/src/wifi_manager/wifi_manager_state.c
```diff
@@ -320,6 +321,55 @@ wifi_manager_result_e _wifimgr_scan(wifi_manager_scan_config_s *config)
...
+	}
+
+	trwifi_scan_multi_configs_s uconf = {0};
+	memset(&uconf, 0, sizeof(trwifi_scan_multi_configs_s));
+	for (int i = 0; i < configs->scan_ap_config_count; i++) {
+		wifi_manager_scan_config_s *config = &configs->ap_configs[i];
+		if (config->ssid_length > 0) {
```
jwei5, line 336: I see that ssid_length was checked in wifi_manager_scan_multi_aps(), do we need to check it again here?

jwei5:
Hi @Jw00, could I ask about the expected behavior for `wifi_manager_scan_multi_aps(wifi_manager_scan_multi_configs_s *configs)`?
Must the user provide the AP's channel number to the API? 
If the desired behavior is only to scan for multiple specific APs, I think we may not need to check `config->channel` in the API.

Also, do you need a copy of our multiple AP scan API patch for testing this PR?

framework/src/wifi_manager/wifi_manager_api.c
Jw00: @jwei5 My opinion is that is better to operate in the same way as single AP scan.
In the future, the single AP scan command may be completely replaced to the multi scan.

framework/src/wifi_manager/wifi_manager_state.c
Jw00: Here, the purpose is to determine whether or not to copy the SSID string.

Jw00:
> Hi @Jw00 , I tried to run the multi scan API with the latest changes. It is working well after I temporarily added the handling of WIFIMGR_CMD_SCAN_MULTI_APS event in wifi_manager_state.c to call into the API. image May I check if you will be modifying the event handlers in this PR? Thanks.

I added that you checked, thank you.

Jw00:
> Hi @Jw00 , I tried to run the multi scan API with the latest changes. It is working well after I temporarily added the handling of WIFIMGR_CMD_SCAN_MULTI_APS event in wifi_manager_state.c to call into the API. image May I check if you will be modifying the event handlers in this PR? Thanks.

I added that you checked, thank you.

jwei5:
> > Hi @Jw00 , I tried to run the multi scan API with the latest changes. It is working well after I temporarily added the handling of WIFIMGR_CMD_SCAN_MULTI_APS event in wifi_manager_state.c to call into the API. image May I check if you will be modifying the event handlers in this PR? Thanks.
> 
> I added that you checked, thank you.

Thanks @Jw00 , can you also add it to the other states that will use WIFIMGR_CMD_SCAN_MULTI_APS event?  
I can only think of **on_connected_state** and **on_softap_state**, not sure if I missed out any other states.

Jw00:
> 
> Thanks @Jw00 , can you also add it to the other states that will use WIFIMGR_CMD_SCAN_MULTI_APS event? I can only think of **on_connected_state** and **on_softap_state**, not sure if I missed out any other states.

Sure, I will check it, thank you, @jwei5 .

Jw00:
@jwei5 , I have checked the other states, would you check the latest code?

Jw00:
@hk-gwak , you need to check this PR, also.

jwei5:
> @jwei5 , I have checked the other states, would you check the latest code?

Hi @Jw00 , I have tested with the latest code, the multi scan API is working well.

hs36-kim (approved):
Thanks

Jw00:
> > @jwei5 , I have checked the other states, would you check the latest code?
> 
> Hi @Jw00 , I have tested with the latest code, the multi scan API is working well.

@jwei5 , thank you for your checking.

PR #6111, opened 2024-03-22
build/tools/amebasmart: Modify Flash Protection Status Register

ZhenBei-Sin:
Changes Note:
Modify the Flash Protection Status Register from 0x2C to 0x28 To protect 512KB of flash


sunghan-chang:
@ZhenBei-Sin Why is this happened? There was no change on this ever.

jeongarmy (approved):
LGTM

sunghan-chang:
@ZhenBei-Sin Please update the `secure storage` test example to verify this. I mean that the app should test range of write protection.

sunghan-chang:
> @ZhenBei-Sin Please update the secure storage test example to verify this. I mean that the app should test range of write protection.

Let's do this with new PR, not in this PR.

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -33,8 +33,13 @@ typedef struct {
...
 } nsfunc_ops_s;
 
-/* Flash Status Bit */
+/* Flash Status Bit Protect Lower 512KB */
+#ifdef CONFIG_AMEBAD_TRUSTZONE
 #define FLASH_STATUS_BITS 0x2c
+#endif
+#ifdef CONFIG_AMEBASMART_TRUSTZONE
+#define FLASH_STATUS_BITS 0x28
+#endif
```
sunghan-chang, line 42: I don't know this is proper. I don't have any requirement change on this. But anyway, when we want to change the region, we should modify several files. And it could be mismatched.
How about making this as a config?

And here are amebad and amebasmart only. How about amebalite? Because we support it, I ask you this.

ziyik: Different IC packaging uses different Flash, it has different Status bits values.
We will base it on previous AmebaLite packaging, thus it's the same as AmebaD.

We will create another PR to change this setting to a config, is that okay?

sunghan-chang: > Different IC packaging uses different Flash, it has different Status bits values.
> We will base it on previous AmebaLite packaging, thus it's the same as AmebaD.

Because of that, I suggest making a config to set for different boards configuration.

> We will create another PR to change this setting to a config, is that okay?

Sure

sunghan-chang (approved):
@ZhenBei-Sin @ziyik Can I merge this now?

ZhenBei-Sin:
> @ZhenBei-Sin @ziyik Can I merge this now?

Yes~ please proceed to merge. Thank you ~

PR #6115, opened 2024-03-25
os/board/rtl8730e: Fix SOC SVACE warning

ziyik:


sunghan-chang (commented):
@ziyik Could you write some explanation in the commit description?

PR #6116, opened 2024-03-25
Fix multiple svace issues 

abhishek-samsung:


abhishek-samsung:
@sunghan-chang , @prabu-r, @kishore-sn, @ewoodev   please review

os/fs/smartfs/smartfs_utils.c
```diff
@@ -1146,7 +1146,10 @@ int smartfs_finddirentry(struct smartfs_mountpt_s *fs, struct smartfs_entry_s *d
 
 	/* Test if this is a request for the root directory */
 
-	if (*relpath == '\0') {
+	if (relpath == NULL) {
+		ret = -EINVAL;
+		goto errout;
```
sunghan-chang, line 1151: Do we need to go to the errout? How about returning here directly?
@Taejun-Kwon Please review this.

sunghan-chang:
@jeongarmy Could you review this?

os/mm/mm_heap/mm_getheap.c
```diff
@@ -84,7 +84,9 @@ void mm_add_app_heap_list(struct mm_heap_s *heap, char *app_name)
 
 	node->is_active = true;
 	node->heap = heap;
-	strncpy(node->app_name, app_name, BIN_NAME_MAX);
+
+	memset(node->app_name, 0, BIN_NAME_MAX);
```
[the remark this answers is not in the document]
abhishek-samsung: Done. Added null character at the end.

os/fs/smartfs/smartfs_utils.c
Taejun-Kwon: prev code is proper. This means that user request something with directory '/mnt'
'mnt' is mount point, smartfs doesn't care path of mount point. hence '\0' is value of relpath of root directory.
Hence svace was wrong.

os/kernel/binary_manager/binary_manager_verify.c
```diff
@@ -180,7 +180,8 @@ int binary_manager_read_header(int type, char *devpath, void *header_data, bool
 		while (bin_size > 0) {
 			read_size = bin_size < crc_bufsize ? bin_size : crc_bufsize;
 			ret = read(fd, (void *)crc_buffer, read_size);
-			if (ret != read_size) {
+			/* prevent going into infinite loop if readsize is zero */
+			if (ret != read_size && read_size == 0) {
```
sunghan-chang, line 184: I don't know this is necessary.
1. by line 181, read_size could be the bin_size, but it's checked in line 180. It can't be zero
2. by line 181, read_size could be the crc_bufsize, but if crc_bufsize is zero, then by kmm_malloc, line 172, it will be returned at line 173. (malloc(0) returns null)

os/kernel/task/task_terminate.c
```diff
@@ -160,21 +160,19 @@ int task_terminate(pid_t pid, bool nonblocking)
...
 
 	dtcb = sched_gettcb(pid);
 	if (!dtcb) {
 		/* This PID does not correspond to any known task */
-		ret = -ESRCH;
-		goto errout_with_lock;
+		return -ESRCH;
 	}
 
+	trace_begin(TTRACE_TAG_TASK, "task_terminate");
```
sunghan-chang, line 174: Regardless of this PR, how about removing TTRACE? Because noone uses and maintains this.

os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -712,7 +712,9 @@ int binary_manager_execute_loader(int cmd, int bin_idx)
 	int ret;
 	uint8_t loader_priority;
 	main_t loader_func;
-	char data_str[1];
+	/* int limits -2,147,483,648 to 2,147,483,647, 11 characters + 1 null character.
+	 * We need to allocate 12bytes atleast to prevent buffer overflows during itoa */
+	char data_str[12];
```
jeongarmy, line 717: `data_str` variable is used only for `bin_idx` which is a index of binary table and I think it will be maximum under 5. Is it needed to expand?

os/kernel/task/task_terminate.c
```diff
@@ -171,8 +171,6 @@ int task_terminate(pid_t pid, bool nonblocking)
 		return -ESRCH;
 	}
 
-	trace_begin(TTRACE_TAG_TASK, "task_terminate");
```
sunghan-chang: Sorry but need to remove entire ttrace code. not only here.

sunghan-chang: Let's revert this pr and do with new pr for this.

abhishek-samsung: Oh okay. I understood it wrong. I dropped this commit.

sunghan-chang (approved):

ewoodev (approved):

PR #6119, opened 2024-03-26
os/pm: checks for wifi keep alive send

ritesh55555:
it adds a new member to global pm timer that is is_wifi_send_required. It checks if the wifi keep alive signal has been send and accordingly set PM transition.

### DIFF
--- a/os/pm/pm_checkstate.c
+++ b/os/pm/pm_checkstate.c
@@ -65,6 +65,12 @@
 #ifdef CONFIG_PM
 
 /****************************************************************************
+ * External Definitons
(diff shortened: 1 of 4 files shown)


os/pm/pm_checkstate.c
```diff
@@ -156,6 +162,13 @@ enum pm_state_e pm_checkstate(int domain)
 		}
 	}
 
+	/* Consider if Wifi keep alive sending is required or not */
```
kishore-sn, line 165: Consider -> Check ??

PR #6122, opened 2024-03-27
pm/pm_procfs: fix issue with lock timer not performing pm_unlock 

ewoodev:


os/pm/pm_changestate.c
```diff
@@ -110,6 +110,10 @@ static void pm_timer(int domain)
 
 	if (!pdom->wdog) {
 		pdom->wdog = wd_create();
+		if (!pdom->wdog) {
+			pmdbg("Fail to create wdog\n");
```
kishore-sn, line 114: Since this is error scenario, should we use error log instead of dbg log?

ewoodev: Thank you for review
It is calling by idle thread. so It is appred to change to the lldbg.

I have chagned for svace, but we are not using this function. so let me clean-up the this code.
thank you~

sunghan-chang (approved):

jeongarmy (approved):
LGTM

PR #6123, opened 2024-03-28
os/board/rtl8721csm: fix BLE get server connection status issue

yeetee179:
- added in check to make sure the input connection handle belongs to server before returning the connection status


sunghan-chang:
How about RTL8730E? No need to apply because ble sw is different?

yeetee179:
> How about RTL8730E? No need to apply because ble sw is different?

For 8730E it can return the correct server connection status. There is no need for this fix.

hs36-kim (approved):
LGTM
It had fixed at RS6.

PR #6127, opened 2024-04-01
os/driver/pm: Adds pm driver

ritesh55555:
- This commit adds pm driver that will be used to perform PM operations.
- Also , some of the ioctl commands are going to change when we add pm timer changes

### DIFF
--- a/os/drivers/pm/Make.defs
+++ b/os/drivers/pm/Make.defs
@@ -1,0 +1,63 @@
+############################################################################
+##
+## Copyright 2024 Samsung Electronics All Rights Reserved.
+##
+## Licensed under the Apache License, Version 2.0 (the "License");
+## you may not use this file except in compliance with the License.
+## You may obtain a copy of the License at
+##
+## http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing,
+## software distributed under the License is distributed on an
+## "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+## either express or implied. See the License for the specific
+## language governing permissions and limitations under the License.
+##
+############################################################################
+############################################################################
+# drivers/pm/Make.defs
+#
+#   Copyright (C) 2008, 2011-2012 Gregory Nutt. All rights reserved.
+#   Author: Gregory Nutt <gnutt@nuttx.org>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in
+#    the documentation and/or other materials provided with the
+#    distribution.
+# 3. Neither the name NuttX nor the names of its contributors may be
+#    used to endorse or promote products derived from this software
+#    without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+# OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+# AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+############################################################################
+
+ifeq ($(CONFIG_PM),y)
+
+CSRCS += pm.c
+
+# Include PM driver build support
+
+DEPPATH += --dep-path pm
+VPATH += :pm
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(TOPDIR)$(DELIM)drivers$(DELIM)pm}
+
+endif
--- a/os/drivers/pm/pm.c
+++ b/os/drivers/pm/pm.c
@@ -1,0 +1,206 @@
+/****************************************************************************
+ *
+ * Copyright 2024 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


os/drivers/pm/Make.defs
```diff
@@ -0,0 +1,63 @@
...
+##
+## Unless required by applicable law or agreed to in writing,
+## software distributed under the License is distributed on an
+## "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+## either express or implied. See the License for the specific
+## language governing permissions and limitations under the License.
+##
+############################################################################
+############################################################################
+# drivers/compression/Make.defs
```
sunghan-chang, line 19: Please check the path.

os/drivers/pm/Make.defs
```diff
@@ -0,0 +1,63 @@
...
+
+ifeq ($(CONFIG_PM),y)
+
+CSRCS += pm.c
+
+# Include Compression driver build support
+
+DEPPATH += --dep-path pm
+VPATH += :pm
+CFLAGS += ${shell $(INCDIR) $(INCDIROPT) "$(CC)" $(TOPDIR)$(DELIM)drivers$(DELIM)pm}
```
sunghan-chang, line 61: Could you let me know why this is necessary?

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+#include <tinyara/fs/ioctl.h>
+#include <tinyara/pm/pm.h>
+#include <tinyara/clock.h>
+
+#include <errno.h>
+#include <debug.h>
+#include <assert.h>
+
+/* Added to record timer countdown interrupt */
+struct pm_timer_s g_pm_timer = {0, 0, 0};
```
sunghan-chang, line 33: It looks that pm.c file uses only. Then, let's add `static`.

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+    	case PMIOC_LOCK:
+        	/* Always lock PM_NORMAL state, check again whether this implementation
+        	is enough to handle, if not, consider below methods...
+        	1a. PM_ACTIVITY is invoked in some of the driver, if application utlized it, it should draw state back to PM_NORMAL
+        	1b. If 1a cannot cover all condition, add one more syscall api, as a bridge for application to call pm_activity
+        	2a. Else, we can consider to lock/unlock for all state other than PM_SLEEP, ie. fetch state: pm_querystate(PM_IDLE_DOMAIN)
+        	2b. Need to verify if the transition is going to happen, but lock state appears suddenly, what will happen?
+        	*/
+        	pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
+        	fvdbg("State locked!\n");
```
sunghan-chang, line 112: fvdbg is the filesystem debug message. You should use `pmvdbg`.

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+        /****************************************************************************
+         * Description:
+         *   apps can use this method to get the current time tick.
+         *
+         * Returned Value:
+         *  Time tick value of current time
+         ****************************************************************************/
+        case PMIOC_TIMETICK:
+        	*((size_t *)arg) = clock_systimer();
+			ret = OK;
```
sunghan-chang, line 169: Could you align indentation?

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+         *  in g_pm_timer.last_wifi_alive_send_time that will be used to calculate the 
+         *  duration of required sleep later just before sleep. If another
+         *  interrupt occurs during this interval, the system will still
+         *  wake up.
+         *  Also we dont want to do pm_set_timer to PM_WAKEUP_TIMER if the
+         *  system is already in PM_LOCK_TIMER state. If system is in 
+         *  PM_LOCK_TIMER state, then power_unlock_write can only change that state.
+         *
+         ****************************************************************************/
+        case PMIOC_SET_WIFI_WAKEUP:
```
sunghan-chang, line 153: I don't agree the name. PM should not have some module specific name. If you add other periodic timer, will you add that with different ioctl command?

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+        case PMIOC_WAKEUP_TIMER:
+        	if (g_pm_timer.timer_type == PM_NO_TIMER && arg > 0) {
+        		pm_set_timer(PM_WAKEUP_TIMER, arg);
+        	}	 
+        	ret = OK;
+        	break;
+        /****************************************************************************
+         * Description:
+         *  Sets timer interval for next sleep cycle after sending of wifi keep 
+         *  alive signal from UART. It also stores the current time tick
```
sunghan-chang, line 143: "after sending of wifi keep alive signal from UART" can be an example of this command.
But it should not be a main content.

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+        	ret = OK;
+        	break;
+        /****************************************************************************
+         * Description:
+         *   apps can use this method to get the current time tick.
+         *
+         * Returned Value:
+         *  Time tick value of current time
+         ****************************************************************************/
+        case PMIOC_TIMETICK:
```
sunghan-chang, line 167: Why do we support this in PM? The clock module already supports, if not, we should add it in the clock, not here.

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+        *  Change the operating frequency of the core to save power
+        *  consumption during active mode.
+        ****************************************************************************/
+        case PMIOC_TUNEFREQ:
+        	/* Buffer string need not to be considered, just div level is enough */
+        	/* Available frequencies for changing mode of power consumption
+        	0 -> 1.2GHz
+        	1 -> 600MHz
+        	2 -> 400Mhz
+        	3 -> 300Mhz
```
sunghan-chang, line 184: This is so rtl8730e specific. PM should not be a chipset or a module specific.

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,204 @@
...
+        	*((size_t *)arg) = clock_systimer();
+			ret = OK;
+        	break;
+
+#ifdef CONFIG_PM_DVFS
+        /****************************************************************************
+        * Description:
+        *  Change the operating frequency of the core to save power
+        *  consumption during active mode.
+        ****************************************************************************/
```
sunghan-chang, line 177: Good for explanation of each command. But it should be in the header, not the source code. Because user will check the header, not the source file.

os/include/tinyara/fs/ioctl.h
```diff
@@ -103,6 +103,7 @@
 #define _CPULOADBASE    (0x2800)	/* cpuload ioctl commands */
 #define _THERMALBASE	(0x2900)	/* thermal camera control ioctl commands */
 #define _COMPBASE       (0x2a00)	/* compress ioctl commands */
+#define _PMBASE         (0x2b00)    /* pm ioctl commands */
```
sunghan-chang, line 106: could you align the comment?

os/include/tinyara/pm/pm.h
```diff
@@ -247,11 +247,7 @@
 /* Defines default sleep duration */
 #define DEFAULT_PM_SLEEP_DURATION    10000000 /* 10 sec in microsecond*/
 
-#define PM_LOCK_PATH					"/proc/power/domains/0/pm_lock"
-#define PM_UNLOCK_PATH					"/proc/power/domains/0/pm_unlock"
-#ifdef CONFIG_PM_DVFS
-#define PM_TUNEFREQ_PATH				"/proc/power/domains/0/pm_tunefreq"
-#endif
+#define PM_DRVPATH     "/dev/pm"
```
sunghan-chang, line 250: Except the pm.c, does some other file use this? If not, let's move this into the pm.c.

os/drivers/pm/pm.c
edwakuwaku: Hi @sunghan-chang , I think os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c and os/pm/pm_activity.c is still using this.

ritesh55555: yes , it is a global variable

PR #6129, opened 2024-04-03
os/pm, os/driver/pm: Add pm_timer module 

ritesh55555:
- This commit adds pm_timer module in os/pm to use multiple timers for PM operation and PM driver also.
- It defines a structure for multiple wakeup timers. All the timers are stored in a list and just before system sleeps, the timer with nearest future expiration time is started.


os/pm/pm_timer/pm_timer.h
```diff
@@ -0,0 +1,117 @@
...
+ * g_pmTimer_list. 
+ */
+
+extern uint16_t g_pmTimer_nfree;
+
+/* g_pmTimer_list is a list of pre-allocated pm timers. The number of pm timers
+ * in the list is a configuration item.
+ */
+
+extern pm_wakeup_timer_t g_pmTimer_list[CONFIG_PM_MAX_WAKEUP_TIMER];
```
sunghan-chang, line 106: Could you let me know why you want to use pre-allocated and fixed array?

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,155 @@
...
+/************************************************************************************
+ * Name: pm_ioctl
+ *
+ * Command description:
+ *   PMIOC_LOCK - for locking PM transition (both permanent and timer)
+ *   PMIOC_UNLOCK - for unlocking PM transiton
+ *   PMIOC_TIMER_CREATE - to create a wakeup timer
+ *   PMIOC_TIMER_SET - to set a wakeup timer
+ *   PMIOC_TIMER_STOP - to stop a specific wakeup timer
+ *   PMIOC_TIMER_DELETE - to delete a specifi wakeup timer and make it available for other apps
```
edwakuwaku, line 82: typo: specific

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,155 @@
...
+        	*((size_t *)arg) = clock_systimer();
+        	ret = OK;
+        	break;
+#ifdef CONFIG_PM_DVFS
+        case PMIOC_TUNEFREQ:
+        	pm_dvfs(arg);
+        	ret = OK;
+        	break;
+#endif
+        }
```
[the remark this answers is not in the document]
edwakuwaku: you may want to add some logs here if input type is out of bounds

edwakuwaku:
Hi @ritesh55555 , 
Initially, I thought the timer resource manager should be held under /dev/timer, as timer is not only specific to PM module.
In os/board/rtl8730e/src/rtl8730e_boot.c, we initialize the timers w.r.t the timer driver (ie. /dev/timer), . Could you please let me know how are we going to link this pm timer with the lower level timer driver control?
If you check this PR https://github.com/Samsung/TizenRT/pull/6064, the final setup for timer is during up_set_pm_timer in os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c. In the ioctl for pm driver, other than state control and dxfs, others are all related to timer.
My question is, how do we realise the relation between /dev/timer and /dev/pm?

ritesh55555:
> Hi @ritesh55555 , Initially, I thought the timer resource manager should be held under /dev/timer, as timer is not only specific to PM module. In os/board/rtl8730e/src/rtl8730e_boot.c, we initialize the timers w.r.t the timer driver (ie. /dev/timer), . Could you please let me know how are we going to link this pm timer with the lower level timer driver control? If you check this PR #6064, the final setup for timer is during up_set_pm_timer in os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c. In the ioctl for pm driver, other than state control and dvfs, others are all related to timer operation. My question is, how do we realise the relation between /dev/timer and /dev/pm?

the new modification still uses g_pm_timer, pm_set_timer() and pm_init_handler() as before . The PM LOCK TIMER is same as before. the moment app will request pm_lock with timer, it will set pm_set_timer(PM_LOCK_TIMER, timer_interval) and do the up_set_pm_timer() and start the timer as before. 
But we wont be using pm_unlock with timer , that has been removed becuase that didnot make sense.
Here we can add multiple wake up timers (struct pm_wakeup_timer_s) in a list , so whenever board is going to sleep , just before that , we can get the right time , find its delay and then do pm_set_timer(PM_WAKEUP_TIMER, delay).

i hope this explains!

os/include/tinyara/fs/ioctl.h
```diff
@@ -451,6 +452,19 @@
 #define COMPIOC_GET_COMP_NAME      _COMPIOC(0x0003)
 #define COMPIOC_DECOMPRESS	   _COMPIOC(0x0004)
 
+/* Pm driver ioctl definitions *****************************/
+#define _PMIOCVALID(c)      (_IOC_TYPE(c) == _PMBASE)
+#define _PMIOC(nr)          _IOC(_PMBASE, nr)
+
+#define PMIOC_LOCK               _PMIOC(0x0001)
+#define PMIOC_UNLOCK             _PMIOC(0x0002)
+#define PMIOC_TIMER_CREATE       _PMIOC(0x0003)
```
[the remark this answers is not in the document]
ritesh55555: i have assumed whenever an app wants to use a timer, it will have to first use TIMER_CREATE to create a timer. 
Then to set the timer, app need to use TIMER_SET just after executing its required function. For the case of wifi sending, after sending wifi , it should always do TIMER_SET. 
TIMER_STOP will be used to stop the timer temporarily.
TIMER_DELETE will be used to completely remove the timer from app , so that other app can use it.

os/pm/pm_timer/pm_timer.h
ritesh55555: Here are the following reasons:
1. code complexity is very less. It is easy to find the timer node using index. High efficiency(faster) and less cases to handle.
2. If we use dynamically allocated timer, we need a way to find the different timer whenever an apps call a specific timer, we will need an indexing method because we should not provide the address to the app. Apart from that decrease in efficiency but good memory usage.
3. If we dont need a lot of timers , then i think it is more efficient to just allocate around 10 timer statically ?

os/include/tinyara/pm/pm.h
```diff
@@ -314,11 +314,28 @@ enum pm_timer_type_e {
...
-	clock_t last_wifi_alive_send_time;       /* Time tick of the last wifi keep alive signal sent time */
+	uint8_t timer_type;		                 /* Bits here are set according to the timer that is to be used */
+	uint32_t timer_interval;	             /* The interval for whichever timer is to be used */
 };
 
+enum pm_timer_status_e {
+	FREE = 0,
+	INACTIVE,
+	ACTIVE,
+	RUNNING,
```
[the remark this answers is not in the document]
ritesh55555: a timer is said to be ACTIVE when it is in the list but not yet started. 
a timer is said to be RUNNING when the timer is started and system should be in sleep.
if some uart interrupt occurs before a timer has expired, then RUNNING status will be helpful to handle that case.

edwakuwaku:
> @edwakuwaku Hi currently for pm_lock with timer , we dont have a way to stop the timer . Suppose we want to stop it inbetween by calling pm_unlock. It will only pm_relax() but it will not stop the timer that was started by pm_lock timer. Then during expiration, it will call pm_relax() again and it will crash. Can we have an api in amebasmart_pmhelpers to stop the timers. currently there is only up_set_pm_timer() which only starts. I think we need another api to stop the timers also.

Can you try
```
void up_stop_pm_timer(void) {
    gtimer_stop(&g_timer1);
    RTIM_INTClear(TIMx[TIMER1]);
}
```

edwakuwaku:
> @edwakuwaku Hi currently for pm_lock with timer , we dont have a way to stop the timer . Suppose we want to stop it inbetween by calling pm_unlock. It will only pm_relax() but it will not stop the timer that was started by pm_lock timer. Then during expiration, it will call pm_relax() again and it will crash. Can we have an api in amebasmart_pmhelpers to stop the timers. currently there is only up_set_pm_timer() which only starts. I think we need another api to stop the timers also.

Let me confirm with my understanding, for the stop it in between before pm_unlock, I assume that you do not need this timer to trigger wakeup or handle some other operation. Thus you would like to stop it, so that it will not trigger callback to do pm_relax twice, is that correct?

ritesh55555:
> > @edwakuwaku Hi currently for pm_lock with timer , we dont have a way to stop the timer . Suppose we want to stop it inbetween by calling pm_unlock. It will only pm_relax() but it will not stop the timer that was started by pm_lock timer. Then during expiration, it will call pm_relax() again and it will crash. Can we have an api in amebasmart_pmhelpers to stop the timers. currently there is only up_set_pm_timer() which only starts. I think we need another api to stop the timers also.
> 
> Let me confirm with my understanding, for the stop it in between before pm_unlock, I assume that you do not need this timer to trigger wakeup or handle some other operation. Thus you would like to stop it, so that it will not trigger callback to do pm_relax at the same state more than once, is that correct?

yes , i just want to stop it . No callback needed.

os/include/tinyara/pm/pm.h
```diff
@@ -314,9 +308,35 @@ enum pm_timer_type_e {
 };
 
 struct pm_timer_s {
```
ewoodev, line 310: It appears to be unused. Please remove any unnecessary definitions.

os/pm/pm_changestate.c
```diff
@@ -300,11 +300,9 @@ int pm_changestate(int domain_indx, enum pm_state_e newstate)
 		pm_changeall(domain_indx, newstate);
 		pdom->state = newstate;
 		
-		/* Adjusting the sleep duration, so that system can revert to NORMAL
-		*  state when next Wifi keep alive signal is required to be sent. 
-		*/
 		if (newstate == PM_SLEEP) {
-			pm_adjust_sleep_duration();
+			/* Set the required wakeup timer from the g_pmTimer_activeList */
+			ret = pm_set_wakeup_timer();
```
ewoodev, line 305: The threshold value for the timer that will put the main core to sleep should be checked here.

If the set wake-up timer interval value is less than the time during the main core actually sleeps, it is an unnecessary sleep.

os/pm/pm_timer/pm_timer_create.c
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * Parameters:
+ *   is_periodic - if the timer should be periodic
+ *
+ * Return Value:
+ *   id of the timer
+ *
+ ************************************************************************/
+
+int pm_timer_create(bool is_periodic)
```
ewoodev, line 94: We do not need to create a PM timer for the user area.

Please change it so that memory is allocated and freed inside the PM.

We can only provide the set of pm_timer.
If necessary, timer cancel can be provided, and it is possible to control the PID with this.

This is because a single PID cannot set both pm_lock_timer and pm_wakeup_timer.

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,293 @@
...
+        (void)sq_rem((FAR sq_entry_t *)timer, &g_pmTimer_activeList);
+        timer->status = INACTIVE;
+        pmdbg("pm timer with id %d is removed from list\n", timer->id);
+        /* check for extra operation or change any fields of the timer ??? */
+}
+
+void add_pm_timer(pm_wakeup_timer_t *timer, unsigned int timer_interval) 
+{
+        /* updating the expire time of the wakeup timer */
+        timer->expire_timetick = clock_systimer() + timer_interval;
```
ewoodev, line 106: We are able to store remain valuse (header interval - setting interval) like to kernel/wdog

Please change the pm timer so that it does not have a dependency on the system time.
The system time can be changed by the user at any time.

We need to discuss how to handle the expiration of the header's time. We have two options:
1. Use a hardware timer to expire the header.(However, interrupts caused by hardware timer expiration during waking hours are unnecessary.)
2. Decrease the watchdog tick when it is decreased, but this option raises concerns about potential delays with the tick.

ewoodev (changes requested):

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,154 @@
...
+{
+	int ret = -ENOSYS;
+    
+	/* Handle built-in ioctl commands */
+	switch (cmd) {
+    	case PMIOC_LOCK:
+        	pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
+        	pmvdbg("State locked!\n");
+        	if (arg > 0) {
+            		pm_set_timer(PM_LOCK_TIMER, arg);
```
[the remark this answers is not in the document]
ewoodev: Because I think it is better separate the pm lock and pm timer within the code entirely. 
These two operations are completely distinct from one another. 
the pm timer use just pm_lock.

os/include/tinyara/pm/pm.h
ritesh55555: it is used to set lock or wakeup timer

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,293 @@
...
+
+pm_wakeup_timer_t *get_pm_timer(int id)
+{
+        /* If we also use malloc to allocate memory for timers,
+         * We need a good way to return the required timer ??? 
+         * For now, we just return timer from the g_pmTimer_list. */
+
+        /* handle the case if it is a valid id */
+        if (id < 0 || id >= CONFIG_PM_MAX_WAKEUP_TIMER) {
+                return NULL;
```
Taejun-Kwon, line 84: 1. It seems this error case can be happened very easily by user's mistake, because it can be refered by ioctl.
Hence you should put some debug message include value of id here.
2. 2nd state (id > CONFIG_PM_MAX_WAKEUP_TIMER) will be happened more than first state. when you use if state to check error case, you should consider this..

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        return &g_pmTimer_list[id];
+}
+
+void remove_pm_timer(pm_wakeup_timer_t *timer)
+{
+        (void)sq_rem((FAR sq_entry_t *)timer, &g_pmTimer_activeList);
+        timer->status = INACTIVE;
```
Taejun-Kwon, line 98: remove from timer queue, then I think FREE is proper state. INACTIVE usually use for  something like a'stopped but not removed' state(c.f paused)

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,293 @@
...
+ * Return Value:
+ *   0 - success
+ *   -1 - error
+ *
+ ************************************************************************/
+
+int pm_timer_set(int id, unsigned int timer_interval)
+{
+        pm_wakeup_timer_t *timer = get_pm_timer(id);
+        DEBUGASSERT(timer);
```
Taejun-Kwon, line 178: I think DEBUGASSERT is too high option here. Return error & debug message is enough....please discuss with your team about this.

os/pm/pm_timer/pm_timer_set.c
ritesh55555: yeah, the function is suppose to stop the timer. Not free. 
the name was kept remove pm timer becasue it removes from the list and will be used again.

os/drivers/pm/pm.c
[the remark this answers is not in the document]
ritesh55555: done

os/pm/pm_timer/pm_timer_create.c
ritesh55555: Now it supports only timer set and timer cancel to user. Updated.

os/pm/pm_timer/pm_timer_set.c
ritesh55555: I have updated the code , please review again.

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,198 @@
...
+        is_pm_lock[timer->pid] = PM_PID_UNLOCK;
+        leave_critical_section(state);
+        
+        return PM_TIMER_SUCCESS;
+}
+
+int pm_timer_cancel(void) 
+{
+        int pid = getpid();
+
```
ewoodev, line 186: free() is not nessarary in pm_timer_cancel?

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * Returned Value:
+ *   None.
+ * 
+ * Assumption: This should be also implemented for CONFIG_SCHED_TICKLESS and
+ * CONFIG_SCHED_TICKSUPRESS.
+ *
+ ****************************************************************************/
+
+void pm_timer_update(void)
```
ewoodev, line 192: we need this function for only -- tick?

How about use below fucntion pm_wakeup_handler use only with argument tick=1?

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * Parameters:
+ *   missing_tick : missing ticks after sleep
+ *
+ * Return Value:
+ *   None
+ *
+ ************************************************************************/
+
+void pm_wakeup_handler(clock_t missing_tick)
```
ewoodev, line 222: This operation appear to only for update pm timer tick.

How about change function name to pm_timer_update? and merge above fucntion?

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,244 @@
...
+		 * for the specific pid, so that the specific thread can complete 
+		 * its function */
+		sq_remfirst(&g_pmTimer_activeList);
+		if (is_pm_lock[head->pid] == PM_PID_UNLOCK) {
+			is_pm_lock[head->pid] = PM_PID_LOCK;
+			pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
+			pmdbg("pm is locked for process with id = %d\n", head->pid);
+		}
+
+		if (PM_ISSTATIC(head)) {
```
ewoodev, line 99: If we have create fucntion, we need to remove or deleate fucntion also. It is pair.

Please make a pm_timer_delete function and call here

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,244 @@
...
+ *   None
+ *
+ ************************************************************************/
+
+void pm_wakeup_handler(clock_t missing_tick)
+{
+	pm_wakeup_timer_t *timer;
+	int decr;
+	/* Check the boundary case, if missing tick is more than INTMAX*/
+	int ticks = missing_tick;
```
ewoodev, line 227: It it un-nessarary code.
please remove missing_tick.

os/pm/pm_timer/pm_timer_create.c
```diff
@@ -0,0 +1,132 @@
...
+                 * count of free pm timers */
+                timer = (pm_wakeup_timer_t *)sq_remfirst(&g_pmTimer_freeList);
+
+                if (timer) {
+                        DEBUGASSERT(g_pmTimer_nfree > 0);
+                        g_pmTimer_nfree--;
+                        timer->next = NULL;
+                        timer->pid = getpid();
+                        timer->flags = PM_STATIC;
+                        pmdbg("pid of the timer is %d\n", timer->pid);
```
ewoodev, line 113: It just infomation log, Please use pmvdbg

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,244 @@
...
+		/* Check if this needs to be atomic */
+		state = enter_critical_section();
+
+		/* Remove the pm timer from active list and lock the system 
+		 * for the specific pid, so that the specific thread can complete 
+		 * its function */
+		sq_remfirst(&g_pmTimer_activeList);
+		if (is_pm_lock[head->pid] == PM_PID_UNLOCK) {
+			is_pm_lock[head->pid] = PM_PID_LOCK;
+			pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
```
ewoodev, line 95: why here set lock?

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,151 @@
...
+        case PMIOC_UNLOCK:
+            	pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
+            	pmvdbg("State unlocked!\n");
+            	ret = OK;
+            	break;
+	case PMIOC_TIMER_LOCK:
+		if (arg > 0) {
+			pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
+			pmvdbg("State locked!\n");
+			pm_set_timer(PM_LOCK_TIMER, arg);
```
ewoodev, line 116: pm_set_timer is h/w timer.

we don't need h/w timer in pm timer of locking.
Please use sw timer(wdog)

os/drivers/pm/pm.c
```diff
@@ -0,0 +1,151 @@
...
+#include <tinyara/clock.h>
+
+#include <errno.h>
+#include <debug.h>
+#include <assert.h>
+
+#define PM_DRVPATH           "/dev/pm"
+
+/* Added to record timer countdown interrupt */
+struct pm_timer_s g_pm_timer = {0, 0};
```
edwakuwaku, line 35: I think we don't need this now~

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/************************************************************************
+ * pm/pm_timer/pm_timer_create.c
```
edwakuwaku, line 19: file name seems incorrect~

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,198 @@
...
+ * Parameters:
+ *   timer_interval - expected board sleep duration
+ *
+ * Return Value:
+ *   0 - success
+ *   -1 - error
+ *
+ ************************************************************************/
+
+int pm_timer_set(unsigned int timer_interval)
```
edwakuwaku, line 155: There are 2 API, pm_timer_set and pm_set_timer. Although both of them serve different purpose, but the naming is too hard to differentiate. Would you to change one of these?
For pm_set_timer, it is used to set the HW timer to wakeup, so I think it is feasible to named as such.
For pm_timer_set here, actually it is adding a virtual timer to the pm_timer pool, also the core logic seems to be the function above (ie. add_pm_timer, no point create a helper function to do that since there is only one caller, create another function incur additional overhead for branching, so you can just inline it). Thus, I suggest to merge the api above to here. And perhaps just rename to pm_timer_add (ie. both API and filename).

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,198 @@
...
+                pm_relax(PM_IDLE_DOMAIN, PM_NORMAL);
+                pmdbg("pm is unlocked for process with id = %d\n", timer->pid);
+        }
+        is_pm_lock[timer->pid] = PM_PID_UNLOCK;
+        leave_critical_section(state);
+        
+        return PM_TIMER_SUCCESS;
+}
+
+int pm_timer_cancel(void)
```
edwakuwaku, line 183: How about create another file for pm_timer_cancel?

os/pm/pm_timer/pm_timer_set.c
```diff
@@ -0,0 +1,198 @@
...
+
+/************************************************************************
+ * Private Functions
+ ************************************************************************/
+
+void add_pm_timer(pm_wakeup_timer_t *timer, unsigned int timer_interval) 
+{
+        /* updating the delay of the wakeup timer */
+        timer->delay = timer_interval;
+        pmdbg("timer->delay is %d\n", timer->delay);
```
edwakuwaku, line 80: There seems many places using debug log (ie. pmdbg) instead of information log (ie. pmvdbg). Please check and revise~

os/pm/pm_timer/pm_timer_add.c
```diff
@@ -0,0 +1,146 @@
...
+ * 
+ * Parameters:
+ *   pm_wakeup_timer_s pointer
+ *
+ * Return Value:
+ *   None
+ * 
+ ************************************************************************/
+
+void pm_timer_add(pm_wakeup_timer_t *timer)
```
ewoodev, line 92: How about func name pm_timer_start?

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,180 @@
...
+
+	while (g_pmTimer_activeList.head && ((pm_wakeup_timer_t *)g_pmTimer_activeList.head)->delay <= 0) {
+		
+    		head = (pm_wakeup_timer_t *)g_pmTimer_activeList.head;
+
+		/* Remove the pm timer from active list */
+        	sq_remfirst(&g_pmTimer_activeList);
+
+		/* As the timer is expired, give back the semaphore to unlock the thread */
+		sem_post(&head->pm_sem);
```
ewoodev, line 87: We can change this code to execute the registered timer callback in order to implement pm_timer in a general-purpose and scalable manner.

sem_post is only for pm_sleep() and it can be handled in the registered callback.

os/pm/pm_timedStay.c
```diff
@@ -0,0 +1,171 @@
...
+	if (pid_timerMap[pid] != NULL) {
+		pmdbg("There is already a lock timer running for this process\n");
+		return PM_TIMER_FAIL;
+	}
+	
+	WDOG_ID wdog = wd_create();
+	pid_timerMap[pid] = wdog;
+
+	/* Lock the pm transition and Start the wdog timer */
+	pm_stay(PM_IDLE_DOMAIN, PM_NORMAL);
```
ewoodev, line 123: wa can call pm_timedStay for other pm state as pm_idle, pm_standby.

Please add argument state and domain.

os/pm/pm_timedStay.c
```diff
@@ -0,0 +1,171 @@
...
+ ************************************************************************/
+
+/************************************************************************
+ * Private Variables
+ ************************************************************************/
+
+/* This array maps the pid to their respective wdog timer. 
+ * Assumption: when a thread does a timed lock , only the same thread 
+ * can unlock the "timed lock" before it expire */
+static WDOG_ID pid_timerMap[CONFIG_MAX_TASKS];
```
ewoodev, line 74: This may cause potential operational errors in the unloading of the binary or in the terminating situation of the task.

Isn't it necessary to handle this part using APIs such as atexit() and on_exit()?

ewoodev: The pid can be over than CONFIG_MAX_TASKS.

#define MAX_PID_MASK	(CONFIG_MAX_TASKS - 1)
#define PIDHASH(pid)	((pid) & MAX_PID_MASK)
task_assignpid()

Please ref this logic

os/pm/pm_changestate.c
ewoodev: Please check this comment.

we need threshold

ewoodev:
How about seperate pm driver commit?
Could you let me know you opinion?

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -116,29 +116,15 @@ int SOCPS_AONWakeReason(void)
...
-	}
-	// Reset the global struct
-	g_pm_timer.timer_type = PM_NO_TIMER;
-	g_pm_timer.timer_interval = 0;
+	// Switch status back to normal mode after wake up from interrupt
+	pm_activity(PM_IDLE_DOMAIN, 9);		
 }
 
-void up_set_pm_timer(void) {
+void up_set_pm_timer(unsigned int timer_interval) {
```
ewoodev, line 123: [up_set_pm_timer](https://github.com/Samsung/TizenRT/blob/46fb6e5bd9fc0f2138be0982a8100ae82b9ab061/os/arch/arm/src/amebasmart/amebasmart_idle.c#L95)

we don't need to change this api??

os/pm/pm_timedStay.c
ewoodev: let's seperate pm_timedStay_cancel

we can apply about pm_timedStay_cancel().
Please remove in this PR and raise new PR

os/include/tinyara/pm/pm.h
```diff
@@ -315,12 +318,16 @@ struct pm_arg_s {
...
-struct pm_timer_s {
-	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
-	uint32_t timer_interval;	/* The interval for whichever timer is to be used */
-	clock_t last_wifi_alive_send_time;       /* Time tick of the last wifi keep alive signal sent time */
+struct pm_wakeup_timer_s {	
+	struct pm_wakeup_timer_s *next;   /* pointer to next timer in the linked list */
+	int pid;                          /* id of process that created pm timer */
+	uint8_t flags;                    /* See pm timer definations above*/
+	int delay;		          /* refers to time of sleep expected */
+	sem_t pm_sem;			  /* semaphore to lock the thread that started the pm timer */
```
ewoodev, line 326: you are providing this as a public API so that pm_wakeup_timer can be used elsewhere besides pm_sleep.

Then pm_sem should not be here.

Add a callback to be called when the timer expires and change it to receive pm_sem as an argument.

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,180 @@
...
+
+		/* Make the pm timer free */
+		pm_timer_delete(head);
```
ewoodev, line 90: This seems strange.

pm_timer_delete() should be called at the same location as pm_timer_create().

Move this to pm_sleep().

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,180 @@
...
+	/* Remove the wakeup timers that are already expired. 
+	 * Although this must have already handled by pm_timer_update() */
+	while (curr->delay <= 0 && curr->next) {
+		prev = curr;
+		curr = curr->next;
+	}
+
+	if (curr->delay > 0) {
+		pmdbg("Setting timer and Board will wake up after %d millisecond\n", curr->delay);
+		up_set_pm_timer(curr->delay * 1000);
```
ewoodev, line 136: This is 1 tick of TizenRT.

We can change the standard of 1 tick.

Thus, simply multiplying this by 1000 is incorrect operation.

Also, up_set_pm_timer is an API only for pm wakeup timer. I do not know why they are setting time based on hw timer in the api for pm.

Please reconsider the interval unit of up_set_pm_timer.

os/pm/pm_timedStay.c
ewoodev: Let's seperate PR about this and pm_timedStay_cancel.

we can apply pm_timedStay_cancel later.
This PR being delayed.

For fast merge, Please remove it in this PR and please raise new PR.

ritesh55555: done

os/pm/pm_timer/pm_timer_add.c
ritesh55555: we are actually adding the timer in the list , pm_timer_start doenot match the functionality of the function !

os/pm/pm_timer/pm_timer_activity.c
ritesh55555: multiplying by 1000 refers to changing the value from millisecond to nanosecond.

os/pm/pm_timer/pm_timer_create.c
```diff
@@ -0,0 +1,132 @@
...
+ *
+ ************************************************************************/
+
+pm_wakeup_timer_t *pm_timer_create()
+{
+        pm_wakeup_timer_t *timer;
+        irqstate_t state;
+
+        /* Creating a timer should be atomic */
+        state = enter_critical_section();
```
ewoodev, line 99: It is disabling irq
so this locking must be minium

but it has kmm_malloc
Please apply very briefly only to the minimum code you need to protect

os/pm/pm_sleep.c
```diff
@@ -0,0 +1,123 @@
...
+        timer->delay = timer_interval;
+        pm_timer_add(timer);
+
+        leave_critical_section(state);
+
+        /* sem_wait untill the timer expires */      
+        sem_wait(&timer->pm_sem);
+
+        return PM_TIMER_SUCCESS;
+}
```
ewoodev, line 123: Please add endline

os/drivers/pm/pm.c
```diff
@@ -72,13 +72,15 @@ static ssize_t pm_write(FAR struct file *filep, FAR const char *buffer, size_t l
...
  *   PMIOC_TUNEFREQ - for changing the operating frequency of the core to save power
  * 
  * Arguments:
  *   filep is ioctl fd, cmd is required command, arg is required argument for
  *   the command. 
- *   for PMIOC_SUSPEND, arg is unsigned long representing PM STATE
- *   for PMIOC_RESUME, arg is unsigned long representing PM STATE
+ *   for PMIOC_SUSPEND, arg is an enum representing PM STATE
+ *   for PMIOC_RESUME, arg is an enum representing PM STATE
+ *   for SLEEP, arg should be an int.
```
edwakuwaku, line 83: Is this PMIOC_SLEEP?

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,180 @@
...
+
+/************************************************************************
+ * Public Variables
+ ************************************************************************/
+
+/************************************************************************
+ * Private Functions
+ ************************************************************************/
+
+void pm_timer_expire(void)
```
edwakuwaku, line 75: It seems only be used in current file, maybe we can mark a static?

os/include/tinyara/pm/pm.h
```diff
@@ -241,6 +242,15 @@
 											 */
 #endif
 
+/* PM timer Definitions *************************************************/
+/* Flag bits for the flags field of struct pm_wakeup_timer_s */
+
+#define PM_STATIC      (1 << 0)	        /* Bit 0: pm timer is statically allocated */
```
ewoodev, line 248: Please move internal header

os/pm/pm_sleep.c
```diff
@@ -0,0 +1,134 @@
...
+ * Public Functions
+ ************************************************************************/
+
+static void pm_timer_callback(pm_wakeup_timer_t* timer)
+{
+        /* As the timer is expired, give back the semaphore to unlock the thread */
+	sem_post(&timer->pm_sem);
+
+	/* Make the pm timer free */
+	pm_timer_delete(timer);
```
ewoodev, line 80: Please move this to last line in pm_sleep() 
delete is big operation. Please remove in ISR

os/include/tinyara/pm/pm.h
```diff
@@ -307,20 +317,30 @@ enum pm_timer_type_e {
...
+	uint8_t timer_type;		                 /* Bits here are set according to the timer that is to be used */
+	uint32_t timer_interval;	                 /* The interval for whichever timer is to be used */
+};
+
+struct pm_wakeup_timer_s {	
+	struct pm_wakeup_timer_s *next;   		      /* pointer to next timer in the linked list */
+	int pid;                          		      /* id of process that created pm timer */
+	uint8_t flags;                    		      /* See pm timer definations above*/
+	int delay;		          		      /* refers to time of sleep expected */
+	void (*callback)(struct pm_wakeup_timer_s *timer);    /* function to be executed when timer expires */
```
ewoodev, line 338: it is some strange..
please check this

os/pm/Kconfig
```diff
@@ -45,6 +45,15 @@ config PM_METRICS_DURATION
 
 endif
 
+config PM_MAX_WAKEUP_TIMER
```
ewoodev, line 48: how about PM_MAX_STATIC_TIMER ?

And this pr is having both pm_wakeup_timer and pm_timer.
How about unifying the terms into one?

os/pm/pm_timer/pm_timer_delete.c
```diff
@@ -0,0 +1,126 @@
...
+
+        } else if (PM_ISALLOCED(timer)) {
+                /* Free the dynamically allocated timer 
+                * If the timer was released from an interrupt handler, 
+                * sched_kfree() will defer the actual deallocation of the
+                * memory until a more appropriate time. 
+                * 
+                * We don't need interrupts disabled to do this */
+                leave_critical_section(state);
+                sched_kfree(timer);
```
ewoodev, line 118: This is to request lpwork to do kmm_free().

pm_timer does not need this function because there is no load as much as wdog.

os/pm/pm_timer/pm_timer_create.c
```diff
@@ -0,0 +1,134 @@
...
+
+        /* If we have pre-allocated timer available then use that */
+        if (g_pmTimer_nfree > 0) {
+                /* Remove the pm timer from the free list and decrement the 
+                 * count of free pm timers */
+                timer = (pm_wakeup_timer_t *)sq_remfirst(&g_pmTimer_freeList);
```
ewoodev, line 105: Can we apply enter_critical_section only to this code?

os/pm/pm_timer/pm_timer_delete.c
```diff
@@ -0,0 +1,126 @@
...
+                * We don't need interrupts disabled to do this */
+                leave_critical_section(state);
+                sched_kfree(timer);	
+
+        } else {
+                /* This should not be the case because a pm timer 
+                * is either statically allocated or dynamically */
+                pmdbg("This should not be any case\n");
+                leave_critical_section(state);
+        }
```
edwakuwaku, line 125: All condition here does leave_critical_section, how about pull it out and do it once at the end of pm_timer_delete?

apps/shell/tash_command.c
```diff
@@ -561,12 +561,11 @@ int tash_execute_cmd(char **args, int argc)
 			if (tash_cmds_info.cmd[cmd_idx].exec_type == TASH_EXECMD_SYNC) {
 				/* function call to execute SYNC command */
 #ifdef CONFIG_PM
-				pm_arg_t pm_arg = {PM_IDLE_DOMAIN, PM_NORMAL};
```
ewoodev, line 564: Please seperate commit

os/drivers/pm/pm.c
```diff
@@ -92,15 +97,31 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
         case PMIOC_RESUME:
-		pm_relax(((pm_arg_t *)arg)->domain, ((pm_arg_t *)arg)->state);
+		pm_relax(PM_IDLE_DOMAIN, (enum pm_state_e)arg);
 		pmvdbg("State unlocked!\n");
 		ret = OK;
 		break;
+	case PMIOC_SLEEP:
+		if (arg > 0) {
+			/* Converting from milliseconds to ticks */
+			arg = ((arg * CLOCKS_PER_SEC) /  1000);
```
ewoodev, line 112: Do we not need to provide an API in milliseconds for kernel functions that use pm_sleep rather than driver?

I think it is better to do this in the pm_sleep() function. And if necessary, please change the name to pm_msleep().

os/include/tinyara/pm/pm.h
```diff
@@ -307,20 +305,25 @@ enum pm_timer_type_e {
...
 };
 
-typedef struct pm_arg_s pm_arg_t;
+typedef struct pm_suspend_arg_s pm_suspend_arg_t;
 
-struct pm_timer_s {
-	uint8_t timer_type;		/* Bits here are set according to the timer that is to be used */
-	uint32_t timer_interval;	/* The interval for whichever timer is to be used */
-	clock_t last_wifi_alive_send_time;       /* Time tick of the last wifi keep alive signal sent time */
+struct pm_wakeup_timer_s {
```
ewoodev, line 316: We use pm_timer in all APIs, But only in this structure data type, we call it pm_wakeup_timer.

Please change it to pm_timer as well.

Comment that our pm_timer means waking up the main core on time.

os/pm/pm_timer/pm_timer_activity.c
```diff
@@ -0,0 +1,184 @@
...
+		curr = curr->next;
+	}
+
+	/* CONFIG_PM_MIN_SLEEP_TIME is the minimum time 
+	 * below which system should not go to sleep. 
+	 * Note that this value should be in millisecond */
+	if (curr->delay > CONFIG_PM_MIN_SLEEP_TIME) {
+		pmvdbg("Setting timer and Board will wake up after %d millisecond\n", curr->delay);
+		/* Converting ticks to nanoseconds */
+		up_set_pm_timer((curr->delay * 1000000) / CLOCKS_PER_SEC);
```
ewoodev, line 138: Please use TICK2NSEC(curr->delay)

os/pm/pm_timedStay.c
```diff
@@ -0,0 +1,134 @@
...
+ *
+ ************************************************************************/
+
+int pm_timedStay(enum pm_state_e state, unsigned int timer_interval)
+{
+	int pid = getpid();
+
+	/* Check if there is already a wdog lock timer running for 
+	 * the process */
+	if (pid_timerMap[pid] != NULL) {
```
ewoodev, line 116: If there is already a timer, it is recommended that we do not clear the timer and start again after wd_cancel.

And if timer_interval is entered as 0, please call wd_cancel and wd_delete only(not start).

os/pm/pm_sleep.c
```diff
@@ -0,0 +1,138 @@
...
+ * Parameters:
+ *   timer_interval - expected board sleep duration (in millisecond)
+ *
+ * Return Value:
+ *   0 - success
+ *   -1 - error
+ *
+ ************************************************************************/
+
+int pm_msleep(int timer_interval)
```
ewoodev, line 102: please change file name to pm_msleep()

os/pm/pm_timer/pm_timer_delete.c
```diff
@@ -0,0 +1,120 @@
...
+        DEBUGASSERT(timer);
+
+        if (PM_ISSTATIC(timer)) {
+                /* Check if this needs to be atomic */
+                state = enter_critical_section();
+
+                /* Put the pm timer back on the free list */
+                sq_addlast((FAR sq_entry_t *)timer, &g_pmTimer_freeList);
+                g_pmTimer_nfree++;
+                DEBUGASSERT(g_pmTimer_nfree <= CONFIG_PM_MAX_STATIC_TIMER);
```
ewoodev, line 107: g_pm_timer_nfree

os/pm/pm_timer/pm_timer.h
```diff
@@ -0,0 +1,156 @@
...
+ * pm wakeup timer expiration time.
+ */
+
+extern sq_queue_t g_pmTimer_activeList;
+
+/* This is the number of free, pre-allocated pm wakeup timer structures in the
+ * g_pmTimer_freeList. 
+ */
+
+extern uint16_t g_pmTimer_nfree;
```
ewoodev, line 111: g_pm_timer_nfree

os/pm/pm_timer/pm_timer.h
```diff
@@ -0,0 +1,156 @@
...
+/* The g_pmTimer_freeList is a singly linked list of pm timers available
+ * to the system */
+
+extern sq_queue_t g_pmTimer_freeList;
+
+/* The g_pmTimer_activeList data structure is a singly linked list ordered by
+ * pm wakeup timer expiration time.
+ */
+
+extern sq_queue_t g_pmTimer_activeList;
```
ewoodev, line 105: g_pm_timer_activelist

os/pm/pm_timer/pm_timer.h
```diff
@@ -0,0 +1,156 @@
...
+#define PM_STATIC      (1 << 0)	        /* Bit 0: pm timer is statically allocated */
+#define PM_ALLOCED      (1 << 1)	/* Bit 1: pm timer is allocated at run time */
+
+#define PM_ISALLOCED(w)  (((w)->flags & PM_ALLOCED) != 0)
+#define PM_ISSTATIC(w)   (((w)->flags & PM_STATIC) != 0)
+
+/* The g_pmTimer_freeList is a singly linked list of pm timers available
+ * to the system */
+
+extern sq_queue_t g_pmTimer_freeList;
```
ewoodev, line 99: g_pm_timer_freelist

os/pm/pm_timedStay.c
```diff
@@ -0,0 +1,135 @@
...
+ *   state - state to be suspended
+ *   timer_interval - expected lock duration in millisecond
+ *
+ * Return Value:
+ *   0 - success
+ *   -1 - error
+ *
+ ************************************************************************/
+
+int pm_timedStay(enum pm_state_e state, unsigned int timer_interval)
```
ewoodev, line 111: Please change name to pm_timedstay.
and filename also

os/pm/pm_timedStay.c
```diff
@@ -0,0 +1,135 @@
...
+ ************************************************************************/
+
+/************************************************************************
+ * Private Variables
+ ************************************************************************/
+
+/* This array maps the pid to their respective wdog timer. 
+ * Assumption: when a thread does a timed lock , only the same thread 
+ * can unlock the "timed lock" before it expire */
+static WDOG_ID pid_timerMap[CONFIG_MAX_TASKS];
```
ewoodev, line 75: Please check coding rule.

g_pid_timer_map[]

os/pm/pm_timedStay.c
ewoodev: I would be failed when a new thread with the same pid hash comes in.

ewoodev (approved):

sunghan-chang (approved):

apps/shell/tash_command.c
```diff
@@ -561,12 +561,11 @@ int tash_execute_cmd(char **args, int argc)
...
 				pmdrv_fd = open("/dev/pm", O_WRONLY);
 				if (pmdrv_fd < 0) {
 					shdbg("open /dev/pm failed(%d), \n", get_errno());
 				} else {
-					pm_suspended = ioctl(pmdrv_fd, PMIOC_SUSPEND, &pm_arg);
```
edwakuwaku, line 569: Hi @ewoodev @gSahitya-samsung @ritesh55555 , the argument for suspend/resume now takes domain instead of PM state. Please revise this part accordingly. Thank you.

PR #6137, opened 2024-04-09
board/common : Support 2nd 32mb of flash

Taejun-Kwon:
1. Update Logic regarding automount
2. Expand w25 driver to support w25q256 series
3. Fix smart.c to support big size of flash

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_boot.c
+++ b/os/board/rtl8730e/src/rtl8730e_boot.c
@@ -141,7 +141,8 @@ void board_spi_initialize(void)
 	}
 #endif
 #endif
-#ifdef CONFIG_AMEBASMART_SPI1
+
+#if defined(CONFIG_AMEBASMART_SPI1) &&  !defined(CONFIG_SECOND_FLASH_PARTITION)
 	spi = up_spiinitialize(1);
 #ifdef CONFIG_SPI_USERIO
 	if (spi_uioregister(1, spi) < 0) {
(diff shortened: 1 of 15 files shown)


Taejun-Kwon:
New Defconfig is necessary, will be done when after it provided.

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -276,7 +276,7 @@ static struct amebasmart_spidev_s g_spi1dev = {
 	.spi_mosi = PB_28,
 	.spi_miso = PB_27,
 	.spi_sclk = PB_25,
-	.spi_cs0 = PB_31,
+	.spi_cs0 = PB_26,
```
sunghan-chang, line 279: https://github.com/Samsung/TizenRT/commit/f90a81026a213f45b239fde173ec9e30367297dc
This is modified already. Let's revert this from this commit.

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -141,7 +141,8 @@ void board_spi_initialize(void)
 	}
 #endif
 #endif
-#ifdef CONFIG_AMEBASMART_SPI1
+
+#if defined(CONFIG_AMEBASMART_SPI1) &&  !defined(CONFIG_SECOND_FLASH_PARTITION)
```
sunghan-chang, line 145: How can we make sure that second flash will be connected to SPI1 always?

Taejun-Kwon: if another external flash attach to another spi, then code need to be expanded in future, I just aligned code with current setting.

sunghan-chang (approved):

PR #6143, opened 2024-04-15
os/: Update docker version for RLT8730E

gSahitya-samsung:
Update toolchain version to 10.3.1 for AI Dual only


edwakuwaku:
AIDual is project name, you should use RTL8730E instead?

build/configs/rtl8730e/loadable_apps/defconfig
```diff
@@ -32,6 +32,7 @@ CONFIG_PASS1_TARGET="all"
 CONFIG_PASS1_OBJECT=""
 CONFIG_BOARD_BUILD_DATE="200204"
 # CONFIG_BINARY_SIGNING is not set
+CONFIG_DOCKER_VERSION="2.0.0"
```
abhishek-samsung, line 35: Please add the config using Kconfig. Otherwise it will disapper when someone uses menuconfig.

sunghan-chang (commented):
Let's split this as two commits.
1. os: set different docker version from defconfig
2. configs/rtl8730e: set a docker version to 2.0.0

build/configs/rtl8730e/loadable_apps/defconfig
gSahitya-samsung: Done

gSahitya-samsung:
> Update toolchain version to 10.3.1 for AI Dual only

Done

gSahitya-samsung:
> Let's split this as two commits.
> 
> 1. os: set different docker version from defconfig
> 2. configs/rtl8730e: set a docker version to 2.0.0

Done

os/Kconfig
```diff
@@ -39,6 +39,12 @@ config DEFAULT_SMALL
 		have already been selected in your configuration file.  This applies
 		only to new settings that require a default value.
 
+config DOCKER_VERSION
+	string "Docker Image Version"
+	default "1.5.8"
+	---help---
+		This config indicates docker image version.
```
sunghan-chang, line 46: Could you add more explanation for this? like `Different chipset requires different toolchain so that this config provides docker version selection for this.`
You should rewrite my sentence with proper English grammar.

sunghan-chang: I am not sure, but we use 6.3.1 for RTL8730E until now. 2.0.0 docker version needs to enable c++ exception on the loadable build. We need to check where we leave the NOTE Prabu said.

gSahitya-samsung: Done

os/Kconfig
```diff
@@ -39,6 +39,18 @@ config DEFAULT_SMALL
...
+	string "Docker Image Version"
+	default "1.5.8"
+	---help---
+		This config indicates docker image version. To ensure compatibility with
+		different chipsets, we have developed distinct version of docker images tailored
+		to each arm-none-eabi toolchain required. This approach streamlines development
+		by providing specific environments that align with the chipset's needs.
+
+		NOTE: RTL8730E supports arm-node-eabi toolchain version 10.3.1. So, please ensure
+		docker version 2.0.0 for RTL8730E boards.
```
sunghan-chang, line 52: @gSahitya-samsung @abhishek-samsung @r-prabu Could you check this location is proper or not? How about c++ exception for RTL8721CSM? With docker version 1.5.8, is it working on it? If not, do we need to mention like this? It looks it does not have a dependency with a chip.

abhishek-samsung: cpp exception need 2.0.0 on tp1x also. By default exceptions are not enabled on tp1x. But if enabled, using 1.5.8 will result in build error.

lib/libxx/Kconfig
```diff
@@ -61,6 +61,11 @@ if LIBCXX
 config LIBCXX_EXCEPTION
 	bool "Enable Exception Suppport"
 	default y
+	--help--
+		Enable exception support in libc++.
+
+		NOTE: C++ exception support requires toolchain version 10.3.1. If
+		you are	enabling C++ exception then use docker image version 2.0.0.
```
edwakuwaku, line 68: Hello @gSahitya-samsung , I have a minor question here. After this PR is merged, if there is some upgrades on other tools or package, then for GCC toolchain v6.3.1 based, the docker version will be eg. 1.5.9/1.6.0?
And for GCC toolchain v10.3.1, the docker version will be eg. 2.0.1/2.1.0?
As I understood, the change of major version is because of difference in GCC toolchain version. Then how will you address other changes in the future~? Will it be similar as mentioned above?

abhishek-samsung: Hello @edwakuwaku , yes, what you said is right. It's the reason why we left a lot of gap and went to 2.0.0 from 1.5.8 .

Fox example if some new package is added to docker, it will be added to both 6.3.1 and 10.3.1 docker images and we will update their version as you said to 2.0.1 and 1.5.9 respectively.

lib/libxx/Kconfig
```diff
@@ -61,6 +61,11 @@ if LIBCXX
 config LIBCXX_EXCEPTION
 	bool "Enable Exception Suppport"
 	default y
+	--help--
```
sunghan-chang, line 64: Let's use three - at each side like others.

lib/libxx/Kconfig
sunghan-chang, line 68: use docker image version 2.0.0 by setting CONFIG_DOCKER_VERSION=2.0.0.

I suggest adding how to set at explanation.

edwakuwaku: Hiiiii @abhishek-samsung , thank you for confirming~

lib/libxx/Kconfig
gSahitya-samsung: Resolved

lib/libxx/Kconfig
gSahitya-samsung: Done

os/Kconfig
```diff
@@ -39,6 +39,20 @@ config DEFAULT_SMALL
 		have already been selected in your configuration file.  This applies
 		only to new settings that require a default value.
 
+config DOCKER_VERSION
+	string "Docker Image Version"
+	default "1.5.8"
```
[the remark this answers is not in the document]
gSahitya-samsung: @sunghan-chang , I had a discussion with @abhishek-samsung, we want to keep "Docker Image Version" inside "Build Setup". Since LIBCXX_EXCEPTION is define in "Built-in Libraries" (below "Build Setup"), hence we can't use this variable for DOCKER_VERSION.

os/dbuild.sh
```diff
@@ -51,7 +50,19 @@ fi
...
 {
+
+	unset CONFIG_DOCKER_VERSION
 	# check existing docker image for specified version
+	if [  -f ${CONFIGFILE} ]; then
+		source ${CONFIGFILE}
+		if [ -n "$CONFIG_DOCKER_VERSION" ]; then
+			DOCKER_VERSION=${CONFIG_DOCKER_VERSION}
+		else
+			DOCKER_VERSION="1.5.8"
```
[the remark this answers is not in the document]
gSahitya-samsung: Done

r-prabu (approved):

sunghan-chang:
@gSahitya-samsung could you check build breaks?

gSahitya-samsung:
> @gSahitya-samsung could you check build breaks?

@sunghan-chang fix the issue. Build error was coming because we were pointing docker tizenrt:latest -> tizenrt:2.0.0

lib/libxx/Kconfig
```diff
@@ -61,6 +61,13 @@ if LIBCXX
 config LIBCXX_EXCEPTION
 	bool "Enable Exception Suppport"
 	default y
+	---help---
+		Enable exception support in libc++.
+
+		NOTE: C++ exception support requires toolchain version 10.3.1. If
+		you are	enabling C++ exception then use docker image version 2.0.0
+		by setting CONFIG_DOCKER_VERSION=2.0.0 or using menuconfig
```
sunghan-chang, line 69: `by setting CONFIG_DOCKER_VERSION=2.0.0 or using menuconfig` This could give wrong information.
When we modify a value of some config, we must use the menuconfig.
It should not be `or`.
Using the menuconfig, we should set CONFIG_DOCKER_VERSION with 2.0.0 value.

sunghan-chang (approved):

PR #6144, opened 2024-04-15
os/board/rtl8721csm: add ble one-shot advertising function

yeetee179:
- add in one-shot-adv function to send an adv
- warning fixes


hs36-kim (approved):
LGTM

PR #6145, opened 2024-04-15
PM: add suppression operation in pm

ewoodev:
- os/arch, os/pm: Remove chip-dependent elements in pm module.
- os: Add tick suppression operation in PM

- rtl8730e: Modify to compensate for sys-ticks within pm


os/pm/pm_wake.c
```diff
@@ -0,0 +1,62 @@
...
+ *   missing_tick - Missed ticks while the core was sleeping.
+ *
+ * Returned value:
+ *   None
+ *
+ ****************************************************************************/
+
+void pm_wake(int missing_tick)
+{
+	irqstate_t flags  = irqsave();
```
edwakuwaku, line 52: I think enter/leave_critical_section should be used here. Since we ported SMP, we are using enter/leave_critical_section. Even without SMP, enter/leave_critical_section will be redirected to irqsave/restore. Please check if a change is necessary.

os/kernel/clock/clock_initialize.c
```diff
@@ -339,4 +339,11 @@ void clock_timer(void)
 
 	g_system_timer++;
 }
+
+#ifdef CONFIG_SCHED_TICKSUPPRESS
+void clock_timer_nohz(int ticks)
+{
+	g_system_timer += ticks;
```
edwakuwaku, line 346: g_system_timer is a u64 variable. But the input ticks is set as int type. What is the strategy to handle overflow cases? Even if the input type is int, shouldn't it be casted to (u64)?

os/include/tinyara/pm/pm.h
```diff
@@ -670,6 +670,24 @@ int pm_changestate(int domain, enum pm_state_e newstate);
 
 enum pm_state_e pm_querystate(int domain);
 
+#ifdef CONFIG_PM_TICKSUPPRESS
+/****************************************************************************
+ * Name: pm_wakeup
```
edwakuwaku, line 675: pm_wake?

os/pm/pm_dvfs.c
```diff
@@ -80,7 +80,7 @@
  *
  * Description:
  *   This internal function is called to reduce the frequency of operation
- *   of the secondary AP core of AI_Dual chipset.
```
edwakuwaku, line 83: This internal function is called to tune (ie. Not recommend to use reduce here because it is dynamic tuning) the clock frequency of the application core. Do note that, it will only affect the power consumption during active mode, it can be applied on some cases which low loading tasks is expected. Please keep in mind that there is a tradeoff between processing speed and cpu active time interval. Which means, we will expect to see lower power consumption during that period of time (ie. reduced clock frequency), but the time interval to handle some task will also consume longer time.

os/pm/pm_wake.c
```diff
@@ -0,0 +1,62 @@
...
+#include <arch/irq.h>
+
+#include "pm.h"
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+/****************************************************************************
+ * Name: pm_wakeup
```
edwakuwaku, line 36: pm_wake?

os/pm/pm_wake.c
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Input parameters:
+ *   missing_tick - Missed ticks while the core was sleeping.
+ *
+ * Returned value:
+ *   None
+ *
+ ****************************************************************************/
+
+void pm_wake(int missing_tick)
```
ritesh55555, line 50: Can you check the variable type of missing_tick ? will it be int or something else?

os/board/rtl8730e/src/component/soc/amebad2/misc/ameba_tizenrt_pmu.c
```diff
@@ -302,9 +302,7 @@ void tizenrt_pre_sleep_processing(uint32_t *expected_idle_time)
 #ifndef CONFIG_PLATFORM_TIZENRT_OS
 	vTaskStepTick(ms_passed); /*  update kernel tick */
 #else
-	irqstate_t flags  = irqsave();
-	g_system_timer += (u64)ms_passed; /*  update kernel tick */
-	irqrestore(flags);
+	pm_wake(ms_passed); /*  update kernel tick */
```
ritesh55555, line 305: we should also add config_pm check here

os/pm/pm_wake.c
ewoodev: Thank you for view. Changed

os/kernel/clock/clock_initialize.c
ewoodev: Thank you for view. Changed

os/include/tinyara/pm/pm.h
ewoodev: Thank you for view. Changed

os/pm/pm_wake.c
ewoodev: Done

os/board/rtl8730e/src/component/soc/amebad2/misc/ameba_tizenrt_pmu.c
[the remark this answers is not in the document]
ewoodev: Thank you for reivew, added

os/pm/pm_wake.c
ewoodev: Done

os/board/rtl8730e/src/component/soc/amebad2/misc/ameba_tizenrt_pmu.c
ewoodev: Done

os/board/rtl8730e/src/component/soc/amebad2/misc/ameba_tizenrt_pmu.c
```diff
@@ -256,6 +256,15 @@ VOID pg_aontimer_int(u32 Data)
 }
 #endif
 
+
+#ifdef CONFIG_PM_TICKSUPPRESS
+static void (*tizenrt_sleep_handler)(clock_t);
+void up_register_wakehandler(void (*handler)(clock_t))
+{
+	tizenrt_sleep_handler = handler;
```
ewoodev, line 264: @edwakuwaku Hello edward, We added callback code to separate the strong dependency of the upper layer, PM, and the lower layer, BSP code. This is a temporary code. We will refactor PM's code through PR in the future. Please refer to it.

os/include/tinyara/arch.h
```diff
@@ -2141,6 +2141,45 @@ int up_rtc_getdatetime(FAR struct tm *tp);
 int up_rtc_settime(FAR const struct timespec *tp);
 #endif
 
+/****************************************************************************
+ * Name: up_register_wakehandler
+ *
+ * Description:
+ *   This function registers a handler to be called when the core wakes up
```
edwakuwaku, line 2148: Suggest to change as below:
"to be called"-> "for systick compensation"

os/board/rtl8730e/src/component/soc/amebad2/misc/ameba_tizenrt_pmu.c
edwakuwaku: 안녕하세요 은우님~ Yes I agree with you that there is strong dependency with current upper layer (application), middle/FW layer (PM driver), and lower layer which is BSP. I see there are ongoing development for clear separation between those layers from you and Ritesh.
This changes looks fine to me, please let me know if you need any help regarding PM revision.

os/include/tinyara/arch.h
ewoodev: Thank you for your suggestion.
I changed~

PR #6147, opened 2024-04-17
os/drivers/lcd : st7789 init code update

anjana348:
issue: music app was showing wrong geometry

solution: driver code was updated in product_AIDual_Bixby but not in TizenRT st7789_init is called in st7789_lcdinitialize


sunghan-chang (commented):
I don't get the point why this is necessary. Could you leave more details?

sunghan-chang (commented):
Is this a right sequence regardless what we did in internal repo? I want to know what is right code in LCD driver point of view.

anjana348:
> Is this a right sequence regardless what we did in internal repo? I want to know what is right code in LCD driver point of view.

Yes, it is right. It is verified

sunghan-chang (approved):

PR #6148, opened 2024-04-17
os/: st7701 LCD driver with MIPI

namanjain7:


os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,351 @@
...
+};
+
+struct mipi_irq {
+	u32 irq_num;
+	u32 irq_data;
+	u32 irq_priority;
+};
+
+/* Helpers */
+static void set_clock();
```
sunghan-chang, line 70: need `void`

`(void)`

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,351 @@
...
+#if defined(CONFIG_DEBUG_MIPI)
+#define mipiinfo(format, ...)   printf(format, ##__VA_ARGS__)
+#define mipierr(format, ...)    printf(format, ##__VA_ARGS__)
+#else
+#define mipiinfo(format, ...)
+#endif
+#define mipierr(format, ...)   printf(format, ##__VA_ARGS__)
+
+u32 vo_freq;
+u32 send_cmd_done = 0;
```
sunghan-chang, line 55: maybe `static`

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,208 @@
...
+		lldbg("LCD driver register success\n");
+	}
+	rtl8730_st7701_enable_lcdc();
+	
+	rtl8730e_st7701_reset_pin(1);
+	DelayMs(10);
+	rtl8730e_st7701_reset_pin(0);
+	DelayMs(10);
+	rtl8730e_st7701_reset_pin(1);
+	DelayMs(120);
```
sunghan-chang, line 205: There are `DelayMs`s between each controlling the reset pin.
Does they come from the data sheet of ST7701?

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,208 @@
...
+	config.YPixels = YRES;
+
+	struct mipi_dsi_host *dsi_host = (struct mipi_dsi_host *)mipi_dsi_host_initialize(&config);
+	struct mipi_dsi_device *dsi_device = (struct mipi_dsi_device *)mipi_dsi_device_register(dsi_host, "dsi", 0);
+	struct lcd_dev_s *dev = (struct lcd_dev_s *)st7701_lcdinitialize(dsi_device, &g_rtl8730_st7701_config_s);
+	LcdcInitValues(config);
+	rtl8730_st7701_lcd_init();
+
+	if (lcddev_register(dev) < 0) {
+		lldbg("ERROR: LCD driver register fail\n");
```
sunghan-chang, line 194: No need to return before next code execution? I mean 
is it ok to execute line 198 even this is failed?

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,208 @@
...
+	rtl8730_st7701_enable_lcdc();
+	
+	rtl8730e_st7701_reset_pin(1);
+	DelayMs(10);
+	rtl8730e_st7701_reset_pin(0);
+	DelayMs(10);
+	rtl8730e_st7701_reset_pin(1);
+	DelayMs(120);
+
+	return;
```
sunghan-chang, line 207: Unnecessary code. Let's remove.

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -84,6 +84,7 @@
 #include "ameba_soc.h"
 #include "osdep_service.h"
 #include "platform_opts_bt.h"
+#include "rtl8730e_st7701.c"
```
sunghan-chang, line 87: looks strange. You already add it in Make.defs. I think you need a header for `rtl8730e_st7701_initialize` prototype.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -234,6 +234,49 @@ static void test_orientation(void)
 	close(fd);
 }
 
+#ifdef CONFIG_LCD_ST7701
+static void test_color_st7701(void)
```
sunghan-chang, line 238: Could you explain why we need this? I mean 
why can't we use the existed example?

sunghan-chang:
@namanjain7 You already implement mipi lowerhalf for amebasmart. Right? @zhongnuo-tang Could you review the first commit for RTL8730e?

namanjain7:
Since there are lots of files in one PR, so I have fixed the comments and committed to this PR which added new commits. Please ignore the number of commits, as I will raise a new PR once all the comments are fixed.

os/board/rtl8730e/src/rtl8730e_st7701.c
namanjain7: Yes sir, it's from the data-sheet of st7701. Page number: 54
The reset time should be longer than 9ms. And the maximum time for reset is 120ms.

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,208 @@
...
+		lcdc_init_struct.layerx[idx].LCDC_LayerImgBaseAddr = (u32) lcd_img_buffer1;
+		lcdc_init_struct.layerx[idx].LCDC_LayerHorizontalStart = 1;
+		lcdc_init_struct.layerx[idx].LCDC_LayerHorizontalStop = config.XPixels;
+		lcdc_init_struct.layerx[idx].LCDC_LayerVerticalStart = 1;
+		lcdc_init_struct.layerx[idx].LCDC_LayerVerticalStop = config.YPixels;
+		lcdc_init_struct.layerx[idx].LCDC_LayerColorKeyingEn = ENABLE;
+	}
+}
+
+static u32 LCDCRgbHvIsr(void *Data)
```
zhongnuo-tang, line 73: This irq handler is not registered in anywhere. 
Shall we register this in rtl8730_st7701_lcd_init?

namanjain7: yes, this irq was registered in original file, but I didnt' see any usecase for this. The LCD is working fine without this irq.
Instead there is another irq handler in amebasmart_mipi.c file which seems useful .
But since you mentioned it, I am rechecking this irq handler from documentation again.

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,198 @@
...
+	lcdc_init_struct.layerx[0].LCDC_LayerVerticalStart = y1;
+	lcdc_init_struct.layerx[0].LCDC_LayerVerticalStop = y2;
+	*/
+	LCDC_LayerConfig(pLCDC, 0, &lcdc_init_struct.layerx[0]);
+	LCDC_TrigerSHWReload(pLCDC);
+}
+
+static void rtl8730e_st7701_reset_pin(u8 Newstatus)
+{
+	u32 pin_name = _PA_14;
```
edwakuwaku, line 147: How about changing this to `#define GPIO_RESET_PIN _PA_14`

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,198 @@
...
+		lcdc_init_struct.layerx[layer].LCDC_LayerEn = DISABLE;
+	}
+}
+
+static void rtl8730_st7701_lcd_put_area(u8 *lcd_img_buffer, u32 x1, u32 y1, u32 x2, u32 y2)
+{
+	lcdc_init_struct.layerx[0].LCDC_LayerImgBaseAddr = (u32) lcd_img_buffer;
+
+	//Uncomment below lines in case LCD window size needs to be changed...
+	/*
```
edwakuwaku, line 135: I'm not sure is this the only way to achieve change in window size, perhaps using some #if CONFIG..... to control?
Also, if the input arguments is not used, you should void it.
eg: `(void) x1; ......`

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+		MIPIx->MIPI_DPHY_ERR = reg_dphy_err;
+		if (MIPIx->MIPI_CONTENTION_DETECTOR_AND_STOPSTATE_DT & MIPI_MASK_DETECT_ENABLE) {
+			MIPIx->MIPI_CONTENTION_DETECTOR_AND_STOPSTATE_DT &= ~MIPI_MASK_DETECT_ENABLE;
+
+			MIPIx->MIPI_DPHY_ERR = reg_dphy_err;
+			MIPI_DSI_INTS_Clr(MIPIx, MIPI_BIT_ERROR);
+		}
+
+		if (MIPIx->MIPI_DPHY_ERR == reg_dphy_err) {
+			lldbg("LPTX Still Error\n");
```
edwakuwaku, line 241: There are many error log using lldbg, please use dbg instead.

edwakuwaku:
Hi @namanjain7 , since this PR is verified LCD working MIPI, thus could you leave simple comment regarding performance for different settings? 
Eg: RGB565: xx FPS
RGB888: xx FPS

os/drivers/mipidsi/Make.defs
```diff
@@ -0,0 +1,30 @@
...
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+############################################################################
+
+ifeq ($(CONFIG_MIPI_DSI),y)
+
+CSRCS += mipi_dsi_device.c mipi_dsi_host.c mipi_dsi_packet.c mipi_dsi_device_driver.c mipi_dsi_host_driver.c
+
+ifeq ($(CONFIG_MIPI_DSI_DRIVER),y)
```
edwakuwaku, line 23: Is there panel_st7701.c? If not, how about just remove this?

apps/examples/lcd_test/example_lcd.c
namanjain7: For SPI, the putarea coordinates are : (0, 0) -> (XRES - 1, YRES - 1)
but for MIPI, the putarea coordinates are: (1, 1) -> (XRES, YRES)
So, either we have to change the existing putarea and add separate conditions for both LCD or define a new function separately MIPI based LCD.

namanjain7: It's the property of LCDC structure, that the coordinates begin with (1, 1) and end with (480, 800)

sunghan-chang: LCD Driver provides a HAL feature and above is chip-specific. Is there no way to convert above in LCD driver? If not, when we apply new HW, app should be modified. Then it means LCD Driver does not provide a HAL feature.

sunghan-chang: I think this is wrong.

os/board/rtl8730e/src/rtl8730e_boot.c
sunghan-chang: Could you check this comment?

os/drivers/mipidsi/mipi_dsi_device.c
```diff
@@ -0,0 +1,981 @@
...
+
+	msg.channel = device->channel;
+	msg.type = MIPI_DSI_COMPRESSION_MODE;
+	msg.tx_len = sizeof(tx);
+	msg.tx_buf = tx;
+	msg.flags = 0;
+
+	ret = mipi_dsi_transfer(device, &msg);
+	return ret < 0 ? ret : 0;
+}
```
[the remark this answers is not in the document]
sunghan-chang: Oh.. I didn't finish my suggestion. What I want is making one definition for common usages and each function inputs different values like tx and type.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,359 @@
...
+{
+	return 0;
+}
+
+static int amebasmart_mipi_detach(FAR struct mipi_dsi_host *dsi_host, FAR struct mipi_dsi_device *dsi_device)
+{
+	return 0;
+}
+
+static int amebasmart_mipi_transfer(FAR struct mipi_dsi_host *dsi_host, FAR const struct mipi_dsi_msg *msg)
```
edwakuwaku, line 320: There are 2 modes under MIPI DSI, one is video mode and one is command mode. During initialization stage, we are in reset status which is command mode, and we will send in a table of command list for configuring the LCD. But, if we need to change some configuration during runtime (ie. sending DCS command), I think we should check current mode, if we are under video mode, then we should toggle the state before sending a command, else it might not work.
So, we should add a check current DSI mode, and if it's video mode, we should invoke `MIPI_DSI_Mode_Switch(MIPI, DISABLE);` before `amebasmart_mipidsi_send_cmd`, and `MIPI_DSI_Mode_Switch(MIPI, ENABLE);` after sending command.

namanjain7: ok, will add the check, 
I have a question: Will there be any case that during runtime we will change the LCD configuration. As far as I know, LCD configuration should required to set in bootup. Why anyone will change the configuration when it's running?

edwakuwaku: It depends on application usage, but I believe more request will come in along the way, thus we should get prepared here.
From my understanding, there is already a scenario which is when integrating with PM module, we will have to set backlight dimming according to different state (ie. STANDBY state), and that requires sending DCS command during runtime. (ie. This is just one of the example use case, it should not be dependent to just PM module)
In general, this API should be applicable for both initialisation and runtime for configuring LCD, instead of just setting it once during bootup sequence.

os/board/rtl8730e/src/rtl8730e_boot.c
namanjain7: thanks for pointing this out, this doesn't needed. I have removed this.

apps/examples/lcd_test/example_lcd.c
namanjain7: yes, I didn't realised that. 
So, instead of doing this in  application part, I have added the changes in st7701.c driver.

namanjain7: so, now application wont' be needed to reconfigured differently for mipi and spi.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -257,4 +257,5 @@ int lcd_test_main(int argc, char *argv[])
 		sleep(1);
 		count++;
 	}
+	return 0;
```
namanjain7, line 260: Added this due to warning message during build.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -149,8 +149,8 @@ static void test_put_run(void)
 	run.npixels = NOPIXELS;
 
 	uint8_t spi_data[2 * NOPIXELS + 1];
-	run.data = &spi_data;
-	for (i = 0; i <= (NOPIXELS * 2); i += 2) {
+	run.data = spi_data;
+	for (i = 0; i < (NOPIXELS * 2); i += 2) {
```
namanjain7, line 153: Fixed the logical error. It will go out of bound of array.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+		if (packet.header[1] == 0) {
+			amebasmart_mipidsi_send_cmd(priv->MIPIx, packet.header[0], 0, NULL);
+		} else {
+			amebasmart_mipidsi_send_cmd(priv->MIPIx, packet.header[0], 1, packet.header + 1);
+		}
+	} else {
+		amebasmart_mipidsi_send_cmd(priv->MIPIx, packet.header[0], packet.payload_length, packet.payload);
+	}
+
+	while (!ameabsmart_is_transfer_completed()) {
```
edwakuwaku, line 341: This seems having much delay (ie. 1ms), are we supposed to wait for such long time? Just polling is not enough?
The intention of adding delay here seems that we are sending a table of command list during initialization stage. But if we are just sending command individually, perhaps the delay is unnecessary.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+	if (reg_val) {
+		dbg("LPTX Error Occur: 0x%x\n", reg_val);
+	}
+
+	if (reg_val2) {
+		dbg("error occured #\n");
+	}
+}
+
+static void ameabsmart_check_freq(struct lcd_data *data)
```
edwakuwaku, line 252: Some of the API naming is having typo, could you please check?

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+	u32 mipi_ckd = PLL_CLK / vo_freq - 1;
+	HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LSYS_CKD_GRP0, (HAL_READ32(SYSTEM_CTRL_BASE_LP, REG_LSYS_CKD_GRP0) & ~LSYS_MASK_CKD_MIPI) | LSYS_CKD_MIPI(mipi_ckd));
+	HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LSYS_CKD_GRP0, (HAL_READ32(SYSTEM_CTRL_BASE_LP, REG_LSYS_CKD_GRP0) & ~LSYS_MASK_CKD_HPERI) | LSYS_CKD_HPERI(3));
+
+	RCC_PeriphClockCmd(APBPeriph_NULL, APBPeriph_HPERI_CLOCK, ENABLE);
+	RCC_PeriphClockCmd(APBPeriph_LCDC, APBPeriph_LCDCMIPI_CLOCK, ENABLE);
+
+}
+
+static void ameabsmart_register_interrupt(void)
```
edwakuwaku, line 274: Please check naming

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+{
+	InterruptDis(mipi_irq_info.num);
+	InterruptUnRegister(mipi_irq_info.num);
+	InterruptRegister((IRQ_FUN) amebasmart_mipidsi_isr, mipi_irq_info.num, (u32) mipi_irq_info.data, mipi_irq_info.priority);
+	InterruptEn(mipi_irq_info.num, mipi_irq_info.priority);
+}
+
+static bool ameabsmart_is_transfer_completed(void)
```
edwakuwaku, line 282: Please check naming

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,355 @@
...
+		amebasmart_mipidsi_send_cmd(priv->MIPIx, packet.header[0], packet.payload_length, packet.payload);
+	}
+
+	while (!ameabsmart_is_transfer_completed()) {
+		usleep(1000);
+	}
+	return OK;
+}
+
+struct mipi_dsi_host *ameabsmart_mipi_dsi_host_initialize(struct lcd_data *config)
```
edwakuwaku, line 347: Please check naming

os/drivers/lcd/st7701.c
```diff
@@ -0,0 +1,362 @@
...
+
+	struct mipi_dsi_msg msg;
+	while (1) {
+		cmd = table[send_cmd_idx_s].cmd;
+		switch (cmd) {
+		case REGFLAG_DELAY:
+			usleep(table[send_cmd_idx_s].count * 1000);
+			break;
+		case REGFLAG_END_OF_TABLE:
+			printf("LPTX CMD Send Done\n");
```
edwakuwaku, line 135: Is this log necessary? Even it is, why not using `lcdvdbg`?

os/drivers/mipidsi/mipi_dsi_packet.c
```diff
@@ -0,0 +1,235 @@
...
+ ****************************************************************************/
+
+int mipi_dsi_create_packet(FAR struct mipi_dsi_packet *packet, FAR const struct mipi_dsi_msg *msg)
+{
+	if (packet == NULL) {
+		printf("packet is NULL \n");
+		return -EINVAL;
+	}
+	if (msg == NULL) {
+		printf("msg is NULL \n");
```
edwakuwaku, line 192: MIPI seems to be an individual driver which is tightly coupled with LCD, maybe you want to define another logging type for it. etc: `mipidbg`, `mipivdbg`? Please find os/include/debug.h.
I'm not very sure, maybe @sunghan-chang could provide better comment on this.

os/drivers/lcd/st7701.c
namanjain7: I think from information perspective, it would be good idea to print when the last command has been transferred.
It should help debug in future that the bootup sequence is completed.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
namanjain7: I have removed the delay and changed the while loop to this:

	while(true){
		if(send_cmd_done == 1){
			break;
		}
		count++;
		if(count >=20){
			dbg("Error: amebasmart mipi send cmd timeout\n");
		}
	}

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,250 @@
...
+		lcdc_init_struct.layerx[layer].LCDC_LayerEn = ENABLE;
+	} else {
+		lcdc_init_struct.layerx[layer].LCDC_LayerEn = DISABLE;
+	}
+}
+
+static void rtl8730_st7701_lcd_put_area(u8 *lcd_img_buffer, u32 x_start, u32 y_start, u32 x_end, u32 y_end)
+{
+	lcdc_init_struct.layerx[LCD_LAYER].LCDC_LayerImgBaseAddr = (u32) lcd_img_buffer;
+#ifdef ENABLE_ST7701_CHANGE_WINDOW_SIZE
```
edwakuwaku, line 135:
```
#ifndef ENABLE_ST7701_CHANGE_WINDOW_SIZE
	lcdc_init_struct.layerx[LCD_LAYER].LCDC_LayerHorizontalStart = x_start;
	lcdc_init_struct.layerx[LCD_LAYER].LCDC_LayerHorizontalStop = x_end;
	lcdc_init_struct.layerx[LCD_LAYER].LCDC_LayerVerticalStart = y_start;
	lcdc_init_struct.layerx[LCD_LAYER].LCDC_LayerVerticalStop = y_end;
#else
	(void) x_start;
	(void) y_start;
	(void) x_end;
	(void) y_end;
#endif
```

edwakuwaku: May I know where will you define ENABLE_ST7701_CHANGE_WINDOW_SIZE?

os/arch/arm/src/amebasmart/amebasmart_mipi.c
edwakuwaku: I have 2 questions:
1) How do you get the value of 20?
2) What if send_cmd_done never gets updated? Maybe we should add some error handling instead of just printing timeout logs (ie. return false?)

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -0,0 +1,352 @@
...
+	amebasmart_mipi_fill_buffer(&(priv->dsi_host));
+	MIPI_Init(priv->MIPIx, priv->MIPI_InitStruct);
+
+	MIPI_DSI_TO1_Set(priv->MIPIx, DISABLE, 0);
+	MIPI_DSI_TO2_Set(priv->MIPIx, ENABLE, 0x7FFFFFFF);
+	MIPI_DSI_TO3_Set(priv->MIPIx, DISABLE, 0);
+
+	amebasmart_register_interrupt();
+	MIPI_DSI_INT_Config(priv->MIPIx, DISABLE, ENABLE, FALSE);
+	MIPI_DSI_init(priv->MIPIx, priv->MIPI_InitStruct);
```
edwakuwaku, line 176: The `MIPI_DSI_init` here can be removed, it has been satisfied in `MIPI_Init `already.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
namanjain7: Hello, I tried to switch the mode before and after the transfer function. I observed that it requires some sleep after mode switch. If there is no sleep, then there is a crash. So, we would need to apply sleep twice for both mode switch.
So, call is like:
modeswitch from video -> cmd
some sleep
transfer
modeswitch from cmd -> video
return

I dont' get it why mode switching requires sleep. And if it's required to apply sleep then, would it be good idea to apply sleep in transfer?

os/board/rtl8730e/src/rtl8730e_st7701.c
namanjain7: I will change the naming to CONFIG_ENABLE_ST7701_CHANGE_WINDOW_SIZE

os/arch/arm/src/amebasmart/amebasmart_mipi.c
edwakuwaku: Hello @namanjain7 , I'm not sure why is there a crash for mode switch. Could you attach the crash log via email or some other platforms?
Also, other than mode switch, I think we should also do a 
`MIPI_DSI_INT_Config(MIPIx, DISABLE, ENABLE, FALSE);` to enable the interrupt for command mode.
And do a `MIPI_DSI_INT_Config(MIPIx, DISABLE, DISABLE, FALSE);` after sending command

os/arch/arm/src/amebasmart/amebasmart_mipi.c
namanjain7: Hello, I have changed this code. It seems that it's not possible to exit from while loop without adding sleep. It gets stuck in the while loop. Not sure about the reason. I have added the 1ms delay. It's the lowest I could go when I ran the code in hardware.
The reason it was working before is that I have added some debug logs before the while loop. It seems that debug log results in some delay. When I remove the debug log just before the while loop, then code gets stuck in the while loop.

The new code is working properly , ... I verified many times.

os/board/rtl8730e/src/rtl8730e_st7701.c
namanjain7: I have changed this.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
namanjain7: [error_logs.txt](https://github.com/Samsung/TizenRT/files/15219893/error_logs.txt)

This file attached contains error logs.

namanjain7: Error in hv_isr

r0bindal:
> Hi @namanjain7 , since this PR is verified LCD working MIPI, thus could you leave simple comment regarding performance for different settings? Eg: RGB565: xx FPS RGB888: xx FPS

 RGB565 : 357.398 FPS
 RGB888 : 201.89 FPS

os/arch/arm/src/amebasmart/amebasmart_mipi.c
edwakuwaku: Hi @namanjain7 , could you please revert this request first? I mean you can ignore the runtime sending DCS command, just leave it as the initial working codes.
I will add the necessary changes together with PM with MIPI in another PR.

Sorry for inconvenience caused.

os/drivers/mipidsi/mipi_dsi_device.c
```diff
@@ -0,0 +1,951 @@
...
+ *
+ ****************************************************************************/
+
+int mipi_dsi_dcs_set_page_address(FAR struct mipi_dsi_device *device, uint16_t start, uint16_t end)
+{
+	uint8_t payload[4] = {
+		start >> 8,
+			  start & 0xff,
+			  end >> 8,
+			  end & 0xff
```
sunghan-chang, line 735: Could you align them? Same for other payloads

os/drivers/lcd/st7701.c
```diff
@@ -0,0 +1,368 @@
...
+	priv->dev.getplaneinfo = st7701_getplaneinfo;
+	priv->dev.getvideoinfo = st7701_getvideoinfo;
+	priv->dev.getpower = st7701_getpower;
+	priv->dev.setpower = st7701_setpower;
+	priv->dev.getcontrast = st7701_getcontrast;
+	priv->dev.setcontrast = st7701_setcontrast;
+	priv->dev.init = st7701_init;
+	priv->dsi_dev = dsi;
+	priv->config = config;
+	if(st7701_send_init_cmd(priv->dsi_dev, ST7701S_init_cmd_g) == OK){
```
sunghan-chang, line 361: if (....) {

Let's add a space before and after `( )`.

os/drivers/lcd/st7701.c
```diff
@@ -0,0 +1,368 @@
...
+		printf("LCD Init sequence completed\n");
+	}
+	else{
```
sunghan-chang, line 364: } else {

os/drivers/lcd/st7701.c
```diff
@@ -0,0 +1,368 @@
...
+
+static struct st7701_dev_s g_lcdcdev;
+
+typedef struct lcm_setting_table {
+	u8 cmd;
+	u8 count;
+	u8 para_list[128];
+} lcm_setting_table_t;
+
+static lcm_setting_table_t ST7701S_init_cmd_g[] = {	/* DCS Write Long */
```
edwakuwaku, line 53: As I have checked, this is consuming up much RAM, around ~5kB, since we are using it only for initialization phase, it should be changed to dynamic allocation, so we can free up the space after initializing the LCD. Please check if a change is necessary.

edwakuwaku:
Hi @namanjain7 , may I know the status of this PR, could you help to clear the remaining comments? FYI, the next step is to integrate PM with MIPI LCD, and that will take time for us to make relevant changes. Thus, this PR is mandatory to be merged asap.

namanjain7:
> remaining comments? FYI, the next step is to integrate PM with MIPI LCD, and that will take time for us to make relevant changes. Thus, this PR is mandatory to be merged asap.

Hello sir, yes I am clearing the remaining comments. I have fixed almost all of them except in one comment where I am getting some error when adding the changes in debug.h file. I am fixing that and will update the PR by tomorrow for sure.

os/drivers/lcd/st7701.c
ziyik: Or maybe it can be changed to constant, so it only consumes Flash? Since the values are used to initial LCD without modifying its original values.

os/board/rtl8730e/src/rtl8730e_st7701.c
```diff
@@ -0,0 +1,249 @@
...
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <tinyara/lcd/st7701.h>
+#include <tinyara/mipidsi/mipi_dsi.h>
+#include "objects.h"
+#include "PinNames.h"
+#include "gpio_api.h"
+
+#define MIPI_GPIO_RESET_PIN _PA_14
```
ziyik, line 29: Is the latest AILite used PA15 as LCD Reset? Could you check?

namanjain7:
The PR #6188 has been merged. I am closing this.

PR #6150, opened 2024-04-19
os/arch/arm/src/amebasmart, os/drivers/lcd: Integrate SPI driver with PM

edwakuwaku:


sunghan-chang (commented):
@edwakuwaku I am sorry but I think little bit different.
Is this a reinit of LCD? If we use a SPI for other device and we don't support a LCD, then how can we call the reinit?
I think regardless of LCD, it's a necessary operation in SPI when hw does not manage this.
1. reinit should be in each peripheral like spi, mipi or something else which should be reinit.
2. Some chip will manage this case automatically. I mean sw does not do this with some chip. Then, we should add this code with new config like `CONFIG_PM_SW_RESTORE_PERI` (name is much poor, need to rename).

sunghan-chang:
@ewoodev could you check this?

ewoodev (approved):

sunghan-chang (approved):

PR #6152, opened 2024-04-23
arm/src/amebasmart: revert changes made SPI pins to support old board

abhishek-samsung:
SPI pins were modified to a newer board version, which is no longer used. So, revert the change.


sunghan-chang (approved):

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -234,7 +234,7 @@ static struct amebasmart_spidev_s g_spi0dev = {
 	.spi_sclk = PB_6,
 	.spi_cs0 = PB_5,
 #if defined(CONFIG_SPI_CS) && defined(CONFIG_AMEBASMART_SPI0_CS)
-	.spi_cs1 = PB_26,
```
r-prabu, line 237: Don't we have a provision to select the Pins from config?

abhishek-samsung: No, we don't have it as of now. Once the board configuration is fixed, we will not need to change the pins. So, I am not sure if it's really required.

PR #6153, opened 2024-04-23
Virtual Key Driver Review update

anjana348:
uint16 is used as uint8 can contain only 255 entries. It was giving wrong values.

semaphore was not released in return condition, which might have caused issues.

Indentation is done uniform with tabs

Updated review comments


sunghan-chang (commented):
1. We don't allow empty description. Could you write what and why you modify in the commit descrption? (not in pr description)
2. `review comments update` for commit title looks not good. Any other understandable title what this commit is from a title?

PR #6154, opened 2024-04-23
os/arch/arm/src/amebasmart: optimize SPI DMA

zhongnuo-tang:
1. remove delay in spi_dmatxwait
2. replace spi_master_write_stream_dma with spi_master_write_read_stream_dma
3. implement data slicing

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_spi.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_spi.c
@@ -527,7 +527,7 @@ static void spi_dmatxwait(FAR struct amebasmart_spidev_s *priv)
 	}
 	/*DMA triggers irq when all data has transferred from memory to SPI fifo, it is around 0.5ms earlier than SPI transfer all data in FIFO out
 	 *compensate by adding 0.5ms delay to ensure that all data in fifo are transferred out before manually pull up CS pin by GPIO*/
-	DelayUs(500); 
+
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -1157,54 +1157,93 @@ static void amebasmart_spi_exchange(FAR struct spi_dev_s *dev,
...
-		spi_dmatxwait(priv);
-		memcpy(rxbuffer, rxbuff_aligned, nwords * mode_16bit);
-		rtw_mfree(rxbuff_aligned, 0);
-		rtw_mfree(txbuff_aligned, 0);
-	} else if (txbuffer) {
-		uint8_t *txbuff_aligned =  NULL;
-		txbuff_aligned = (uint8_t *)rtw_zmalloc(SPI_DMA_MAX_BUFFER_SIZE);
-		if(txbuff_aligned == NULL) {
-			return;
+		while (len >= SPI_DMA_MAX_BUFFER_SIZE) {
```
sunghan-chang, line 1181:
```
state = 0;
if (txbuffer) {
   alloc txbuff_alinged with SPI_DMA_MAX_BUFFER_SIZE;
   state += 1;
}
if (rxbuffer) {
  alloc rxbuff_aligned with SPI_DMA_MAX_BUFFER_SIZE;
  state += 2;
}

while (remain_data_len > 0) {
   if (remain_data_len >= SPI_DMA_MAX_BUFFER_SIZE) {
      send_data_len = SPI_DMA_MAX_BUFFER_SIZE;
   } else {
      send_data_len = remain_data_len;
   }

  switch (state) {
  case 3: // tx+rx
    do r/tx;
    break;
  case 2: // rx only
     do rx only
     break;
  case 1: // tx only
      do tx only
      break;
   case 0: // do nothing
      break;
   }

   remain_data_len -= send_data_len;
}

if (txbuff_aligned)
  free

if (rxbuff_aligned)
  free
```

It looks it has duplicated code, and even rxbuffer is null, it allocates rxbuff_aligned always.
How about above?

abhishek-samsung: rx buffer is required as we asked @zhongnuo-tang to remove the delay (0.5ms) after spi tx transfer. Delay was slower when we compared it using read_write (both tx and rx) for tx.

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -1157,54 +1157,93 @@ static void amebasmart_spi_exchange(FAR struct spi_dev_s *dev,
...
 		/* 16-bit mode */
 		mode_16bit++;
 	}
-	DEBUGASSERT((nwords * mode_16bit) <= SPI_DMA_MAX_BUFFER_SIZE);
-	if (txbuffer && rxbuffer) {
-		uint8_t *rxbuff_aligned = NULL;
+	len = nwords * mode_16bit;
+	uint8_t *rxbuff_aligned = NULL;
+	uint8_t *txbuff_aligned =  NULL;
+	if (len >= SPI_DMA_MAX_BUFFER_SIZE) {
```
abhishek-samsung, line 1169: I understand that max DMA page size is 64KB. However, I think 64KB is too big. Also, as exchange return type is NULL, currently there is no way to know if the transfer has failed because of no memory. Hence, I think it would be good to allocate the tx and rx buffers with spi initialize. Since 64KB * 4 (2 spi ports, 2 buffers tx, rx) = 256KB is a lot, may be we can use a smaller buffers of 4KB (16KB total usage). I think this will also increase the performance a bit as we skip the malloc calls in exchange function.

zhongnuo-tang: Hi @abhishek-samsung,
So where do we free this malloc, i don't see there is a deinit in spi operation.
Also, how much will the performance increase with skipping the malloc calls in exchange function? I think it will add to the complicity during exchange, as we need to consider which set of tx/rx buffer to use... 
I agree that using 64KB might take up alot of memory, but we also need to consider the fact there will be a slight gap between each of the slicing data (~130us), if we slice it to 4K there will be more gaps, so im afraid this might also affect performance. 
Please share your thoughts, thank you.

abhishek-samsung: Okay,
1. I think we should add some logs so that we get to know when SPI transfer failed due to malloc.
2. Okay, lets use 64KB then.

os/arch/arm/src/amebasmart/amebasmart_spi.c
zhongnuo-tang: Hi @sunghan-chang,
I have removed duplicate code as suggested, and rxbuffer is required like what Abhishek had explained,

os/arch/arm/src/amebasmart/amebasmart_spi.c
zhongnuo-tang: Hi @abhishek-samsung ,
I have added lldbg("rxbuff_aligned malloc failed\n") when malloc failed.

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -1157,54 +1157,64 @@ static void amebasmart_spi_exchange(FAR struct spi_dev_s *dev,
...
-		spi_flush_rx_fifo(&priv->spi_object);
-		spi_master_read_stream_dma(&priv->spi_object, (char *) rxbuff_aligned, (uint32_t)nwords * mode_16bit);
-		spi_dmarxwait(priv);
-		memcpy(rxbuffer, rxbuff_aligned, nwords * mode_16bit);
-		rtw_mfree(rxbuff_aligned, 0);
 	}
-
+	if (txbuffer) {
+		rtw_mfree(txbuff_aligned, 0);
+	}
```
sunghan-chang, line 1216: When txbuffer has a some pointer (not null), `txbuff_aligned` is allocated. Yes it's right. But anyway they are different variables. How about checking `txbuff_aligned` itself?
I mean
```
if (txbuff_aligned) {
   rtw_mfree(....);
}
```

sunghan-chang (approved):

sunghan-chang:
@abhishek-samsung Could you let us know test result for this?
@jeongarmy Once Abhisehk confirms, could you merge this?

abhishek-samsung:
> @abhishek-samsung Could you let us know test result for this?

No issues with the PR. Everything works fine.

ewoodev (approved):

jeongarmy (approved):
LGTM

PR #6155, opened 2024-04-23
os/binfmt, build/configs: add support for xip loadable apps

abhishek-samsung:
added an alternative binary loader for executing applications from flash.


build/configs/rtl8730e/scripts/xipelf/userspace_app2.ld
```diff
@@ -0,0 +1,153 @@
+/****************************************************************************
+ *
+ * Copyright 2019 Samsung Electronics All Rights Reserved.
+ *
```
r-prabu, line 4: minor copyright info update

build/configs/rtl8730e/scripts/xipelf/userspace_app1.ld
```diff
@@ -0,0 +1,157 @@
...
+ *
+ ****************************************************************************/
+
+MEMORY
+{
+    /* apps actual data should start after the crc and header, so adding 48bytes to the start 
+     * also, if userbinary signing is done, then we need to add another 32bytes to this, so need to find some generic way
+       Split this into memory_app1.ld and auto-generate this based on config?
+     */
+    uflash (rx)      : ORIGIN = 0x844B000 + 0x30, LENGTH = 0x8545000 - 0x844B030
```
r-prabu, line 25: Is the address is only change for multiple linker scripts. Can we have this information taken from config/makefile for generating linker scripts based on the configs which will be easy for the maintenance.?

os/binfmt/libxipelf/xipelf.c
```diff
@@ -0,0 +1,188 @@
...
+		if (g_bin_addr_list[bin_idx].text_addr != 0) {
+			lldbg("[%s] Text Addr : %p, Text Size : %u\n", BIN_NAME(bin_idx), g_bin_addr_list[bin_idx].text_addr, g_bin_addr_list[bin_idx].text_size);
+		}
+	}
+}
+
+void elf_save_bin_section_addr(struct binary_s *bin)
+{
+	if (bin != NULL) {
+		uint8_t bin_idx = bin->binary_idx;
```
r-prabu, line 155: can we check the array index?

os/include/tinyara/binary_manager.h
```diff
@@ -149,6 +149,9 @@ struct user_binary_header_s {
 	uint32_t bin_ramsize;
 	uint32_t bin_stacksize;
 	uint32_t kernel_ver;
+#ifdef CONFIG_XIP_ELF
```
r-prabu, line 152: why do you need padding?  If we strictly required word aligned memory for this struct then can we remove the  __attribute__((__packed__)); instead of manually adding padding ourself which might create issues if incase we add some more members in near future.

abhishek-samsung: The pc value should be 32 bit aligned, so if we add headers in front of the binary, it wont align without the current padding. I added some comments in mkbinheader.py file in the PR regarding this.

os/binfmt/libxipelf/xipelf.c
abhishek-samsung: Done

os/binfmt/libxipelf/xipelf.c
```diff
@@ -0,0 +1,188 @@
...
+
+bin_addr_info_t *get_bin_addr_list()
+{
+	return g_bin_addr_list;
+}
+
+void elf_delete_bin_section_addr(uint8_t bin_idx)
+{
+	/* Clear binary section address information */
+
```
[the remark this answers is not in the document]
abhishek-samsung: Done

build/configs/rtl8730e/scripts/xipelf/userspace_app1.ld
abhishek-samsung: Done

edwakuwaku:
Hi @abhishek-samsung , as per my understanding, there a few advantage of loadable_apps running on ram and kernel running on XIP
1) For prioritising application codes
2) Clear separation between kernel and application

Your PR seems to be placing app binary to run XIP, may I know the reason? Other than that, may I know which kind of app will apply such layout?

r-prabu (approved):

abhishek-samsung:
> Hi @abhishek-samsung , as per my understanding, there a few advantage of loadable_apps running on RAM and kernel running on XIP
> 
> 1. For prioritising application codes
> 2. Clear separation between kernel and application
> 
> Your PR seems to be placing app binary to run XIP, may I know the reason? Other than that, may I know which kind of app will apply such layout? Are we doing this because of some RAM shortage?

Yes, we are doing this because of RAM shortage, no other reason.

sunghan-chang (approved):

PR #6160, opened 2024-04-24
build/configs/rtl8730e, rtl8721csm_make_bin.sh: Fix binary signing build error

jylee9613:
Public does not support binary signing, but there was a build error when enable binary signing.
This commit fix binary signing build error.
It may not work properly if enable binary signing, but the binary will be built.


os/Kconfig
```diff
@@ -315,7 +315,8 @@ config BINARY_SIGNING
 
 config USER_SIGN_PREPEND_SIZE
 	int "Prepended user binary signing size"
-	default 32
+	default 32 if !ARCH_CHIP_AMEBASMART 
+	default 4096 if ARCH_CHIP_AMEBASMART
```
sunghan-chang, line 319: Is this configurable for user? If this is fixed for each board, how about hiding this and move this into board or chip Kconfig?

os/include/tinyara/signature.h
```diff
@@ -29,6 +29,9 @@
 #define USER_SIGN_PREPEND_SIZE     0
 #endif
 
+#define SIGNATURE_VERIFY_TRUE      0
+#define SIGNATURE_VERIFY_FALSE     1
```
sunghan-chang, line 33: 1. Need to apply this to amebad as well.
2. `VERIFY_TRUE` and `FALSE`? `SIGNATURE_VERIFIED` and `SIGNATURE_INVALID`?

jylee9613: Amebad returns 0(OK) when verify success, but amebasmart returns 1 when success. So I added SIGNATURE_VERIFY_TRUE / SIGNATURE_VERIFY_FALSE value only for amebasmart. 
This is already applied in our product code for now.

jylee9613: 2. I'll change the name like this
`SIGNATURE_VALID / SIGNATURE_INVALID`

sunghan-chang: > Amebad returns 0(OK) when verify success, but amebasmart returns 1 when success. So I added SIGNATURE_VERIFY_TRUE / SIGNATURE_VERIFY_FALSE value only for amebasmart.

It should not be. Every chip should provide the same returns. If not, will you modify binary manager for each board?

sunghan-chang (commented):
This PR has too many different topics.
1. change a permission for sh to 755
2. move CONFIG_USER_SIGN_PREPEND_SIZE into each board
3. add amebasmart_signature.c
4. make non-signing binary with CONFIG_BINARY_SIGNING
5. add new definitions for signing verification

os/arch/arm/src/amebad/amebad_signature.c
```diff
@@ -20,17 +20,18 @@
  * Included Files
  ****************************************************************************/
 #include <tinyara/config.h>
+#include <tinyara/signature.h>
 #include <sys/types.h>
```
sunghan-chang, line 24: Because we change OK to SIG....VALID, this is unnecessary. Let's remove this including.

os/arch/arm/src/amebalite/amebalite_signature.c
```diff
@@ -20,17 +20,18 @@
  * Included Files
  ****************************************************************************/
 #include <tinyara/config.h>
+#include <tinyara/signature.h>
 #include <sys/types.h>
```
sunghan-chang, line 24: Same as previous

os/arch/arm/src/amebad/amebad_signature.c
jylee9613: Erased

sunghan-chang (approved):

PR #6162, opened 2024-04-24
os/kernel/init/os_bringup.c: Change security_level log

jylee9613:
If security_level is not set yet, it shows error log. 
Because of the log "error", it can be mistaken for abnormal operation. 
Therefore, this commit change the security level log to prevent confusion.

### DIFF
--- a/os/kernel/init/os_bringup.c
+++ b/os/kernel/init/os_bringup.c
@@ -274,10 +274,7 @@ static inline void os_do_appstart(void)
 #endif
 
 #ifdef CONFIG_SECURITY_LEVEL
-	int ret = set_security_level();
-	if (ret != SECURITY_LEVEL_OK) {
-		dbg("[set_security_level] error ret : %d\n", ret);
-	}
+	set_security_level();
 #endif
 
 #ifdef CONFIG_IOTDEV
--- a/os/se/security_level/security_level.c
+++ b/os/se/security_level/security_level.c
@@ -92,26 +92,30 @@ int set_security_level(void)
 	output.data = (char *)kmm_malloc(output.data_len);
 
 	if (output.data == NULL) {
+		dbg("Alloc fail. Security level is high\n");
 		return SECURITY_LEVEL_ALLOC_FAILED;
 	}
 
 	hal_ret = se->ops->read_storage(SS_SLOT_INDEX_SECURITY_LEVEL, &output);
 	if (hal_ret != HAL_SUCCESS) {
 		kmm_free(output.data);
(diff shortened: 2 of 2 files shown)


os/kernel/init/os_bringup.c
```diff
@@ -276,7 +276,13 @@ static inline void os_do_appstart(void)
 #ifdef CONFIG_SECURITY_LEVEL
 	int ret = set_security_level();
 	if (ret != SECURITY_LEVEL_OK) {
-		dbg("[set_security_level] error ret : %d\n", ret);
+		if (ret == SECURITY_LEVEL_READ_FAILED) {
```
jsdosa, line 279: set_security_level() This function has a total of 4 return values.
1 success value, 3 failure values So, I think, why not print an error message for else as well? (ALLOC_FAILED)

I only offer opinions, you don't have to accept them.

sunghan-chang, line 279: 1. You print the success case in the function. But you add failure cases at out of the function. Is there any reason? If not, let's get the same policy for them.
2. Log messages occupy some flash memory. To optimize memory usage, let's remove duplicated words.

jylee9613: I added log for alloc failed. Thank you for your review.

jylee9613: 1. fixed to pring log inside of `set_security_level`
2. I erased duplicated words.

os/kernel/init/os_bringup.c
```diff
@@ -275,9 +275,6 @@ static inline void os_do_appstart(void)
 
 #ifdef CONFIG_SECURITY_LEVEL
 	int ret = set_security_level();
```
sunghan-chang, line 277: The `ret` looks unncessary. Let's remove.

os/se/security_level/security_level.c
```diff
@@ -108,10 +110,11 @@ int set_security_level(void)
 		security_level = 0;
 	} else {
 		kmm_free(output.data);
+		dbg("Wrong security_level value\n");
 		return SECURITY_LEVEL_INVALID_VALUE;
 	}
 
-	printf("[set_security_level] security_level : %d\n", security_level);
+	dbg("[set_security_level] security_level : %d\n", security_level);
```
sunghan-chang, line 117: Need `[set_security_level]`?

sunghan-chang (commented):
You added some failure messages in detail. But it's not our purpose of this.
for example,
Line 113 `security_level is not set`, so? what happened? `not set` is an issue?

PR #6163, opened 2024-04-24
os/drivers: update Kconfig for VIRTKEY_QUEUE_LEN

anjana348:
New config was added for Virtual key driver event queue length. It was not added in Kconfig, adding it after review.

### DIFF
--- a/os/drivers/Kconfig
+++ b/os/drivers/Kconfig
@@ -59,6 +59,15 @@ config VIRTKEY
 	default n
 	depends on NFILE_DESCRIPTORS != 0
 
+if VIRTKEY
+config VIRTKEY_QUEUE_LEN
+	int "Virtual key event queue length"
+	default 1024
+	---help---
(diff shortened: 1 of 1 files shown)


os/drivers/Kconfig
```diff
@@ -59,6 +59,13 @@ config VIRTKEY
 	default n
 	depends on NFILE_DESCRIPTORS != 0
 
+config VIRTKEY_QUEUE_LEN
+	int "Virtual key event queue length"
```
sunghan-chang, line 63: This should be enabled when VIRTKEY is enabled. Let's add a conditional.
```
config VIRTKEY
.....

if VIRTKEY
config VIRTKEY_QUEUE_LEN
....
endif # VIRTKEY
```

sunghan-chang (commented):
@anjana348 Could you remove double quotation marks at the commit title?

jsdosa (approved):
looks nice to me.

sunghan-chang (approved):

PR #6164, opened 2024-04-26
kernel/log_dump: Add for detecting device hang occurrence

ewoodev:
Currently, the log dump structure prevents logging output until compression is complete, which can cause device hangs.

Although exceptions have already been processed in several cases, there may be other exceptions. Also, when a hang occurs, the log is not printed, making it difficult to determine the cause.

Previous device hangs generated in the log dump occurred because a simple busy wait was performed with a while statement to wait for compression.
In other words, if you cannot get out of the while loop, a device hang may have occurred.

Therefore, if the while statement is not exited within a certain time, it calls PANIC to notify that a device hang has occurred and debugs by printing the assert log.


os/kernel/log_dump/log_dump.c
```diff
@@ -62,6 +63,10 @@
 
 #define LOG_CHUNK_SIZE			sizeof(struct log_dump_chunk_s)
 
+#define LOG_DUMP_HANG_CHECK_SEC		60	/* 60s */
```
sunghan-chang, line 66: We use 3 mins for watchdog. How about 2 and half mins?
I suggest this because this is for debugging.

And no need to cover them with debug new feature?

ewoodev: I have confirmed the case where an assert occurs because it exceeds the post max of another IRQ even though it has been less than two minutes since the hang occurred.

In this case, it is also possible to debug with assert, but debugging may be difficult if a hang occurs because the compression thread in log_dump is not completed.

How about reducing the time to one minute and thirty seconds for this?

If compression is performed normally, the priority of the compression thread is high, so the actual counting period will delay. This count only works normally when the compression thread stops.

sunghan-chang: Ok, Anyway let's leave some explanation of those values.

os/kernel/log_dump/log_dump.c
```diff
@@ -62,6 +63,10 @@
...
+#define LOG_DUMP_COMP_CHECK_RETRY_NSEC		10000	/* 10ms */
```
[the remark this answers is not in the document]
ewoodev: oh.. Thank you. fixed

os/kernel/log_dump/log_dump.c
```diff
@@ -62,6 +63,10 @@
...
+#define LOG_DUMP_COMP_CHECK_RETRY_CNT (LOG_DUMP_HANG_CHECK_SEC * 1000000 / LOG_DUMP_COMP_CHECK_RETRY_NSEC)
```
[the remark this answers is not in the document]
ewoodev: Done~!

os/kernel/log_dump/log_dump.c
```diff
@@ -185,8 +190,15 @@ int log_dump_read_wake(void)
...
 	/* wait for the completion of the partially filled block compression */
+	int hang_check_count = 0;
 	while (compress_last_block) {
-		usleep(10000);
+		hang_check_count++;
+		usleep(LOG_DUMP_COMP_CHECK_RETRY_NSEC);	
+
+		if (hang_check_count > LOG_DUMP_COMP_CHECK_RETRY_CNT) {
+			lldbg("Compression failed for %d seconds, The device hang is suspected.\n", LOG_DUMP_HANG_CHECK_SEC);
+			PANIC();
```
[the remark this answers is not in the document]
ewoodev: Changed to DEBUGPANIC()

os/kernel/log_dump/log_dump.c
```diff
@@ -180,13 +188,23 @@ int log_dump_read_wake(void)
...
 		return ERROR;
 	}
 
 	/* wait for the completion of the partially filled block compression */
+#ifdef CONFIG_LOG_DUMP_DEBUG_DETECT_HANG
+	int hang_check_count = 0;
+#endif
 	while (compress_last_block) {
-		usleep(10000);
+		usleep(LOG_DUMP_COMP_CHECK_RETRY_USEC);
```
sunghan-chang, line 200: This should be in the conditional. Because the definition is defined in the conditional. I think you dont build test without the config enabled.

ewoodev: oh right, I fixed and test without config enable.

sunghan-chang (approved):

PR #6068, opened 2024-02-19
apps/platform, os/binfmt: add exceptions support in loadable

abhishek-samsung:
Add support for cpp exceptions in loadable for RTL8730E and RTL8721CSM.

PR #6165, opened 2024-04-26
Enable audio configs on loadable_ext of RTL8730E

abhishek-samsung:


abhishek-samsung:
None of the test code for media is enabled in the current PR as it will result in build failure. The build issues can be resolved by applying exceptions PR https://github.com/Samsung/TizenRT/pull/6068 . However, cout is still not working, so it must be avoided for now. I have tested media player (adding code without cout) after applying https://github.com/Samsung/TizenRT/pull/6068 and its working fine.

jsdosa (approved):
looks good to me.

jsdosa (changes requested):
@abhishek-samsung Would you like to check for conflicts?

abhishek-samsung:
> @abhishek-samsung Would you like to check for conflicts?

Done

framework/src/media/Kconfig
```diff
@@ -8,7 +8,7 @@ config MEDIA
 	default n
 	depends on AUDIO
 	select HAVE_CXX
-	select HAVE_CXXINITIALIZE
+	select HAVE_CXXINITIALIZE if BUILD_FLAT
```
sunghan-chang, line 11: How about loadable xip? Isn't it the same as flat?

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -17,6 +17,7 @@ config ARCH_CHIP_RTL8730E
 	select ARMV7A_HAVE_GICv2
 	select ARCH_HAVE_FPU
 	select AMEBASMART_WIFI
+	select ARCH_BOARD_HAVE_SECOND_FLASH
```
sunghan-chang, line 20: It should be valid with HW Rev 5. Is it possible to give any conditional on this? @ewoodev Could you check this together?

framework/src/media/Kconfig
abhishek-samsung: In the case of flat, HAVE_CXXINITIALIZE does static constructors initialization. In the case of loadable static constructors are initialized in binfmt code for both xip loadable and elf loadable and is not dependent on HAVE_CXXINITIALIZE.

sunghan-chang: Oh ok, Loadable XIP does with BINFMT_CONSTRUCTORS as same as loadable ELF.
Then, don't you need a selection for that?
```
select BINFMT_CONSTRUCTORS if (ELF || XIP_ELF)
```

abhishek-samsung: https://github.com/Samsung/TizenRT/blob/cf41d8ad8333a95dc375331578c7d5920d4acc9d/lib/libxx/Kconfig#L51-L57

They are selected be default when we enable c++, so we need not add it here.

sunghan-chang: Then, why do we keep this `select HAVE_CXXINITIALIZE if BUILD_FLAT` here?

abhishek-samsung: Agreed, removed that line.

framework/src/media/Kconfig
```diff
@@ -8,7 +8,6 @@ config MEDIA
 	default n
 	depends on AUDIO
 	select HAVE_CXX
-	select HAVE_CXXINITIALIZE
```
sunghan-chang, line 11: Could you leave where this is concerned in a Kconfig in the commit description?

os/arch/arm/src/amebasmart/Kconfig
ewoodev: https://github.com/Samsung/TizenRT/blob/cf41d8ad8333a95dc375331578c7d5920d4acc9d/os/board/rtl8730e/Kconfig#L39-L44

Please add condition if RTL8730E_BOARD_REVISION > 5

framework/src/media/Kconfig
abhishek-samsung: Sorry, I didnt understand. Do I need to update the commit description on why I removed this line?

sunghan-chang: Because this `select` is done in other config, we remove this.
I want to leave
1. why this config should be enabled
2. who enables this automatically instead of here

abhishek-samsung: I have updated the commit description.

PR #6166, opened 2024-04-30
os/board/rtl8730e: dimming support for lcd st7789 using pwm

anjana348:
pwm duty cycle is being changed to change brightness of lcd. It can go from 0 to 100 percent.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_st7789.c
+++ b/os/board/rtl8730e/src/rtl8730e_st7789.c
@@ -25,6 +25,7 @@
 #include "objects.h"
 #include "PinNames.h"
 #include "gpio_api.h"
+#include "pwmout_api.h"
 
 #ifndef CONFIG_LCD_ST7789_SPI_PORT
 #define CONFIG_LCD_ST7789_SPI_PORT 1
@@ -34,14 +35,26 @@
 #define GPIO_PIN_BACKLIGHT 	PA_9
 #define GPIO_PIN_CMDDATA 	PA_10
 
+/****************************************************************************
+ * Private Types
+ ****************************************************************************/
+
+struct rtl8730e_st7789_info_s {
+        struct st7789_config_s lcd_config;
+        pwmout_t pwm_led;
+};
+
+
 static int rtl8730_st7789_gpio_reset(void);
 static int rtl8730_st7789_gpio_cmddata(uint8_t cmd);
-static int rtl8730_st7789_gpio_backlight(uint8_t level);
+static int rtl8730_st7789_control_backlight(uint8_t level);
 
-struct st7789_config_s g_rtl8730_config_dev_s = {
-	.reset = rtl8730_st7789_gpio_reset,
-	.cmddata = rtl8730_st7789_gpio_cmddata,
-	.backlight = rtl8730_st7789_gpio_backlight,
+struct rtl8730e_st7789_info_s g_rtl8730_config_dev_s = {
+	.lcd_config = {
+		.reset = rtl8730_st7789_gpio_reset,
+		.cmddata = rtl8730_st7789_gpio_cmddata,
+		.backlight = rtl8730_st7789_control_backlight,
+	},
 };
 
 static int rtl8730_st7789_gpio_reset(void)
@@ -61,16 +74,16 @@ static int rtl8730_st7789_gpio_cmddata(uint8_t cmd)	// cmd - 0, data - 1
 	return OK;
 }
 
-static int rtl8730_st7789_gpio_backlight(uint8_t level)	//on - 1, off - 0
+static int rtl8730_st7789_control_backlight(uint8_t level)	// range  0~100 percentage of brightness level
 {
-	GPIO_WriteBit(GPIO_PIN_BACKLIGHT, level);
+	float pwm_level = level/100.0;
+	pwmout_write(&g_rtl8730_config_dev_s.pwm_led, pwm_level);
 	return OK;
 }
 
 static void gpio_pins_init(void)
 {
 	GPIO_InitTypeDef ResetPin;
-	GPIO_InitTypeDef RSPin;
 	GPIO_InitTypeDef CmdPin;
 
 	Pinmux_Swdoff();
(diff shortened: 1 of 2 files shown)


jsdosa (approved):
looks nice to me.

os/board/rtl8730e/src/rtl8730e_st7789.c
```diff
@@ -63,7 +77,8 @@ static int rtl8730_st7789_gpio_cmddata(uint8_t cmd)	// cmd - 0, data - 1
 
 static int rtl8730_st7789_gpio_backlight(uint8_t level)	//on - 1, off - 0
```
ewoodev, line 78: Could you check this description?

ewoodev (approved):
I checked operation dimming

os/board/rtl8730e/src/rtl8730e_st7789.c
```diff
@@ -34,14 +35,27 @@
...
-	.reset = rtl8730_st7789_gpio_reset,
-	.cmddata = rtl8730_st7789_gpio_cmddata,
-	.backlight = rtl8730_st7789_gpio_backlight,
+struct rtl8730e_st7789_info_s g_rtl8730_config_dev_s = {
+	.lcd_config = {
+		.reset = rtl8730_st7789_gpio_reset,
+		.cmddata = rtl8730_st7789_gpio_cmddata,
+		.backlight = rtl8730_st7789_gpio_backlight,
+	},
+	.pwm_led.pwm_idx = 1,
```
sunghan-chang, line 58: What does it mean `1`? If possible, how about changing it to definition? or let's remove a meaning of that.

os/board/rtl8730e/src/rtl8730e_st7789.c
```diff
@@ -61,9 +75,10 @@ static int rtl8730_st7789_gpio_cmddata(uint8_t cmd)	// cmd - 0, data - 1
 	return OK;
 }
 
-static int rtl8730_st7789_gpio_backlight(uint8_t level)	//on - 1, off - 0
+static int rtl8730_st7789_gpio_backlight(uint8_t level)	// range  0~100 percentage of brightness level
```
sunghan-chang, line 78: How about renaming the function? `rtl8730e_st7789_control_backlight`?

os/board/rtl8730e/src/rtl8730e_st7789.c
sunghan-chang: This was removed. Is it unnecessary?

anjana348: Actually if we don't set any id, it by default take id as 0 in pwm_init function

anjana348: Yes, it was unnecessary because in pwmout_init, idx value is set already using defined structure in which it is 1.
board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/pwmout_api.c 
const u32 pin2chan[20][2] = {
        {PA_7,  0},
        {PA_9,  1},
        {PA_10, 2},
        {PA_11, 3},
        {PA_12, 4},
        {PA_13, 5},
        {PA_14, 0},
        {PA_8,  2},
        {PA_4,  3},
        {PA_5,  4},
        {PA_6,  5},
        {PB_14, 0},
        {PB_15, 1},
        {PB_16, 2},
        {PB_21, 3},
        {PB_22, 4},
        {PB_25, 0},
        {PB_26, 1},
        {PB_27, 2},
        {PB_30, 3}
};

os/drivers/lcd/st7789.c
```diff
@@ -907,7 +908,7 @@ FAR struct lcd_dev_s *st7789_lcdinitialize(FAR struct spi_dev_s *spi, struct st7
 	priv->spi = spi;
 	priv->config = config;
 
-	priv->config->backlight(1);
+	priv->config->backlight(100);
```
sunghan-chang, line 911: not CONFIG_LCD_MAXPOWER?
To use here, didn't you define that?

os/board/rtl8730e/src/rtl8730e_st7789.c
```diff
@@ -92,14 +106,18 @@ static void gpio_pins_init(void)
 	CmdPin.GPIO_PuPd = GPIO_PuPd_NOPULL;
 	CmdPin.GPIO_Mode = GPIO_Mode_OUT;
 	GPIO_Init(&CmdPin);
+
+	pwmout_init(&g_rtl8730_config_dev_s.pwm_led, GPIO_PIN_BACKLIGHT);
+
+	GPIO_WriteBit(GPIO_PIN_BACKLIGHT, 1);
```
sunghan-chang, line 112: That pin is inited as pwm. Is it ok to call GPIO API after that?
@edwakuwaku Is it ok in BSP?

sunghan-chang: What does it mean `1`? Backlight on?

edwakuwaku: Theoretically, `pmwout_init` and `pwmout_write` should be enough.
I think you can remove these few lines:
```
	Pinmux_Config(GPIO_PIN_BACKLIGHT, PINMUX_FUNCTION_GPIO);
	RSPin.GPIO_Pin = GPIO_PIN_BACKLIGHT;
	RSPin.GPIO_PuPd = GPIO_PuPd_NOPULL;
	RSPin.GPIO_Mode = GPIO_Mode_OUT;
	GPIO_Init(&RSPin);
```
and also remove 
`GPIO_WriteBit(GPIO_PIN_BACKLIGHT, 1);`

In `pwmout_init`, it will invoke `Pinmux_Config `to overwrite the function assigned to the pin.
Which means `Pinmux_Config(GPIO_PIN_BACKLIGHT, PINMUX_FUNCTION_GPIO);` will not be useful.

Could you try again with removing the few lines as stated above? It should work fine for PWM without invoking GPIO API.

The 1 in `GPIO_WriteBit `is to pull the pin state to high.

anjana348: yes , it works without gpio. I have updated PR

edwakuwaku (approved):
LGTM

sunghan-chang (approved):

PR #6167, opened 2024-04-30
arm/src/amebasmart: fix assert issue cause by wrong errno usage

abhishek-samsung:
errno should be only checked if sem_wait fails, but here we are calling ASSERT by checking the errno even in the success case. Hence, modified the loop logic to check errno in case of failure.


Taejun-Kwon (approved):

sunghan-chang:
@zhongnuo-tang @jnd91 This is the same as amebad. Please make changes there as well.

zhongnuo-tang:
> @zhongnuo-tang @jnd91 This is the same in the ameba d. Please make changes there as well.

Hi @sunghan-chang, I have made the changes for amebad in PR6168, please check.

PR #6171, opened 2024-05-03
apps/shell/tash_command: Add pm suspend while executing SYNC cmd

ewoodev:
We check the information or status of the program by using the TASH SYNC command. TASH is only used for debugging and testing, so it does not affect power consumption in actual operation. Therefore, for proper debugging and testing, we apply pm suspend to the TASH SYNC command to ensure operation.

### DIFF
--- a/apps/shell/tash_command.c
+++ b/apps/shell/tash_command.c
@@ -36,6 +36,12 @@
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
 #include <tinyara/reboot_reason.h>
 #endif
+#ifdef CONFIG_PM
+#include <tinyara/pm/pm.h>
+#include <fcntl.h>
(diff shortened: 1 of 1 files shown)


ritesh55555:
looks good

apps/shell/tash_command.c
```diff
@@ -538,6 +544,10 @@ int tash_execute_cmd(char **args, int argc)
 {
 	int cmd_idx;
 	int cmd_found = 0;
+#ifdef CONFIG_PM
+	int fd;
+	int ret;
```
sunghan-chang, line 549: Let's use meaningful names. I suggest belows
fd -> pmdrv_fd
ret -> pm_suspended

sunghan-chang (approved):

sunghan-chang (approved):

PR #6174, opened 2024-05-07
build/tools/amebasmart: Fix PG wakeup hang issue

edwakuwaku:
- During wakeup from PG sleep, PSRAM need to be reinitialized to prevent from hang


ritesh55555:
> @edwakuwaku Could you explain what this is? Is there a hang at wakeup? How is going with log uart wakeup?

Hi , I tested it for log uart wakeup , it is working fine . And PM is also working fine after appyling this PR

sunghan-chang (approved):

edwakuwaku:
Hi @sunghan-chang , during wakeup flow, the KM4 is not being waken up as we thought, due to PSRAM is not initialized correctly, and without KM4 being awake, CA32 cannot be powered up.
This situation causes all wakeup src to fail waking up as the problem is happening on the wakeup flow, and loguart is just happen to be one of those.

PR #6176, opened 2024-05-07
build/tools/amebasmart: Revert bootloader toolchain compiler flag

edwakuwaku:
- Revert bootloader toolchain compiler flag from -Os to -O2
- Reason: -Os is causing hotfix verfication failure


sunghan-chang:
@edwakuwaku This is the perfectly same one you gave before. Right? I mean, do we need to verify this again?

PR #6178, opened 2024-05-08
os/include/tinyara/security_hal.h: Add AES-GCM mode

jylee9613:
Add AES-GCM mode and parameter to support AES-GCM encryption/decryption. AES-GCM mode needs 4 additional parameter.
- AAD (Additional Authentication Data)
- AAD length
- TAG (Authentication Tag)
- TAG length

### DIFF
--- a/os/include/tinyara/security_hal.h
+++ b/os/include/tinyara/security_hal.h
@@ -149,6 +149,11 @@ typedef enum {
 	HAL_KEY_UNKNOWN,
 } hal_key_type;
 
+typedef enum {
+	HAL_GCM_AES,
+	HAL_GCM_UNKNOWN,
+} hal_gcm_type;
+
 typedef struct _hal_init_param {
 	uint32_t i2c_port;
 	uint32_t gpio;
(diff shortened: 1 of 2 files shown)


os/se/ameba/Kconfig
```diff
@@ -44,6 +44,13 @@ config HW_AES_ENC
 	---help---
 		Encrypts a data based on hardware.
 
+config HW_AES_GCM_ENC
+	bool "HW AES GCM encryption decryption"
```
sunghan-chang, line 48: The config name is `ENC` but explanation says `encryption descryption`. What is correct?

os/include/tinyara/security_hal.h
```diff
@@ -103,6 +103,9 @@ typedef enum {
 	HAL_AES_CBC_PKCS7,
 	HAL_AES_CTR,
 	HAL_AES_CFB128,
+#ifdef CONFIG_HW_AES_GCM_ENC
+	HAL_AES_GCM,
+#endif
```
sunghan-chang, line 108: How about adding that without the conditional?

os/se/ameba/Kconfig
jylee9613: GCM support both enc and dec. But other config(AES, RSA) name has only ENC, so I used enc only in the same way.
If I need to change this, I'll change all config name in the other commit.

sunghan-chang (approved):

PR #6179, opened 2024-05-08
tools/trap: add trap tool support for xip loadable apps 

neel-samsung:
add trap tool support for xip loadable apps

### DIFF
--- a/tools/debug/debugsymbolviewer.py
+++ b/tools/debug/debugsymbolviewer.py
@@ -22,7 +22,6 @@ from array import *
 import os
 import sys
 
-bin_path = '../../build/output/bin/'	# Output bin path
 ksymbol_lookup_table = []		# Lookup table created from contents of system map file for kernel symbols
 asymbol_lookup_table = [[]] * 10	# Lookup table created from contents of appX map files for application symbols
 debug = False				# Variable to print debug logs
@@ -220,6 +219,7 @@ def is_kernel_text_address(address):
 
 # Function to Parse the input log file (which contains stackdump during assert) and to print stack values
 def print_stack(log_file):
+	global xip_enabled
 	stack_addr = 0x00000000
 	stack_val = 0x00000000
 	partition_str = "==========================================================="
@@ -274,7 +274,8 @@ def print_stack(log_file):
 					is_app_symbol = is_app_text_address(hex(stack_val))	# app index in case of application symbol
 					if (is_app_symbol):
 						#If yes, print it's corresponding symbol
-						stack_val = stack_val - int(g_stext_app[is_app_symbol - 1])
+						if not xip_enabled:
+							stack_val = stack_val - int(g_stext_app[is_app_symbol - 1])
 						print_symbol(stack_addr, stack_val, is_app_symbol)
 
 
@@ -282,6 +283,7 @@ def print_stack(log_file):
 def print_wrong_sp(log_file):
 	stack_addr = 0x00000000
 	format_print = True
+	global xip_enabled
 	# Parse the contents based on tokens in log file.
 	with open(log_file) as searchfile:
 		for line in searchfile:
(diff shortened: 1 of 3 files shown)


tools/debug/debugsymbolviewer.py
```diff
@@ -28,6 +28,7 @@
 debug = False				# Variable to print debug logs
 pc_value = 0x00000000			# Program counter value
 format_print = True
+CONFIG_PATH = '../../os/.config'
```
r-prabu, line 31: instead of hardcode path can we have CONFIG_PATH like which is used in the ramdumpparser.py?

neel-samsung: we are also setting ```CONFIG_PATH``` in ramdumpParser.py at start.
https://github.com/Samsung/TizenRT/blob/849427eb66d1800ad584c0e57556edec8d1ad6aa/tools/trap/ramdumpParser.py#L56
Since, we need it here, then we need to define here also.

sunghan-chang: As far as I know, the tool provides selectable options for this. 
1. use default values
2. get user input for this

Is it the same? We should concern this can be executed without source code.

sunghan-chang (commented):
Please leave TRAP result with any assert in the commit description.

tools/debug/debugsymbolviewer.py
neel-samsung: Yes, That is it. Apart from ```CONFIG_PATH```, there was one other variable ```bin_path```, which was hard-coded in debugsymbolviewer.py. I have changed it and passed both variable from ramdumpParser.py as argument.

sunghan-chang (commented):
@neel-samsung Could you leave non-loadable xip result with this PR in the commit description? loadable elf and flat.

tools/debug/debugsymbolviewer.py
sunghan-chang: @neel-samsung I think you modified it to get as an new argument always. But that's not what I said.
If user inputs the path with an argument, then the script uses given path. If user does not give, then the script should use default path.

sunghan-chang: @neel-samsung That's for user input. If the script is not for user, no need to do that like above.

sunghan-chang: @neel-samsung Regardless of the comments above, what I think is that
we should guarantee the trap working without code build. In other words, when we get assert log and elf only, we want to run the trap without code build and on custom path (elf is not in the build/output/bin). That's why I asked you to support user input.

neel-samsung: @sunghan-chang The change I have made is in ```debugsymbolviewer.py``` and I thought this file will execute only when ```ramdumpParser.py``` will call internally. If that's not the case, we need to handle it. But currently, trap will run both the ways, like when user has passed the path as argument, it will take that or else default one. I haven't altered it in ```ramdumpParser.py```. 
Do we want ```debugsymbolviewer.py``` to run individually? .. Then, we need to handle it. 
https://github.com/Samsung/TizenRT/blob/c4d2e718f0cb1991d1a6393a811f2b852b9cb0d6/tools/trap/ramdumpParser.py#L55
https://github.com/Samsung/TizenRT/blob/c4d2e718f0cb1991d1a6393a811f2b852b9cb0d6/tools/trap/ramdumpParser.py#L56

neel-samsung:
> @neel-samsung Could you leave non-loadable xip result with this PR in the commit description? loadable elf and flat.

@sunghan-chang I have added required TRAP outputs to commit desc. Please check.

tools/debug/debugsymbolviewer.py
sunghan-chang: @neel-samsung Is there any usecase to run it individually? I think TRAP script is enough for user. If you think no usecase, we can keep current. No need to handle both cases (user input, default path). Please let me know your opinion.

neel-samsung: I also think TRAP script is enough for the user. We can keep the current change.

neel-samsung:
@sunghan-chang Can you merge this PR?

neel-samsung:
@sunghan-chang can you review it once? It looks ready.

sunghan-chang (approved):

PR #6181, opened 2024-05-10
os/arch/arm/src/amebasmart: I2C/I2S integration with PM

edwakuwaku:


os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1226,3 +1249,149 @@ int up_i2creset(FAR struct i2c_dev_s *dev)
...
+ *
+ ****************************************************************************/
+#ifdef CONFIG_PM
+static void amebasmart_i2c_pmnotify(FAR struct pm_callback_s *cb, int domain,
+                                   enum pm_state_e pmstate)
+{
+	switch (pmstate)
+	{
+		case PM_NORMAL:
+			pmu_unregister_sleep_callback(PMU_I2C_DEVICE);
```
sunghan-chang, line 1321: This means the pm state should be normal at wakeup? I heard it could be the standby when LCD is used. For example, lcd should not be turned on but something should be working (main core should be wakeup without lcd on). In this case, is this right? @ritesh55555 please check this.

ritesh55555: 1. Yes , for now , whenever board wakes up with an interrupt, pm state goes to NORMAL .
2. When LCD is  used, according the design given (lcd and pm) , pm should go to stand by state . So if we do that , we need to change this part and also in some other files also.

ewoodev: PM_NORMAL can be set in several situations.

This may not be the case when the main core wakes up.

I think it's a better way to restore this function directly from the pmu code in BSP than this.

The noti callback of this PM is not a feature for the BSP stage.

edwakuwaku: Hi all, I have pushed a commit to move the pre/post sleep callbacks to being done only during initialization stage.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1809,3 +1836,183 @@ struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port)
...
+}
+#endif
+
+
+#ifdef CONFIG_PM
+void i2s_pminitialize(void)
+{
+	int ret;
+	/* Register to receive power management callbacks */
+	ret = pm_register(&g_i2spm.pm_cb);
```
ewoodev, line 2013: If you do not use this, Could you please remove it?
This can cause unnecessary memory and sleep-wakeup time to increase.

edwakuwaku: Yes, I understand your concern~ Removed accordingly

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1809,3 +1812,86 @@ struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port)
...
+#if defined(I2S_HAVE_RX) && (0 < I2S_HAVE_RX)
+	priv->rx_isr_handler = (i2s_irq_handler)&i2s_transfer_rx_handleirq;
+#endif
+
+#if defined(I2S_HAVE_TX) && (0 < I2S_HAVE_TX)
+	priv->tx_isr_handler = (i2s_irq_handler)&i2s_transfer_tx_handleirq;
+#endif
+
+	ret = amebasmart_i2s_isr_initialize(priv);
+	if (ret != OK) {
```
ritesh55555, line 1858: ret is not defined in this function, can you please update?

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1809,3 +1812,86 @@ struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port)
...
+		i2serr("I2S initialize: isr fails\n");
+		goto errout_with_alloc;
+	}
```
ritesh55555, line 1861: errout_with_alloc is not defined in this function

os/arch/arm/src/amebasmart/amebasmart_i2s.c
edwakuwaku: sorry for mistake, it has been fixed.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
edwakuwaku: Removed.

kishore-sn (approved):

kishore-sn:
hi @edwakuwaku 
If we enable, CONFIG_AMEBASMART_I2C2, then the up_i2cinitialize is called two times (duplicate) during boot up. Is this ok?
os/board/rtl8730e/src/rtl8730e_boot.c:	                i2c = up_i2cinitialize(bus);
os/board/rtl8730e/src/rtl8730e_alc1019.c:		i2c = up_i2cinitialize(ALC1019_I2C_PORT);

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1236,3 +1240,50 @@ int up_i2creset(FAR struct i2c_dev_s *dev)
...
+{
+	(void)expected_idle_time;
+	(void)param;
+	/* For PG Sleep, I2C 1/2 HW will be lost power, thus a reinitialization is required here 
+	I2C0 is under SYSON block, so it will not require to be reinitialized*/
+#ifdef CONFIG_AMEBASMART_I2C1
+	(void) up_i2cinitialize(1);
+#endif
+#ifdef CONFIG_AMEBASMART_I2C2
+	(void) up_i2cinitialize(2);
```
kishore-sn, line 1278: hi @edwakuwaku
If we enable, CONFIG_AMEBASMART_I2C2, then the up_i2cinitialize is called two times (duplicate) during boot up. Is this ok?
os/board/rtl8730e/src/rtl8730e_boot.c: i2c = up_i2cinitialize(bus);
os/board/rtl8730e/src/rtl8730e_alc1019.c: i2c = up_i2cinitialize(ALC1019_I2C_PORT);

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1163,6 +1163,10 @@ FAR struct i2c_dev_s *up_i2cinitialize(int port)
 		return NULL;
 	}
 
+	/* If I2C has been initialized, return directly */
```
edwakuwaku, line 1166: Hi @kishore-sn , if there is multiple attempt for initialisation, the condition here will return directly.

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1312,40 +1296,25 @@ static void amebasmart_serial_pmnotify(FAR struct pm_callback_s *cb, int domain,
...
-			 * UART1_DEV: hp uart1
-			 * UART2_DEV: hp uart2
-			 * UART3_DEV: hp uart3_bt
-			 * LOGUART_DEV: KM0 log uart */
+	if (pmstate == PM_SLEEP) {
 #ifdef CONFIG_RTL8730E_UART0
-			if ((g_uart0priv.txint_enable) && (!serial_writable(sdrv[0]))) {		/* If Tx init enable and FIFO not empty */
-					return ERROR;
-			}
+		if ((g_uart0priv.txint_enable) && (!serial_writable(sdrv[0]))) {		/* If Tx init enable and FIFO not empty */
```
ewoodev, line 1301: amebasmart_serial_pmprepare
amebasmart_spi_pmprepare

Please make changes so that serial and spi do not use the pm_register function, just like i2c and i2s.

You can raise new PR about it.

ewoodev (approved):
LGTM

sunghan-chang (approved):

os/arch/arm/src/amebasmart/amebasmart_serial.c
edwakuwaku: @ewoodev , but these checks are from driver as a last protection, to prevent from system going to sleep, if some app is still running with utilizing those drivers.
I2C and I2S are for pre/post sleep process, both are different.

PR #6182, opened 2024-05-13
os/se/security_level: Set security level high in default

jylee9613:
Set security level high in default (in case of failing to set security level or security level is not setting yet)


sunghan-chang (commented):
You should set it at line 117 as well.

os/se/security_level/security_level.c
```diff
@@ -51,7 +51,8 @@
 /****************************************************************************
  * Private data
  ****************************************************************************/
-static int security_level = 0;
+/* Default security_level value: high */
```
sunghan-chang, line 54: Security level value high means?

sunghan-chang (approved):

PR #6183, opened 2024-05-13
os/board/rtl8730e: Adding the missing PWM pins channel

ziyik:


sunghan-chang (commented):
1. It has 22 ports, but what I have is 14.
PA 4, 5, 9, 10, 13, 14
PB 10, 11, 21, 22, 25, 26, 27, 30
Can we use 22 all?

2. We don't allow no description in a commit. Could you write any explanation of this?
 I think RTL8730E's PWM port list and this variable explanation is fine.

ziyik:
> 1. It has 22 ports, but what I have is 14.
>    PA 4, 5, 9, 10, 13, 14
>    PB 10, 11, 21, 22, 25, 26, 27, 30
>    Can we use 22 all?

But you do not have physical pins or ports. There is no way for you to output the PWM.

sunghan-chang:
@ziyik Sorry but I think we mis-communicate.
What I said is that
1. as far as I know, RTL8730E has only 14 available PWM pins. 
2. but your table (pin2chan) has 22 pins.

Why does it have 22 pins, not 14 pins?
It's my question.

Dane-Kang:
> @ziyik Sorry but I think we mis-communicate. What I said is that
> 
> 1. as far as I know, RTL8730E has only 14 available PWM pins.
> 2. but your table (pin2chan) has 22 pins.
> 
> Why does it have 22 pins, not 14 pins? It's my question.

Hi @sunghan-chang ,
this table is set to support all of our Amebasmart packages (100 pins, 144 pins)
As you can see below our pin descriptions, RTL8730EAH is set 100 pins and there are not existed PA6,PA7, PA8 but have from RTL8730ELH which made to use 144 pins.
image

So, even pwm table has 22, we can use 14 PWM pins due to RTL8730EAH packages which has 100pins.

sunghan-chang (approved):

PR #6187, opened 2024-05-16
os/arch/arm/src/amebasmart: Improve power efficiency for RTL8730E

edwakuwaku:
- In idle thread, apply "WFE" for cpu to enter low power state, under all PM states other than PM_SLEEP, as there is nothing to be done for idle thread under those PM states.
- If a state is locked (ie. state transition will not be happening), but we might still enter idle loop under some circumstances, in this case, WFE should also be applied as well, because there will be no system variable changes happening from idle thread

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_idle.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_idle.c
@@ -77,12 +77,23 @@ static void up_idlepm(void)
 			oldstate = newstate;
 		}
 		/* MCU-specific power management logic */
+		/* TODO: When LCD is integrated, some of the state operation might require revision
+		   PM_IDLE: We will expect LCD to go through background light dimming
+		   PM_STANDBY: LCD shut down, single core might be able to handle remaining tasks
+		   			   Then, we can do vPortSecondaryOff() for SMP case (ie. shutdown secondary core)
+		*/
 		switch (newstate) {
 			case PM_NORMAL:
+				/* In PM_NORMAL, we have nothing to do, set core to WFE */
+				__asm(" WFE");
 				break;
 			case PM_IDLE:
+				/* In PM_IDLE, we have nothing to do, set core to WFE */
+				__asm(" WFE");
 				break;
 			case PM_STANDBY:
+				/* In PM_STANDBY, we have nothing to do, set core to WFE */
+				__asm(" WFE");
 				break;
 			case PM_SLEEP:
 				/* TODO: When enabling SMP, PM state coherency should be verified for 
@@ -193,6 +204,14 @@ REJECTED:
 			default:
 				break;
 		}
+	} else {
+		/* If a state is locked, we will not be able to do state transition
+		But idle thread might still have chance to be doing judgement under some
+		conditions (ie. context switched to idle loop), thus if there is no
+		state transition happening, we invoke WFE for the core to rest (ie. a kind of HW sleep),
+		to prevent from unecessary power consumption
+		*/
+		__asm(" WFE");
 	}
 }
 #else


sunghan-chang:
@ritesh55555 Could you share our test result with this PR?

ewoodev (commented):
who is wakeup core?

edwakuwaku:
In my understanding, the WFE is a cpu sleep, and pending for any event to happen, which means any task with higher priority will be able to trigger the core to handle some operation.
This WFE is different from the power save (ie. Power gating mode), although we call both of them as sleep.
But PG sleep is shutting down the core, whereas WFE is just a kind of rest (?) for the cpu.

ritesh55555:
> In my understanding, the WFE is a cpu sleep, and pending for any event to happen, which means any task with higher priority will be able to trigger the core to handle some operation (ie. context switch). This WFE is different from the power save (ie. Power gating mode), although we call both of them as sleep. But PG sleep is shutting down the core (ie. power lost), whereas WFE is just a kind of rest (?) for the cpu.

hi @edwakuwaku , i have a doubt . Sometimes state changes happen very fast like within milliseconds . In that case , if we make the core to WFE mode and wake up again quickly , will it save power or it will have a negative impact . 
Is this scenario tested?

edwakuwaku:
> > In my understanding, the WFE is a cpu sleep, and pending for any event to happen, which means any task with higher priority will be able to trigger the core to handle some operation (ie. context switch). This WFE is different from the power save (ie. Power gating mode), although we call both of them as sleep. But PG sleep is shutting down the core (ie. power lost), whereas WFE is just a kind of rest (?) for the cpu.
> 
> hi @edwakuwaku , i have a doubt . Sometimes state changes happen very fast like within milliseconds . In that case , if we make the core to WFE mode and wake up again quickly , will it save power or it will have a negative impact . Is this scenario tested?

I will not say the core really went to sleep and will be waken up like what we did in the power save (ie. PG sleep case).
The core is actually still actively pending for event to trigger for doing operation.
If we simplify the routine, you may think it as:
while(1) {// do nothing, the cpu might treat it as an "NOP", thus execution higher power consumption}
while(1) {"WFE", when cpu execute this instruction, it should go to "sleep" and pending for event to happen, thus saving power)

I have tested the basic power save cycle, and found that the total power consumption is much lesser, thus I conclude that "WFE" was taking effect.
In conclusion, I don't think this will cause any negative impact as the asm instruction goes directly for cpu to stay pending for event.

sunghan-chang:
> @ritesh55555 Could you share our test result with this PR?

@ritesh55555 Any update for test result? on operation and on power or current?

ritesh55555:
> > @ritesh55555 Could you share our test result with this PR?
> 
> @ritesh55555 Any update for test result? on operation and on power or current?

hi , operation is same as before. It is working as it should . But because of lack of measuring device, i have not yet measured power or current. We need to set up a measuring tool to measure that. i will update the power status once we get a measuring device.

ewoodev:
> In my understanding, the WFE is a cpu sleep, and pending for any event to happen, which means any task with higher priority will be able to trigger the core to handle some operation (ie. context switch). This WFE is different from the power save (ie. Power gating mode), although we call both of them as sleep. But PG sleep is shutting down the core (ie. power lost), whereas WFE is just a kind of rest (?) for the cpu.

hello @edwakuwaku 
If the core enters the core while saving power with the WFE command, does the core will wakeup by the next tick irq or another hw irq?

is it used to just reduce meaningless behavior in idle thread?

edwakuwaku:
> > In my understanding, the WFE is a cpu sleep, and pending for any event to happen, which means any task with higher priority will be able to trigger the core to handle some operation (ie. context switch). This WFE is different from the power save (ie. Power gating mode), although we call both of them as sleep. But PG sleep is shutting down the core (ie. power lost), whereas WFE is just a kind of rest (?) for the cpu.
> 
> hello @edwakuwaku If the core enters the core while saving power with the WFE command, does the core will wakeup by the next tick irq or another hw irq?
> 
> is it used to just reduce meaningless behavior in idle thread?

Hi @ewoodev , any ISR event will wakeup the core, HW irq is used to wakeup system from sleep. Both concept are different.
I'm not sure what are you referring to "meaningless behaviour". This asm instruction is mainly just keeping core in low power state when there is no tasks ongoing, that's why we apply it in idle thread.

sunghan-chang:
@edwakuwaku We can't verify this yet. Once we did, let you know and let me merge this. But it could be little bit some time later.

sunghan-chang:
@gSahitya-samsung Once you check, please let us know result in operation and in power consumption.

ewoodev (approved):

sunghan-chang (approved):

PR #6189, opened 2024-05-16
os, build/configs: add config for common data sections size, clean up xip elf code

abhishek-samsung:
1. added config for common ram size when using xip loadable apps
2. calculated heap allocation for kernel based on app ram size configs
3. added checks for 4KB (l2 page alignment) aligned for app flash and ram partitions


build/configs/rtl8730e/loadable_apps_xip/defconfig
```diff
@@ -8,6 +8,7 @@
 #
 # CONFIG_EXPERIMENTAL is not set
 # CONFIG_DEFAULT_SMALL is not set
+CONFIG_DOCKER_VERSION="2.0.0"
```
sunghan-chang, line 11: Do we still need 2.0.0 with loadable XIP?

build/configs/rtl8730e/loadable_apps_xip/defconfig
```diff
@@ -27,6 +28,7 @@ CONFIG_NUM_APPS=2
 CONFIG_SUPPORT_COMMON_BINARY=y
 CONFIG_COMMON_BINARY_VERSION="200204"
 CONFIG_COMMON_BINARY_NAME="common"
+CONFIG_COMMON_BIN_DYN_RAMSIZE=524288
```
sunghan-chang, line 31: What does it mean? A common binary uses separate heap?

build/configs/rtl8730e/loadable_apps_xip/defconfig
abhishek-samsung: It doesn't require newer toolchain. We can use older toolchain also.

build/configs/rtl8730e/loadable_apps_xip/defconfig
abhishek-samsung: Its the space in RAM we allocate for data and bss sections of common binary.

abhishek-samsung: I am currently using the same configs CONFIG_*_BIN_DYN_RAMSIZE for the space allocated for each app or common in the ram. If its common that space is used only for data and bss. If its app, its used for data, bss and heap. I think using the name DYN_RAMSIZE for common is wrong for common. I will change it to CONFIG_COMMON_BIN_STATIC_RAMSIZE

sunghan-chang: Yes, we need to rename it as you said.
But do we allocate a data and bss in `CONFIG_APP1_BIN_DYN_RAMSIZE`??

abhishek-samsung: Yes, data and bss for apps are allocated in CONFIG_APP1_BIN_DYN_RAMSIZE itself. We first allocate data and bss, then rest of it is allocated as heap.

build/configs/rtl8730e/loadable_apps_xip/defconfig
edwakuwaku: I recommend to use new toolchain, although there is no significant improvement from overall code size.
But along the way the toolchain version improve (ie. v6.3.1 -> v10.3.1), it includes some bug fix and optimization.

os/tools/mkldscript.py
```diff
@@ -79,26 +79,42 @@
 
 ram_end = int(CONFIG_RAM_SIZE) + int(CONFIG_RAM_START, 16)
 
+if ram_end % 4096 != 0 :
+    print("RAM end should be 4KB aligned")
```
r-prabu, line 83: Is the 4KB check is coupled with l2 page size?

r-prabu (approved):

os/tools/mkldscript.py
abhishek-samsung: Yes. For apps we always allocate l2 pages.

build/configs/rtl8730e/loadable_apps_xip/defconfig
abhishek-samsung: As discussed, we will take up changing the desing to remove fixed CONFIG_COMMON_BIN_STATIC_RAMSIZE

ewoodev (approved):

sunghan-chang (approved):

r-prabu (approved):

ritesh55555 (approved):

PR #6190, opened 2024-05-17
build/configs/rtl8730e, os/board/rtl8730e: Move data to external flash

edwakuwaku:
- Load rtk data binary to external flash
- RTK Data binary Address offset starts from 0x0
- Use SPI to read BT FW patch during BT init from external flash


build/configs/rtl8730e/scripts/rlx8730e_img2.ld
```diff
@@ -113,6 +113,8 @@ MEMORY
 	CA32_BL3_DRAM_NS (rwx) :                ORIGIN = 0x60100000, LENGTH = 0x60800000 - 0x60100000	/* CA32 BL3 DRAM NS: 7MB */
 	KM0_BD_DRAM (rwx) :                     ORIGIN = 0x6FFFFFFF, LENGTH = 0x6FFFFFFF - 0x6FFFFFFF	/* KM0 BD DRAM */
 	KM4_DRAM_HEAP_EXT (rwx) :               ORIGIN = 0x6FFFFFFF, LENGTH = 0x6FFFFFFF - 0x6FFFFFFF	/* KM4 PSRAM HEAP EXT */
+
+	CA32_FLASH_EXT (r) :               		ORIGIN = 0x09000000, LENGTH = 0x09020000 - 0x09000000	/* 128KB of external flash */
```
ziyik, line 117: Can we align this? To standardize with other

os/board/rtl8730e/src/component/bluetooth/driver/hci/hci_process/hci_process.c
```diff
@@ -171,6 +185,55 @@ static uint8_t hci_process_reset_baudrate(uint16_t opcode)
...
+    uint8_t *rtlbt_fw_hdr_ptr = NULL;
+    uint32_t rtlbt_fw_addr = 0;
+
+    /* SPI Read size*/
+    rtlbt_fw_hdr_ptr = osif_mem_alloc(RAM_TYPE_DATA_ON, EXTERNAL_FLASH_BINARY_HEADER_SIZE);
+    SPI_SELECT(spi, 1, true);
+    external_flash_read_stream(spi, (char *)rtlbt_fw_hdr_ptr, EXTERNAL_FLASH_BINARY_HEADER_SIZE, HCI_PATCH_ADDR_OFFSET);
+    SPI_SELECT(spi, 1, false);
+    rtlbt_fw_addr = (rtlbt_fw_hdr_ptr[12]) | (rtlbt_fw_hdr_ptr[13] << 8) | (rtlbt_fw_hdr_ptr[14] << 16) | (rtlbt_fw_hdr_ptr[15] << 24);
+    rtlbt_fw_size = (rtlbt_fw_hdr_ptr[8]) | (rtlbt_fw_hdr_ptr[9] << 8) | (rtlbt_fw_hdr_ptr[10] << 16) | (rtlbt_fw_hdr_ptr[11] << 24);
```
ziyik, line 223: Should we check the address and size in case it's empty?

os/board/rtl8730e/src/component/bluetooth/driver/hci/hci_process/hci_process.c
```diff
@@ -171,6 +185,57 @@ static uint8_t hci_process_reset_baudrate(uint16_t opcode)
...
+    /* Below condition is minimal requirement to meet */
+    DEBUGASSERT(rtlbt_fw_size > 0 && rtlbt_fw_addr >= EXT_FLASH_BASE_ADDRESS);
```
ziyik, line 225: When the Flash is empty, what is the default value? (FF/00)
If the default value is FF then the checking should add FF checking.

sunghan-chang:
@edwakuwaku
Could you elaborate this more? Is it the data which is in the "ftl" partition? The "data" is the "BT FW patch"?
There is no change on the partition map, but you set another 128KB in the external flash. Could you let me know where it moves from?
Is this working well without external flash?

edwakuwaku:
Hi @sunghan-chang ,

The data is not in the "ftl" partition, it is included in the kernel binary. Currently, we have yet to find other data's that can be included into the data binary, thus BT FW patch is equivalent to the whole data binary.
The partition map is not needed to be changed, we are extracting the data which was included in the kernel binary, and move out to the external flash.
As you can see, the changes are covered under the macro CONFIG_SECOND_FLASH_PARTITION, which means only if there is external flash, we will move the data out, or else it will still be included in kernel binary.

I hope this clear your doubts.

edwakuwaku:
I tested with/without external flash, both are working fine from my side.

sunghan-chang:
@edwakuwaku Thank you for explanation. It's clear for me. Can I merge this now?

edwakuwaku:
> @edwakuwaku Thank you for explanation. It's clear for me. Can I merge this now?

Yes please~

sunghan-chang:
@edwakuwaku Sorry but are you able to leave kernel package size change in the commit description?

edwakuwaku:
Hi @sunghan-chang , done, please check~

sunghan-chang (approved):

PR #6191, opened 2024-05-21
os: CPU hotplug mode implementation

edwakuwaku:
- This PR introduces cpu hotplug mode, which is a necessary feature to integrate PM and SMP.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_idle.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_idle.c
@@ -36,6 +36,7 @@
 
 #ifdef CONFIG_SMP
 #include "gic.h"
+#include "smp.h"
 #endif
 
 /****************************************************************************
@@ -61,7 +62,6 @@ static void up_idlepm(void)
 	/* Check if new attainable state is different from current state */
 	if (newstate != oldstate)
 	{
-		/* TODO: SMP yet to be verified*/
 		/* Perform board-specific, state-dependent logic here */
 	  	pmvdbg("newstate= %d oldstate=%d\n", newstate, oldstate);
 
@@ -80,8 +80,9 @@ static void up_idlepm(void)
 		/* TODO: When LCD is integrated, some of the state operation might require revision
 		   PM_IDLE: We will expect LCD to go through background light dimming
 		   PM_STANDBY: LCD shut down, single core might be able to handle remaining tasks
-		   			   Then, we can do vPortSecondaryOff() for SMP case (ie. shutdown secondary core)
 		*/
+		/* Consider to perform DVFS in different PM state, if the average power consumption
+		   is still not up to satisfactory */
 		switch (newstate) {
 			case PM_NORMAL:
 				/* In PM_NORMAL, we have nothing to do, set core to WFE */
@@ -102,10 +103,15 @@ static void up_idlepm(void)
 				   EG: Secondary core should be in hotplug mode, primary core should check
 				   the secondary core state before going to sleep
 				*/
-#ifndef CONFIG_SMP
-				up_set_pm_timer();
-#endif
 				if (up_cpu_index() == 0) {
+#ifdef CONFIG_SMP
+/* For CPU_RUNNING case 
+CPU_RUNNING:
+Send SGI for core 1 to do task migration and shut down, we will expect that to be handled
+nicely before the next cycle of core 0 reaches
+*/
+RESLEEP_1:
+#endif
 					/* mask sys tick interrupt*/
 					arm_arch_timer_int_mask(1);
 					up_timer_disable();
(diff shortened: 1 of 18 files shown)


os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -201,15 +194,21 @@ void smp_init(void)
...
+			dbg("CPU%d: failed to boot: %d\n", (int)xCoreID, (int)err);
 		}
 	}
 #endif
-
+	/* We should check whether it is a warm boot here
+	   If yes, we should send SGI1
+	 */
+#ifdef CONFIG_CPU_HOTPLUG
+	if (up_get_secondary_cpu_state(1) == CPU_WAKE_FROM_SLEEP) {
```
kishore-sn, line 210: Can we have #define for core id instead of hardcoding?

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -201,15 +194,21 @@ void smp_init(void)
...
 		if (xCoreID == up_cpu_index()) {
-			// pmu_set_secondary_cpu_state(xCoreID, CPU1_RUNNING);
+			up_set_secondary_cpu_state(xCoreID, CPU_RUNNING);
 			continue;
 		}
-
 		err = psci_cpu_on(xCoreID, (unsigned long)__cpu1_start);
 		if (err < 0) {
-			printf("CPU%d: failed to boot: %d\n", (int)xCoreID, (int)err);
+			dbg("CPU%d: failed to boot: %d\n", (int)xCoreID, (int)err);
```
kishore-sn, line 202: Is it ok to just print the error msg here and proceed?
When we return from this API, we assume that the second core has booted up.

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -0,0 +1,98 @@
...
+	arm_savestate(rtcb->xcp.regs);
+	rtcb->task_state = TSTATE_TASK_ASSIGNED;
+	CURRENT_REGS = NULL;
+
+	/* MINOR: PSCI is a general interface for controlling power of cortex-A cores 
+	It is one of the feature in arm-trusted-firmware (ie. Trustzone-A)
+	Currently, the definitions are provided by chip specific arch layer
+	*/
+	/* Shut down the secondary core */
+   	psci_cpu_off(0);
```
kishore-sn, line 94: I think we are always passing zero to this API. Is there any case where other value needs to be passed?
In nuttx, this API does not take any parameters.

os/arch/arm/src/amebasmart/amebasmart_smp.c
edwakuwaku: Done~

os/arch/arm/src/amebasmart/amebasmart_smp.c
edwakuwaku: Your point make sense, the operation is irreversible (ie. unrecoverable). So, we should make a crash here.

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
edwakuwaku: As I have checked in our ATF layer, it seems that we will not be using any parameter passed in as well. Thus, it will be removed.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -79,10 +80,20 @@ static void up_idlepm(void)
 		/* MCU-specific power management logic */
 		switch (newstate) {
 			case PM_NORMAL:
+#ifdef CONFIG_PM_DVFS
+				up_set_dvfs(0);
+#endif
 				break;
 			case PM_IDLE:
+#ifdef CONFIG_PM_DVFS
+				up_set_dvfs(1);
```
ewoodev, line 89: What kind of Sinairo is this?

We also use the highest freq of the highest cpu for IDLE and STANDBY.

Does this return to the top freq on the next tick irq?

edwakuwaku: This is added for saving power in different state. I mean we do it in idle thread because PM state change is expected when application has lock/unlock to the idle domain.
Since IDLE and STANDBY still use highest frequency, then which scenario apply DVFS?

edwakuwaku:
Hi @kishore-sn , any update on SMP verification? Other than that, may I know the status for this PR? Any test conducted or verification result from your side?

sunghan-chang:
@kishore-sn Could you review and leave approval?

os/arch/arm/src/amebasmart/amebasmart_idle.c
ewoodev: Thank you for suggesting
There is no DVFS scenario that we are reviewing yet.

os/kernel/init/os_smpstart.c
```diff
@@ -79,8 +79,10 @@ void os_idle_trampoline(void)
 
 	for (; ; ) {
 		/* Perform any processor-specific idle state operations */
-
-		up_idle();
+		/* Let primary core control the PM idle loop, other 
+		   cores may stay at WFE and wait for signal from primary core
+		*/
+		SP_WFE();
```
ewoodev, line 85: It looks good to me.

If second core entering the wfe, power can be reduced than two core is working?

edwakuwaku: Yes, the power consumption should be as below:
(primary core running + secondary core wife) < (primary core running + secondary core running)
But, the core task assignment depends on kernel SMP mechanism.

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -201,15 +168,30 @@ void smp_init(void)
...
+#endif
+	/* We should check whether it is a warm boot here
+	   If yes, we should send SGI1
+	 */
+#ifdef CONFIG_CPU_HOTPLUG
+	/* In our case, CONFIG_SMP_NCPUS is always 2, thus
+	xCoreID will be identified as the last core we have
+	from above loop, so we will use xCoreID here
+	*/
+	if (up_get_secondary_cpu_state(xCoreID) == CPU_WAKE_FROM_SLEEP) {
```
kishore-sn, line 193: When we come out of the for loop, the value of xCoreID will be 2. 
So, it will never enter this if condition. Please set core id to 1.

edwakuwaku: Thank you for identifying this! Sorry for the mistake, I have fixed it.

kishore-sn (approved):

PR #6192, opened 2024-05-22
lib: Apply DEBUG_SET_CALLER_ADDR macro

ewoodev:
When a memory is alloc, we store the caller address requested by which code. However, if we wrap the alloc api and use it, the caller only looks at the wrapped api.

Therefore, we apply the DEBUG_SET_CALLER_ADDR macro to actually store the caller's address in alloc.

### DIFF
--- a/lib/libc/misc/lib_envpath.c
+++ b/lib/libc/misc/lib_envpath.c
@@ -65,6 +65,7 @@
 #include <assert.h>
 
 #include <tinyara/envpath.h>
+#include <tinyara/mm/mm.h>
 
 #include "libc.h"
 
@@ -138,7 +139,7 @@ ENVPATH_HANDLE envpath_init(FAR const char *name)
 		printf("Allocate a container failed\n");
 		return (ENVPATH_HANDLE) NULL;
 	}
-
+	DEBUG_SET_CALLER_ADDR(envpath);
 	/* Populate the container */
 
 	strcpy(envpath->path, path);
--- a/lib/libc/misc/lib_hashmap.c
+++ b/lib/libc/misc/lib_hashmap.c
@@ -97,6 +97,7 @@ struct hashmap_s *hashmap_create(int startsize)
 	if (hm == NULL) {
 		return NULL;
 	}
+	DEBUG_SET_CALLER_ADDR(hm);
 
 	if (!startsize) {
 		startsize = TABLE_STARTSIZE;
@@ -185,6 +186,7 @@ unsigned long* hashmap_get_keyset(struct hashmap_s *hash)
 
 		keyset = (unsigned long *)lib_malloc(sizeof(unsigned long) * hash->count);
 		if (keyset != NULL) {
+			DEBUG_SET_CALLER_ADDR(keyset);
 			for (i = 0; i < hash->size; i++) {
 				if (hash->table[i].flags & ACTIVE) {
 					keyset[idx++] = hash->table[i].key;
--- a/lib/libxx/libxx_newa.cxx
+++ b/lib/libxx/libxx_newa.cxx
@@ -55,6 +55,7 @@
 //***************************************************************************
 
 #include <tinyara/config.h>
+#include <tinyara/mm/mm.h>
 #include <cstddef>
 #include <debug.h>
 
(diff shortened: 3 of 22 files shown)


lib/libxx/libxx_newa.cxx
```diff
@@ -111,6 +111,8 @@ void *operator new[](unsigned int nbytes)
     // we cannot throw an exception!  We are bad.
 
     dbg("Failed to allocate\n");
+  } else {
+	DEBUG_SET_CALLER_ADDR(alloc);
```
sunghan-chang, line 115: Let's use the same indentation in the file.

lib/libc/misc/lib_hashmap.c
```diff
@@ -185,6 +186,7 @@ unsigned long* hashmap_get_keyset(struct hashmap_s *hash)
 
 		keyset = (unsigned long *)lib_malloc(sizeof(unsigned long) * hash->count);
 		if (keyset != NULL) {
+			DEBUG_SET_CALLER_ADDR(envpath);
```
sunghan-chang, line 189: `keyset`, not `envpath`.

sunghan-chang (approved):

os/include/tinyara/mm/mm.h
```diff
@@ -258,7 +258,7 @@ typedef size_t mmsize_t;
 /* This macro updates a caller return address in the 'mem' node.
  * This will show the real owner of the 'mem' even it's allocated through wrapping APIs of malloc.
  */
-#define DEBUG_SET_CALLER_ADDR(mem) heapinfo_set_caller_addr(mem, __builtin_return_address(0))
+#define DEBUG_SET_CALLER_ADDR(mem) heapinfo_set_caller_addr(mem, (mmaddress_t)__builtin_return_address(0))
```
sunghan-chang, line 261: @r-prabu @kishore-sn @abhishek-samsung We will change a type of the `mmaddress_t` from `size_t` to `void *`. Because it's an address.
```
typedef size_t mmaddress_t;             /* 32 bit address space */

->

typdef void *mmaddress_t;
```
How do you think?

kishore-sn: I think that would be good.

lib/libc/misc/lib_envpath.c
```diff
@@ -138,7 +139,7 @@ ENVPATH_HANDLE envpath_init(FAR const char *name)
 		printf("Allocate a container failed\n");
 		return (ENVPATH_HANDLE) NULL;
 	}
-
+	DEBUG_SET_CALLER_ADDR(envpath);
```
kishore-sn, line 142: `lib_malloc' is a preprocessor directive. It is replaced with `malloc` or `kmm_malloc`.
So, in this case we are not actually using a wrapper API. So, do we actually need DEBUG_SET_CALLER_ADDR in this case?

sunghan-chang: The main purpose of saving the owner is getting debug point of memory issue like memory leak or corruption. But allocation in common codes like libc does not save real owner. We can't find who handles the allocation memory. This will resolve above. That's why we add this here.

os/include/tinyara/mm/mm.h
```diff
@@ -266,7 +266,7 @@ typedef size_t mmsize_t;
 /* typedef is used for defining size of address space */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-typedef size_t mmaddress_t;             /* 32 bit address space */
+typedef void* mmaddress_t;             /* 32 bit address space */
```
sunghan-chang: void* mmaddress_t
->
void *mmaddress_t

sunghan-chang (commented):
typo in commit description
ardument -> argument

sunghan-chang (commented):
```
os/kernel/semaphore: Change ARCH_GET_RET_ADDRESS __builtin_return_add…
…ress(0)

Changes the use of __builtin_return_address() in semaphore to check for callers.
```
Could you leave why?

ewoodev:
> ```
> os/kernel/semaphore: Change ARCH_GET_RET_ADDRESS __builtin_return_add…
> …ress(0)
> 
> Changes the use of __builtin_return_address() in semaphore to check for callers.
> ```
> 
> Could you leave why?

I added description in commit description

sunghan-chang (approved):

PR #6193, opened 2024-05-23
os/board/rtl8730e: add API to return BLE indication queue count

yeetee179:
- add ble_server_get_indicate_pending_count API that returns current indication queue count to application


framework/src/ble_manager/ble_manager_event.h
```diff
@@ -62,6 +62,7 @@ typedef enum {
 	BLE_CMD_GET_PROFILE_COUNT,
 	BLE_CMD_CHARACT_NOTI,
 	BLE_CMD_CHARACT_INDI,
+	BLE_CMD_GET_INDICATE_PENDDING_CNT,
```
kishore-sn, line 65: PENDDING -> PENDING

framework/src/ble_manager/ble_manager_api.c
```diff
@@ -343,6 +343,15 @@ ble_result_e ble_server_charact_indicate(ble_attr_handle attr_handle, ble_conn_h
 	RETURN_RESULT(res, msg);
 }
 
+ble_result_e ble_server_get_indicate_pending_count(ble_conn_handle con_handle, uint8_t *count)
+{
```
kishore-sn, line 347: Pls perform NULL check for `count` before using it.

yeetee179: Added a NULL for count

framework/src/ble_manager/ble_manager_event.h
yeetee179: Modified as suggested

giwon-nam (approved):
LGTM.
I have confirmed that it works on the product code.

hs36-kim (approved):
LGTM we tested.

PR #6197, opened 2024-05-24
os/arch/arn/src/amebasmart: Modify i2c pins for board revision 5th

ziyik:
-Modify pins for i2c0 and i2c1 according to board revision 5th

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2c.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2c.c
@@ -301,8 +301,13 @@ static const struct amebasmart_i2c_config_s amebasmart_i2c0_config = {
 	//.busy_idle = CONFIG_I2C0_BUSYIDLE,
 	//.filtscl = CONFIG_I2C0_FILTSCL,
 	//.filtsda = CONFIG_I2C0_FILTSDA,
+#if CONFIG_RTL8730E_BOARD_REVISION >= 5
+	.scl_pin = PB_30,
+	.sda_pin = PB_29,
+#else
 	.scl_pin = PA_10,
 	.sda_pin = PA_9,
+#endif
 #ifndef CONFIG_I2C_SLAVE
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@abhishek-samsung Could you confirm this?

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -301,8 +301,13 @@ static const struct amebasmart_i2c_config_s amebasmart_i2c0_config = {
 	//.busy_idle = CONFIG_I2C0_BUSYIDLE,
 	//.filtscl = CONFIG_I2C0_FILTSCL,
 	//.filtsda = CONFIG_I2C0_FILTSDA,
+#if CONFIG_RTL8730E_BOARD_REVISION >= 5
+	.scl_pin = PA_29,
+	.sda_pin = PA_28,
+#else
```
abhishek-samsung, line 307: https://github.com/Samsung/TizenRT/blob/9522f9b6f99fcf192de3cb2a6de3db240b46375f/os/arch/arm/src/amebasmart/amebasmart_i2c.c#L359-L365

I2C port 2 is already using PA_28 and PA_29 and it's the same mentioned in the data shared by h/w team. So, I am not sure if its the right change.

ziyik: This is a mistake on my side, please ignore this change, I have reverted it.

sunghan-chang (commented):
i2c0 also has pin map changed.
pa_9 -> pb29
pa_10 -> pb30

Could you check it as well?

ziyik:
> #if CONFIG_RTL8730E_BOARD_REVISION >= 5
> 	.scl_pin = PA_29,
> 	.sda_pin = PA_28,
> #else

Yes, you are right

sunghan-chang (approved):

sunghan-chang (approved):

PR #6199, opened 2024-05-24
build/tools/amebasmart/gnu_utility/: fix km4 crash when running wificsi

zhongnuo-tang:
1. crash was caused by memset a nullpointer, add condition to prevent such case


hiarungupta:
PR looks fine, no crash observed.

sunghan-chang (approved):

PR #6200, opened 2024-05-24
os: Fix XIP_ELF runtime loading crash issue

edwakuwaku:
-  In TRPK file, it might include several binaries together, which means the binary running on TizenLite might not be the starting address of the TRPK file.
- In return, we will face crash issue, since there is a address mismatch between the correct address and the current address of the apps binary (ie. the address of the apps binary should be calculated from the binary which is running on TizenLite, not the trpk file itself)

### DIFF
--- a/os/tools/mkldscript.py
+++ b/os/tools/mkldscript.py
@@ -37,7 +37,13 @@ CONFIG_APP_BINARY_SEPARATION = util.get_value_from_file(cfg_file, "CONFIG_APP_BI
 CONFIG_SUPPORT_COMMON_BINARY = util.get_value_from_file(cfg_file, "CONFIG_SUPPORT_COMMON_BINARY=").rstrip('\n')
 CONFIG_USE_BP = util.get_value_from_file(cfg_file, "CONFIG_USE_BP=").rstrip('\n')
 CONFIG_NUM_APPS = util.get_value_from_file(cfg_file, "CONFIG_NUM_APPS=").rstrip('\n')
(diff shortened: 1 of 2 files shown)


sunghan-chang (approved):

edwakuwaku:
Hi @abhishek-samsung , could you please help to review the changes?

os/tools/mkldscript.py
```diff
@@ -37,7 +37,13 @@
...
 CONFIG_USE_BP = util.get_value_from_file(cfg_file, "CONFIG_USE_BP=").rstrip('\n')
 CONFIG_NUM_APPS = util.get_value_from_file(cfg_file, "CONFIG_NUM_APPS=").rstrip('\n')
-CONFIG_FLASH_START_ADDR = util.get_value_from_file(cfg_file, "CONFIG_FLASH_START_ADDR=").rstrip('\n')
+# Get flash virtual remapped address instead of physical address
+# CONFIG_FLASH_VSTART_LOADABLE = util.get_value_from_file(cfg_file, "CONFIG_FLASH_VSTART_LOADABLE=").rstrip('\n')
+
+# Hardcoded as of now. REASON: Kernel TRPK might include more than one binary information, the exact starting address
+# of loadable apps binary has to be shifted accordingly (ie. Minus the offset here), to make sure XIP_ELF is working fine 
+# TODO: Dynamically get the offset from Kernel TRPK binary file
+CONFIG_FLASH_VSTART_LOADABLE = "0x0DF96000"
```
sunghan-chang, line 46: This value is set here. Then, do we need to add it in Kconfig?

abhishek-samsung (approved):

PR #6203, opened 2024-05-27
os, build: Add support for loadable XIP_ELF offset calculation

edwakuwaku:
- mkldscript for loadable is generic tool, to break the dependency, chip specific should implement the logic independently
- If trpk file contains multiple binary, the logic should be applied
- Add flash virtual starting address and trpk multiple binary macros in defconfig

### DIFF
--- a/build/tools/amebasmart/gnu_utility/loadable_xip_elf.py
+++ b/build/tools/amebasmart/gnu_utility/loadable_xip_elf.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+###########################################################################
+#
+# Copyright 2024 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+import os
+
+filepath = os.path.dirname(__file__) + "/km0_km4_app.bin"
+alignment = 4096
+
+def get_offset():
+    bin_size = os.path.getsize(filepath)
+    # Binary need to be aligned to 4K
+    bin_size = (bin_size + alignment - 1) // alignment * alignment
+    # 4K manifest before AP binary (ie. running on Tizen Lite) starts
+    bin_size = bin_size + alignment
+    return hex(bin_size)
+
--- a/build/tools/amebasmart/gnu_utility/rtk_data_binary.py
+++ b/build/tools/amebasmart/gnu_utility/rtk_data_binary.py
@@ -1,4 +1,21 @@
 #!/usr/bin/env python
+###########################################################################
+#
+# Copyright 2024 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
 
 import os
 import sys
--- a/os/board/rtl8730e/Kconfig
+++ b/os/board/rtl8730e/Kconfig
@@ -20,6 +20,21 @@ config FLASH_START_ADDR
         A start address of flash (in hex).
         This is fixed value, so user doesn't need to change it.
 
+config TRPK_CONTAINS_MULTIPLE_BINARY
+    bool
+    default n
+    depends on XIP_ELF
+    help
+        This macro is to differentiate whether the kernel trpk contains
+        more than 1 binary. If yes, we have to recalculate the offset
+        according to the binary that is running on Tizen Lite, to make
+        sure the loadable_apps for XIP_ELF is placed to the correct
+        virtual flash address (ie. Only Loadable_apps with running on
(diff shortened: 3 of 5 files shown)


edwakuwaku:
Hi @abhishek-samsung , could you please help to verify this PR on internal branch?

build/tools/amebasmart/gnu_utility/loadable_xip_elf.py
```diff
@@ -0,0 +1,31 @@
...
+filepath = os.path.dirname(__file__) + "/km0_km4_app.bin"
+alignment = 4096
+
+def get_offset():
+    bin_size = os.path.getsize(filepath)
+    # Binary need to be aligned to 4K
+    bin_size = (bin_size + alignment - 1) // alignment * alignment
+    # 4K padding before AP binary (ie. running on Tizen Lite) starts
+    bin_size = bin_size + alignment
+    return hex(bin_size)
```
abhishek-samsung, line 30: I have a minor question. Does this also handle the size of Samsung header that we add at the start of the trpk file? If it doesn't then we will have issues at some cases where the alignment goes wrong.

abhishek-samsung (commented):
@edwakuwaku , this PR works fine on the internal branch.

build/tools/amebasmart/gnu_utility/loadable_xip_elf.py
edwakuwaku: Yes, the km0_km4 binary has a 4kb of header which is same as trpk file, despite of the values. Which means calculating from km0_km4 will yield the correct result.
In detail: 4KB header -> 4KB km4_km0 manifest -> km4_km0 image size with 4KB alignment -> 4KB ca32 manifset
Thus, your concern should be handled in the above logic.

abhishek-samsung: Okay, thanks for explaining

edwakuwaku: If there is no further question, could you leave an approval? Thanks~

build/tools/amebasmart/gnu_utility/rtk_data_binary.py
```diff
@@ -1,4 +1,21 @@
 #!/usr/bin/env python
+###########################################################################
```
abhishek-samsung, line 2: minor comment, this change doesn't seem to fit with the commit. Could it be moved to a new commit?

abhishek-samsung (approved):

build/tools/amebasmart/gnu_utility/rtk_data_binary.py
edwakuwaku: done~

abhishek-samsung (approved):

os/board/rtl8730e/Kconfig
```diff
@@ -20,6 +20,17 @@ config FLASH_START_ADDR
...
+config TRPK_CONTAINS_MULTIPLE_BINARY
+    bool
+    default n
+    depends on XIP_ELF
+    help
+        This macro is to differentiate whether the kernel trpk contains
+        more than 1 binary. If yes, we have to recalculate the offset
+        according to the binary that is running on Tizen Lite, to make
+        sure the loadable_apps for XIP_ELF is placed to the correct
+        virtual flash address.
```
sunghan-chang, line 32: The description is the same with the flat. But we don't need this for the flat. Are you able to leave why `XIP_ELF` requires this more?

build/configs/rtl8730e/loadable_apps/defconfig
```diff
@@ -313,6 +313,8 @@ CONFIG_FLASH_MINOR=0
 CONFIG_FLASH_PART_SIZE="60,40,12,400,1844,4808,384,1844,4808,384,512,1280,8,"
 CONFIG_FLASH_PART_TYPE="none,none,none,none,kernel,bin,bin,kernel,bin,bin,smartfs,ftl,bootparam,"
 CONFIG_FLASH_PART_NAME="bl1,reserved,ftl,ss,kernel,app1,app2,kernel,app1,app2,userfs,reserved,bootparam,"
+CONFIG_TRPK_CONTAINS_MULTIPLE_BINARY=y
+CONFIG_FLASH_VSTART_LOADABLE=0xe000000
```
sunghan-chang, line 317: I think this defconfig does not enable `XIP_ELF` which you make a dependency. 
@abhishek-samsung Could you guide him which defconfig should be enabled?

edwakuwaku: But in https://github.com/Samsung/TizenRT/pull/6185, the loadable configs will apply XIP_ELF.
Also as I have known, there seems no more loadable ram cases for RTL8730E in the future, due to ram shortage.

abhishek-samsung: > @abhishek-samsung Could you guide him which defconfig should be enabled?

in the current code loadable_apps_xip/defconfig is the only one using xip apps. So changing only that particular config should be enough. I think #6185 can apply these changes required in that PR self.

edwakuwaku: okay, then I will only update to loadable_app_xip/defconfig

os/board/rtl8730e/Kconfig
edwakuwaku: Hi @sunghan-chang , could you check the comments again?

os/board/rtl8730e/Kconfig
```diff
@@ -20,6 +20,21 @@ config FLASH_START_ADDR
         A start address of flash (in hex).
         This is fixed value, so user doesn't need to change it.
 
+config TRPK_CONTAINS_MULTIPLE_BINARY
+    bool
+    default n
+    depends on XIP_ELF
+    help
+        This macro is to differentiate whether the kernel trpk contains
```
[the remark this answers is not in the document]
edwakuwaku: Yes, but is there any conflict in between? We used kernel trpk to calculate the offset for common and app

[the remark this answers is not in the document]
sunghan-chang: Oh ok. Then the common and the app get affected from the kernel that includes km4. Right?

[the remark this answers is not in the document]
edwakuwaku: yes correct, but we won't state "km4/km0" here, that's why this is just a bool to differentiate whether kernel trpk contains multiple binary

sunghan-chang (approved):

PR #6208, opened 2024-05-27
os/pm: fix WID:35656010	defect caused by no null check and wrong casting

ritesh55555:
- this commit fixes defect in pm module. One fix is to add a null check. Another fix is to use appropriate data type in pm timer module.


sunghan-chang:
@ewoodev Please review this.

ewoodev (approved):

os/pm/pm_changestate.c
```diff
@@ -110,6 +110,10 @@ static void pm_timer(int domain)
 
 	if (!pdom->wdog) {
```
ewoodev, line 111: Please remove pm_timer function.
we are not using this.

ewoodev (approved):

sunghan-chang (approved):

PR #6209, opened 2024-05-27
framework/src/eventloop: Fix Svace issues

kishore-sn:
Fix Svace issue related to out of bounds array access. OVERFLOW_UNDER_CHECK	Buffer 'g_thread_safe_cb_list' of size 128 accessed at eventloop_async.c:130 can overflow, since its index 'list_idx' can have value 128 that is out of range, as indicated by preceding conditional expression at eventloop_async.c:125


jeongarmy (approved):
lgtm

PR #6212, opened 2024-05-30
os/arch/arm/src/armv7-a/arm_assert.c: Add \n in security_level lldbg

jylee9613:
When assert occurs, the assert log is printed continuously with the previous log. (It occurs only in AIDual)

```
E[MEDIA][audio_manager.c]::get_device_process_handler_mesecurity level: 0
===========================================================
Assertion details
===========================================================
```

This commit adds \n in front of security_level log to distinguish between normal log and assert log.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -555,6 +555,8 @@ void up_assert(const uint8_t *filename, int lineno)
 	ARCH_GET_RET_ADDRESS(kernel_assert_location)
 	struct tcb_s *fault_tcb = this_task();
 
+	/* Add new line to distinguish between normal log and assert log.*/
+	lldbg_noarg("\n");
 
 	board_autoled_on(LED_ASSERTION);
 
--- a/os/arch/arm/src/armv7-m/up_assert.c
+++ b/os/arch/arm/src/armv7-m/up_assert.c
@@ -533,6 +533,9 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	irqstate_t flags = irqsave();
 
+	/* Add new line to distinguish between normal log and assert log.*/
+	lldbg_noarg("\n");
+
 	struct tcb_s *fault_tcb = this_task();
 	bool is_kmm_corruption = false;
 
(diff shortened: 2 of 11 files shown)


os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -580,7 +580,7 @@ void up_assert(const uint8_t *filename, int lineno)
 
 	irqstate_t flags = irqsave();
 #ifdef CONFIG_SECURITY_LEVEL
-	lldbg_noarg("security level: %d\n", get_security_level());
+	lldbg_noarg("\nsecurity level: %d\n", get_security_level());
```
sunghan-chang, line 583: If you modify here, when someone adds logs before this, this should be changed again. And need to be checked whether someone adds `\n` at start. I suggest adding `\n` at start of the function.

Other question is
why only `armv7-a` file is changed? how about `armv8-m`, `armv7-m`, `armv7-r`, hardfault? memfault and other fault files?

jylee9613: Your suggest means that adding `\n` at start of up_assert?

As I mentioned in commit message, I only added in `armv7-a` because log continues only in `armv7-a`. 
`arm7-m, arm7-r, arm8-m` is not occurred.

sunghan-chang:
@jylee9613 Could you check up/arm_xxxfault functions as well?

jylee9613:
> @jylee9613 Could you check up/arm_xxxfault functions as well?

Updated. `arm7-a` & `arm7-r` already have `\n`, so I only updated `arm8-m` and `arm7-m`

sunghan-chang (approved):

os/arch/arm/src/armv7-m/up_memfault.c
```diff
@@ -148,6 +148,9 @@ static inline void print_memfault_detail(uint32_t *regs, uint32_t cfsr, uint32_t
 
 int up_memfault(int irq, FAR void *context, FAR void *arg)
 {
+	/* Add new line to distinguish between normal log and assert log.*/
+	lldbg("\n");
```
sunghan-chang, line 152: fault has the irqsave at start of the function. Let's do this after the irqsave.

sunghan-chang (dismissed):

sunghan-chang (approved):

PR #6214, opened 2024-05-30
os/arch/arm/src/amebasmart: Amebasmart rtc fix

kaikaikaime:
* Replace function name with 'rtk'

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_rtc.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_rtc.c
@@ -118,8 +118,8 @@ int up_rtc_gettime(FAR struct timespec *tp)
 {
 	time_t timer;
 	int ret;
-	timer = rtc_read();
-
+	timer = rtk_rtc_read();
+	
 	ret = gmtime_r(&timer, tp);
 	if (ret == 0) {
 		return -EOVERFLOW;
@@ -150,7 +150,7 @@ int up_rtc_gettime(FAR struct timespec *tp)
 
 time_t up_rtc_time(void)
 {
-	return rtc_read();
+		return rtk_rtc_read();
 }
 
 /************************************************************************************
@@ -171,11 +171,11 @@ time_t up_rtc_time(void)
 int up_rtc_settime(FAR const struct timespec *ts)
 {
 	if (ts) {
-		rtc_write(ts->tv_sec);
+		rtk_rtc_write(ts->tv_sec);
 
 		return OK;
 	} else {
-		return ERROR;
+				return ERROR;
 	}
 }
 
@@ -202,7 +202,7 @@ int up_rtc_initialize(void)
 {
 #ifdef CONFIG_RTC_DRIVER
 	/* Perform RTC initialization */
-
+	
 	rtc_init();
 #endif
 
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/rtc_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/rtc_api.c
@@ -181,7 +181,7 @@ int rtc_isenabled(void)
   * @param  t: Seconds from 1970.1.1 00:00:00 to specified data and time which is to be set.
   * @retval  none
   */
-void rtc_write(time_t t)
+void rtk_rtc_write(time_t t)
 {
 	/* Convert the time in to a tm*/
 	struct tm *timeinfo = localtime(&t);
@@ -204,7 +204,7 @@ void rtc_write(time_t t)
   * @param  none
   * @return Current timestamp in seconds which is calculated from 1970.1.1 00:00:00.
   */
-time_t rtc_read(void)
+time_t rtk_rtc_read(void)
 {
 	time_t t;
 	struct tm tm_temp;


abhishek-samsung:
I applied your patch on my test code and still the values are same as before and it seems to not work. Could you also please explain how changing the name would fix this issue?

branch : https://github.com/abhishek-samsung/TizenRT/tree/rtc_issue_updated
config : flat_dev
command : rtc

```
TASH>>rtc
TASH>>=== RTC Accuracy Test ===
amebasmart_rdtime: value fo timer : 4294967295 <--- same value is repeated
rtc_ioctl: tm_sec : 15, tm_min : 28
[Total 0m passed] Time Difference : 0s
amebasmart_rdtime: value fo timer : 4294967295
rtc_ioctl: tm_sec : 15, tm_min : 28
[Total 0m passed] Time Difference : 0s
amebasmart_rdtime: value fo timer : 4294967295
rtc_ioctl: tm_sec : 15, tm_min : 28
[Total 0m passed] Time Difference : 0s
amebasmart_rdtime: value fo timer : 4294967295
rtc_ioctl: tm_sec : 15, tm_min : 28
[Total 0m passed] Time Difference : 0s
```

kaikaikaime:
Dear Abhishek,

Upon further inspection, this is found that the problem lies in the usage parameter for the example of rtc_main. 
Without the changes of this PR:

The lowest date allowed is : 
1970-1-1
reference: [rtc man pages](https://man7.org/linux/man-pages/man4/rtc.4.html)

Therefore the parameters at minimum should be:
```
prev_time.tm_year = 70; // 1970 - 1900
prev_time.tm_mday = 1; //1~31
prev_time.tm_mon = 0;  //0~11 
```
instead of memset everything to 0.

I can revert the changes of 'rtk_' as this does not fix the current issue. 
Instead shall we add a warning or an assert if the dates set is below this set of values?

Chuen Kai

abhishek-samsung:
> Dear Abhishek,
> 
> Upon further inspection, this is found that the problem lies in the usage parameter for the example of rtc_main. Without the changes of this PR:
> 
> The lowest date allowed is : 1970-1-1 reference: [rtc man pages](https://man7.org/linux/man-pages/man4/rtc.4.html)
> 
> Therefore the parameters at minimum should be:
> 
> ```
> prev_time.tm_year = 70; // 1970 - 1900
> prev_time.tm_mday = 1; //1~31
> prev_time.tm_mon = 0;  //0~11 
> ```
> 
> instead of memset everything to 0.
> 
> I can revert the changes of 'rtk_' as this does not fix the current issue. Instead shall we add a warning or an assert if the dates set is below this set of values?
> 
> Chuen Kai

Thanks for checking with the test code. After applying the initialization that you provided for prev_time, rtc test seems to run fine. However, with rtl8721csm, the example works fine without the prev_time init. Could you let us know why there is a difference?

branch : https://github.com/abhishek-samsung/TizenRT/tree/rtc_issue_updated
config : rtl8721csm/hello

```
TASH>>rtc
TASH>>=== RTC Accuracy Test ===
rtc_ioctl: tm_sec : 1, tm_min : 0
[Total 0m passed] Time Difference : 1s
rtc_ioctl: tm_sec : 2, tm_min : 0
[Total 0m passed] Time Difference : 1s
rtc_ioctl: tm_sec : 3, tm_min : 0
[Total 0m passed] Time Difference : 1s
rtc_ioctl: tm_sec : 4, tm_min : 0
[Total 0m passed] Time Difference : 1s
rtc_ioctl: tm_sec : 5, tm_min : 0
```

kaikaikaime:
rtl8721csm does not have a year input, therefore 0 will be acceptable. 
Since rtl8730e has added the year parameter in the rtc, it is required to conform to 1970-1-1 restriction.

sunghan-chang:
```
config RTC_DATETIME
	bool "Date/Time RTC Support"
	default n
	---help---
		There are two general types of RTC:  (1) A simple battery backed
		counter that keeps the time when power is down, and (2) a full
		date / time RTC the provides the date and time information, often in
		BCD format.  If RTC_DATETIME is selected, it specifies this second kind
		of RTC. 

if !RTC_DATETIME

config RTC_HIRES
	bool "Hi-Res RTC Support"
```
We can use two types of RTCs like above.
@kaikaikaime 
You mean RTL8730E is RTC_DATETIME and RTL8721CSM is RTC_HIRES ?

kaikaikaime:
Dear Chang &  Abhishek, 

Both RTL8730E & RTL8721CSM used datetime format only. The difference is that RTL8730E has a additional field for year thus imposing additional restriction.
Both RTL8730E & RTL8721CSM are not RTC_HIRES.

Regards,
Chuen Kai

sunghan-chang:
@abhishek-samsung @kaikaikaime We can't get the board config at app side which uses RTC. Then
1. introduce new common config for year. But it looks strange.
2. just add year value when we call the RTC api for all boards. Then RTL8721CSM is ok? It just ignore year value?
3. Any other suggestion?

kaikaikaime:
Dear Chang & Abhishek,

This is the conclusion I gathered:
1. After Turning on Device Drivers->"RTC Device Support".
mktime API used to process rtc_time struct in lib_mktime.c
value of 99 (1999) tm_year automatically will be added. 
Both RTL8730E & RTL8721CSM can run. 

2. In addition of using built-in libraries-> "Localtime API call support"
mktime API used to process rtc_time struct in lib_localtime.c
tm_year is 0, due to memset 0 from rtc_main.c
Both RTL8730E & RTL8721CSM will **not** able to start. 
Only by adding 1970-1-1 in rtc_main.c, both RTL8730E & RTL8721CSM will be able to run normally.

3. RTL8721CSM does not ignore the year value, as it still counts value of leap year etc or lib_mktime was used previously (99 was added).

Regards,
Chuen Kai

sunghan-chang:
@r-prabu Please check this.

sunghan-chang:
@kaikaikaime Will merge @abhishek-samsung 's PR #6242 . Let's close this. But rtc read without local time config looks wrong. @abhishek-samsung will send you a mail. Please check it.

PR #6216, opened 2024-05-31
os/binfmt, os/tools: support binary signing with xipelf

abhishek-samsung:
support xipelf binary loading with binary signing

### DIFF
--- a/os/tools/mkldscript.py
+++ b/os/tools/mkldscript.py
@@ -98,25 +98,32 @@ app2_ram_str = hex(ram_offset) + str1 + hex(int(CONFIG_APP2_BIN_DYN_RAMSIZE)) + 
 
 # common at the end of the region
 
+CONFIG_USER_SIGN_PREPEND_SIZE = util.get_value_from_file(cfg_file, "CONFIG_USER_SIGN_PREPEND_SIZE=").rstrip('\n')
+
+if CONFIG_USER_SIGN_PREPEND_SIZE == 'None' :
+    signing_offset = 0
+else :
+    signing_offset = int(CONFIG_USER_SIGN_PREPEND_SIZE)
+
 for name in NAME_LIST :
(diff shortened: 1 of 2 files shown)


os/tools/mkldscript.py
```diff
@@ -98,25 +98,32 @@
 
 # common at the end of the region
 
+CONFIG_USER_SIGN_PREPEND_SIZE = util.get_value_from_file(cfg_file, "CONFIG_USER_SIGN_PREPEND_SIZE=").rstrip('\n')
+
+if CONFIG_USER_SIGN_PREPEND_SIZE == 'None' :
+    signing_offset = 0
+else :
+    signing_offset = int(CONFIG_USER_SIGN_PREPEND_SIZE)
```
sunghan-chang, line 106: The xipelf has the same compensation. Is it possible to do at here or there only?
If we do that here, you just can refer some global variable like _sbss.

abhishek-samsung: We need to mention the address in linkerscript from where we can place the data. So, if partition starts from 0x0ea00000 then we can only place the code and data starting from 0x0ea00000 + signing size + header size. So we cant skip it here.

And then during loading we need to read the userspace object which is located at an offset of (signing size + header size). I am not sure how we can do it at only one place.

abhishek-samsung: also I think I didnt understand your question. Could you give some example?

abhishek-samsung: The other alternative way is to add this compensation to the offset we get from the realtek script. In that case we don't need to make changes here. However, Since the realtek script is board specific and mkldscript.py is common script, its better to include signing changes in common script as board has nothing to do with it. Also, this way, the flow stays similar to normal elf loading.

sunghan-chang (approved):

PR #6218, opened 2024-06-03
os/drivers/lcd : update spi frequency to maximum 50 MHz

anjana348:
Update spi frequency to 50MHz from 48MHz. Removed unnecessary code which was using slowed down api after DMA support

### DIFF
--- a/os/drivers/lcd/st7789.c
+++ b/os/drivers/lcd/st7789.c
@@ -80,7 +80,7 @@
 /* SPI frequency */
 
 #ifndef CONFIG_LCD_ST7789_FREQUENCY
-#define CONFIG_LCD_ST7789_FREQUENCY 48000000
+#define CONFIG_LCD_ST7789_FREQUENCY 50000000
 #endif
 
 /* Check contrast selection */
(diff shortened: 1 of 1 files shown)


os/drivers/lcd/st7789.c
```diff
@@ -547,37 +547,6 @@ static void st7789_rdram(FAR struct st7789_dev_s *dev, FAR uint8_t *buff, size_t
 	SPI_SELECT(dev->spi, SPIDEV_DISPLAY, false);
 }
 
-/****************************************************************************
```
sunghan-chang, line 550: Let's split commit for this. And could you write more detail in commit description for this?

os/drivers/lcd/st7789.c
```diff
@@ -799,9 +768,6 @@ static int st7789_init(FAR struct lcd_dev_s *dev)
 
 	st7789_setorientation(dev, ST7789_DEFAULT_ORIENTATION);
 
-	lcd_lock(priv->spi);
```
sunghan-chang, line 802: No need when DMA is disabled?

anjana348: yes even when DMA is disabled, it is not needed. It was just to fill initial garbage data with color before any app runs. It can be avoided if app itself turns on lcd or backlight when app is ready to show some valid frames.

sunghan-chang (approved):

PR #6219, opened 2024-06-03
fs/mtd: Add partition number to partition procfs data

jeongarmy:
Add partition number to partition procfs data.
It is read by "/proc/partitions"


os/fs/driver/mtd/mtd_partition.c
```diff
@@ -869,4 +869,4 @@ int mtd_setpartitionname(FAR struct mtd_dev_s *mtd, FAR const char *name)
 	strncpy(priv->name, name, MTD_PARTNAME_LEN);
 	return OK;
 }
-#endif
+#endif
```
sunghan-chang, line 872: Let's add the line end.

sunghan-chang (approved):

PR #6222, opened 2024-06-05
os/tools: Check package size with both internal and external flashes

jeongarmy:
This script compares a package size with partition size and show the ratio of usage.
It checked internal partitions so I added external checking if external flash exist.


sunghan-chang:
@jeongarmy Please add test results in the commit description for internal flash only and for internal + external flash.

sunghan-chang (dismissed):

sunghan-chang (approved):

PR #6223, opened 2024-06-05
pm: clean up

ewoodev:
pm_test is not used and does not work properly.
And pm_test is not suitable for the current pm structure, and the operation for using user api is not covered.

Therefore, Remove un-used pm_test


edwakuwaku:
Hi @ewoodev , you should remove the PM_TEST from os/pm/Kconfig as well. And all the defconfig which has CONFIG_PM_TEST should be removed. Please also check os/pm/Makefile and remove the macro build.

gSahitya-samsung (approved):
The PM code clean changes looks good to me

jeongarmy (approved):
lgtm

PR #6227, opened 2024-06-06
os/board/rtl8720e&rtl8721csm&rtl8730e: Flash Protection

ZhenBei-Sin:
os/board/rtl8720e&rtl8721csm&rtl8730e: Modify defconfig for Flash Protection

Change Note:
- make the flash status bit for enable flash protection as config
- define the FLASH_STATUS_BITS to CONFIG_FLASH_STATUS_BITS
- add new function use for verify the Flash Protection Bit is enabled|

apps/examples/board_specific_demo: Enhance the Secure Storage Area example

Change Notes:
- Enhance the verification process to ensure the protection of the Secure Storage Area.
- Verify the Secure Storage area is protected and cannot be erased.
- Sequence of operations includes Secure Storage read, flash read, Secure Storage write, Secure Storage read, flash read.
- Identification of data exists area, retrieval flash status bit, erase existing data area, and verification of protection status.


ZhenBei-Sin:
As previously discussed at the [PR](https://github.com/Samsung/TizenRT/pull/6111).
Enhance for the sstorage example.

Attached verification:
image
[rtl8730e_FlashNonProtect.log](https://github.com/user-attachments/files/15614626/rtl8730e_FlashNonProtect.log)
[rtl8730e_FlashProtected.log](https://github.com/user-attachments/files/15614623/rtl8730e_FlashProtected.log)

image
[rtl8721csm_FlashProtected.log](https://github.com/user-attachments/files/15614720/rtl8721csm_FlashProtected.log)
[rtl8721csm_FlashNonProtect.log](https://github.com/user-attachments/files/15614721/rtl8721csm_FlashNonProtect.log)

os/board/rtl8720e/src/component/filesystem/ftl/ftl.c
```diff
@@ -122,8 +122,6 @@ extern bool ftl_page_erase(struct Page_T *p);
 void ftl_mapping_table_init(void);
 uint16_t read_mapping_table(uint16_t logical_addr);
 
-/* Flash Status Bit */
-#define FLASH_STATUS_BITS 0x2c
```
ziyik, line 126: To reduce the changes maybe you can just change the value to CONFIG_FLASH_STATUS_BITS .

os/board/rtl8721csm/src/component/common/file_system/ftl/ftl.c
```diff
@@ -138,8 +138,6 @@ typedef struct {
 } ns_ftl_func_struc;
 #endif
 
-/* Flash Status Bit */
-#define FLASH_STATUS_BITS 0x2c
```
ziyik, line 142: Same comment as rtl8720e

os/board/rtl8730e/src/component/file_system/ftl/ftl.c
```diff
@@ -131,8 +131,6 @@ uint16_t read_mapping_table(uint16_t logical_addr);
 #define FTL_KEY_ADDR 0x390
 #endif
 
-/* Flash Status Bit */
-#define FLASH_STATUS_BITS 0x28 //Protect Lower 512KB
```
ziyik, line 135: Same comment as rtl8720e

os/board/rtl8720e/Kconfig
```diff
@@ -19,6 +19,13 @@ config FLASH_SIZE
         A size of flash (in bytes).
         This is fixed value, so user doesn't need to change it.
 
+config FLASH_STATUS_BITS
+	hex "Flash Status Register Bits"
+	default 0x2c
+	---help---
+		Setting this value according to the IC package's Flash Model.
+		To protect the lower 512KB of the flash memory.
```
sunghan-chang, line 27: Is this configurable? If not, let's remove `"Flash Status Register Bits"`. (Then, user can't change the value)

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -97,6 +89,28 @@ int rtl_ss_flash_read(uint32_t address, uint32_t len, uint8_t *data, int en_disp
...
+
+	status_register = 0xFF & flash_get_status(&flash);
+	printf("Status Register: 0x%02x\n", status_register);
+
+	if ((status_register & FLASH_STATUS_BITS) == FLASH_STATUS_BITS) {
+		return 1;
+	} else {
+		return 0;
+	}
+}
```
sunghan-chang, line 112: Internal flash size is 8MB, but we protect only first 512KB. Even this situation, can we say that a flash is protected?
Is there anyway where we protect? range?

os/se/ameba/rtl_security_api_wrapper_tz.c
```diff
@@ -22,6 +22,9 @@
 #include "flash_api.h"
 #include "device_lock.h"
 
+/* Flash Status Bit to protect lower 512KB */
```
sunghan-chang, line 25: lower 512KB is fixed?

sunghan-chang (commented):
Let's split 1st commit for os change and config change.

os/se/ameba/rtl_security_api_wrapper_tz.c
ZhenBei-Sin: Yes, it has been fixed.

Refer to the flash layout:
image

The lower 512KB flash protected area contains the bootloader, FTL, and SS.
After the MP process, the protected area cannot be changed, and only the user image can be updated.

os/board/rtl8720e/Kconfig
ZhenBei-Sin: It was NOT configurable. I will remove the "Flash Status Register Bits".
Thank You.

os/se/ameba/rtl_security_api_wrapper_tz.c
ZhenBei-Sin: Flash Protection as commented at below:
For Samsung security concern, we only focus on protect the lower 512KB.

This API is to checking whether the flash register bit was enabled.
For protect how much size of flash it was need refer to the flash datasheet.

ZhenBei-Sin:
> Let's split 1st commit for os change and config change.

The os part change is located at the ftl.c file, defining a macro and will be replaced by CONFIG_xxx
#define FLASH_STATUS_BITS CONFIG_FLASH_STATUS_BITS

If we separate it into 2 commits, CONFIG_FLASH_STATUS_BITS will be undefined and leading to a compilation error. 
Should we proceed with this separation?

sunghan-chang:
@ZhenBei-Sin You mean that when we split it to 2 commits and we get just one commit for os without config, it will cause build break. Is that right what you mean? If right, yes. Even it causes, let's split them.

ZhenBei-Sin:
> @ZhenBei-Sin You mean that when we split it to 2 commits and we get just one commit for os without config, it will cause build break. Is that right what you mean? If right, yes. Even it causes, let's split them.

Hi @sunghan-chang 
my mistake.. it should not be affect.
defconfig add the config name -> os part with remain the config name (1st commit)
os part change the name (2nd commit)

Done separated the commit.
For the flash protection 512KB, may I know still have any unclear part?

Thank you.

sunghan-chang (approved):

PR #6232, opened 2024-06-13
os/arch/arm/src/amebasmart: Remap timer index for app usage

edwakuwaku:
- Normally, app uses lowest index for HW timer, such as /dev/timer0. But, currently the kernel and driver has allocated timer 0~2 for several usage:
- timer0: LP systimer
- timer1: PM one-shot wakeup timer
- timer2: Crash detector for KM4/KM0 cpu status
- Due to above reason, we have to remap the timer index, so that when app is using lower index timer, it will not cause conflict with the timer which already has designated usage.
- For example, if app use /dev/timer0, it actually uses HW timer 3, and the rules follow by incrementing the index


sunghan-chang:
@edwakuwaku Can't we reassign higher timers for current usages instead of remapping?

sunghan-chang (approved):

PR #6236, opened 2024-06-17
os/tools, os/Makefile.unix: verify whether l2 page table per app are sufficient or not

neel-samsung:
Check whether L2 page table assigned per app are sufficient enough or not. We have to throw an error and stop further execution if they are not sufficient enough and print the exact amount of L2 page table required per app.

### DIFF
--- a/os/Makefile.unix
+++ b/os/Makefile.unix
@@ -476,7 +476,7 @@ ifeq ($(CONFIG_RAW_BINARY),y)
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O binary $(BIN) $(BIN).bin
 endif
 
-# Verify app binary to make sure there are no undefined symbols
+# Verify app binary to make sure there are no undefined symbols and to check the l2 pages per app are sufficient enough
 define VERIFY_APP
 	$(Q) echo "Verify $(OUTBIN_DIR)/$1"
 	$(Q) if [ "`nm -u $(OUTBIN_DIR)/$1 | wc -l`" != "0" ]; then \
@@ -485,6 +485,7 @@ define VERIFY_APP
 		rm $(OUTBIN_DIR)/$1; \
 		exit 1; \
 	fi
+	$(Q) $(if $(CONFIG_ARCH_USE_MMU), $(TOPDIR)/tools/mkverifyappsize.py $(OUTBIN_DIR)/$1 $2,)
 endef
 
 # Strip debug symbols from binary if CONFIG_ELF_EXCLUDE_SYMBOLS is enabled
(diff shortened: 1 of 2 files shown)


os/Makefile.unix
```diff
@@ -485,6 +485,7 @@ define VERIFY_APP
 		rm $(OUTBIN_DIR)/$1; \
 		exit 1; \
 	fi
+	$(Q) $(TOPDIR)/tools/mkverifyappsize.py $(OUTBIN_DIR)/$1 $2
```
kishore-sn, line 488: This must be done only in case of chipsets which support MMU. Please check for mmu config here.

os/tools/mkverifyappsize.py
```diff
@@ -0,0 +1,77 @@
...
+l2_pg_need = 0
+os.system('size ' + file_path + ' > ' + APP_SIZE_FILE)
+app_data = open(APP_SIZE_FILE, 'rb')
+line1 = app_data.readline().split()
+line2 = app_data.readline().split()
+size = dict(zip(line1,line2))
+if util.check_config_existence(cfg_path, 'CONFIG_XIP_ELF') == True :
+    textsize = int(size["text"])
+    datasize = int(size["data"])
+    bsssize = int(size["bss"])
```
kishore-sn, line 57: We can move the above 3 lines out of the if statement.

os/tools/mkverifyappsize.py
```diff
@@ -0,0 +1,77 @@
...
+datasize = 0
+bsssize = 0
+l2_pg_per_app = int(util.get_value_from_file(cfg_path, "CONFIG_NUM_L2_PER_APP=").replace('"','').replace('\n',''))
+l2_pg_need = 0
+os.system('size ' + file_path + ' > ' + APP_SIZE_FILE)
+app_data = open(APP_SIZE_FILE, 'rb')
+line1 = app_data.readline().split()
+line2 = app_data.readline().split()
+size = dict(zip(line1,line2))
+if util.check_config_existence(cfg_path, 'CONFIG_XIP_ELF') == True :
```
kishore-sn, line 54: Can you please add comment as to why we need to calculate it differently in case of xip elf ? (flash and ram address are different)

os/Makefile.unix
neel-samsung: I have added if statement for config check.

os/tools/mkverifyappsize.py
neel-samsung: ok sure.

os/tools/mkverifyappsize.py
neel-samsung: I have added comment for xip elf case (calculating l2 page table). Please check.

os/Makefile.unix
```diff
@@ -476,7 +476,8 @@ ifeq ($(CONFIG_RAW_BINARY),y)
 	$(Q) $(OBJCOPY) $(OBJCOPYARGS) -O binary $(BIN) $(BIN).bin
 endif
 
-# Verify app binary to make sure there are no undefined symbols
+# Verify app binary to make sure there are no undefined symbols and to check the l2 pages per app are sufficient enough
+ifeq ($(CONFIG_ARCH_HAVE_MMU),y)
```
kishore-sn, line 480: Instead of adding condition here, you can add condition just for calling the mkverifyappsize.py.
Also, you need to check for CONFIG_ARCH_USE_MMU... not HAVE_MMU.

sunghan-chang:
@kishore-sn Once review is done, please leave an approval.

neel-samsung:
@kishore-sn Please review it and merge the PR, as it looks ready to merge.

kishore-sn (approved):

sunghan-chang (approved):

PR #6237, opened 2024-06-18
add performance test for internal & external flash for Read, write & …

r-prabu:
Add performance test for internal & external flash for Read, Write & Erase.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/prabu.r/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/prabu.r/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

Memory | Read_4K /1MB | Write_4K / 1MB | Erase_4K / 1MB | Remarks
-- | -- | -- | -- | --
Internal | < 1 ms / 29 ms | 7 ms / 1713 ms | 41 ms / 10567 ms | Internal   Flash Erase takes more time comared to External flash
External | 6 ms / 1496 ms | 20 ms / 3008 ms | 6 ms / 8572 ms |  
  |   |   |   |  
Inter_DMA | < 1 ms / 29 ms | 6 ms  / 1715ms | 40 ms /10598 ms | No   impact of DMA enable on internal flash
Ext_DMA | 2 ms/ 569 ms | 22 ms / 3620 ms | 6 ms / 8548 ms | DMA        - Flash read has improved.     - Flash Write time increased compared to without DMA

</body>

</html>


r-prabu:
@edwakuwaku DMA does has some improvement but only in the External flash read. 
Also i'm not sure why write performance is decreased. Is DMA not feasible for the internal flash?

r-prabu:
> Hi @r-prabu , for RTL8730E, there is an SPIC designated for internal flash operation. So, I don't think you need to test internal flash performance individually. For DMA write to ext flash, it should have better performance than polling. We might need to verify the performance of such test again.

 > @edwakuwaku Can you clarify little more on the SPI DMA enable. Will it enable the DMA for all the SPI interfaces?

sunghan-chang (commented):
This uses many kernel internal functions, this will not run on the loadable.
I think we can use bch to write and to read a flash without fs.
@jeongarmy Could you confirm this?

apps/system/flash_test/flash_test.c
```diff
@@ -85,16 +101,53 @@ static int flash_read(unsigned long addr, int size)
...
+			return ERROR;
+		}
+#elif defined(CONFIG_MTD_W25)
+		dev_mtd = w25_initialize(spi);
+		if (dev_mtd == NULL) {
+			lldbg("w25 Init failed\n");
+			return ERROR;
+		}
+		printf("w25_initialize: done for secondary flash \n");
+#endif
```
sunghan-chang, line 138: This kind of code should be done in kernel, not here. And they are done already. 
As I left a comment before, let's use bch to read / write some data from/to a flash.

sunghan-chang (commented):
@r-prabu What we should do is performance comparison between internal and external.
So, up_xxxxx, w25_xxxx, jedec_xxx should not be used here. Kernel already does.
Let's use bch to access a flash.

r-prabu:
ok. 
@sunghan-chang , Tested with bch for the flash performance. Getting similar values for read, slight increase for the write.  If we use the bch then will not be able to get the erase info separately. 
Also write data time is (erase + write). Can we have a separate app for this testing or update in this app itself?

sunghan-chang:
> ok.
@sunghan-chang , Tested with bch for the flash performance. Getting similar values for read, slight increase for the write. If we use the bch then will not be able to get the erase info separately.
Also write data time is (erase + write). Can we have a separate app for this testing or update in this app itself?

@Taejun-Kwon Any suggestion on erase?

sunghan-chang:
@r-prabu TJ said there is an ioctl command for erase, it can be executed using bch. Could you refer the format command?

sunghan-chang:
@r-prabu Let's finishi this by this week. Could you change this to use bch and rtc?

r-prabu:
> @r-prabu TJ said there is an ioctl command for erase, it can be executed using bch. Could you refer the format command?

> yes i could use the ioctl to erase.

r-prabu:
> @r-prabu What we should do is performance comparison between internal and external. So, up_xxxxx, w25_xxxx, jedec_xxx should not be used here. Kernel already does. Let's use bch to access a flash.

> @sunghan-chang, to check the flash performance it better to check like as it is implemented in the flash_test, for this we need to have above in the app layer to get the reference.  
For bch we dont need it. So i have updated the app to have both , 1.Direct flash access  (read, write, erase) 2. Using BCH

r-prabu:
> @r-prabu Let's finishi this by this week. Could you change this to use bch and rtc?

> @sunghan-chang, Added commands to use bch as well. We can close this without rtc. Meanwhile i will check for using some other alternate high resolution timer.

apps/system/flash_test/flash_test.c
```diff
@@ -84,18 +351,55 @@ static int flash_read(unsigned long addr, int size)
...
-	dev_mtd = up_flashinitialize();
-	if (!dev_mtd) {
-		printf("up_flashinitialize Failed\n");
-		return ERROR;
+	}
+#ifdef CONFIG_SECOND_FLASH_PARTITION
+
+	else if (flash == FLASH_EXT) {
+		printf("Received Flash Read cmd for external flash offset 0x%x, read size %d\n", addr, size);
+		struct spi_dev_s *spi = up_spiinitialize(1);
```
edwakuwaku, line 374: Will this be callable using loadable defconfig? It should get a compile error.

apps/system/flash_test/flash_test.c
```diff
@@ -84,18 +351,55 @@ static int flash_read(unsigned long addr, int size)
...
+	long long total_duration_ms = 0;
+	systick_time_ms = 0;
 
 	if (addr == 0 || size <= 0) {
 		printf("Invalid arguments, offset 0x%x, read size %d\n", addr, size);
 		return ERROR;
 	}
+	if (flash == FLASH_INTERNAL) {
+		printf("Received Flash Read cmd for internal flash offset 0x%x, read size %d\n", addr, size);
+		dev_mtd = up_flashinitialize();
```
edwakuwaku, line 363: Please check below comment

edwakuwaku:
I don't think this test case can work on loadable, is this meant to be run on flat defconfig only? Could you please check?

r-prabu:
> I don't think this test case can work on loadable, is this meant to be run on flat defconfig only? Could you please check?

> @edwakuwaku yes currently we are testing only on the flat build only.

r-prabu:
@edwakuwaku looks device is hanging during the external flash erase with DMA enable. can you check this behavior on your side?

edwakuwaku:
Hi @r-prabu , we will check with running your test example. But, please also provide 2 things to us:
1) Your .config file, we need to know which additional config is enabled. you can share it via email.
2) The test logs of facing hang issue.
Also, please share the reproduction rate. Is it reproducible every time?

> Hi @r-prabu , you constrained the buffer size to 4KB, that could lead to performance drop. Please check comment here: [#6154 (comment)](https://github.com/Samsung/TizenRT/pull/6154#discussion_r1577420423) You could just provide whole buffer size, our driver will handle the slicing.

Also as mentioned in the previous comment, you need not constraint the buffer size. But as I checked, you still limit the buffer size to 4K. May I know what is the reason to do so?
Please correct if my understanding is wrong.

apps/system/flash_test/flash_test.c
r-prabu: using the flash driver only for flat build update the app accordingly

apps/system/flash_test/flash_test.c
r-prabu: using the flash driver for the flat app config. update the app accordingly

apps/system/flash_test/flash_test.c
r-prabu: Direct flash access used for flat config only. update the app accordingly.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+	//sys_end_time = TICK2MSEC(clock_systimer());
+	sys_end_time = get_time();
+	printf("\nSystick time taken for Flash Write %ld(ms)\n", (sys_end_time - sys_start_time));
+
+	printf("\n");
+	free(buffer);
+	return OK;
+}
+
+static int flash_erase(unsigned long addr, int size, flash_t flash)
```
sunghan-chang, line 549: This is the function to test flash by accessing mtd directly. Let's rename the function with mtd.

apps/system/flash_test/flash_test.c
```diff
@@ -143,19 +633,68 @@ int flash_test_main(int argc, char *argv[])
 #endif
 {
 	off_t offset;
-	int read_size;
+	int size;
+	flash_t flash_type;
 
-	if (argc == 4 && !strncmp(argv[1], READ_CMD, sizeof(READ_CMD) + 1)) {
+	if ((argc == 4) && (!strncmp(argv[1], BCDEV_CMD, sizeof(BCDEV_CMD) + 1))) {
+		flash_type = (flash_t) atoi(argv[3]);
```
sunghan-chang, line 640: Let's remove a space after typecasting. There are several lines for the same.

kishore-sn (dismissed):

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+		dev_mtd = w25_initialize(spi);
+		if (dev_mtd == NULL) {
+			lldbg("w25 Init failed\n");
+			return ERROR;
+		}
+#endif
+	}
+#endif
+	/* Get the geometry of the FLASH device */
+	ret = dev_mtd->ioctl(dev_mtd, MTDIOC_GEOMETRY, (unsigned long)((uintptr_t) & geo));
```
sunghan-chang, line 589: (uintptr_t)&geo

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+		ret = MTD_WRITE(dev_mtd, addr + offset, write_size, (const uint8_t *)buffer);
+		if (ret <= 0) {
+			printf("Write Failed : %d\n", ret);
+			free(buffer);
+			return ERROR;
+		}
+		offset += ret;
+		remaining -= ret;
+	}
+	lldbg("lldb: Flash Write End\n");
```
sunghan-chang, line 539: Is there any reason to be lldbg, not printf? We should not use lldbg at normal code.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+long sys_end_time = 0;
+long systick_time_ms = 0;
+
+char blockname[32];
+
+static inline long get_time(void)
+{
+	struct timeval tv;
+
+	gettimeofday(&tv, NULL);
```
sunghan-chang, line 110: Let's use RTC instead of system tick. Tick is much inaccurate.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+};
+typedef enum flash_e flash_t;
+
+FAR struct mtd_geometry_s geo;
+
+long sys_start_time = 0;
+long sys_end_time = 0;
+long systick_time_ms = 0;
+
+char blockname[32];
```
sunghan-chang, line 104: Let's add `static`

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,570 @@
...
+#define ERASE_CMD   "erase"
+#define BCDEV_CMD	"bch"
+
+#ifdef CONFIG_BUILD_FLAT
+#define INT_FLASH_PART 7
+#define EXT_FLASH_PART 13
+#else
+#define INT_FLASH_PART 11
+#define EXT_FLASH_PART 17
+#endif
```
sunghan-chang, line 86: The numbers mean?

sunghan-chang (commented):
@r-prabu As you know, we don't allow empty commit description. Please add some explanation of the commit.

apps/system/flash_test/flash_test.c
r-prabu: updated the function as mtd_read, mtd_write & mtd_erase.

apps/system/flash_test/flash_test.c
r-prabu: updated

apps/system/flash_test/flash_test.c
r-prabu: nothing specific on using updated it to use printf.

apps/system/flash_test/flash_test.c
r-prabu: Regarding the RTC , currently it is only seconds resolution. Also as i mentioned before , averaging may not be good idea. since RTC is in seconds resolution we might need to run the tests for minimum number of times about (1000 times) which will make us to o wait for hours :(

apps/system/flash_test/flash_test.c
r-prabu: done

apps/system/flash_test/flash_test.c
r-prabu: this is the partition number which will be used during the mtd interaction . currently i have fixed to this partition. If we want then we can get as an input from the user itself.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,553 @@
...
+
+	/* Now write the offset into every block */
+	printf("Writing to media: buff Size %d, times %d\n", BUF_SIZE, (SIZE_1MB/BUF_SIZE));
+
+	offset = 0;
+	/* Fill the block with the offset */
+	fill_buffer(buffer);
+
+	sys_start_time = get_time();
+	printf("lldb: BCH Write start\n");
```
sunghan-chang, line 195: What's the meaning of `lldb`?

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,553 @@
...
+		g_partno = INT_FLASH_PART;
+	} else if (flash == FLASH_EXT) {
+		g_partno = EXT_FLASH_PART;
+	}
+	/* Allocate a buffer */
+	buffer = (FAR uint32_t *)malloc(BUF_SIZE);
+
+	if (!buffer) {
+		printf("ERROR: failed to allocate a sector buffer\n");
+		fflush(stdout);
```
sunghan-chang, line 174: Is this necessary?

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,553 @@
...
+	sys_end_time = get_time();
+	printf("\nSystick time taken for Flash bch Write %ld(ms)\n", (sys_end_time - sys_start_time));
+	goto out;
+out_free_buffer:
+	free(buffer);
+out_error:
+	return ERROR;
+out:
+	free(buffer);
+	return OK;
```
sunghan-chang, line 218:
```
	int ret = ERROR;

	if (!buffer) {
		printf("ERROR: failed to allocate a sector buffer\n");
		return ret;
	}

	if (fd < 0) {
		printf("ERROR: open /dev/mtd%d failed: %d\n", g_partno, errno);
		goto out_free_buffer;
	}

		if (nbytes < 0) {
			printf("ERROR: write to /dev/mtd%d failed: %d\n", g_partno, errno);
			goto out_close_fd;
		}

	printf("lldb: BCH Write End\n");
	sys_end_time = get_time();
	printf("\nSystick time taken for Flash bch Write %ld(ms)\n", (sys_end_time - sys_start_time));
	ret = OK;

out_close_fd:
	close(fd);
out_free_buffer:
	free(buffer);
	return ret;
}
```

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,553 @@
...
+static int mtd_read(unsigned long addr, int size, flash_t flash)
 {
 	int ret;
 	int offset;
 	int buf_idx;
 	int read_size;
 	int remaining;
-	FAR uint8_t *buffer;
+	FAR uint32_t *buffer;
+	FAR uint8_t *rd_buffer;
```
sunghan-chang, line 348: I understand what `rd_buffer` is from the name. But I don't know what `buffer` is. Could you rename it?

apps/system/flash_test/flash_test.c
r-prabu: updated the code to scan the specific partition ("flash_type") & use it for the testing.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,553 @@
...
+#include <tinyara/config.h>
+#include <tinyara/fs/ioctl.h>
+#include <fcntl.h>
 /****************************************************************************
  * Definitions
  ****************************************************************************/
-#define BUF_SIZE   1024
-
+#define BUF_SIZE   65536
+#define SIZE_1MB   1024*1024
```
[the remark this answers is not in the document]
r-prabu: updated

apps/system/flash_test/flash_test.c
r-prabu: mybad :( corrected...

apps/system/flash_test/flash_test.c
r-prabu: updated the code control flow :)

apps/system/flash_test/flash_test.c
r-prabu: rename the buffer -> wr_buffer.

apps/system/flash_test/flash_test.c
r-prabu: not sure. extended the existing usage. anyway removed this.

apps/system/flash_test/flash_test.c
```diff
@@ -59,80 +59,641 @@
...
  * Definitions
  ****************************************************************************/
-#define BUF_SIZE   1024
+#define BUF_SIZE   65536
+#define SIZE_1MB   1048576
 
 #define READ_CMD    "read"
+#define WRITE_CMD   "write"
+#define ERASE_CMD   "erase"
+#define BCDEV_CMD	"bch"
```
gSahitya-samsung, line 79: Please check indentation

gSahitya-samsung, line 79: Please check indentation

PR #6243, opened 2024-06-19
os/arch/arm/src/amebasmart: Add refs count in SPI to prevent from multiple initialization

edwakuwaku:
- Add refs count in SPI to prevent from multiple initialization


os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -1588,6 +1591,10 @@ FAR struct spi_dev_s *up_spiinitialize(int port)
 
 		priv = &g_spi0dev;
 
+		if (priv->refs > 0) {
+			dbg("SPI port%d has been initialized before!\n", port);
+			return (FAR struct spi_dev_s *)priv;
+		}
```
kishore-sn, line 1597: Can you please add a check for CONFIG_AUDIO_NDP120 not defined while initializing SPI0 in board_spi_initialize API?

sunghan-chang (approved):

PR #6247, opened 2024-06-21
os/arch/arm/src/amebasmart: Revise BSP to follow PM new structure

edwakuwaku:
- pm_register is removed from pm driver, changed to domain registration according to different driver module
- Old way: Register callback to PM driver, every PM state change will inform driver, some driver status will be checked to determine whether a PM state change is given approval
- Revised way: Register domain to PM driver, suspend/resume domain according to operation in different driver module

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
@@ -50,7 +50,8 @@
 #include "osdep_service.h"
 #include "timer_api.h"
 
-gtimer_t g_timer1;
+static gtimer_t g_timer1;
+static int g_bsp_domain_id[BSP_DOMAIN_MAX];
 
 void SOCPS_SetAPWakeEvent_MSK0(u32 Option, u32 NewStatus)
 {
@@ -112,19 +113,41 @@ int SOCPS_AONWakeReason(void)
 	return reason;
 }
 
-void pg_timer_int_handler(void *Data)
+void up_set_pm_timer(unsigned int interval_us) 
 {
-	pmvdbg("PM Timer interrupt handler!!\n");
-}
-
-void up_set_pm_timer(unsigned int interval_us) {
 	// Check whether timer interrupt need to be set
 	if (interval_us > 0) {
 		gtimer_init(&g_timer1, TIMER1);
 		/* Pass in timer obj to avoid compile warning, the last argument will not be used in the callback handler */
-		gtimer_start_one_shout(&g_timer1, interval_us, (void *)pg_timer_int_handler, (uint32_t)&g_timer1);
+		gtimer_start_one_shout(&g_timer1, interval_us, NULL, (uint32_t)&g_timer1);
 	}
 	return;
+}
+
+void bsp_pm_domain_register(char *domain_name, int bsp_drv_id)
+{
+	int domain_id = pm_domain_register(domain_name);
+	if (domain_id < 0) {
+		pmdbg("Unable to register %s DOMAIN\n", domain_name);
+	} else {
+		g_bsp_domain_id[bsp_drv_id] = domain_id;
+	}
+}
+
+void bsp_pm_domain_control(int bsp_drv_id, bool is_suspend)
+{
+	/* Retrive the domain id from the bsp driver id */
+	int domain_id = g_bsp_domain_id[bsp_drv_id];
+	if (is_suspend) {
+		/* Check BSP_DRV_DOMAIN enum for more info */
+		if (pm_suspend(domain_id) != OK) {
+			pmdbg("Unable to suspend bsp_drv_id: %d!\n", bsp_drv_id);
+		}
+	} else {
+		if (pm_resume(domain_id) != OK) {
+			pmdbg("Unable to resume bsp_drv_id: %d!\n", bsp_drv_id);
+		}
+	}
 }
 
 /* Interrupt callback from wifi-keepalive, which LP received designated packet*/
--- a/os/arch/arm/src/amebasmart/amebasmart_serial.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_serial.c
@@ -264,13 +264,6 @@ static void rtl8730e_up_txint(struct uart_dev_s *dev, bool enable);
 static bool rtl8730e_up_txready(struct uart_dev_s *dev);
 static bool rtl8730e_up_txempty(struct uart_dev_s *dev);
 
-#ifdef CONFIG_PM
-static void amebasmart_serial_pmnotify(FAR struct pm_callback_s *cb,
-										enum pm_state_e pmstate);
-static int  amebasmart_serial_pmprepare(FAR struct pm_callback_s *cb,
-										enum pm_state_e pmstate);
-#endif
-
 /****************************************************************************
  * Private Data
  ****************************************************************************/
@@ -513,18 +506,6 @@ static uart_dev_t g_uart4port = {
 	.ops = &log_g_uart_ops,
 	.priv = &g_uart4priv,
 };
-#endif
-
-#ifdef CONFIG_PM
-static struct
-{
-	struct pm_callback_s pm_cb;
-} g_serialpm =
-	{
-		.pm_cb.name = "rtl8730e_serial",
-		.pm_cb.notify  = amebasmart_serial_pmnotify,
-		.pm_cb.prepare = amebasmart_serial_pmprepare,
-	};
 #endif
 
 /****************************************************************************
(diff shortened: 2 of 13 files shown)


os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -112,21 +112,32 @@ int SOCPS_AONWakeReason(void)
...
-		gtimer_start_one_shout(&g_timer1, interval_us, (void *)pg_timer_int_handler, (uint32_t)&g_timer1);
+		gtimer_start_one_shout(&g_timer1, interval_us, NULL, (uint32_t)&g_timer1);
 	}
 	return;
 }
 
+void bsp_pm_domain_control(char *domain, bool is_suspend)
+{
+	/* Retrive the domain id from the domain string */
+	int domain_id = pm_domain_register(domain);
```
ritesh55555, line 129: i think its better if we create a  pm_get_domainId(domain_name) for this? , @gSahitya-samsung

ritesh55555:
hi @edwakuwaku , there are some comments related to this PR in my another PR (https://github.com/Samsung/TizenRT/pull/6259) , which i have taken patch from here , 
Can you check it?

edwakuwaku:
> hi @edwakuwaku , there are some comments related to this PR in my another PR (#6259) , which i have taken patch from here , Can you check it?

Hi @ritesh55555 , as I have discussed with @ewoodev , he suggested another way such as store domain id for each peripheral driver, so that we don't need to invoke pm_domain_register() every time. It is not a logic error.
But, I think current implementation is also fine. Since the interface is able for us to retrieve the domain id according to each driver.

ewoodev (approved):

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -746,8 +726,16 @@ static inline void amebasmart_spi_master_set_delays(FAR struct amebasmart_spidev
 static int amebasmart_spi_lock(FAR struct spi_dev_s *dev, bool lock)
 {
 	FAR struct amebasmart_spidev_s *priv = (FAR struct amebasmart_spidev_s *)dev;
-
+	/* When a lock is invoked, means that we are going to do some SPI operation,
+	   thus we should decide to suspend/resume SPI domain here
+	*/
+#ifdef CONFIG_PM
+	int domain = pm_domain_register("SPI");
+#endif
```
sunghan-chang, line 734: Could you let me know why this is necessary? There is no `domain` variable usage.

os/board/rtl8730e/src/component/bluetooth/api/rtk_bt_power_control.c
```diff
@@ -75,18 +53,14 @@ static uint32_t rtk_bt_suspend(uint32_t expected_idle_time, void *param)
 
 	pmvdbg("[BT_PS] Enter rtk_bt_suspend\r\n");
 
+	system_enable_power_save();
```
sunghan-chang, line 56: Is that enabling ble power save mode?
Then no need to call disabling? I mean no need to call it in pair?

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1351,12 +1242,11 @@ void up_serialinit(void)
 #endif
 
 #ifdef CONFIG_PM
-	/* Register to receive power management callbacks */
-	ret = pm_register(&g_serialpm.pm_cb);
+	/* Domain registration for operation control */
+	(void)pm_domain_register("UART");
+	(void)pm_domain_register("LOGUART");
```
sunghan-chang, line 1247: What is difference between UART and LOGUART? and what is the usage of them?

ewoodev, line 1247: Is it necessary to distinguish between UART and LOGUART?

pm_domain manages the count of pm_suspend in domain units.
This is to ensure that the use of pm by other domains does not affect it.

If the uart and loguart processes the same logic, "UART" is enough.

os/arch/arm/src/amebasmart/amebasmart_spi.c
edwakuwaku: Sorry it was mistake, I will remove it.

os/arch/arm/src/amebasmart/amebasmart_serial.c
edwakuwaku: LOGUART is console uart, UART is MICOM uart. In tash_command.c, there is a domain suspend operation. Thus as mentioned by Eunwoo. I will keep just UART.
The usage for UART will be to suspend/resume domain when there is communication/operation with MICOM.

edwakuwaku: Hi Eunwoo, I agree with you. LOGUART will be removed.

os/board/rtl8730e/src/component/bluetooth/api/rtk_bt_power_control.c
edwakuwaku: Hi @ziyik , could you help to reply this?

ziyik: This was added in the early stage to ensure Wifi LPS and IPS are enabled.
This is not needed anymore, Wifi module will control its IPS and LPS.
Edward has removed it.

sunghan-chang (commented):
@edwakuwaku As far as I know, our concept is
1. register new domain (get an integer value with return)
2. control pm using the integer value instead of domain name string

But you use a string at every pm function instead of an integer value. It's not good at performance. Could you let use know if you feel difficult to use an integer?

sunghan-chang:
> @edwakuwaku As far as I know, our concept is
> 
> 1. register new domain (get an integer value with return)
> 2. control pm using the integer value instead of domain name string
> 
> But you use a string at every pm function instead of an integer value. It's not good at performance. Could you let use know if you feel difficult to use an integer?

@ewoodev @gSahitya-samsung Please confirm above.

gSahitya-samsung:
> > @edwakuwaku As far as I know, our concept is
> > 
> > 1. register new domain (get an integer value with return)
> > 2. control pm using the integer value instead of domain name string
> > 
> > But you use a string at every pm function instead of an integer value. It's not good at performance. Could you let use know if you feel difficult to use an integer?
> 
> @ewoodev @gSahitya-samsung Please confirm above.

@sunghan-chang You are right, we need to use domain_id instead of domain_name to control pm operations for performance improvement. @edwakuwaku It is advisable to register domain once and use corresponding domain_id multiple times.

edwakuwaku:
> > @edwakuwaku As far as I know, our concept is
> > 
> > 1. register new domain (get an integer value with return)
> > 2. control pm using the integer value instead of domain name string
> > 
> > But you use a string at every pm function instead of an integer value. It's not good at performance. Could you let use know if you feel difficult to use an integer?
> 
> @ewoodev @gSahitya-samsung Please confirm above.

No, there is no difficulties to apply int domain_id. Initially, I think that it is unnecessary to store the domain_id since the pm_domain_register can search and return accordingly.
But as you mentioned, it might cause performance drop since we have to loop through the pm domain buffer everytime.
I have refine it accordingly.

sunghan-chang:
@edwakuwaku Could you check a build break?
```
CC:  chip/amebasmart_pmhelpers.c
chip/amebasmart_pmhelpers.c: In function 'bsp_pm_domain_register':
chip/amebasmart_pmhelpers.c:130:6: error: 'domain' undeclared (first use in this function)
  if (domain < 0) {
      ^~~~~~
```

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -112,21 +113,42 @@ int SOCPS_AONWakeReason(void)
...
+{
+	/* Retrive the domain id from the domain string */
+	int domain_id = g_bsp_domain_id[bsp_drv_id];
+	if (is_suspend) {
+		if (pm_suspend(domain_id) != OK) {
+			pmdbg("Unable to suspend %s domain!\n", domain);
+		}
+	} else {
+		if (pm_resume(domain_id) != OK) {
+			pmdbg("Unable to resume %s domain!\n", domain);
```
sunghan-chang, line 147: Here are `domain`s.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -112,21 +113,42 @@ int SOCPS_AONWakeReason(void)
...
+		gtimer_start_one_shout(&g_timer1, interval_us, NULL, (uint32_t)&g_timer1);
 	}
 	return;
 }
 
+void bsp_pm_domain_register(char *domain_name, int bsp_drv_id)
+{
+	int domain_id = pm_domain_register(domain_name);
+	if (domain_id < 0) {
+		pmdbg("Unable to register BLE DOMAIN\n");
```
sunghan-chang, line 131: It looks not `BLE` specific.

sunghan-chang (approved):

PR #6248, opened 2024-06-21
loadable_apps, binfmt: add config to use --whole-archieve option

abhishek-samsung:
If we use whole-archieve option, then full library gets included. Instead if not enabled, then only required symbols from the library will be picked

About ~350KB (7%) size reduction was observed on internal branch.

### DIFF
--- a/os/binfmt/Kconfig
+++ b/os/binfmt/Kconfig
@@ -48,6 +48,16 @@ config XIP_ELF
 endchoice
 
 
+if XIP_ELF
+
+config XIP_ELF_WARCHIVE
+	bool "Use Whole Archieve option to include full libraries"
+	default n
+	---help---
+		Use --whole-archieve flag for including full libraries
+
+endif
+
 if ELF
 source binfmt/libelf/Kconfig
 endif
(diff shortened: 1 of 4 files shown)


sunghan-chang:
@abhishek-samsung Is this verified with ?

abhishek-samsung:
I verified kernel tc and its workign fine. Internal branch verification is in progress.

sunghan-chang:
@abhishek-samsung Could you let me know the result of internal branch verification?

abhishek-samsung:
> @abhishek-samsung Could you let me know the result of internal branch verification?

It works well on the internal branch. No issues.

os/binfmt/Kconfig
```diff
@@ -48,6 +48,15 @@ config XIP_ELF
 endchoice
 
 
+if XIP_ELF
+
+config XIP_ELF_WARCHIVE
+	bool "Use Whole Archieve option to include full libraries"
+	---help---
```
sunghan-chang, line 55: Let's add default value with y.

sunghan-chang (approved):

PR #6249, opened 2024-06-24
mm: replace watchdog stop to watchdog keepalive

ewoodev:
If watchdog is active, the large amount of logging in heapinfo_parse_heap() can cause reboots by Watchdog To prevent this, we added code to stop Watchdog in mm_manage_allocfail, but some chipsets do not support stopping Watchdog.

Therefore, we modified it so that Watchdog keepalive is called along with log output inside heapinfo_parse_heap(),
 so that even if Watchdog stop is not supported, it will not be rebooted during log output.

### DIFF
--- a/os/mm/mm_heap/mm_heapinfo_parse_heap.c
+++ b/os/mm/mm_heap/mm_heapinfo_parse_heap.c
@@ -59,6 +59,11 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <syslog.h>
+#ifdef CONFIG_WATCHDOG
+#include <tinyara/watchdog.h>
+#include <fcntl.h>
+#include <errno.h>
+#endif
 
 /****************************************************************************
  * Pre-processor Definitions
@@ -66,6 +71,28 @@
 #define HEAPINFO_INT INT16_MAX
 #define HEAPINFO_NONSCHED (INT16_MAX - 1)
 
+/****************************************************************************
+ * Private Functions
+ ****************************************************************************/
+#ifdef CONFIG_WATCHDOG
+static inline int check_watchdog_state(int wd_fd, bool *is_watchdog_running)
+{
+	struct watchdog_status_s wd_status;
+	if (wd_fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+		return ERROR;
+	}
+
+	if (ioctl(wd_fd, WDIOC_GETSTATUS, (unsigned long)&wd_status) != OK) {
+		mfdbg("Fail to get watchdog state, errno %d\n", get_errno());
+		return ERROR;
+	}
+
+	*is_watchdog_running = wd_status.flags & WDFLAGS_ACTIVE;
+
+	return OK;
+}
+#endif
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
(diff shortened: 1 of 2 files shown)


ewoodev:
> Previous code does not print details with wd stop failure. How about new one?

Hello @sunghan-chang,
I have checked that all heap info detail is printed with this PR.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -250,7 +263,9 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 			}
 		}
 	}
-
+#ifdef CONFIG_WATCHDOG
+	close(wd_fd);
```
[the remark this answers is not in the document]
ewoodev: Thanks for review, I applied.

ewoodev:
> Previous code does not print details with wd stop failure. How about new one?

I updated the content we considered privous. To prevent reboot by watchdog when CONFIG_MM_ASSERT_ON_FAIL is disabled, we decided not to output logs. 

This is updated. Please review it.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -182,6 +198,24 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 					heap_dbg("0x%x | %8d |   %c    |            |       |\n", node, node->size, 'F');
 				}
 			}
+#ifdef CONFIG_WATCHDOG
+			if (wd_fd > 0) {
```
sunghan-chang, line 202: Line 121 checks wd_fd so that this line looks unnecessary. It could be a defect with always true on the SVACE.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -110,6 +115,17 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 
 	ASSERT(mm_check_heap_corruption(heap) == OK);
 
+#ifdef CONFIG_WATCHDOG
+	int wd_fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDONLY);
+#ifndef CONFIG_MM_ASSERT_ON_FAIL
+	if (wd_fd < 0) {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+		mfdbg("The heap usage dump doesn't printed\n");
```
sunghan-chang, line 123: Let's change the sentence as below
`It will not print detail of heap usages to prevent watchdog reset.`

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -110,6 +115,34 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
...
+	} else {
+		mfdbg("Fail to open %s, errno %d\n", CONFIG_WATCHDOG_DEVPATH, get_errno());
+#ifdef CONFIG_MM_ASSERT_ON_FAIL
+		mfdbg("Watchdog can reboot during the printing heap usage log.\n");
+#else
+		mfdbg("It doesn't print the heap usage dump.\n");
+		return;
+#endif
+	}
+#endif
```
sunghan-chang, line 144: Let's make it as a separate function.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -182,6 +215,26 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 					heap_dbg("0x%x | %8d |   %c    |            |       |\n", node, node->size, 'F');
 				}
 			}
+#ifdef CONFIG_WATCHDOG
```
sunghan-chang, line 218: same as previous

os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -110,6 +115,34 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 
 	ASSERT(mm_check_heap_corruption(heap) == OK);
 
+#ifdef CONFIG_WATCHDOG
+	bool is_watchdog_running = false;
+	struct watchdog_status_s wd_status;
+	int wd_fd = open(CONFIG_WATCHDOG_DEVPATH, O_RDONLY);
+	if (wd_fd >= 0) {
```
sunghan-chang, line 122: Is it possible to get 0 for success?

ewoodev: file descriotion can be 0

os/mm/mm_heap/mm_heapinfo_parse_heap.c
ewoodev: thanks you for suggestion.

I separated checking wd fd and checking wd state.

os/mm/mm_heap/mm_heapinfo_parse_heap.c
ewoodev: This is difficult to seperate other function because it has to be output in the middle of outputting the log.

sunghan-chang (approved):

PR #6253, opened 2024-06-24
build/configs/rtl8730e: Revise rtk data binary build flow

edwakuwaku:
- Get value from .config, to differentiate whether the data binary build is necessary
- Skip the operation if external flash is invalid

### DIFF
--- a/build/configs/rtl8730e/rtl8730e_make_bin.sh
+++ b/build/configs/rtl8730e/rtl8730e_make_bin.sh
@@ -63,7 +63,7 @@ cat $BINDIR/ram_2.bin > $BINDIR/ca32_image2_all.bin
 
 python $GNUUTL/code_analyze.py
 
-python $GNUUTL/rtk_data_binary.py
+ext_flash_support=$(python $GNUUTL/rtk_data_binary.py)
 
 echo "========== Image manipulating end =========="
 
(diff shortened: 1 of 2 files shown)


edwakuwaku:
Hi @jeongarmy , could you please help to try again with this on internal branch?

sunghan-chang (commented):
Sorry but I don't know what this resolves the situation of a problem. Could you elaborate the issue and how to run with this?

edwakuwaku:
> Sorry but I don't know what this resolves the situation of a problem. Could you elaborate the issue and how to run with this?

Hi @sunghan-chang , after discussion with Ahreum, there are 2 things for this PR:
1) Not need to build the data binary for defconfig without external flash enabled
2) The data binary is not generated on some of the internal defconfig, thus I revised the flow and request to check again.

sunghan-chang (approved):

build/tools/amebasmart/gnu_utility/rtk_data_binary.py
```diff
@@ -32,8 +35,18 @@ def text_to_binary(input_file, output_file):
...
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../../os/tools')))
+import config_util as util
+CONFIG_SECOND_FLASH_PARTITION = util.get_value_from_file(cfg_file, "CONFIG_SECOND_FLASH_PARTITION=").rstrip('\n')
+if CONFIG_SECOND_FLASH_PARTITION == "y":
     input_file = dirpath + "/rtk_data_binary.txt"
     output_file = dirpath + "/rtk_ext_flash_data.bin"
 
     text_to_binary(input_file, output_file)
+    # Ext flash valid
+    print("1")
```
jeongarmy, line 47: Could you change the log to correct message?

edwakuwaku: This print is to send the value back to the bash. You can find sys_exit(0) at below. It will be used to determine whether we pass the binary to do prepend header.

build/configs/rtl8730e/rtl8730e_make_bin.sh
```diff
@@ -83,10 +83,14 @@ $GNUUTL/prepend_header.sh $BINDIR/bl1_sram.bin  __ca32_bl1_sram_start__  $BINDIR
 $GNUUTL/prepend_header.sh $BINDIR/bl1.bin  __ca32_bl1_dram_start__  $BINDIR/target_img2.map
 $GNUUTL/prepend_header.sh $BINDIR/fip.bin  __ca32_fip_dram_start__  $BINDIR/target_img2.map
 
-$GNUUTL/prepend_header.sh $GNUUTL/rtk_ext_flash_data.bin  __flash_ext_data_start__  $BINDIR/target_img2.map
-cp $GNUUTL/rtk_ext_flash_data_prepend.bin $BINDIR/rtk_ext_flash_data.bin
-rm $GNUUTL/rtk_ext_flash_data.bin
-rm $GNUUTL/rtk_ext_flash_data_prepend.bin
+if [ "$ext_flash_support" = "1" ];then
```
edwakuwaku, line 86: @jeongarmy Please check here~

edwakuwaku: According to the logic here, if we enter this condition, means the binary creation is successful. Else, it should print the logs for condition `ext_flash_support == 0`.
I tried on flat and loadable config, it is working as expected.
If you think we still need some logs, then I shall add it here.

PR #6255, opened 2024-06-24
[DO NOT REVIEW] os/drivers/lcd : partial are update

anjana348:
[DO NOT REVIEW] st7785 partial update issue

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -144,6 +144,69 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
 	return OK;
 }
 
+static int lcd_setarea(fb_coord_t row_start, fb_coord_t row_end, fb_coord_t col_start, fb_coord_t col_end)
+{
+	int transfer_status = 0;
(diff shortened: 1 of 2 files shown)


os/drivers/lcd/mipi_lcd.c
```diff
@@ -144,6 +144,69 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
...
 
+static int lcd_setarea(fb_coord_t row_start, fb_coord_t row_end, fb_coord_t col_start, fb_coord_t col_end)
+{
+	int transfer_status = 0;
+	struct mipi_dsi_msg msg_row, msg_col, msg;
+	FAR struct mipi_lcd_dev_s *priv = &g_lcdcdev;
+
+	priv->config->lcd_mode_switch(false);
+
+	u8 cmd_addr_0[1] = {0x00};
```
edwakuwaku, line 155: In here, you sent 0x29, 0x2B, 0x29, 0x2A, 0x2C. As I tested, sending just 0x2B and 0x2A is also working fine (ie update row and column).
May I know the reason to send 0x29 and 0x2C?

PR #6256, opened 2024-06-24
os/arch/arm/src/armv7-a: Update hotplug logic for PM + SMP

edwakuwaku:
- When the secondary core is not in idle thread and system went into sleep, we faced some unexpected behaviour during system wakeup (ie. crash)
- Thus, this PR added a check to the hotplug handler, to determine whether the secondary core is already in idle thread and ready for shutdown
- If the secondary core is still running, the primary core will restore it's irq, so that it will be able to react to some task/scheduler change which is requested by secondary core (ie. pause request)


edwakuwaku:
Hi @ewoodev  @kishore-sn  @ritesh55555 , let's discuss about the PM + SMP sleep/wakeup operation in this PR.

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -83,24 +86,41 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 			   (ie. TizenRT Kernel logic is not ready to handle such case yet)
 			   Thus, we will be facing a deadlock when cpuA fire SGI2 to cpuB (ie. pausing CPU)
 	*/
+	int cpu = this_cpu();
+	char curr_task[CORE_IDLE_TASK_CHAR_COUNT];
+	snprintf(curr_task, CORE_IDLE_TASK_CHAR_COUNT, "CPU%d IDLE", cpu);
+	/* If current cpu is in idle thread, then proceed for hotplug operation */
+	if (!strncmp(g_running_tasks[cpu]->name, curr_task, CORE_IDLE_TASK_CHAR_COUNT)) {
```
ritesh55555, line 93: we can directly check with task pid as idle task pid = cpu

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -83,24 +86,41 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
...
-	/* If the cpu off operation failed, we should face a problem during wakeup
-	booting the secondary core, thus we can ignore the return value here first
-	*/
-	/* Shut down the secondary core */
-   	(void)psci_cpu_off(); 
+		/* If SMP is enabled, the secondary core should take care of setting up the HW timer */
+		/* We need to add some flag to indicate the action has been completed, or else for 
+		   core number > 2, this code piece might be executed several times when each core goes hotplug,
+		   which might lead to some unexpected problem
+		*/
```
kishore-sn, line 107: Hi...
I don't understand why the second core needs to setup the hw timer in case of smp? Can you please elaborate?

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
edwakuwaku: Thank you for suggesting, this seems more clear-cut to me.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -140,6 +140,7 @@ But in case that happens, we have to introduce some strategy to handle that
 								arm_cpu_sgi(GIC_IRQ_SGI4, (1 << 1));
 								arm_arch_timer_int_mask(0);
 								up_timer_enable();
+								pmdbg("Secondary core is still running!\n");
```
kishore-sn, line 143: At this point, we have already called pm_changestate and informed all drivers that the board is going into sleep. However, if the secondary core never goes to idle or if there is a delay in secondary core going to idle, then the system remains in previous power state, but the drivers assume sleep state. 

How about calling pm_changestate at line 160? In that case, we dont have to worry about the hw timer also.

edwakuwaku: There are some changes in the PM structure, now we will use suspend/resume as according to each domain (ie. this current flow of pm_changestate() will be modified in other PR). Thus, it will not have the concept of whether each of the driver to assume any PM state. In return, if the secondary core still utilise any driver, it should be fine.
@ewoodev  Could you please share your opinion regarding this?

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -83,24 +83,39 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 			   (ie. TizenRT Kernel logic is not ready to handle such case yet)
 			   Thus, we will be facing a deadlock when cpuA fire SGI2 to cpuB (ie. pausing CPU)
 	*/
+	int cpu = this_cpu();
+	/* If current cpu is in idle thread, then proceed for hotplug operation */
+	if (g_running_tasks[cpu]->pid == cpu) {
+		/* Set secondary core to hotplug state */
```
ritesh55555, line 89: Also i saw , g_running_tasks list is not getting updated anywhere in the code . if I am right , then i think we should use g_assignedtasks[cpu] to check

os/arch/arm/src/amebasmart/amebasmart_idle.c
ewoodev: I believe that if the board sleep is called as Kishore said, it must enter sleep mode. Otherwise, there would be no point in changing the state in the PM. At least, if the board sleep entry fails, it needs to be informed with a return error format.

And I am not sure if it is the right way to check the scheduling status of the second CPU in the BSP code.
If the second CPU is executing regular threads, the PM state should not become sleep either.

I suggest checking this in the pm_idle. We can check this by calling sched_cpu_pause. While the PM is trying to enter sleep, no CPU should be assigned a tick ISR. Scheduling should stop.

edwakuwaku: If then, the logic for checking secondary core status should be moved out to pm_changestate() (ie, As I knew, this will decide whether we can do a state change?), and firing the SGI4 (ie. hotplug handler) to secondary core should be triggered in pm_idle().

edwakuwaku: As I have checked, there seems to be no sched_cpu_pause in current codebase. Do you mean up_cpu_pause...?

ewoodev: sorry,.. sched_pause_cpu()
i think, the logic for checking secondary core status should be moved out to pm_idle() only, and Can SGI4 stay up_pm_board_sleep()?

But please confirm if this is possible first.

edwakuwaku: I have updated the logic, pm_idle() will handle checking all other core status according to CONFIG_SMP_NCPUS.
I don't know why we need to invoke sched_cpu_pause, could you explain the reason?
SGI4 can be done in up_pm_board_sleep(), but there is a slight time gap between "finish checking secondary core status"<->"send SGI in BSP". I'm afraid that the secondary core undergo context switch or some interrupt between this time period. (ie. There is no additional checking in hotplug handler now)

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
edwakuwaku: Thank you for reminding~

os/arch/arm/src/amebasmart/amebasmart_idle.c
ewoodev: the reason why we need to call sched_pause_cpu is also related to your concern.

Even if we check the scheduling status of the second cpu in pm_idle, the second cpu can receive tick ISR until SGI4 is called. If this is wrong, please correct it. It is thought that the scheduling status of the second cpu is allowed to change.

edwakuwaku: Yes, I understand the pause is to freeze the status of secondary core. Then, the pause will be done in pm_idle. And as I understand, the resume need to be called so that the secondary core can handle SGI4, which means the resume will be done in amebasmart_idle.c, is that acceptable? Please correct if my understanding is wrong.

os/pm/pm_idle.c
```diff
@@ -65,12 +65,29 @@ void pm_idle(void)
...
 	/* Perform state-dependent logic here */
-	pmvdbg("newstate= %d\n", newstate);
+	/* For SMP case, we need to check secondary core status
+	   If secondary core status is not in idle thread, abort
+	   the sleep and check again on next cycle
+	*/
+#ifdef CONFIG_SMP
+	if (newstate == PM_SLEEP) {
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			tcb = current_task(cpu);
```
kishore-sn, line 84: Even if we check the cpu 1 idle state here, there are many steps happening, before SGI4 is actually sent to cpu1. Due to this time difference, it is possible that cpu1 might start running a different task. So, I think we should check for the cpu1 idle state in the SGI4 handler only.

os/pm/pm_idle.c
```diff
@@ -65,12 +65,29 @@ void pm_idle(void)
...
+			if (tcb->pid != cpu) {
+				goto EXIT;
+			}
+		}
+	}
+#endif
 	/* Then force the global state change */
 	if (pm_changestate(newstate) < 0) {
```
kishore-sn, line 92: I think pm_changestate must not be called until cpu1 has gone to sleep. 
As I mentioned above, we need to move the cpu1 idle check into sgi4 handler. So, there might be a delay in the cpu's actually going to sleep. My suggestions is that we need to pass pm_changestate API as a function pointer to up_pm_board_sleep(). After cpu1 has gone to sleep, up_pm_board_sleep() must call the changestate API, and then put cpu0 to sleep.

os/pm/pm_idle.c
edwakuwaku: Hi @kishore-sn , as I have discussed with @ewoodev , we think that pm_idle could check the state of cores other than primary core (ie. only primary core conduct PM logic). Thus, we will utilize SGI2 to control other core status, to prevent from context switch or other interrupt

os/pm/pm_idle.c
edwakuwaku: As mentioned in above comment, after we confirmed the core status, we will pause it immediately. And we will fire the SGI4 once we complete the change state to PM_SLEEP.

os/pm/pm_idle.c
```diff
@@ -65,12 +65,32 @@ void pm_idle(void)
...
+#ifdef CONFIG_SMP
+	int cpu;
+	int paused_count = 0;
+	FAR struct tcb_s *tcb = NULL;
 #endif
 	/* Decide, which power saving level can be obtained */
 	flags = enter_critical_section();
 	newstate = pm_checkstate();
 	/* Perform state-dependent logic here */
-	pmvdbg("newstate= %d\n", newstate);
```
edwakuwaku, line 73: This log is removed due to infinite printing if pm log is enabled.

os/pm/pm_idle.c
```diff
@@ -65,22 +65,54 @@ void pm_idle(void)
...
+	*/
+#ifdef CONFIG_SMP
+	if (newstate == PM_SLEEP) {
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			tcb = current_task(cpu);
+			if (tcb->pid != cpu) {
+				paused_count = cpu - 1;
+				goto EXIT;
+			}
+			(void)sched_pause_cpu(tcb);
```
kishore-sn, line 90: Why do we not use up_cpu_pause() here?

os/pm/pm_idle.c
```diff
@@ -65,22 +65,54 @@ void pm_idle(void)
...
+		pmvdbg("newstate= %d\n", newstate);
 	}
 	/* If current state is not good to go sleep then do core power saving*/
 	if (g_pmglobals.state != PM_SLEEP) {
 		goto EXIT;
 	}
+#ifdef CONFIG_SMP
+	/* Send signal to shutdown other cores here */
+	for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+		up_cpu_hotplug(cpu);
```
kishore-sn, line 109: If we perform the hotplug operation here, then better remove duplicate SGI4 from up_pm_board_sleep().

os/pm/pm_idle.c
ewoodev: @edwakuwaku I understand Kishore's concern. Since we have paused the CPU, but it may already be in an tick IRQ or late to be paused, the checked CPU state may differ.

cpu sched checking in the SGI4 is better.
Thank you for consider and explain.

How about requesting SGI4 in up_cpu_hotplug and returning it with a flag? 
If the up_cpu_hotplug returns fail, we don't need to call pm_changestate.

os/pm/pm_idle.c
edwakuwaku: Changed~

os/pm/pm_idle.c
edwakuwaku: Updated~

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -391,4 +391,35 @@ int up_cpu_resume(int cpu)
...
+ *   previously been called within the same critical section. This
+ *   function should only be called by primary core which conducts PM logic.
+ *
+ ****************************************************************************/
+void up_cpu_hotplug(int cpu)
+{
+  DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
+
+  /* Fire SGI for cpu to enter hotplug */
+  arm_cpu_sgi(GIC_IRQ_SGI4, (1 << cpu));
```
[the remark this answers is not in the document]
edwakuwaku: Hi @kishore-sn , the status check for each of the cpu has been done before we do pm_changestate, can you check the initial part of pm_idle()?

os/pm/pm_idle.c
```diff
@@ -65,22 +65,62 @@ void pm_idle(void)
...
+		(void)up_cpu_resume(cpu);
+		/* Check whether each of the cpu has entered hotplug */
+		while(up_get_secondary_cpu_state(cpu) != CPU_HOTPLUG);
```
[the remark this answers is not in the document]
edwakuwaku: This is related to above discussion, if needed, we can add a retry count when checking the core status.
The design implementation here is I assume that all the cores are already in idle thread and ready to go hotplug.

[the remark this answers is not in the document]
kishore-sn: ok sounds good.

os/pm/pm_idle.c
```diff
@@ -65,22 +65,62 @@ void pm_idle(void)
...
+#ifdef CONFIG_SMP
+	if (newstate == PM_SLEEP) {
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			/* If the CPU is just back from sleep, abort the sleep */
+			if (up_get_secondary_cpu_state(cpu) == CPU_WAKE_FROM_SLEEP) {
+				goto EXIT;
+			}
+			tcb = current_task(cpu);
+			/* Lock the cpu first, before checking which task it is handling */
+			(void)up_cpu_pause(tcb);
```
ewoodev, line 92: i think, we need to bring tcb after pause so that can be protected from context switching.
```suggestion
			/* Lock the cpu first, before checking which task it is handling */
			(void)up_cpu_pause(cpu);
			tcb = current_task(cpu);

```

edwakuwaku:
Hello @kishore-sn  @ewoodev @ritesh55555  , I have pushed the latest implementation. Instead of pause/resume, I changed to use gating as I think that method is not affecting the task/scheduler block.
When I tested with pause/resume, there is some deadlock happening after secondary core is paused. Thus, I think there are still some misunderstanding from our side regarding the usage of cpu pause/resume.
I tested this changes with both cores in idle thread, it works fine as previously.
Please help to test this changes on the condition of "cpu0: idle thread, cpu1: not idle thread". The sleep should be aborted under this condition.

os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -40,7 +40,22 @@
 #include "barriers.h"
 
 #ifdef CONFIG_CPU_GATING
-volatile uint32_t ulFlashPG_Flag = 0;
+static volatile uint32_t g_cpugating_flag[CONFIG_SMP_NCPUS];
+
+void up_set_gating_flag_status(uint32_t CoreID, uint32_t val)
+{
+	g_cpugating_flag[CoreID] = val;
```
ewoodev, line 47: This doesn't seem to fit the pair.

when flag setting is plus 1 `g_cpugating_flag[cpu]++;`,
but  flag release is set 0 foced.

Is this okay when the request comes at the same time?

os/pm/pm_idle.c
```diff
@@ -65,12 +69,45 @@ void pm_idle(void)
...
+			up_cpu_gating(cpu);
+			/* Polling the status of the target CPU, 
+			   The flag should be > 0 if it is already gated
+			*/
+			while(!up_get_gating_flag_status(cpu));
+			tcb = current_task(cpu);
+			/* Check if current cpu is in idle thread, and whether there is pending
+			   pause request on primary core 
+			*/
+			if (tcb->pid != cpu || up_cpu_pausereq(me)) {
```
ewoodev, line 104: do wa need to check still up_cpu_pausereq?

os/arch/arm/src/armv7-a/arm_cpugating.c
edwakuwaku: Hi eunwoo, I think there is some restriction on the usage here, and it should only be applied on some special cases (ie. we need to freeze the state of some specific cores).
However, when a gating request is sent, the requestor normally should do a polling to check the status of the gated cpu. And the freeing of flag should be called after some checking/operation is completed. In this case, it should work as a pair.
So, I think some assumption or limitation should be explained more clearly if someone is applying this.

os/pm/pm_idle.c
edwakuwaku: I think we could remove it.

edwakuwaku:
@ritesh55555  @kishore-sn , Could you please test with this PR on internal branch and leave the test results here?
Thanks.

gSahitya-samsung:
@edwakuwaku Please check merge conflicts

ritesh55555:
> @ritesh55555 @kishore-sn , Could you please test with this PR on internal branch and leave the test results here? Thanks.

Can you please rebase the PR, when i merge and test i am getting stuck just after booting

edwakuwaku:
I have rebased to latest master.

os/pm/pm_idle.c
```diff
@@ -98,6 +136,15 @@ void pm_idle(void)
...
 		}
+#ifdef CONFIG_SMP
+		/* Send signal to shutdown other cores here */
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			up_cpu_hotplug(cpu);
+			up_set_gating_flag_status(cpu, 0);
+			/* Check whether each of the cpu has entered hotplug */
+			while(up_get_secondary_cpu_state(cpu) != CPU_HOTPLUG);
+		}
+	#endif
```
ritesh55555, line 147: wrong identation

os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -40,7 +40,22 @@
...
+	ARM_DSB();
+	/* Flag already reach 0 */
+	if (!g_cpugating_flag[CoreID]) {
+		SP_SEV();
```
ritesh55555, line 51: can you replace it by  __asm__ __volatile__ ("sev" : : : "memory"); 
to use SP_SEV(), we need to define CONFIG_ARM_HAVE_WFE_SEV but it is not .
This was causing the hang issue

os/pm/pm_idle.c
edwakuwaku: fixed~

ritesh55555:
This PR is working fine with product code. There is no more hand issue during sleep and wakeup. It is working fine

ritesh55555:
This PR is working fine with product code. There is no more hang issue during sleep and wakeup. It is working fine

os/pm/pm_idle.c
```diff
@@ -84,7 +94,35 @@ void pm_idle(void)
...
-		pmvdbg("newstate= %d\n", newstate);
+		/* For SMP case, we need to check secondary core status
+	     * If secondary core status is not in idle thread, abort
+	     * the sleep and check again on next cycle
+		 */
+#ifdef CONFIG_SMP
+		if (newstate == PM_SLEEP) {
+			for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+				/* If the CPU is just back from sleep, abort the sleep */
+				if (up_get_secondary_cpu_state(cpu) == CPU_WAKE_FROM_SLEEP) {
```
kishore-sn, line 105: Can we change this api name as up_get_cpu_state and up_set_cpu_state? 
As these apis are used for all cpus in the system, the name secondary_cpu seems wrong.

os/pm/pm_idle.c
```diff
@@ -26,8 +26,12 @@
 #include <tinyara/clock.h>
 #include <tinyara/irq.h>
 #include <tinyara/arch.h>
+#include "../kernel/sched/sched.h"
```
gSahitya-samsung, line 29: Please use `<tinyara/sched.h>`

os/pm/pm_idle.c
```diff
@@ -72,6 +76,12 @@ void pm_idle(void)
 	clock_t now;
 #ifdef CONFIG_PM_TIMEDWAKEUP
 	clock_t delay;
+#endif
+#ifdef CONFIG_SMP
+	int me = this_cpu();
```
gSahitya-samsung, line 81: Is this variable required ?

os/pm/pm_idle.c
```diff
@@ -84,7 +94,35 @@ void pm_idle(void)
 		/* Decide, which power saving level can be obtained */
 		newstate = pm_checkstate();
 		/* Perform state-dependent logic here */
-		pmvdbg("newstate= %d\n", newstate);
+		/* For SMP case, we need to check secondary core status
+	     * If secondary core status is not in idle thread, abort
+	     * the sleep and check again on next cycle
+		 */
+#ifdef CONFIG_SMP
+		if (newstate == PM_SLEEP) {
```
gSahitya-samsung, line 102: Please move this section below `pm_metrics` code

gSahitya-samsung (changes requested):

gSahitya-samsung:
Can we keep `SMP Hotplug` logic outside `pm_idle` by creating new APIs ?

os/pm/pm_idle.c
edwakuwaku: no, the definition needed is defined in os/kernel/sched/sched.h. If using the path you mentioned, it will face compilation error.

edwakuwaku:
> Can we keep `SMP Hotplug` logic outside `pm_idle` by creating new APIs ?

I have talked with @ritesh55555 about this, after this PR is merged, cleanup can be done from Samsung team on PM + SMP which you think it is necessary.

os/pm/pm_idle.c
edwakuwaku: nope it's not, I will remove it

os/pm/pm_idle.c
edwakuwaku: sure

os/pm/pm_idle.c
edwakuwaku: I agree, renamed it accordingly

os/arch/arm/src/armv7-a/arm_cpugating.c
ewoodev: And does gating mean forced getting the WFE state?

I couldn't find cpu gating, I just found power gating of core.

edwakuwaku: > It doesn't seem necessary for the CPU gating code to know about flags. Wouldn't a function like up_cpu_gating_release() be needed?

What is the benefit of doing so? In SGI2, there is also a up_cpu_pausereq to get the flag of cpu. I don't think we need another API to do that.

> And if there are more secondary CPUs, it seems that the release of this state should only occur when all cores have been released. Is there any reason why the flag needs to be initialized to zero by force?

 I don't understand what you are referring to.

edwakuwaku: > And does gating mean forced getting the WFE state?
> 
> I couldn't find cpu gating, I just found power gating of core.

CPU gating means, the target CPU is blocked to execute any operation, thus we use an infinite loop to force blocking the CPU. WFE is applied to inhibit additional power consumption.

ewoodev: > > It doesn't seem necessary for the CPU gating code to know about flags. Wouldn't a function like up_cpu_gating_release() be needed?
> 
> What is the benefit of doing so? In SGI2, there is also a up_cpu_pausereq to get the flag of cpu. I don't think we need another API to do that.

up_cpu_pausereq is using the SP_LCOK and limit this flag in arm_cpupause. It dosen't allow setting by other code.
If you should use flag, you need to limit in this file.
If you allow set flag using up_set_gating_flag_status, we can't avoid wrong used it form other module. 

> > And if there are more secondary CPUs, it seems that the release of this state should only occur when all cores have been released. Is there any reason why the flag needs to be initialized to zero by force?
> 
> I don't understand what you are referring to.

`g_cpugating_flag[cpu]++;` is counting structure. 
but up_set_gating_flag_status(cpu, 0) is calling with 0 or 1 argument.

I thought this was for a 4-core case.
It's because all CPUs have to maintain gating until it is released for gating. However, this seems to be handled by using duplicate delay interrupts.
Do we need to have a counting structure right?

edwakuwaku: > up_cpu_pausereq is using the SP_LCOK and limit this flag in arm_cpupause. It dosen't allow setting by other code.
> If you should use flag, you need to limit in this file.
> If you allow set flag using up_set_gating_flag_status, we can't avoid wrong used it form other module.

Understood. But, why would we use it wrongly? This should only be used under specific circumstances.

> g_cpugating_flag[cpu]++; is counting structure.
> but up_set_gating_flag_status(cpu, 0) is calling with 0 or 1 argument.
> 
> I thought this was for a 4-core case.
> It's because all CPUs have to maintain gating until it is released for gating. However, this seems to be handled by using duplicate delay interrupts.
> Do we need to have a counting structure right?

Yes, this supports up to 4 cores case. The structure follows CONFIG_SMP_NCPUS. 0 or 1 argument is set flag to control CPU state. What do you mean by duplicate delay interrupts? Currently, we gate each core and check its condition one by one, and if one of the secondary CPU violate the condition, we release gated core in descending order. Is there any misunderstanding...?

ewoodev: > > up_cpu_pausereq is using the SP_LCOK and limit this flag in arm_cpupause. It dosen't allow setting by other code.
> > If you should use flag, you need to limit in this file.
> > If you allow set flag using up_set_gating_flag_status, we can't avoid wrong used it form other module.
> 
> Understood. But, why would we use it wrongly? This should only be used under specific circumstances.
> 
> > g_cpugating_flag[cpu]++; is counting structure.
> > but up_set_gating_flag_status(cpu, 0) is calling with 0 or 1 argument.
> > I thought this was for a 4-core case.
> > It's because all CPUs have to maintain gating until it is released for gating. However, this seems to be handled by using duplicate delay interrupts.
> > Do we need to have a counting structure right?
> 
> Yes, this supports up to 4 cores case. The structure follows CONFIG_SMP_NCPUS. 0 or 1 argument is set flag to control CPU state. What do you mean by duplicate delay interrupts? Currently, we gate each core and check its condition one by one, and if one of the secondary CPU violate the condition, we release gated core in descending order. Is there any misunderstanding...?

`we release gated core in descending order.`
Could you let me know where code release gating count?
current code, only set 1 and 0....

edwakuwaku: @ewoodev , Please check the code here https://github.com/Samsung/TizenRT/pull/6256/files#diff-34272b827c1d9af918fa7e69ffe694e7e31445afb6d3e04081179ea1a2bb5225R168-R174.
If one of the secondary cores cannot go hot plug, we set the flag to 0 and do SEV.
And then, https://github.com/Samsung/TizenRT/pull/6256/files#diff-af4e58c133b5aca9d3f92e02b429cae7c5ad60207bfee9e537c5cb97ebb9ef2cR80-R81. Once the condition does not meet, it will be released from gated state and resume its operation.

ewoodev: @edwakuwaku Thank you for explain detail.
I still couldn't understand this.

You call up_cpu_hotplug() for CPU gating. However, to release the gating, you directly set the flag using up_set_gating_flag_status().

There is no setting of a flag for CPU gating, but for release gating we need to set flag manually. I think this doesn't match general idea of a lock/unlock pair.

Secondly, for CPU gating, you increment g_cpugating_flag with ++. But when releasing the gating, you set it to 0. I can't understand why there's a counting structure instead of a boolean(set 0, 1) If we did ++, then we should do --.

Could you let me know your opinion?

edwakuwaku: Hello @ewoodev , I understand your concern now.
The reason I added the g_cpugating_flag increment, is to ensure that the gating will be success in any circumstances (ie. if any user invoke gating SGI, but flag is not set).
Anyway, your point make sense to me. We could add setting flag to 1 when invoke up_cpu_gating. In that case, it would be a pair.
Also as you mentioned, if there is other usage applying this, then a counting structure is necessary to differentiate that as well. But, if any user need to apply this, it has to be used with care, if synchronization is not done carefully, it would cause hang issue easily.

ewoodev (approved):

sunghan-chang (approved):

PR #6257, opened 2024-06-25
os/pm: Clean PM Code

gSahitya-samsung:
This commit adds registered domain_id checks in new PM APIs. With new domain registration API, we need to register TASH domain separately and use pm_suspend and pm_resume with tash's domain_id. It also corretcs typos and missing description in PM code.


apps/shell/tash_command.c
```diff
@@ -564,22 +567,27 @@ int tash_execute_cmd(char **args, int argc)
...
 				if (pmdrv_fd < 0) {
 					shdbg("open /dev/pm failed(%d), \n", get_errno());
+					ret = ERROR;
 				} else {
-					pm_suspended = ioctl(pmdrv_fd, PMIOC_SUSPEND, PM_NORMAL);
-					if (pm_suspended != OK) {
+					if (ioctl(pmdrv_fd, PMIOC_DOMAIN_REGISTER, &pm_domain_arg) != OK) {
+						shdbg("pm_domain_register failed(%d)\n", get_errno());
+						ret = ERROR;
+					} else if (ioctl(pmdrv_fd, PMIOC_SUSPEND, pm_domain_arg.domain_id) != OK) {
```
ewoodev, line 575: we need to check pm_domain_arg.domain_id > 0

apps/shell/tash_command.c
```diff
@@ -564,22 +567,27 @@ int tash_execute_cmd(char **args, int argc)
...
 #endif
 				(*tash_cmds_info.cmd[cmd_idx].cb) (argc, args);
 #ifdef CONFIG_PM
-				if (pmdrv_fd > 0) {
-					if (pm_suspended == OK) {
-						/* pm_resume should only be called when the pm_suspend is executed correctly. */
-						if (ioctl(pmdrv_fd, PMIOC_RESUME, PM_NORMAL) != OK) {
-							shdbg("pm_resume failed(%d)\n", get_errno());
-						}
+				if (ret == OK) {
```
ewoodev, line 583: you don't return using the ret.
so, how to keep this checking value name to pm_suspended

apps/shell/tash_command.c
gSahitya-samsung: No, if domain_id < 0, then ioctl(PMIOC_DOMAIN_REGISTER) return -1, so we don't call ioctl(PMIOC_SUSPEND) in else-if part

Taejun-Kwon:
Dear HQ,
1. First of all, you should define PM interface include operations that is need to be implmented by Vendor easily. (up_set_dvfs can be one of operation)
```
FAR struct inode *inode = filep->f_inode;
FAR struct pm_dev_s dev = inode->i_private;
int ret;

switch(cmd) {
case PMIOC_TUNEFREQ:
	if (dev->pm_dvfs != NULL) {
		ret = dev->pm_dvfs(arg);
	} else {
		ret -ENOSYS;
	}
	break;
...
```
2. So you should define some dev structure include operations & value(include pm_domain_map) this is necessary to protect access to each of value without open driver.
3. In ioctl, you call set_errno, it should just return one of errno, set_errno will be called by fs_ioctl.

I think because of development schedule, this seems that hard to modified for now, but should be done in someday.

ewoodev:
> Dear HQ,
> 
> 1. First of all, you should define PM interface include operations that is need to be implmented by Vendor easily. (up_set_dvfs can be one of operation)
> 
> ```
> FAR struct inode *inode = filep->f_inode;
> FAR struct pm_dev_s dev = inode->i_private;
> int ret;
> 
> switch(cmd) {
> case PMIOC_TUNEFREQ:
> 	if (dev->pm_dvfs != NULL) {
> 		ret = dev->pm_dvfs(arg);
> 	} else {
> 		ret -ENOSYS;
> 	}
> 	break;
> ...
> ```
> 
> 2. So you should define some dev structure include operations & value(include pm_domain_map) this is necessary to protect access to each of value without open driver.
> 3. In ioctl, you call set_errno, it should just return one of errno, set_errno will be called by fs_ioctl.
> 
> I think because of development schedule, this seems that hard to modified for now, but should be done in someday.

Thank you for suggestion. For apply this, Let us dissus detail. Thank you

os/kernel/init/os_start.c
```diff
@@ -745,6 +745,17 @@ void os_start(void)
...
 
+#ifdef CONFIG_PM
+	/* Initialize the power management subsystem.  This MCU-specific function
+	 * must be called *very* early in the initialization sequence *before* any
+	 * other device drivers are initialized (since they may attempt to register
+	 * with the power management subsystem).
+	 */
+	pm_initialize();
+	/* We cannot enter low power state until boot complete */
+	pm_suspend(PM_IDLE_DOMAIN);
```
ewoodev, line 756: We call pm_suspend here, but we need to call pm_resume in the app.
It seems like an unconventional calling structure that we can only know about. 

How about setting the initial state of pm as 'not stated' and adding a call to 'pm_start'?

r-prabu (approved):

os/kernel/init/os_start.c
ewoodev: Or, how about call pm_resume(idle) before starting the idle loop.
The app code calls suspend when the app starts and resumes when init ends.
This is because it cannot fall into an idle loop while the app thread is being created.
binary manager also.

And please separate PR for this commit.

os/pm/pm.h
```diff
@@ -140,20 +140,14 @@ struct pm_global_s {
 
 	/* state       - The current state for this PM domain (as determined by an
 	 *               explicit call to pm_changestate())
-	 * recommended - The recommended state based on the PM algorithm.
 	 */
 
 	uint8_t state;
-	uint8_t recommended;
```
ewoodev, line 143: won't be used this in lcd structure?
Please seperate pr about this.

gSahitya-samsung: Its not needed for LCD structure

ewoodev (approved):

os/pm/pm_checkstate.c
```diff
@@ -103,50 +102,39 @@ enum pm_state_e pm_checkstate(void)
...
-	 * slicing is controlled by driver activity.  In either case, the duration
-	 * of the time slice is only approximate; during times of heavy activity,
-	 * time slices may be become longer and the activity level may be over-
-	 * estimated.
-	 */
+	/* Board should remain wakeup for minimunu CONFIG_PM_MIN_WAKEUP_TIME */
 
 	now = clock_systimer();
-	if (now - g_pmglobals.stime >= TIME_SLICE_TICKS) {
+	if (now - g_pmglobals.stime >= MSEC2TICK(CONFIG_PM_MIN_WAKEUP_TIME)) {
```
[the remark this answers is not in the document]
gSahitya-samsung: We are removing it in PR #6293

ewoodev (approved):

PR #6258, opened 2024-06-25
os/fs/procfs, os/kernel/init: Fix issue in running the ps command with smp configs enabled

neel-samsung:
```ps``` command was not working properly with smp configs enabled. Idle task stack are not stored in heap region, instead they are stored in .bss region. For retrieving idle task stack, we should not search in heap region And this case was not handled properly for all cpu's idle task. So, we handle the case for every cpu's idle task.

### DIFF
--- a/os/fs/procfs/fs_procfsproc.c
+++ b/os/fs/procfs/fs_procfsproc.c
@@ -385,7 +385,7 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 	curr_heap = -1;
 	peak_heap = -1;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	if (tcb->pid == 0) {
+	if (tcb->pid < CONFIG_SMP_NCPUS) {
 		/* Idle task normally uses heap with index 0. */
 		heap = kmm_get_heap_with_index(0);
 	} else {
--- a/apps/system/utils/utils_ps.c
+++ b/apps/system/utils/utils_ps.c
@@ -72,6 +72,9 @@ static const char *utils_statenames[] = {
 	"INVALID ",
 	"PENDING ",
 	"READY   ",
+#ifdef CONFIG_SMP
+	"ASSIGNED",		
+#endif
 	"RUNNING ",
 	"INACTIVE",
 	"WAITSEM ",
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: os/fs/procfs/fs_procfsproc.c [hunk 1]
```diff
@@ -385,7 +385,7 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 	curr_heap = -1;
 	peak_heap = -1;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	if (tcb->pid == 0) {
+	if (tcb->pid < CONFIG_SMP_NCPUS) {
```
ritesh55555, line 388: do we need to see if #ifdef CONFIG_SMP is enabled?

ritesh55555 (approved):
looks good

edwakuwaku:
Hi @neel-samsung , could you please share the information printed when input "ps" to tash?

neel-samsung:
> Hi @neel-samsung. For reference, could you please share the information printed for SMP when input "ps" to tash,

I have added the ```ps``` command output to commit description. Please check.

edwakuwaku:
> > Hi @neel-samsung. For reference, could you please share the information printed for SMP when input "ps" to tash,
> 
> I have added the `ps` command output to commit description. Please check.

Sorry I have one question. As I checked your commit, the CPU0 Idle shows running, but CPU1 Idle shows inactive. I assume when you capture "ps" command, both cores should be in their own idle thread. Also, I think in SMP case, there should always be 2 tasks showing RUNNING state, because we have 2 cores running concurrently.
Could you please enlighten me on the above question?

neel-samsung:
> > > Hi @neel-samsung. For reference, could you please share the information printed for SMP when input "ps" to tash,
> > 
> > 
> > I have added the `ps` command output to commit description. Please check.
> 
> Sorry I have one question. As I checked your commit, the CPU0 Idle shows running, but CPU1 Idle shows inactive. I assume when you capture "ps" command, both cores should be in their own idle thread. Also, I think in SMP case, there should always be 2 tasks showing RUNNING state, because we have 2 cores running concurrently. Could you please enlighten me on the above question?

Thank you for addressing this.
So, there was task state which was missed due to which some of task state were not printed correctly. We also need to take care of Assign (Ready) state in case of SMP.

I have updated the code and ```ps``` command output. Now, in this case, tash and CPU1 idle are ```RUNNING```.

edwakuwaku (approved):
LGTM.

sunghan-chang (approved):

PR #6259, opened 2024-06-25
os/pm: Removes pm register related code from PM

ritesh55555:
This commit removes the pm register related code from pm. Before this change, each driver used to register pm_callback structure to notify PM during state change. That code part is removed now.

This PR uses the patch from https://github.com/Samsung/TizenRT/pull/6247


os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1284,6 +1284,7 @@ static uint32_t rtk_i2c_resume(uint32_t expected_idle_time, void *param)
 #ifdef CONFIG_PM
 void i2c_pminitialize(void)
 {
+	(void)pm_domain_register("I2C");
```
Taejun-Kwon, line 1287: I think enum is better option instead of string. then you can use like below.

```
enum pm_domain_e
{
	PM_DOMAIN_I2C = 0,
	PM_DOMAIN_I2S,
	...
	PM_DOMAIN_MAX,
};

int pm_domain_register(char *domain) {
	...
	for (index = 0; index < PM_DOMAIN_MAX; index++) {
		...
	}
	...
}
```
Then you should not define CONFIG_PM_NDOMAINS.
Is there any specific reason for string value?

ewoodev: We don't know which app to use PM.
And there was a request that it was not right to change the enum provided by the interface frequently.

We are going to operate both static and dynamic tables in the future for performance
It seems necessary to modify the name of CONFIG_PM_NDOMAINS.
This is simply a static table size.

os/pm/pm_changestate.c
```diff
@@ -227,23 +118,10 @@ int pm_changestate(enum pm_state_e newstate)
...
 	 */
 	if ((newstate != PM_RESTORE) && (newstate != g_pmglobals.state)) {
-		ret = pm_prepall(newstate);
-		if (ret != OK) {
-			/* One or more drivers is not ready for this state change.  Revert to
-			* the preceding state.
-			*/
-
-			g_pmglobals.recommended = g_pmglobals.state;
-			g_pmglobals.btime = clock_systimer();
```
gSahitya-samsung, line 237: If g_pmglobals.btime is not needed, can you remove it from header too.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -112,21 +112,32 @@ int SOCPS_AONWakeReason(void)
...
-		gtimer_start_one_shout(&g_timer1, interval_us, (void *)pg_timer_int_handler, (uint32_t)&g_timer1);
+		gtimer_start_one_shout(&g_timer1, interval_us, NULL, (uint32_t)&g_timer1);
 	}
 	return;
 }
 
+void bsp_pm_domain_control(char *domain, bool is_suspend)
+{
+	/* Retrive the domain id from the domain string */
+	int domain_id = pm_domain_register(domain);
```
ewoodev, line 129: Wrapping this again doesn't seem right.

after calling pm_suspend, pm_resume must be called.
there is no need to call the pm_domain_register redundantly.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
```diff
@@ -112,21 +112,32 @@ int SOCPS_AONWakeReason(void)
...
+	if (is_suspend) {
+		if (pm_suspend(domain_id) != OK) {
+			pmdbg("Unable to suspend %s domain!\n", domain);
```
ewoodev, line 132: if pm_suspaned is fail, pm_resume must not be called.

os/arch/arm/src/amebasmart/amebasmart_pmhelpers.c
edwakuwaku: The place to register the domain and the exact operation to utilize driver is in different spot, and invoking it repeating is redundant, that's why I created another API to do that.
For instance: https://github.com/Samsung/TizenRT/pull/6247/files#diff-4f2740a1a1d4b82c7b9746da9a15a73362d83e7028ac4d45244ea72fc5fe683aR1287

os/pm/pm_changestate.c
```diff
@@ -227,23 +118,10 @@ int pm_changestate(enum pm_state_e newstate)
...
-		*/
-		pm_changeall(newstate);
 		g_pmglobals.state = newstate;
+		g_pmglobals.stime = clock_systimer();
 	}
-EXIT:
+
 	/* Restore the interrupt state */
 	leave_critical_section(flags);
 	return ret;
```
gSahitya-samsung, line 127: You can directly return OK here.  Also please remove pm_checkstate conditional check from pm_idle & other bsp code

gSahitya-samsung (changes requested):

edwakuwaku:
Hi @ritesh55555 , PR#6247 has been merged, you can rebase your PR now.
Thanks.

os/pm/pm_wakehandler.c
```diff
@@ -66,7 +66,6 @@ void pm_wakehandler(clock_t missing_tick, pm_wakeup_reason_code_t wakeup_src)
 
 	/* After wakeup change PM State to STANDBY and reset the time slice */
 	pm_changestate(PM_STANDBY);
-	g_pmglobals.stime = clock_systimer();
```
ewoodev, line 69: Please revert it

r-prabu (approved):

kishore-sn (approved):

os/pm/pm_changestate.c
```diff
@@ -186,69 +82,32 @@ static inline void pm_changeall(enum pm_state_e newstate)
...
-	irqstate_t flags;
-	int ret = OK;
-
-	/* Disable interrupts throught this operation... changing driver states
-	 * could cause additional driver activity that might interfere with the
-	 * state change.  When the state change is complete, interrupts will be
-	 * re-enabled.
-	 */
-
-	flags = enter_critical_section();
```
ewoodev, line 223: please revert it
```
g_pmglobals.state = newstate;
```

this code is needed to protect in critical section

os/pm/pm_procfs.c
```diff
@@ -315,41 +313,6 @@ static size_t power_curstate_read(FAR struct file *filep, FAR char *buffer, size
 	return copysize;
 }
 
-/****************************************************************************
- * Name: power_devices_read
- ****************************************************************************/
-
-static size_t power_devices_read(FAR struct file *filep, FAR char *buffer, size_t buflen)
```
ewoodev, line 322: Instead of erasing it, how about change to return domain strings?

os/pm/pm_register.c
```diff
@@ -1,103 +0,0 @@
...
- * Input parameters:
- *   callbacks - An instance of struct pm_callback_s providing the driver
- *               callback functions.
- *
- * Returned value:
- *    Zero (OK) on success; otherwise a negated errno value is returned.
- *
- ****************************************************************************/
-
-int pm_register(FAR struct pm_callback_s *callbacks)
```
ewoodev, line 86: Using the notification function of pm in the BSP code should be removed.

However, I think we need to prepare a new function that notifies the application.

This should be run with user thread context, not with pm running at kernel end. Like sigaction,

Could you let me know you opinion?

neel-samsung:
@ritesh55555 Looks like, the branch has some conflicts. Please Resolve it.

PR #6265, opened 2024-06-27
apps/examples/lcd_test/: add FPS calculation in sample lcd app

namanjain7:


apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd_lcd);
+		free(lcd_data);
+		return;
+	}
+
+	struct rtc_time start_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+	struct rtc_time end_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+
+	ioctl(fd, RTC_SET_TIME, (unsigned long)&start_time);
+	ioctl(fd, RTC_SET_TIME, (unsigned long)&end_time);
```
ewoodev, line 300: is it nessesary duplicated reset?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+	area.col_start = 0;
+	area.col_end = xres - 1;
+	area.stride = 2 * xres;
+	area.data = lcd_data;
+
+	sprintf(port, LCD_DEV_PATH, p);
+	fd_lcd = open(port, O_RDWR | O_SYNC, 0666);
+	if (fd_lcd < 0) {
+		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd_lcd);
+		free(lcd_data);
```
ewoodev, line 292: please add close(fd) before return

ewoodev: Pelase check return type

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+	fd = open("/dev/rtc0", O_RDWR);
+	if (fd < 0) {
+		printf("ERROR: LCD FPS test, Fail to open rtc.\n");
+		return;
+	}
+
+	len = xres * yres * 2 + 1;
+	uint8_t *lcd_data = (uint8_t *)malloc(len);
+	if (lcd_data == NULL) {
+		printf("malloc failed for lcd data : %d\n", len);
```
ewoodev, line 271: Please add close(fd) here also

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
 	free(lcd_data);
 }
 
+static int test_fps_helper(int color){
+	int fd = 0, fd_lcd = 0, p = 0;
```
sunghan-chang, line 258: violations of coding rules
```
static int ...(int color)
{
    int fd = 0;
    int fd_lcd = 0;
    int p = 0;
```

Let's rename fd to fd_rtc like fd_lcd.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+	}
+
+	struct rtc_time start_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+	struct rtc_time end_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+
+	ioctl(fd, RTC_SET_TIME, (unsigned long)&start_time);
+	ioctl(fd, RTC_SET_TIME, (unsigned long)&end_time);
+
+	ioctl(fd, RTC_RD_TIME, (unsigned long)&start_time);
+	for(int itr = 0; itr < FPS_TEST_FRAME_COUNT; itr++){
```
sunghan-chang, line 303: 1. coding rule
```
for (...) {
```
2. I don't know how lcd works with non-changed data (color). Are you able to change the color at every loop? For exampe, print red -> print yellow -> print red like that.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+	memcpy(&start_tm, (struct tm *)&start_time, sizeof(struct tm));
+	memcpy(&end_tm, (struct tm *)&end_time, sizeof(struct tm));
+
+	start = mktime(&start_tm);
+	end = mktime(&end_tm);
+
+	free(start_tm);
+	free(end_tm);
+	free(start_time);
+	free(end_time);
```
sunghan-chang, line 324: looks unnecessary. They are not dynamic allocated buffers.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,94 @@ static void test_bit_map(void)
...
+
+	if(difftime(end, start) != 0){
```
sunghan-chang, line 326: if (...) {

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,93 @@ static void test_bit_map(void)
 	free(lcd_data);
 }
 
+static int test_fps_helper(int color){
+	int fd_rtc = 0, fd_lcd = 0, p = 0;
```
sunghan-chang, line 258: below is missing.

violations of coding rules
```
static int ...(int color)
{
    int fd_rtc = 0;
    int fd_lcd = 0;
    int p = 0;
```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,93 @@ static void test_bit_map(void)
...
+		free(lcd_data);
+		close(fd_rtc);
+		return 0;
+	}
+
+	struct rtc_time start_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+	struct rtc_time end_time = RTC_TIME_INITIALIZER(1970, 1, 1, 0, 0, 0);
+
+	ioctl(fd_rtc, RTC_SET_TIME, (unsigned long)&start_time);
+	ioctl(fd_rtc, RTC_SET_TIME, (unsigned long)&end_time);
```
sunghan-chang, line 302: 1. As @ewoodev left, no need to set time twice.
2. If the RTC is already running, this example changes the RTC time. I think we can read current time for start and then read again at the end.

namanjain7: Yes it was unnecessary, I have removed one of them.

apps/examples/lcd_test/example_lcd.c
namanjain7: done

apps/examples/lcd_test/example_lcd.c
namanjain7: I am switching between red and blue for 1000 loop.

apps/examples/lcd_test/example_lcd.c
sunghan-chang: Not done yet. Could you check the comment again?

apps/examples/lcd_test/example_lcd.c
sunghan-chang: 2nd is not concerned. If you keep RTC_SET_TIME with 1970-01-01, the system time will be changed by example running. I think it's wrong. I think we don't need to set the time. Just read is enough. Please check it again

edwakuwaku:
Hi @namanjain7 , please add commit description, we don't allow empty message

apps/examples/lcd_test/example_lcd.c
namanjain7: sir, I have already removed the test_fps_helper function. Since the test is now running like putarea(red) -> putarea(blue) -> putarea(red). So, this function was no longer required.

namanjain7:
I have tested on 7785 lcd. It' running perfectly and below is the print log for FPS.

TASH>>lcd_test
TASH>>=== LCD demo ===
xres : 240, yres:320
count :1
xres : 240, yres:320
count :2
xres : 240, yres:320
count :3
xres : 240, yres:320
count :4
xres : 240, yres:320
count :5
FPS: 58.82

apps/examples/lcd_test/example_lcd.c
namanjain7: yes sir, I have tested, and ioctl were unnecessary. i have removed both the ioctl which does the set time.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,105 @@ static void test_bit_map(void)
 	free(lcd_data);
 }
 
+static void test_fps(void){
```
sunghan-chang, line 257: I left this comment third times.
This line has a coding rule violation. Please fix it.
```
static void test_fps(void){
-->
static void test_fps(void)
{
```
`{` should be located at next line for function.

namanjain7: done

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,106 @@ static void test_bit_map(void)
...
+	time_t end;
+	start = mktime((FAR struct tm *)&start_time);
+	end = mktime((FAR struct tm *)&end_time);
+
+	if (difftime(end, start) != 0) {
+		float fps = FPS_TEST_FRAME_COUNT / difftime(end, start);
+		printf("FPS: %.2f\n", fps);
+		return;
+	}
+	printf("Error: FPS test failed\n");
```
edwakuwaku, line 347: This log is ambiguous. I suggest to add more meaningful comment. In here, it seems it is failed due to the start and end time is not captured due to some reason..
```
	if (difftime(end, start) != 0) {
		float fps = FPS_TEST_FRAME_COUNT / difftime(end, start);
		printf("FPS: %.2f\n", fps);
		return;
	} else {
	        printf("FPS calculation failed! Please check start/end timediff!\n");
        }
```

namanjain7: done. I have added this printf which gives more clarified description.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,105 @@ static void test_bit_map(void)
...
+		return;
+	}
+
+	len = xres * yres * 2 + 1;
+	uint8_t *lcd_data_red = (uint8_t *)malloc(len);
+	uint8_t *lcd_data_blue = (uint8_t *)malloc(len);
+	if (lcd_data_red == NULL || lcd_data_blue == NULL) {
+		printf("FPS TEST, malloc failed for lcd data : %d\n", len);
+		close(fd_rtc);
+		return;
```
sunghan-chang, line 277: This could cause memory leak.
For example, the `lcd_data_red` is allocated, but `lcd_data_blue` is failed. Then, this if conditional will return without freeing `red`. Same as opposite case.
Need to check both separately and to free preallocated memory before return.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,105 @@ static void test_bit_map(void)
...
+	}
+	for (int i = 0; i < len-1; i += 2) {
```
sunghan-chang, line 279: coding rule
```
len-1
->
len - 1
```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +254,105 @@ static void test_bit_map(void)
...
+	time_t start;
+	time_t end;
+	start = mktime((FAR struct tm *)&start_time);
+	end = mktime((FAR struct tm *)&end_time);
+
+	if (difftime(end, start) != 0) {
+		float fps = FPS_TEST_FRAME_COUNT / difftime(end, start);
+		printf("FPS: %.2f\n", fps);
+	} else {
+		printf("FPS calculation failed! Please check start/end timediff!\n");
```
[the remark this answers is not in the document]
sunghan-chang: > 1 frame is taking like 16 ms. So if the number of frames executed is less than 62, then the start and end will be same because rtc records time in seconds.

I want to print `failed, so how to resolve`. When we meet this case, you can increase `FPS_TEST_FRAME_COUNT` to resolve the failure. If right, let's explain this. But this will requires config for FPS_TEST_FRAME_COUNT definition.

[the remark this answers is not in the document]
namanjain7: ok..  so,   how about if we make a config "CONFIG_EXAMPLE_LCD_FPS_TEST". If this value is not defined, then we won't execute this test, otherwise this test case will be executed with the value defined.
CONFIG_EXAMPLE_LCD_FPS_TEST=1000

apps/examples/lcd_test/example_lcd.c
```diff
@@ -67,10 +68,12 @@
 #define COLINDEX 10
 #define ROWINDEX 10
 #define NOPIXELS 200
-
 static int xres;
 static int yres;
 
+#ifndef EXAMPLE_LCD_FPS_TEST
+#define EXAMPLE_LCD_FPS_TEST 5000
+#endif
```
sunghan-chang, line 76: This should be
```
#ifdef CONFIG_EXAMPLE_LCD_FPS_TEST
#define EXAMPLE_LCD_FPS_TEST CONFIG_EXAMPLE_LCD_FPS_TEST
#else
#define EXAMPLE_LCD_FPS_TEST 5000
#endif
```

apps/examples/lcd_test/Kconfig
```diff
@@ -9,6 +9,14 @@ config EXAMPLES_LCD
 	---help---
 		Enable LCD Example
 
+if EXAMPLES_LCD
+config EXAMPLE_LCD_FPS_TEST
+	int "LCD FPS test"
```
sunghan-chang, line 14: In the menuconfig, this `" "` sentence explains what it is. But I can't get it from the sentence.
```
"the number of frames execution"
```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -254,13 +256,114 @@ static void test_bit_map(void)
...
+	start = mktime((FAR struct tm *)&start_time);
+	end = mktime((FAR struct tm *)&end_time);
+
+	int time_elapsed = difftime(end, start);
+	if (time_elapsed != 0) {
+		float fps = EXAMPLE_LCD_FPS_TEST / time_elapsed;
+		printf("FPS Test: %d frames executed in %d sec, FPS: %.2f\n", EXAMPLE_LCD_FPS_TEST, time_elapsed, fps);
+		return;
+	} else {
+		printf("FPS calculation failed! Please increase the number of frames execution!\n");
```
sunghan-chang, line 356: printf("FPS calculation failed! Please increase the number of frames execution using CONFIG_EXAMPLE_LCD_FPS_TEST!\n");

sunghan-chang: I assume some other person can run this app so that we should guide detail as possible. So I wanna add `CONFIG_XXXX` in the message as above.

sunghan-chang (approved):

PR #6268, opened 2024-07-01
os/driver/lcd: initial black screen added in INIT ioctl

anjana348:
To avoid garbage frame on screen, keep display off untill first frame call

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -208,7 +208,6 @@ static void rtl8730e_enable_lcdc(void)
 {
 	LCDC_Cmd(pLCDC, ENABLE);
 	while (!LCDC_CheckLCDCReady(pLCDC)) ;
-	mipidsi_mode_switch(true);
 }
 
 void rtl8730e_mipidsi_underflowreset(void)
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -36,6 +36,11 @@
 #define CONFIG_LCD_MAXPOWER 100
 #endif
 
+typedef enum {
+	COMMAND,
+	VIDEO,
+} LCD_MODE;
+
 struct mipi_lcd_dev_s {
 	/* Publicly visible device structure */
 
@@ -48,34 +53,29 @@ struct mipi_lcd_dev_s {
 	int fb_alloc_count;
 	uint8_t power;
 	struct mipi_lcd_config_s *config;
+	LCD_MODE mode;
 };
 
 static struct mipi_lcd_dev_s g_lcdcdev;
 
-static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
+static int send_cmd(struct mipi_lcd_dev_s *priv, bool display_on)
 {
-	int transfer_status = OK;
-	u8 cmd = command.cmd;
-	u8 *cmd_addr = command.para_list;
-	u32 payload_len = command.count;
-	struct mipi_dsi_msg msg;
-	msg.channel = cmd;
-	if (payload_len == 0) {
-		msg.type = MIPI_DSI_DCS_SHORT_WRITE_0_PARAM;
-	} else if (payload_len == 1) {
-		msg.type = MIPI_DSI_DCS_SHORT_WRITE_1_PARAM;
-	} else {
-		msg.type = MIPI_DSI_DCS_LONG_WRITE;
-	}
-	msg.tx_buf = cmd_addr;
-	msg.tx_len = payload_len;
-	msg.flags = 0;
 	priv->config->lcd_mode_switch(false);
+	int transfer_status = OK;
+	struct mipi_dsi_msg msg;
+	if (display_on) {
+		msg.channel = 0x29;
+	} else {
+		msg.channel = 0x28;
+	}
+	msg.type = MIPI_DSI_DCS_LONG_WRITE;
+	msg.tx_len = 0;
+	msg.flags = 0;
 	transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
-	priv->config->lcd_mode_switch(true);
 	if (transfer_status != OK) {
 		lcddbg("Command %x not sent \n", cmd);
 	}
+	priv->config->lcd_mode_switch(true);
 	return transfer_status;
 }
 
@@ -167,11 +167,10 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
 static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord_t row_end, fb_coord_t col_start, fb_coord_t col_end, FAR const uint8_t *buffer, fb_coord_t stride)
 {
 	FAR struct mipi_lcd_dev_s *priv = (FAR struct mipi_lcd_dev_s *)dev;
-	//coordinate shift from (0,0) -> (1,1) and (XRES-1,YRES-1) -> (XRES,YRES)
-	row_start += 1;
-	row_end += 1;
-	col_start += 1;
-	col_end += 1;
+	if (priv->mode == COMMAND) {
+		send_cmd(priv, true);
+		priv->mode = VIDEO;
+	}
 	priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
 	return OK;
 }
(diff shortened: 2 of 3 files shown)


os/drivers/lcd/mipi_lcd.c
```diff
@@ -278,14 +280,15 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 static int lcd_init(FAR struct lcd_dev_s *dev)
 {
 	FAR struct mipi_lcd_dev_s *priv = (FAR struct mipi_lcd_dev_s *)dev;
-	priv->config->reset();
-	if (send_init_cmd(priv, lcd_init_cmd_g) == OK) {
-		lcdvdbg("LCD Init sequence completed\n");
-	} else {
-		lcddbg("ERROR: LCD Init sequence failed\n");
+	int len = LCD_XRES * LCD_YRES * 2;
+	uint8_t *buffer = malloc(len);
```
edwakuwaku, line 284: Please add condition check for malloc failed.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -85,6 +93,7 @@ LCDC_TypeDef *pLCDC = LCDC;
 LCDC_InitTypeDef lcdc_init_struct;
 static u32 UnderFlowCnt = 0;
 static u8 lcdc_nextframe = 0;
+uint8_t *lcd_data = NULL;
```
edwakuwaku, line 96: static

sunghan-chang:
@anjana348 With this PR, no more garbage frame on screen?

anjana348:
> @anjana348 With this PR, no more garbage frame on screen?

Yes, but buffer is freed in INIT ioctl so if INIT ioctl is not called, it will keep occupying memory.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -204,6 +213,31 @@ static void rtl8730e_control_backlight(uint8_t level)
...
+	if (lcd_data == NULL) {
+		lcddbg("malloc failed for lcd data : %d\n", len);
+		rtl8730e_control_backlight(CONFIG_LCD_MAXPOWER);
+		return;
+	}
+	for (int i = 0; i < len; i += 2) {
+		lcd_data[i] = (INITCOLOR & 0xFF00) >> 8;
+		lcd_data[i + 1] = INITCOLOR & 0x00FF;
+	}
+	rtl8730e_lcd_put_area(lcd_data, 0, 0, xres, yres);
```
edwakuwaku, line 229: In here you are using put area, which is in video mode to update black screen during initialization.
As I heard, the request seems only rendering one time, and when UIFW done initializing, it will start to update the frame.
Thus, the flow here should be changed to update the pixels by command mode, instead of video mode (ie. Use 0x2C command, and throw the buffer into the msg).

edwakuwaku:
```
msg_col.channel = 0x2C;
msg_col.type = MIPI_DSI_DCS_LONG_WRITE;
msg_col.tx_buf = lcd_data;
msg_col.tx_len = len;
msg_col.flags = 0;
```

anjana348: What will be benefit of using command mode for this instead of video mode?
We have to malloc and free buffer in both cases.

sunghan-chang: No need to wait time to free. With the video mode, we must get the init command by upper layer. If it's missing in upper layer, it will be a leak. It means it has a dependency. But if we use the command mode, we can free it after sending. No need to get dependency with upper layer. It's what I understand.

anjana348: @edwakuwaku 
We are finding strange behaviour when using cmd mode for sending data during initialization. Please confirm
1. After sending data via cmd mode, LCD does not show content till we change to video mode.
2. LCD shows noise, when we free the memory(lcd_data here) for cmd mode data
msg_col.tx_buf = lcd_data;
3. Is there any need to initalize the memory pointer(lcdc_init_struct.layerx[idx].LCDC_LayerImgBaseAddr) even when we are sending data via cmd mode

zhongnuo-tang: Hi @anjana348,
If you want to use cmd mode to update the frame, you need to program the LCM to cmd mode.  
{0xB0, {0x00, 0x00, 0x00, 0xF1}} before using cmd mode to send the image buffer.

anjana348: @zhongnuo-tang 
I tried below sequence, but no data refresh on LCD
MIPI CMD Mode
Send {0xB0, {0x00, 0x00, 0x00, 0xF1}} 
Send 0x2A (coordinates), 0x2B (coordinates)
Send 0x2C (lcd buffer)
Did you verify this working? Or do I need to change anything else?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -48,34 +53,160 @@ struct mipi_lcd_dev_s {
...
+{
+	int transfer_status = OK;
+	struct mipi_dsi_msg msg_data, msg;
+	uint32_t size = CONFIG_AMEBASMART_MIPI_PAYLOAD_LEN;
+	fb_coord_t curr_row_start = 0;
+	fb_coord_t curr_row_end;
+	fb_coord_t curr_col_start = 0;
+	fb_coord_t curr_col_end = LCD_XRES - 1;
+	fb_coord_t width_size = LCD_XRES * 2;
+	uint8_t *init_buffer = (uint8_t *)kmm_malloc(CONFIG_AMEBASMART_MIPI_PAYLOAD_LEN);
```
edwakuwaku, line 134: Hello @anjana348, in TizenLite driver structure, we cannot apply chipset specific config directly. Thus, I think some alternative method should be applied.
For example, adding another variable such as `u32 mipi_buffer_max_size;` in the `struct mipi_lcd_dev_s`. And then during initialization, the chipset specific code can pass in the variable and store in the struct.
eg: `priv-> mipi_buffer_max_size = variable passed in from chipset BSP`
So, all the condition check can rely on the priv structure instead of chipset specific.

sunghan-chang:
@anjana348 Is this ready? Update is done?

anjana348:
> @anjana348 Is this ready? Update is done?

yes, you can review the code.
Command mode is being used to send initial black color for screen

os/drivers/lcd/mipi_lcd.c
```diff
@@ -48,34 +53,159 @@ struct mipi_lcd_dev_s {
...
+		curr_row_end = curr_row_start + (priv->config->mipi_buffer_max_size / width_size);
+		if (curr_row_end > LCD_YRES) {
+			curr_row_end = LCD_YRES;
+		}
+		if (size > (curr_row_end - curr_row_start) * LCD_XRES * 2) {
+			size = (curr_row_end - curr_row_start) * LCD_XRES * 2;
+		}
+		curr_row_end--;
+		uint8_t cmd_col_addr[8] = {0x00, (curr_col_start >> 8) & 0xFF, 0x00, curr_col_start & 0xFF, 0x00, (curr_col_end >> 8) & 0xFF, 0x00, curr_col_end & 0xFF};
+		msg.channel = 0x2A;
```
edwakuwaku, line 157: Hello @anjana348 , I don't think the command id to configure LCD parameters are the same across different vendors. Thus, it seems not a good way to hardcode here (ie. including both channel id and cmd size).
As I understood, the command id seems vendor specific and not defined as to MIPI protocol.
Thus, could you please confirm this? If yes, we might need to change to be set dynamically in header or Kconfig.

anjana348: Hi @edwakuwaku 
I have changed logic of keeping initial garbage screen
We can keep display off till first frame comes, this doesn't need extra overhead of sending data through command mode.
Is there any issue with this way ?

edwakuwaku: Hello @anjana348 , it should be fine. But, what I mean is in the os/drivers/lcd layer. The msg.channel is hardcoded to specific index (ie  0x28, 0x29, 0x2a, 0x2b... etc). I don't think it is good way to maintain that, because that seems vendor device specific index, instead of MIPI protocol. Which means if we switch a LCD model, this code might not be compatible.
@sunghan-chang , could you please advise here?

sunghan-chang:
Let's apply Samsung logo instead of this. Let me close this.

PR #6269, opened 2024-07-01
os/board/rtl8730e: Update BSP driver irq priority running on cortex-A

edwakuwaku:
- Currently the driver priority setting is according to cortex-M value range (ie. NVIC), but rtl8730e is running on cortex-A, which has it's own value range for setting irq priority (ie. GIC).
- This PR revise the wrong usage on irq priority for peripheral drivers
- Due to performance issue, Increase irq priority level for LCDC

### DIFF
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_sport.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_sport.c
@@ -1100,7 +1100,7 @@ BOOL AUDIO_SP_TXGDMA_Init(
 	assert_param(GDMA_InitStruct != NULL);
 	DCache_CleanInvalidate((u32)pTXData, Length);
 	/*obtain a DMA channel and register DMA interrupt handler*/
-	GdmaChnl = GDMA_ChnlAlloc(0, CallbackFunc, (u32)CallbackData, INT_PRI_MIDDLE);
+	GdmaChnl = GDMA_ChnlAlloc(0, CallbackFunc, (u32)CallbackData, INT_PRI_HIGH);
 	if (GdmaChnl == 0xFF) {
 		// No Available DMA channel
 		return _FALSE;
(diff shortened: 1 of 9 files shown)


os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -89,7 +89,7 @@ static u8 lcdc_nextframe = 0;
 struct irq lcdc_irq_info = {
 	.num = LCDC_IRQ,
 	.data = (u32) LCDC,
-	.priority = INT_PRI_MIDDLE,
+	.priority = INT_PRI_HIGH,
 };
 
 extern struct irq mipi_irq_info;
```
kishore-sn, line 95: Do we need to increase the priority of mipi also? It is currently set to middle.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_hp/ameba_sport.c
```diff
@@ -1312,7 +1312,7 @@ BOOL AUDIO_SP_LLPTXGDMA_Init(
 	u32 pTxData = Lli[0].LliEle.Sarx;
 
 	/*obtain a DMA channel and register DMA interrupt handler*/
-	GdmaChnl = GDMA_ChnlAlloc(0, CallbackFunc, (u32)CallbackData, INT_PRI_MIDDLE);
+	GdmaChnl = GDMA_ChnlAlloc(0, CallbackFunc, (u32)CallbackData, INT_PRI_HIGH);
```
kishore-sn, line 1315: Why are we increasing the priority of dma irq? What effect will this have?

kishore-sn (commented):
How about other places in code where we are using the old system?
For example amebasmart_serial.c, etc..

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
edwakuwaku: MIPI irq is only used during sending init command table. When running LCD app, only LCD irq is being applied.

edwakuwaku:
> How about other places in code where we are using the old system? For example amebasmart_serial.c, etc..

Ya, that should be changed as well... Thank you for reminding

edwakuwaku:
Hello @r0bindal , thanks for updating the results.
May I know the LCD result is tested individually without Bixby audio? For this PR, I expect that when LCD + Bixby is up and running together with other driver modules, both LCD and Bixby can be prioritised.
I am not sure whether are you aware of that, but currently the parameter settings for LCD is according to 60FPS. Thus, the test results is making sense to me.

sunghan-chang:
@edwakuwaku We've verified that this and PR #6271 resolves external flash hang. I am going to merge if possible. Is there any reason to mark `DO NOT MERGE`?

edwakuwaku:
> @edwakuwaku We've verified that this and PR #6271 resolves external flash hang. I am going to merge if possible. Is there any reason to mark `DO NOT MERGE`?

In the second commit of this PR, I would like to check whether it works better if we increase the irq priority for driver used by bixby audio and LCD. But it seems it was not tested together yet.
Anyway, since that will not affect normal operation, it can be merged.

sunghan-chang (approved):

PR #6271, opened 2024-07-03
os/arch/arm/src/amebasmart: Update missing leave_critical_section

edwakuwaku:
- Missing leave_critical_section for SPI initialization lead to several unexpected behaviour, thus add leave_critical_section to necessary places (ie. early return condition)


sunghan-chang:
@edwakuwaku With this and PR #6269 , we still get the same hang issue. So now, @ewoodev are checking where it gets stuck.

Tirtha-Samsung:
This PR ```os/arch/arm/src/amebasmart: Update missing leave_critical_section``` looks good to merge. Verified with these 2 PRs
https://github.com/Samsung/TizenRT/pull/6269 and https://github.com/Samsung/TizenRT/pull/6237 and it is not hanging

sunghan-chang (approved):

PR #6273, opened 2024-07-05
os/pm: fix pm_timedsuspend cancel

gSahitya-samsung:
When a thread call pm_timedsuspend multiple times within timer expiration, then pm_timedsuspend doesn't stoping the timer, instead it was rasing error. Now when pm_timedsuspend call multiple times within timer expiration, then it changes timer duration to remain suspended for given time if it is greater than current expiration time. If application provide zero time interval, then it cancel the timer immediately and resume the suspended domain.

### DIFF
--- a/os/pm/pm_timedsuspend.c
+++ b/os/pm/pm_timedsuspend.c
@@ -70,26 +70,23 @@
  * Private Variables
  ************************************************************************/
 
-/* This array maps the pid to their respective wdog timer. 
- * Assumption: when a thread does a timed lock , only the same thread 
- * can unlock the "timed lock" before it expire */
-static WDOG_ID pid_timer_map[CONFIG_MAX_TASKS];
+/* This array maps the domain to their respective wdog timer */
+static WDOG_ID domain_timer_map[CONFIG_PM_NDOMAINS];
 
 /************************************************************************
  * Private Functions
  ************************************************************************/
 
-static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
+static void timer_timeout(int argc, int domain_id)
 {
-	DEBUGASSERT(pid_timer_map[hash_pid] != NULL);
+	DEBUGASSERT(domain_timer_map[domain_id] != NULL);
 	/* PM transition will be resume here */
 	if (pm_resume(domain_id) != OK) {
 		pmlldbg("Unable to resume domain: %s\n", pm_domain_map[domain_id]);
 	}
-	(void)wd_delete(pid_timer_map[hash_pid]);
-	pid_timer_map[hash_pid] = NULL;
+	(void)wd_delete(domain_timer_map[domain_id]);
+	domain_timer_map[domain_id] = NULL;
 }
-
 /************************************************************************
  * Public Functions
  ************************************************************************/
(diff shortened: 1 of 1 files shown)


os/pm/pm_timedsuspend.c
```diff
@@ -79,17 +79,33 @@ static WDOG_ID pid_timer_map[CONFIG_MAX_TASKS];
...
+{
+	WDOG_ID wdog;
+	uint32_t pid;
+	irqstate_t flags;
+	/* Don't let actual timer to expire during this process */
+	flags = enter_critical_section();
+	pid = getpid();
+	wdog = pid_timer_map[pid];
+	if (wdog) {
+		timer_timeout(2, wdog->parm[0], pid);
```
kishore-sn, line 103: What is the hardcoded value 2? Can we either use a preprocessor definition for it or add a comment?

os/pm/pm_timedsuspend.c
```diff
@@ -112,18 +128,17 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
 		return ERROR;
 	}
 
 	/* Check if there is already a wdog lock timer running for 
-	 * the process */
-	if (pid_timer_map[hash_pid] != NULL) {
-		pmdbg("There is already a lock timer running for this process pid %d with domain %s\n", getpid(), pm_domain_map[domain_id]);
-		return ERROR;
+	 * the process, if so then cancel it */
+	if (pid_timer_map[pid] != NULL) {
```
kishore-sn, line 140: I dont think this check is necessary here. You are already checking it inside the function.

os/pm/pm_timedsuspend.c
gSahitya-samsung: `2` is no. of argument

os/pm/pm_timedsuspend.c
```diff
@@ -79,17 +79,33 @@ static WDOG_ID pid_timer_map[CONFIG_MAX_TASKS];
...
 }
 
+static void pm_timedsuspend_cancel(void)
+{
+	WDOG_ID wdog;
+	uint32_t pid;
+	irqstate_t flags;
+	/* Don't let actual timer to expire during this process */
+	flags = enter_critical_section();
+	pid = getpid();
```
abhishek-samsung, line 100: looks like this getpid is unnecessary as we already fetch the pid before calling this function. It would be better to pass the pid as an argument instead.

os/pm/pm_timedsuspend.c
```diff
@@ -150,7 +161,7 @@ int pm_timedsuspend(int domain_id, unsigned int milliseconds)
 	pmvdbg("PM is locked for pid %d and timer started for %d milliseconds\n", getpid(), milliseconds);
 	return OK;
 errout:
-	pid_timer_map[hash_pid] = NULL;
-	(void)wd_delete(pid_timer_map[hash_pid]);
+	pid_timer_map[pid] = NULL;
+	(void)wd_delete(pid_timer_map[pid]);
```
abhishek-samsung, line 165: This errout looks wrong. We are passing NULL to wd_delete.

os/pm/pm_timedsuspend.c
gSahitya-samsung: If we use `getpid()` instead of `pid` argument, we can use this function for both internal and external API

sunghan-chang: What do you mean the `external API`? This API is in os/pm and is a `static`.

gSahitya-samsung: Yes, but if in future if we want to give this as an API to other application. We can simply provide them without changing the code logic.

os/pm/pm_timedsuspend.c
kishore-sn: Ok

os/pm/pm_timedsuspend.c
```diff
@@ -106,51 +103,62 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
  ************************************************************************/
 
 int pm_timedsuspend(int domain_id, unsigned int milliseconds)
 {
-	int hash_pid = PIDHASH(getpid());
-
-	if (domain_id >= CONFIG_PM_NDOMAINS || pm_domain_map[domain_id]==NULL) {
+	irqstate_t flags;
+	WDOG_ID wdog;
+	int ret = OK;
```
ewoodev, line 114: let's change init value OK to ERROR

kishore-sn (dismissed):

os/pm/pm_timedsuspend.c
```diff
@@ -106,51 +103,62 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
-		
+	/* New delay is less than running timer, no need to do anyting */
+	if (delay <= wd_gettime(wdog)) {
+		pmvdbg("Domain: %s is already suspended for %d milliseconds\n", pm_domain_map[domain_id], TICK2MSEC(wd_gettime(wdog)));
+		goto exit;
 	}
-	if (wd_start(wdog, MSEC2TICK(milliseconds), (wdentry_t)timer_timeout, 2, domain_id, (uint32_t)hash_pid) != OK) {
+	/* Start the timer */
+	if (wd_start(wdog, delay, (wdentry_t)timer_timeout, 1, domain_id) != OK) {
 		pmvdbg("Error starting Wdog timer\n");
```
edwakuwaku, line 154: Let's use pmdbg for error log

ewoodev (approved):

ewoodev:
@r-prabu @kishore-sn Could you review and merge?

os/pm/pm_timedsuspend.c
```diff
@@ -106,51 +103,64 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
-	if (pid_timer_map[hash_pid] != NULL) {
-		pmdbg("There is already a lock timer running for this process pid %d with domain %s\n", getpid(), pm_domain_map[domain_id]);
-		return ERROR;
+	flags = enter_critical_section();
+	wdog = domain_timer_map[domain_id];
+	/* If delay is zero then cancel the timer (PM Policy) */
+	if (delay == 0) {
+		if (wdog) {
+			timer_timeout(NULL, domain_id);
+		}
```
r-prabu, line 127: Why when delay == 0, assume timer is already created?. can we check the negative case at start and exit if delay is 0 & no timer available?

os/pm/pm_timedsuspend.c
```diff
@@ -106,51 +103,64 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
-
-	if (domain_id >= CONFIG_PM_NDOMAINS || pm_domain_map[domain_id]==NULL) {
+	irqstate_t flags;
+	WDOG_ID wdog;
+	int tick_remain;
+	int ret = ERROR;
+	int delay = MSEC2TICK(milliseconds);
+	if ((domain_id < 0) || (domain_id >= CONFIG_PM_NDOMAINS) || (pm_domain_map[domain_id] == NULL)) {
 		set_errno(EINVAL);
-		return ERROR;
```
r-prabu, line 119: better to add pmdbg messages on error cases.

os/pm/pm_timedsuspend.c
gSahitya-samsung: There is PM Policy to cancel the domain timedsuspend timer if delay is 0.

Delay can't be negative since `milliseconds` argument is of type `uint`.

r-prabu: negative case to check, which i mentioned is delay 0 & no timer  (!wdog) at the start.

gSahitya-samsung: In that case we simply exiting from the code `line:128`

gSahitya-samsung: If we check delay 0 & no timer (!wdog) at start, we need to again check delay==0 for the delay & timer (wdog) case. So, I think nested if-clause is better for less comparisions

os/pm/pm_timedsuspend.c
```diff
@@ -106,51 +103,65 @@ static void timer_timeout(int argc, int domain_id, uint32_t hash_pid)
...
-	return OK;
-errout:
-	pid_timer_map[hash_pid] = NULL;
-	(void)wd_delete(pid_timer_map[hash_pid]);
-	return ERROR;
+	pmvdbg("Domain: %s is suspended for %d milliseconds\n", pm_domain_map[domain_id], milliseconds);
+	ret = OK;
+exit:
+	leave_critical_section(flags);
+	return ret;
```
Tirtha-Samsung, line 166: This PR ```os/pm: fix pm_timedsuspend cancel``` looks good to me

r-prabu (approved):

Tirtha-Samsung (approved):

kishore-sn (approved):

PR #6278, opened 2024-07-08
os/board/rtl8721csm: fix BLE function array mismatch due to PR6193

yeetee179:
- NULL added to g_trble_drv_ops to match function in trble_ops struct

### DIFF
--- a/os/board/rtl8721csm/src/component/os/tizenrt/rtk_blemgr.c
+++ b/os/board/rtl8721csm/src/component/os/tizenrt/rtk_blemgr.c
@@ -144,6 +144,7 @@ struct trble_ops g_trble_drv_ops = {
 	trble_netmgr_get_profile_count,
 	trble_netmgr_charact_notify,
 	trble_netmgr_charact_indicate,
+	NULL,
 	trble_netmgr_attr_set_data,
 	trble_netmgr_attr_get_data,
 	trble_netmgr_attr_reject,


sunghan-chang:
@yeetee179 Is this fix of some commit which was merged before?

yeetee179:
Yes, the PR for AIdual https://github.com/Samsung/TizenRT/pull/6193 has modified trble_ops, which will affect TP1x as well.

sunghan-chang:
@yeetee179 How about writing that at the title? Then, we can easily find related change and can easily find merging together for other repository.

yeetee179:
@sunghan-chang  I have modified title to include the AIdual PR number that cause this issue in TP1x

os/board/rtl8721csm/src/component/os/tizenrt/rtk_blemgr.c
```diff
@@ -486,6 +487,7 @@ trble_result_e trble_netmgr_stop_adv(struct bledev *dev)
 
 trble_result_e trble_netmgr_start_adv(struct bledev *dev)
 {
+	printf("[######## %s : %d]\n", __FUNCTION__, __LINE__);
```
hs36-kim, line 490: would you remove this comment if not necessary?

hs36-kim (changes requested):

os/board/rtl8721csm/src/component/os/tizenrt/rtk_blemgr.c
yeetee179: I have removed the line above

hs36-kim (approved):
LGTM

PR #6279, opened 2024-07-08
build/tools/amebasmart/gnu_utility: move pinmap_init from KM4 to AP

zhongnuo-tang:
1. remove pinmap_init in KM4
2. add pinmap_init in AP app start to prevent PUPD status overwritten to default due to timing issue


os/board/rtl8730e/src/component/usrcfg/rtl8730e/Make.defs
```diff
@@ -52,7 +52,8 @@
 CSRCS += ameba_intfcfg.c \
 	 ameba_sleepcfg.c \
 	 ameba_wificfg.c \
-	 ameba_flashcfg.c 
+	 ameba_flashcfg.c \
+	 ameba_pinmapcfg.c
```
sunghan-chang, line 56: requires `\` at EOL

edwakuwaku (approved):
LGTM.

sunghan-chang (commented):
@zhongnuo-tang When we merge this, os/board/rtl8730e/src/component/usrcfg/rtl8730e/ameba_pinmapcfg.c is the file to configure pin map actually used?

zhongnuo-tang:
> @zhongnuo-tang When we merge this, os/board/rtl8730e/src/component/usrcfg/rtl8730e/ameba_pinmapcfg.c is the file to configure pin map actually used?

Hi @sunghan-chang, This is the file that configure the initial PIN status, but subsequently usage of the PIN can still modify the PUPD.

sunghan-chang:
@zhongnuo-tang Just question (not suggestion, not request to modify),
why don't you maintain pin setting like gpio, alternative function in the same table?
If you do, we can prevent duplicated setting and can check entire pin usage.

sunghan-chang (approved):

PR #6280, opened 2024-07-09
[DO NOT MERGE] os/board/rtl8730e/src: MIPI command mode implementation

edwakuwaku:
- MIPI command mode implementation


os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -48,6 +48,8 @@
 static u32 vo_freq;
 static u32 send_cmd_done = 0;
 
+#define MIPI_PAYLOAD_MAX_LEN   3840    /* (480 * 64) bits */
```
Taejun-Kwon, line 51: If it never changed based on type of LCD, it should be defined in mipi_dsi.h or some header, because both of here and mipi_lcd.c refer to this.
But it seems that it's not referred by code here.

edwakuwaku: Hello @Taejun-Kwon , this is a PoC on using MIPI command mode for frame update. As you can see this PR is still messy, many places still require cleanup to follow the structure it supposed to (ie. driver layer). It is still pending for discussion on how to revise the LCD driver to support both MIPI command + video mode during runtime.
Also, this definition is not a MIPI protocol definition, whereas it is specific to chipset MIPI driver, RTL8730E has a MIPI buffer of size = 480 * 64 bits. I added this just for the convenience of initial implementation.
Anyway, thank you for your review~

edwakuwaku:
Hello @anjana348 , I have rebased my PR~

anjana348:
Hi @edwakuwaku 
When we are updating full frame using command mode, the last pixels of each subarea remains unwritten.
Means we are sending data of length 3840 , it first set coordinates
0 0 5 319 so the last pixel can be seen without data, it looks like noise on boundary.
Can you please check on your side ?

edwakuwaku:
> Hi @edwakuwaku When we are updating full frame using command mode, the last pixels of each subarea remains unwritten. Means we are sending data of length 3840 , it first set coordinates 0 0 5 319 so the last pixel can be seen without data, it looks like noise on boundary. Can you please check on your side ?

Hello Anjana, we need to carry the command id in the first byte of the buffer (ie. cmd_buffer[0]), thus, I think the max length should be (3840 - 2) bytes. Because we are using RGB565 (ie. 2 bytes) for each pixel.

PR #6284, opened 2024-07-10
os/pm: fix pm_sleep memory leak

gSahitya-samsung:
This commit resolve memory leak issue with pm_sleep API. pm_sleep allocating timer from heap and releasing memory after timer expiration. In event of thread termination, the timer will remain active forever and cause memory leak. To prevent this issue now we are using tcb->waitdog. During the event of thread termination the task_recover API will call and it delete the timer to prevent memory leak.

### DIFF
--- a/os/pm/pm_sleep.c
+++ b/os/pm/pm_sleep.c
@@ -102,40 +102,47 @@ static void pm_timer_callback(int argc, uint32_t sem)
 
 int pm_sleep(int milliseconds)
 {
-	int ret;
-	WDOG_ID wdog;
 	sem_t pm_sem;
-
+	irqstate_t flags;
+	int ret = ERROR;
+	/* TODO - Since PM & Kernel are separate, we should not use tcb inside pm.
+	 * We need to remove tcb in future.
+	 */
+	FAR struct tcb_s *rtcb = this_task();
 	/* initialize the timer's semaphore. It will be used to lock the
 	 * thread before sleep and unlock after expire */
 	sem_init(&pm_sem, 0, 0);
-
-	wdog = wd_create();
-	if (!wdog) {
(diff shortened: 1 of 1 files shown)


ewoodev (approved):

os/pm/pm_sleep.c
```diff
@@ -102,40 +102,42 @@ static void pm_timer_callback(int argc, uint32_t sem)
...
 		ret = sem_wait(&pm_sem);
 		DEBUGASSERT(ret == 0 || errno == EINTR);
 	} while (ret < 0);
-
 	/* When the semaphore is freed, make the pm timer free */
-	wd_delete(wdog);
-
-    return OK;
+	wd_delete(rtcb->waitdog);
+	rtcb->waitdog = NULL;
```
kishore-sn, line 139: Let us move this line after errout so that we set waitdog to null in all fail cases.

os/pm/pm_sleep.c
```diff
@@ -102,40 +102,42 @@ static void pm_timer_callback(int argc, uint32_t sem)
...
+	if (!rtcb->waitdog) {
 		set_errno(EAGAIN);
-		return ERROR;
+		goto errout;
 	}
-
-	if (wd_setwakeupsource(wdog) != OK) {
-		wd_delete(wdog);
-		return ERROR;
+	/* set this timer as wakeup source */
```
r-prabu, line 120: can we have pmdbg messages on all the error paths?

ewoodev (approved):

sunghan-chang:
@gSahitya-samsung @ewoodev Need to discuss more for this.

ewoodev (approved):

sunghan-chang (approved):

PR #6289, opened 2024-07-11
Add API for setting channel plan in runtime

jwei5:
- Support setting of channel plan in runtime in km4
- Add API in CA32 to set channel plan during runtime


jwei5:
Prepared an API wifi_netmgr_utils_set_chplan() in rtk_netmgr.c for wifi manager to call into the API for setting channel plan

os/board/rtl8730e/src/component/os/tizenrt/rtk_netmgr.c
```diff
@@ -794,6 +796,24 @@ trwifi_result_e wifi_netmgr_utils_ioctl(struct netdev *dev, trwifi_msg_s *msg)
 	return TRWIFI_NOT_SUPPORTED;
 }
 
+trwifi_result_e wifi_netmgr_utils_set_chplan(struct netdev *dev, uint8_t chplan)
+{
+	trwifi_result_e wuret = TRWIFI_FAIL;
+	if ((wifi_is_connected_to_ap() == RTK_STATUS_SUCCESS) || (g_mode == RTK_WIFI_SOFT_AP_IF)){
```
hk-gwak, line 802: Hello @jwei5, @Dane-Kane,
As I heard, the situation that I cannot call this API is softAP mode && connected to client.
But looks like I cannot call this API when device is softAP mode && not connected to any client(mobile).
Can you check?

jwei5: Hello @hk-gwak ,

> But looks like I cannot call this API when device is softAP mode && not connected to any client(mobile).

Do you mean the condition shows that the API cannot be called when device is in softAP mode and there is no client connected to softAP?
The API should not be called in any of these conditions:
1. Module is in STA mode and connected to an AP
2. Module is in SoftAP mode (even if no devices are connected to SoftAP)

The API should be called in this condition:
1. Module is in STA mode and not connected to AP

hk-gwak: OK, I got it. Thanks.

hk-gwak (approved):
LGTM

PR #6292, opened 2024-07-12
os/arch/arm/src/amebasmart: I2S driver sampling fix

kaikaikaime:
*Wordlength correction has been added

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -604,6 +604,8 @@ static uint32_t i2s_txdatawidth(struct i2s_dev_s *dev, int bits)
 	else if (bits == I2S_BITS_PER_SAMPLE_32BIT)
 		priv->i2s_object->word_length = WL_32b;
 
+	i2s_set_param(priv->i2s_object, priv->i2s_object->channel_num, priv->i2s_object->sampling_rate, priv->i2s_object->word_length);
+
 	return priv->bits_per_sample * priv->sample_rate;
 #endif
 
@@ -967,6 +969,8 @@ static uint32_t i2s_rxdatawidth(struct i2s_dev_s *dev, int bits)
 		priv->i2s_object->word_length = WL_24b;
 	else if (bits == I2S_BITS_PER_SAMPLE_32BIT)
 		priv->i2s_object->word_length = WL_32b;
+
+	i2s_set_param(priv->i2s_object, priv->i2s_object->channel_num, priv->i2s_object->sampling_rate, priv->i2s_object->word_length);
 
 	return priv->bits_per_sample * priv->sample_rate;
 #endif
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
@@ -350,8 +350,8 @@ void i2s_rx_irq_handler(i2s_t *obj, i2s_irq_handler handler, uint32_t id)
   * @brief  Set I2S data transfer direction.
   * @param  obj: I2S object defined in application software.
   * @param  trx_type: Transfer direction. This parameter can be one of the following values:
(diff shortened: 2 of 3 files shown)


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -594,6 +594,9 @@ static uint32_t i2s_txdatawidth(struct i2s_dev_s *dev, int bits)
 	/* Support 16, 24, 32 bits */
 	DEBUGASSERT(priv && (bits == I2S_BITS_PER_SAMPLE_16BIT || bits == I2S_BITS_PER_SAMPLE_32BIT || bits == I2S_BITS_PER_SAMPLE_24BIT));
 
+	/* Unregister Port direction */
+	i2s_deinit(&priv->i2s_object);
```
abhishek-samsung, line 598: Hello @kaikaikaime , I think this should be as follows. Could you please confirm? I am getting a crash if I run the current code.
```suggestion
	i2s_deinit(priv->i2s_object);
```
Also even after changing it as above, it doesn't seem to work. As soon as deinit is called, uart logs are getting corrupted.

kaikaikaime: Hi abhishek, I have remove ampersasnd and disable of the clock which interferes with other peripherals

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -499,6 +501,7 @@ void i2s_init(i2s_t *obj, PinName sck, PinName ws, PinName sd_tx, PinName sd_rx,
   */
 void i2s_deinit(i2s_t *obj)
 {
+	/* Disabling general clock interferes with other peripheral
```
ziyik, line 504: Since it's not necessary could you delete it instead of commenting on it?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -507,7 +510,7 @@ void i2s_deinit(i2s_t *obj)
 	case PLL_CLOCK_98P304M:
 		PLL_I2S_98P304M(DISABLE);
 		break;
-	}
+	}*/
 
 	if (obj->i2s_idx == I2S2) {
```
ziyik, line 515: In this deinit function it only does SPORT clock disable.
Do we need GDMA_ChnlFree() and AUDIO_SP_Deinit()?
Both are called in i2s_disable() when PM try to suspend I2S activity.

ziyik:
Could you check amebasmart_i2s_initialize()?
priv->i2s_object is not free in the error case, this will cause memory lack.

kaikaikaime:
Dear Abhishek,

The current push e5cd4a, does not do deinit in other to set a new sampling rate. Instead, could you check the wordlength is working according to your needs? Especially the audio sound is correct from 48k and 16k sampling rate.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -430,16 +403,61 @@ void i2s_set_param(i2s_t *obj, int channel_num, int rate, int word_len)
...
+{
+	uint32_t clock_mode = 0;
+
+	obj->channel_num = channel_num;
+	obj->sampling_rate = rate;
+	obj->word_length = word_len;
+	clock_mode = i2s_clock_select(obj);
+
+	/* Sport Deinit */
+    AUDIO_SP_Unregister(obj->i2s_idx, obj->direction);
```
edwakuwaku, line 447: Please check alignment

kaikaikaime: Thank you @edwakuwaku, this is fixed now

abhishek-samsung (approved):
I have verified this PR (48KHz,16bit -> 16KHz, 32bit) and both sampling rate and data width change well.

Taejun-Kwon (approved):

PR #6293, opened 2024-07-15
os/pm: Add PM Quick Sleep

gSahitya-samsung:
This PR adds the feature to make board sleep quickly after wakeup. The quick sleep feature is needed to save upmost power when board need to wakeup aperiodically to serve timer expiration callback. For frequent timer expiration case if we do not allow quick sleep then board will remain wakeup for very long time resulting in no power saving.

Note - Now board can sleep immediately after wakeup. It's better to use pm_suspend if application want to suspend board working state.


os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -90,52 +90,42 @@ Theoretically, we will not expect PM state to drop to sleep again in such short
...
+		if (pmu_get_sleep_type() == SLEEP_PG) {
+			/* CPU1 is in task schedular, tell CPU1 to enter hotplug */
+			if (up_get_secondary_cpu_state(1) == CPU_RUNNING) {
+				/* CPU1 may in WFI idle state. Wake it up to enter hotplug itself */
+				up_irq_enable();
+				arm_cpu_sgi(GIC_IRQ_SGI4, (1 << 1));
+				arm_arch_timer_int_mask(0);
 				up_timer_enable();
-				arm_arch_timer_set_compare(arm_arch_timer_count() + 50000);
+				DelayUs(100);
```
r-prabu, line 102: why do we need DelayUs(100); here?

gSahitya-samsung: @edwakuwaku ??

edwakuwaku: Hello... This bsp idle logic will be revised in the PR: https://github.com/Samsung/TizenRT/pull/6256
The delay was added previously when the secondary core is failed to enter hotplug, as a kind of buffer time for secondary core to handle the hotplug signal....
As you know, the PM driver is keep on changing... Thus, BSP is forced to follow the PM driver change as well....
Please ignore this as of now as all of these logic will be revised later on...

edwakuwaku:
@gSahitya-samsung  could you please revert ameba_tizenrt_pmu.c/.h?  I will make necessary changes in my PR..

os/arch/arm/src/amebasmart/amebasmart_idle.c
r-prabu: @gSahitya-samsung mark it with #todo to revise it later point of time.

r-prabu (approved):

ewoodev (approved):

kishore-sn (dismissed):

os/pm/Kconfig
```diff
@@ -70,8 +60,8 @@ config PM_TIMEDWAKEUP
 if PM_TIMEDWAKEUP
 
 config PM_MIN_SLEEP_TIME
```
kishore-sn, line 62: I feel the naming of this config is not correct. MIN_SLEEP_TIME seems to indicate minimum time pm can sleep or something else. Can we change this config to PM_SLEEP_ENTRY_WAIT_MS or something else?

os/pm/Makefile
```diff
@@ -24,6 +24,7 @@ CSRCS += pm_changestate.c pm_checkstate.c pm_initialize.c pm_idle.c
 CSRCS += pm_register.c pm_domain_register.c pm_unregister.c pm_procfs.c
 CSRCS += pm_suspend.c pm_resume.c pm_timedsuspend.c
 CSRCS += pm_suspendcount.c
+CSRCS += pm_wakehandler.c
```
kishore-sn, line 27: Minor. We could add this file to previous line.

os/pm/pm_idle.c
```diff
@@ -111,11 +109,8 @@ void pm_idle(void)
 			}
 		}
 #endif
-#ifdef CONFIG_PM_TICKSUPPRESS
-	up_pm_board_sleep(pm_wakehandler);
-#else
-	up_pm_board_sleep(NULL);
-#endif
+		up_pm_board_sleep(pm_wakehandler);
+		stime = clock_systimer();
```
ritesh55555, line 113: what is the use of now and stime , it will be always now > stime . 
what is the use of if (now > stime) condition?

gSahitya-samsung: In one tick `pm_idle` can run multiple times, to avoid this we have added this condition

ritesh55555:
Hi @gSahitya-samsung , is the code tested properly . Because after board wakes up, it will stay in idle task until next tick . In that idle task, as other thread will have no chance to add a wakeup timer, then board might go to permanent sleep quickly in the idle task itself. Will this case be handled in future?

os/arch/arm/src/amebasmart/amebasmart_idle.c
gSahitya-samsung: Done

os/pm/Kconfig
gSahitya-samsung: Done

os/pm/pm_wakehandler.c
```diff
@@ -60,14 +60,13 @@ void pm_wakehandler(clock_t missing_tick, pm_wakeup_reason_code_t wakeup_src)
 #endif
 	pmllvdbg("wakeup source code = %d\n", wakeup_src);
 	pmllvdbg("missing_tick: %llu\n", missing_tick);
-
+#ifdef CONFIG_PM_TICKSUPPRESS
```
ewoodev, line 63: the reason why add it is getting build error?

ewoodev (approved):
@r-prabu @kishore-sn 
Please review and merge thie pr

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -169,6 +159,10 @@ void up_idle(void)
 
 	nxsched_process_timer();
 #else
+	/* power saving when idle*/
+	__asm(" DSB");
+	__asm(" WFI");
+	__asm(" ISB");
 	/* set core to WFE */
 	__asm("WFE");
```
ewoodev, line 167: Could you check it?
WFI is already call, but after get irq, WFE will be called again.

ewoodev: And Please check if it is correct to call WFI here.

This location can be called even when not in pm_sleep state, so in a normal situation, even though there are events on the secondary core, the primary core may not wake up.

gSahitya-samsung: @ewoodev I asked @edwakuwaku, he will do necessary cleanup required for SMP in PR #6256

ewoodev: Thank you,
when we use single core only, we can use WFI instead WFE.
WFI can reduce power more.

gSahitya-samsung: Yes WFI is better than WFE

ewoodev (approved):

kishore-sn (approved):

PR #6295, opened 2024-07-15
lib_buffer: add DEBUG_SET_CALLER_ADDR macro

ewoodev:
The apb_alloc api is wrapping malloc apis.
So we don't know who was the requesting alloc memory using heapinfo.

Therefore, Add DEBUG_SET_CALLER_ADDR to update the caller address that requested the actual memory.

### DIFF
--- a/lib/libc/audio/lib_buffer.c
+++ b/lib/libc/audio/lib_buffer.c
@@ -149,6 +149,7 @@ int apb_alloc(FAR struct audio_buf_desc_s *bufdesc)
 		ret = -ENOMEM;
 	} else {
 		/* Populate the buffer contents */
+		DEBUG_SET_CALLER_ADDR(*bufdesc->u.ppBuffer);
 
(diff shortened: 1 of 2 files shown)


lib/libc/audio/lib_buffer.c
```diff
@@ -149,6 +149,7 @@ int apb_alloc(FAR struct audio_buf_desc_s *bufdesc)
 		ret = -ENOMEM;
 	} else {
 		/* Populate the buffer contents */
+		DEBUG_SET_CALLER_ADDR(*bufdesc->u.ppBuffer);
```
sunghan-chang, line 152: all other locations use `apb`. How about using it here as well?

sunghan-chang:
@ewoodev Let's write a location at the title.
```
libc/audio/lib_buffer.c: update a caller address of memory allocation

blah blah
```

sunghan-chang (approved):

PR #6297, opened 2024-07-15
kmm_realloc: fix crash issue case of oldmem is NULL

ewoodev:
if oldmem argument is NULL or not in heap address, we get a crash in mm_free_delaylist:103

When we call the mm_malloc API, we guarantee that the argument's heap is valid before calling it. Therefore, there is no NULL check for the heap argument in mm_malloc. So, when calling mm from kmm or umm, we need to pass the correct heap variable as an argument.

In this issue, when oldmem was passed as NULL in kmm_realloc, mm_get_heap returned NULL and caused a crash because mm_realloc was called with a NULL argument. To fix this,

Change to so that, if we cannot find the heap using oldmem, mm_realloc is not called with NULL.

### DIFF
--- a/os/mm/kmm_heap/kmm_realloc.c
+++ b/os/mm/kmm_heap/kmm_realloc.c
@@ -144,21 +144,30 @@ FAR void *kmm_realloc(FAR void *oldmem, size_t newsize)
 	mmaddress_t caller_retaddr = 0;
 	ARCH_GET_RET_ADDRESS(caller_retaddr)
 #endif
-	struct mm_heap_s *kheap_origin = mm_get_heap(oldmem);
+	struct mm_heap_s *kheap_origin;
 	struct mm_heap_s *kheap_new;
 
-	if (newsize == 0) {
-		mm_free(kheap_origin, oldmem);
-		return NULL;
-	}
+	if (oldmem) {
+		kheap_origin = mm_get_heap(oldmem);
+
+		/* The oldmem given by first argument is not a dynamically
+		 * allocated address. This will cause ASSERT like Linux.
+		 */
+		ASSERT(kheap_origin);
+
+		if (newsize == 0) {
+			mm_free(kheap_origin, oldmem);
+			return NULL;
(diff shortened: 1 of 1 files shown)


ewoodev:
For verify this, we can execute cpuload -s 10

os/mm/kmm_heap/kmm_realloc.c
```diff
@@ -144,21 +144,26 @@ FAR void *kmm_realloc(FAR void *oldmem, size_t newsize)
...
+	struct mm_heap_s *kheap_origin;
 	struct mm_heap_s *kheap_new;
 
-	if (newsize == 0) {
-		mm_free(kheap_origin, oldmem);
-		return NULL;
-	}
+	if (oldmem) {
+		kheap_origin = mm_get_heap(oldmem);
+		ASSERT(kheap_origin);
```
sunghan-chang, line 152: Let's leave some explanation about this. How about below?
```
// The oldmem given by first argument is not a dynamically allocated address. This will cause ASSERT like Linux.
```

os/mm/kmm_heap/kmm_realloc.c
```diff
@@ -144,21 +144,28 @@ FAR void *kmm_realloc(FAR void *oldmem, size_t newsize)
...
+	struct mm_heap_s *kheap_origin;
 	struct mm_heap_s *kheap_new;
 
-	if (newsize == 0) {
-		mm_free(kheap_origin, oldmem);
-		return NULL;
-	}
+	if (oldmem) {
+		/* The oldmem given by first argument is not a dynamically
+		 * allocated address. This will cause ASSERT like Linux.*/
```
sunghan-chang, line 152:
```
/*
 *
 */
```
like others.

sunghan-chang: And move this comment before the `ASSERT`, next line.

sunghan-chang (approved):

PR #6298, opened 2024-07-16
build/tools/amebasmart/bootloader: remove download blocking

zhongnuo-tang:
1. remove download blocking in all case


build/tools/amebasmart/bootloader/km4_boot_all.bin
jylee9613, line 1: Hello @zhongnuo-tang 
I confirmed download works after running ATDS.
But when I checked prodconfig, it printed the log that both HW Debug and download are disabled.
```
HW Debug : Off
Download : Hotfix Only Available
```

I talked with Eric, would you please check it with him?

zhongnuo-tang: Hello @jylee9613,
I have added the check in prodconfig for download, please help to verify
Thank you

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -118,7 +118,13 @@ int up_check_prodswd(void)
 }
 int up_check_proddownload(void)
 {
-	return up_check_prod();
+	u8 hw_download, sw_download;
+	OTP_Read8(SEC_CFG2, &hw_download);
+	OTP_Read8(0x369, &sw_download);
+	if ((hw_download & 1) && (sw_download & 0x08)) {
+		return OK;
+	}
```
jylee9613, line 126: I checked this with two boards.

The board which SWD disabled show below log, so it's fine.
```
HW Debug : Off
Download : Downloader & Hotfix Available
```

But when I tested with SWD enabled board, it shows below log...
```
HW Debug : On
Download : Hotfix Only Available
```

jylee9613: Normal log in SWD enable board should like below.

```
HW Debug : On
Download : Downloader & Hotfix Available
```

zhongnuo-tang: But when I tested with SWD enabled board, it shows below log...
--> download is available but log shows it is not? could you help to print out the value of hw_download and sw_download.

jylee9613: Yes right. Both download and HW debug available but log shows download disable.

Oh sorry I missed a information.  I see this log in signing board. (signing + swd enable).
```
HW Debug : On
Download : Hotfix Only Available
```

Is swd related to signing?

zhongnuo-tang: swd is not related to signing, could you print out the value of hw_download and sw_download using signing module

jylee9613: Sorry for confusing.. There was typo error in my side..
I checked the logs coming out properly even if signing board.
```
HW Debug : On
Download : Downloader & Hotfix Available
```

Thank you for checking.

jylee9613 (approved):

PR #6300, opened 2024-07-16
os/board/rtl8730e/src/component: change SPI dma channel allocation/free

zhongnuo-tang:
1. from our experiment, frequent allocating/free gdma channel is causing unexpected behavior (ie. system stability issue)
2. thus we only allocate it once during the first spi operation

### DIFF
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
@@ -57,10 +57,10 @@ typedef struct {
 	/* mbed var */
 	u32 dma_en;
 
-	bool is_readwrite;
+	u32 is_readwrite;
 } HAL_SSI_ADAPTOR, *PHAL_SSI_ADAPTOR;
 
-HAL_SSI_ADAPTOR ssi_adapter_g[2];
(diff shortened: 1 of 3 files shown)


os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/spi_api.c
```diff
@@ -57,10 +57,10 @@ typedef struct {
 	/* mbed var */
 	u32 dma_en;
 
-	bool is_readwrite;
+	u32 is_readwrite;
```
sunghan-chang, line 60: Is this related with `change SPI DMA channel allocation/free`?

ewoodev (approved):
I tested spi operation

sunghan-chang (approved):

sunghan-chang:
@zhongnuo-tang We found that this change does not resolve the issue yet. We reproduced the issue again with this change. Please talk with Eric.

zhongnuo-tang:
> @zhongnuo-tang We found that this change does not resolve the issue yet. We reproduced the issue again with this change. Please talk with Eric.

@sunghan-chang Hi Mr Chang, okay, I will find out from Eric what is the difference between the old/new commits

PR #6301, opened 2024-07-16
app/examples/testcase/le_tc/kernel : Fix kernel_tc testcase for rtl8721csm board

Kr159:
Make required changes to resolve the failed kernel testcase. There was two tc is failing: First in tc_task.c and 2nd in tc_timer.c In tc_task.c: it was working only in case of when SMP was enabled,now it is working in both enable and disable cases. And in tc_timer.c: it is a minor change to check the set time(ie: st_timer_spec_set.it_value.tv_sec) be greater than the get time(st_timer_spec_get.it_value.tv_sec).

### DIFF
--- a/apps/examples/testcase/le_tc/kernel/tc_task.c
+++ b/apps/examples/testcase/le_tc/kernel/tc_task.c
@@ -289,7 +289,12 @@ static void tc_task_task_restart(void)
 	ret_chk = task_restart(pid);
 
 	sleep(SEC_1);
+#ifdef CONFIG_SMP
+	//lldbg("ret_chk=  0x%08x\n", ret_chk);
 	TC_ASSERT_EQ("task_restart", ret_chk, 0);
+#else
+	TC_ASSERT_NEQ("task_restart", ret_chk, 0);
+#endif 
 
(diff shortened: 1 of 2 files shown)


sunghan-chang (commented):
@Kr159 Could you add more explanation in the commit description? about the reason of the failure and why we should modify like this

sunghan-chang:
`app/examples/testcase/le_tc/kernel : Fix kernel_tc testcase for rtl8721csm board`
title looks little bit long.
How about
`examples/testcase/kernel : Fix ?? for SMP disabled`
or something else?

And why do you add `for rtl8721csm board` at the title? Is it ok for other chipsets?

apps/examples/testcase/le_tc/kernel/tc_task.c
```diff
@@ -289,7 +289,12 @@ static void tc_task_task_restart(void)
 	ret_chk = task_restart(pid);
 
 	sleep(SEC_1);
+#ifdef CONFIG_SMP
+	//lldbg("ret_chk=  0x%08x\n", ret_chk);
```
neel-samsung, line 293: Is this comment necessary? If not then, can you please remove it.

PR #6315, opened 2024-07-25
os/drivers/pm: fix errno in pm_driver

gSahitya-samsung:
This commit resolve ioctl overwrite issue of errno. Now pm_ioctl returning errno directly to fs_ioctl, which will set errno with given error.

### DIFF
--- a/os/drivers/pm/pm.c
+++ b/os/drivers/pm/pm.c
@@ -144,8 +144,11 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	default:
 		pmvdbg("Invalid command passed!\n");
 		break;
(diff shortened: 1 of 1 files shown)


sunghan-chang (commented):
You should NOT mix different topics in a single commit.
Please split beautifuler and fixing errno as different commits.

ewoodev (approved):

sunghan-chang (approved):

r-prabu (approved):

abhishek-samsung (approved):

r-prabu (approved):
fixing overwrite issue of errno look good.

kishore-sn (approved):

PR #6318, opened 2024-07-29
os/include/tinyara/lcd/: add comments for describing each init comman…

namanjain7:
…ds and remove landscape settings for st7701sn LCD

add comments for send init commands for LCD st7701sn and remove landscape settings for LCD st7701sn. This LCD do not have landscape mode support.


namanjain7:
I have added comments for init commands. However, some new init commands definition are not present in datasheet, so have skipped those.

os/include/tinyara/lcd/st7701sn.h
```diff
@@ -30,7 +30,8 @@
  * Pre-processor Definitions
  ****************************************************************************/
 /* System Function Command Table 1 */
-
+#define LCD_XRES CONFIG_LCD_XRES
+#define LCD_YRES CONFIG_LCD_YRES
```
sunghan-chang, line 34: How about leaving some comment to say this does not support rotation in hw?

sunghan-chang: And let's split commit for this and adding comment at each command.

sunghan-chang (approved):

PR #6320, opened 2024-07-31
[DO NOT MERGE] os/drivers/lcd/: add rotation support for LCD ST7701SN

namanjain7:
LCD ST7701SN does not have support for LANDSCAPE mode. This commit rotate the screen from PORTRAIT to LANDSCAPE in lcd driver

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -38,6 +38,12 @@
 #define CONFIG_LCD_MAXPOWER 100
 #endif
 
+#ifdef CONFIG_LCD_ST7701SN
+#define NUM_OF_LCD_BUFFER	(2)
+uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER] = { NULL, NULL };
+int lcd_buffer_index = 0;
+#endif
+
 struct mipi_lcd_dev_s {
 	/* Publicly visible device structure */
 
@@ -174,7 +180,24 @@ static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord
 	row_end += 1;
 	col_start += 1;
 	col_end += 1;
+#ifdef CONFIG_LCD_ST7701SN
+	int src_buffer_index = 0;
+	int dest_buffer_index = 0;
+	uint8_t *lcd_rotate_buffer = buffer;
+	for (int src_row = 0; src_row < CONFIG_LCD_YRES; src_row++) {
+		for (int src_col = CONFIG_LCD_XRES - 1; src_col >= 0 ; src_col--) {
+			lcd_rotate_buffer = lcd_buffer[lcd_buffer_index];
+			src_buffer_index = ((CONFIG_LCD_XRES * src_col) + src_row) * 2;
+			lcd_rotate_buffer[dest_buffer_index] = buffer[src_buffer_index];
+			lcd_rotate_buffer[dest_buffer_index + 1] = buffer[src_buffer_index + 1];
+			dest_buffer_index += 2;
+		}
+	}
+	lcd_buffer_index = (lcd_buffer_index == 0) ? 1 : 0;
+	priv->config->lcd_put_area((u8 *)lcd_rotate_buffer, row_start, col_start, row_end, col_end);
+#else
 	priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
+#endif
 	return OK;
 }
 
@@ -348,6 +371,17 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 		lcddbg("ERROR: LCD Init sequence failed\n");
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
-
+#ifdef CONFIG_LCD_ST7701SN
+	for (int itr = 0; itr < NUM_OF_LCD_BUFFER; itr++) {
+		lcd_buffer[itr] = (uint8_t *)malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1);
+		if (!lcd_buffer[itr]) {
+			lcddbg("ERROR: LCD rotate buffer memory allocation failed\n");
+			for (int itr2 = 0; itr2 < itr; itr2++) {
+				free(lcd_buffer[itr2]);
+			}
+			break;
+		}
+	}
+#endif
 	return &priv->dev;
 }


os/drivers/lcd/mipi_lcd.c
```diff
@@ -38,6 +38,11 @@
 #define CONFIG_LCD_MAXPOWER 100
 #endif
 
+#ifdef CONFIG_LCD_ST7701SN
+uint8_t *lcd_buffer_1 = NULL, *lcd_buffer_2 = NULL;
+bool is_lcd_buffer_1 = false;
+#endif
```
sunghan-chang, line 44: 1. Let's make a new config for sw orientation or sw rotation.
2. Let's add `static`.
3. Let's define each at each line.
```
uint8_t *lcd...._1 = NULL;
uint8_t *lcd..._2 = NULL;
bool .... = false;
```
4. How about below?
```
#define NUM_OF_LCD_BUFFER  (2)
uint8_t lcd_buffer[NUM_OF_LCD_BUFFER] = { NULL, NULL };
uint8_t lcd_buffer_index = 0;
```

os/drivers/lcd/mipi_lcd.c
```diff
@@ -169,12 +174,35 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
...
+	}
+
+	for(int src_row = 0; src_row < CONFIG_LCD_YRES; src_row++) {
+		for(int src_col = CONFIG_LCD_XRES - 1; src_col >= 0 ; src_col--) {
+			src_buf_ind = ((CONFIG_LCD_XRES * src_col) + src_row) * 2;
+			final_buffer[dest_buf_ind] = buffer[src_buf_ind];
+			final_buffer[dest_buf_ind+1] = buffer[src_buf_ind+1];
+			dest_buf_ind += 2;
+		}
+	}
```
sunghan-chang, line 198: many coding rule violations
1. for (
2. ind + 1
3. if (
4. } else {

os/drivers/lcd/mipi_lcd.c
```diff
@@ -348,6 +376,9 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 		lcddbg("ERROR: LCD Init sequence failed\n");
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
-
+#ifdef CONFIG_LCD_ST7701SN
+	lcd_buffer_1 = (uint8_t *)malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1);
+	lcd_buffer_2 = (uint8_t *)malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1);
+#endif
```
sunghan-chang, line 382: There is no free. Then, do we need to get dynamic allocation instead of just global static array?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -169,12 +174,35 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
 static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord_t row_end, fb_coord_t col_start, fb_coord_t col_end, FAR const uint8_t *buffer, fb_coord_t stride)
 {
 	FAR struct mipi_lcd_dev_s *priv = (FAR struct mipi_lcd_dev_s *)dev;
+
+	uint8_t *final_buffer = buffer;
+#ifdef CONFIG_LCD_ST7701SN
```
sunghan-chang, line 179: Currently this supports landscape only. But we can do all of orientations in SW. Let's add them as well later.
```
#ifdef CONFIG_LCD_LANDSCAPE
  code
#elif defined(CONFIG_LCD_PORTRAIT)
  code
#elif defined(CONFIG_LCD_RPORTRAIT)
....
```

os/drivers/lcd/mipi_lcd.c
sunghan-chang: And any other algorithm to enhance performance?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -169,12 +174,35 @@ static int lcd_putrun(FAR struct lcd_dev_s *dev, fb_coord_t row, fb_coord_t col,
...
+		final_buffer = lcd_buffer_1;
+		is_lcd_buffer_1 = true;
+	}
+
+	for(int src_row = 0; src_row < CONFIG_LCD_YRES; src_row++) {
+		for(int src_col = CONFIG_LCD_XRES - 1; src_col >= 0 ; src_col--) {
+			src_buf_ind = ((CONFIG_LCD_XRES * src_col) + src_row) * 2;
+			final_buffer[dest_buf_ind] = buffer[src_buf_ind];
+			final_buffer[dest_buf_ind+1] = buffer[src_buf_ind+1];
+			dest_buf_ind += 2;
```
edwakuwaku, line 196: This seems to assume the resolution type used is always RGB565. As I knew, ST7701SN seems using higher resolution such as RGB888, then each pixel will be 3bytes.
Please check whether some macro should be used to differentiate that.

namanjain7:
This commit is working in flat build. For loadable_ext, both the malloc and kmm_zalloc are failing to allocate two buffers.
I think we need to increase kernel memory allocation for loadable_ext.

os/drivers/lcd/mipi_lcd.c
namanjain7: done.

os/drivers/lcd/mipi_lcd.c
namanjain7: unlike tizen, there is no exit function in driver.
What I was thinking is: 
https://github.com/Samsung/TizenRT/blob/faaec07247ec92cca066d88da53ff0dafe8cd154/os/drivers/lcd/lcd_dev.c#L153

in this lcd_dev.c , we keep the count of fd. So if this fd becomes 0, then we can free the buffer.

os/drivers/lcd/mipi_lcd.c
sunghan-chang: 1st and 2nd are not done.
For 1st, you use the config for ST7701SN. But it's not a feature of that. Let's make NEW config for SW rotation and use it.

os/drivers/lcd/mipi_lcd.c
sunghan-chang: Need to check here

os/drivers/lcd/mipi_lcd.c
sunghan-chang: Need to check

sunghan-chang:
@namanjain7 Yes, we have a PR for that, PR #6324

sunghan-chang:
@namanjain7 Update is too late. Could you check remain comments?

edwakuwaku:
Hi @namanjain7 , could you share the difference in FPS after applying this change? 
With this frame rotation method, is it able to guarantee 60FPS still? 
If the gap is too big (ie. currently calculation causes too big overhead), we should think of some way to optimize the rotation algorithm...

edwakuwaku:
> This commit is working in flat build. For loadable_ext, both the malloc and kmm_zalloc are failing to allocate two buffers. I think we need to increase kernel memory allocation for loadable_ext.

Using malloc/free, we cannot guarantee enough memory space (ie. free heap) everytime. It would cause experience drop due to image update failing... Do you have any backup solution for this?
If not, static allocation should be considered.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -348,6 +371,17 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 		lcddbg("ERROR: LCD Init sequence failed\n");
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
-
+#ifdef CONFIG_LCD_ST7701SN
+	for (int itr = 0; itr < NUM_OF_LCD_BUFFER; itr++) {
+		lcd_buffer[itr] = (uint8_t *)malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1);
```
sunghan-chang, line 376: This should be `kmm_malloc` to get memory at kernel side. (Because the driver is in kernel context)
To use kmm_malloc, you should include <tinyara/kmalloc.h>

PR #6322, opened 2024-08-01
os: Add pm_start api

gSahitya-samsung:
This commit enable applications to start power
management system. This API should be called
when board is ready to transition into low
power state. This API helps PM to remove the
use of `PM_IDLE_DOMAIN` outside the scope of PM.

`pm_start` - This API meant to be called during app
initialization to enable low power state transition.
In PM side we can reduce some of the costlier operation
based on the `is_running` flag. In the event of crash
and app reloading, this API can be used for reinitialization.

`pm_resume` - This API removes pm state transition lock for
specific domain. It has domain information, which helps PM
to do domain specific operation, like "when LCD domain get
resume, do backlight dimming". This should be used in pair
with `pm_suspend` API.

`pm_suspend` and `pm_resume` are like locks, which should be used
in pairs to prevent board sleep in critical section. The current
usage `pm_suspend(0)` in OS and `pm_resume(0)` in application, to prevent
board sleep before app initialization is not good design. So, we kept
`PM (OFF)` by default and pm_start API will do `PM (ON)` when application
is ready to do power saving.

### DIFF
--- a/os/drivers/pm/pm.c
+++ b/os/drivers/pm/pm.c
@@ -76,6 +76,7 @@ static ssize_t pm_write(FAR struct file *filep, FAR const char *buffer, size_t l
  *   PMIOC_METRICS - to get pm metrics data for given time
  *   PMIOC_TUNEFREQ - for changing the operating frequency of the core to save power
  *   PMIOC_SUSPEND_COUNT - to get suspend count of pm domain
+ *   PMIOC_START - to start PM functionality to make board sleep
  * 
  * Arguments:
  *   filep is ioctl fd, cmd is required command, arg is required argument for
@@ -88,6 +89,7 @@ static ssize_t pm_write(FAR struct file *filep, FAR const char *buffer, size_t l
  *   for PMIOC_METRICS, arg should be an int type.
  *   for PMIOC_TUNEFREQ, arg should be an int type.
  *   for PMIOC_SUSPEND_COUNT, arg should be an int type.
+ *   for PMIOC_START, arg should be NULL
  *
  * Description:
  *   This api can be used to perform PM operation.
@@ -103,6 +105,7 @@ static ssize_t pm_write(FAR struct file *filep, FAR const char *buffer, size_t l
  *   PMIOC_METRICS           -   return OK on success
  *   PMIOC_TUNEFREQ          -   return OK on success
  *   PMIOC_SUSPEND_COUNT     -   return non-negative suspend count of domain
+ *   PMIOC_START             -   return OK
  *
  ************************************************************************************/
 static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
@@ -117,6 +120,12 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
         	break;
         case PMIOC_RESUME:
 		ret = pm_resume((int)arg);
+		/* Note: Below code is temporary fix to support old 'system_pm_start' API.
+		 * It needs to remove for later version.
+		 */
+		if (arg == PM_IDLE_DOMAIN) {
+			pm_start();
+		}
 		pmvdbg("State unlocked!\n");
 		break;
 	case PMIOC_SLEEP:
@@ -143,6 +152,10 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 				ret = OK;
 			}
 		}
+		break;
+	case PMIOC_START:
+		pm_start();
+		ret = OK;
 		break;
 	case PMIOC_SUSPEND_COUNT:
 		ret = pm_suspendcount((int)arg);
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -294,6 +294,25 @@ EXTERN const char *wakeup_src_name[PM_WAKEUP_SRC_COUNT];
 void pm_driver_register(void);
 
 #ifdef CONFIG_PM
+
+/****************************************************************************
+ * Name: pm_start
+ *
+ * Description:
+ *   This function is called by the application thread to start the Power
+ *   Management system. This fucntion sets the is_running flag which
+ *   enables pm to transition between low and high power states.
+ *
+ * Input parameters:
+ *   None.
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
+
+void pm_start(void);
+
 /****************************************************************************
  * Name: pm_initialize
  *
@@ -550,6 +569,7 @@ int pm_metrics(int milliseconds);
  * avoid so much conditional compilation in driver code when PM is disabled:
  */
 
+#define pm_start()
 #define pm_initialize()
 #define pm_register(cb)         (0)
 #define pm_unregister(cb)       (0)
--- a/os/pm/pm.h
+++ b/os/pm/pm.h
@@ -135,6 +135,9 @@ struct pm_global_s {
 
 	/* No. of Registered Domains */
 	uint16_t ndomains;
+
+	/* Indicates Board is Ready to State Change */
+	bool is_running;
 };
 
 /****************************************************************************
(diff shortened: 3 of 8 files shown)


os/include/tinyara/pm/pm.h
```diff
@@ -292,6 +292,25 @@ extern "C" {
...
+
+/****************************************************************************
+ * Name: pm_start
+ *
+ * Description:
+ *   This function is called by the application thread to start the Power
+ *   Management system. This fucntion sets the ready_to_state_change flag which
+ *   enables pm to transition between low and high power states.
+ *
+ * Input parameters:
```
r-prabu, line 304: Why this interface is required? Can you specify the needs of it?

os/pm/pm_initialize.c
```diff
@@ -81,6 +81,26 @@ struct pm_global_s g_pmglobals;
...
+ * Input parameters:
+ *   None.
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
+
+void pm_start(void) {
+	g_pmglobals.ready_to_state_change = true;
```
r-prabu, line 101: Even if we need this interface, cant we need to check this flag as well in the places where we do the state changes for the pm?

os/pm/pm_initialize.c
```diff
@@ -81,6 +81,26 @@ struct pm_global_s g_pmglobals;
...
+ *   Management system. This fucntion sets the ready_to_state_change flag which
+ *   enables pm to transition between low and high power states.
+ *
+ * Input parameters:
+ *   None.
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
```
r-prabu, line 98: Also it seems, If any one domain/platform user once starts this will enable the pm. Is this our intension? Also once set there is no provision to clear this. Any reason for this

os/pm/pm_initialize.c
```diff
@@ -81,6 +81,26 @@ struct pm_global_s g_pmglobals;
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
+ * Name: pm_start
+ *
+ * Description:
+ *   This function is called by the application thread to start the Power
+ *   Management system. This fucntion sets the ready_to_state_change flag which
```
ritesh55555, line 89: minor : fucntion -> function

os/pm/pm_initialize.c
gSahitya-samsung: the `pm_changestate` is the only function which do pm state transition, we have used this flag in it to stop state transition

os/pm/pm.h
```diff
@@ -132,6 +132,9 @@ struct pm_global_s {
 	/* Timer to decrease state */
 
 	WDOG_ID wdog;
+
+	/* Indicates Board is Ready to State Change */
+	bool ready_to_state_change;
```
ritesh55555, line 137: do we need to initialize this value in pm_initialize?

os/include/tinyara/pm/pm.h
```diff
@@ -550,6 +569,7 @@ int pm_metrics(int milliseconds);
  * avoid so much conditional compilation in driver code when PM is disabled:
  */
 
+#define pm_start()              (0)
```
ritesh55555, line 572: should be #define pm_start() only , its a void funtion

os/pm/pm_initialize.c
gSahitya-samsung: Our intention is to give product team an API, which they can call to start PM after `START TASK MAIN`

os/include/tinyara/pm/pm.h
gSahitya-samsung: Previously we were using `PM_IDLE_DOMAIN` to suspend PM state transition in `os_start`. Since we are suspending `PM_IDLE_DOMAIN`, so to enable pm state transition, we were using `system_pm_start` which resume the `PM_IDLE_DOMAIN`. This was the temporary fix.

Now with this interface we can avoid the sharing `PM_IDLE_DOMAIN` to `os_start` and `system_pm_start`. It gives better encapsulation to PM Module.

os/pm/pm.h
gSahitya-samsung: by defaut all code section allocation are initializing with `0`

kishore-sn (approved):

ewoodev:
We previously considered using pm_start as well as adding pm_suspend and pm_resume to application init.

pm_start can also be a temporary fix. This is because it cannot satisfy app unloading and reloading.

I think it would be better if we add pm_suspend in each booting section.

Could you consider to add pm_suspend binary manager and application init code?

neel-samsung (approved):
Add pm_start PR looks good to merge

os/pm/pm_changestate.c
```diff
@@ -221,6 +221,10 @@ int pm_changestate(enum pm_state_e newstate)
 	 */
 
 	flags = enter_critical_section();
+	/* State change only if PM is ready to state change */
+	if (!g_pmglobals.ready_to_state_change) {
```
ewoodev, line 225: Let's move this into pm_idle.

os/drivers/pm/pm.c
```diff
@@ -130,6 +131,10 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 			}
 		}
 		break;
+	case PMIOC_START:
+		pm_start();
```
ewoodev, line 135: For existing compatibility, case PMIOC_RESUME: also needs a temporary modification to call pm_start when it is 0.

os/include/tinyara/pm/pm.h
gSahitya-samsung: Done

os/pm/pm_changestate.c
gSahitya-samsung: Since pm state transition can happen at various places like `Idle Thread, timer, wake handlers, etc`. Isn't it good to keep this in `pm_changestate` ?

os/drivers/pm/pm.c
gSahitya-samsung: Done

os/drivers/pm/pm.c
```diff
@@ -117,6 +120,9 @@ static int pm_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
         	break;
         case PMIOC_RESUME:
 		ret = pm_resume((int)arg);
+		if (ret == OK) {
```
ewoodev, line 123: arg???

os/pm/pm_changestate.c
ewoodev: In the current structure, the initial start for PM is from pm_idle. It seems like a good idea to have the idle thread do nothing at all.

ewoodev (approved):

sunghan-chang (commented):
@gSahitya-samsung @ewoodev Could you leave more details in the description? Why this is necessary? How did pm work without this? What is different between resume and start?

gSahitya-samsung:
`pm_start` - This API meant to be called during app initialization to enable low power state transition. In PM side we can reduce some of the costlier operation based on the `ready_to_state_change` flag. In the event of crash and app reloading, this API can be used for reinitializations.

`pm_resume` - This API removes pm state transition lock for particular domain. It has domain information, which helps PM to do domain specific operation, like "when LCD domain get resume do backlight dimming". This should be used in pair with `pm_suspend` API.

`pm_suspend` and `pm_resume` are like locks, which should be used in pairs to prevent board sleep in critical section. The current usage `pm_suspend(0)` in OS and `pm_resume(0)` in application, to prevent board sleep before app initialization is not good design. So, we kept `PM (OFF)` by default and `pm_start` API will do `PM (ON)` when application is ready to do power saving.

sunghan-chang:
@ewoodev @gSahitya-samsung Can I merge this now?

ewoodev (approved):

ritesh55555 (approved):
i have also verfied, looks good to merge

r-prabu (approved):

PR #6328, opened 2024-08-07
build/configs/common_download.sh: Download using ttyUSB(n-2) as the n…

Tirtha-Samsung:
…ew rtl8730e open 4 ports

New rtl8730e board open 4 ports(eg. ttyUSB0, ttyUSB1, ttyUSB2 and ttyUSB3), however it is only getting downloaded through ttyUSB(n-2)(eg. ttyUSB1)

### DIFF
--- a/build/configs/common_download.sh
+++ b/build/configs/common_download.sh
@@ -31,27 +31,19 @@ CONFIG=${OS_PATH}/.config
 source ${CONFIG}
 SMARTFS_BIN_PATH=${BIN_PATH}/${CONFIG_ARCH_BOARD}_smartfs.bin
 
-for device in $(ls /dev/ttyACM* | sort -V); do
-	if [ -c "$device" ]; then
-		DEFAULT_PORT=$device
-	fi
-done
-
-if [ -z ${DEFAULT_PORT} ]; then
-	for device in $(ls /dev/ttyUSB* | sort -V); do
-		if [ -c "$device" ]; then
-			DEFAULT_PORT=$device
-		fi
-	done
-fi
 BOARD_CONFIG=${TOP_PATH}/build/configs/${CONFIG_ARCH_BOARD}/board_metadata.txt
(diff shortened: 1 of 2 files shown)


build/configs/common_download.sh
```diff
@@ -43,7 +43,15 @@ if [ -z ${DEFAULT_PORT} ]; then
 			DEFAULT_PORT=$device
 		fi
 	done
+	if [ -n ${DEFAULT_PORT} ]; then
+		if [ -n ${CONFIG_ARCH_BOARD_RTL8730E} ]; then
+			highest_number=$(echo "$DEFAULT_PORT" | grep -o '[0-9]\+')
+			desired_number=$((highest_number - 2))
+			DEFAULT_PORT="/dev/ttyUSB${desired_number}"
```
kishore-sn, line 50: Will this work for all versions of the 8730 board ?? 
If we know that 8730 needs ttyUSB1, then why do we need to do this arithematic ? We can just hardcode the required port number.

build/configs/common_download.sh
```diff
@@ -43,7 +43,15 @@ if [ -z ${DEFAULT_PORT} ]; then
 			DEFAULT_PORT=$device
 		fi
 	done
+	if [ -n ${DEFAULT_PORT} ]; then
```
abhishek-samsung, line 46: I think its better to have the following in boart_metadata.txt,

DEFAULT_PORT1 and DEFAULT_PORT2 (will be present if only required, currently its only for rtl8730e).

Then, we can write the logic as follows,

```
if (DEFAULT_PORT1 is available) {
    PORT = DEFAULT_PORT1;
} else if (DEFAULT_PORT2 defined) {
    if (DEFAULT_PORT2 is available) {
        PORT = DEFAULT_PORT2;
    else {
         exit("No available port");
    }
} else {
   exit("No available port");
}
``` 

In the current way, even for some boards like imxrt which use ttyACM0, we still attempt to download to the ttyUSB port if its present.

build/configs/common_download.sh
Tirtha-Samsung: yes we can, but if it opens 3 4 5 6 , then we need to download using ttyUSB4 right?

PR #6304, opened 2024-07-18
[UnderVerification] os/board/rtl8730e: Modify I2S Pin

ZhenBei-Sin:
build/configs/rtl8730e: Enable CONFIG_GPIO by default

os/board/rtl8730e: Modify I2S Pin
Change Notes:
- Configure the I2S BCLK pin for both normal and sleep modes with Pull Down initialization.
- Configure the I2S BCLK pin to operate as a GPIO input pin when enter sleep mode.
- Configure the PA23 pin for normal mode with Pull Down initialization.

PR #6330, opened 2024-08-09
Allow I2S BCLK as PDM clock

cvejlbo:
Changed I2S setup to use 16-bit channel length for 1.536MHz I2S_BCLK to be able to use it as PDM CLK. 
Enable DMIC power on PA_25.  
Configured NDP120 to use external PDM 1.536MHZ clock.  
Note that no handoff between using I2S_BCLK and NDP120 is currently implemented

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -1663,7 +1663,12 @@ static void i2s_getdefaultconfig(struct amebasmart_i2s_s *priv)
 	}
 	priv->txenab = priv->config->txenab;
 	if (priv->config->txenab) {
+#if defined(CONFIG_AUDIO_NDP120)
+		/* use channel length pof 16bits to give 1.536MHz I2S_BCLK which is shared with PDM clock */
+		priv->i2s_object->channel_length = SP_TXCL_16;
+#else
 		priv->i2s_object->channel_length = SP_TXCL_32;
+#endif
 		priv->i2s_object->fifo_num = SP_TX_FIFO2;
 		priv->i2s_object->direction = SP_DIR_TX;
 		priv->i2s_object->role = MASTER;
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -1663,12 +1663,7 @@ static void i2s_getdefaultconfig(struct amebasmart_i2s_s *priv)
 	}
 	priv->txenab = priv->config->txenab;
 	if (priv->config->txenab) {
-#if defined(CONFIG_AUDIO_NDP120)
-		/* use channel length pof 16bits to give 1.536MHz I2S_BCLK which is shared with PDM clock */
-		priv->i2s_object->channel_length = SP_TXCL_16;
-#else
 		priv->i2s_object->channel_length = SP_TXCL_32;
-#endif
 		priv->i2s_object->fifo_num = SP_TX_FIFO2;
 		priv->i2s_object->direction = SP_DIR_TX;
 		priv->i2s_object->role = MASTER;
--- a/os/board/rtl8730e/src/rtl8730e_alc1019.c
+++ b/os/board/rtl8730e/src/rtl8730e_alc1019.c
@@ -62,9 +62,6 @@ extern FAR struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port, bool is_re
 /*other pin config */
 #define ALC1019_GPIO_RESET_PIN		PA_22
 
-#if defined(CONFIG_AUDIO_NDP120)
-#define ALC1019_GPIO_DMIC_EN        PA_25
-#endif
 
 #define ALC1019_AVAILABLE_MINOR_MIN	0
 #define ALC1019_AVAILABLE_MINOR_MAX	25
@@ -215,13 +212,6 @@ int rtl8730e_alc1019_initialize(int minor)
 			goto errout_with_pcm;
 		}
 
-#if defined(CONFIG_AUDIO_NDP120)
-		auddbg("Enable DMIC\n");
-		gpio_t gpio_dmic_en;
-		gpio_init(&gpio_dmic_en, ALC1019_GPIO_DMIC_EN);
-		gpio_dir(&gpio_dmic_en, PIN_OUTPUT);
-		gpio_mode(&gpio_dmic_en, PullUp);
-#endif
 		/* Now we are initialized */
 
 		initialized = true;
--- a/os/board/rtl8730e/src/rtl8730e_alc1019.c
+++ b/os/board/rtl8730e/src/rtl8730e_alc1019.c
@@ -62,7 +62,6 @@ extern FAR struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port, bool is_re
 /*other pin config */
 #define ALC1019_GPIO_RESET_PIN		PA_22
 
-
 #define ALC1019_AVAILABLE_MINOR_MIN	0
 #define ALC1019_AVAILABLE_MINOR_MAX	25
 
--- a/os/board/rtl8730e/src/rtl8730e_alc1019.c
+++ b/os/board/rtl8730e/src/rtl8730e_alc1019.c
@@ -217,12 +217,11 @@ int rtl8730e_alc1019_initialize(int minor)
 
 #if defined(CONFIG_AUDIO_NDP120)
 		auddbg("Enable DMIC\n");
-        gpio_t gpio_dmic_en;
-        gpio_init(&gpio_dmic_en, ALC1019_GPIO_DMIC_EN);
+		gpio_t gpio_dmic_en;
+		gpio_init(&gpio_dmic_en, ALC1019_GPIO_DMIC_EN);
 		gpio_dir(&gpio_dmic_en, PIN_OUTPUT);
 		gpio_mode(&gpio_dmic_en, PullUp);
 #endif
-        
 		/* Now we are initialized */
 
 		initialized = true;
--- a/os/board/rtl8730e/src/rtl8730e_alc1019.c
+++ b/os/board/rtl8730e/src/rtl8730e_alc1019.c
@@ -62,6 +62,10 @@ extern FAR struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port, bool is_re
 /*other pin config */
 #define ALC1019_GPIO_RESET_PIN		PA_22
 
+#if defined(CONFIG_AUDIO_NDP120)
+#define ALC1019_GPIO_DMIC_EN        PA_25
+#endif
+
 #define ALC1019_AVAILABLE_MINOR_MIN	0
 #define ALC1019_AVAILABLE_MINOR_MAX	25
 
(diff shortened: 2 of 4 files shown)


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1663,7 +1663,12 @@ static void i2s_getdefaultconfig(struct amebasmart_i2s_s *priv)
 	}
 	priv->txenab = priv->config->txenab;
 	if (priv->config->txenab) {
+#if defined(CONFIG_AUDIO_NDP120)
```
edwakuwaku, line 1666: I think this macro is only in internal codebase. Is it ok to apply it in public?

cvejlbo: I am not 100% sure who should really answer this, however:
`CONFIG_AUDIO_NDP120` is already in the public codebase, here is a quick grep from a fresh clone of TizenRT repo done today (HEAD at commit 2901b73b196c9b49e9970bd4a76db6ae69788deb)
```
~/tmp/TizenRT (master ✔) grep -riIn CONFIG_AUDIO_NDP120
./os/drivers/audio/Make.defs:83:ifeq ($(CONFIG_AUDIO_NDP120),y)
./os/board/rtl8730e/src/Make.defs:75:ifeq ($(CONFIG_AUDIO_NDP120),y)
./os/board/rtl8730e/src/rtl8730e_ndp120.c:40:#ifdef CONFIG_AUDIO_NDP120
./os/board/rtl8730e/src/rtl8730e_boot.c:488:#ifdef CONFIG_AUDIO_NDP120
./os/include/tinyara/audio/ndp120.h:32:#ifdef CONFIG_AUDIO_NDP120
./os/include/tinyara/audio/ndp120.h:129:#endif							/* CONFIG_AUDIO_NDP120 */
./build/configs/rtl8730e/audio/defconfig:531:CONFIG_AUDIO_NDP120=y
~/tmp/TizenRT (master ✔)
```

Taejun-Kwon: @edwakuwaku
Hello edward, I think you can provide some config regarding this.
Something like a 'CONFIG_AMEBASMART_TXCL'
Then we can set default value as 16 or 32, we will take 16 when we use ndp120.
How do you think?
@abhishek-samsung  Please let me know your opinion too.

os/board/rtl8730e/src/rtl8730e_alc1019.c
```diff
@@ -62,6 +62,10 @@ extern FAR struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port, bool is_re
 /*other pin config */
 #define ALC1019_GPIO_RESET_PIN		PA_22
 
+#if defined(CONFIG_AUDIO_NDP120)
```
Taejun-Kwon, line 65: Actually alc1019 doesn't control this pin by itself, but this is required to support AEC, between amp & ndp120.
Anyway As I remember, this gpio pin will not be changed, then I think it is need to be moved to rtl8730e_ndp120.c

os/drivers/ai-soc/ndp120/src/ndp120_api.c
```diff
@@ -77,6 +77,10 @@
 /* currently the max size of tank is limited to ~580ms  */
 #define AUDIO_BEFORE_MATCH_MS	(1500)
 
+/* when defined, the let the NDP use an external PDM clock */
+#define USE_EXTERNAL_PDM_CLOCK
```
Taejun-Kwon, line 81: Is there any case we do not use external PDM clock? we should use for all the time, am I right?
Even if there are some case we should not use external PDM clock, actually we should check gpio pin you defined in above file instead of constant value.
```
#ndo120.h
struct ndp120_lower_s {
	/* SPI related configs */
        struct spi_io_config spi_config;

	/* callback function: attach the ndp120 interrupt handler to the GPIO interrupt */
        CODE int (*attach)(ndp120_handler_t handler, FAR char *arg);

	/* eanble/disable ndp120 interrupts hadling on rlt8730e */
	CODE void (*irq_enable)(bool enable);
	CODE void (*external_pdm_enabled)(void); // new function
};

#define EXTERNAL_PDM_CLOCK_RATE 1536000
#define INTERNAL_PDM_CLOCK_RATE 768000
static int configure_audio(struct syntiant_ndp_device_s *ndp, unsigned int pdm_in_shift,
				unsigned int audio_tank_ms, unsigned int *audio_frame_bytes)
{
	memset(&pdm_config, 0, sizeof(pdm_config));
	
	/* if external pdm need to be used, then change clk mode * pdm rate */
	if (ndp->dev->lower->external_pdm_enabled()) {
		pdm_config.clk_mode = SYNTIANT_NDP120_CONFIG_VALUE_PDM_CLK_MODE_EXTERNAL;
		pdm_config.pdm_rate = EXTERNAL_PDM_CLOCK_RATE;
	} else {
		pdm_config.clk_mode = SYNTIANT_NDP120_CONFIG_VALUE_PDM_CLK_MODE_DUAL_INTERNAL;
		pdm_config.pdm_rate = INTERNAL_PDM_CLOCK_RATE;
	}
	pdm_config.interface = 0;
	pdm_config.clk = SYNTIANT_NDP120_CONFIG_VALUE_PDM_CLK_ON;
	pdm_config.sample_rate = 16000;
	....
}
```
I think this is proper approach.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
edwakuwaku: > I am not 100% sure who should really answer this, however: `CONFIG_AUDIO_NDP120` is already in the public codebase, here is a quick grep from a fresh clone of TizenRT repo done today (HEAD at commit [2901b73](https://github.com/Samsung/TizenRT/commit/2901b73b196c9b49e9970bd4a76db6ae69788deb))
> 
> ```
> ~/tmp/TizenRT (master ✔) grep -riIn CONFIG_AUDIO_NDP120
> ./os/drivers/audio/Make.defs:83:ifeq ($(CONFIG_AUDIO_NDP120),y)
> ./os/board/rtl8730e/src/Make.defs:75:ifeq ($(CONFIG_AUDIO_NDP120),y)
> ./os/board/rtl8730e/src/rtl8730e_ndp120.c:40:#ifdef CONFIG_AUDIO_NDP120
> ./os/board/rtl8730e/src/rtl8730e_boot.c:488:#ifdef CONFIG_AUDIO_NDP120
> ./os/include/tinyara/audio/ndp120.h:32:#ifdef CONFIG_AUDIO_NDP120
> ./os/include/tinyara/audio/ndp120.h:129:#endif							/* CONFIG_AUDIO_NDP120 */
> ./build/configs/rtl8730e/audio/defconfig:531:CONFIG_AUDIO_NDP120=y
> ~/tmp/TizenRT (master ✔)
> ```

Hi @cvejlbo , yeap it was merged recently, thank you for reminding~

edwakuwaku: Hello Taejun님, do you mean something like this? https://github.com/Samsung/TizenRT/pull/6348

Taejun-Kwon: > @Taejun-Kwon Hello Taejun님, do you mean something like this? #6348

Hello,
Yes. I mean this.
BTW is channel length of i2s can be changed dynamically in the middle of sending?

abhishek-samsung: Hello @Taejun-Kwon , I think this should be okay. However, the new amp is not verfied with datawidth as 16, so we will have to check with it. @cvejlbo , can ndp also work with 3MHz instead of 1.5MHz?

edwakuwaku: Hello~ The channel length cannot be changed dynamically during runtime. As it will affect the BLCK and result in error during data transfer.

os/board/rtl8730e/src/rtl8730e_alc1019.c
cvejlbo: Added a commit that moves the enabling of DMIC via PA_25 to `rtl8730e_ndp120_initialize()` function instead

os/drivers/ai-soc/ndp120/src/ndp120_api.c
cvejlbo: From a couple of discussions (also with @abhishek-samsung) it sounds like the optimal approach is to integrate the selection of external (I2S BCLK) vs internal (NDP generated) PDM clock with the PM module, where we can register callbacks to be called when entering sleep state and normal/wakeup state respectively. 

My suggestion would be to keep the current behaviour (where it's controlled by a #define) in this PR, and then we can add a new PR with the relevant changes with PM callbacks subsequently.
Is the PM code for RTL8730e in a functional state where we already implement and test this?

os/arch/arm/src/amebasmart/amebasmart_i2s.c
cvejlbo: > Hello @Taejun-Kwon , I think this should be okay. However, the new amp is not verfied with datawidth as 16, so we will have to check with it. @cvejlbo , can ndp also work with 3MHz instead of 1.5MHz?

In general, the NDP can work with 3.072MHz PDM clock as well, however the DMIC will definitely have higher power consumption, and it may also have a different sensitivity given that clock rate.
We'd also need to check/review the impact on NDP side (adjust for sensitivity levels, etc) with the given DMIC at 3.072MHz if the PDM clock rate is changed.

os/drivers/ai-soc/ndp120/src/ndp120_api.c
edwakuwaku: Hello @cvejlbo , previously we got a request regarding the integration of I2S BCLCK and NDP PDM CLOCK.
This PR is doing the necessary operation as per our understanding, please check https://github.com/Samsung/TizenRT/pull/6304.
But, it was yet to verified with the NDP, could you help to review the PR and check whether it meets the requirement?

abhishek-samsung: Hello @cvejlbo , yes. PM related changes are taking a bit more time than expected. Also, we havent enabled pm on product code yet for bixby. So, the current PR would be okay to merge.

@edwakuwaku , in PR #6304 , we are only setting the gpio pin state. However, we also need to run some api's from NDP side in addition to changing the gpio pin state. As a result, that PR is not fully enough. Currently PM related engineers are working on verifying the callback feature in PM. Once its done, I think, we should combine the changes in PR #6304 + ndp changes and add them to callbacks.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
cvejlbo: > @cvejlbo , thanks for explaining. Let us stick with 1.5MHz. This file change can be removed as we will merge #6348 for this.

@abhishek-samsung, Sounds good. In order to not break anything, I guess it's OK if I wait with reverting the changes in `amebasmart_i2s.c` until #6348 is merged?

abhishek-samsung: Sure, we will merge #6348 asap.

abhishek-samsung: @cvejlbo , we merged https://github.com/Samsung/TizenRT/pull/6348, can you please update the PR to include the new changes?

cvejlbo: > @cvejlbo , we merged #6348, can you please update the PR to include the new changes?

Done

abhishek-samsung (approved):

sunghan-chang:
@cvejlbo Could you clean up the PR? You can squash all of commits as a single commit and can rewrite the title with our format and can write details in the commit description for this.

Taejun-Kwon (approved):

Taejun-Kwon:
Please resolve conflicts.

os/drivers/ai-soc/ndp120/src/ndp120_api.c
edwakuwaku: Hello @abhishek-samsung , thanks for updating~

PR #6340, opened 2024-08-14
os/drivers: change recording flow of ndp120 as per Syntiant review

abhishek-samsung:


cvejlbo:
I have looked at the changes, and they mostly look OK. 
However, when running this for testing the behavior, I notice that sometimes a bit of audio between keyword & query part is missing.
Looking further at this, I suggest the following changes:

```
diff --git a/os/drivers/ai-soc/ndp120/src/ndp120_api.c b/os/drivers/ai-soc/ndp120/src/ndp120_api.c
index 525fd38e1..72f4b01dd 100644
--- a/os/drivers/ai-soc/ndp120/src/ndp120_api.c
+++ b/os/drivers/ai-soc/ndp120/src/ndp120_api.c
@@ -1118,14 +1118,10 @@ int ndp120_irq_handler(struct ndp120_dev_s *dev)
        return work_queue(HPWORK, &ndp120_work, ndp120_irq_handler_work, (void *)dev, 0);
 }
 
-int ndp120_set_match_per_frame(struct ndp120_dev_s *dev, int on)
+int ndp120_set_sample_ready_int(struct ndp120_dev_s *dev, int on)
 {
        int s;
-       syntiant_ndp120_config_misc_t ndp120_config;
-       memset(&ndp120_config, 0, sizeof(ndp120_config));
-       ndp120_config.set = SYNTIANT_NDP120_CONFIG_SET_MISC_MATCH_PER_FRAME_ON;
-       ndp120_config.match_per_frame_on = on;
-       s = syntiant_ndp120_config_misc(dev->ndp, &ndp120_config);
+       s = syntiant_ndp120_config_notify_on_sample_ready(dev->ndp, on);
        return s;
 }
 
@@ -1144,10 +1140,6 @@ int ndp120_extract_audio(struct ndp120_dev_s *dev, uint8_t *audio_out_buffer,
        if (keyword_bytes_left != 0) {
                memcpy(audio_out_buffer, &keyword_buffer[dev->keyword_bytes - keyword_bytes_left], dev->sample_size);
                keyword_bytes_left -= dev->sample_size;
-               if (keyword_bytes_left == 0) {
-                       /* turn on MPF interrutps */
-                       s =  ndp120_set_match_per_frame(dev, 1);
-               }
                return SYNTIANT_NDP_ERROR_NONE;
        } else {
                /* wait for MPF interrupt */
@@ -1215,6 +1207,8 @@ int ndp120_start_sample_ready(struct ndp120_dev_s *dev)
 
        dev->recording = true;
 
+       s =  ndp120_set_sample_ready_int(dev, 1);
+
        if (include_keyword) {
                unsigned int extract_bytes = dev->keyword_bytes;
                s = syntiant_ndp_extract_data(dev->ndp, SYNTIANT_NDP_EXTRACT_TYPE_INPUT,
@@ -1229,13 +1223,13 @@ int ndp120_start_sample_ready(struct ndp120_dev_s *dev)
                include_keyword = false;
        }
 
-       return SYNTIANT_NDP_ERROR_NONE;
+       return s;
 }
 
 int ndp120_stop_sample_ready(struct ndp120_dev_s *dev)
 {
        int s;
-       s = ndp120_set_match_per_frame(dev, 0);
+       s = ndp120_set_sample_ready_int(dev, 0);
 
        dev->recording = false;
```

I believe the reason for the missing audio that I see would be that MPF/sample-rdy enabling syncs some buffer pointers between DSP and host, so that sync needs to be done at the proper time, and it is likely too late after all the per-frame-copying done for the keyword. Sinc we'll revert to using the sample ready interrupt anyhow going forward, we might as well do it here also.
It does not hurt to enable sample ready interrupt prior to actually extracting the keyword, so that's what I am doing here.
I also suggest that we now rename the `ndp120_set_match_per_frame` to `ndp120_set_sample_ready_int` to reflect what it's actually doing now.
It should be possible to get rid of using `SYNTIANT_NDP_EXTRACT_TYPE_INPUT_ANNOTATED` (and the corresponding skip of the annotations) and instead just use `SYNTIANT_NDP_EXTRACT_TYPE_INPUT` - but maybe to keep things simpler, we can also push that change ourselves as part of upcoming update.

abhishek-samsung:
Hello @cvejlbo , thanks for the review.  I have included the changes you provided in the diff. Also, I have removed annotated data extraction used normal SYNTIANT_NDP_EXTRACT_TYPE_INPUT instead. I want to understand few things. What is the difference between MPF and sample ready? Because after applying all the changes, the audio extraction flow is mostly same (even we enabled sample ready before keyword extraction) , only difference is that keyword is extracted in one go. So, I am trying to understand how this is better compared to older approach.

os/drivers/ai-soc/ndp120/src/ndp120_api.c
```diff
@@ -629,10 +638,14 @@ static int configure_audio(struct syntiant_ndp_device_s *ndp, unsigned int pdm_i
 		goto errout_configure_audio;
 	}
 
-	audio_frame_bytes_ = misc_config.audio_frame_step
-		* SYNTIANT_NDP120_AUDIO_SAMPLES_PER_WORD;
+	*sample_size_orig_annot = misc_config.audio_sample_size_bytes;
```
Taejun-Kwon, line 641:
```
configure_audio(dev, DMIC_768KHZ_PDM_IN_SHIFT_FF, AUDIO_TANK_MS, &dev->sample_size, &dev->sample_size_orig_annot); //use dev instead of dev->ndp
...
dev->sample_size_orig_annot = misc_config.audio_sample_size_bytes;
```

os/drivers/audio/ndp120_voice.c
```diff
@@ -469,8 +467,8 @@ static int ndp120_enqueuebuffer(FAR struct audio_lowerhalf_s *dev, FAR struct ap
 	}
 
 	sq_entry_t *tmp;
-	apb->nbytes = CONFIG_NDP120_BUFFER_SIZE;
-	ndp120_extract_audio(priv, apb->samp, CONFIG_NDP120_BUFFER_SIZE);
+	apb->nbytes = priv->sample_size;
+	ndp120_extract_audio(priv, apb->samp, priv->sample_size);
```
Taejun-Kwon, line 471:
```
ndp120_extract_audio(priv, apb);
```
and then ndp120_extract_audio should handle it & fill buffer & update apb->nbytes.

Taejun-Kwon (approved):

PR #6342, opened 2024-08-16
Add CPU option for ps and cpuload commands

kishore-sn:
The new cpu option lets you print the details for the required cpu only.


sunghan-chang (approved):

apps/system/utils/utils_cpuload.c
```diff
@@ -116,18 +117,21 @@ static void cpuload_stop(void)
...
-#else
-	if (strcmp(load_info, "0.0") == 0) {
-#endif
-		return false;
-	}
+	bool is_nonzero[CONFIG_SMP_NCPUS + 1] = {0, };
+	is_nonzero[0] = (strncmp(load_info, "0.0", 3) != 0);
+	is_nonzero[2] = (strncmp(&load_info[len - 4], "-0.0", 4) != 0);
+	load_info = strchr(load_info, '-'); 
+	is_nonzero[1] = (strncmp(load_info, "-0.0", 4) != 0);
```
ritesh55555, line 131: seems like we are only taking 2 core case with smp. Is it okay?

apps/system/utils/utils_cpuload.c
```diff
@@ -223,19 +239,21 @@ static void cpuload_print_normal(void)
...
+				}
+			} else {
+				printf("  CPU%d |", j);
+			}
 		}
 #endif
-#ifndef CONFIG_SCHED_MULTI_CPULOAD
-		printf("  Avg  |");
-#else
-		printf("  Avg%d |", i - PROC_STAT_CPULOAD_SHORT + 1);
```
ritesh55555, line 236: We dont want an extra column of average of two core's cpuload?

kishore-sn: No. I feel the avg load is misleading in smp case. Let us remove it.

apps/system/utils/utils_cpuload.c
kishore-sn: Thank you for pointing it out. I fixed it.

ritesh55555 (approved):

apps/system/utils/utils_cpuload.c
```diff
@@ -116,18 +117,30 @@ static void cpuload_stop(void)
  * So for zero load case, we check if avg_load is 0.0 or not.
  * Note: we compare with "-0.0" because if avg_load is 90.0, then
  * also we will get true if we compare with 0.0 */
+
 static bool has_cpuload(char *load_info)
 {
-#ifdef CONFIG_SMP
+#ifndef CONFIG_SMP
```
sunghan-chang, line 123: `#ifndef`
`n` is difficult to see. How about changing the conditional?
```
#ifdef CONFIG_SMP
    int len = strlen(load_info);
    if (cpu == CONFIG_SMP_NCPUS) {
    ...
#else
   return .....;
#endif
```

r-prabu (approved):
changes of adding CPU option for ps and cpuload commands looks good.

apps/system/utils/utils_cpuload.c
```diff
@@ -116,18 +117,30 @@ static void cpuload_stop(void)
...
+		return (strncmp(load_info, "0.0", 3) != 0);
+	} else {
+		char *str = load_info;
+		for (int i = 1; i < CONFIG_SMP_NCPUS && i <= cpu; i++) {
+			str = strchr(str, '-') + 1;
+		}
+		return (strncmp(str, "0.0", 3) != 0);
+	}
+#else 
+	return (strcmp(load_info, "0.0") != 0);
```
sunghan-chang, line 142: If you don't use the `strncmp`, SVACE will give you a warning I think. And other lines already use strncmp.

And Is it ok with length 3 for "0.0"? not 4? ok with length 4 for "-0.0"?

apps/system/utils/utils_ps.c
```diff
@@ -98,12 +98,19 @@ static const char *utils_ttypenames[4] = {
 	"--?--  "
 };
 
+#ifdef CONFIG_SMP
+uint32_t cpu;
```
sunghan-chang, line 102: Global variable is necessary?
If yes, how about static?

kishore-sn: Changed to static.

apps/system/utils/utils_cpuload.c
kishore-sn: Changed to strncmp. 
And it is ok to compare just 3 chars for 0.0 and 4 for -0.0 because it might not be end of string.

apps/system/utils/utils_cpuload.c
```diff
@@ -223,19 +248,21 @@ static void cpuload_print_normal(void)
...
 #ifdef CONFIG_SMP
 		for (int j = 0; j < CONFIG_SMP_NCPUS; j++) {
-			printf("  CPU%d |", j);
+			if (cpu < CONFIG_SMP_NCPUS) {
+				if (cpu == j) {
+					printf("  CPU%d |", j);
+				}
+			} else {
+				printf("  CPU%d |", j);
+			}
```
sunghan-chang, line 262:
```
if ((cpu >= CONFIG_SMP_NCPUS) || (cpu == j)) {
    printf(....);
}
```
I don't know whether this reduces flash usage or not.

kishore-sn: Done.

apps/system/utils/utils_cpuload.c
```diff
@@ -169,9 +182,15 @@ static void cpuload_print_pid_value(char *buf, void *arg)
...
+			for (int j = 0; j < CONFIG_SMP_NCPUS; j++) {
 				avgload[j] = strtok_r(avgload[j], "-", &avgload[j + 1]);
-				printf(" %5s |", avgload[j]);
+				if (cpu < CONFIG_SMP_NCPUS) {
+					if (cpu == j) {
+						printf(" %5s |", avgload[j]);
+					}
+				} else {
+					printf(" %5s |", avgload[j]);
+				}
```
sunghan-chang, line 193: Here and line205~211 have the same flow. You can ignore my suggestion but I think we should do the same way for the same.

kishore-sn: Fixed. Thanks.

sunghan-chang (approved):

PR #6343, opened 2024-08-16
os/board: temporary change to enable DMIC on rev 6 board

abhishek-samsung:


sunghan-chang:
@abhishek-samsung `temporary` means?

abhishek-samsung:
> @abhishek-samsung `temporary` means?

I thought this gpio change was required for rev 6 boards with modifications (speaker related), so raised the PR. But upon testing this is not required. I will be closing this PR as the modified boards work even without this PR.

PR #6347, opened 2024-08-20
sched/sched_setpriority: fix cond wait deadlock issue

ewoodev:
There is an issue where a deadlock occurs when using pthread_cond_wait().
If the priority of the thread calling pthread_cond_signal is higher than that of the thread
calling pthread_cond_wait, a deadlock is occuring.

The cause is that even though sched_lock was called within the pthread_cond_wait function,
during the PRIORITY INHERITANCE operation in the sem_post operation.

So, It should have take sem of cond after give mutex in pthread_cond_wait, It can't take sem
because it give up cpu in this time, resulting in a leak of cond signals.

In the sched_setpriority() function, it changes its own priority while giving up the CPU.

Therefore, in the case of sched_lock, so as not to give up the CPU, threads with priorities
higher than the changed priority of the running thread are changed to pending, ensuring
the sched lock operation of the running thread.

### DIFF
--- a/os/kernel/sched/sched_setpriority.c
+++ b/os/kernel/sched/sched_setpriority.c
@@ -174,10 +174,11 @@ static FAR struct tcb_s *sched_nexttcb(FAR struct tcb_s *tcb)
 int sched_setpriority(FAR struct tcb_s *tcb, int sched_priority)
 {
 	FAR struct tcb_s *rtcb = this_task();
+	FAR struct tcb_s *ntcb;
 	tstate_t task_state;
 	irqstate_t saved_state;
+	
 #ifdef CONFIG_SMP
-	FAR struct tcb_s *ntcb;
 	int cpu;
 #endif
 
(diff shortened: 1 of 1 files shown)


os/kernel/sched/sched_setpriority.c
```diff
@@ -211,14 +212,36 @@ int sched_setpriority(FAR struct tcb_s *tcb, int sched_priority)
...
+					ntcb->task_state = TSTATE_TASK_PENDING;
 
-			up_reprioritize_rtr(tcb, (uint8_t)sched_priority);
+#ifdef CONFIG_SMP
+					ntcb = sched_nexttcb(tcb);
+#else
+					ntcb = tcb->flink;
+#endif
+				}
+				while (sched_priority < ntcb->sched_priority);
```
sunghan-chang, line 237: coding rule.

} while (...);

vibhor-m:
Thank You, I check in sample application. These changes work fine.

r-prabu (approved):
changes in the TSTATE_TASK_RUNNING state looks fine for addressing the deadlock case.

sunghan-chang (approved):

kishore-sn (approved):

PR #6348, opened 2024-08-20
os/arch/arm/src/amebasmart: Add macro for i2s tx channel length

edwakuwaku:
- Added I2S tx channel length macro for user to modify according to usage
- Supported variable range between 8~32


abhishek-samsung (approved):

Taejun-Kwon (approved):
I think this can be applied.

Taejun-Kwon:
@edwakuwaku because of urgent schedule, I modified title & merge. please understand us.

edwakuwaku:
> @edwakuwaku because of urgent schedule, I modified title & merge. please understand us.

Hello @Taejun-Kwon , sorry for late reply due to other tasks.

PR #6354, opened 2024-08-21
drivers/audio: add volume formula for syu645b

ritesh55555:
This commits add volume formula for syu645b audio driver such that the volume range of speaker hardware is normalized to media volume from 1 to 10 . 

The formula to convert media volume to hardware volume is approximated to a linear equation .It is observed that there is minimum hardware volume below which we cannot hear the volume, so by taking that value as minimum value, this formula as made linear equation (taking slope as (hardware maximum - hardware minimum) / (10 - 0) )


os/drivers/audio/syu645b.c
```diff
@@ -173,7 +173,8 @@ static void syu645b_setvolume(FAR struct syu645b_dev_s *priv)
 			if (priv->volume == 0) {
 				codec_set_master_volume_script[0].val[0] = 0;
 			} else {
-				codec_set_master_volume_script[0].val[0] = SYU645B_HW_VOL_MAX - (SYU645B_SPK_VOL_MAX - priv->volume) * SYU645B_HW_VOL_STEP;
+				/* Linear approximation is done to convert media volume to hardware volume */
+				codec_set_master_volume_script[0].val[0] = SYU645B_HW_VOL_MIN_BOUND +  SYU645B_HW_VOL_SLOPE * priv->volume;
```
kishore-sn, line 177: Remove extra space : ` +  SYU645B_HW_VOL_SLOPE`

os/drivers/audio/syu645b.h
```diff
@@ -62,11 +62,12 @@
...
-#define SYU645B_HW_VOL_MAX		255
-#define SYU645B_HW_VOL_STEP		7
+#define SYU645B_HW_VOL_MAX	    255
 
-/* SYU645B_HW_VOL_DEF = (SYU645B_HW_VOL_MAX - (SYU645B_SPK_VOL_MAX - SYU645B_SPK_VOL_DEF) * SYU645B_HW_VOL_STEP) */
-#define SYU645B_HW_VOL_DEF		234
+/* SYU645B_HW_VOL_DEF = SYU645B_HW_VOL_MIN_BOUND +  SYU645B_HW_VOL_SLOPE * SYU645B_SPK_VOL_DEF */
+#define SYU645B_HW_VOL_DEF	    214
+#define SYU645B_HW_VOL_MIN_BOUND    130       /* refers to the minimum hardware volume audible */
+#define SYU645B_HW_VOL_SLOPE        (SYU645B_HW_VOL_MAX - SYU645B_HW_VOL_MIN_BOUND) / SYU645B_SPK_VOL_MAX
```
kishore-sn, line 70: The slope actually comes to 12.5. But it might get rounded off to 12 if considered as int. Then if we try to set media volume 10, our hw volume comes to 250 instead of 255. How about using (float) cast?

os/drivers/audio/syu645bscripts.h
```diff
@@ -73,7 +73,8 @@ t_codec_init_script_entry codec_initial_script[] = {
 	{ 0x13, {0x06,}, 30, SYU645B_REG_D_2BYTE},
 	{ 0x14, {0x06,}, 30, SYU645B_REG_D_2BYTE},
 	{ 0x1E, {0x05,}, 30, SYU645B_REG_D_2BYTE},
-	{ 0x07, {0xFF,}, 30, SYU645B_REG_D_2BYTE},
+	/* setting to default volume */
+	{ 0x07, {SYU645B_HW_VOL_DEF,}, 30, SYU645B_REG_D_2BYTE},
```
edwakuwaku, line 77: All the scripts seems to be read only, how about mark it as const, it could save up spaces for ram.

Taejun-Kwon:
I don't mean this, we are doing commecialization.
Supose that 0 db is 1,
then 10db is 10, and 20db is 100.
So we should define min & max first.
For example, if we define min as 80 and max as 100,
then there's 100 level of sound is there.
We should calculate based on log function to make each step increase/decrease 10 level of volume. It's not simple logic as this PR

abhishek-samsung:
Hello @ritesh55555 , I found this article online and it explains the issue suggested by @Taejun-Kwon very well,

https://www.dr-lex.be/info-stuff/volumecontrols.html

You can even compare between linear and log based volume sliders. Please take a look at it when you get time. 

Also, I think we can customize the curve as per our requirement. May be we can even make it more general as many amplifiers only allow us to set volume in dB, so these changes will be same for  all of them with varying parameters to tune the curve.

Taejun-Kwon (approved):
Many thing need to be modify later.
1. Proper volume value.
2. delay in init script
3. i2s time out based on buffer size & samplerate & pcm format & number of channel

PR #6355, opened 2024-08-22
Minor fixes in assert log and trap tool  (Under Verification)

neel-samsung:
This PR resolves issue faced while using TRAP when the code asserts in IRQ state.


sunghan-chang (commented):
```
tools/trap/cli: minor fixes in trap tool

TRAP fails to to generate output when code asserts in IRQ state.
So, this commit fix this issue and TRAP output gets generated.
```
In my opinion, `minor fixes` can be non-operation change. It looks it's not minor.
And I can't get what commit does from the title. Let's rewrite the title to represent what commit does.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -287,7 +287,7 @@ static void check_sp_corruption(uint32_t sp, uint32_t *stackbase, uint32_t *stac
 {
 	lldbg_noarg("===========================================================\n");
 	lldbg_noarg("Asserted task's stack details\n");
-	lldbg_noarg("===========================================================\n");	
+	lldbg_noarg("===========================================================\n");
```
Tirtha-Samsung, line 290: Please revert this extra space at the end

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -306,12 +306,12 @@ static void check_sp_corruption(uint32_t sp, uint32_t *stackbase, uint32_t *stac
 {
 	lldbg_noarg("===========================================================\n");
 	lldbg_noarg("Asserted task's stack details\n");
-	lldbg_noarg("===========================================================\n");	
+	lldbg_noarg("===========================================================\n");
```
Tirtha-Samsung, line 309: Please revert this extra space at the end

os/arch/arm/src/armv7-a/arm_assert.c
neel-samsung: There is no extra space at the end rather I have removed it.

os/arch/arm/src/armv8-m/up_assert.c
neel-samsung: same as above

sunghan-chang (approved):

r-prabu (approved):

PR #6356, opened 2024-08-22
apps/shell: fix board sleep during TASH Input

gSahitya-samsung:
This commit resolve the issue of board sleep during TASH Input. With the
PM Quick Sleep the board is frequently going to sleep whenever user
tries to enter the command in tash prompt. This issue cause the board to
wake frequently resulting in high power consumption. To avoid this we
have suspended the PM Sleep for TASH_PM_TIMEDSUSPEND_TIME_IN_MS whenever
TASH read user's input.


apps/shell/tash_init.c
```diff
@@ -58,6 +58,11 @@
 #include <tinyara/usb/cdcacm.h>
 #endif
 #include "tash_internal.h"
+#include <sys/ioctl.h>
```
sunghan-chang, line 61: Is this necessary regardless of CONFIG_PM?

apps/shell/tash_internal.h
```diff
@@ -91,6 +91,11 @@ extern int tash_sleep(int argc, char **args);
 #ifdef CONFIG_TASH_USLEEP
 extern int tash_usleep(int argc, char **args);
 #endif
+#ifdef CONFIG_PM
+#define TASH_PM_DOMAIN_NAME "TASH"
+#define PM_DRIVER "/dev/pm"
+extern int tash_pm_domain_id;
```
sunghan-chang, line 97: Instead of using global variable, let's set it as a static variable and provide a new function to get or to set it.

apps/shell/tash_init.c
```diff
@@ -160,6 +165,20 @@ static int tash_nonftdi_usbconinit(void)
...
+	if (pmdrv_fd >= 0) {
+		if (ioctl(pmdrv_fd, PMIOC_DOMAIN_REGISTER, &pm_domain_arg) != OK) {
+			shdbg("pm_domain_register failed(%d)\n", get_errno());
+		}
+		tash_pm_domain_id = pm_domain_arg.domain_id;
+		close(pmdrv_fd);
+	} else {
+		shdbg("open %s failed(%d), \n", PM_DRIVER, get_errno());
+	}
+#endif
```
sunghan-chang, line 181: How about making a new file like tash_pm.c and using tash_pm function in tash main files instead of coding them at several location?

apps/shell/tash_main.c
```diff
@@ -156,9 +167,18 @@ char *tash_read_input_line(int fd)
 				shdbg("TASH: can not read uart\n");
 				return buffer;
 			}
-
+#ifdef CONFIG_PM
+			pmdrv_fd = open(PM_DRIVER, O_WRONLY);
```
kishore-sn, line 171: We dont need to open the pm device for each character. We can open it in tash_init and keep it open as long as tash is alive.

apps/shell/tash_main.c
```diff
@@ -156,9 +167,18 @@ char *tash_read_input_line(int fd)
...
+			if (pmdrv_fd < 0) {
+				shdbg("open %s failed(%d), \n", PM_DRIVER, get_errno());
+			}
+#endif
 			for (char_idx = 0; char_idx < nbytes; char_idx++) {
-
+#ifdef CONFIG_PM
+				if ((pmdrv_fd >= 0) && (ioctl(pmdrv_fd, PMIOC_TIMEDSUSPEND, &pm_suspend_arg) != OK)) {
```
kishore-sn, line 178: At this point, we have multiple characters in our buffer and we are processing one character at a time. I dont think we need to suspend the pm here. The pm suspend must be called before read/poll call, it at the first line of the do-while loop.

apps/shell/tash_internal.h
```diff
@@ -91,6 +91,11 @@ extern int tash_sleep(int argc, char **args);
 #ifdef CONFIG_TASH_USLEEP
 extern int tash_usleep(int argc, char **args);
 #endif
+#ifdef CONFIG_PM
+#define TASH_PM_DOMAIN_NAME "TASH"
+#define PM_DRIVER "/dev/pm"
```
r-prabu, line 96: Is this ("/dev/pm") not exposed in the <tinyara/pm/pm.h>?

apps/shell/tash_init.c
gSahitya-samsung: No

apps/shell/tash_internal.h
gSahitya-samsung: Done

apps/shell/tash_init.c
gSahitya-samsung: Done

apps/shell/tash_main.c
gSahitya-samsung: Done

apps/shell/tash_main.c
gSahitya-samsung: I have created the new `tash_pm_timedsuspend` API

apps/shell/tash_internal.h
gSahitya-samsung: No

apps/shell/tash_main.c
sunghan-chang: @gSahitya-samsung Could you check @kishore-sn 's comment again? Changed code also opens and closes the pm device at every character.

apps/shell/tash_internal.h
sunghan-chang: pm.c has `#define PM_DRVPATH  "/dev/pm". I think @r-prabu said you should use the same define for the same driver.
There are two files which separately defines `"/dev/pm"` already. It causes mismatch or missing when we modify the driver name. Let's define it at some pm file and use it at every file.

gSahitya-samsung: Shall I add it to `<tinyara/pm/pm.h>` ?

apps/shell/tash_main.c
```diff
@@ -55,6 +55,10 @@ enum tash_input_state_e {
 #define TASH_PROMPT           "TASH>>"
 #endif							/* CONFIG_TASH */
 
+#ifdef CONFIG_PM
+#define TASH_PM_TIMEDSUSPEND_TIME_IN_MS 1000
```
sunghan-chang, line 59: After input the enter key, can we input next always in 1s? Sometimes yes, but sometimes no. Let me suggest 5s. It's the TASH so that 1s -> 5s does not affect power consumption almost. And because it's user intend so that it's ok with more power consumption.

gSahitya-samsung: Ok I will change it 5s

apps/shell/tash_internal.h
sunghan-chang: I think yes. You can decide the location.

apps/shell/tash_main.c
```diff
@@ -149,14 +153,17 @@ char *tash_read_input_line(int fd)
 		FD_SET(fd, &tfd);
 
 		if ((select(fd + 1, &tfd, NULL, NULL, &stimeout)) && FD_ISSET(fd, &tfd)) {
+#endif
+#ifdef CONFIG_PM
+			/* Ensure board does not go to sleep for TASH_PM_TIMEDSUSPEND_TIME_IN_MS*/
+			tash_pm_timedsuspend(TASH_PM_TIMEDSUSPEND_TIME_IN_MS);
```
abhishek-samsung, line 159: Placement of this looks a bit weird, I think we are expected to wait for N seconds before going to sleep after the user has entered a character, right? Here we are suspending sleep for N seconds even when user has not entered anything. Shouldnt this be below the read function?

```
			nbytes = read(fd, &buffer[pos], (bufsize - pos));
			if (nbytes < 0) {
				shdbg("TASH: can not read uart\n");
				return buffer;
			}
                        /* Ensure board does not go to sleep for TASH_PM_TIMEDSUSPEND_TIME_IN_MS as user started entering input*/
		------->tash_pm_timedsuspend(TASH_PM_TIMEDSUSPEND_TIME_IN_MS);
```

apps/shell/tash_main.c
gSahitya-samsung: Done

apps/shell/tash_internal.h
gSahitya-samsung: Done

apps/shell/tash_main.c
gSahitya-samsung: Done

sunghan-chang (approved):

kishore-sn (dismissed):

apps/shell/tash_main.c
```diff
@@ -357,6 +368,9 @@ static int tash_main(int argc, char *argv[])
 
 		tash_free(line_buff);
 	} while (tash_running);
+#ifdef CONFIG_PM
+	tash_pm_close_driver();
+#endif
```
r-prabu, line 373: why can't we do this inside the tash_stop function?

apps/shell/tash_command.c
```diff
@@ -564,9 +564,9 @@ int tash_execute_cmd(char **args, int argc)
 			if (tash_cmds_info.cmd[cmd_idx].exec_type == TASH_EXECMD_SYNC) {
 				/* function call to execute SYNC command */
 #ifdef CONFIG_PM
-				pmdrv_fd = open("/dev/pm", O_WRONLY);
+				pmdrv_fd = open(PM_DRVPATH, O_WRONLY);
```
r-prabu, line 567: Do we need to open the pm driver here as well?

apps/shell/tash_main.c
gSahitya-samsung: We are not using the `pm_driver` after `while` loop. So, it's better to close it here only.

r-prabu: yes but tash_running this cleared in the tash_stop function only.

apps/shell/tash_command.c
gSahitya-samsung: Yes, to suspend `TASH` domain

apps/shell/tash_main.c
gSahitya-samsung: Yes, moved the `tash_pm_close_driver` inside `tash_stop`

r-prabu (approved):

PR #6357, opened 2024-08-22
Enable affinity apis irrespective of SMP

kishore-sn:
**Enable affinity setting even if SMP is disabled**

The kernel ABI should remain unchanged due to SMP feature, so that apps
can be easily ported irrespective of SMP. So, we make affinity related
variables and functions available to apps irrespective of SMP feature.
If SMP is disabled by kernel, then using the affinity by app returns an
error code.

**Add kernel tc for affinity apis:**
- pthread_setaffinity_np
- pthread_getaffinity_np
- sched_setaffinity
- sched_getaffinity


sunghan-chang:
Could you let me know builds for smp enabled and disabled?

kishore-sn:
> Could you let me know builds for smp enabled and disabled?

I have tested it by manually enabling and disabling smp on loadable apps config.

r-prabu (approved):

apps/examples/testcase/le_tc/kernel/tc_sched.c
```diff
@@ -673,6 +673,33 @@ static void tc_sched_task_setcanceltype(void)
...
+static void tc_sched_set_get_affinity(void)
+{
+	cpu_set_t affinity = 1 << 0;	/* Set affinity to core zero */
+#ifdef CONFIG_SMP
+	TC_ASSERT_EQ("sched_setaffinity", sched_setaffinity(0, sizeof(cpu_set_t), &affinity), OK);
+	affinity = 0;
+	TC_ASSERT_EQ("sched_getaffinity", sched_getaffinity(0, sizeof(cpu_set_t), &affinity), OK);
+	TC_ASSERT_EQ("sched_getaffinity", affinity, (1 << 0));
+#else
+	TC_ASSERT_EQ("sched_setaffinity", sched_setaffinity(0, sizeof(cpu_set_t), &affinity), -EINVAL);
```
abhishek-samsung, line 696: I think the goal is to use same code for both single core and multi core case. If so, this will not work. For example if the below code exists in some thread, it will only move forward in multi core case.

```
.....
if (sched_setaffinity(0, sizeof(cpu_set_t), &affinity) != OK) {
        printf("setaffinity failed, exit\n");
        exit();
} 
/* do work related to thread */
.....
```
and it fails for non smp case always.

kishore-sn: Yes. You are right. It will fail and return error. 
This is what @sunghan-chang  asked me to do.

os/include/sys/syscall.h
```diff
@@ -99,13 +99,13 @@
 #define SYS_sched_yield                (CONFIG_SYS_RESERVED + 12)
 #define SYS_set_errno                  (CONFIG_SYS_RESERVED + 13)
 
-#ifdef CONFIG_SMP
 #define SYS_sched_getaffinity          (CONFIG_SYS_RESERVED + 14)
 #define SYS_sched_setaffinity          (CONFIG_SYS_RESERVED + 15)
+#ifdef CONFIG_SMP
```
sunghan-chang, line 104: In the same point of my request, we should enable all syscall API regardless of related configs. But the config is disabled, then we can do above among the options. 
We can do this with new PR.

apps/examples/testcase/le_tc/kernel/tc_sched.c
kishore-sn: As per discussion, I modified to ENOTSUP

apps/examples/testcase/le_tc/kernel/tc_pthread.c
```diff
@@ -1605,6 +1605,51 @@ static void tc_pthread_pthread_mutex_consistent(void)
...
+	TC_ASSERT_EQ("pthread_getaffinity_np", pthread_getaffinity_np(pthread, sizeof(attr.affinity), &affinity), OK);
+	TC_ASSERT_EQ("pthread_getaffinity_np", affinity, (1 << 0));
+	ret_chk = pthread_join(pthread, &p_value);
+	TC_ASSERT_EQ("pthread_join", ret_chk, OK);
+#else
+	ret_chk = pthread_create(&pthread, &attr, self_pthread_join_n_exit, NULL);
+	TC_ASSERT_EQ("pthread create", ret_chk, EINVAL);
+	ret_chk = pthread_create(&pthread, NULL, self_pthread_join_n_exit, NULL);
+	TC_ASSERT_EQ("pthread create", ret_chk, OK);
+	TC_ASSERT_EQ("pthread_setaffinity_np", pthread_setaffinity_np(pthread, sizeof(attr.affinity), &affinity), EINVAL);
```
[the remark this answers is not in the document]
kishore-sn: Modfied

r-prabu (dismissed):

apps/examples/testcase/le_tc/kernel/tc_pthread.c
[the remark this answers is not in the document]
sunghan-chang: Really sorry @kishore-sn . To keep the same with Linux, could you change it to Invalid? So sorry.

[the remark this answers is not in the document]
kishore-sn: Done. Thanks.

sunghan-chang:
@r-prabu Let's merge this.

PR #6358, opened 2024-08-22
Media framework minor updates

vibhor-m:


sunghan-chang (commented):
@vibhor-m We don't allow empty description in commit. Could you write details there?

Taejun-Kwon (approved):

PR #6360, opened 2024-08-23
media/utils: Add check for ID3v2 tag in mp3 file format

aadotverma:
media/utils: Add check for ID3v2 tag in mp3 file format

### DIFF
--- a/framework/src/media/utils/MediaUtils.cpp
+++ b/framework/src/media/utils/MediaUtils.cpp
@@ -83,6 +83,8 @@ namespace utils {
 				| ((header[7] & 0x7f) << 14) \
 				| ((header[8] & 0x7f) << 7) \
 				| ((header[9] & 0x7f)) )
+// MP3 ID3v2 footer length
+#define MP3_ID3V2_FOOTER_LENGTH 10
 
 // Frame size = frame samples * (1 / sample rate) * bitrate / 8 + padding
 //            = frame samples * bitrate / 8 / sample rate + padding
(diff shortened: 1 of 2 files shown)


vibhor-m:
Code changes are OK.

Taejun-Kwon (commented):
I think you can optimize isMP3 for both of streambuffer & file.

I think it will not take long time but if you think it need more time then you can put some TODO comment.

vibhor-m:
> I think you can optimize isMP3 for both of streambuffer & file.
> 
> I think it will not take long time but if you think it need more time then you can put some TODO comment.

changes can be done now

vibhor-m (approved):
code changes are OK

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -1002,23 +1028,19 @@ bool file_header_parsing(FILE *fp, audio_type_t audioType, unsigned int *channel
...
-			medvdbg("string id3v2 tag found\n");
-			/* todo: check for extended header and parse title, artist etc details */
-			fseek(fp, 128, SEEK_SET);
-			break;
-		} else {
-			medvdbg("string id3v2 tag not found\n");
-			fseek(fp, 0, SEEK_SET);
-			break;
+		unsigned char id3Header[MP3_ID3V2_HEADER_LENGTH];
+		ret = fread(id3Header, sizeof(unsigned char), MP3_ID3V2_HEADER_LENGTH, fp);
```
vibhor-m, line 1039: lets perform this fread operation also inside checkId3V2Header header. we have passed isFile (true/false) param.

framework/src/media/utils/MediaUtils.cpp
```diff
@@ -617,18 +619,42 @@ static bool isMpeg2Ts(const unsigned char *buffer, size_t size)
 	return true;
 }
 
-static bool isMp3(const unsigned char *buffer, size_t size)
+static bool checkId3V2Header(const unsigned char *buffer, size_t size, bool isFile, FILE *fp = nullptr)
```
vibhor-m, line 622: only fp pointer is enough, if fp is NULL lets use buffer else consider file

aadotverma: Changes done

vibhor-m (approved):
code changes are OK

jsdosa (approved):
LGTM.

PR #6362, opened 2024-08-26
os/board/rtl8730e: ensure cache coherency on csi config API

jwei5:
- CSI config parameters will be passed from CA32 to KM4 for processing. Need to ensure config's buffer is aligned to prevent cache coherency issues
- Add an aligned temp buffer to store application's csi config to pass into KM4 to ensure the above point is met

### DIFF
--- a/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api_ext.c
+++ b/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api_ext.c
@@ -897,10 +897,19 @@ int wifi_csi_config(rtw_csi_action_parm_t *act_param)
 	int ret = 0;
 	u32 param_buf[1];
 
-	param_buf[0] = (u32)act_param;
-	DCache_Clean((u32)act_param, sizeof(rtw_csi_action_parm_t));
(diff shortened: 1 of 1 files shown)


sunghan-chang (dismissed):

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api_ext.c
```diff
@@ -897,10 +897,19 @@ int wifi_csi_config(rtw_csi_action_parm_t *act_param)
 	int ret = 0;
 	u32 param_buf[1];
 
-	param_buf[0] = (u32)act_param;
-	DCache_Clean((u32)act_param, sizeof(rtw_csi_action_parm_t));
+	rtw_csi_action_parm_t *act_param_temp = (rtw_csi_action_parm_t *)rtw_zmalloc(sizeof(rtw_csi_action_parm_t));
```
sunghan-chang, line 900: After this line, you do memcpy for entire buffer size. Do you need zmalloc? The zmalloc do malloc + memset with zero. `memset with zero` will be redundant.

jwei5: Agreed, zmalloc will not be necessary. I will make the changes, thanks

PR #6350, opened 2024-08-20
build/configs/common_download.sh: Changes needed for DDR

Tirtha-Samsung:
This change is needed for the external flash to be downloaded properly using package 1

With current flash partition and merged changes, loadable_ext wont build, we need to change the partition size to this
-CONFIG_FLASH_PART_SIZE="60,40,12,400,1844,5404,684,1844,5404,684,8,"
+CONFIG_FLASH_PART_SIZE="60,40,12,400,10036,5404,684,10036,5404,684,8,"

PR #6364, opened 2024-08-27
os/board: Define start address of second flash for binary manager

jeongarmy:


os/board/common/partitions.c
```diff
@@ -92,6 +94,7 @@ struct partition_data_s g_flash_part_data = {
 
 #ifdef CONFIG_SECOND_FLASH_PARTITION
 struct partition_data_s g_second_flash_part_data = {
+	CONFIG_SECOND_FLASH_START_ADDR,
```
Taejun-Kwon, line 97: #ifndef CONFIG_SECOND_FLASH_START_ADDR
#define CONFIG_SECOND_FLASH_START_ADDR 0
#endif

os/board/common/partitions.c
```diff
@@ -380,7 +383,7 @@ int configure_mtd_partitions(struct mtd_dev_s *mtd, int minor, partition_info_t
 
 #ifdef CONFIG_BINARY_MANAGER
 		if (!strncmp(types, "kernel,", 7)) {
-			binary_manager_register_kpart(g_partno, partsize, partoffset * geo.blocksize);
+			binary_manager_register_kpart(g_partno, partsize, part_data.start_addr + partoffset * geo.blocksize);
```
Taejun-Kwon, line 386: This is required only for binary manager?
I mean flash startaddress is stil 0x00000000 but binary manager should use 0x0a0000000 ?

jeongarmy: It seems that only binary manager is used now. If there is any hard-coded address for external flash access in realtek codes, they can user this value I think.

signing verification function has a parameter of actual flash address. So it should be set as actual address.

Taejun-Kwon: Then I think both of prev code & current approach also wrong.
Binary Manager also should use pointer of mtd to read data from flash, without information of start address...
Anyway you can't change it right now, but make sure that current design is wrong.

os/board/rtl8730e/Kconfig
```diff
@@ -65,6 +65,16 @@ config BOARD_FLASH_32M
 
 endchoice
 
+if SECOND_FLASH_PARTITION
+config SECOND_FLASH_START_ADDR
+    hex
+    default 0x09000000 if BOARD_FLASH_16M
+    default 0x0A000000 if BOARD_FLASH_32M
```
sunghan-chang, line 72: This could give confusion.
How about renaming them like?
```
BOARD_FLASH_XX -> BOARD_PRI_FLASH_XX
SECOND_FLASH_XX -> BOARD_SEC_FLASH_XX
```
PRI is primary, SEC is secondary.
This can be done with new PR.

os/board/common/partitions.c
jeongarmy: I agree with your opinion. BM already access flash with mtd driver. But signing API is served by realtek and it requires actual address so I made it like this.

os/board/common/partitions.c
jeongarmy: signing verification function has a parameter of actual flash address. So it should be set as actual address.
So I added #error condition to make build break.

os/board/rtl8730e/Kconfig
jeongarmy: Okay. I will update with new PR later.

sunghan-chang: @jeongarmy Please find @Tirtha-Samsung 's comment on PR #6350 . He said, with latest downloader, we can't write something between 0x0900 0000 ~ 0x0A00 0000 regardless of the internal flash size.

jeongarmy: I updated always 0xA000000 regardless flash size.

sunghan-chang (approved):

PR #6370, opened 2024-08-30
build/tools/amebasmart: change wifi fw log level

jwei5:
- Change log level of "warning report" log from warning to error to reduce print frequency
- Added condition to only print the log if there is an error status


sunghan-chang:
@jwei5 The `warning report` is txdma warning. Right?

jwei5:
Hi @sunghan-chang , yes this commit is for changing the txdma warning log. There is no dependency with other PR, this KM4 library does not include the changes from PR6368

sunghan-chang (approved):

PR #6374, opened 2024-09-03
os/board/rtl8730e: Change LCD update polling to semaphore

ziyik:
- Remove while loop waiting for next frame display
- Introduce sem_wait to signal the next frame is ready to update
- Reduce CPU loading from continue checking, LCD threads will be block and be scheduled only when the resource becomes available.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -87,6 +87,7 @@ LCDC_TypeDef *pLCDC = LCDC;
 LCDC_InitTypeDef lcdc_init_struct;
 static u32 UnderFlowCnt = 0;
 static u8 lcdc_nextframe = 0;
+static sem_t g_next_frame_block;
 
 struct irq lcdc_irq_info = {
 	.num = LCDC_IRQ,
@@ -122,6 +123,7 @@ static void rtl8730e_lcd_init(void)
 	rtl8730e_register_lcdc_isr();
 	LCDC_LineINTPosConfig(pLCDC, LCD_YRES * 4 / 5);
 	LCDC_INTConfig(pLCDC, LCDC_BIT_LCD_LIN_INTEN | LCDC_BIT_DMA_UN_INTEN | LCDC_BIT_LCD_FRD_INTEN, ENABLE);
+	sem_init(&g_next_frame_block, 0, 1);
 }
 
 static void rtl8730e_gpio_reset(void)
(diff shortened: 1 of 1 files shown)


os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -171,11 +173,9 @@ static void rtl8730e_lcd_put_area(u8 *lcd_img_buffer, u32 x_start, u32 y_start,
 #endif
 	LCDC_LayerConfig(pLCDC, LCD_LAYER, &lcdc_init_struct.layerx[LCD_LAYER]);
 	DCache_CleanInvalidate((u32)lcd_img_buffer, LCDC_IMG_BUF_SIZE);
-	lcdc_nextframe = 0;
+	sem_wait(&g_next_frame_block);
+	lcdc_nextframe = 1;
```
ewoodev, line 177: this flag is checked on thread and IRQ, So I thick, It is needed to protect critical section using irqsave()

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -171,11 +173,9 @@ static void rtl8730e_lcd_put_area(u8 *lcd_img_buffer, u32 x_start, u32 y_start,
...
 	LCDC_TrigerSHWReload(pLCDC);
```
ewoodev, line 178: this flag is checked on ISR, So I thick, This codes is needed to protect critical section using irqsave()

ewoodev: In between this codes, If get a IRQ vsync, our sem count can be broken or we can loss the vsync IRQ

ziyik: enter_critical_section() will disable IRQ, this case LCD IRQ will not be trigger, both are waiting for each other.

ewoodev:
```suggestion
	flag = irqsave()
	lcdc_nextframe = 1;
	LCDC_TrigerSHWReload(pLCDC);
	irqrestore(flag)
```

is it also not commig LCD IRQ??
is the thread waiting in LCDC_TrigerSHWReload() function?

ziyik: I misunderstand your previous point.
This should be working, as only 178 and 179 is protected, should we use enter_critical_section()?

sunghan-chang (approved):

PR #6377, opened 2024-09-03
audio/tinyalsa : Do not enqueue when apb buffer is not full to preven…

Taejun-Kwon:
…t noise

Some contents, for example mp3, we enqueue very small size of data. In this case some padding also included, it always make noise.

### DIFF
--- a/framework/src/tinyalsa/tinyalsa.c
+++ b/framework/src/tinyalsa/tinyalsa.c
@@ -346,7 +346,6 @@ static int pcm_set_config(struct pcm *pcm, const struct pcm_config *config)
 	pcm->config.period_count = buf_info.nbuffers;
 	pcm->buffer_size = buf_info.buffer_size;
 	pcm->buffer_cnt = buf_info.nbuffers;
-
 	return 0;
 }
 
@@ -483,6 +482,7 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
 	struct audio_msg_s msg;
 	unsigned int size;
 	int prio;
+	int ret;
 
 	if (pcm == NULL || data == NULL || frame_count == 0) {
 		return -EINVAL;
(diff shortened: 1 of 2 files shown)


framework/src/tinyalsa/tinyalsa.c
```diff
@@ -497,15 +497,17 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
...
 	while (pending > 0) {
-		nbytes = pending > pcm_frames_to_bytes(pcm, pcm->buffer_size) ? pcm_frames_to_bytes(pcm, pcm->buffer_size) : pending;
-		if (pcm->buf_idx < pcm->buffer_cnt) {
-			/* If we have empty buffers, fill them first */
-			memcpy(pcm->pBuffers[pcm->buf_idx]->samp, (char *)data + offset, nbytes);
-			apb = pcm->pBuffers[pcm->buf_idx];
-			pcm->buf_idx++;
-		} else {
+		bool found = false;
+		for (int i = 0; i < pcm->buffer_cnt; i++)
```
abhishek-samsung, line 502: I think we can make the logic a bit better, pcm->next_buf is not used in case of PCM_OUT type. How about using it instead?

I think it will be a bit faster too,


```
                if (pcm->next_buf != NULL) {
			apb = pcm->next_buf; --------->here
		if (pcm->buf_idx < pcm->buffer_cnt) {
			/* If we have empty buffers, fill them first */
			apb = pcm->pBuffers[pcm->buf_idx];
			pcm->next_buf = apb; ------------------>here
			pcm->buf_idx++;
		} else {
			/* We dont have any empty buffers. wait for deque message from kernel */
			size = mq_receive(pcm->mq, (FAR char *)&msg, sizeof(msg), &prio);
			if (size != sizeof(msg)) {
				/* Interrupted by a signal? What to do? */
				return oops(pcm, EINTR, "Interrupted while waiting for deque message from kernel\n");
			}
			if (msg.msgId == AUDIO_MSG_DEQUEUE) {
				apb = (struct ap_buffer_s *)msg.u.pPtr;
				pcm->next_buf = apb; ------------------> here
			} else if (msg.msgId == AUDIO_MSG_XRUN) {
				/* Underrun to be handled by client */
				return -EPIPE;
			} else {
				return oops(pcm, EINTR, "Recieved unexpected msg (id = %d) while waiting for deque message from kernel\n", msg.msgId);
			}
		}
```
and then pcm->next_buf can be enqueued in pcm_drain.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -497,53 +497,62 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
...
+		medvdbg("pending : %d nbytes : %d offset : %d apb : %x\n", pending, nbytes, offset, apb);
+		apb->nbytes += nbytes;
 		apb->curbyte = 0;
 		apb->flags = 0;
+
+		offset += nbytes;
+		pending -= nbytes;
+		
 		bufdesc.numbytes = apb->nbytes;
 		bufdesc.u.pBuffer = apb;
```
abhishek-samsung, line 540: These two lines can be moved under the if condition below them.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -497,53 +497,62 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
...
 #ifdef CONFIG_AUDIO_MULTI_SESSION
 		bufdesc.session = pcm->session;
 #endif
-		apb->nbytes = nbytes;
+		nbytes = pending > pcm->buffer_size - apb->nbytes ? pcm->buffer_size - apb->nbytes : pending;	
+		memcpy(apb->samp + apb->nbytes, (char *)data + offset, nbytes);
+		medvdbg("pending : %d nbytes : %d offset : %d apb : %x\n", pending, nbytes, offset, apb);
+		apb->nbytes += nbytes;
 		apb->curbyte = 0;
 		apb->flags = 0;
```
abhishek-samsung, line 534: These two look repeated, may be we can assign them when the apb is assigend for the first time? You can ignore this, not major.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -1163,16 +1172,27 @@ int pcm_drain(struct pcm *pcm)
...
 	if (pcm->flags & PCM_OUT) {
+		struct ap_buffer_s *apb;
 		struct audio_msg_s msg;
 		unsigned int size;
 		int prio;
-
+		/* If pending buffer exist, consume it */
+		if (pcm->next_buf != NULL) {
+			apb = pcm->next_buf;
+			if ((apb->flags & AUDIO_APB_OUTPUT_ENQUEUED) != AUDIO_APB_OUTPUT_ENQUEUED) {
```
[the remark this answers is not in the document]
Taejun-Kwon: yeah maybe but remained it as a double check. to prevent below while loop wait message forever.

framework/src/tinyalsa/tinyalsa.c
```diff
@@ -497,53 +497,61 @@ int pcm_writei(struct pcm *pcm, const void *data, unsigned int frame_count)
...
+		nbytes = pending > pcm->buffer_size - apb->nbytes ? pcm->buffer_size - apb->nbytes : pending;	
+		memcpy(apb->samp + apb->nbytes, (char *)data + offset, nbytes);
+		medvdbg("pending : %d nbytes : %d offset : %d apb : %x\n", pending, nbytes, offset, apb);
+		apb->nbytes += nbytes;
 		offset += nbytes;
 		pending -= nbytes;
+		
+		if (apb->nbytes == pcm->buffer_size) {
+			apb->curbyte = 0;
+			apb->flags = 0;
```
abhishek-samsung, line 538: I think these change should also be added to pcm_drain before sending last buffer.

abhishek-samsung (approved):
lgtm

jsdosa (approved):
LGTM.

PR #6380, opened 2024-09-04
board/ndp120 : Add new ioctl to control dmic

Taejun-Kwon:
Some product use external mic, in this case we should set down gpio pin
Usage
```

#include <tinyara/audio/audio.h>
#include <sys/ioctl.h>
#include <fcntl.h>

        ....
	int fd = open("/dev/audio/pcmC0D0c", O_RDONLY);
	if (fd < 0) {
		return -1;
	}
        /* For Internal DMIC */
	bool enable = false;
	int ret = ioctl(fd, AUDIOIOC_ENABLEDMIC, enable);
	if (ret < 0) {
		return -1;
	}
        /* For External DMIC */
	enable = true;
	ret = ioctl(fd, AUDIOIOC_ENABLEDMIC, enable);
	if (ret < 0) {
		return -1;
	}
```


jsdosa (approved):
LGTM.

PR #6381, opened 2024-09-04
audio/tinyalsa : Fix pcm_drain logic properly

Taejun-Kwon:
apb->flags changed in wrong place, reset it when dequeue message received.


jsdosa (approved):
looks nice to me.

PR #6382, opened 2024-09-05
Implement scan_all flag in scan_multi_config struct

jwei5:
- Implement scan_all flag in scan_multi_config struct to control whether multi scan API should return all results from probe request


framework/include/wifi_manager/wifi_manager.h
```diff
@@ -244,6 +244,7 @@ typedef struct {
 typedef struct {
 	wifi_manager_scan_config_s ap_configs[WIFIMGR_SPECIFIC_SCAN_AP_CNT];    /**< The AP configurations to scan */
 	unsigned int scan_ap_config_count;                                    /**< The count of AP to scan       */
+	char scan_all;											/**<  Flag to enable scanning specific AP + other APs responding to NULL probe req  */
```
hk-gwak, line 247: Is there any reason for this char data type for the flag? If not, how about boolean type?
```suggestion
	bool scan_all;											/**<  Flag to enable scanning specific AP + other APs responding to NULL probe req  */
```

os/include/tinyara/net/if/wifi.h
```diff
@@ -227,6 +227,7 @@ typedef struct {
 typedef struct {
 	trwifi_scan_config_s scan_ap_config[SCAN_SSID_CNT];
 	unsigned int scan_ap_config_count;                         /**<  the count of SSID to scan  */
+	char scan_all;											/**<  Flag to enable scanning specific AP + other APs responding to NULL probe req  */
```
hk-gwak, line 230: Same above.
```suggestion
	bool scan_all;											/**<  Flag to enable scanning specific AP + other APs responding to NULL probe req  */
```

framework/include/wifi_manager/wifi_manager.h
jwei5: Hi @hk-gwak , no specific reason for using char, ok will change it to boolean type

jwei5:
Changed scan_all to bool type

hk-gwak (approved):
Looks good to me.

PR #6392, opened 2024-09-11
 Add rotation support and add kernel thread flush lcd data

namanjain7:


os/drivers/lcd/mipi_lcd.c
```diff
@@ -349,5 +416,17 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
 
+#if defined(CONFIG_LCD_SW_ROTATION)
+	for (int itr = 0; itr < NUM_OF_LCD_BUFFER; itr++) {
+		lcd_buffer[itr] = (uint8_t *)kmm_malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1);
```
sunghan-chang, line 421: How about allocating entire size at once instead of allocating each node?
for example, if 1 frame size is 100,
we can allocate 200 at once, and internally we can assign the address on each array.

Why I suggest this is
1. each allocation needs a chunk at start of the node. Its size is 16 byte. We can reduce it.
2. Because this allocation is done at booting time, there is no adv in possibility of allocation. 
    Because this allocation size is big, if we do this at middle of app operation, we could get failure on huge size allocation. But this will be ok with 1 huge node at booting time.

os/drivers/lcd/lcd_dev.c
```diff
@@ -319,6 +331,34 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	return ret;
 }
 
+static int lcd_flushing_thread(void)
+{
+	int ret;
+	while(true) {
```
sunghan-chang, line 337: while (

os/drivers/lcd/mipi_lcd.c
```diff
@@ -38,6 +38,62 @@
...
+	int dst_inc = 2 * CONFIG_LCD_XRES;
+	short int val0;
+	short int val1;
+	short int val2;
+	short int val3;
+	short int *psrc;
+	short int *pdst;
+
+#if defined(CONFIG_LCD_LANDSCAPE)
+	lldbg("Line number: %d\n", __LINE__);
```
sunghan-chang, line 59: Is this necessary?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -38,6 +38,62 @@
...
+			val3 = *(psrc + CONFIG_LCD_YRES + 1);
+			psrc += 2;
+			*(pdst + 0) = val0;
+			*(pdst + 1) = val2;
+			*(pdst - CONFIG_LCD_XRES) = val1;
+			*(pdst - CONFIG_LCD_XRES + 1) = val3;
+			pdst -= dst_inc;
+		}
+	}
+#endif
```
sunghan-chang, line 93: Let's add `#else` with #error NOT Support blah blah orientation message.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -38,6 +38,62 @@
 #define CONFIG_LCD_MAXPOWER 100
 #endif
 
+#if defined(CONFIG_LCD_SW_ROTATION)
+#define NUM_OF_LCD_BUFFER	2
+uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER] = { NULL, NULL };	//Two lcd buffers to avoid screen tearing
+int lcd_buffer_index = 0;
```
sunghan-chang, line 44: Maybe they are used in this file only. Let's add `static`.

os/drivers/lcd/lcd_dev.c
```diff
@@ -319,6 +331,34 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	return ret;
 }
 
+static int lcd_flushing_thread(void)
```
sunghan-chang, line 334: I don't know this thread is necessary always. If not, we should introduce a new config and need to enable this for ST7701 only. @namanjain7 Please let me know your opinion on this.

build/configs/rtl8730e/loadable_ext_ddr/defconfig
```diff
@@ -587,6 +587,7 @@ CONFIG_LCD_MAXPOWER=100
 #
 # CONFIG_LCD_ST7701 is not set
 CONFIG_LCD_ST7701SN=y
+CONFIG_LCD_SW_ROTATION=y
```
sunghan-chang, line 590: I think you add this manually. When I check the Kconfig, this should be added at different line.
Could you do this using menuconfig?
Using menuconfig, you can modify .config which is in os folder
and then you can copy necessary line into the defconfig.

os/drivers/lcd/lcd_dev.c
namanjain7: The flushing will never decrease performance. FPS may increase or remain same across lcd.
In case the performance remain same with and without flush, then there is no point in keeping this code. It will be good to just disable it for those lcd's.
I think it would be good to add a option for thread.

build/configs/rtl8730e/loadable_ext_ddr/defconfig
namanjain7: fixed.

os/drivers/lcd/mipi_lcd.c
namanjain7: done

namanjain7:
Fixed all above mentioned comments. Pls review.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -349,5 +417,15 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
 
+#if defined(CONFIG_LCD_SW_ROTATION)
+	uint8_t *mem = (uint8_t *)kmm_malloc((CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1) * NUM_OF_LCD_BUFFER);	// each pixel is 8bit int
+	for (int i = 0; i < NUM_OF_LCD_BUFFER; i++) {
+		lcd_buffer[i] = mem + i * CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2;
+	}
+	if (!mem) {
```
anjana348, line 425: this check should be before lcd_buffer[i] = mem + i*...

os/drivers/lcd/mipi_lcd.c
sunghan-chang: How about using `#error` instead of debug message.
Why I suggested is
we can't support what user wants in this case. But debug message makes user know it after build, download and running. It's too late (User will feel bad). I think we can notice it to user at build time using `#error`

os/drivers/lcd/mipi_lcd.c
```diff
@@ -349,5 +417,17 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
 
+#if defined(CONFIG_LCD_SW_ROTATION)
+	uint8_t *mem = (uint8_t *)kmm_malloc((CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1) * NUM_OF_LCD_BUFFER);	// each pixel is 8bit int
+	if (!mem) {
+		lcddbg("ERROR: LCD rotate buffer memory allocation failed\n");
+		free(mem);
```
sunghan-chang, line 424: 1. This statement is failed to allocate so that no memory to free. This line should be removed.
2. Even we fail to allocate the memory, it returns `&priv->dev`. Is it ok? No need to return NULL?

os/drivers/lcd/lcd_dev.c
```diff
@@ -80,8 +81,16 @@ struct lcd_s {
 	struct lcd_planeinfo_s planeinfo[MAX_NO_PLANES];
 	sem_t sem;
 	int16_t crefs;
-};
 
+#if defined(CONFIG_LCD_FLUSH_THREAD)
+	uint8_t *lcd_kbuffer;
+	FAR const struct lcddev_area_s *g_lcd_area;
```
anjana348, line 87: g_lcd_area and g_wait should not have g_ as they are private variables

os/drivers/lcd/lcd_dev.c
```diff
@@ -340,30 +398,39 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
 		return -ENOMEM;
 	}
 
-	priv->dev = dev;
+	lcd_info->dev = dev;
+#if defined(CONFIG_LCD_FLUSH_THREAD)
+	sem_init(&lcd_info->flushing_sem, 0, 0);
+	lcd_info->g_wait = false;
+	lcd_info->empty = true;
+	lcd_info->lcd_kbuffer = (uint8_t *)kmm_malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * sizeof(uint16_t));
```
anjana348, line 418: if memory allocation fails here, return NULL, no need to continue

os/drivers/lcd/lcd_dev.c
```diff
@@ -319,6 +347,36 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+				continue;
+			}
+			return;
+		}
+		lcd_info->empty = false;
+		size_t cols = lcd_info->g_lcd_area->col_end - lcd_info->g_lcd_area->col_start + 1;
+		size_t pixel_size = lcd_info->planeinfo[lcd_info->g_lcd_area->planeno].bpp > 1 ? lcd_info->planeinfo[lcd_info->g_lcd_area->planeno].bpp >> 3 : 1;
+		size_t row_size = lcd_info->g_lcd_area->stride > 0 ? lcd_info->g_lcd_area->stride : cols * pixel_size;
+		if (lcd_info->planeinfo[lcd_info->g_lcd_area->planeno].putarea) {
+			ret = lcd_info->planeinfo[lcd_info->g_lcd_area->planeno].putarea(lcd_info->dev, lcd_info->g_lcd_area->row_start, lcd_info->g_lcd_area->row_end, lcd_info->g_lcd_area->col_start, lcd_info->g_lcd_area->col_end, lcd_info->lcd_kbuffer, row_size);
```
anjana348, line 367: break this into multiple lines

ewoodev (approved):
LGTM

sunghan-chang (approved):

os/drivers/lcd/mipi_lcd.c
```diff
@@ -349,5 +417,17 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
 
+#if defined(CONFIG_LCD_SW_ROTATION)
+	uint8_t *mem = (uint8_t *)kmm_malloc((CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1) * NUM_OF_LCD_BUFFER);	// each pixel is 8bit int
```
ZhenBei-Sin, line 421: Hi @sunghan-chang  @namanjain7 }

With the enable CONFIG_LCD_SW_ROTATION, for the loadable_ext_psram config is unable to bootup.

(CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2 + 1) * NUM_OF_LCD_BUFFER)
CONFIG_LCD_XRES=480
CONFIG_LCD_YRES=800
CONFIG_LCD_XRES=2
(480*800*2+1)*2 = 1,536,002

up_allocate_kheap: start = 0x60119600 size = 1468928

It should be no enough heap causing the issue.
Please help check on it.

Thank you

namanjain7: Hello Mr ZhenBei
We have tested the commit on loadable_ext_ddr. It's working in that.
I think loadable_ext_psram is for 5th revision board and it has less ram compared to next one. So, heap allocation is failing.
You can disable the rotation
CONFIG_LCD_SW_ROTATION=n
It won't crash then. But st7701sn won't display the screen correctly as LANDSCAPE is not supported in hardware side.

ZhenBei-Sin: Hi @namanjain7 
I understand your provided statements clearly because I tested and debugged them yesterday before I comment at here.

I'm clear on the config (CONFIG_LCD_SW_ROTATION is not set) can be disable to prevent the crashes for loadable_ext_psram config.

But in your commit, the default setting enabled CONFIG_LCD_SW_ROTATION=y for all configurations and was merged into the master branch.

When the issue occurred, it wasted our time as we had to review various configurations, settings, and code changes to pinpoint the cause of the problem.

Hi @sunghan-chang I hope you will take note of this.
This is public code, so please ensure it works for different configurations, including the basic boot-up.

PR #6394, opened 2024-09-11
log_dump: Add api to get size of the log dump

ritesh55555:
This commit adds an api to get the size of the compressed log dump.

### DIFF
--- a/apps/examples/log_dump/log_dump_main.c
+++ b/apps/examples/log_dump/log_dump_main.c
@@ -43,6 +43,7 @@ int log_dump_main(int argc, char *argv[])
 #endif
 {
 	int ret = 1;
+	int total_read = 0;
 	char buf[READ_BUFFER_SIZE];
 	int fd = OPEN_LOGDUMP();
 
@@ -78,10 +79,12 @@ int log_dump_main(int argc, char *argv[])
 
 	while (ret > 0) {
 		ret = READ_LOGDUMP(fd, buf, sizeof(buf));
+		total_read += ret;
 		for (int i = 0; i < ret; i++) {
 			printf("%c", buf[i]);
 		}
 	}
+	printf("log_dump_size = %d, and total_read = %d, Both should be equal\n", GET_LOGDUMP_SIZE(fd), total_read);
 
         printf("\n*********************   LOG DUMP END  ***********************\n");
 
--- a/os/fs/procfs/fs_procfslogsave.c
+++ b/os/fs/procfs/fs_procfslogsave.c
@@ -120,10 +120,6 @@ static int logsave_open(FAR struct file *filep, FAR const char *relpath, int ofl
 {
 	FAR struct logsave_file_s *attr = (struct logsave_file_s *)kmm_zalloc(sizeof(struct logsave_file_s));
 	filep->f_priv = (FAR void *)attr;
-	if (log_dump_read_wake() != OK) {
-		fdbg("ERROR: log dump read wake fail\n");
-		return ERROR;
-	}
 	fvdbg("Open '%s'\n", relpath);
 	return OK;
 }
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -100,6 +100,7 @@ static int compress_curbytes = 0;	/* number of bytes in the current chunk that w
 static int compress_rdptr = 0;		/* current pointer for compressed data read */
 static size_t log_dump_size = 0;
 static struct log_dump_chunk_s *read_node;
+static int log_dump_chunk_node_cnt;
 static unsigned char uncomp_buf[CONFIG_LOG_DUMP_NUMBUFS][CONFIG_LOG_DUMP_CHUNK_SIZE];	/* initial uncompressed log dump data */
 static bool uncomp_buf_full[CONFIG_LOG_DUMP_NUMBUFS];
 static int uncomp_idx = 0;
@@ -164,11 +165,13 @@ int log_dump_init(void)
 	comp_idx = 0;
 	compress_curbytes = 0;
 	compress_rdptr = 0;
+	log_dump_chunk_node_cnt = 0;
 	sq_addfirst((sq_entry_t *)node, &log_dump_chunks);
 	/* set the size of chunks to 1 chunk */
 	log_dump_size = LOG_CHUNK_SIZE;
 	/* Initialize to save the log by default */
 	is_started_to_save = true;
+	last_comp_block_size = 0;
 
 	return LOG_DUMP_OK;
 
@@ -231,6 +234,7 @@ int log_dump_compress_lastblock(void)
 	compress_ret = compress_block(&last_comp_block[4],&last_comp_block_size,  (unsigned char *)uncomp_buf[uncomp_idx], uncomp_curbytes);
 
 	if (compress_ret != LOG_DUMP_OK) {
+		last_comp_block_size = 0;
 		ldpdbg("Fail to compress compress_ret = %d\n", compress_ret);
 		return -compress_ret;
 	}
@@ -265,6 +269,7 @@ static void log_dump_mem_check(size_t max_size)
 			next_chunk = (struct log_dump_chunk_s *)sq_remfirst(&log_dump_chunks);
 			compress_rdptr = 0;	/* reset the read pointer as the head is removed */
 			log_dump_size -= LOG_CHUNK_SIZE;
+			log_dump_chunk_node_cnt -= 1;
 			kmm_free(next_chunk);
 		}
 	}
@@ -278,6 +283,14 @@ int log_dump_set(FAR const char *buffer, size_t buflen)
 		is_started_to_save = true;
 	} else if (strncmp(buffer, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) == 0) {
 		is_started_to_save = false;
+		/* Now we should compress the partially filled uncompressed buffer */
+		if (log_dump_read_wake() != OK) {
+			ldpdbg("ERROR: log dump read wake fail\n");
+			/* Should we return error from here?? because partially filled uncompressed buffer
+			 * might not be compressed but log_dump save has stopped now, so operation is success */
+		}
+	} else if (strncmp(buffer, LOGDUMP_GET_SIZE, strlen(LOGDUMP_GET_SIZE) + 1) == 0) {
+		return log_dump_get_size();
 	} else {
 		return LOG_DUMP_OPT_FAIL;
 	}
@@ -332,6 +345,7 @@ static int log_dump_tobuffer(char ch, size_t *free_size)
 				set_comp_head = false;
 			}
 			log_dump_size += LOG_CHUNK_SIZE;
+			log_dump_chunk_node_cnt += 1;
 
 		} else {
 			/* logs reached memory limit, reuse the head and free extra logs */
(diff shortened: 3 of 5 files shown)


sunghan-chang (commented):
Let's remove `log_dump.c` from the title. You don't make a change only in that file.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -43,6 +44,9 @@ int log_dump_main(int argc, char *argv[])
 #endif
 {
 	int ret = 1;
+#ifdef CONFIG_BUILD_FLAT
```
sunghan-chang, line 47: Could you let me know why this conditional is necessary?

os/kernel/log_dump/log_dump.c
```diff
@@ -505,6 +505,26 @@ int log_dump_save(char ch)
 	return LOG_DUMP_OK;
 }
 
+size_t log_dump_get_size(void)
+{
+	size_t size = 0;
+	struct log_dump_chunk_s* head = (struct log_dump_chunk_s *)sq_peek(&log_dump_chunks);
```
sunghan-chang, line 511: In code readability, head != tail? It looks strange.
How about renaming the variable like log_node?
It does not have the head always.

os/kernel/log_dump/log_dump.c
```diff
@@ -505,6 +505,26 @@ int log_dump_save(char ch)
...
+	while (head != (struct log_dump_chunk_s *)sq_tail(&log_dump_chunks)) {
+		head = (struct log_dump_chunk_s *)sq_next((sq_entry_t *)head);
+		size += CONFIG_LOG_DUMP_CHUNK_SIZE;
+	}
```
sunghan-chang, line 515: 1. Don't you need to do lock / unlock for log dump?
2. If tail should be fixed, I think we can get it once. Not every while loop.
3. How about calculating current size at every node allocation and free in some static global variable? We already need to get current size to adjust node total size based on free heap size.

os/kernel/log_dump/log_dump.c
```diff
@@ -505,6 +505,26 @@ int log_dump_save(char ch)
...
+	/* compress_curbytes is the length of compressed logs in latest log_dump_chunk */
+	size += compress_curbytes;
+
+	/* This is the size of compressed partially filled uncomp_buf.
+	 * Note: It is recommended to use this log_dump_get_size api after
+	 * log_dump_read_wake() is performed, so that it has correct value of last_comp_block_size.
```
sunghan-chang, line 521: App will use this API and App can't know whether `log_dump_read_wake` is performed or not. This recommendation looks redundant. Is there any other way to do it automatically?

os/kernel/log_dump/log_dump.c
```diff
@@ -505,6 +505,26 @@ int log_dump_save(char ch)
...
+	 * Else it might add old last_comp_block_size which is already present in log_dump_chunk list */
+	size += last_comp_block_size;
```
sunghan-chang, line 523: You return compressed log size + uncompressed log size. When we read saved log dump, don't we compress uncompressed log?? If we don't, then how can we know how many data should be uncompressed and how many data should be used as it is?

sunghan-chang:
@ritesh55555 One question, after getting the size using new API, is there any chance that node size is changed?
If it can be, app will do abnormal operation.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -78,10 +82,16 @@ int log_dump_main(int argc, char *argv[])
...
 		ret = READ_LOGDUMP(fd, buf, sizeof(buf));
+#ifdef CONFIG_BUILD_FLAT
+		total_read += ret;
+#endif
 		for (int i = 0; i < ret; i++) {
 			printf("%c", buf[i]);
 		}
 	}
+#ifdef CONFIG_BUILD_FLAT
+	printf("log_dump_size = %d, and total_read = %d, Both should be equal\n", log_dump_get_size(), total_read);
```
ewoodev, line 93: Please use procfs about log dump

ritesh55555:
> @ritesh55555 One question, after getting the size using new API, is there any chance that node size is changed? If it can be, app will do abnormal operation.

It is recommended that log dump size and log dump read should be done after stopping log dump save process. Else app might not get the value wanted . 
Apart from that if app tries to get log dump size during log dump process is happening, then it will get the size of compressed logs till that point.

os/kernel/log_dump/log_dump.c
ritesh55555: Earlier, when we were opening the procfs file of log dump, it was compressing the partially filled log buffer. Now the compressing the partially filled log buffer has been shifted to when log dump save is stopped . So that whenever we stop log dump process, we will compress the partially filled log buffer and update the size. 
two cases while getting log dump size:
1. When log dump save is true and log dump is saving character : In this case , log dump size will return the size of compressed logs till now in the log_dump_chuck and it will not include last_comp_block_size becuase it is not valid.
2. when log dump save is false and log dump is not saving : In this case , log dump size will include compressed logs in log_dump_chunk + last_comp_block_size becuase after log dump save is stopped we will compress the last partially filled logs.

apps/examples/log_dump/log_dump_main.c
kishore-sn: Please implement size api in procfs. Then this is not required.

os/kernel/log_dump/log_dump.c
```diff
@@ -505,6 +515,17 @@ int log_dump_save(char ch)
 	return LOG_DUMP_OK;
 }
 
+size_t log_dump_get_size(void)
+{
+	/* If log_dump save is stopped, that means partially uncompressed buffer
+	 * is compressed and size should include last_comp_block_size */
+	if (is_started_to_save == false) {
+		return log_dump_chunk_node_cnt * CONFIG_LOG_DUMP_CHUNK_SIZE + compress_curbytes + last_comp_block_size;
+	} else {
```
kishore-sn, line 524: The else case means, top_logsave has not been called. So, if we return some size here and there is some timegap before read, then the size will differ. It is better to return error code here. This api must be called only when log dump has been stopped.

ritesh55555: done

apps/examples/log_dump/log_dump_main.c
```diff
@@ -25,6 +25,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <tinyara/log_dump/log_dump.h>
+#include <tinyara/log_dump/log_dump_internal.h>
```
kishore-sn, line 28: We can remove this

kishore-sn (approved):

apps/examples/log_dump/log_dump_main.c
```diff
@@ -43,6 +43,7 @@ int log_dump_main(int argc, char *argv[])
 #endif
 {
 	int ret = 1;
+	int total_read = 0;
```
Tirtha-Samsung, line 46: This PR ```log_dump: Add api to get size of the log dump``` looks good to merge

kishore-sn:
@ritesh55555 Pls add the output of the log dump app in the commit description.

apps/examples/log_dump/log_dump_main.c
sunghan-chang, line 46: Sorry but, I want to keep separate commit for main change and app change. Could you split a commit for this app? It will be easy to sync later.

os/fs/procfs/fs_procfslogsave.c
```diff
@@ -120,10 +120,6 @@ static int logsave_open(FAR struct file *filep, FAR const char *relpath, int ofl
 {
 	FAR struct logsave_file_s *attr = (struct logsave_file_s *)kmm_zalloc(sizeof(struct logsave_file_s));
 	filep->f_priv = (FAR void *)attr;
-	if (log_dump_read_wake() != OK) {
```
sunghan-chang, line 123: Could you let me know why this is removed?

os/kernel/log_dump/log_dump.c
```diff
@@ -278,6 +283,14 @@ int log_dump_set(FAR const char *buffer, size_t buflen)
 		is_started_to_save = true;
 	} else if (strncmp(buffer, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) == 0) {
 		is_started_to_save = false;
+		/* Now we should compress the partially filled uncompressed buffer */
+		if (log_dump_read_wake() != OK) {
+			fdbg("ERROR: log dump read wake fail\n");
```
sunghan-chang, line 288: It should be ldpdbg, not fdbg.

os/kernel/log_dump/log_dump.c
```diff
@@ -278,6 +283,14 @@ int log_dump_set(FAR const char *buffer, size_t buflen)
 		is_started_to_save = true;
 	} else if (strncmp(buffer, LOGDUMP_SAVE_STOP, strlen(LOGDUMP_SAVE_STOP) + 1) == 0) {
 		is_started_to_save = false;
+		/* Now we should compress the partially filled uncompressed buffer */
```
sunghan-chang, line 286: If we do stop and start again without read, what happened?

os/fs/procfs/fs_procfslogsave.c
ritesh55555: This was called whenever we open procfs logsave file which was not needed. We are reading log dump after doing log dump stop . So calling log_dump_read_wake() when log dump stop is called is enough. 

log_dump_read_wake() will start compressing the partially filled log dump buffer . so i dont see why we should call it while opening procfs logsave file.

os/kernel/log_dump/log_dump.c
ritesh55555: Then the last comp block will be present with some old compressed log data and log dump will continue filling latest log dump. But if we try to get size of log dump , it will return error as is_started_to_save is true .
when log dump stop is called again, last comp block will be replaced by latest compressed partially filled buffer and return size will be valid.

Basically , do stop and start again, last_comp_block will have some old compressed log data untill stop is called again .During this period we are not supposed to get log dump size.

kishore-sn: @ritesh55555 
The question is whether the data remains valid or not?
If we call stop > start > stop > size > read > start .... will we get valid data on the call to size and read?
If we call stop > size > start > stop > size > read > start ... will we get valid data on the second call to size?

ritesh55555: yes , we will get valid data . I also tested the way you said it. It is workin fine

os/kernel/log_dump/log_dump.c
```diff
@@ -278,6 +283,14 @@ int log_dump_set(FAR const char *buffer, size_t buflen)
...
+		if (log_dump_read_wake() != OK) {
+			ldpdbg("ERROR: log dump read wake fail\n");
+			/* Should we return error from here?? because partially filled uncompressed buffer
+			 * might not be compressed but log_dump save has stopped now, so operation is success */
+		}
+	} else if (strncmp(buffer, LOGDUMP_GET_SIZE, strlen(LOGDUMP_GET_SIZE) + 1) == 0) {
```
sunghan-chang, line 292: There are 3 strncmp s to check a command. It looks not good.
Because the data is a just character, how about using switch / case?

os/kernel/log_dump/log_dump.c
sunghan-chang: Here you added the else case to check stop first. How about adding it at the read function as well? Because it's the same.

sunghan-chang:
@ritesh55555 I left two comments, but I think it's different topic against this PR so that you can do that with new PR or new commit.

kishore-sn (approved):

PR #6402, opened 2024-09-17
build/tools/amebasmart/gnu_utility: fix stuck tash when key held on boot

lhenry-realtek:
This fix does the following:

After KM0 core start, IER bit set to 0 to disable LOGUART
After CA32 ipc_send_message in board_initialize(), set IER bit to 1 in KM0 IPC command handler (ROM LOGUART is disabled here)
CA32 can use tizen RAM handler now, and tash shell is working properly

Root cause:

- Due to multicore environment, LOGUART control is passed from KM4->KM0->CA32 on boot, where TASH shell is executing on AIDual. 
- Control is passed by disabling interrupt handling for LOGUART on each core, until only 1 core is using it
- When TizenRT is starting up, in board_initialize there is a short period of time before CA32 sends IPC to disable KM0's LOGUART interrupt handler. In this short period it is possible for CA32 LOGUART handler and KM0 LOGUART handler to fire.
- Due to handling of the Rx interrupt event enable (ERBFI) bit in each handler, it is possible for KM0 side to read and set an invalid value of the enable bit as CA32 side is also accessing it
- When the ERBFI bit is 0 after KM0 finishes IRQ and then KM0 LOGUART is disabled, CA32 now has access to the LOGUART peripheral but no events will fire, as the bit is 0, thus tash shell also hangs


sunghan-chang:
> Please check rtk jira SDAIOT30E-291 for more info, there is another proposed fix for same issue

Did you leave this for RTK engineers?

ewoodev (commented):
Could you please explain rootcause? and now issue state?

sunghan-chang:
@lhenry-realtek Could you leave what you said in the commit description?
@jylee9613 Could you verify this?

jylee9613 (approved):
LGTM

ewoodev (approved):

PR #6403, opened 2024-09-17
Moves RO and TEXT to RAM in XIP ELF configuration

agnel-samsung:
Updates .ld script and xipelf loader to move text and rodata sections from flash to RAM


abhishek-samsung (commented):
We should be able to use XIP elf normally like before. You will need to add a config, if that config is eanbled, then we need to execute everthing from RAM. So, we should not change XIP elf directly.

PR #6407, opened 2024-09-23
os/board/rtl8730e: Fix module sleep while UART buffer Tx buffer not empty

ziyik:
- Do not suspend and resume UART domain in IRQ, it only check FIFO not Tx buffer
- Suspend UART domain PM sleep while Tx init is set
- Correct the clock set when wake up

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_serial.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_serial.c
@@ -210,6 +210,9 @@
  */
 
 static serial_t* sdrv[MAX_UART_INDEX + 1] = {NULL, NULL, NULL, NULL, NULL}; //uart 0~4, uart4 is configured as log uart
+#ifdef CONFIG_PM
+static bool uart_active_state = 0;
+#endif
 
 struct rtl8730e_up_dev_s {
 	uint8_t parity;				/* 0=none, 1=odd, 2=even */
@@ -893,9 +896,6 @@ void rtl8730e_uart_irq(uint32_t id, SerialIrq event)
 	struct uart_dev_s *dev = (struct uart_dev_s *)id;
 	struct rtl8730e_up_dev_s *priv = (struct rtl8730e_up_dev_s *)dev->priv;
 
-#ifdef CONFIG_PM
-	bsp_pm_domain_control(BSP_UART_DRV, 1);
-#endif
 	if (event == RxIrq) {
 		uart_recvchars(dev);
 	}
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1103,6 +1097,9 @@ static void rtl8730e_up_txint(struct uart_dev_s *dev, bool enable)
 	struct rtl8730e_up_dev_s *priv = (struct rtl8730e_up_dev_s *)dev->priv;
 	DEBUGASSERT(priv);
 	priv->txint_enable = enable;
+#ifdef CONFIG_PM
+	bsp_pm_domain_control(BSP_UART_DRV, enable);
```
ewoodev, line 1101: Please check `bsp_pm_domain_control: Unable to resume bsp_drv_id: 4!` logs

pm_suspend and pm_resume should be called pair

ziyik: bsp_pm_domain_control(BSP_UART_DRV, enable); will be set to 1 and 0.
"enable" is set based on UART Tx request, where there is still data in the ring buffer, this txinit bit will be set to 1.
In uart_write(), uart_disabletxint() will disable UART Tx init due to there being no data in the buffer.

Thus, bsp_pm_domain_control(BSP_UART_DRV, enable); is called in pair.

ewoodev: now, you can get `bsp_pm_domain_control: Unable to resume bsp_drv_id: 4!` error.

Could you please check again in your test?
Please check the errno number. if it is ERANGE, you are not matching pair of pm_suspend /pm_resume

edwakuwaku: > bsp_pm_domain_control(BSP_UART_DRV, enable); will be set to 1 and 0. "enable" is set based on UART Tx request, where there is still data in the ring buffer, this txinit bit will be set to 1. In uart_write(), uart_disabletxint() will disable UART Tx init due to there being no data in the buffer.
> 
> Thus, bsp_pm_domain_control(BSP_UART_DRV, enable); is called in pair.

It is possible that uart_disabletxint() is invoked multiple times even the suspended domain has been resumed. Then, you might encounter the error mentioned by eunwoo.
Please check again the flow in os/drivers/serial/serialirq.c, serial.c. The BSP code is invoked by the TizenLite driver, and there is no guarantee that layer does that operation in pair, since both layer are maintained independently.

ewoodev: you can use `uint16_t pm_suspendcount(int domain_id)`

ziyik: I have updated the PR, from my side I couldn't reproduce the issue.
I have changed the way of calling bsp_pm_domain_control().

ewoodev (approved):

sunghan-chang (approved):

PR #6408, opened 2024-09-24
os/arch/arm/src/amebasmart: low level api for USB logging

zhongnuo-tang:
1. provide usb_up_lowputc to route low level debugging such as lldbg etc..

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_usb.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_usb.c
@@ -139,6 +139,7 @@ struct amebasmart_usbdev_s {
 	u8 remote_wakeup_en : 1;		/* Remote wakeup enable or not, 0-disabled, 1-enabled */
 	u8 remote_wakeup : 1;			/* Remote wakeup */
 	sem_t txsem;
+	sem_t exclsem;
 };
 static struct amebasmart_usbdev_s g_usbdev;
 /* Serial driver UART operations */
@@ -333,6 +334,7 @@ static void usb_xmitchars(FAR uart_dev_t *dev)
 {
 	struct rtl8730e_up_dev_s *priv = (struct rtl8730e_up_dev_s *)dev->priv;
 	DEBUGASSERT(priv);
+	int ret;
 	uint16_t nbytes = 0;
 	uint8_t *buffer = NULL;
 	buffer = (uint8_t *)rtw_zmalloc(CONFIG_CDC_ACM_BULK_IN_XFER_SIZE);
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/amebasmart/amebasmart_usb.c
```diff
@@ -800,4 +816,53 @@ void usb_initialize(void)
...
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+void usb_up_lowputc(uint8_t *ch, uint16_t len)
+{
+	uint16_t remain_data_len = len;
+	uint16_t send_data_len = 0;
+	uint8_t *buffer = NULL;
+	buffer = (uint8_t *)rtw_zmalloc(CONFIG_CDC_ACM_BULK_IN_XFER_SIZE);
```
ewoodev, line 837: we can't use alloc during log level log output

os/arch/arm/src/amebasmart/amebasmart_usb.c
```diff
@@ -800,4 +816,53 @@ void usb_initialize(void)
...
+	while (remain_data_len > 0) {
+		if (remain_data_len >= CONFIG_CDC_ACM_BULK_IN_XFER_SIZE) {
+			send_data_len = CONFIG_CDC_ACM_BULK_IN_XFER_SIZE;
+		} else {
+			send_data_len = remain_data_len;
+		}
+		memcpy(buffer, ch, send_data_len);
+		int ret;
+		do {
+			ret = sem_wait(&g_usbdev.exclsem);
```
ewoodev, line 851: we can't use sem wait during low level log outputting,
because it is used in ISR and fault handler

zhongnuo-tang: Hi @ewoodev,
Do you have any suggestion for this? We need to ensure that there will only be one place calling usbd_cdc_acm_transmit

os/arch/arm/src/amebasmart/amebasmart_usb.c
zhongnuo-tang: I will change it to static buffer instead

os/arch/arm/src/amebasmart/amebasmart_usb.c
ewoodev: We need more lower layer api. 
I think, we can't use usbd_cdc_acm_transmit.

Could you please create new api which is lower than usbd_cdc_acm_transmit?
we can't use sem, so, we need to use busy wait in most low usb cdc acm. and it should not wait to complate of normall usb transmit as usbd_cdc_acm_transmit
is it possible?

sunghan-chang:
@zhongnuo-tang This is necessary on lldbg and it's used when os is not working. So, we CAN NOT use any os function like sem, malloc and etc. Please refer up_lowputc. It should only access hw register directly.

zhongnuo-tang:
> @zhongnuo-tang This is necessary on lldbg and it's used when os is not working. So, we CAN NOT use any os function like sem, malloc and etc. Please refer up_lowputc. It should only access hw register directly.

Hi @sunghan-chang,
Understood, let me check on this.

sunghan-chang:
@zhongnuo-tang I heard this is impossible now. Do we need to keep this PR open? If not, let's close.

PR #6409, opened 2024-09-24
os/arch/arm/amebasamart: optimize I2S tx

zhongnuo-tang:
1. fill up all the dma pages during the first tx, this is to prevent if there are any delays/thread blocking, CPU might not have sufficient time to update the next page in IRQ handler, which will lead to noise
2. fill up all available dma pages in IRQ handler, to handle cases when there is any interrupt missed previously
3. increase page size from 2K to 4K, allow more time for CPU to update the next available page before dma trigger tx done irq. Tested and verified with Ai-lite + on-boading + bixby application original flow with 4K page size -> noise can be heard within 5mins with this commit -> no noise can be heard for more than 40mins

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -119,7 +119,7 @@
 
 #define OVER_SAMPLE_RATE (384U)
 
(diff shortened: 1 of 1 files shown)


jsdosa (approved):
Looks nice to me.

PR #6412, opened 2024-09-24
fs/driver/mtd/mtd_jedec: Add missing jedec_lock

ewoodev:
In the function "jedec_waitwritecomplete", it is assumed that "jedec_lock" has already been called, and it unlocks and locks again. However, the function "jedec_bytewrite" (which is called by "jedec_write") does not call "jedec_lock" first.

The jedec_lock contains a spi_lock, which prevents simultaneous access to spi. So, when the jedec_lock is used out of pairs, it simultaneously accesses the spi and causes a spi malfunction resulting in a device hang.

Therefore, To ensure atomic operation of the spi, Add a missing jedec_lock.


Taejun-Kwon (approved):

Taejun-Kwon:
welcome to fs world!

PR #6415, opened 2024-09-25
framework/ble_manager: Deinitialize global scan context before call state change callback function

giwon-nam:
After calling the callback function and setting the value of the global context to 0, a problem occurred when the ble scan started inside the callback function resulting in deinitialization. Therefore, I changed it so that the global context deinit is performed first and then the callback function is called.


hs36-kim (approved):
LGTM
confirmed by rtk.
--------- Original Message ---------
Sender : eric.kim@semitron.co.kr <eric.kim@semitron.co.kr>
Date : 2024-10-07 17:34 (GMT+9)
Title : RE: (2) (2) (2) ble scan state changed callback 호출 순서 변경
 
안녕하세요.
 Review comment는 아래와 같습니다. 특별한 이슈는 없을 듯합니다.
It seems that the author wants to assign new values to g_scan_ctx attributes in the callback function. With the current code, the newly assigned value will get cleared after being assigned in the callback function.
LGTM
 
감사합니다.

PR #6416, opened 2024-09-27
build/configs/rtl8730e/loadable_ext_psram/defconfig: Disable screen r…

namanjain7:
…otation and enable portrait mode for 5th revision and lower board

For 5th revision and less board, it's not possible to rotate screen due to low memory, This commit disable the screen rotate and enable the portrait mode


ewoodev (commented):
Please disable LCD configuration in this defconfig

build/configs/rtl8730e/loadable_ext_psram/defconfig
```diff
@@ -586,8 +586,9 @@ CONFIG_LCD_ST7701SN=y
 # CONFIG_LCD_ST7789 is not set
 CONFIG_LCD_XRES=480
 CONFIG_LCD_YRES=800
-CONFIG_LCD_SW_ROTATION=y
-CONFIG_LCD_LANDSCAPE=y
+CONFIG_LCD_SW_ROTATION=n
```
sunghan-chang, line 589: There is no syntax for `=n`. It means you modified manually which you should not.
Please find the first line
```
#
# Automatically generated file; DO NOT EDIT.
# TinyAra Configuration
#
```
When you want to modify some defconfig, you should modify .config first through menuconfig.
And then, copy changes to defconfig.

And as @ewoodev said, we should disable lcd here, not only for rotation...

PR #6420, opened 2024-09-27
os/kernel/pthread: Fix issue during pthread_create with affinity

kishore-sn:
If pthread_create is called with non-zero value for affinity, then we need to check if SMP is enabled. We need to take the affinity value only if SMP enabled. However, if SMP is disabled, the pthread_create API must not fail, so that the app layer can still use the pthread API irrespective of SMP config in kernel. So, instead of returning an error code at this point, we just print an error message and continue with the pthread_create API.


r-prabu (dismissed):
Changes to fix pthread_create with affinity 0 when smp is disabled looks good.

os/kernel/pthread/pthread_create.c
```diff
@@ -281,7 +281,6 @@ int pthread_create(FAR pthread_t *thread, FAR const pthread_attr_t *attr, pthrea
 #ifndef CONFIG_SMP
 	if (attr->affinity != 0) {
 		sdbg("ERROR: Non zero affinity value not allowed when CONFIG_SMP is disabled\n");
```
sunghan-chang, line 283: Let's change to warning. When SMP is disabled, setting affinity does not cause any issue. Someone can think this is a problem with this error log.

ritesh55555 (approved):
not causing error for non zero affinity in case of single core, looks good

sunghan-chang (approved):

r-prabu (approved):

PR #6422, opened 2024-10-02
build/tools/amebasmart/image_tool: Fix CRC checksum error for interna…

zhongnuo-tang:
…l flash download

- fix crc checksum error corner case when using Windows Imagetool to download to internal flash, only applicable to Windows imagetool.


sunghan-chang:
@zhongnuo-tang What about Linux?

zhongnuo-tang:
> @zhongnuo-tang What about Linux?

Dear @sunghan-chang,
The corner case only happen if we use USB download with two images in a particular order using Windows imagetool. For Linux it does not happen because image are download one by one.

sunghan-chang (approved):

sunghan-chang:
@jylee9613 could you confirm this?

jylee9613 (approved):

PR #6425, opened 2024-10-04
audio/syu645scripts.h: remove unnecessary setting of default values

ritesh55555:
This commit removes the code to set the register values of amplifier after reset. Some values were getting set to already set default values again after reset which was unnecessary. Also the commit increases the delay for reset.

### DIFF
--- a/os/drivers/audio/syu645bscripts.h
+++ b/os/drivers/audio/syu645bscripts.h
@@ -45,7 +45,7 @@ typedef struct {
 
 t_codec_init_script_entry codec_reset_script[] = {
 	/* 0x0F for soft reset */
-	{ 0x0F, {0x01,}, 0, SYU645B_REG_D_2BYTE}
+	{ 0x0F, {0x01,}, 30, SYU645B_REG_D_2BYTE}
 };
 t_codec_init_script_entry codec_stop_script[] = {
 	/* set volume 0x0007 to zero on stop */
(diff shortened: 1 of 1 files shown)


os/drivers/audio/syu645bscripts.h
```diff
@@ -45,7 +45,7 @@ typedef struct {
 
 t_codec_init_script_entry codec_reset_script[] = {
 	/* 0x0F for soft reset */
-	{ 0x0F, {0x01,}, 0, SYU645B_REG_D_2BYTE}
+	{ 0x0F, {0x01,}, 30, SYU645B_REG_D_2BYTE}
```
r-prabu, line 48: what is this 30? Is this related to delay?

ritesh55555: yes , it adds a delay of that millisecond

Taejun-Kwon:
I wIll test this and merge.

samsung-singh (approved):
default values from settings have been removed and modified, this change is good to merge.

neel-samsung (commented):
PR looks ready to get merge

os/drivers/audio/syu645bscripts.h
abhishek-samsung, line 48: I was uanble to find where they mentioned the delay after reset as 30ms. It looks like we can read the register until the bit is set to 0 once reset is done. Please check it later . It might be much faster than 30ms.

Taejun-Kwon:
Currently pause-resume is not added because of remained issue in DMA, but it will be added soon.
Currently driver execute stop script when audio_release requested.(unprepare)
During prepare, it should execute some script to start coded again.
Some value in syu645b_dev_s structure need to be handled when execute start script again...

Taejun-Kwon:
Any update? Please refer to this.
https://github.com/Samsung/TizenRT/pull/6464#discussion_r1815966797

os/drivers/audio/syu645bscripts.h
Taejun-Kwon: which page it mentioned?

os/drivers/audio/syu645bscripts.h
```diff
@@ -54,34 +54,25 @@ t_codec_init_script_entry codec_stop_script[] = {
 
 t_codec_init_script_entry codec_initial_script[] = {
 	/* Initialize AMP sequence */
-	{ 0x0F, {0x01,}, 0, SYU645B_REG_D_2BYTE},
-	{ 0x00, {0x1A,}, 0, SYU645B_REG_D_2BYTE},
+	{ 0x0F, {0x01,}, 30, SYU645B_REG_D_2BYTE},
```
gSahitya-samsung, line 57: Please check merge conflicts and rebase

PR #6426, opened 2024-10-04
Add test APIs for clock switching NDP120 PDM <-> I2S BCLK

cvejlbo:
This Draft PR adds test code too allow switching between internally and externally generated PDM clock.

This test code in `configure_audio()` called during init sets up for internal clock and then adds intermediate test apis for dynamic use to either:

1.  switch to external clock (or back again) - for use when buffer is in use
or
2.  switch to using PCLK1 input as passthrough (or back again) - for the case where I2S BCLK is connected to PCLK1 input

The assumption in this code is that at NDP initialization time, the buffer on i2s bclk is set to disabled (OE=0) so NDP can take control of the PDM clock. 

To switch the clock in the two different cases:

1. with "buffer approach":  Generally, this needs to be done in "break before make" fashion
- Switching from internal -> external
I) Switch NDP to external using `ndp120_test_internal_external_switch(dev, 0)` API, this will stop the PDM clock from NDP without a hard reconfig of the rest of the PDM interface. 
II) Set buffer OE=1 to allow i2s clk to DMIC

- Switching from external -> internal
I) Set buffer to OE=0, to enable the tri-state of i2s clk to DMIC
II) Switch NDP to internal using `ndp120_test_internal_external_switch(dev, 1)` API, this will start the PDM clock from NDP

2. with "PDM passthrough" approach: (assumes buffer OE=0)
-Switching from internal -> external: Call `ndp120_test_internal_passthrough_switch(dev, 0)` API
-Switching from external -> internal: Call `ndp120_test_internal_passthrough_switch(dev, 1)` API

### DIFF
--- a/os/drivers/ai-soc/ndp120/src/ndp120_api.c
+++ b/os/drivers/ai-soc/ndp120/src/ndp120_api.c
@@ -77,9 +77,7 @@
 /* currently the max size of tank is limited to ~580ms  */
 #define AUDIO_BEFORE_MATCH_MS	(1500)
 
-/* when defined, the let the NDP use an external PDM clock */
-#define USE_EXTERNAL_PDM_CLOCK
-#define EXTERNAL_PDM_CLOCK_PDM_RATE 1536000
+#define PDM_CLOCK_PDM_RATE 1536000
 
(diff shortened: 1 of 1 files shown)


Taejun-Kwon:
Seems that work fine (pclk). we will test it more.

PR #6432, opened 2024-10-08
framework/ble_manager: Rename api for setting gap_device_name

giwon-nam:
- Since the `gap device name` is not used only in ble server, I modified it to define an API in ble_manager.h.
- Additionally, I defined the length of the `gap device name`, which is being used identically on each board, so that a string with corresponding length can be inputted.


hs36-kim (approved):
LGTM

PR #6435, opened 2024-10-08
ai-soc/ndp120: fix issue with missing data after keyword

abhishek-samsung:
If recording start is delayed after keyword extraction, then some audio data is lost. Fix this by moving the unread pointer to the match point so we start extraction right after the key word in the buffer.


r-prabu (approved):

Taejun-Kwon (approved):
Great

PR #6438, opened 2024-10-09
os/arch/arm/src/armv7-a: Modify assert handler for SMP case

kishore-sn:
If an assert happens when SMP is handled, we need to do the following:
- Enter critical section
- If crash is in kernel, then stop all other cores
- Perform assert handling
- Leave critical section

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -129,8 +129,13 @@ extern int g_irq_num;
 #define CONFIG_BOARD_RESET_ON_ASSERT 0
 #endif
 
+#ifdef CONFIG_APP_BINARY_SEPARATION
 #define IS_FAULT_IN_USER_THREAD(fault_tcb)		((void *)fault_tcb->uheap != NULL)
 #define IS_FAULT_IN_USER_SPACE(asserted_location)	(is_kernel_space((void *)asserted_location) == false)
+#else
+#define IS_FAULT_IN_USER_THREAD(fault_tcb)		(false)
+#define IS_FAULT_IN_USER_SPACE(asserted_location)	(false)
+#endif
 
 #if CONFIG_TASK_NAME_SIZE > 0
 #define LOG_TASK_NAME  fault_tcb->name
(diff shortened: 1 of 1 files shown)


samsung-singh (approved):
assert handler is modified in the case of SMP, this change is good to merge.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -582,10 +567,36 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = (uint32_t)user_assert_location;
 		user_assert_location = 0x0;
 	} else {
-		asserted_location = (uint32_t)kernel_assert_location;
+		ARCH_GET_RET_ADDRESS(asserted_location)
```
sunghan-chang, line 570: In this place, is this working? As far as I know, if we can this at middle of a function, it does not give a proper address. If we want to move this here, then I think we should use __builtin_..... instead of ARCH_GET....

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -582,10 +567,36 @@ void up_assert(const uint8_t *filename, int lineno)
...
+	}
+
+#ifdef CONFIG_SMP
```
sunghan-chang, line 573: How about leaving some comment to explain below?

sunghan-chang:
@kishore-sn There is a build break with flat_apps.
```
make[2]: Leaving directory '/*******/os/board/rtl8730e/src'
LD: tinyara
/*******/os/../build/output/libraries/libarch.a(arm_assert.o): In function `up_assert':
/*******/os/arch/arm/src/armv7-a/arm_assert.c:574: undefined reference to `is_kernel_space'
Makefile:207: recipe for target '../build/output/bin/tinyara.axf' failed
make[1]: *** [../build/output/bin/tinyara.axf] Error 1
```

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -582,10 +567,36 @@ void up_assert(const uint8_t *filename, int lineno)
...
+	if (!IS_FAULT_IN_USER_SPACE(asserted_location)) {
```
ritesh55555, line 574: there is a build break in flat because this is only for loadable config, please check

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -582,10 +567,36 @@ void up_assert(const uint8_t *filename, int lineno)
...
+		int me = sched_getcpu();
+		for (int i = 0; i < CONFIG_SMP_NCPUS; i++) {
```
edwakuwaku, line 576: let's use meaningful variable, maybe `int cpu`

os/arch/arm/src/armv7-a/arm_assert.c
kishore-sn: Right.. i reverted it. Thank you.

os/arch/arm/src/armv7-a/arm_assert.c
kishore-sn: Done. Thanks.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -553,25 +558,13 @@ static inline void print_assert_detail(const uint8_t *filename, int lineno, stru
 
 void up_assert(const uint8_t *filename, int lineno)
 {
-	/* ARCH_GET_RET_ADDRESS should always be
-	 * called at the start of the function */
```
sunghan-chang, line 557: Let's leave this.

sunghan-chang: @r-prabu After this, let's merge this PR without my approval.

PR #6441, opened 2024-10-11
os/arch/arm/src/amebasmart: allow booting normally when USB not attached

zhongnuo-tang:
1. allow it to boot normally even with no USB attached, use task instead of busy wait for usb init.
2. once USB attached, logs and tash can be input from usb port.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_usb.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_usb.c
@@ -83,6 +83,7 @@ static _sema cdc_acm_attach_status_changed_sema;
 extern int cdc_acm_ready_flag;
 static uint8_t receive_data;
 static int received_flag = 0;
+static int attach_status = -1;
 /****************************************************************************
  * Private Type Definitions
  ****************************************************************************/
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/amebasmart/amebasmart_usb.c
```diff
@@ -761,29 +761,35 @@ int amebasmart_up_usbinitialize(struct amebasmart_usbdev_s *priv)
...
 {
 	while(!cdc_acm_ready_flag){
-		usleep(1000);
+		usleep(100);
 	}
 	USB_DEV.isconsole = true;
 	uart_register("/dev/ttyACM0", &USB_DEV);
 	/*unregister /dev/console because serial was sharing the same fd during serial initialization*/
 	unregister_driver("/dev/console");
+	tash_stop();
```
sunghan-chang, line 776: You should not call the apl API at kernel side. And USB BSP should not controll the TASH. Please find another way.

zhongnuo-tang: Dear @sunghan-chang,

In my understanding, during the initalization of the TASH task, it opened the path /dev/console, and this fd will be used in a do while loop until TASH stopped. So during boot up stage, if USBD is not ready, /dev/console will be registered to loguart and logs and input will be direct to loguart in TASH task. Even if i re-register it to USB (when USB device is connected,) the logs will not redirect to usb unless i can close the fd and re-open it. 

Is it possible to change tash_main.c file to add flag check for usb device?

sunghan-chang: Hello @zhongnuo-tang .
Regardless of USB connection, when we enable USB console, then logs should be out through USB only. We should not redirect it to UART. We enable USB console due to performance issue with UART. If we redirect logs to UART according to USB connection, we don't need to enable USB console. (Almost products has no USB connection)

Let's make it as UART does. Even UART is not connected, it does not block Tizen Lite booting. They should be independent. ("They" are booting and UART connection in physical)

zhongnuo-tang: Dear @sunghan-chang,
Understood, can i assume that all the logs during USB disconnection period can be ignored/loss? Initially the busy-wait was to ensure that USB logging should only start working when USB device is connected and ready.

sunghan-chang: In UART, it's loss, even uart sends logs but there is no receiver.
Anything is ok. BUT we have the log dump module, it saves logs regardless of UART or USB connection. It should be working.

sunghan-chang (approved):

PR #6448, opened 2024-10-15
rtl8730e/i2s : Rollback DMA size to 16K(temp code)

aadotverma:
If we use 4K of dma but 16k of audio buffer, then pause->start doesn't work properly Until Realtek fix this issue, keep 16k * 2 of buffer for both of dma & audio


Taejun-Kwon:
It already merged in other PR, close it. thanks!

PR #6454, opened 2024-10-18
rtl8730e_boot.c: Add led pin to gpio driver

ritesh55555:
This commit adds led pin (PB_20) to gpio driver.

### DIFF
--- a/apps/examples/led_test/Kconfig
+++ b/apps/examples/led_test/Kconfig
@@ -1,0 +1,22 @@
+#
+#For a description of the syntax of this configuration file,
+#see kconfig-language at https:	//www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
+#
+
+config EXAMPLES_LED
+	bool "LED example"
+	default n
+	---help---
+		Enable LED Example
+
+if EXAMPLES_LED
+config EXAMPLES_LED_PIN
+	int "LED pin number"
+	default 52
+	---help---
+		Defines LED pin number				
+endif 
+
+config USER_ENTRYPOINT
+	string
+	default "led_test" if ENTRY_LED
--- a/apps/examples/led_test/Makefile
+++ b/apps/examples/led_test/Makefile
@@ -1,0 +1,162 @@
+###########################################################################
+#
+# Copyright 2024 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+############################################################################
+# apps/examples/led_test/Makefile
+#
+#   Copyright (C) 2008, 2010-2013 Gregory Nutt. All rights reserved.
+#   Author: Gregory Nutt <gnutt@nuttx.org>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in
+#    the documentation and/or other materials provided with the
+#    distribution.
+# 3. Neither the name NuttX nor the names of its contributors may be
+#    used to endorse or promote products derived from this software
+#    without specific prior written permission.
+#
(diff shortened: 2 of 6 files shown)


os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -240,6 +240,7 @@ void board_gpio_initialize(void)
 		u32 pinpull;
 	} pins[] = {
 				{PA_23, PIN_INPUT, PullNone},
+				{PB_20, PIN_OUTPUT, PullNone},
```
sunghan-chang, line 243: Let's set it as pd.

apps/examples/led_test/Makefile
```diff
@@ -0,0 +1,162 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+#
+###########################################################################
+############################################################################
+# apps/examples/led_test/Makefile
+#
+#   Copyright (C) 2008, 2010-2013 Gregory Nutt. All rights reserved.
+#   Author: Gregory Nutt <gnutt@nuttx.org>
```
abhishek-samsung, line 22: Is this required? I think the files in the PR are written by us entirely, so I am not sure if this is requried.

sunghan-chang: No need it in c file, let's remove.
But I either am not sure for makefile. Because format is almost the same.

sunghan-chang:
@ritesh55555 For first commit, `rtl8730e_boot.c: Add led pin to gpio driver`, let's use `board/rtl8730e:` instead of `rtl8730e_boot.c:`.

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -240,6 +240,7 @@ void board_gpio_initialize(void)
 		u32 pinpull;
 	} pins[] = {
 				{PA_23, PIN_INPUT, PullNone},
```
kishore-sn, line 242: Add a comment to indicate what purpose this pin is used for.

apps/examples/led_test/example_led.c
```diff
@@ -0,0 +1,95 @@
...
+#else
+int led_test_main(int argc, char *argv[])
+#endif
+{
+	printf("=== LED demo ===\n");
+	int fd = 0;
+	fd = open(LED_DEV_PATH, O_RDWR | O_SYNC, 0666);
+	if (fd < 0) {
+		printf("ERROR: Failed to open led gpio driver : %s error:%d\n", LED_DEV_PATH, fd);
+		return;
```
kishore-sn, line 72: Change to "return 1" or "return -1" to avoid compiler warning as this API is supposed to return an int value.

apps/examples/led_test/example_led.c
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	printf("LED should light up(for 1 second) for 5 times at an interval of 1 second\n");
```
kishore-sn, line 75: Change as below:
`Check LED operation on device. LED will blink 5 times with ON-OFF period of 1 sec.`

apps/examples/led_test/example_led.c
```diff
@@ -0,0 +1,95 @@
...
+ *
+ ****************************************************************************/
+
+#include <tinyara/config.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define LED_DEV_PATH "/dev/gpio52"
```
kishore-sn, line 56: How do we know that gpio52 is supposed to be used for this LED?
Please leave a comment. 
Also, how do with know that GPIO pins 20 and 23 map to this port? Pls leave comment in driver side also.

sunghan-chang:
@ritesh55555 Could you resolve all comments today? We should give this tomorrow.

os/board/rtl8730e/src/rtl8730e_boot.c
ritesh55555: it is already defined as PB by realtek in bsp code

ritesh55555:
> @ritesh55555 Could you resolve all comments today? We should give this tomorrow.

updated all comments, please review

os/board/rtl8730e/src/rtl8730e_boot.c
sunghan-chang: Oh, I mean `pulldown`, not `PD_20`.

apps/examples/led_test/Kconfig
```diff
@@ -0,0 +1,27 @@
...
+
+config EXAMPLES_LED
+	bool "LED example"
+	default n
+	---help---
+		Enable LED Example
+
+if EXAMPLES_LED
+
+if ARCH_BOARD_RTL8730E
```
sunghan-chang, line 14: Let's remove this. If we want to add this, then we should locate this in `apps/example/board_specific`.
Just add `config LED_PIN` only. but, we should distinguish the config for app and driver. I think we should add `EXAMPLES_` prefix, I mean `config EXAMPLES_LED_PIN` is better.

sunghan-chang (approved):

kishore-sn (approved):

PR #6460, opened 2024-10-22
os/drivers/serial: Avoid Board Sleep in UART Tx

gSahitya-samsung:
This commit make use of pm_suspend and pm_resume APIs to avoid board sleep during uart_write operation. It register PM_UART_DOMAIN during uart_open to access PM APIs. The prevention of board sleep is required to avoid data loss during uart_write operation.

### DIFF
--- a/os/drivers/serial/serial.c
+++ b/os/drivers/serial/serial.c
@@ -77,6 +77,10 @@
 #include <tinyara/log_dump/log_dump.h>
 #include <tinyara/log_dump/log_dump_internal.h>
 #endif
+#ifdef CONFIG_PM
+#include <tinyara/pm/pm.h>
+#endif
+
 
 /************************************************************************************
  * Definitions
@@ -86,6 +90,10 @@
 
 #ifndef CONFIG_ARCH_LOWPUTC
 #error "Architecture must provide up_putc() for this driver"
+#endif
+
+#ifdef CONFIG_PM
+#define PM_UART_DOMAIN "UART"
 #endif
 
 #define uart_putc(ch) up_putc(ch)
@@ -113,6 +121,10 @@ static int uart_poll(FAR struct file *filep, FAR struct pollfd *fds, bool setup)
 /************************************************************************************
  * Private Variables
  ************************************************************************************/
+
+#ifdef CONFIG_PM
+static int pm_uart_domain_id = -1;
+#endif 
 
 static const struct file_operations g_serialops = {
 	uart_open,					/* open */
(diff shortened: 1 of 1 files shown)


ewoodev (approved):

os/drivers/serial/serial.c
```diff
@@ -1214,6 +1235,18 @@ static int uart_open(FAR struct file *filep)
...
 
+#ifdef CONFIG_PM
+	/* Register PM_UART_DOMAIN to access PM APIs during UART operations. */
+	if (pm_uart_domain_id == -1) {
+		ret = pm_domain_register(PM_UART_DOMAIN);
+		if (ret < 0) {
+			return ret;
+		} else {
+			pm_uart_domain_id = ret;
+		}
```
ewoodev, line 1246:
```suggestion
pm_uart_domain_id = ret;
```

it has return, we don't need to eles

sunghan-chang:
@gSahitya-samsung If we registet UART tx IRQ as a wakeup source, what happen?
@ziyik Is it possible? (Sorry, I don't know Dane's user id)

sunghan-chang:
@gSahitya-samsung The title is too long. How about `make pm suspend in uart tx`?

ziyik:
> @gSahitya-samsung If we registet UART tx IRQ as a wakeup source, what happen? @ziyik Is it possible? (Sorry, I don't know Dane's user id)

It does no make sense as register UART tx IRQ as a wakeup source, because when there is something to send the device should be in as wake condition since the data is send from the application or kernel.

ziyik:
@sunghan-chang, I have fixed an issue related to my previous PR, which is more related to the driver side.
Please refer to https://github.com/Samsung/TizenRT/pull/6407

sunghan-chang:
> @gSahitya-samsung The title is too long. How about `make pm suspend in uart tx`?

@gSahitya-samsung After modifying the title, let's merge. Could you check my comment?

sunghan-chang:
> > @gSahitya-samsung If we registet UART tx IRQ as a wakeup source, what happen? @ziyik Is it possible? (Sorry, I don't know Dane's user id)
> 
> It does no make sense as register UART tx IRQ as a wakeup source, because when there is something to send the device should be in as wake condition since the data is send from the application or kernel.

@ziyik Thank you for opinion.

gSahitya-samsung:
> > @gSahitya-samsung The title is too long. How about `make pm suspend in uart tx`?
> 
> @gSahitya-samsung After modifying the title, let's merge. Could you check my comment?

Done

sunghan-chang (approved):

ewoodev (approved):

sunghan-chang:
@gSahitya-samsung hmmmm. much minor but.. you also use pm_resume. And using pm_suspend is too surface(?) title? It does not represent what you modify. You want to keep wakeup state at tx operation instead of entering pm sleep.

sunghan-chang:
@gSahitya-samsung Let me suggest `avoid board sleep in UART Tx`. "use pm api" is no need.

sunghan-chang (approved):

PR #6466, opened 2024-10-25
utils/fscmd : Fix cat to display contents properly

Taejun-Kwon:
Previously, cat print out entire of buffer at once. But if there's new line or null character included in contents, it doesn't work properly. Hence print out character by character


jsdosa (approved):
LGTM.

PR #6467, opened 2024-10-28
Syntiant sdk update v112.3.5 samsung

cvejlbo:
Syntiant SDK update to v112.3.5-Samsung.  
Added NDP120 health check thread implementation


Taejun-Kwon:
I think we need version.h !!

Taejun-Kwon:
dsp_firmware_id49_v09_18_2024_sdkv112.2.0.synpkg
hi_bixby_model_581ua++v3_sdkv112.2.0.synpkg
mcu_fw_demo_2024_0818

cvejlbo:
> I think we need version.h !!

Just to be sure, do you mean you'd like to see a header file referencing the SDK version number?

cvejlbo:
> dsp_firmware_id49_v09_18_2024_sdkv112.2.0.synpkg hi_bixby_model_581ua++v3_sdkv112.2.0.synpkg mcu_fw_demo_2024_0818

The existing `hi_bixby_model_581ua++v3_sdkv112.2.0.synpkg` will work with SDK v112.3.5 as well, so that can be used.
For the DSP & MCU FW binaries, those are included in the SDK v112.3.5-Samsung SDK tarball available on Zendesk. 
For easy reference, I have added those two FW binaries explicitly to our Zendesk release page as well

Taejun-Kwon:
> > I think we need version.h !!
> 
> Just to be sure, do you mean you'd like to see a header file referencing the SDK version number?

Yes!

Taejun-Kwon:
> > dsp_firmware_id49_v09_18_2024_sdkv112.2.0.synpkg hi_bixby_model_581ua++v3_sdkv112.2.0.synpkg mcu_fw_demo_2024_0818
> 
> The existing `hi_bixby_model_581ua++v3_sdkv112.2.0.synpkg` will work with SDK v112.3.5 as well, so that can be used. For the DSP & MCU FW binaries, those are included in the SDK v112.3.5-Samsung SDK tarball available on Zendesk. For easy reference, I have added those two FW binaries explicitly to our Zendesk release page as well

Thanks I checked loading works properly.
BTW, we should update ours based on this PR, is this PR need to be updated much? can we start it now?

cvejlbo:
Right now only thing we have is `os/drivers/ai-soc/ndp120/include/syntiant_ilib/syntiant_ndp_ilib_version.h`- this has our release hash for the "ilib" host code part.  
We have a version file in our normal SDK release that I may be able to use, I'll need to double check and then if possible I can add a commit with this.  
Once done, I think we are OK to merge (I'm also OK to if we merge first and then add version file afterwards, if you prefer)

Taejun-Kwon:
> Right now only thing we have is `os/drivers/ai-soc/ndp120/include/syntiant_ilib/syntiant_ndp_ilib_version.h`- this has our release hash for the "ilib" host code part. We have a version file in our normal SDK release that I may be able to use, I'll need to double check and then if possible I can add a commit with this. Once done, I think we are OK to merge (I'm also OK to if we merge first and then add version file afterwards, if you prefer)

ok! one more question, can we use 32bit for i2s ? asking this because logic of barge-in has changed.

Taejun-Kwon:
> > Thanks, more question, can we use 32bit for i2s ? asking this because logic of barge-in has changed.
> > I think that might be tricky, but please can you elaborate here? What does "because logic of barge-in has changed." mean?
> 
> For older boards, when you say those older boards need to be supported, please can you describe the level of support needed?
I mean below.
new board : wake up + record + AEC
old board : wakeupo + record
Can you reply about bitrate of i2s too? if possible I want to rollback it to 32bit.

cvejlbo:
Using 32bit for i2s would mean the resulting BCLK is 3.072MHz when playing back @ 48kHz and also it would mean that the system would have to use a PDM clock of 3.072MHz. I have to discuss internally with respect to this to make sure I don't overlook anything with respect to that.

cvejlbo:
Circling back to your comment on version file, our SDK has plain `VERSION` file that we can add, it'll look something like this:
```
v112.3.5-Samsung
```
This can be read by humans to identify the Syntiant SDK version included in the code - but otherwise won't be included in the build (we have that other unique `SYNTIANT_NDP_ILIB_RELEASE_HASH` string which can be used for that)

Taejun-Kwon:
we will apply PR first.
https://github.com/Samsung/TizenRT/pull/6475

Taejun-Kwon:
Cade based this PR already merged, hence close it

PR #6468, opened 2024-10-29
os/board/rtl8730e: fix BLE GATT attribute notify & indicate

yeetee179:
- fixed issue of unable to set an attribute as notify and indicate at the same time


giwon-nam (approved):
It was confirmed that the BLE server operated normally when both INDICATE and NOTIFY were added to the properties.

hs36-kim (approved):
LGTM

PR #6471, opened 2024-10-30
os/arch/arm/src/amebasmart: Modify incorrect #if to #ifdef

ziyik:


sunghan-chang:
@ziyik Could you write some description in the commit?
Because CONFIG_RAM_DDR has two states, defined and undefined, not value change, it should be checked with "defined"

sunghan-chang (approved):

PR #6473, opened 2024-11-01
os/arch/arm/src/amebasmart: stop sending i2s data if there is no pend…

zhongnuo-tang:
…ing buffer containing in tx.pend list

1. pause i2s send when there is no buffer in tx.pend list
2. it removes noise, (mute instead) if there is insufficient time to get the next buffer from application due to heavy loading


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -557,6 +558,8 @@ static void i2s_tx_schedule(struct amebasmart_i2s_s *priv, int result)
 
 		/* Start next transfer */
 		amebasmart_i2s_tx(priv, bfcontainer);
+	} else {
+		ameba_i2s_pause(priv->i2s_object);
```
Taejun-Kwon, line 562: To avoid unexpected case by DMA Timeout, I suggest below
```
		if ((priv->apb_tx->nbytes - priv->apb_tx->curbyte) <= 0)
			ameba_i2s_pause(priv->i2s_object);
		}
```

sunghan-chang (approved):

PR #6476, opened 2024-11-03
os/arch/arm/src/amebasmart: update I2S PM domain control

zhongnuo-tang:
1. domain control should be added at start i2s DMA tx and end of i2s DMA tx.
2. original place is the end of CPU filling up DMA page, but not the end of DMA transfer, resulting in enter sleep before I2S tx done.
3. add i2s_lock_state flag to keep track current domain status, since i2s_send will be called multiple times by application, it should only lock for the first time, and release the lock only when i2s_stop is called.
4. verified that it will only enter sleep after I2S tx done, and after wakeup it can start I2S tx again.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -122,6 +122,10 @@
 #define I2S_DMA_PAGE_SIZE 4096 	/* 4 ~ 16384, set to a factor of APB size */
 #define I2S_DMA_PAGE_NUM 4	/* Vaild number is 2~4 */
 
+#ifdef CONFIG_PM
+static volatile bool i2s_lock_state = 0;
+#endif
+
 struct amebasmart_buffer_s {
 	struct amebasmart_buffer_s *flink; /* Supports a singly linked list */
 	i2s_callback_t callback;		   /* Function to call when the transfer completes */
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -122,6 +122,10 @@
 #define I2S_DMA_PAGE_SIZE 4096 	/* 4 ~ 16384, set to a factor of APB size */
 #define I2S_DMA_PAGE_NUM 4	/* Vaild number is 2~4 */
 
+#ifdef CONFIG_PM
+static bool i2s_lock_state = 0;
```
edwakuwaku, line 126: static volatile

sunghan-chang: @zhongnuo-tang Could you check @edwakuwaku 's comment?

zhongnuo-tang:
Before this PR:
The module will enter PG while player is playing audio, and after wakeup the audio is not working anymore.
With this PR:
The module will only enter PG after the audio is played finished/paused. After wakeup audio can be played again, working correctly. 
Tested with soundplayer example

sunghan-chang:
@gSahitya-samsung @abhishek-samsung Could you verify this on bixby when Gaurav comes here?

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -655,7 +653,12 @@ static int i2s_send(struct i2s_dev_s *dev, struct ap_buffer_s *apb, i2s_callback
 	struct amebasmart_i2s_s *priv = (struct amebasmart_i2s_s *)dev;
 
 	DEBUGASSERT(priv && apb);
-
+#ifdef CONFIG_PM
+	if (!i2s_lock_state) {
+		i2s_lock_state = 1;
```
gSahitya-samsung, line 658: Verified basic operations and bixby

ewoodev (approved):

PR #6482, opened 2024-11-05
wdog/wd_timer_nohz: Move correcting wd missing tick to pm_wakehandler

ewoodev:
There is an issue where after waking up from UART RX, even if pm_timedsuspend is called, it immediately enters enter_board_sleep again.

since, The missed ticks while the board is asleep will be compensated for after waking up. At this time, the wd timer is corrected in the first tick ISR that comes after wakeup.

but UART RX ISR is comming first than tick ISR.
So even if call pm_timedsuspend whick is using wd timer, it will expired in first tick ISR after wakeup.

The issue flow likes below:
WAKE HANDLER -> UART RX ISR -> UART READ THREAD -> call pm_timedsuspend -> TICK ISR (Collect missing tick and expire)

Therefore, Move correcting wd missing tick to pm_wakehandler. but The wd timer is corrected in pm_wakehandler and the timer expiration is performed in the first ISR after wakeup. This is to avoid delaying the response of the UART or other HW IRQ ISR due to a timer expiration operation.

The fixed flow:
WAKE HANDLER (Collect missing tick) -> UART RX ISR -> UART READ THREAD -> call pm_timedsuspend -> TICK ISR (expire)

### DIFF
--- a/os/kernel/wdog/wd_start.c
+++ b/os/kernel/wdog/wd_start.c
@@ -107,9 +107,6 @@ typedef void (*wdentry4_t)(int argc, uint32_t arg1, uint32_t arg2, uint32_t arg3
 /****************************************************************************
  * Private Variables
  ****************************************************************************/
-#ifdef CONFIG_SCHED_TICKSUPPRESS
-static clock_t g_wd_ticks;
-#endif
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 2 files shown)


edwakuwaku (approved):

sunghan-chang (dismissed):
Approved. But need to rewrite some sentences.

sunghan-chang (dismissed):

os/kernel/wdog/wd_start.c
```diff
@@ -494,6 +470,21 @@ void wd_timer(void)
...
+	for (wdog = g_wdactivelist.head; ticks > 0 && wdog; wdog = wdog->next) {
+
+		/* Decrement the lag for this watchdog. */
+
+		decr = MIN(wdog->lag, ticks);
+
+		/* There are.  Decrement the lag counter */
+
+		wdog->lag -= decr;
+		ticks -= decr;
```
ziyik, line 485: Since it has been modify from g_wd_ticks to ticks, I think we can remove this line?

ziyik: Other than this I have no comment on the PR.

sunghan-chang: The decr can be `wdog->lag`, not `ticks`. This case, we need this line.
When the decr is `ticks`, this line is unnecessary.

sunghan-chang: Did I get your point exactly?

ziyik: yes, correct

os/kernel/wdog/wd_start.c
```diff
@@ -494,6 +470,21 @@ void wd_timer(void)
 #ifdef CONFIG_SCHED_TICKSUPPRESS
 void wd_timer_nohz(clock_t ticks)
 {
-	g_wd_ticks += ticks;
```
sunghan-chang, line 497:
```
#ifdef CONFIG_SCHED_TICKSUPPRESS
static clock_t g_wd_ticks;
#endif
```
This looks unnecessary. Let's remove.

sunghan-chang (approved):

PR #6483, opened 2024-11-05
os/board/rtl8730e: Replace Semaphore with Polling in IPC_wait_idle

ZhenBei-Sin:
Change Notes:
1. Removed semaphore related in IPC_wait_idle.
2. Replaced semaphore wait logic with polling mechanism to check ipc_Semaphore[IPC_ChNum].
3. Introduced a polling loop that repeatedly checks ipc_Semaphore[IPC_ChNum] every 5 microseconds until it becomes NULL.
4. Moved clearing of ipc_Semaphore[ChanNum] to IPC_TXHandler by setting it to NULL after handling the interrupt.

### DIFF
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_ipc_api.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_ipc_api.c
@@ -128,14 +128,7 @@ void IPC_TXHandler(VOID *Data, u32 IrqStatus, u32 ChanNum)
 
 	IPC_INTConfig(IPCx, ChanNum, DISABLE);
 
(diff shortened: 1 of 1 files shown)


ZhenBei-Sin:
HI @sunghan-chang  @ziyik 

This may not be the final version, as it represents a temporary change and fix for the current stage.
Let's have discussion about this.

sunghan-chang:
@ZhenBei-Sin @ziyik If we move the logic of sem_wait for ipc before interrupt lock instead of this PR, what happen?
The reason why we make this PR is to prevent context switching by sem_wait after interrupt lock. If we do it like above, it will not do. How do you think?

sunghan-chang:
@ZhenBei-Sin @ziyik Unfortunately @jeongarmy meets the issue again with this PR.....

PR #6484, opened 2024-11-05
build/tools/amebasmart: Sync PR#6102 into the SE

ZhenBei-Sin:
Change Notes:
[In ATF SE]
1. Add check spic busy condition before entering user mode for spic
2. Apply DSB and ISB usage to enter/exit user mode to protect from randomness of program execution from cpu


sunghan-chang (approved):

sunghan-chang:
@ZhenBei-Sin @ziyik We will stop today's debugging now. Thank you for support. Let's continue at Thursday. Good morning..

PR #6487, opened 2024-11-07
[Do Not Merge] os/pm: Change pm_suspend and pm_resume API logic

gSahitya-samsung:
This commit change the logic of pm_suspend and pm_resume API. Now, instead of incrementing or decrementing suspend counts, we maintain only suspended domain flag to prevent board sleep.


neel-samsung:
Please close the PR if not needed

PR #6491, opened 2024-11-08
Fix AiLite Hang issues during boot and onboarding

kishore-sn:


os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -61,6 +61,23 @@ uint32_t up_get_gating_flag_status(uint32_t CoreID)
 	return g_cpugating_flag[CoreID];
 }
 
+
+void up_do_gating()
```
sunghan-chang, line 65: () -> (void)

os/kernel/irq/irq_csection.c
```diff
@@ -264,13 +265,17 @@ irqstate_t enter_critical_section(void)
...
 					 */
-				if (!paused)
-				{
-					up_cpu_paused_save();
-				}
-
-					DEBUGVERIFY(up_cpu_paused(cpu));
-				paused = true;
+					if (up_cpu_pausereq(cpu)) {
+						if (!paused)
```
sunghan-chang, line 269: if ( ) {

os/arch/arm/src/armv7-a/arm_cpugating.c
kishore-sn: Done

os/kernel/irq/irq_csection.c
kishore-sn: done

os/kernel/irq/irq_csection.c
```diff
@@ -123,7 +123,7 @@ static bool irq_waitlock(int cpu)
 	while (spin_trylock_wo_note(&g_cpu_irqlock) == SP_LOCKED) {
 		/* Is a pause request pending? */
 
-		if (up_cpu_pausereq(cpu)) {
+		if (up_cpu_pausereq(cpu) || up_get_gating_flag_status(cpu) == 1) {
```
edwakuwaku, line 126: This will break the standardisation for kernel codes, it's difficult to understand what does flag status = 1 correlates to, it is too chipset specific (ie. for RTL8730E flash operation).
Some countermeasure should be taken to address this issue later.

edwakuwaku:
Let's change topic, AILite -> RTL8730E

os/kernel/irq/irq_csection.c
```diff
@@ -264,13 +265,16 @@ irqstate_t enter_critical_section(void)
...
-				paused = true;
+					if (up_cpu_pausereq(cpu)) {
+						if (!paused) {
+							up_cpu_paused_save();
+						}
+
+						DEBUGVERIFY(up_cpu_paused(cpu));
+						paused = true;
+					} else {
+						up_do_gating();
```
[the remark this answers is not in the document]
edwakuwaku: If you are forcing up_do_gating here, the pending gating SGI might cause issue at later stage.
So, we can restore irq here to handle the request first, and we can return back here to continue the rest.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_flash_ram.c
```diff
@@ -86,15 +86,15 @@ void FLASH_Write_Lock(void)
 #endif
 	/* disable irq */
 	/* We do not need to acquire lock for this core, as the other core will enter gating */
-	PrevIrqStatus = irqsave();
+	PrevIrqStatus = enter_critical_section();
```
edwakuwaku, line 89: Thanks for identifying this, the irqsave/restore was set here purposely during previous work. I found some booting hang issue and think that the core doing flash operation should not compete for spin lock.
May I know if this change is able to cover all the cases, I mean under complex situation? (With considering the change in irq_csection.c)

os/kernel/irq/irq_csection.c
```diff
@@ -264,13 +265,16 @@ irqstate_t enter_critical_section(void)
...
-				if (!paused)
-				{
-					up_cpu_paused_save();
-				}
-
-					DEBUGVERIFY(up_cpu_paused(cpu));
-				paused = true;
+					if (up_cpu_pausereq(cpu)) {
+						if (!paused) {
+							up_cpu_paused_save();
```
r-prabu, line 270: Minor: Is there any alignment issue here?

r-prabu (approved):

samsung-singh (approved):
hang issue during bootup and onboarding has been fix for ai lite board, change is good to merge.

os/kernel/irq/irq_csection.c
kishore-sn: You are right. But since the code in the kernel folder is interacting with arch code, i think it is necessary to call low level apis. As of now, I am not sure how we can change this. Maybe in future, if we add more chipsets, then we can handle this in a better way.

edwakuwaku (approved):
Good to merge

sunghan-chang (approved):

PR #6500, opened 2024-11-11
os/arch/arm/src/amebasmart: fix module cannot enter PG when LCD and M…

zhongnuo-tang:
…IPI are enabled

1. domain control is not added at MIPI end of transmission/failure to create packet, result in counter not decrement to 0, causing board not go to PG even all data are tx
2. add bsp_pm_domain_control to ensure it is in pairs.


edwakuwaku (approved):

zhongnuo-tang:
before this PR:
When using defconfig loadable_ext_ddr, when LCD/MIPI is enabled, the module is not able to enter PG, due to bsp_pm_domain_control not called in amebasmart_mipi_transfer at the end of transmission, as a result suspend_count for BSP_MIPI_DRV domain is kept at 1, blocking the module to enter PG.

This issue is fixed with this PR, now the module can enter PG, and after wakeup example_lcd is working fine.

ritesh55555 (approved):
It is working fine, verified

ewoodev (approved):

sunghan-chang (approved):

PR #6501, opened 2024-11-11
system/utils: fixes in cpuload measurement interface

ritesh55555:
Due to addition of extra column(cpu for SMP case) in proc state, there was some issues while printing cpuload measurement for single core case. Also some title were missing in cpu load measurement output for single core case. So handled those cases for the single core case.

This is the new output:

```
--------------------------------------------------                                                       
Non-Zero CPU utilization trend (updated every 5s)                                                        
PID | Pri |  CPU  | Task Name  
--------------------------------------------------
  0 |   0 |  98.2 | Idle Task
 11 | 203 |   0.1 | binary_manager
 14 | 180 |   0.1 | app1
 16 | 125 |   0.1 | tash
 19 | 100 |   2.6 | CPULoadMonitor
--------------------------------------------------

```


apps/system/utils/utils_ps.c
```diff
@@ -133,12 +133,17 @@ static void ps_print_values(char *buf, void *arg)
...
 	printf("%5s | %4s | %4s | %7s | %c%c | %8s | %3s", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
 		flags & TCB_FLAG_ROUND_ROBIN ? "RR  " : "FIFO", utils_ttypenames[(flags & TCB_FLAG_TTYPE_MASK) >> TCB_FLAG_TTYPE_SHIFT], \
 		flags & TCB_FLAG_NONCANCELABLE ? 'N' : ' ', flags & TCB_FLAG_CANCEL_PENDING ? 'P' : ' ', \
 		utils_statenames[state], stat_info[PROC_STAT_CPU]);
+#else
+	printf("%5s | %4s | %4s | %7s | %c%c | %8s ", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
+		flags & TCB_FLAG_ROUND_ROBIN ? "RR  " : "FIFO", utils_ttypenames[(flags & TCB_FLAG_TTYPE_MASK) >> TCB_FLAG_TTYPE_SHIFT], \
+		flags & TCB_FLAG_NONCANCELABLE ? 'N' : ' ', flags & TCB_FLAG_CANCEL_PENDING ? 'P' : ' ', \
+		utils_statenames[state]);
+#endif
```
sunghan-chang, line 146: Let's reduce duplicated codes.
```
	printf("%5s | %4s | %4s | %7s | %c%c | %8s", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
		flags & TCB_FLAG_ROUND_ROBIN ? "RR  " : "FIFO", utils_ttypenames[(flags & TCB_FLAG_TTYPE_MASK) >> TCB_FLAG_TTYPE_SHIFT], \
		flags & TCB_FLAG_NONCANCELABLE ? 'N' : ' ', flags & TCB_FLAG_CANCEL_PENDING ? 'P' : ' ', \
		utils_statenames[state]);

#ifdef CONFIG_SMP
	cpu_idx = atoi(stat_info[PROC_STAT_CPU]);
	if (cpu_idx != cpu && cpu != CONFIG_SMP_NCPUS) {
		return;
	}

	printf(" | %3s", stat_info[PROC_STAT_CPU]);
#endif
```

1. You should remove a space after `%8s`.
2. You can keep #else code only, please find above.

And this is a change in ps, not in cpuload. Let's split a commit.

sunghan-chang: Instead of splitting, let's modify the title and description.

apps/system/utils/utils_cpuload.c
```diff
@@ -249,7 +249,10 @@ static void cpuload_print_normal(void)
 				printf("  CPU%d |", j);
 			}
 		}
+#else
+		printf("  CPU  |");
 #endif
+		printf(" Task Name  ");
```
sunghan-chang, line 255: Line 210,
```
#if (CONFIG_TASK_NAME_SIZE > 0)
	printf(" %s", stat_info[PROC_STAT_NAME]);
#endif
	printf("\n");
```
When the config is not enabled, it does not print thread name, but title prints it always.
How about modifying line 210 like below?
```
#if (CONFIG_TASK_NAME_SIZE > 0)
	printf(" %s\n", stat_info[PROC_STAT_NAME]);
#else
	printf(" NA\n"); 
#endif
```
Need to remove line 213

apps/system/utils/utils_ps.c
```diff
@@ -128,17 +128,19 @@ static void ps_print_values(char *buf, void *arg)
...
 
+	printf("%5s | %4s | %4s | %7s | %c%c | %8s", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
+		flags & TCB_FLAG_ROUND_ROBIN ? "RR  " : "FIFO", utils_ttypenames[(flags & TCB_FLAG_TTYPE_MASK) >> TCB_FLAG_TTYPE_SHIFT], \
+		flags & TCB_FLAG_NONCANCELABLE ? 'N' : ' ', flags & TCB_FLAG_CANCEL_PENDING ? 'P' : ' ', \
+		utils_statenames[state]);
+
 #ifdef CONFIG_SMP
 	cpu_idx = atoi(stat_info[PROC_STAT_CPU]);
 	if (cpu_idx != cpu && cpu != CONFIG_SMP_NCPUS) {
 		return;
```
sunghan-chang, line 139: What happen if we meet this return?
Just `return` is enough? Could you verify that?

sunghan-chang: Then, how about checking this at very first?
Even others have `return`, they are error cases (I mean it's difficult to meet).
but this is a normal case (easy to meet). Adjusting the location will be good I think,

ritesh55555: done that already

apps/system/utils/utils_ps.c
```diff
@@ -128,17 +133,14 @@ static void ps_print_values(char *buf, void *arg)
...
-	printf("%5s | %4s | %4s | %7s | %c%c | %8s | %3s", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
+	printf("%5s | %4s | %4s | %7s | %c%c | %8s", stat_info[PROC_STAT_PID], stat_info[PROC_STAT_PRIORITY], \
 		flags & TCB_FLAG_ROUND_ROBIN ? "RR  " : "FIFO", utils_ttypenames[(flags & TCB_FLAG_TTYPE_MASK) >> TCB_FLAG_TTYPE_SHIFT], \
 		flags & TCB_FLAG_NONCANCELABLE ? 'N' : ' ', flags & TCB_FLAG_CANCEL_PENDING ? 'P' : ' ', \
-		utils_statenames[state], stat_info[PROC_STAT_CPU]);
+		utils_statenames[state]);
+
+#ifdef CONFIG_SMP
+	printf(" | %3s", stat_info[PROC_STAT_CPU]);
+#endif
```
sunghan-chang, line 143: For SMP and for name, how about using the same way as cpuload?
I mean
1. printing all titles including cpu and name
2. when you print real value here, if cpu or name are disabled, then just print N/A

I wanna keep the code simple. It has too many conditionals now.

ritesh55555: done

r-prabu (approved):

sunghan-chang (approved):

kishore-sn (approved):

PR #6503, opened 2024-11-12
os/kernel/task: Fix wrong output in asserted tcb's syscall details

neel-samsung:
When assert happens in syscall, we need to know the return address of syscall. Currently, we print the address of the buffer instead of data (return address of syscall). This patch fix this improper print issue.


samsung-singh (approved):
fixed output in asserted tcb system call details, this change is good to merge.

os/kernel/task/task_dumptasks.c
```diff
@@ -106,8 +109,10 @@ void task_show_tcbinfo(struct tcb_s *tcb)
 	lldbg("Sigdeliver  : %p\n", tcb->xcp.sigdeliver);
 #endif
 #ifdef CONFIG_LIB_SYSCALL
-	lldbg("Nsyscalls   : %u\n", tcb->xcp.nsyscalls);
-	lldbg("Syscall     : %p\n", tcb->xcp.syscall);
+	lldbg("Nsyscalls   : %u\n", nsyscall);
+	for (int i = 0; i < nsyscall; i++) {
+		lldbg("Syscall     : %p\n", tcb->xcp.syscall[i].sysreturn);
```
sunghan-chang, line 114: How about printing index?
And what does it mean, index 0 and index 1?
what data is in index 0 and index 1?

sunghan-chang (commented):
@neel-samsung Could you check the TRAP with this change? This value should be parsed in the TRAP.

os/kernel/task/task_dumptasks.c
kishore-sn: System call return address is stored in the syscall[] array.
It points to the next instruction after syscall is executed.
And, it seems that syscalls can be nested. So each index in the array indicates the nesting level.

@neel-samsung Please print the index also as requested by Chang.

neel-samsung: I have added the index in assert log as requested

neel-samsung:
> @neel-samsung Could you check the TRAP with this change? This value should be parsed in the TRAP.

Yes, I have parsed the whole TCB information ( State, Flags, and Syscall) as described in the commit description.

os/kernel/task/task_dumptasks.c
```diff
@@ -106,8 +109,10 @@ void task_show_tcbinfo(struct tcb_s *tcb)
 	lldbg("Sigdeliver  : %p\n", tcb->xcp.sigdeliver);
 #endif
 #ifdef CONFIG_LIB_SYSCALL
-	lldbg("Nsyscalls   : %u\n", tcb->xcp.nsyscalls);
-	lldbg("Syscall     : %p\n", tcb->xcp.syscall);
+	lldbg("Nsyscalls   : %u\n", nsyscall);
+	for (int i = 0; i < nsyscall; i++) {
+		lldbg("Syscall %d   : %p\n", i+1, tcb->xcp.syscall[i].sysreturn);
```
sunghan-chang, line 114: 1. i + 1
2. Could you let me know why you print the value with `+1`?

sunghan-chang: Oh.. ok. It's matched with nsyscall.
I found some assert code prints different index number. I don't remember what it is but current assert prints now. Could you check and use the same index range?
I mean in assert code, same buffer is printed twice.
First, it prints index `0~1`. Second, it prints index `1~2`.

r-prabu (approved):

tools/trap/cli/logParser.py
```diff
@@ -574,6 +578,90 @@ def print_wrong_sp(self):
 								stack_val = stack_val - int(self.g_stext_app[is_app_symbol - 1])
 							utils.print_symbol(stack_addr, stack_val, is_app_symbol, self.bin_path, self.app_name)
 
+	# API to parse asserted tcb information
+	def parse_tcb_info(self):
+		TCB_FLAG_TTYPE_SHIFT      = (0)	                         # Bits 0-1: thread type
```
gSahitya-samsung, line 583: Do we need to declare constants local to function?
Can we move them inside class declarations?

os/kernel/task/task_dumptasks.c
neel-samsung: okay.. I have changed the indexing based upon the indexing present in the base code. That is 0-based indexing for syscall.

tools/trap/cli/logParser.py
neel-samsung: Okay.. It's better to declare them global rather confine them to just class?

os/kernel/task/task_dumptasks.c
```diff
@@ -106,8 +109,10 @@ void task_show_tcbinfo(struct tcb_s *tcb)
 	lldbg("Sigdeliver  : %p\n", tcb->xcp.sigdeliver);
 #endif
 #ifdef CONFIG_LIB_SYSCALL
-	lldbg("Nsyscalls   : %u\n", tcb->xcp.nsyscalls);
-	lldbg("Syscall     : %p\n", tcb->xcp.syscall);
+	lldbg("Nsyscalls   : %u\n", nsyscall);
+	for (int i = 0; i < nsyscall; i++) {
+		lldbg("Syscall %d   : %p\n", i, tcb->xcp.syscall[i].sysreturn);
```
sunghan-chang, line 114: How about `syscall #%d` to say that it's an index?

sunghan-chang (approved):

kishore-sn (approved):

PR #6504, opened 2024-11-12
os/board/rtl8730e: get BLE MAC address before BLE init

yeetee179:
- add hci_platform_get_ble_mac_address api to get address from efuse
- hci_platform_get_ble_mac_address can be used before BLE init


hs36-kim (approved):
LGTM

PR #6506, opened 2024-11-13
os/arch/arm/src/armv7-a: Add logic for cpu pause

kishore-sn:
If a cpu is already paused and again up_cpu_pause is called, then it might result in a lock up due to waiting on semaphores. So, in up_cpu_pause we check whether a pause reques is pending or whether the cpu is already paused. Then we return without sending another pause request.


edwakuwaku:
Under what condition will multiple pause occurs?

neel-samsung (approved):
CPU Pause PR looks ready to merge

samsung-singh (approved):
code changes done for pausing the cpu and checking cpu pause state, change is good to merge.

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -222,8 +242,8 @@ int up_cpu_paused_restore(void)
 	 * will be made when the interrupt returns.
 	 */
 
-  up_restoretask(tcb);
-  arm_restorestate(tcb->xcp.regs);
+	up_restoretask(tcb);
+	arm_restorestate(tcb->xcp.regs);
```
sunghan-chang, line 246: Let's split commit for beautifuller.

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -300,7 +320,15 @@ int arm_pause_handler(int irq, void *context, void *arg)
...
 int up_cpu_pause(int cpu)
 {
-  DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
+	DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
+
+	/* If a pause request is already pending or if the cpu is already in
+	 * paused state, then dont send the pause request again.
+	 */
+
+	if (up_cpu_pausereq(cpu) || up_is_cpu_paused(cpu)) {
```
gSahitya-samsung, line 329: cpu pause logic looks fine to merge

gSahitya-samsung (approved):

os/arch/arm/src/armv7-a/arm_cpupause.c
kishore-sn: Done. Thanks.

kishore-sn:
> Under what condition will multiple pause occurs?

It is very unlikely. 
However, we are currently facing this issue when we try to reload apps. 
arm_assert -> pause cpu to until reload apps completed -> start app reload -> mq operation -> sched operation -> again calls cpu_pause. 

I raised this PR for this case. However, I think if the cpu is already in pause state, then we must not do the pause operation again.

edwakuwaku (approved):
Change seems decent, good to merge.

sunghan-chang (approved):

PR #6508, opened 2024-11-13
os/arc/arm/src/amebasmart: set loguart isr priority as lowest

zhongnuo-tang:
1. loguart isr should set lower than other peripheral priority


jeongarmy (approved):
lgtm

PR #6513, opened 2024-11-16
Media : Add new volume logic & update wakerec

Taejun-Kwon:
1. New volume logic based on stream policy added.
2. Refactor wakerec to work based on latest code


jsdosa (approved):
LGTM.

PR #6514, opened 2024-11-18
os/arch/arm/src/amebasmart: fix LOGUART Tx IRQ handling for Tx FIFO f…

zhongnuo-tang:
…ull case

1. remove previous workaround
2. issue found at up_lowgetc where it disable loguart irq during get char and set it back to 1 after get char
3. it should read the IMR status and restore it since now tx irq is set


zhongnuo-tang:
Before this PR:
During printing of logs, if loguart rx irq is triggered, and at the same time TX FIFO is full, it will stop printing and result in watchdog reboot after 3 mins. Previous workaround was to manually enable tx irq.

The root cause was identified in up_lowgetc where it disabled irq and reset it back to 1, instead of restoring the previous irq value.
with this PR, there is no hang when loguart rx irq is triggered.

ewoodev (approved):

sunghan-chang (approved):

PR #6518, opened 2024-11-20
[DO NOT MERGE] os/drivers/input : TSP firmware update and Alive check recovery

anjana348:
Test code to check reset is working or not with TSP firmware update

>> SRID does not have test board with TSP reset connection so Please check if reset is working fine.
On running test app "lcd_test", it will update firmware and call reset.

In SRID, reset doesn't work so touch interrupt stops working after firmware update.
Need to manually remove power and comment update firmware code to run touch again.


os/drivers/input/ist415.c
```diff
@@ -81,17 +87,116 @@ struct touchscreen_ops_s g_ist415_ops = {
...
+}
 
 /****************************************************************************
- * Name: ist415_get_touch_data
+ * Name: ist415_check_status
  ****************************************************************************/
+static void ist415_check_status(FAR void *arg)
+{
+	u32 touch_status = 0;
+	ct += 5;
```
sunghan-chang, line 177: 1. Where is the ct definition?
2. What does 5 mean?

os/drivers/input/ist415.c
```diff
@@ -157,10 +267,13 @@ static int ist415_get_touch_data(struct ist415_dev_s *dev, FAR void *buf)
 				data->point[i].flags = TOUCH_UP;
 				break;
 			}
-			touchvdbg("COORDINATES: id %d status %d type %d x : %d y : %d\n", data->point[i].id, p_evt_coord->tsta, (p_evt_coord->ttype_3_2 << 2) | p_evt_coord->ttype_1_0, data->point[i].x, data->point[i].y);
+			printf("COORDINATES: id %d status %d type %d x : %d y : %d\n", data->point[i].id, p_evt_coord->tsta, (p_evt_coord->ttype_3_2 << 2) | p_evt_coord->ttype_1_0, data->point[i].x, data->point[i].y);
```
sunghan-chang, line 270: Could you let me know why you do this?

anjana348:
@sunghan-chang - The PR is not clean, it is raised to test reset operation in HQ as SRID board doesn't have reset pin.
I will change it to draft mode

PR #6522, opened 2024-11-21
[DO NOT REVIEW / MERGE] os/drivers/input : TSP firmware update and Alive check recovery

anjana348:
Test code to check reset is working or not with TSP firmware update-
By default, firmware update is disabled. Alive check works every 2 seconds.

If need to enable firmware update, please copy firmware file to 
tinyara/input/240619_AV_Display_V02010206.h

and enable CONFIG_TOUCH_FIRMWARE_UPDATE=y
run "lcd_test update"
It will update firmware and call reset
If reset pin works, touch should work otherwise need to manually power off and run "lcd_test" again to check touch.

### DIFF
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -85,6 +85,7 @@ static int yres;
 #define EXAMPLE_LCD_FPS_TEST 5000
 #endif
 
+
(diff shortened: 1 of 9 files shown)


sunghan-chang:
@anjana348 This is merged in the PR #6579 . Let's close this.

PR #6526, opened 2024-11-22
build/tools/amebasmart: Update wififw and adjust RA table

jwei5:
- Update wififw version codes and adjust RA table to improve open space performance


hs36-kim (approved):
temp merge
(after testing, this commit can be revert)

PR #6532, opened 2024-11-25
os/board/rtl8730e/src/ change UART1 PuPd during PG

zhongnuo-tang:
1. instead of setting pupd to DOWN, it should remain its last status before sleep


zhongnuo-tang:
initially in pinmap config, UART1 pins are set to pull down when enter PG, Samsung request it to be kept the last status before sleep, so GPIO_PuPd_KEEP should be used

sunghan-chang:
@zhongnuo-tang We will get exact usages for all of pins later. Then let us make changes. Let's close this at this time.

PR #6534, opened 2024-11-25
os/arch/arm/src/amebasmart: Enable/disable loguart rx irq according t…

zhongnuo-tang:
…o driver call

1. Loguart rx irq should be handled properly with enable/disable

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_serial.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_serial.c
@@ -717,10 +717,11 @@ static void rtl8730e_log_up_rxint(struct uart_dev_s *dev, bool enable)
 	struct rtl8730e_up_dev_s *priv = (struct rtl8730e_up_dev_s *)dev->priv;
 	DEBUGASSERT(priv);
 	priv->rxint_enable = enable;
-	//if (enable)
-		//LOGUART_RxCmd(LOGUART_DEV, ENABLE);
-	//else
-		//LOGUART_RxCmd(LOGUART_DEV, DISABLE);
+	if (enable) {
+		LOGUART_INTConfig(LOGUART_DEV, LOGUART_BIT_ERBI, ENABLE);
+	} else {
+		LOGUART_INTConfig(LOGUART_DEV, LOGUART_BIT_ERBI, DISABLE);
+	}
 }
 
 /****************************************************************************
--- a/os/board/rtl8730e/src/rtl8730e_boot.c
+++ b/os/board/rtl8730e/src/rtl8730e_boot.c
@@ -519,10 +519,6 @@ void board_initialize(void)
 	ipc_msg_loguart.rsvd = 0; /* for coverity init issue */
 	ipc_send_message(IPC_AP_TO_LP, IPC_A2L_DISLOGUART, &ipc_msg_loguart);
 
-	/* Tizen: re-enable the LOGUART Rx event interrupt so that CA32 can use it, after KM0 LOGUART is disabled */
-	irqstate_t flags = enter_critical_section();
-	LOGUART_INTConfig(LOGUART_DEV, LOGUART_BIT_ERBI, ENABLE);
-	leave_critical_section(flags);
 }
 #else
 #error "CONFIG_BOARD_INITIALIZE MUST ENABLE"


os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -717,10 +717,13 @@ static void rtl8730e_log_up_rxint(struct uart_dev_s *dev, bool enable)
...
 	DEBUGASSERT(priv);
 	priv->rxint_enable = enable;
-	//if (enable)
+	if (enable) {
+		LOGUART_INTConfig(LOGUART_DEV, LOGUART_BIT_ERBI, ENABLE);
 		//LOGUART_RxCmd(LOGUART_DEV, ENABLE);
-	//else
+	} else {
+		LOGUART_INTConfig(LOGUART_DEV, LOGUART_BIT_ERBI, DISABLE);
 		//LOGUART_RxCmd(LOGUART_DEV, DISABLE);
```
ewoodev, line 725: why you keep this code??

ewoodev (commented):
I would like to check how it is currently going from booting to when LOG UART RX is first enabled in this regard.

if we enable here, do we need to set privious changs of LOG UART enable during boot?

ewoodev:
The UART porting layer of TizenRT remains unchanged from amebaD.
I think that routing the LOG UART through up_lowgetc and up_lowputc has caused a number of issues.
Could you please explain why this structure was changed and why it is being maintained?

zhongnuo-tang:
> I would like to check how it is currently going from booting to when LOG UART RX is first enabled in this regard.
> 
> if we enable here, do we need to set privious changs of LOG UART enable during boot?

Hi @ewoodev, we can remove the enable during boot with this changes.

zhongnuo-tang:
> The UART porting layer of TizenRT remains unchanged from amebaD. I think that routing the LOG UART through up_lowgetc and up_lowputc has caused a number of issues. Could you please explain why this structure was changed and why it is being maintained?

amebaD's loguart is a common UART which can use the same set of structure like UART1.. However, in amebasmart, there is a dedicated UART for LOGUART only, that is why we are using a separate set of flow different from UART0/1/2..
image

os/arch/arm/src/amebasmart/amebasmart_serial.c
zhongnuo-tang: removed

sunghan-chang:
@zhongnuo-tang Could you let us know some information as below?
1. How did you verify the PR?
2. If we don't apply this, what happen? (I mean what was the issue?)

sunghan-chang:
@ewoodev @gSahitya-samsung @neel-samsung @ritesh55555 This is necessary to review and to verify. Could you check that?

ewoodev:
Hello @zhongnuo-tang 
Could you please verify uart stuck if enter tash during booting??

zhongnuo-tang:
> Hello @zhongnuo-tang Could you please verify uart stuck if enter tash during booting??

verified with loadable_ext DDR, spamming tash during booting, no stuck occurred.

sunghan-chang (approved):

ewoodev:
> > Hello @zhongnuo-tang Could you please verify uart stuck if enter tash during booting??
> 
> verified with loadable_ext DDR, spamming tash during booting, no stuck occurred.

Could you please verify psram also?

zhongnuo-tang:
> > > Hello @zhongnuo-tang Could you please verify uart stuck if enter tash during booting??
> > 
> > 
> > verified with loadable_ext DDR, spamming tash during booting, no stuck occurred.
> 
> In my test, the stuck is occurred. Could you please verify psram also?

Hi @ewoodev,
Using latest tizenrt sha:511aa520a8c9b080331d60522fe85aa7ba1bfca1
Verified with loadable_ext_psarm: spamming tash during booting, no stuck occurred.
Verified with loadable_ext_ddr: spamming tash during booting, no stuck occurred.
Verified with loadable_apps: spamming tash during booting, no stuck occurred.

Maybe you can share with the git sha u tested on? or which defconfig u are using?

ewoodev (approved):

jylee9613 (approved):

sunghan-chang (approved):

PR #6535, opened 2024-11-25
build/tools/amebasmart/gnu_utility: limit frequency of sending probe response

ZivLow:
1. limit the frequency to send probe response packet by saving wifi hal bandwith in noisy environment to ensure softap connection


hs36-kim (approved):
LGTM
 3층에서 작동확인함

PR #6543, opened 2024-11-28
os/drivers/lcd/: Add Kconfig to enable and disable LCD image show during bootup

namanjain7:
This commit fix build time error if logo is not enabled


os/drivers/lcd/mipi_lcd.c
```diff
@@ -436,6 +436,7 @@ FAR void lcd_put_logo(FAR struct lcd_dev_s *dev)
 	 * then filling it with logo data on specific index in while loop */
 	memset(lcd_logo_fullscreen_data, LCD_BLACK_VAL, CONFIG_LCD_XRES * CONFIG_LCD_YRES * 2);
 
+#if defined(CONFIG_LCD_LOGO)
```
sunghan-chang, line 439: The function name is `lcd_put_logo`. But the conditional is added only here. 
Is that to print black screen when log is disabled?
If yes, how about renaming the function and the variables like lcd_init_xxxx?

sunghan-chang: lcd put(or set0 initial screen function prints
1. logo with the logo config enabled
2. black screen with logo config disabled.

I think this makes sense in name (You already make that in operation, I mean in name)

os/drivers/lcd/lcd_logo.h
```diff
@@ -25,6 +25,9 @@
 #ifdef CONFIG_LCD_LOGO_52_340
 #define LOGO_XRES 52
 #define LOGO_YRES 340
+#else
+#define LOGO_XRES 52
+#define LOGO_YRES 340
```
sunghan-chang, line 30: The same values look strange. if we set them with all 0 in this conditional, what happen?

namanjain7: if logo_xre and logo_yres is set to 0
then, it will bydefault show black screen
because it won't go inside this [while loop](https://github.com/Samsung/TizenRT/pull/6543/files#diff-0dcbbbfee32bf5989cf8e2ad4117f850d4a44a336017e1f0e7f7e4ef484d5e2fL439:~:text=*%202))%20%7B-,while%20(logo_arr_index%20%3C%20(LOGO_YRES%20*%20LOGO_XRES%20*%202))%20%7B,lcd_logo_fullscreen_data%5Blcd_data_index%5D%20%3D%20lcd_logo_raw_data%5Blogo_arr_index%2B%2B%5D%3B,-lcd_init_fullscreen_image%5Blcd_data_index)

I have changed it to 0.

namanjain7: so, by default it will be black screen if no logo is selected

os/drivers/lcd/mipi_lcd.c
namanjain7: ok, I have changed the name of this function.

sunghan-chang (approved):

PR #6545, opened 2024-12-02
arch/boardctl: Add printing board reset caller pid.

ewoodev:
To check who called to reset the board, add outputting the pid running the boardctl.

### DIFF
--- a/os/arch/boardctl.c
+++ b/os/arch/boardctl.c
@@ -62,6 +62,7 @@
 #include <stdint.h>
 #include <errno.h>
 #include <assert.h>
+#include <unistd.h>
 
 #include <tinyara/board.h>
 #include <tinyara/arch.h>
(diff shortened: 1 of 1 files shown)


os/arch/boardctl.c
```diff
@@ -150,7 +151,7 @@ int boardctl(unsigned int cmd, uintptr_t arg)
 		sched_lock();
 		/* Add 100ms delay for flushing another logs like printf. */
 		up_mdelay(100);
-		lldbg("Board will Reboot now.\n");
+		lldbg("Board will Reboot now. pid: %d\n", getpid());
```
sunghan-chang, line 154: Let's add some sleep before the board_sleep. I found many logs which do not print entire message.

sunghan-chang (approved):

PR #6549, opened 2024-12-03
os/arch/arm/src/amebasmart: remove PB22 for I2S2 MCLK

zhongnuo-tang:
1. Only pinmux MCLK if required
2. remove I2S2 MCLK since it is not required in pinmux table


zhongnuo-tang:
Tested with loadable psram+ bixby + PM, working well.

ewoodev (approved):

sunghan-chang (approved):

PR #6552, opened 2024-12-04
os/arch/arm/include/amebasmart: Correct DRAM Size to follow Memory config

ziyik:


sunghan-chang:
@ziyik Please leave description in the commit. What is wrong, why this is necessary, and something else

sunghan-chang (approved):

PR #6554, opened 2024-12-04
os/board/rtl8730e: Add touch power on off using gpio

anjana348:
Touch IC power on/off functionality using gpio pin setting


sunghan-chang:
@anjana348 This is merged in the PR #6579 . Let's close this.

PR #6488, opened 2024-11-07
Sync NDP drivers code

abhishek-samsung:
when "SYNTIANT_NDP_ERROR_DATA_REREAD" error occurs, zero is returned in the argument extract size as no bytes are extracted. However, in the next iteration, zero is passed as extract size, causing to extract the entire audio tank, hence corrupting the heap. This exactly matches the crash we observed as the heap corrupted (around 60KB) matches the size of the audio tank (60KB)

PR #6555, opened 2024-12-04
enable NDP aliveness thread and add recovery for NDP

abhishek-samsung:


os/drivers/audio/ndp120_voice.c
```diff
@@ -505,8 +509,20 @@ static int ndp120_enqueuebuffer(FAR struct audio_lowerhalf_s *dev, FAR struct ap
 			break;
 		}
 	}
+
+	if (ret != OK) {
```
Taejun-Kwon, line 513: hmm I think below logic is required.
```
static void *ndp120_app_device_health_check(void)
{
	if (ndp_120_died) {
		dev->alive = false;
		reset_reload;
	}
	...
}

void reset_reload()
{
	do_reset_reload;
	dev->alive = true;
}

int ndp120_extract_audio(struct ndp120_dev_s *dev, struct ap_buffer_s *apb)
{
	if (!dev->alive) {
		return SYNTIANT_NDP_ERROR_UNINIT;
	}
	...
}

static int ndp120_enqueuebuffer(FAR struct audio_lowerhalf_s *dev, FAR struct ap_buffer_s *apb)
{
	...
	int ret = ndp120_extract_audio(priv, apb);
	/* Regardless result, dequee from pendq */
		for (tmp = (sq_entry_t *)sq_peek(&priv->pendq); tmp; tmp = sq_next(tmp)) {
		if (tmp == (sq_entry_t *)apb) {
			sq_rem(tmp, &priv->pendq);
			audvdbg("found the apb to remove 0x%x\n", tmp);
			break;
		}
	}
	
	if (ret == SYNTIANT_NDP_ERROR_UNINIT) {
		priv->dev.upper(priv->dev.priv, AUDIO_CALLBACK_UNREACHABLE, NULL, OK);
	}
	...
}
```

abhishek-samsung: Yes, thank you. This makes more sense for our case. I have updated it.

Taejun-Kwon (dismissed):

os/drivers/ai-soc/ndp120/src/ndp120_api.c
```diff
@@ -984,7 +1009,7 @@ ndp120_app_device_health_check(void *d)
 			printf("Error: %d in check_firmware_aliveness\n", s);
 			goto out;
 		}
-		pm_sleep(5000);
+		pm_sleep(3000);
```
r-prabu, line 1012: What is this magic number about? Can we have config for this?

os/drivers/ai-soc/ndp120/src/ndp120_api.c
```diff
@@ -937,6 +937,8 @@ void ndp120_show_debug(int include_spi)
 	}
 }
 
+int ndp120_init(struct ndp120_dev_s *dev, bool reinit);
```
r-prabu, line 940: shouldn't this be at declaration section at the top?

os/drivers/ai-soc/ndp120/src/ndp120_api.c
```diff
@@ -1012,29 +1037,33 @@ int ndp120_init(struct ndp120_dev_s *dev)
...
-		auddbg("failed to initialize mb sync mutex variable\n");
-	}
+	if (!reinit) {
 
-	s = pthread_mutex_init(&dev->ndp_mutex_mcu_mb_in, NULL);
-	if (s) {
-		auddbg("failed to initialize mb in mutex variable\n");
-	}
+		s = pthread_mutex_init(&dev->ndp_mutex_mbsync, NULL);
+		if (s) {
```
r-prabu, line 1043: Is it fine to proceed with the initialization if there is failure in the middle?

os/drivers/ai-soc/ndp120/src/ndp120_api.c
abhishek-samsung: Done

os/drivers/ai-soc/ndp120/src/ndp120_api.c
abhishek-samsung: Done

os/drivers/ai-soc/ndp120/src/ndp120_api.c
abhishek-samsung: This is not related to this PR. You are right, initialization should fail incase mutex_init fails. Let me add it in a new PR along with the comments from #6488 .

PR #6556, opened 2024-12-05
os/pm: Fix board hang issue

gSahitya-samsung:
This commit resolves the board hang issue, which is occuring with AMP & PM.
The CPU1 hotplug code is not taking care of EXIT case in PM TIMEDWAKEUP
code. When PM decides to make board sleep and the wakeup time is less than
CONFIG_PM_SLEEP_ENTRY_WAIT_MS, then it simply exit from pm_idle without
making board sleep, but in this case CPU1 was already in HOTPLUG state.
Later when application thread in CPU0 tries to gate CPU1, then CPU0 gets
stuck in while loop waiting for CPU1 to gate but since CPU1 is
in HOTPLUG state, it won't able to change its state, which leads to hang
issue. This commit moves the hotplug code after PM TIMEDWAKEUP code
which ensure the hotplug lead to board sleep without any failure.

### DIFF
--- a/os/pm/pm_idle.c
+++ b/os/pm/pm_idle.c
@@ -106,7 +106,10 @@ void pm_idle(void)
 				/* If the CPU is just back from sleep, abort the sleep */
 				if (up_get_cpu_state(cpu) == CPU_WAKE_FROM_SLEEP) {
 					goto EXIT;
+				} else if (up_get_cpu_state(cpu) == CPU_HOTPLUG) {
+					continue;
 				}
+
 				/* Gate the cpu first, before checking which task it is handling */
 				if (!up_get_gating_flag_status(cpu)) {
 					up_set_gating_flag_status(cpu, 1);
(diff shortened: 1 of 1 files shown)


os/pm/pm_idle.c
```diff
@@ -165,6 +158,18 @@ void pm_idle(void)
...
+#ifdef CONFIG_SMP
+		/* Send signal to shutdown other cores here */
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			if (up_get_cpu_state(cpu) == CPU_RUNNING) {
+				up_cpu_hotplug(cpu);
+			}
+			/* Reset core gating status flag */
+			up_set_gating_flag_status(cpu, 0);
+			/* Check whether each of the cpu has entered hotplug */
+			while(up_get_cpu_state(cpu) != CPU_HOTPLUG);
```
sunghan-chang: `while (`
Let's add a space.

os/pm/pm_idle.c
```diff
@@ -165,6 +158,18 @@ void pm_idle(void)
...
 		}
+#endif
+#ifdef CONFIG_SMP
+		/* Send signal to shutdown other cores here */
+		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
+			if (up_get_cpu_state(cpu) == CPU_RUNNING) {
+				up_cpu_hotplug(cpu);
+			}
+			/* Reset core gating status flag */
+			up_set_gating_flag_status(cpu, 0);
```
sunghan-chang, line 166: How about using some definitions instead of 0 here and 1 at line 164?

r-prabu (approved):

samsung-singh (approved):
board hang issue is fixed in os / pm, this change is good to go.

ewoodev (approved):

sunghan-chang (approved):

PR #6559, opened 2024-12-05
os/arch/arm/src/amebasmart: add related macro on peripherals priv struct

zhongnuo-tang:
1. each priv struct should only be made available if it is enabled in menuconfig
2. it will return NULL when port is not enabled, error handling should be added in places calling intialization

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2c.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2c.c
@@ -295,7 +295,7 @@ static const struct i2c_ops_s amebasmart_i2c_ops = {
 };
 
 /* I2C device structures */
-
+#ifdef CONFIG_AMEBASMART_I2C0
 static const struct amebasmart_i2c_config_s amebasmart_i2c0_config = {
 	//.base = amebasmart_I2C0_BASE,
 	//.busy_idle = CONFIG_I2C0_BUSYIDLE,
@@ -330,7 +330,9 @@ static struct amebasmart_i2c_priv_s amebasmart_i2c0_priv = {
 	.flags = 0,
 	.status = 0
 };
+#endif
 
+#ifdef CONFIG_AMEBASMART_I2C1
 static const struct amebasmart_i2c_config_s amebasmart_i2c1_config = {
 	//.base = amebasmart_I2C1_BASE,
 	//.busy_idle = CONFIG_I2C1_BUSYIDLE,
@@ -365,7 +367,9 @@ static struct amebasmart_i2c_priv_s amebasmart_i2c1_priv = {
 	.flags = 0,
 	.status = 0
 };
+#endif
 
+#ifdef CONFIG_AMEBASMART_I2C2
 static const struct amebasmart_i2c_config_s amebasmart_i2c2_config = {
 	//.base = amebasmart_I2C2_BASE,
 	//.busy_idle = CONFIG_I2C2_BUSYIDLE,
@@ -395,6 +399,7 @@ static struct amebasmart_i2c_priv_s amebasmart_i2c2_priv = {
 	.flags = 0,
 	.status = 0
 };
+#endif
 /************************************************************************************
  * Private Functions
  ************************************************************************************/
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1159,11 +1164,17 @@ FAR struct i2c_dev_s *up_i2cinitialize(int port)
...
+#ifdef CONFIG_AMEBASMART_I2C1
 		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c1_priv;
+#endif
 	} else if (port == 2) {
+#ifdef CONFIG_AMEBASMART_I2C2
 		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c2_priv;
+#endif
 	} else {
 		printf("Please select I2CO, I2C1 or I2C2 bus\n");
 		return NULL;
```
sunghan-chang, line 1180: You should check this block again.
If we disable I2C1 and call port 1 in upper layer, the `priv` is null by line 1162.
Then, line 1184 does null dereference. (priv->i2c_object)

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -1159,11 +1164,17 @@ FAR struct i2c_dev_s *up_i2cinitialize(int port)
...
 		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c0_priv;
+#endif
 	} else if (port == 1) {
+#ifdef CONFIG_AMEBASMART_I2C1
 		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c1_priv;
+#endif
 	} else if (port == 2) {
+#ifdef CONFIG_AMEBASMART_I2C2
 		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c2_priv;
+#endif
```
abhishek-samsung, line 1177: May be something like below should be done to check only ports enabled by config.
```

#ifdef CONFIG_AMEBASMART_I2C0
	if (port == 0) {
		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c0_priv;
	} else
#endif
#ifdef CONFIG_AMEBASMART_I2C1
	if (port == 1) {
		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c1_priv;
	} else 
#endif
#ifdef CONFIG_AMEBASMART_I2C2
	if (port == 2) {
		priv = (struct amebasmart_i2c_priv_s *)&amebasmart_i2c2_priv;
	} else
#endif
	{
		printf("Please select I2CO, I2C1 or I2C2 bus\n");
		return NULL;
```

os/arch/arm/src/amebasmart/amebasmart_spi.c
```diff
@@ -1383,9 +1386,9 @@ FAR struct spi_dev_s *amebasmart_spibus_initialize(int bus)
 	
 	if (bus == 0) {
 		/* Select SPI0 */
-
+#ifdef CONFIG_AMEBASMART_SPI0
 		priv = &g_spi0dev;
```
ziyik, line 1390: This change will cause hardfault in the later line 1395 or 1397.
Maybe it should return null here?

zhongnuo-tang:
Instead of debugassert to reboot, it is changed to return NULL, and null check should be added during initialization for error handling.

Please select I2CO, I2C1 or I2C2 bus
[RTK] Link callback handles: registered
Please select I2CO, I2C1 or I2C2 bus
board_initialize: ALC1019 initialization failed

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1785,11 +1787,17 @@ struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port, bool is_reinit)
...
+#endif
+#ifdef CONFIG_AMEBASMART_I2S3
+	if (port == I2S_NUM_3) {
 		hw_config_s = (struct amebasmart_i2s_config_s *)&amebasmart_i2s3_config;
-	} else {
+	} else 
+#endif
+	{
 		i2serr("Please select I2S2 or I2S3 bus\n");
 		return NULL;
```
sunghan-chang, line 1802: This looks strange. Is this function only for I2S2 and I2S3?
Don't you call this with I2S0 or I2S1?
If you do, this shows wrong error message.

zhongnuo-tang: As i check the history, 
Because I2S0/1 are for internal codec, that is why I2S2/3 are used. 
I2S2 for TX and I2S3 for RX. 
Error message will be added if I2S0/1 is used.

sunghan-chang (approved):

PR #6560, opened 2024-12-06
[Do not merge] os/pm: restructure pm for LCD changes

gSahitya-samsung:


os/pm/pm_domain_register.c
```diff
@@ -81,7 +81,7 @@ int pm_check_domain(int domain_id)
  *
  ****************************************************************************/
 
-int pm_domain_register(char *domain)
```
ewoodev, line 84: Could you please mrege pm_domain_register and pm_register

gSahitya-samsung: done

os/include/tinyara/pm/pm.h
```diff
@@ -208,60 +200,49 @@ struct pm_callback_s {
...
-	 *   method will also be recalled when reverting from lower back to higher
-	 *   power consumption modes (say because another driver refused a lower
-	 *   power state change).  Drivers are not permitted to return non-zero
-	 *   values when reverting back to higher power consumption modes!
+	 *   and will cause the state change to be aborted.
 	 *
 	 **************************************************************************/
 
-	int (*prepare)(FAR struct pm_callback_s *cb, enum pm_state_e pmstate);
+	int (*sleep)(FAR struct pm_callback_s *cb);
```
ewoodev, line 225: do we need prepare?

gSahitya-samsung: `prepare` is not needed in new design because we have `wake` callback now. If before sleep, a driver failed to change its state according to the new PM state, it will raise `ERROR` then PM will wake all the slept driver immediately. The drivers must support `wake` operation. The `prepare` is only needed if we want to separate prepare functionality from `sleep` callback. Considering current `NDP` and `Audio` driver code this separation is not needed.

PR #6562, opened 2024-12-06
os/include/tinyara/input/touchscreen.h: Add missing header

jylee9613:
There was a build error in UI like below.
```
include/tinyara/input/touchscreen.h:191:2: error: 'sem_t' does not name a type
  191 |  sem_t sem;
      |  ^~~~~
include/tinyara/input/touchscreen.h:194:2: error: 'sem_t' does not name a type
  194 |  sem_t pollsem;
```

Add missing header(semaphore.h) to fix build error.


jeongarmy (approved):
lgtm

PR #6563, opened 2024-12-06
os/board/rtl8730e: Workaround fix I2C hang issue during boot

ewoodev:
If you call GPIO_WriteBit right after calling I2c_write, there is an issue where the FIFO does not become empty after putting data in I2C FIFO. (I2C_BIT_TFE)

we have to check two point.
1. Why is there a problem with using I2C immediately after writing to GPIO?
2. When writing to I2C, do we have to put it in the FIFO and wait for the FIFO to be empty?

The changes is workarount adding delay


jeongarmy (approved):
lgtm

PR #6564, opened 2024-12-06
apps/examples/wakerec : Update wakerec to support record without KD

Taejun-Kwon:


apps/examples/wakerec/wakerec.cxx
```diff
@@ -327,34 +328,51 @@ extern "C" {
...
-		if (!gBuffer) {
-			printf("memory allocation failed\n");
+		printf("###################################\n");
+		printf("#### Wait for wakeup triggered ####\n");
+		printf("###################################\n");
+		
+		if (sd->getKeywordBufferSize(&bufferSize) == true) {
+			printf("KD buffer size %d\n", bufferSize);
+			gBuffer = new uint8_t[bufferSize];
+			if (!gBuffer) {
```
dalvinder-s, line 360: if !NULL only work when used like new(std::nothrow)

dalvinder-s (commented):
use new(sdt:nothrow)

apps/examples/wakerec/wakerec.cxx
Taejun-Kwon: nullptr is \0 so I think this is okay.

PR #6565, opened 2024-12-09
build/tools/amebasmart: Fix PSRAM Latency

ZhenBei-Sin:
Change Notes:
Root Cause: There is no backup latency for PSRAM Vendor A.
Possible Risk: There will be an issue when lpcnp PSRAM enter sleep mode.
Only lp can run normally and the psram memory will be unstable.
Solution: Regardless of the PSRAM vendor, the backup latency code for KM0 during sleep mode will also be implemented.


ZhenBei-Sin:
Hi @sunghan-chang , @ewoodev 

This change is not due to any reported issues, but because we internally discovered a bug.
We made the change in case we encounter any future instability with the PSRAM memory.

This is the latest bootloader version, which can be viewed from the bootup log.
bootloader_version km4_bootloader_ver_1d94f05280_2024/12/09-12:35:36

**Verification**:
loadable_apps config
pm_resume 0; pm_suspend

**Result**: 
Able bootup, Able perform pm test sleep and wake.

image

Thank You.

sunghan-chang:
@ewoodev @gSahitya-samsung @neel-samsung @ritesh55555 Could you check this?
@ZhenBei-Sin How can we know this is valid or not?

ZhenBei-Sin:
> How can we know this is valid or not?

Hi @sunghan-chang,
Whether this is valid or not...... (Apologies I cannot provide a 100% guarantee on this) but no issues have been reported so far from TizenRT.
However, it’s important to consider the potential risks involved.
While there seems to be no harm in including these changes, we can only confirm their validity by ensuring it do not affect the current boot and PM flow.

sunghan-chang (approved):

sunghan-chang:
@ZhenBei-Sin This requires the bootloader update. Right? I can't make sure that we can apply this in a real product...

ZhenBei-Sin:
> @ZhenBei-Sin This requires the bootloader update. Right? I can't make sure that we can apply this in a real product...

Yes bootloader updates. 
Don’t worry, I understand on your concern for real product.

PR #6566, opened 2024-12-10
os/board/rtl8721csm: improve one-shot adv CPU usage

yeetee179:
- adv data will not be updated if it is the same as the previous adv sent
- reduced CPU usage due to less updating of adv data


hs36-kim (approved):
LGTM
작동확인 되었고, 이 후 mem leak 발생 부분 삭제됨을 확인함

PR #6569, opened 2024-12-10
framework/Media: Add reset API for recorder and player.

s1-vivek:
This commit add reset API for recorder and player which will help application to do reset for recorder and player to IDLE state.


Taejun-Kwon (approved):

framework/include/media/MediaPlayer.h
```diff
@@ -134,6 +134,15 @@ class MediaPlayer
 	 */
 	player_result_t unprepare();
 
+	/**
+	 * @brief Reset allocated resources related to the player.
+	 * @details @b #include <media/MediaPlayer.h>
+	 * This function is a synchronous API
+	 * @return The result of the reset operation
+	 * @since TizenRT v2.0
```
aadotverma, line 142: Is this correct?
TizenRT v2.0

aadotverma (commented):
Code changes looks fine

framework/include/media/MediaPlayer.h
s1-vivek: Thanks.
Updated.

framework/src/media/MediaPlayerImpl.cpp
```diff
@@ -299,6 +299,29 @@ void MediaPlayerImpl::unpreparePlayer(player_result_t &ret)
...
+player_result_t MediaPlayerImpl::reset()
+{
+	LOG_STATE_INFO(mCurState);
+	meddbg("MediaPlayer reset mPlayer : %x\n", &mPlayer);
+	
+	if (mCurState == PLAYER_STATE_READY || mCurState == PLAYER_STATE_PLAYING || mCurState == PLAYER_STATE_PAUSED) {
+		mInputHandler.close();
+		if (reset_audio_stream_out(mStreamInfo->id) != AUDIO_MANAGER_SUCCESS) {
+			meddbg("MediaPlayer reset fail : reset_audio_stream_out fail\n");
+			return PLAYER_ERROR_INTERNAL_OPERATION_FAILED;
```
Taejun-Kwon, line 311: In my opinion, if another thread already use audio manager, it should keep going on release resource in media.
so I think some warning message is enough, we should not return error.

s1-vivek: Thanks.
Changes Done as requested.

PR #6570, opened 2024-12-10
[Do not merge]fs/mqueue/mq_check.c: group level error check for mq operations

ritesh55555:
This commit creates an api mq_check() for mqueue module. This api will check the mq descriptor value in the mq des list of the calling task's group. This will help to remove unwanted error in mq_send and mq_receive apis.


sunghan-chang:
@ritesh55555 @kishore-sn I want to check whether this improves stability even it causes performance degradation by checking single queue every send and receive or not.

os/kernel/mqueue/mq_receive.c
```diff
@@ -150,6 +150,11 @@ ssize_t mq_receive(mqd_t mqdes, FAR char *msg, size_t msglen, FAR int *prio)
 		return ERROR;
 	}
 
+	if (mq_check(mqdes) != OK) {
```
kishore-sn, line 153: Let us perfrom mq_check before mq_verify. It is done like this in nuttx also.

os/fs/mqueue/mq_check.c
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * Return Value:
+ *   OK - if mqdes is present in the calling task group list of mqdes
+ *   ERROR - if mqdes is not present in the calling task group of mqdes
+ *
+ ************************************************************************/
+
+int mq_check(mqd_t mqdes)
+{
+	int ret = ERROR;
```
kishore-sn, line 89: Let us change error code to -EBADF

os/kernel/mqueue/mq_send.c
```diff
@@ -154,6 +154,11 @@ int mq_send(mqd_t mqdes, FAR const char *msg, size_t msglen, int prio)
 		return ERROR;
 	}
 
+	if (mq_check(mqdes) != OK) {
+		leave_cancellation_point();
+		return ERROR;
+	}
+
```
kishore-sn, line 161: We need to check the mqdes in all mq functions like send, receive, getattr, setattr, notify, etc..

neel-samsung (commented):
Please check the github checks which are failing

kishore-sn:
> ue every send and receive

Hi @sunghan-chang, 
Do you have any suggestion on how to check stability? 
In my opinion, we can run all our test cases. But apart from that, I am not sure of any other method.

os/kernel/mqueue/mq_getattr.c
```diff
@@ -105,6 +105,10 @@ int mq_getattr(mqd_t mqdes, struct mq_attr *mq_stat)
 {
 	int ret = ERROR;
 
+	if (mq_check(mqdes) != OK) {
+		return ERROR;
```
kishore-sn, line 109: We need to set the errno before we return.

os/fs/mqueue/mq_check.c
```diff
@@ -0,0 +1,108 @@
...
+ * Parameters:
+ *   mqdes - Message queue descriptor
+ *
+ * Return Value:
+ *   OK - if mqdes is present in the calling task group list of mqdes
+ *   ERROR - if mqdes is not present in the calling task group of mqdes
+ *
+ ************************************************************************/
+
+int mq_check(mqd_t mqdes)
```
sunghan-chang, line 89: Let's rename the function to show "check what".

os/fs/mqueue/mq_check.c
```diff
@@ -0,0 +1,108 @@
...
+ * Description:
+ *   This function checks if a message queue descriptor is present in the
+ *   calling task's group list of mq des.
+ *
+ * Parameters:
+ *   mqdes - Message queue descriptor
+ *
+ * Return Value:
+ *   OK - if mqdes is present in the calling task group list of mqdes
+ *   ERROR - if mqdes is not present in the calling task group of mqdes
```
sunghan-chang, line 85: It does not return ERROR, it does EBADF.

os/include/mqueue.h
```diff
@@ -188,6 +188,13 @@ int mq_setattr(mqd_t mqdes, FAR const struct mq_attr *mq_stat, FAR struct mq_att
  * @since TizenRT v1.0
  */
 int mq_getattr(mqd_t mqdes, FAR struct mq_attr *mq_stat);
+/**
+ * @brief get message queue attributes
+ * @details @b #include <mqueue.h> \n
+ * POSIX API (refer to : http://pubs.opengroup.org/onlinepubs/9699919799/)
```
sunghan-chang, line 194: Is this the POSIX API?

os/include/mqueue.h
```diff
@@ -188,6 +188,13 @@ int mq_setattr(mqd_t mqdes, FAR const struct mq_attr *mq_stat, FAR struct mq_att
  * @since TizenRT v1.0
  */
 int mq_getattr(mqd_t mqdes, FAR struct mq_attr *mq_stat);
+/**
+ * @brief get message queue attributes
```
sunghan-chang, line 192: description looks wrong.

os/kernel/mqueue/mq_notify.c
```diff
@@ -161,6 +161,11 @@ int mq_notify(mqd_t mqdes, const struct sigevent *notification)
 		return ERROR;
 	}
 
+	if (mq_check(mqdes) != OK) {
+		set_errno(EBADF);
```
sunghan-chang, line 165: I think you should use the return value instead of hard coding here.

os/kernel/mqueue/mq_receive.c
```diff
@@ -145,6 +145,12 @@ ssize_t mq_receive(mqd_t mqdes, FAR char *msg, size_t msglen, FAR int *prio)
 	/* mq_receive() is a cancellation point */
 	(void)enter_cancellation_point();
 
+	if (mq_check(mqdes) != OK) {
+		leave_cancellation_point();
+		set_errno(EINVAL);
```
sunghan-chang, line 150: When it has EINVAL, when it has EBADF?

ritesh55555:
> @ritesh55555 @kishore-sn I thought when we close or unlink a mq, if we get wrong or invalid mqdes or mqname, it could corrupt mq inode. If we use invalid mqdes for mq send and receive, is it possible to be corrupted?

i dont think the inode will get corrupted if we use invalid mqdes for mq send and receive . But will lead to unintended behavior

os/fs/mqueue/mq_check.c
```diff
@@ -0,0 +1,108 @@
...
+ *
+ ************************************************************************/
+
+int mq_check(mqd_t mqdes)
+{
+	int ret = -EBADF;
+    	mqd_t mqdes_ptr;
+
+	FAR struct task_group_s *group = sched_self()->group;
```
jeongarmy, line 94: I think we need to check mqdes and group.
`DEBUGASSERT(mqdes != NULL && group != NULL);`

os/kernel/mqueue/mq_setattr.c
```diff
@@ -114,6 +114,10 @@ int mq_setattr(mqd_t mqdes, const struct mq_attr *mq_stat, struct mq_attr *oldst
 {
 	int ret = ERROR;
 
+	if (mq_check(mqdes) != OK) {
+		return ERROR;
```
jeongarmy, line 118: Set errno is missnig

os/include/mqueue.h
```diff
@@ -188,6 +188,13 @@ int mq_setattr(mqd_t mqdes, FAR const struct mq_attr *mq_stat, FAR struct mq_att
...
+ * @details @b #include <mqueue.h> \n
+ * POSIX API (refer to : http://pubs.opengroup.org/onlinepubs/9699919799/)
+ * @since TizenRT v1.0
+ */
+int mq_check(mqd_t mqdes);
```
abhishek-samsung, line 197: This looks like an internal api for mqueue, we need not expose it to outsize. Please check if we can we move this to some internal header file.

os/kernel/mqueue/mq_send.c
```diff
@@ -123,7 +123,8 @@
  *
  *   EAGAIN   The queue was empty, and the O_NONBLOCK flag was set for the
  *            message queue description referred to by mqdes.
- *   EINVAL   Either msg or mqdes is NULL or the value of prio is invalid.
+ *   EINVAL   Either msg or mqdes is NULL or the value of prio is invalid,
+ * 	 		  or mqdes is present in task group's mq list.
```
abhishek-samsung, line 127: I think this should be,
```suggestion
 * 	 		  or mqdes is not present in task group's mq list.
```

But here too, I think EBADF is better instead of EINVAL.

os/fs/mqueue/mq_check.c
```diff
@@ -0,0 +1,108 @@
...
+ *   OK - if mqdes is present in the calling task group list of mqdes
+ *   ERROR - if mqdes is not present in the calling task group of mqdes
+ *
+ ************************************************************************/
+
+int mq_check(mqd_t mqdes)
+{
+	int ret = -EBADF;
+    	mqd_t mqdes_ptr;
+
```
kishore-sn, line 93: We need to take care of interrupt handler case here. Mq can be used to send msg from irq handler to any task. In that case, mq may be opened by any task which calls the device driver. But when irq is being handled, sched_self() will return the task running at the time of irq. This task might be different than the task which opened the device driver and the mq. So, if we are in irq hanlder, then we must NOT perform this check.

os/kernel/mqueue/mq_send.c
```diff
@@ -149,6 +150,12 @@ int mq_send(mqd_t mqdes, FAR const char *msg, size_t msglen, int prio)
 	/* mq_send() is a cancellation point */
 	(void)enter_cancellation_point();
 
+	if (mq_check(mqdes) != OK) {
+		leave_cancellation_point();
+		set_errno(EINVAL);
+		return ERROR;
+	}
+
```
kishore-sn, line 158: This check needs to be done in all mq apis. Please check timedsend and timed receive.

PR #6571, opened 2024-12-11
os/board/rtl8730e: add timeout for i2c writebrk api

lhenry-realtek:
This commit addresses issue where i2c driver may get stuck in infinite loop while checking for status bit due to unstable voltage on SDA/SCL line caused by hardware

Added a counter-based timeout as watchdog cannot be used as interrupts are disabled before I2C operation happens

The workaround message for ist415 driver is replaced to indicate the root cause due to LDO soft startup voltage unstable

### DIFF
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
@@ -613,9 +613,11 @@ void I2C_MasterWrite(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 {
 	u32 cnt = 0;
+	u32 InTimeoutCount = 0;	/* timeout */
 
 	/* Write in the DR register the data to be sent */
 	for (cnt = 0; cnt < len; cnt++) {
+		InTimeoutCount = 1000;
 
 		while ((I2C_CheckFlagState(I2Cx, I2C_BIT_TFNF)) == 0);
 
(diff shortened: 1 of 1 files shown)


ewoodev (approved):

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -631,6 +633,14 @@ u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 				I2C_ClearAllINT(I2Cx);
 				return cnt;
 			}
+
+			DelayUs(1);
+
+			if (InTimeoutCount == 0) {
+				RTK_LOGW(TAG, "MasterWriteBrk_TimeOut\n");
+				return cnt;
```
ewoodev, line 641: If cnt resturn is 0, does it return error??

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -631,6 +633,14 @@ u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 				I2C_ClearAllINT(I2Cx);
 				return cnt;
 			}
+
+			DelayUs(1);
```
sunghan-chang, line 637: DelayUS is necessary?

sunghan-chang: `if not required`
I think you should confirm whether this is required or not.
Anyway delay causes performance degradation even it's small. If possible, I think removal is better.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -613,9 +613,11 @@ void I2C_MasterWrite(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 {
 	u32 cnt = 0;
+	u32 InTimeoutCount = 0;	/* timeout */
 
 	/* Write in the DR register the data to be sent */
 	for (cnt = 0; cnt < len; cnt++) {
+		InTimeoutCount = 1000; /* 1ms timeout */
 
 		while ((I2C_CheckFlagState(I2Cx, I2C_BIT_TFNF)) == 0);
```
anjana348, line 622: This while can also go in inifinite loop. shouldn't this also timeout?

sunghan-chang:
@lhenry-realtek Could you check my last comment and Anjana's comment?

sunghan-chang:
@lhenry-realtek Thank you for checking. But is this verified after removing delay?

sunghan-chang:
> i2c_uioregister

1. I think we should use `i2cerr` or `i2cwarn` instead of lldbg and RTK_LOGW.
2. But regardless of escape from while loop, this means the board is still unstable for i2c in HW?

sunghan-chang (approved):

abhishek-samsung:
> @abhishek-samsung Are you able to verify the PR with your audio device? If it works well, let's merge

Hello @sunghan-chang , I verified this PR with audio device. It works well.

sunghan-chang (approved):

PR #6537, opened 2024-11-26
build/tools/amebasmart/gnu_utility: refined frequency of sending probe response

ZivLow:
1. further reduced the frequency of sending probe response to increase chance of successful SoftAP connection in noisy environment

PR #6575, opened 2024-12-12
os/board/rtl8730e: limit probe response using silent table and timer

ZivLow:
- Improve success rate of SoftAP connection under very noisy environment by controlling number of probe responses sent
- If is a new STA MAC address not stored in silent table, send probe response
- If STA MAC address has already been recorded in silent table, do not send again until timer interval is reached
- Add option to enable and adjust silent table interval
- Silent table enabled by default


sunghan-chang:
@ZivLow Is this different with PR #6537 ?

ZivLow:
@sunghan-chang
Hi, both #6575 and #6537 have the same purpose of limiting probe responses. But, they use different algorithm.
This PR #6575 uses silent table and a timer to limit the number of probe responses sent. 
On the other hand, #6537 is cruder, and limits the number of probe responses sent by simply skipping probe responses.

Please use this PR #6575, instead of #6537.

hk-gwak:
Confirmed that issue does not happen on several tests.

hk-gwak (approved):

PR #6576, opened 2024-12-13
os/board/rtl8730e: Modify PinMode AILite

ZhenBei-Sin:
Change Notes:
The current pinmapcfg is based on the AILite modules (consolidated AILite, AILite+ and AILite+W).
Samsung requires modifications to the PuPd for both cases: Func PuPd and Sleep PuPd.
Refer to the UM0602_RTL8730E_pinmux_v1.3(85339)_241212.xls.

### DIFF
--- a/os/board/rtl8730e/src/component/usrcfg/rtl8730e/ameba_pinmapcfg.c
+++ b/os/board/rtl8730e/src/component/usrcfg/rtl8730e/ameba_pinmapcfg.c
@@ -27,78 +27,78 @@
 /* DSLP(RTC+KEYSCAN, TOUCH OFF): 3uA under power supply */
 /* DSLP(RTC+KEYSCAN+TOUCH): 3.9uA under power supply (peak is 3.7mA) */
 const PMAP_TypeDef pmap_func[] = {
-//Pin Name		Func PU/PD					Slp PU/PD
-	{_PA_0,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
-	{_PA_1,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
-	{_PA_2,			GPIO_PuPd_UP,			GPIO_PuPd_UP},			 //
-	{_PA_3,			GPIO_PuPd_UP,			GPIO_PuPd_UP},			 //
-	{_PA_4,			GPIO_PuPd_UP,			GPIO_PuPd_UP},			 //
-	{_PA_5,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
-	{_PA_6,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
-	{_PA_7,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
-	{_PA_8,			GPIO_PuPd_DOWN,			GPIO_PuPd_DOWN},			 //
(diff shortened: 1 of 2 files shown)


os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -239,10 +239,10 @@ void board_gpio_initialize(void)
 		u32 pinmode;
 		u32 pinpull;
 	} pins[] = {
-				{PA_23, PIN_INPUT, PullNone},
+				{PA_23, PIN_INPUT, PullDown},
```
sunghan-chang, line 242: 1. If rtl8730e_boot.c and ameba_pinmapcfg.c has different setting for the same pin, what happen?
2. This single commit has two changes, pin map configuration change and adding code comment. Due to this, it's difficult to see which pins are changed. Could you split a commit for adding code comment?

os/board/rtl8730e/src/component/usrcfg/rtl8730e/ameba_pinmapcfg.c
```diff
@@ -27,78 +27,78 @@
...
-	{_PB_31,			GPIO_PuPd_UP,				GPIO_PuPd_DOWN},			 //
-	{_PC_0,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PC_1,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PC_2,			GPIO_PuPd_UP,				GPIO_PuPd_DOWN},			 //flash clk pull down for PG or 50uA
-	{_PC_3,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PC_4,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PC_5,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PC_6,			GPIO_PuPd_UP,				GPIO_PuPd_UP},			 //
-	{_PNC,			GPIO_PuPd_KEEP,				GPIO_PuPd_KEEP},			 //table end
+	//Pin Name		Func PU/PD			Slp PU/PD
```
sunghan-chang, line 30: This file still shows every line is modified in the first commit. Could you leave real change in the first?

os/board/rtl8730e/src/rtl8730e_boot.c
ZhenBei-Sin: 1.
pinmap_init() is config all pins to the right function and pull state based on pmap_func table.
board_gpio_initialize() is intended to configure specific GPIO pins initialization for board-level functions.

The sequence pinmap_init() -> board_gpio_initialize().
If the pull-up/down configurations set by the two functions are different, usually the latter (board_gpio_initialize()) will override the former, so conflicts should be avoided.
If it is clear that a specific pin is used for GPIO and no other functionality is required, the pull-up/down configuration can be set directly in pinmap_init(), which can reduce unnecessary configurations and conflicts.

2. Done separated to two commits.


Additional Info:
From the email discussion: AIDual and AILite pinmapcfg is not compatible.
Did you have any macro to separate modules pinmap.
Will Samsung take back responsibility for the pinmapcfg management in the future?

sunghan-chang: Hello @ZhenBei-Sin .
For both, I suggest to manage all pin configurations in a single file regardless of pin types (gpio, i2c, i2s and etc).
It's difficult to know where it configures and can be missing.
If not, SS has no plan to do. We can't know where they are for each chipset.

PR #6578, opened 2024-12-13
os/pm: fix board hang issue

gSahitya-samsung:
This commit resolve hang issue of pm with amp. The issue happens when CPU0 gated the CPU1 and exited from critical section without undoing gating of CPU1. Later when tick interrupt came, the gated CPU0 tries to pause CPU1, but since CPU1 is in gating it can not release spinlock require by CPU0 to proceed CPU pause request. This creates deadlock which results in hang. To prevent this case we are setting gated_cpu_count just after performing gating and at EXIT case it will ensure the gated CPUs get ungated.

### DIFF
--- a/os/pm/pm_idle.c
+++ b/os/pm/pm_idle.c
@@ -99,6 +99,16 @@ void pm_idle(void)
 #ifdef CONFIG_PM_METRICS
 		pm_metrics_update_idle();
 #endif
+#ifdef CONFIG_PM_TIMEDWAKEUP
+		/* get wakeup timer */
+		if (newstate == PM_SLEEP) {
+			delay = wd_getwakeupdelay();
+			if ((delay > 0) && (delay < MSEC2TICK(CONFIG_PM_SLEEP_ENTRY_WAIT_MS))) {
+				pmvdbg("Wdog Timer Delay: %ldms is less than SLEEP_ENTRY_WAIT: %ldms\n", TICK2MSEC(delay), CONFIG_PM_SLEEP_ENTRY_WAIT_MS);
+				goto EXIT;
+			}
+		}
+#endif
 		/* Perform state-dependent logic here */
 		/* For SMP case, we need to check secondary core status
 	     * If secondary core status is not in idle thread, abort
@@ -118,12 +128,12 @@ void pm_idle(void)
 				if (!up_get_gating_flag_status(cpu)) {
 					up_set_gating_flag_status(cpu, 1);
 					up_cpu_gating(cpu);
+					gated_cpu_count = cpu;
 				}
 				while (up_get_gating_flag_status(cpu) == 1) {
 					/* If there is a pause request, we should handle it first */
 					if (up_cpu_pausereq(up_cpu_index())) {
-						pmdbg("Sleep abort! CPU%d task: %s!\n", cpu, tcb->name);
-						gated_cpu_count = cpu;
+						pmdbg("Sleep abort! CPU%d\n", cpu);
 						goto EXIT;
 					}
 				}
@@ -133,20 +143,9 @@ void pm_idle(void)
 				 * pause request on primary core 
 				 */
 				if (tcb->pid != cpu) {
-					pmdbg("Sleep abort! CPU%d task: %s!\n", cpu, tcb->name);
-					gated_cpu_count = cpu;
+					pmvdbg("Sleep abort! CPU%d task: %s!\n", cpu, tcb->name);
 					goto EXIT;
 				}
-			}
-		}
-#endif
-#ifdef CONFIG_PM_TIMEDWAKEUP
-		/* get wakeup timer */
-		if (newstate == PM_SLEEP) {
-			delay = wd_getwakeupdelay();
-			if ((delay > 0) && (delay < MSEC2TICK(CONFIG_PM_SLEEP_ENTRY_WAIT_MS))) {
-				pmvdbg("Wdog Timer Delay: %ldms is less than SLEEP_ENTRY_WAIT: %ldms\n", TICK2MSEC(delay), CONFIG_PM_SLEEP_ENTRY_WAIT_MS);
-				goto EXIT;
 			}
 		}
 #endif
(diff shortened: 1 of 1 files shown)


ewoodev (commented):
Could you please share how to reproduce this issue and teat result?

os/pm/pm_idle.c
```diff
@@ -98,6 +98,16 @@ void pm_idle(void)
 		newstate = pm_checkstate();
 #ifdef CONFIG_PM_METRICS
 		pm_metrics_update_idle();
+#endif
+#ifdef CONFIG_PM_TIMEDWAKEUP
+		/* get wakeup timer */
+		if (newstate == PM_SLEEP) {
```
r-prabu, line 104: is the placement/flow of the PM_TIMEDWAKEUP code matters to resolve the this hang case? If yes then what is the reason?

gSahitya-samsung: Yes, this placement will reduce the probability of `exiting out` from `pm_idle`. I believe the frequent `exit` cases should be kept at top to avoid going/recovering from unwanted state.

ewoodev (approved):

sunghan-chang:
@r-prabu @kishore-sn Once your team review is done, let's merge this.

r-prabu (dismissed):

os/pm/pm_idle.c
```diff
@@ -186,7 +186,9 @@ void pm_idle(void)
 #ifdef CONFIG_SMP
 	/* Check if any core is gated, resume it */
 	while (gated_cpu_count) {
-		up_set_gating_flag_status(gated_cpu_count, 0);
+		if (up_get_cpu_state(gated_cpu_count) == CPU_RUNNING) {
```
edwakuwaku, line 189: Is there other possible state that a gated core can obtain? If yes, there should be else statement to handle that condition. If not, we don't need to add this if condition.

os/pm/pm_idle.c
```diff
@@ -164,13 +163,14 @@ void pm_idle(void)
...
 		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
 			if (up_get_cpu_state(cpu) == CPU_RUNNING) {
+				/* Reset core gating status flag */
+				up_set_gating_flag_status(cpu, 0);
 				up_cpu_hotplug(cpu);
 			}
-			/* Reset core gating status flag */
-			up_set_gating_flag_status(cpu, 0);
 			/* Check whether each of the cpu has entered hotplug */
 			while(up_get_cpu_state(cpu) != CPU_HOTPLUG);
```
edwakuwaku, line 171: can you please help to correct the code here? 
`while (`

os/pm/pm_idle.c
```diff
@@ -164,13 +163,14 @@ void pm_idle(void)
 		/* Send signal to shutdown other cores here */
 		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
 			if (up_get_cpu_state(cpu) == CPU_RUNNING) {
+				/* Reset core gating status flag */
+				up_set_gating_flag_status(cpu, 0);
 				up_cpu_hotplug(cpu);
 			}
-			/* Reset core gating status flag */
-			up_set_gating_flag_status(cpu, 0);
```
edwakuwaku, line 170: You cannot move this line above hot plug command, the reason of such design is to ensure that the gated core can handle the pending hot plug request when it has been granted access to exit the gated condition.
If you restore the gated core first, it will be possible that a context switch happened immediately before the hot plug request is sent out and received by the target core(ie. due to timing issue), then the target core will enter hot plug state but it is not in idle thread.
Please revert this change.

os/pm/pm_idle.c
gSahitya-samsung: we are only doing gating at `CPU_RUNNING` state only. Before resuming CPU from gating, we should check it states is intact i.e. `CPU_RUNNING` state. Yes, we should handle else condition also.

os/pm/pm_idle.c
gSahitya-samsung: `pm_idle` is inside `critical_section`. The `tick interrupts` are disabled at this time for all CPU cores. The context switch won't happen. Also, can you confirm, do we really need to gate CPUs inside critical section? Because all CPUs will not change their context inside critical section, and we are doing gating to ensure CPU did not change context before we check all CPUs are idle and ready to put into sleep.

edwakuwaku: Critical section only disable interrupt as per each specific cores. pm_idle will only be executed by primary core. The generic clock for CA32 is disabled in the BSP layer which is invoked in the last step of pm_idle. Meaning, secondary cores are still possible to happen context switch when they exit from gated condition.
I don't understand your intention of swapping the sequence, but it shouldn't be allowed.

os/pm/pm_idle.c
gSahitya-samsung: Done

os/pm/pm_idle.c
gSahitya-samsung: done

os/pm/pm_idle.c
gSahitya-samsung: done

edwakuwaku (approved):
Good to merge.

ewoodev:
I think that the reason why SMP PM code has a complex structure is because it cannot turn on a CPU that has already been hot-plugged in a failed state.

Since entering hotplus of secondery core is here, but exiting hotplug is in bsp wakeup code.
```
void SOCPS_SleepPG(void)
#if ( CONFIG_SMP_NCPUS > 1 )
#ifdef CONFIG_CPU_HOTPLUG
	/* Now init core1 */
	up_set_cpu_state(1, CPU_WAKE_FROM_SLEEP);
#endif
	smp_init();
#endif
```

we should get controllability which exiting cpu hotplug also in os side
we need to improve this pm enter sleep structure.
Could you please make sequence and consider calling order?
- disable irq
- suspend device
- freezing(disable) secondary cpus
- enter board sleep in bsp side
- wake handler (it also doesn't need in excuting in bsp side)
- enable secondary cpus
- resume device
- enable irq
 
For this, Please raise additional PRs

kishore-sn (approved):

r-prabu (approved):

PR #6585, opened 2024-12-17
os/include: resolve build warnings

gSahitya-samsung:
Issue: Getting below build warnings

```
include/signal.h:590:83: WARNING: 'int sigaction(int, const sigaction*, sigaction*)' hides constructor for 'struct sigaction' [-Wshadow]

  590 | int sigaction(int sig, FAR const struct sigaction *act, FAR struct sigaction *oact);

include/stdlib.h:603:30: WARNING: 'mallinfo mallinfo()' hides constructor for 'struct mallinfo' [-Wshadow]

  603 | struct mallinfo mallinfo(void);

```

Cause: the struct `sigaction` and function `sigaction` is declared with same
name. During compilation they were part of same  namespace that's
causing shadowing of struct with function declaration. Same is true for
`mallinfo`

Resolution: Disable shadow warning during function declaration and
enable it after declaration.

Note: GCC>=7 doesn't throw this error with -Wshadow flag. This
temporary changes is only needed to support build with old version of
GCC.

### DIFF
--- a/os/include/signal.h
+++ b/os/include/signal.h
@@ -578,6 +578,18 @@ CODE void (*signal(int sig, CODE void (*func)(int sig)))(int sig);
 /**
  * @}
  */
+/*
+	Following changes are work around to resolve build warnings.
+
+	TODO : The build warnings are there because function `sigaction` shadows struct `sigaction`.
+
+	The proper fix for warning issue is below:
+
+		1. Either change struct or function name
+		2. Move functions declaration into separate header file
+*/
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshadow"
 
 /**
  * @ingroup SIGNAL_KERNEL
@@ -588,6 +600,9 @@ CODE void (*signal(int sig, CODE void (*func)(int sig)))(int sig);
  * @since TizenRT v1.0
  */
 int sigaction(int sig, FAR const struct sigaction *act, FAR struct sigaction *oact);
+
+#pragma GCC diagnostic pop
+
 /**
  * @ingroup SIGNAL_KERNEL
  * @brief examine and change blocked signals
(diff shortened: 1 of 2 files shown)


sunghan-chang (approved):

ritesh55555 (approved):
build warnings resolved

abhishek-samsung (commented):
This looks like a work around. Is it possible that these errors are occuring because we include c headers in cpp files? Will it be possible to resolve them by adding extern C at appropriate places?

gSahitya-samsung:
> This looks like a work around. Is it possible that these errors are occuring because we include c headers in cpp files? Will it be possible to resolve them by adding extern C at appropriate places?

Yes, its work around

The proper fix can be as follow:

1. Either change struct or function name
2. Move functions declaration into separate header file

But in our case, none of above will work.

Can we change `-Wshadow` to `-Wshadow=compatible-local`? It will also resolve the warnings message.

Reference - https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

r-prabu:
> This looks like a work around. Is it possible that these errors are occuring because we include c headers in cpp files? Will it be possible to resolve them by adding extern C at appropriate places?

@abhishek-samsung i think it is not because of cpp files. Also this is applied only for the few required lines which should not create a problem.

r-prabu (dismissed):

os/include/signal.h
```diff
@@ -579,6 +579,9 @@ CODE void (*signal(int sig, CODE void (*func)(int sig)))(int sig);
  * @}
  */
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshadow"
+
```
kishore-sn, line 584: If we are using this approach, then pls add comments indicating why we are doing this.

kishore-sn:
> Can we change -Wshadow to -Wshadow=compatible-local? It will also resolve the warnings message.

I think this might be a better option. Please try it out.

os/include/signal.h
gSahitya-samsung: Done

kishore-sn (approved):

PR #6586, opened 2024-12-17
os/board/rtl8730e: add connection parameter update for BLE

yeetee179:
- add function to update BLE connection parameter for slave and master


hs36-kim (approved):
LGTM

PR #6587, opened 2024-12-17
Change touch event structure

ewoodev:


os/drivers/input/touchscreen.c
```diff
@@ -183,23 +196,60 @@ static int touch_close(FAR struct file *filep)
...
 		return -EINVAL;
 	}
+
+	DEBUGASSERT(dev->tp_buf.buffer);
+
+	tp_buf = &dev->tp_buf;
+
 	/* Wait for semaphore to prevent concurrent reads */
-	touch_semtake(&priv->sem, false);
+	touch_semtake(&tp_buf->sem, false);
```
Taejun-Kwon, line 216: you should consider below.
1. Enable / disable condition (open -> ioctl(TSIOC_DISABLE) -> read case)
2. if user increased size of ringbuffer and it fully filled with touch event, but app request stop in the middle of  read event. 
3. How app can realize proper buflen? (actually this also violate posix, it should return even 1byte.)
4. In posix, write -> read then written value must be readed. Current code violate posix
5. Someday if we apply micro kernel fully, because of differrent address logic in app & kernel, this will not work properly because address of buffer can't be relalized in kernel.

apps/examples/touchscreen/touchscreen_main.c
```diff
@@ -42,31 +42,34 @@ static void touch_test(void)
...
 	while (!g_terminated) {
 		poll(fds, 1, 5000);
 		if (fds[0].revents & POLLIN) {
-			ret = read(fd, &buf, sizeof(struct touch_sample_s));
-			if (ret != - 1) {
-				printf("Total touch points %d\n", buf.npoints);
-				for (int i = 0; i < buf.npoints; i++) {
-					printf("coordinates id: %d, x : %d y : %d touch type: %d\n", buf.point[i].id, buf.point[i].x, buf.point[i].y, buf.point[i].flags);
-					if (buf.point[i].flags == TOUCH_DOWN) {
+			ret = read(fd, buf, sizeof(struct touch_point_s) * 15);
```
Taejun-Kwon, line 60: I think simply you can use pipe.

os/drivers/input/touchscreen.c
ewoodev: Thank you for review.

we need to remove ioctl(TSIOC_DISABLE), I think there is nothing for the application to do in order to read touch data.

Could you please explain more about `In posix, write -> read then written value must be readed. Current code violate posix`?

we are using write read in serial.c also.
is the problem related to no code in write?

anjana348: TSIOC_DISABLE is added for suspend case as was asked that interrupt needs to be disabled. 
App need to ensure to call TSIOC_DISABLE and TSIOC_ENABLE when going to suspend and resume. doesn't it?

PR #6588, opened 2024-12-18
os/pm: clean and improve pm code

gSahitya-samsung:


apps/system/utils/pmcmd.c
```diff
@@ -0,0 +1,242 @@
...
+#endif
+	{"pm_sleep_test", tash_pm_sleep_test, TASH_EXECMD_ASYNC},
+	{NULL           , NULL              , 0               }
+};
+
+/****************************************************************************
+ * Public Functions
+ ****************************************************************************/
+
+void pm_register_utilcmds(void)
```
ewoodev, line 234: Let's merge into power test

os/pm/pm.h
```diff
@@ -140,6 +140,22 @@ struct pm_global_s {
...
+	PM_MET_IDLE,
+	PM_MET_CHANGESTATE,
+	PM_MET_WAKE,
+} pm_metrics_update_e;
+
+typedef struct {
+	clock_t ticks;
+	pm_wakeup_reason_code_t wakeup_src;
+} pm_metrics_wake_t;
+#endif
```
ewoodev, line 157: How about use va_arg?

PR #6591, opened 2024-12-19
os/syscall, os/include, os/kernel: Remove SMP check for getcpu syscall

kishore-sn:
Remove check on CONFIG_SMP for getcpu syscall. If smp is not enabled then this syscall will return zero value.


sunghan-chang (approved):

sunghan-chang:
@kishore-sn @r-prabu This is what I asked before. Let's apply the same in all other syscall for system.

sunghan-chang:
@r-prabu Let's merge if your review is done.

neel-samsung (approved):

r-prabu (commented):
changes for Remove SMP check for getcpu syscall  looks good.

PR #6592, opened 2024-12-19
os/board/rtl8730e: add check for BLE scanned packet size

yeetee179:
- with extended scan, packet size may exceed 31
- limit the packet size to 31 to accomodate for legacy packets only


giwon-nam (approved):
During the one hour scan test, no abort occurred.

hs36-kim (approved):
LGTM

PR #6597, opened 2024-12-24
Remove global structure in LCD HAL Layer

abhinav-s235:
Replaced the global LCD HAL layer instance with a local declaration within the lcddev_register() function to reduce memory usage and passed this instance to LCD_FLUSH_THREAD by converting its address to a string using atoi() function.

### DIFF
--- a/os/drivers/lcd/lcd_dev.c
+++ b/os/drivers/lcd/lcd_dev.c
@@ -93,7 +93,6 @@ struct lcd_s {
 	bool do_wait;
 #endif
 };
-static struct lcd_s *lcd_info;
 /****************************************************************************
  * Private Function Prototypes
  ****************************************************************************/
@@ -376,10 +375,12 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 }
 
 #if defined(CONFIG_LCD_FLUSH_THREAD)
-static void lcd_flushing_thread(void)
+static void lcd_flushing_thread(int argc, char **argv)
 {
 	int ret;
+	DEBUGASSERT(argc == 2);
 	FAR struct lcddev_area_s *lcd_area;
+	struct lcd_s *lcd_info = (struct lcd_s *)strtoul(argv[1], NULL, 16);
 	lcd_area = &(lcd_info->lcd_area);
 	while (true) {
 		while (sem_wait(&lcd_info->flushing_sem) != 0) {
@@ -423,6 +424,12 @@ int lcddev_register(struct lcd_dev_s *dev)
 {
 	char devname[16] = { 0, };
 	int ret;
+	struct lcd_s *lcd_info;
+#if defined(CONFIG_LCD_FLUSH_THREAD)
+	int pid;
+	char *flushing_thread_args[2];
+	char lcd_info_addr[9]; /* for storing 32 bit address */
+#endif
 
 	if (!dev) {
 		return -EINVAL;
@@ -441,12 +448,19 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_info->empty = true;
 	lcd_info->lcd_kbuffer = (uint8_t *)kmm_malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * sizeof(uint16_t));
 	if (!lcd_info->lcd_kbuffer) {
+		sem_destroy(&lcd_info->flushing_sem);
+		kmm_free(lcd_info);
 		lcddbg("ERROR: Failed to allocate memory for LCD flush swap buffer\n");
 		return -ENOMEM;
 	}
-
-	int pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, NULL);
+	itoa((int)lcd_info, lcd_info_addr, 16);
+	flushing_thread_args[0] = lcd_info_addr;
+	flushing_thread_args[1] = NULL;
+	pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, (FAR char *const *)flushing_thread_args);
 	if (pid < 0) {
+		kmm_free(lcd_info->lcd_kbuffer);
+		sem_destroy(&lcd_info->flushing_sem);
+		kmm_free(lcd_info);
 		lcddbg("ERROR: Failed to start LCD Frame Flusing thread\n");
 		return -ENOMEM;
 	}
@@ -458,23 +472,34 @@ int lcddev_register(struct lcd_dev_s *dev)
 #endif
 
 	lcd_init_put_image(dev);
+	sem_init(&lcd_info->sem, 0, 1);
 	if (dev->setpower) {
 		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
 		if (ret != OK) {
-			return ret;
+			goto cleanup;
 		}
 #ifdef CONFIG_PM
 		(void)pm_suspend(lcd_info->pm_domain);
 #endif
 	}
-	sem_init(&lcd_info->sem, 0, 1);
 	if (lcd_info->dev->getplaneinfo) {
 		lcd_info->dev->getplaneinfo(lcd_info->dev, 0, &lcd_info->planeinfo);	//plane no is taken 0 here
 		snprintf(devname, 16, "/dev/lcd0");
-		return register_driver(devname, &g_lcddev_fops, 0666, lcd_info);
-	} else {
-		sem_destroy(&lcd_info->sem);
-		kmm_free(lcd_info);
-		return -ENOSYS;
+		ret = register_driver(devname, &g_lcddev_fops, 0666, lcd_info);
+		if (ret != OK) {
+			goto cleanup;
+		} else {
+			return ret; //successful registration of driver
+		}
 	}
+cleanup:
+	lcddbg("ERROR: Failed to register driver %s\n", devname);
+#if defined(CONFIG_LCD_FLUSH_THREAD)
+	task_delete(pid);
+	sem_destroy(&lcd_info->flushing_sem);
+	kmm_free(lcd_info->lcd_kbuffer);
+#endif
+	sem_destroy(&lcd_info->sem);
+	kmm_free(lcd_info);
+	return -ENOSYS;
 }


os/drivers/lcd/lcd_dev.c
```diff
@@ -444,8 +447,10 @@ int lcddev_register(struct lcd_dev_s *dev)
 		lcddbg("ERROR: Failed to allocate memory for LCD flush swap buffer\n");
 		return -ENOMEM;
 	}
-
-	int pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, NULL);
+	itoa((int)lcd_info, parm_buf, 16);
+	parm[0] = parm_buf;
```
jeongarmy, line 451: `param[0] = itoa((int)lcd_info, parm_buf, 16);`

And don't we need to memset parm_buf?

os/drivers/lcd/lcd_dev.c
```diff
@@ -423,13 +424,15 @@ int lcddev_register(struct lcd_dev_s *dev)
 {
 	char devname[16] = { 0, };
 	int ret;
+	char *parm[2];
+	char parm_buf[9]; /* for storing 32 bit address */
```
sunghan-chang, line 428: The param and param_buf are difficult to distinguish and to know what they are. Could you rename them to know what they are in name?

os/drivers/lcd/lcd_dev.c
sunghan-chang: I don't know why we need to convert it to ascii and to reconvert to integer.
Is it possible to give the address itself without converting?

sunghan-chang (commented):
Could you leave why you modify this in the COMMIT description? I can't get why you did.

os/drivers/lcd/lcd_dev.c
abhinav-s235: It is not possible to pass the address without converting it to string because kernel_thread() function accept the parameter of type FAR char *const argv[].
**int kernel_thread(FAR const char *name, int priority, int stack_size, main_t entry, FAR char *const argv[])**

abhinav-s235: > `param[0] = itoa((int)lcd_info, parm_buf, 16);`
> 
> And don't we need to memset parm_buf?

I don't think it is needed
could you please explain why you think so?

os/drivers/lcd/lcd_dev.c
abhinav-s235: > The param and param_buf are difficult to distinguish and to know what they are. Could you rename them to know what they are in name?

fixed

sunghan-chang:
> Could you leave why you modify this in the COMMIT description? I can't get why you did.

@abhinav-s235 Could you check my comment above? I mean why you did below
```
Replaced the global LCD HAL layer instance with a local declaration within the lcddev_register function
and passed this instance to LCD_FLUSH_THREAD by converting its address to a string using atoi() function.
```

abhinav-s235:
> Could you leave why you modify this in the COMMIT description? I can't get why you did.

Previously, while creating frame flushing thread we were not passing [lcd_info](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L435C2-L435C9) pointer to function [lcd_flushing_thread()](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L378C13-L378C32) that's why we declared it globally but now we are passing it as a parameter to the lcd_flushing_thread() function there is no need to globally declare it.

os/drivers/lcd/lcd_dev.c
```diff
@@ -444,8 +447,10 @@ int lcddev_register(struct lcd_dev_s *dev)
 		lcddbg("ERROR: Failed to allocate memory for LCD flush swap buffer\n");
 		return -ENOMEM;
 	}
-
-	int pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, NULL);
+	itoa((int)lcd_info, lcd_info_addr, 16);
+	task_info[0] = lcd_info_addr;
```
sunghan-chang, line 451: The `lcd_info_addr` is a local pointer variable. Is it ok in lifecycle? When the flushing thread gets the arg, is it still valid?

os/drivers/lcd/lcd_dev.c
```diff
@@ -423,13 +424,15 @@ int lcddev_register(struct lcd_dev_s *dev)
 {
 	char devname[16] = { 0, };
 	int ret;
+	char *task_info[2];
```
sunghan-chang, line 427: How about `flushing_thread_args`?

sunghan-chang:
> > Could you leave why you modify this in the COMMIT description? I can't get why you did.
> 
> Previously, while creating frame flushing thread we were not passing [lcd_info](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L435C2-L435C9) pointer to function [lcd_flushing_thread()](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L378C13-L378C32) that's why we declared it globally but now we are passing it as a parameter to the lcd_flushing_thread() function there is no need to globally declare it.

The point is you want to remove the global variable. Right? Is that to reduce memory usages?
And Let's leave it at the COMMIT description.

os/drivers/lcd/lcd_dev.c
anjana348: @sunghan-chang - Yes, it can happen that thread executes after scope of lcd_info_addr is finished. 
But to encounter this, if we use dynamic(malloc) lcd_info_addr, in thread function, the passed argument's address changes so we can't free it.
So should we keep static struct *lcd_info as global only ?

sunghan-chang: I checked the function - thread_create.
It looks it's ok with local variable in lifecycle.

```
	/* Copy the argv[] array and all strings are to the task's stack.  Data on
	 * the stack is guaranteed to be accessible by the ask no matter what
	 * privilege mode the task runs in.
	 */

	return task_stackargsetup(tcb, argv);
```

os/drivers/lcd/lcd_dev.c
sunghan-chang: @abhinav-s235 Could you check my comment above?

abhinav-s235:
> > > Could you leave why you modify this in the COMMIT description? I can't get why you did.
> > 
> > 
> > Previously, while creating frame flushing thread we were not passing [lcd_info](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L435C2-L435C9) pointer to function [lcd_flushing_thread()](https://github.com/Samsung/TizenRT/blob/0229ed0f928dd0671bfce6fbce65ef2d480d75f2/os/drivers/lcd/lcd_dev.c#L378C13-L378C32) that's why we declared it globally but now we are passing it as a parameter to the lcd_flushing_thread() function there is no need to globally declare it.
> 
> The point is you want to remove the global variable. Right? Is that to reduce memory usages? And Let's leave it at the COMMIT description.

Commit description has been changed.

os/drivers/lcd/lcd_dev.c
abhinav-s235: > How about `flushing_thread_args`?

It has been changed

os/drivers/lcd/lcd_dev.c
```diff
@@ -458,23 +468,34 @@ int lcddev_register(struct lcd_dev_s *dev)
...
-		return -ENOSYS;
+		ret = register_driver(devname, &g_lcddev_fops, 0666, lcd_info);
+		if (ret != OK) {
+			goto cleanup;
+		} else {
+			return OK; //successful registration of driver
+		}
 	}
+cleanup:
+lcddbg("ERROR: Failed to register driver %s\n", devname);
```
anjana348, line 492: indentation should be tab

abhinav-s235: Fixed

sunghan-chang:
@abhinav-s235 Is this verified with real device?

sunghan-chang (approved):

abhinav-s235:
> @abhinav-s235 Is this verified with real device?

Yes

os/drivers/lcd/lcd_dev.c
```diff
@@ -441,12 +444,19 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+		kmm_free(lcd_info);
 		lcddbg("ERROR: Failed to allocate memory for LCD flush swap buffer\n");
 		return -ENOMEM;
 	}
-
-	int pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, NULL);
+	itoa((int)lcd_info, lcd_info_addr, 16);
+	flushing_thread_args[0] = lcd_info_addr;
+	flushing_thread_args[1] = NULL;
+	int pid = kernel_thread("LCD Frame flusing", 204, 8192, lcd_flushing_thread, (FAR char *const *)flushing_thread_args);
```
ewoodev, line 455: It also, Let's move variable declaration on top

abhinav-s235: fixed

os/drivers/lcd/lcd_dev.c
```diff
@@ -423,6 +424,12 @@ int lcddev_register(struct lcd_dev_s *dev)
 {
 	char devname[16] = { 0, };
 	int ret;
+#if defined(CONFIG_LCD_FLUSH_THREAD)
+	int pid;
+#endif
+	struct lcd_s *lcd_info;
+	char *flushing_thread_args[2];
+	char lcd_info_addr[9]; /* for storing 32 bit address */
```
sunghan-chang, line 432: This should be in the `CONFIG_LCD_FLUSH_THREAD` conditional. 
If not, when we disable the config, it will be unused variable warning.

os/drivers/lcd/lcd_dev.c
```diff
@@ -441,12 +448,19 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_info->empty = true;
 	lcd_info->lcd_kbuffer = (uint8_t *)kmm_malloc(CONFIG_LCD_XRES * CONFIG_LCD_YRES * sizeof(uint16_t));
```
sunghan-chang, line 449: The members - do_wait, empty and lcd_kbuffer are valid with the config - CONFIG_LCD_FLUSH_THREAD only?
When the config is disable, they are not used?

os/drivers/lcd/lcd_dev.c
abhinav-s235: Fixed

os/drivers/lcd/lcd_dev.c
abhinav-s235: Yes, these members are only valid when config CONFIG_LCD_FLUSH_THREAD is enabled
[struct lcd_s](https://github.com/Samsung/TizenRT/blob/c0638bbb47ccb167b92aef3392a5ce412b4c6ce1/os/drivers/lcd/lcd_dev.c#L88)

sunghan-chang (approved):

PR #6598, opened 2024-12-25
os/arch/armv7-a: new line to distinguish between normal and crash logs

ritesh55555:
This commit introduces a new line code in assert logs so that user can distinguish between normal logs and crash logs.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_dataabort.c
+++ b/os/arch/arm/src/armv7-a/arm_dataabort.c
@@ -78,6 +78,8 @@ uint32_t system_exception_location;
 
 static inline void print_dataabort_detail(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
 {
+	/* Abort log must always start at a new line.*/
+	lldbg_noarg("\n");
 	_alert("#########################################################################\n");
 	_alert("PANIC!!! Data Abort at instruction : 0x%08x\n",  regs[REG_PC]);
 	_alert("PC: %08x DFAR: %08x DFSR: %08x\n", regs[REG_PC], dfar, dfsr);
--- a/os/arch/arm/src/armv7-a/arm_prefetchabort.c
+++ b/os/arch/arm/src/armv7-a/arm_prefetchabort.c
@@ -76,6 +76,8 @@ extern uint32_t system_exception_location;
 
 static inline void print_prefetchabort_detail(uint32_t *regs, uint32_t ifar, uint32_t ifsr)
 {
+	/* Abort log must always start at a new line.*/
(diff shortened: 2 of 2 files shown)


os/arch/arm/src/armv7-a/arm_prefetchabort.c
```diff
@@ -152,6 +152,8 @@ uint32_t *arm_prefetchabort(uint32_t *regs, uint32_t ifar, uint32_t ifsr)
 		CURRENT_REGS = savestate;
 	} else {
 		if (!IS_SECURE_STATE()) {
+			/* Add new line to distinguish between normal log and assert log.*/
+			lldbg_noarg("\n");
```
sunghan-chang, line 156: How about adding `\n` at the start of the first message in the print_prefecthabort_detail?
I mean line 79.

os/arch/arm/src/armv7-a/arm_dataabort.c
```diff
@@ -183,6 +183,8 @@ uint32_t *arm_dataabort(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
 
 segfault:
 	if (!IS_SECURE_STATE()) {
+		/* Add new line to distinguish between normal log and assert log.*/
+		lldbg_noarg("\n");
```
kishore-sn, line 187: As Chang mentioned, pls move this to the start of the print function to avoid repitition. 
Also, if you are adding comment for this, then you can change it as follows
`Abort log must always start at a new line`

os/arch/arm/src/armv7-a/arm_dataabort.c
```diff
@@ -78,6 +78,8 @@ uint32_t system_exception_location;
 
 static inline void print_dataabort_detail(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
 {
+	/* Abort log must always start at a new line.*/
+	lldbg_noarg("\n");
 	_alert("#########################################################################\n");
```
sunghan-chang, line 83: Let me suggest below
```
_alert("\n###################
```

sunghan-chang: current one is the same as previous. just location is different. But it's not important.

sunghan-chang (approved):

os/arch/arm/src/armv7-a/arm_dataabort.c
ritesh55555: as function name is printed first in _alert api , so putting \n will not work properly , i tested

kishore-sn (approved):

PR #6599, opened 2024-12-26
os/board/rtl8730e: change BLE HCI debug print function

yeetee179:
- Changed HCI_ERR and HCI_INFO to DGB_8195A, which does not use printf


j11591 (approved):
LGTM

PR #6601, opened 2024-12-26
os/board/rtl8730e: fix multi scan ap config count 0

ZivLow:
- Problem: When scan_ap_config_count is 0, will print "Can't malloc memory for channel list"
- Root cause: malloc(0) returns NULL
- Fix: Add check before malloc
- Add multi scan scenario: If config is not NULL, scan_all is false, scan_ap_config_count is 0, then do not scan

### DIFF
--- a/os/board/rtl8730e/src/component/os/tizenrt/rtk_netmgr.c
+++ b/os/board/rtl8730e/src/component/os/tizenrt/rtk_netmgr.c
@@ -506,58 +506,67 @@ trwifi_result_e wifi_netmgr_utils_scan_multi_ap(struct netdev *dev, trwifi_scan_
 	scan_param.scan_user_callback = app_scan_result_handler;
 
 	if (config) {
-		if (config->scan_ap_config_count > SSID_SCAN_NUM) {
-			RTW_API_INFO("ERROR: SSID count exceeded, maximum allowed:%d given:%d\n\r",SSID_SCAN_NUM,config->scan_ap_config_count);
-			TRWIFI_POST_SCANEVENT(ameba_nm_dev_wlan0, LWNL_EVT_SCAN_FAILED, NULL);
-			return TRWIFI_INVALID_ARGS;
-		}
-		for (i = 0; i < config->scan_ap_config_count; i++) {
-			/* Scan all channels if any channel in scan config is set to 0 */
-			if (config->scan_ap_config[i].channel == 0) {
-				scan_all_ch = 1;
-				/* Skip checking of channel validity if scanning all channels */
-				break;
-			}
-			ch_valid = 0;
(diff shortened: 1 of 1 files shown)


os/board/rtl8730e/src/component/os/tizenrt/rtk_netmgr.c
ZhenBei-Sin, line 1: Hi @ZivLow & @jwei5 

Why wasn't the check for config->scan_ap_config_count is 0 performed earlier, at line 509?
If config->scan_ap_config_count is 0, there's no need to enter the for loop at line 514.
for (i = 0; i < config->scan_ap_config_count; i++)

ZivLow: Hi @ZhenBei-Sin ,
multi_scan_scenario

Due to requirement to do full scan in scenario 2:
- Config: not null
- scan_all: true
- scan_ap_config_count: 0

Because scenario 2 need to do full scan, we cannot return TRWIFI_INVALID_ARGS early by checking scan_ap_config_count is 0 at line 509.

ZhenBei-Sin: Hi @ZivLow 
Does the full scan refer to the TRWIFI_POST_SCANEVENT ?

My suggestion:
image

ZivLow: Hi @ZhenBei-Sin 
full scan refers to calling `wifi_scan_networks(&scan_param, 0)` with `scan_param.options = RTW_SCAN_ALL;`
This happens when `config->scan_all` is set.

ZivLow: > Hi @ZivLow & @jwei5
> 
> Why wasn't the check for config->scan_ap_config_count is 0 performed earlier, at line 509? If config->scan_ap_config_count is 0, there's no need to enter the for loop at line 514. for (i = 0; i < config->scan_ap_config_count; i++)

I have just added `if (config->scan_ap_config_count)` check at line 509.
I also removed `if (config->scan_ap_config_count)` check right before malloc for channel_list.

> Hi @ZivLow Does the full scan refer to the TRWIFI_POST_SCANEVENT ?
> 
> My suggestion: image

I did not add the else statement return part if scan_ap_config_count is 0 from your suggestion.
This is because for scenario 2, we still need to check if scan_all is true, so cannot return early in else part.

hs36-kim (approved):
LGTM
jw00.jeong: Tizen Lite 5.0에 포함해서 확인했던 PR 입니다.

PR #6602, opened 2024-12-26
os/board/rtl8730e: Optimize wifi disconnect handler

jwei5:
- Previously, both wifi_disconnect and wifi_join_status_indicate could trigger TizenRT wifi linkdown callback
- Re-arrange flow such that TizenRT wifi linkdown callback is only called from RTK wifi disconnect handler for STA mode


hs36-kim (approved):
LGTM
jw00.jeong: Tizen Lite 5.0에 포함해서 확인했던 PR 입니다.

PR #6605, opened 2024-12-30
os/kernel/wdog: add debug logs in wd_expiration

gSahitya-samsung:
This PR adds debug log in wd_expiration. It logs the  wdog structure
value which can be useful to debug the cause of timer expiration fail.
This PR also modifies the wdog struct to keep track of process pid
which is responsible for timer creation.

### DIFF
--- a/os/include/assert.h
+++ b/os/include/assert.h
@@ -113,11 +113,16 @@ extern char assert_info_str[CONFIG_STDIO_BUFFER_SIZE];
 
 #ifdef CONFIG_DEBUG
 
+#define DEBUGPANIC_INFO(fmt, ...) \
+	do { \
+			snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
+			up_assert((const uint8_t *)__FILE__, (int)__LINE__); \
+	} while(0); \
+
 #define DEBUGASSERT_INFO(f, fmt, ...) \
 	{ \
 		if (!(f)) { \
-			  snprintf(assert_info_str, CONFIG_STDIO_BUFFER_SIZE, fmt, ##__VA_ARGS__); \
-			  up_assert((const uint8_t *)__FILE__, (int)__LINE__); \
+			  DEBUGPANIC_INFO(fmt, ...); \
 		} \
 	}
 
@@ -132,6 +137,7 @@ extern char assert_info_str[CONFIG_STDIO_BUFFER_SIZE];
 
 #else
 
+#define DEBUGPANIC_INFO(fmt, ...)
 #define DEBUGASSERT_INFO(f, fmt, ...)
 #define DEBUGASSERT(f)
 #define DEBUGVERIFY(f) ((void)(f))
--- a/os/include/tinyara/wdog.h
+++ b/os/include/tinyara/wdog.h
@@ -153,6 +153,9 @@ struct wdog_s {
 #ifdef CONFIG_PIC
 	FAR void *picbase;			/* PIC base address */
 #endif
+#ifdef CONFIG_DEBUG
+	int pid;					/* The pid of process which creates wdog timer */
+#endif
 	int lag;					/* Timer associated with the delay */
 	uint8_t flags;				/* See WDOGF_* definitions above */
 	uint8_t argc;				/* The number of parameters to pass */
--- a/os/kernel/wdog/wd_dbg.c
+++ b/os/kernel/wdog/wd_dbg.c
@@ -1,0 +1,94 @@
+/****************************************************************************
+ *
+ * Copyright 2025 Samsung Electronics All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+
+/****************************************************************************
+ * Included Files
+ ****************************************************************************/
+
+#include <sys/types.h>
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdarg.h>
+#include <unistd.h>
+#include <sched.h>
+#include <assert.h>
+#include <errno.h>
+
+#include <tinyara/arch.h>
+#include <tinyara/wdog.h>
+#include <tinyara/mm/mm.h>
+
+#include "sched/sched.h"
+#include "wdog/wdog.h"
+
+static inline void wd_dump(struct wdog_s *wdog)
+{
+	int index = 0;
+	lldbg("Wdog address = 0x%08x\n", wdog);
+	lldbg("lag: %d\n", wdog->lag);
+	lldbg("flags: %u\n", wdog->flags);
+	lldbg("pid: %d\n", wdog->pid);
+	lldbg("func: %p\n", wdog->func);
+	lldbg("argc: %u\n", wdog->argc);
+	for (; index < CONFIG_MAX_WDOGPARMS; index++) {
+		lldbg("parm[%d]: %u\n", index, wdog->parm[index]);
+	}
+}
+
+void wd_corruption_dbg(struct wdog_s *wdog)
+{
+	uint32_t wd_node;
+	lldbg("WDOG INFO\n");
+	wd_dump(wdog);
+
+	if (mm_get_heap(wdog)) {
+
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+		// Wdog struct is in heap
+		lldbg("WDOG is in HEAP. Print neighbour heap nodes\n");
+		wd_node = (uint32_t)wdog - SIZEOF_MM_ALLOCNODE;
+		lldbg("PREV WDOG INFO\n");
+		mm_dump_node((struct mm_allocnode_s *)(wd_node - (((struct mm_allocnode_s *)wd_node)->preceding & ~MM_ALLOC_BIT)), "PREV HEAP NODE");
+		lldbg("NEXT WDOG INFO\n");
+		mm_dump_node((struct mm_allocnode_s *)(wd_node + ((struct mm_allocnode_s *)wd_node)->size), "NEXT HEAP NODE");
+#endif
+
+	} else if (wd_is_prealloc(wdog)) {
+		// Wdog struct is from prealloc list
+
+		lldbg("WDOG is in pre alloc list. Print neighbor wdog nodes\n");
+
+		if (wd_is_prealloc(wdog - 1)) {
+			lldbg("PREV WDOG INFO\n");
+			wd_dump(wdog - 1);
+		} else {
+			lldbg("WDOG is first in pre-alloc list\n");
+		}
+
+		if (wd_is_prealloc(wdog + 1)) {
+			lldbg("NEXT WDOG INFO\n");
+			wd_dump(wdog + 1);
+		} else {
+			lldbg("WDOG is last in pre-alloc list\n");
+		}
+
+	} else {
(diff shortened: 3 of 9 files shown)


os/include/assert.h
```diff
@@ -113,11 +113,18 @@ extern char assert_info_str[CONFIG_STDIO_BUFFER_SIZE];
 
 #ifdef CONFIG_DEBUG
 
-#define DEBUGASSERT_INFO(f, fmt, ...) \
+#define DEBUGPANIC_INFO(fmt, ...) \
 	{ \
```
sunghan-chang, line 117: We can remove here and line 122, `{ }`

os/kernel/wdog/wd_start.c
```diff
@@ -242,11 +242,13 @@ int wd_start(WDOG_ID wdog, int delay, wdentry_t wdentry, int argc, ...)
...
 	/* Verify the wdog */
 
-	if (!wdog || argc > CONFIG_MAX_WDOGPARMS || delay < 0) {
+	if (!wdog || argc > CONFIG_MAX_WDOGPARMS || argc < 0 || delay < 0) {
 		set_errno(EINVAL);
 		return ERROR;
 	}
 
+	/* Store the pid of process to keep track which process is responsible for wdog expiration */
+	wdog->pid = getpid();
```
sunghan-chang, line 251: Let's add an empty line after this to split.

sunghan-chang:
@gSahitya-samsung Could you add more logs that we added internally? If possible, let's add it with debugpanic_info instead of new function.

gSahitya-samsung:
> @gSahitya-samsung Could you add more logs that we added internally? If possible, let's add it with debugpanic_info instead of new function.

Do we need to move `mm_dump_node` from internal also?

os/kernel/wdog/wd_initialize.c
```diff
@@ -100,7 +100,7 @@ uint16_t g_wdnfree;
 * in the pool is a configuration item.
  */
 
-static struct wdog_s g_wdpool[CONFIG_PREALLOC_WDOGS];
+struct wdog_s g_wdpool[CONFIG_PREALLOC_WDOGS];
```
sunghan-chang, line 103: Instead of accessing the variable directly, let's make a new function to access this.
This is not good at SAM.

os/kernel/wdog/wd_start.c
```diff
@@ -111,6 +112,71 @@ typedef void (*wdentry4_t)(int argc, uint32_t arg1, uint32_t arg2, uint32_t arg3
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
+#ifdef CONFIG_DEBUG
+extern struct wdog_s g_wdpool[CONFIG_PREALLOC_WDOGS];
+static inline void wd_dump(struct wdog_s *wdog)
```
sunghan-chang, line 117: Let's make a new file for this.

os/kernel/wdog/wd_start.c
```diff
@@ -111,6 +112,71 @@ typedef void (*wdentry4_t)(int argc, uint32_t arg1, uint32_t arg2, uint32_t arg3
...
+#if CONFIG_MAX_WDOGPARMS > 3
+	lldbg("parm[3]: %u\n", wdog->parm[3]);
+#endif
+}
+
+static inline void wd_corruption_dbg(struct wdog_s *wdog)
+{
+	lldbg("WDOG INFO\n");
+	wd_dump(wdog);
+	lldbg("Wdog address = 0x%08x\n", wdog);
```
sunghan-chang, line 142: I think we can add this in the wd_dump, not outside of it.

os/kernel/wdog/wd_start.c
```diff
@@ -111,6 +112,71 @@ typedef void (*wdentry4_t)(int argc, uint32_t arg1, uint32_t arg2, uint32_t arg3
...
+#endif
+#if CONFIG_MAX_WDOGPARMS > 1
+	lldbg("parm[1]: %u\n", wdog->parm[1]);
+#endif
+#if CONFIG_MAX_WDOGPARMS > 2
+	lldbg("parm[2]: %u\n", wdog->parm[2]);
+#endif
+#if CONFIG_MAX_WDOGPARMS > 3
+	lldbg("parm[3]: %u\n", wdog->parm[3]);
+#endif
```
sunghan-chang, line 135:
```
int parm_idx;
for (parm_idx = 0; parm_idx < CONFIG_MAX_WDOGPARMS; parm_idx++) {
    lldbg("parm[%d]: %u\n", parm_idx, wdog->parm[parm_idx]);
}
```
I think we can reduce the code.

os/kernel/wdog/wd_start.c
```diff
@@ -111,6 +112,71 @@ typedef void (*wdentry4_t)(int argc, uint32_t arg1, uint32_t arg2, uint32_t arg3
...
+	lldbg("WDOG INFO\n");
+	wd_dump(wdog);
+	lldbg("Wdog address = 0x%08x\n", wdog);
+
+	if (mm_get_heap(wdog)) {
+		// Wdog struct is in heap
+		lldbg("WDOG is in HEAP. Print neighbor heap nodes\n");
+		// uint32_t node = (uint32_t)wdog - SIZEOF_MM_ALLOCNODE;
+		// mm_dump_node((struct mm_allocnode_s *)(node - (((struct mm_allocnode_s *)node)->preceding & ~MM_ALLOC_BIT)), "PREV HEAP NODE");
+		// mm_dump_node((struct mm_allocnode_s *)(node + ((struct mm_allocnode_s *)node)->size), "NEXT HEAP NODE");
```
sunghan-chang, line 149: 1. Could you let me know why they are commented?
2. Let's print title as wdog does like line 161 and line 168.

gSahitya-samsung: `mm_dump_node` is not present in public. We need to move code from internal, can we do that?

os/include/assert.h
gSahitya-samsung: Done

os/kernel/wdog/wd_start.c
gSahitya-samsung: Done

os/kernel/wdog/wd_start.c
sunghan-chang: Let's add it here as well. Is there any reason why we can't have it here?

os/kernel/wdog/wd_dbg.c
```diff
@@ -0,0 +1,100 @@
...
+		} else {
+			llwdbg("WDOG is last in pre-alloc list\n");
+		}
+		
+	} else {
+		llwdbg("WDOG Address is invalid\n");
+	}
+
+}
+#endif
```
sunghan-chang, line 100: Could you check this warning?

os/mm/mm_heap/mm_heap_dbg.c
```diff
@@ -76,3 +76,68 @@ void mm_dump_heap_region(uint32_t start, uint32_t end)
 	heap_dbg("#########################################################################################\n");
 }
 
+/************************************************************************
+ * Name: mm_dump_heap_free_node_list
+ *
+ * Description: Print the hex value contents of heap free nodes.
+ ************************************************************************/
+void mm_dump_heap_free_node_list(struct mm_heap_s *heap)
```
sunghan-chang, line 84: This is not a wdog debug log. Let's split commits

os/kernel/wdog/wd_start.c
```diff
@@ -242,11 +245,14 @@ int wd_start(WDOG_ID wdog, int delay, wdentry_t wdentry, int argc, ...)
 
 	/* Verify the wdog */
 
-	if (!wdog || argc > CONFIG_MAX_WDOGPARMS || delay < 0) {
+	if (!wdog || argc > CONFIG_MAX_WDOGPARMS || argc < 0 || delay < 0) {
```
sunghan-chang, line 248: same as previous.
Let's split commits for this.

os/kernel/wdog/wd_dbg.c
```diff
@@ -0,0 +1,100 @@
...
+	llwdbg("pid: %d\n", wdog->pid);
+	llwdbg("argc: %u\n", wdog->argc);
+    for (; index < CONFIG_MAX_WDOGPARMS; index++) {
+	    llwdbg("parm[%d]: %u\n", index, wdog->parm[index]);
+    }
+}
+
+void wd_corruption_dbg(struct wdog_s *wdog)
+{
+    uint32_t wd_node;
```
sunghan-chang, line 58: Need to check indentation.
There are many lines which use spaces instead of tabs

os/kernel/wdog/wd_dbg.c
```diff
@@ -0,0 +1,100 @@
...
+#include <errno.h>
+
+#include <tinyara/arch.h>
+#include <tinyara/wdog.h>
+#include <tinyara/mm/mm.h>
+
+#include "sched/sched.h"
+#include "wdog/wdog.h"
+
+#ifdef CONFIG_DEBUG
```
sunghan-chang, line 40: You should include `<tinyara/config.h>` to use CONFIG_XX.
But let's add this conditional at Make.defs, not here.

os/kernel/wdog/wd_start.c
```diff
@@ -66,6 +66,7 @@
 
 #include <tinyara/arch.h>
 #include <tinyara/wdog.h>
+#include <tinyara/mm/mm.h>
```
sunghan-chang, line 69: I think this is unnecessary here.

os/include/tinyara/wdog.h
```diff
@@ -157,6 +157,7 @@ struct wdog_s {
 	uint8_t flags;				/* See WDOGF_* definitions above */
 	uint8_t argc;				/* The number of parameters to pass */
 	uint32_t parm[CONFIG_MAX_WDOGPARMS];
+	int pid;					/* The pid of process which creates wdog timer */
```
sunghan-chang, line 160: When CONFIG_DEBUG is disabled, this is not used. Let's cover this with the conditional.

os/kernel/wdog/wd_dbg.c
```diff
@@ -0,0 +1,100 @@
...
+
+static inline void wd_dump(struct wdog_s *wdog)
+{
+    int index = 0;
+	llwdbg("Wdog address = 0x%08x\n", wdog);
```
sunghan-chang, line 45: `llwdbg` is the warning message. But when we call PANIC, we print them. So, in my opinion, it should be lldbg, not llwdbg.

os/kernel/wdog/wd_dbg.c
```diff
@@ -0,0 +1,100 @@
...
+	llwdbg("func: %p\n", wdog->func);
```
sunghan-chang, line 46: Much minor comment. But how about printing this before argc? Because the argc and the parm are args of the func.

sunghan-chang (approved):

kishore-sn (approved):

os/kernel/wdog/wd_initialize.c
gSahitya-samsung: Done

os/kernel/wdog/wd_start.c
gSahitya-samsung: Done

r-prabu (approved):
debug log additions for watch dog expiration is good to merge

os/kernel/wdog/wd_initialize.c
```diff
@@ -110,6 +110,36 @@ static struct wdog_s g_wdpool[CONFIG_PREALLOC_WDOGS];
...
+	uintptr_t start = (uintptr_t)(&g_wdpool[0]);
+	uintptr_t end = (uintptr_t)(&g_wdpool[CONFIG_PREALLOC_WDOGS - 1]);
+	
+	if (end < start) {
+		start = start ^ end;
+		end = start ^ end;
+		start = start ^ end;
+	}
+
+	return (wdog_ptr >= start) && (wdog_ptr <= end) && (((wdog_ptr - start) % sizeof(struct wdog_s)) == 0);
```
abhishek-samsung, line 140: I am not sure if this works as intended as all pointers are of type uintptr_t, instead, this would work if they are of type char * .

gSahitya-samsung: These are not pointer instead pointer's value.

Ref: https://cplusplus.com/reference/cstdint/

PR #6608, opened 2024-12-30
Add new API to pause and resume the CPUs and handle it properly

neel-samsung:


os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -583,24 +583,13 @@ void up_assert(const uint8_t *filename, int lineno)
...
-		int me = sched_getcpu();
-		for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
-			if (cpu != me) {
-				/* Pause the CPU */
-				up_cpu_pause(cpu);
-				/* Wait while the pause request is pending */
-				while (up_cpu_pausereq(cpu)) {
-				}
-			}
+		up_cpu_pause_all();
```
sunghan-chang, line 591: Please check indentation.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -636,6 +625,13 @@ void up_assert(const uint8_t *filename, int lineno)
 	lldbg_noarg("\n");
 #endif
 
+#ifdef CONFIG_SMP
+	/* If SMP is enabled then we need to resume all the other cpu's
+	 * which we paused earlier.
+	 */
+	up_cpu_resume_all();
+#endif
```
sunghan-chang, line 633: Is this right to resume all cpus before recovering the fault?

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
+ * Description:
+ *   pause all the CPUs other than the current cpu. Internally, it calls 
+ * 	 up_cpu_pause() api to stop all the CPU's.
+ *
+ * Returned Value:
+ *   Zero on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_cpu_pause_all()
```
sunghan-chang, line 432: should be (void)
same as other new apis

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
 
+/****************************************************************************
+ * Name: up_cpu_pause_all
+ *
+ * Description:
+ *   pause all the CPUs other than the current cpu. Internally, it calls 
+ * 	 up_cpu_pause() api to stop all the CPU's.
+ *
+ * Returned Value:
+ *   Zero on success; a negated errno value on failure.
```
sunghan-chang, line 428: It always returns OK. Need the return?

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
+ ****************************************************************************/
+
+int up_cpu_pause_all()
+{
+	int me = sched_getcpu();
+	for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
+		if (cpu != me) {
+			/* Pause the CPU */
+			up_cpu_pause(cpu);
+			/* Wait while the pause request is pending */
```
sunghan-chang, line 439: What does "the pause request is pending" mean?

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
+ *
+ ****************************************************************************/
+
+int up_cpu_resume_all()
+{
+	int me = sched_getcpu();
+	for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
+		if (cpu != me) {
+			/* Resume the CPU */
+			up_cpu_resume(cpu);
```
sunghan-chang, line 464: Don't you need to check resume here like pause all does?

os/arch/arm/src/armv7-a/arm_assert.c
neel-samsung: So, there will be two cases:
1) If it is kernel fault, then the board will reboot. Resuming all CPUs won't affect this case.
2) If the crash is in app binary, then we will be pausing all the CPUs inside recovery.

os/arch/arm/src/armv7-a/arm_cpupause.c
neel-samsung: I think we make return type void.

os/arch/arm/src/armv7-a/arm_cpupause.c
neel-samsung: We will be waiting till the CPU gets paused.

os/arch/arm/src/armv7-a/arm_cpupause.c
neel-samsung: yes, it's better to wait till the cpu gets resume. Let me add it.

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
+ * Returned Value:
+ *   Zero on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
+int up_cpu_pause_all()
+{
+	int me = sched_getcpu();
+	for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
+		if (cpu != me) {
```
kishore-sn, line 456: We can add the condition in for statement:
`for (int cpu = 0; cpu != me && cpu < CONFIG_SMP_NCPUS; cpu++) {`

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -417,4 +417,53 @@ int up_cpu_resume(int cpu)
...
+			/* Pause the CPU */
+			up_cpu_pause(cpu);
+			/* Wait while the pause request is pending */
+			while (up_cpu_pausereq(cpu)) {
+			}
```
kishore-sn, line 461: We dont need to perform this while loop. When up_cpu_pause() returns, then we are sure that the cpu has been paused.

os/arch/arm/src/armv7-a/arm_assert.c
sunghan-chang: 1. Even we call reboot, why do we need to resume?
2. Who(where) pauses the all cpus in recovery?

os/arch/arm/src/armv7-a/arm_cpupause.c
sunghan-chang: You changed the return type to void, but this line is not changed.

os/arch/arm/src/armv7-a/arm_assert.c
neel-samsung: 1. Yeah, I agree. There is no point in resuming here for this case.
2.  As, we are not pausing the CPUs in recovery as of now, let me revert this change.

os/arch/arm/src/armv7-a/arm_cpupause.c
neel-samsung: I think, it will get fail when me = CPU0, it will exit from the loop and won't pause for CPU1

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -90,6 +90,14 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 {
 	irqstate_t flags = enter_critical_section();
 
+#ifdef CONFIG_SMP
+	/* If SMP is enabled then we need to pause all the other cpu's immediately
+	 * because the kernel state might be invalid at this point. If we don't pause
```
kishore-sn, line 95: I dont think the kernel state will be invalid. Let us just mention that we pause the other cpu to avoid mixed logs.

os/arch/arm/src/armv7-a/arm_assert.c
sunghan-chang: 2. When app fault is done in recovery, all the cpus should be resumed. You should add resuming at somewhere.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -589,17 +589,7 @@ void up_assert(const uint8_t *filename, int lineno)
...
-			if (cpu != me) {
-				/* Pause the CPU */
-				up_cpu_pause(cpu);
-				/* Wait while the pause request is pending */
-				while (up_cpu_pausereq(cpu)) {
-				}
-			}
-
-		}
+		up_cpu_pause_all();
```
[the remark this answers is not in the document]
neel-samsung: We are just calling newly added API. should we do this in separate commit?

os/arch/arm/src/armv7-a/arm_schedulesigaction.c
```diff
@@ -256,18 +256,13 @@ void up_schedule_sigaction(struct tcb_s *tcb, sig_deliver_t sigdeliver)
...
-					/* Pause the CPU */
+					/* Pause the other CPUs */
 
-					up_cpu_pause(cpu);
-
-					/* Wait while the pause request is pending */
-
-					while (up_cpu_pausereq(cpu)) {
-					}
+					up_cpu_pause_all();
```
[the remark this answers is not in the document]
neel-samsung: okay, we can check it by return value of up_cpu_pause() API.

os/arch/arm/src/armv7-a/arm_assert.c
neel-samsung: Let me check it properly.

sunghan-chang:
@r-prabu @kishore-sn please review and merge this.

os/arch/arm/src/armv7-a/arm_assert.c
neel-samsung: As of now, we don't support recovery with SMP, let us add it separately while supporting recovery

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -583,24 +583,13 @@ void up_assert(const uint8_t *filename, int lineno)
...
 
 #ifdef CONFIG_SMP
-	/* If SMP is enabled and there is a crash in kernel space, then we need to
-	 * pause all the other cpu's immediately because the kernel state might be
-	 * invalid at this point. If we dont pause other cpu's then it might lead
-	 * to multiple asserts.
+	/* If SMP is enabled and there is a crash, then we need to pause all
+	 * the other cpu's immediately because the kernel state might be
+	 * invalid at this point. If we dont pause other cpu's then it might
+	 * lead to multiple asserts.
```
kishore-sn, line 589: I think this comment is not valid. Let us change as follows:
"Pause all other cpus to avoid mix up of logs while printing asssert logs"

sunghan-chang:
@ewoodev Could you let us know how to reproduce the issue?
@neel-samsung Once @ewoodev shares the way, could you confirm this PR?
@r-prabu And then let's merge this.

ewoodev:
I havn't reproduct this issue.
but I think, you can make log mixing env with adding printing log cpu1 and making cpu0 alloc fail.

sunghan-chang:
> I havn't reproduct this issue. but I think, you can make log mixing env with adding printing log cpu1 and making cpu0 alloc fail.

@neel-samsung Could you confirm this PR with the way @ewoodev suggested?

neel-samsung:
> > I havn't reproduct this issue. but I think, you can make log mixing env with adding printing log cpu1 and making cpu0 alloc fail.
> 
> @neel-samsung Could you confirm this PR with the way @ewoodev suggested?

yes, I have made cpu0 to mem alloc fail and on the other hand created new thread to run on the core1 to print the log. 
There is no log mixup. Hence, the PR seems to be tested and ready to merge.

sunghan-chang (approved):

sunghan-chang:
@kishore-sn @r-prabu Let's merge

kishore-sn (approved):

PR #6609, opened 2024-12-31
build/tools/amebasmart: Add display log to show the value of BKUP_REG2

ZhenBei-Sin:
Change Notes:
- When a BOOT/KM0/KM4 fault occurs, ensure BKUP_REG2 is successfully written with the value 0x1 and display it.
- Verify that the CA32 5s timer is running, read back the BKUP_REG2 value to determine need to trigger CA32 assert.
- Display the value of BKUP_REG2 during the BOOT stage.[BOOT-I] BKUP_REG2 0x0
- Modify the ASSERT check condition so that it triggers whenever the value of REG2 is not equal to 0x0.


os/arch/arm/src/armv7-a/arm_initialize.c
```diff
@@ -101,7 +102,7 @@ void up_initialize(void)
 {
 #ifdef CONFIG_SYSTEM_REBOOT_REASON
 	up_reboot_reason_init();
-	lldbg("[Reboot Reason] : %d\n", up_reboot_reason_read());
+	lldbg("[Reboot Reason] : %d [BKUP_REG2] 0x%x \n", up_reboot_reason_read(), BKUP_Read(BKUP_REG2));
```
sunghan-chang, line 105: This is a common code, not Realtek specific. So, we can't add it here.
It seems you print it in the rtl8730e_boot.c. Do you need to do here again?

sunghan-chang: Oh, it's the timer to check km4 status. Anyway, we can't add this log here. Please move it into the realtek function.

ZhenBei-Sin: Move to bootloader side to display it.

image

sunghan-chang (approved):

build/tools/amebasmart/bootloader/km4_boot_RELEASE_NOTE.txt
```diff
@@ -1,5 +1,9 @@
 /* == "version" + "Realtek git version" + "compile date" + "compile time" == */
 
+== version a1a0ee7deb_2024/12/31-15:44:33 ==
```
sunghan-chang, line 3: Because some product is over in dev, it could be impossible to apply.
If possible, I want to avoid bootloader change. How about printing at amebasmart_reboot_reason.c?

ZhenBei-Sin: Inside the BOOT_FaultHandler, two changes were included. 

This change can be moved to amebasmart_reboot_reason.c.
- Display the value of BKUP_REG2 during the BOOT stage.[BOOT-I] BKUP_REG2 0x0

However, we cannot avoid the following change:
When a BOOT fault occurs, ensure BKUP_REG2 is successfully written with the value 0x1 and display it.

Therefore, the bootloader img needs to check the BKUP_REG2 value.
Unless Samsung side agrees to only print the BKUP_REG2 value in the case of KM0 & KM4 fault only.

sunghan-chang: What is the `boot fault`? Fault in bootloader with some reason?

sunghan-chang:
@ZhenBei-Sin Even I merge this, the bootloader could not be updated. Then, we should verify this case - old bootloader + latest Tizen Lite. Could you check operation with this case?

build/tools/amebasmart/bootloader/km4_boot_RELEASE_NOTE.txt
ZhenBei-Sin: BOOT_FaultHandler
for example memory corruption, corrupted bootloader etc…

ZhenBei-Sin:
> @ZhenBei-Sin Even I merge this, the bootloader could not be updated. Then, we should verify this case - old bootloader + latest Tizen Lite. Could you check operation with this case?

Hi @sunghan-chang 
The difference between the old and new bootloader is to display out the BKUP_REG2 value only.
So its was compatible old / new bootloader with the existing latest Tizen Lite.

PR #6612, opened 2025-01-02
[Draft] os/se/ameba: Optimize AmebaSmart SE APIs access

ziyik:
- Remove Flash Lock Unlock for all SE APIs
- Call Flash Lock Unlock for necessary APIs only

### DIFF
--- a/os/se/ameba/rtl_se_smccc_wrapper.c
+++ b/os/se/ameba/rtl_se_smccc_wrapper.c
@@ -73,9 +73,7 @@ static unsigned long invoke_securetest(unsigned long function_id,
 									unsigned long arg2, unsigned long arg3)
 {
 	struct arm_smccc_res res;
-	FLASH_Write_Lock();
 	arm_smccc_smc(function_id, arg0, arg1, arg2, arg3, 0, 0, 0, &res);
-	FLASH_Write_Unlock();
 
 	//printf("\n ==== Output SMCCC: [function_id:%x] %x %x %x %x ==== \n", function_id, res.a0, res.a1, res.a2, res.a3);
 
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@ziyik With this PR, could you let us know elapsed time of each function in before and after?

ziyik:
> @ziyik With this PR, could you let us know elapsed time of each function in before and after?

This PR is created to check if it will improve the I2S noise issue, since it's not helping I would like to abandon it.
Is this okay?

sunghan-chang:
@ziyik This could be not enough to resolve the issue. But in my opinion, this is necessary.

os/se/ameba/rtl_se_smccc_wrapper.c
jylee9613, line 1: Hello @ziyik 

Isn't it necessary to call flash lock/unlock in key manager APIs?

```
ameba_hal_set_key
ameba_hal_get_key
ameba_hal_remove_key
ameba_hal_generate_key
```

ziyik: These APIs will modify or operate with the RAM key, and all Flash keys will be loaded in the RAM area during init.
These APIs do not interact with Flash.

jylee9613: and all Flash keys will be loaded in the RAM area during init
-> Do you mean se_initailize while booting?

ziyik: Yes, during init, the factory key will be loaded into the RAM slot.

ziyik:
> @ziyik With this PR, could you let us know elapsed time of each function in before and after?

I think you are concerned about the time needed for FLASH_Write_Lock(); and FLASH_Write_Unlock();.
I have tried to measure the time needed for both APIs, It causes less than a systick.

Below are the code used to measure:
	tickbefore_sleep = SYSTIMER_TickGet();
	FLASH_Write_Lock();
	FLASH_Write_Unlock();
	tickafter_sleep = SYSTIMER_TickGet();

printf("\n tickbefore_sleep = %d \n", tickbefore_sleep);
printf("\n tickafter_sleep = %d \n", tickafter_sleep);
printf("\n tick different = %d \n", tickafter_sleep - tickbefore_sleep);
printf("\n Time in us = %d \n", (((tickafter_sleep - tickbefore_sleep) * 1000 * 1000)/ (32 * 1024)));

Below is the print log:
 tickbefore_sleep = 1032175

 tickafter_sleep = 1032175

 tick different = 0

 Time in us = 0
 
 I execute the same code multiple times, most of the time is consumes 0 ticks, and only a few time it consumes 1 tick, which is like 30us.

zhongnuo-tang:
> is PR, could you let us know elapsed time of each function in before and after?

Hi, I don’t think this aligns with what Mr. Chang was asking. Instead of measuring the time taken for FLASH_Write_Lock() and FLASH_Write_Unlock(), the focus should be on measuring the execution time of each Secure Execution (SE) API.

The main concern is how long each SE API takes, as optimizing and removing unnecessary FLASH_Write_Lock() calls will help reduce the time IRQs are disabled, leading to better performance.

PR #6619, opened 2025-01-08
os/arch/arm/src/armv7-a: Add memory barrier to ensure latest copy of spinlock variable is observed in ISR

lhenry-realtek:
This commit addresses a condition where the spinlock variable becomes un-synced between cores due to out-of-order memory access.

DMB is added to ensure the latest copy of the lock is retrieved before operating on it


gSahitya-samsung:
> Hi @gSahitya-samsung please help to verify this

We are still facing hang issue

gSahitya-samsung:
@lhenry-realtek I have taken your commit in PR #6623 

Thanks for checking

PR #6621, opened 2025-01-08
os/arch/arm/src/amebasmart: increase I2S dma page to 8 to increase to…

zhongnuo-tang:
…lerate of delays

1. there are delays between I2S DMA interrupt caused by other higher priority threads/flash operation, which if exceed the time taken to finish tx all data from dma pages will result in noise
2. with 4K size * 4 page, maximum delays it can tolerate will be ~85.33 ms, delays exceeding this will introduce noise.
3. this commit increases the page from 4 to 8, thus increases the maximum delays it can tolerate to ~170ms.


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -120,7 +120,7 @@
 #define OVER_SAMPLE_RATE (384U)
 
 #define I2S_DMA_PAGE_SIZE 4096 	/* 4 ~ 16384, set to a factor of APB size */
-#define I2S_DMA_PAGE_NUM 4	/* Vaild number is 2~4 */
+#define I2S_DMA_PAGE_NUM 8	/* Vaild number is 2~8 */
```
ewoodev, line 123: Please revert it

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2s_api.c
```diff
@@ -20,7 +20,7 @@
 #include "ameba_audio_clock.h"
 #include "amebasmart_i2s.h"
 
-#define SP_MAX_DMA_PAGE_NUM 8
+#define SP_MAX_DMA_PAGE_NUM 9
```
sunghan-chang, line 23: Is there any reason why you limit the number?

zhongnuo-tang: There is no restriction/limit of the page number for i2s dma page, but this page size will be used to initialize static buffer array for i2s dma (larger page number will means bigger binary size), therefore i have changed it to 9 (8 dma page + 1 zero page) so that we can support 8 pages of I2S DMA buffer.
static struct GDMA_CH_LLI LliTx[SP_MAX_DMA_PAGE_NUM];

sunghan-chang: From you reply, it looks it's a configurable value. Then, how about making a new config for this?
Because this is in BSP C code, it's impossible to know.

zhongnuo-tang: Hi @sunghan-chang,
I have modified this PR to include config for user to set I2S DMA page and size, and the maximum size in bsp is synced with value set by the user. Please help to review this PR again, thank you.

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -182,6 +182,20 @@ config AMEBASMART_I2S_TXCL
...
+	default 4096
+	range 4 16384
+	---help---
+		4 ~ 16384, set to a factor of APB size
+
+config AMEBASMART_I2S_TX_DMA_PAGE
+	int "I2S TX DMA PAGE"
+	default 4
+	---help---
+		DMA page should be more than 2
```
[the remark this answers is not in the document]
zhongnuo-tang: No upper limit, as long as we are okay with the binary size.

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -182,6 +182,20 @@ config AMEBASMART_I2S_TXCL
 	int "I2S TX Channel Length"
 	default 32
 	range 8 32
+
+config AMEBASMART_I2S_TX_DMA_SIZE
+	int "I2S TX DMA SIZE"
+	default 4096
+	range 4 16384
```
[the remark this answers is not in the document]
zhongnuo-tang: there is no independency between TX_DMA_SIZE and TX_DMA_PAGE

sunghan-chang:
@zhongnuo-tang I heard Eric would talk to you about this. Is there any update?

zhongnuo-tang:
> @zhongnuo-tang I heard Eric would talk to you about this. Is there any update?

Let me check with him what is it about..

zhongnuo-tang:
> @zhongnuo-tang I heard Eric would talk to you about this. Is there any update?

Hi @sunghan-chang, here is the update regarding the qns asked

1. DMA Page Size Limitation: Each DMA page has a size limit, with a maximum of 16,384 bytes (4,096 bytes × transfer width). This is the maximum amount of data that can be transferred in a single DMA operation.

2. Multi-Block (Linked List Mode): I2S DMA uses a multi-block transfer approach, where each page can be viewed as a memory region. As long as these pages (or nodes) are properly linked, the DMA controller will sequentially access each memory address and transfer data to or from the peripheral FIFO.

3. DMA Hardware Operation: The actual data transfer between memory and the peripheral FIFO is handled by the DMA hardware. However, the process still requires RAM for tasks like filling each DMA page in the ISR handler, assuming the code is located in RAM.

4. Differences Between AmebaSmart and AmebaD: Unlike the AmebaD platform, which has a dedicated DMA controller for I2S with a maximum of 16K × 4 page, the AmebaSmart platform uses a common DMA controller for all peripherals. This means the DMA is shared across different peripherals on the AmebaSmart platform, rather than being specific to one like in the AmebaD.

Anywhere, i think maybe it would be better to restrict the page size to 2~8 to prevent confusion?

sunghan-chang:
```
Differences Between AmebaSmart and AmebaD: Unlike the AmebaD platform, which has a dedicated DMA controller for I2S with a maximum of 16K × 4 page, the AmebaSmart platform uses a common DMA controller for all peripherals. This means the DMA is shared across different peripherals on the AmebaSmart platform, rather than being specific to one like in the AmebaD.
```

@zhongnuo-tang Could you confirm this lastly? Even it does, no upper limitation? page size 16384, page num 100. Is this working?

sunghan-chang:
If yes, I think no need to modify.

zhongnuo-tang:
> ```
> Differences Between AmebaSmart and AmebaD: Unlike the AmebaD platform, which has a dedicated DMA controller for I2S with a maximum of 16K × 4 page, the AmebaSmart platform uses a common DMA controller for all peripherals. This means the DMA is shared across different peripherals on the AmebaSmart platform, rather than being specific to one like in the AmebaD.
> ```
> 
> @zhongnuo-tang Could you confirm this lastly? Even it does, no upper limitation? page size 16384, page num 100. Is this working?

Hi @sunghan-chang , 
With the current I2S DMA buffer settings (page_num: 100, page_size: 16384), testing with the SoundPlayer example and playing a 2-minute soundtrack works fine. However, some issues have been observed in the product code.

Before Wi-Fi Connection: The audio functions correctly.
After Wi-Fi Connection: There is mute occurs in Bixby.
As Eric is currently occupied with BLE-related issues, we have limited time to investigate the root cause of the muting.

ewoodev (approved):

PR #6622, opened 2025-01-08
kernel/sched: fix logging after stack overflow

ritesh55555:
This commit changes logging method from dbg to lldbg to print stack overflow information. This is done because after stack overflow, using dbg does not work because of more stack memory requirement. But in the lldbg case, it uses low level syslog, so it uses minimal stack memory.

logs before:
```
cycle in normal thread done
stack base ptr of normal thread is 0xdeadbeef
cycle in recursive thread done with cnt 423

security level: 0
===========================================================
Assertion details
===========================================================
print_assert_detail: Assertion failed CPU0 at file: sched/sched_removereadytorun.c line 125 task: recusrsive_thread pid: 20
print_assert_detail: Assert location (PC) : 0x0e01b587
```

logs after changing to lldbg:
```
cycle in normal thread done
stack base ptr of normal thread is 0xdeadbeef
cycle in recursive thread done with cnt 423
stack base 
###############    STACK OVERFLOW at pid 20 (recusrsive_thread) ###################

security level: 0
===========================================================
Assertion details
===========================================================
print_assert_detail: Assertion failed CPU0 at file: sched/sched_removereadytorun.c line 125 task: recusrsive_thread pid: 20
print_assert_detail: Assert location (PC) : 0x0e01b587


```

### DIFF
--- a/os/kernel/sched/Make.defs
+++ b/os/kernel/sched/Make.defs
@@ -62,6 +62,10 @@ CSRCS += sched_lock.c sched_unlock.c sched_lockcount.c sched_self.c
 CSRCS += sched_getaffinity.c sched_setaffinity.c
 CSRCS += sched_getcpu.c
 
+ifeq ($(CONFIG_SW_STACK_OVERFLOW_DETECTION),y)
+CSRCS += sched_checkstackoverflow.c
+endif
+
 ifeq ($(CONFIG_PRIORITY_INHERITANCE),y)
 CSRCS += sched_reprioritize.c
 endif
--- a/os/kernel/sched/sched_addreadytorun.c
+++ b/os/kernel/sched/sched_addreadytorun.c
@@ -124,14 +124,7 @@ bool sched_addreadytorun(FAR struct tcb_s *btcb)
 	bool ret;
 
 #ifdef CONFIG_SW_STACK_OVERFLOW_DETECTION
-	if (*(uint32_t *)(rtcb->stack_base_ptr) != STACK_COLOR) {
-		dbg_noarg("###############    STACK OVERFLOW at pid %d ", rtcb->pid);
-#if CONFIG_TASK_NAME_SIZE > 0
-		dbg_noarg("(%s) ", rtcb->name);
-#endif
-		dbg_noarg("###################\n");
-		PANIC();
-	}
+	sched_checkstackoverflow(rtcb);
 #endif
 	/* Check if pre-emption is disabled for the current running task and if
 	 * the new ready-to-run task would cause the current running task to be
@@ -182,14 +175,7 @@ bool sched_addreadytorun(FAR struct tcb_s *btcb)
 	int cpu;
 
 #ifdef CONFIG_SW_STACK_OVERFLOW_DETECTION
-	if (*(uint32_t *)(rtcb->stack_base_ptr) != STACK_COLOR) {
-		dbg_noarg("###############    STACK OVERFLOW at pid %d ", rtcb->pid);
-#if CONFIG_TASK_NAME_SIZE > 0
-		dbg_noarg("(%s) ", rtcb->name);
-#endif
-		dbg_noarg("###################\n");
-		PANIC();
-	}
+	sched_checkstackoverflow(rtcb);
 #endif
 	/* Check if the blocked TCB is locked to this CPU */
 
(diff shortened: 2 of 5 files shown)


samsung-singh (approved):
fixed logging after stack overflow in kernel sched, this change is good to merge.

os/kernel/sched/sched_addreadytorun.c
```diff
@@ -125,11 +125,11 @@ bool sched_addreadytorun(FAR struct tcb_s *btcb)
...
 #ifdef CONFIG_SW_STACK_OVERFLOW_DETECTION
 	if (*(uint32_t *)(rtcb->stack_base_ptr) != STACK_COLOR) {
-		dbg_noarg("###############    STACK OVERFLOW at pid %d ", rtcb->pid);
+		lldbg_noarg("\n###############    STACK OVERFLOW at pid %d ", rtcb->pid);
 #if CONFIG_TASK_NAME_SIZE > 0
-		dbg_noarg("(%s) ", rtcb->name);
+		lldbg_noarg("(%s) ", rtcb->name);
 #endif
-		dbg_noarg("###################\n");
+		lldbg_noarg("###################\n");
```
sunghan-chang, line 132: It could cause log mix. 
Because we will call the PANIC after this message, how about adding enter_critical_section?

sunghan-chang (commented):
And the same codes are at 4 different locations. How about making a new api for that?

os/kernel/sched/sched_removereadytorun.c
```diff
@@ -88,6 +88,33 @@
  * Public Functions
  ****************************************************************************/
 
+/****************************************************************************
```
sunghan-chang, line 91: Sorry but let's add this with new file, not here.

ritesh55555: done

os/kernel/sched/sched_checkstackoverflow.c
```diff
@@ -0,0 +1,110 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ ****************************************************************************/
+/****************************************************************************
+ * kernel/sched/sched_checkstackoverflow.c
+ *
+ * Redistribution and use in source and binary forms, with or without
```
sunghan-chang, line 21: No need to add this license. Let's remove.

os/kernel/sched/sched_checkstackoverflow.c
```diff
@@ -0,0 +1,110 @@
...
+ *   rtcb - Points to the TCB that is ready-to-run
+ *
+ ****************************************************************************/
+
+void sched_checkstackoverflow(FAR struct tcb_s *rtcb)
+{
+	irqstate_t flags = enter_critical_section();
+
+	if (*(uint32_t *)(rtcb->stack_base_ptr) != STACK_COLOR) {
+		lldbg_noarg("\n###############    STACK OVERFLOW at pid %d ", rtcb->pid);
```
sunghan-chang, line 101: Need to include proper header for lldbg_noarg, PANIC, struct tcb_s, enter/leave_critical_section, STACK_COLOR, uint32_t.

os/kernel/sched/Make.defs
```diff
@@ -60,7 +60,7 @@ CSRCS += sched_setscheduler.c sched_getscheduler.c
 CSRCS += sched_yield.c sched_rrgetinterval.c sched_foreach.c
 CSRCS += sched_lock.c sched_unlock.c sched_lockcount.c sched_self.c
 CSRCS += sched_getaffinity.c sched_setaffinity.c
-CSRCS += sched_getcpu.c
+CSRCS += sched_getcpu.c sched_checkstackoverflow.c
```
sunghan-chang, line 63: The function is called with CONFIG_SW_STACK_OVERFLOW_DETECTION. 
It means this file is valid with the config. Let's add conditional for this.
Or, remove the conditional at all c file, and in header, let's do like below
```
#ifdef CONFIG_XXX
void sched_checkstackoverflow(FAR struct tcb_s *rtcb);
#else
#define sched_checkstackoverflow(x)
#endif
```

sunghan-chang (approved):

kishore-sn (approved):

PR #6623, opened 2025-01-08
os: fix amp hang issue

gSahitya-samsung:
To pause the cpu, cpu should call `up_cpu_paused`. The call to `up_cpu_paused` will get missed when `g_cpu_irqlock` is unlocked (means none of the cpu having access to irqlock), `arm_pause_handler` rely on `enter_critical_section` to pause the cpu, it must handle `up_cpu_pausereq` without any fail, otherwise other cpu will wait forever in spinlock.

Change `while` loop to `do-while` loop in `irq_waitlock`, which will ensure there is no pending `up_cpu_pausereq` before getting lock on `g_cpu_irqlock` otherwise pause request will get missed.

The `SP_DMB` is needed to ensure both cpu read correct spinlock value.

The `spin_lock(&g_cpu_resumed[cpu])` is moved upward in code to avoid race condition between cpus.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_cpupause.c
+++ b/os/arch/arm/src/armv7-a/arm_cpupause.c
@@ -187,15 +187,16 @@ int up_cpu_paused_save(void)
 
 int up_cpu_paused(int cpu)
 {
+	/* Ensure the CPU has been resumed to avoid causing a deadlock */
+
+	spin_lock(&g_cpu_resumed[cpu]);
+
 	/* Release the g_cpu_paused spinlock to synchronize with the
 	 * requesting CPU.
 	 */
 
 	spin_unlock(&g_cpu_paused[cpu]);
 
-	/* Ensure the CPU has been resumed to avoid causing a deadlock */
-
-	spin_lock(&g_cpu_resumed[cpu]);
 
 	/* Wait for the spinlock to be released.  The requesting CPU will release
 	 * the spinlock when the CPU is resumed.
--- a/os/include/tinyara/spinlock.h
+++ b/os/include/tinyara/spinlock.h
@@ -81,11 +81,45 @@ typedef uint8_t spinlock_t;
 #  define __SP_UNLOCK_FUNCTION 1
 #endif
 
-#  define spin_is_locked(l) (*(l) == SP_LOCKED)
 
 /****************************************************************************
  * Public Function Prototypes
  ****************************************************************************/
+
+/****************************************************************************
+ * Name: spin_is_locked
+ *
+ * Description:
+ *   check whether spinlock is locked or not.
+ *
+ * Input Parameters:
+ *   lock  - A reference to the spinlock object.
+ *
+ * Returned Value:
+ *   true  - if spin is locked
+ *   false - otherwise
+ *
+ ****************************************************************************/
+
+inline bool spin_is_locked(volatile FAR spinlock_t *lock)
+{
+	/* In multicore env, there is a possibility that up_cpu_pausereq will be 
+	 * FALSE in the CPU serving this ISR due to out-of-order memory ops in ARM.
+	 * This can happen when Store operation in up_cpu_pause on the spinlock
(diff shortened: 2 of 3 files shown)


sunghan-chang:
> Hi @gSahitya-samsung since your commit includes my change of including DMB in arm_pause_handler, is #6619 still required?

@gSahitya-samsung Let's split commits for @lhenry-realtek 's change. Could you cherry pick his commit here?

ewoodev:
As you know, I think, The rootcause is missed enter_critical_section in up_schedule_sigaction(). #6624 

Could you please update if below commit is not needed, Please update it
[os: fix amp hang issue](https://github.com/Samsung/TizenRT/pull/6623/commits/6ee637af8aa09596917d17900344495aecff32d4)

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -278,6 +278,22 @@ int arm_pause_handler(int irq, void *context, void *arg)
...
+	 * condition where g_cpu_paused[1] is '1' in CPU0 and g_cpu_paused[1] is 
+	 * '0' in CPU1
+	 * 
+	 * While the spinlock may be protected with DSB/DMB already, it may not be 
+	 * broadcasted to a different core such as the one serving this ISR. Thus, 
+	 * we need to ensure latest copy of the lock after Store is observed.
+	 * 
+	 * This instruction does not affect operation with only 1 CPU
+	 */
+	SP_DMB();
```
kishore-sn, line 296: Let us move this to the point where lock is being checked.. 

```
inline bool spin_is_locked(l) 
{
          SP_DMB();
          return (*(l) == SP_LOCKED);
}         
```

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -187,15 +187,16 @@ int up_cpu_paused_save(void)
 
 int up_cpu_paused(int cpu)
 {
+	/* Ensure the CPU has been resumed to avoid causing a deadlock */
+
+	spin_lock(&g_cpu_resumed[cpu]);
```
kishore-sn, line 192: How does this affect our issue?

gSahitya-samsung: This is to avoid race condition between multiple CPUs (more than 2). If 2 or more CPUs tries to pause a given CPU, then in between releasing lock on `g_cpu_pause` and getting lock on `g_cpu_resumed`, CPUs can read garbage value of `g_cpu_resume`, which will trigger multiple SGI to given CPU (unwanted condition which we don't want).

gSahitya-samsung:
> As you know, I think, The rootcause is missed enter_critical_section in up_schedule_sigaction(). #6624
> 
> Could you please update if below commit is not needed, Please update it [os: fix amp hang issue](https://github.com/Samsung/TizenRT/pull/6623/commits/6ee637af8aa09596917d17900344495aecff32d4)

Hi @ewoodev, this commit is required, the `irq_waitlock` should return `false` if there is pending pause request to cpu. 

The `enter_critical_section` in `up_schedule_sigaction` just preventing the above case to occur.

os/arch/arm/src/armv7-a/arm_cpupause.c
gSahitya-samsung: Done

r-prabu (approved):

os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -277,7 +278,7 @@ int arm_pause_handler(int irq, void *context, void *arg)
 	 * interrupt by calling up_cpu_paused().  If the pause event has already
 	 * been processed then g_cpu_paused[cpu] will not be locked.
 	 */
-
+
```
sunghan-chang, line 281: Let's remove this unnecessary change.

sunghan-chang (approved):

sunghan-chang (approved):

kishore-sn (approved):

PR #6629, opened 2025-01-14
kernel/wdog: Add to restore executing wdog and dump in assert

ewoodev:
If wdog is corrupted or if prefetch abort is occuring in wdog callback, current assertion log is not enough for debugging.

Therefore, during the executing wdog expireation, we save the executing wdog address, and printing wdog information during the assert log printing.

### DIFF
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -131,6 +131,8 @@ extern sq_queue_t g_freemsg_list;
 extern uint32_t system_exception_location;
 extern uint32_t user_assert_location;
 extern int g_irq_nums[3];
+
+extern struct wdog_s *g_wdog_debug;
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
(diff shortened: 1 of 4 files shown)


os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -296,6 +298,15 @@ static void check_assert_location(uint32_t *sp, uint8_t *irq_num, bool *is_irq_a
...
 		}
 	}
+
+	if (*is_irq_assert && g_wdog_debug) {
+		/* Assert in wdog excuting*/
+		lldbg("Code asserted in wdog!\n");
+		lldbg_noarg("===========================================================\n");
+		lldbg_noarg("Asserted wdog details\n");
+		lldbg_noarg("===========================================================\n");
+		wd_corruption_dbg(g_wdog_debug);
```
jeongarmy, line 308: Isn't it duplicated with wd_corruption_dbg in wd_start.c:173?
```
			switch (wdog->argc) {
			default:
				wd_corruption_dbg(wdog);
				DEBUGPANIC();
				break;
```

PR #6632, opened 2025-01-16
build/tools/amebasmart: Fix softap no beacon issue

jwei5:
- Fix issue of SoftAP not transmitting beacons when starting it after PM wake


hk-gwak (approved):
Fix issue that softap beacon emission stop for a while after softAP connection.

PR #6634, opened 2025-01-16
Remove config dependency to fix ps command issue

neel-samsung:


os/kernel/sched/sched_getcpu.c
```diff
@@ -63,3 +63,8 @@ int sched_getcpu(void)
 {
   return up_cpu_index();  /* Does not fail */
 }
+
+int sched_getcpucount(void)
+{
+  return CONFIG_SMP_NCPUS;
```
jeongarmy, line 69: What happens if this value is not present?
I think we need to define values for when config is not existing.

jeongarmy: like
```
#if defined(CONFIG_SMP_NCPUS)
   return CONFIG_SMP_NCPUS;
#else
   return 0
```
or
```
In the file,
#if defined(CONFIG_SMP_NCPUS)
#define NCPUS CONFIG_SMP_NCPUS
#else
#define NCPUS 0
...

in sched_getcpucount()
   return NCPUS;
```

apps/system/utils/utils_cpuload.c
```diff
@@ -179,32 +177,24 @@ static void cpuload_print_pid_value(char *buf, void *arg)
 	} else {
 #ifdef CONFIG_SCHED_MULTI_CPULOAD
 		for (i = PROC_STAT_CPULOAD_SHORT; i <= PROC_STAT_CPULOAD_LONG; i++) {
-#ifdef CONFIG_SMP
-			char * avgload[CONFIG_SMP_NCPUS + 1];
+			char * avgload[ncpus + 1];
```
jeongarmy, line 180: `ncpus` is not static value. you define array with not defined value.. Is it ok?

os/kernel/sched/sched_getcpu.c
neel-samsung: We are already defining the config if it is not present. 
https://github.com/Samsung/TizenRT/blob/e28022ba933ae59f45a3b4fbcb7f3a62960866dc/os/tools/mkconfig.c#L302-L304

apps/system/utils/utils_cpuload.c
neel-samsung: I think, this also shouldn't create issue as syscall will return a constant value based on ```CONFIG_SMP_NCPUS```

kishore-sn: Actually, we are not supposed to delare array using a variable as the array size. 
In this case, it might not cause issue because the array is local to the for loop and the ncpus variable is outside of the loop. However, pls check again to make sure you dont get any warning or error during build.

kishore-sn:
Please modify commit descriptions. This work is for ps and cpuload commands.
In app side commit, pls add the output of each of these commands for smp and no smp case.

os/kernel/init/os_start.c
```diff
@@ -296,6 +296,10 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
     &g_readytorun,
     TLIST_ATTR_PRIORITIZED | TLIST_ATTR_RUNNABLE
   },
+  {                                              /* TSTATE_TASK_ASSIGNED */
```
sunghan-chang, line 299: Coding rule violations. Same as other lines

sunghan-chang: To split different topic of change, in this commit, please modify lines you added only. And please add one more commit to modify other lines violations. (or new pr)

apps/system/utils/utils_cpuload.c
neel-samsung: I have made sure that it doesn't give any warning or failure.

kishore-sn:
Please change commit title `remove config dependency to fix ps command issue`  >>> 'Remove ssytem commands dependency on CONFIG_SMPxxx'

neel-samsung:
> Please change commit title `remove config dependency to fix ps command issue` >>> `Remove ssytem commands dependency on CONFIG_SMPxxx`

Updated

kishore-sn (approved):

kishore-sn:
@sunghan-chang 
Please let me know if I can merge this.

sunghan-chang (dismissed):

sunghan-chang:
@kishore-sn Let's merge.
And let's apply this to syscall as well. I want to remove all conditionals in the syscall.

sunghan-chang:
> @kishore-sn Let's merge.
> And let's apply this to syscall as well. I want to remove all conditionals in the syscall.

What I mean above is removing all conditionals (not smp only) with new PR.

sunghan-chang:
@kishore-sn Let's merge this PR.

os/include/pthread.h
```diff
@@ -779,14 +779,12 @@ int pthread_attr_setstacksize(FAR pthread_attr_t *attr, long stacksize);
...
 int pthread_attr_getstacksize(FAR const pthread_attr_t *attr, long *stackaddr);
 
-#ifdef CONFIG_SMP
 /* Set or obtain thread affinity attributes */
 
 int pthread_attr_setaffinity_np(FAR pthread_attr_t *attr, \
                 size_t cpusetsize, FAR const cpu_set_t *cpuset);
 int pthread_attr_getaffinity_np(FAR const pthread_attr_t *attr, \
                 size_t cpusetsize, cpu_set_t *cpuset);
-#endif
```
kishore-sn, line 789: This is not sufficient. 
We need to build the corresponding files in libc. 

Please check below files. We need to remove CONFIG_SMP from all these locations. 
lib/libc/sched/sched_cpucount.c:#ifdef CONFIG_SMP
lib/libc/sched/Make.defs:ifeq ($(CONFIG_SMP),y)
lib/libc/pthread/Make.defs:ifeq ($(CONFIG_SMP),y)

kishore-sn: Please make a separate commit for all the above changes.

neel-samsung: Added as required.

kishore-sn (approved):

PR #6635, opened 2025-01-17
os/board/rtl8730e: added BLE secure parameter setting, passkey display callback and passkey confirmation API

yeetee179:
- ble_manager_set_secure_param API is added to set BLE secure parameter
- ble_server_passkey_display_cb is added inform app layer passkey
- ble_manager_passkey_confirm is added to for passkey confirmation


hs36-kim (approved):
numeric comparision server side  먼저 개발 하기 위해 머지 수행합니다.

PR #6641, opened 2025-01-20
apps/examples/uart_loopback: Add option uart port selection

ewoodev:
Add '-p' and '--port' option
The option is set the uart port ttySx as ttyS1 or ttyS2, etc. The option input type is integer.

### DIFF
--- a/apps/examples/uart_loopback/uart_loopback.c
+++ b/apps/examples/uart_loopback/uart_loopback.c
@@ -47,16 +47,20 @@
 
 static bool is_running;
 
-static int uart_tx_loop(void)
+static int uart_tx_loop(pthread_addr_t *arg)
 {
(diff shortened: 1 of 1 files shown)


apps/examples/uart_loopback/uart_loopback.c
```diff
@@ -82,13 +86,15 @@ static int uart_rx_loop(void)
 	char *read_ptr;
 	int rx_test_count = 0;
 	char read_buf[TEST_STR_LEN];
+	int port_num = (int)arg;
+
+	sprintf(port, UART_DEV_PATH, port_num);
```
jeongarmy, line 91: We need to use `snprinf`.

ewoodev: Thanks , Updated

PR #6646, opened 2025-01-22
apps/system/pm_metrics: increase stack size to 2048

ewoodev:
The stack overflow is occuring when excuting pm_metrics. So increase stack size 1024 to 2048.

The app is excuted from only tash command.


sunghan-chang (dismissed):
Are you able to measure stack usage of pm metrics? We should check the peak value to verify that 2048 is enough or not.

ewoodev:
the peak stack size is 1060, I updated stack size 1280

jeongarmy (approved):
LGTM

sunghan-chang (approved):

PR #6652, opened 2025-01-29
armv7-a/arm_gicv2.c: add api for setting irq affinity

ritesh55555:
This commit adds a new API for gic module. It adds "up_affinity_irq" api that sets affinity of irq(SPI) to a specific core. Note that it does not support assigning multiple core yet.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_gicv2.c
+++ b/os/arch/arm/src/armv7-a/arm_gicv2.c
@@ -509,6 +509,75 @@ int up_prioritize_irq(int irq, int priority)
 }
 
 /****************************************************************************
+ * Name: up_set_irq_affinity
+ *
+ * Description:
+ *   set the affinity of a given irq (shared peripheral interrupts - SPI only) to given cpu set
+ *
+ * Input Parameters:
+ *   irq: irq id
+ *   cpuset: cpu set
+ *   
+ *   Example usage: 
+ *   for cpu 1    -> up_set_irq_affinity(10, 0x10)
+ *   for cpu 0    -> up_set_irq_affinity(10, 0x01)
+ *   for cpu 0, 1 -> up_set_irq_affinity(10, 0x11)
+ * 
+ *   Each bit of the register represents each core affinity.
+ *   Note: register supports 8 bits per interrupt. (So maximum 8 cpus)
+ *
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_set_irq_affinity(int irq, cpu_set_t cpu_set)
+{
+	/* Ignore invalid interrupt IDs */
+
+	if (irq >= GIC_IRQ_SPI && irq < NR_IRQS) {
+
+		/* read the current cpu affinity register */
+		uint32_t reg_value = getreg32(GIC_ICDIPTR(irq));
+
+		/* clear the bits for the current irq */
+		reg_value &= ~GIC_ICDIPTR_ID_MASK(irq);
+
+		/* write the target core set for the irq */
+		reg_value |= GIC_ICDIPTR_ID(irq, cpu_set);
+
+		/* write the updated value back to the register */
+		putreg32(reg_value, GIC_ICDIPTR(irq));	
+
+		return OK;
+	}
+
+	return -EINVAL;
+}
+
+/****************************************************************************
+ * Name: up_reset_irq_affinity
+ *
+ * Description:
+ *   reset the affinity of a given irq (shared peripheral interrupts - SPI only) to cpu 0.
+ *
+ * Input Parameters:
+ *   irq: irq id
+ *
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_reset_irq_affinity(int irq)
+{
+	/* reset affinity of the irq to core 0 */
+	return up_set_irq_affinity(irq, 0x01);
+}
+
+/****************************************************************************
  * Name: arm_gic_irq_trigger
  *
  * Description:
--- a/os/include/tinyara/arch.h
+++ b/os/include/tinyara/arch.h
@@ -1471,6 +1471,49 @@ int up_prioritize_irq(int irq, int priority);
 #endif
 
 /****************************************************************************
+ * Name: up_set_irq_affinity
+ *
+ * Description:
+ *   set the affinity of a given irq (shared peripheral interrupts - SPI only) to given cpu set
+ *
+ * Input Parameters:
+ *   irq: irq id
+ *   cpuset: cpu set
+ *   
+ *   Example usage: 
+ *   for cpu 1    -> up_set_irq_affinity(10, 0x10)
+ *   for cpu 0    -> up_set_irq_affinity(10, 0x01)
+ *   for cpu 0, 1 -> up_set_irq_affinity(10, 0x11)
+ * 
+ *   Each bit of the register represents each core affinity.
+ *   Note: register supports 8 bits per interrupt. (So maximum 8 cpus)
+ *
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_set_irq_affinity(int irq, cpu_set_t cpu_set);
+
+/****************************************************************************
+ * Name: up_reset_irq_affinity
+ *
+ * Description:
+ *   reset the affinity of a given irq (shared peripheral interrupts - SPI only) to cpu 0.
+ *
+ * Input Parameters:
+ *   irq: irq id
+ *
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_reset_irq_affinity(int irq);
+
+/****************************************************************************
  * Tickless OS Support.
  *
  * When CONFIG_SCHED_TICKLESS is enabled, all support for timer interrupts


os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
...
+	// read the current cpu affinity register
+	uint32_t reg_value = getreg32(GIC_ICDIPTR(irq));
+
+	// clear the bits for the current irq
+	reg_value &= ~(0xFF << ((irq % 4) * 8));
+
+	// write the target core for the irq
+	reg_value |= (target_cpu_bit << ((irq % 4) * 8));
+
+	// write the updated value back to the register
```
neel-samsung, line 543: could you please check if comment using ```//``` is allowed or not?

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
 	return -EINVAL;
 }
 
+/****************************************************************************
+ * Name: up_affinity_irq
+ *
+ * Description:
+ *   set the affinity of a given irq (SPI only) to given cpu.
+ *   Currenlty, it doesnot support assigning multiple cpus as affinity.
```
kishore-sn, line 516: doesnot -> does not

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
...
+ *   cpu: cpu id
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void up_affinity_irq(int irq, int cpu)
+{
+	DEBUGASSERT(irq >= GIC_IRQ_SPI && irq < NR_IRQS && cpu >= 0 && cpu < CONFIG_SMP_NCPUS);
```
kishore-sn, line 530: Let us split this into 2 lines for make debugging simpler. One line for irq and other for cpu.

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
...
+ * Input Parameters:
+ *   irq: irq id
+ *   cpu: cpu id
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void up_affinity_irq(int irq, int cpu)
```
kishore-sn, line 528: Let us change the name to `up_set _irq_affinity`

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
...
+	reg_value &= ~(0xFF << ((irq % 4) * 8));
+
+	/* write the target core for the irq */
+	reg_value |= (target_cpu_bit << ((irq % 4) * 8));
+
+	/* write the updated value back to the register */
+	putreg32(reg_value, GIC_ICDIPTR(irq));
+
+}
+
```
kishore-sn, line 547: Let us add another api `up_reset_irq_affinity`. This will set the irq affinity to core 0 only. This is the default setting.

os/arch/arm/src/armv7-a/arm_gicv2.c
kishore-sn, line 528: This api can be easily changed to handle multiple cpu affinity. Instead of using `int cpu` let us use `cpu_set_t`. The same is used for setting affinity of tasks. The cpu set data type will have one bit for each cpu. If the bit is set, then the irq can be taken to that particular cpu.

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,43 @@ int up_prioritize_irq(int irq, int priority)
...
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void up_affinity_irq(int irq, int cpu)
+{
+	DEBUGASSERT(irq >= GIC_IRQ_SPI && irq < NR_IRQS && cpu >= 0 && cpu < CONFIG_SMP_NCPUS);
+
```
kishore-sn, line 531: Even though we have added debugassert above, we need to add if condition here to check whether irq value is within range. If irq parameter is out of range, then we dont do anything in this api. Debug assert will be usefult only if the debug macro is enabled. But when the code goes into production, then this macro is not enabled. But we must still check the valid range of the irq parameter.

kishore-sn (dismissed):

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,92 @@ int up_prioritize_irq(int irq, int priority)
...
+	if (irq >= GIC_IRQ_SPI && irq < NR_IRQS) {
+
+		/* read the current cpu affinity register */
+		uint32_t reg_value = getreg32(GIC_ICDIPTR(irq));
+
+		/* clear the bits for the current irq */
+		reg_value &= ~(0xFF << ((irq % 4) * 8));
+
+		/* write the target core for the irq */
+		reg_value |= (cpu_set << ((irq % 4) * 8));
```
abhishek-samsung, line 544: Please use macros for this, 

https://github.com/apache/nuttx/blob/48b93b8dc807c99a8903827d20f23eb02286ad67/arch/arm/src/armv7-a/arm_gicv2.c#L615-L632

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,71 @@ int up_prioritize_irq(int irq, int priority)
...
+ *   cpuset: cpu set
+ *   Note: the API will only support 8 bit cpu set.
+ *
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_set_irq_affinity(int irq, cpu_set_t cpu_set)
```
sunghan-chang, line 528: Is there any example to show how to use?
If I want to assign irq 10 to cpu 1, how should I call this?
```
up_set_irq_affinity(10, ??);
```
What about cpu 0?

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,71 @@ int up_prioritize_irq(int irq, int priority)
...
+	}
+
+	return -EINVAL;
+}
+
+/****************************************************************************
+ * Name: up_reset_irq_affinity
+ *
+ * Description:
+ *   reset the affinity of a given irq (SPI only) to cpu 0.
```
sunghan-chang, line 559: What does `(SPI only)` mean?
Assign to cpu 0 is `reset`?

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,71 @@ int up_prioritize_irq(int irq, int priority)
...
+ ****************************************************************************/
+
+int up_set_irq_affinity(int irq, cpu_set_t cpu_set)
+{
+	DEBUGASSERT(irq >= GIC_IRQ_SPI && irq < NR_IRQS);
+	DEBUGASSERT(sizeof(cpu_set) == sizeof(uint8_t));
+
+	/* Ignore invalid interrupt IDs */
+
+	if (irq >= GIC_IRQ_SPI && irq < NR_IRQS) {
```
abhishek-samsung, line 535: this check looks redundant as we have DEBUGASSERT above with same condition.

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,71 @@ int up_prioritize_irq(int irq, int priority)
...
+ * Returned Value:
+ *   OK : operation successful
+ *   EINVAL: irq out of range
+ *
+ ****************************************************************************/
+
+int up_set_irq_affinity(int irq, cpu_set_t cpu_set)
+{
+	DEBUGASSERT(irq >= GIC_IRQ_SPI && irq < NR_IRQS);
+	DEBUGASSERT(sizeof(cpu_set) == sizeof(uint8_t));
```
abhishek-samsung, line 531: Does this mean this api will only for chips with NCPU less than or equal to 8?

https://github.com/Samsung/TizenRT/blob/8b5b5d3da6e6814785b060c7fac17e51b46a2c73/os/include/sys/types.h#L296-L306

sunghan-chang: Do we need to change the type? If we use uint32_t always, what happen? Just waste some memory maximum 3 bytes?

os/arch/arm/src/armv7-a/arm_gicv2.c
sunghan-chang: `//` is ok to use.

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: 1. SPI means shared peripheral interrupt - these are the interrupts that can be shared by each core
    PPI (private peripheral interrupt), SGI  - every core has one of their own set of these interrupts. These are not shared.

2. yes

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: > Does this mean this api will only for chips with NCPU less than or equal to 8?
> 
> https://github.com/Samsung/TizenRT/blob/8b5b5d3da6e6814785b060c7fac17e51b46a2c73/os/include/sys/types.h#L296-L306

currently from my understanding, the register to assign cpu affinity for a interrupt only supports 8 bits. It can only support upto 8 cpus.

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: for cpu 1     ->  up_set_irq_affinity(10, 0x10)
for cpu 0     ->  up_set_irq_affinity(10, 0x01)
for cpu 0, 1 ->  up_set_irq_affinity(10, 0x11)

Each bit of the register represents each core affinity.
and register supports 8 bits per interrupt.

os/arch/arm/src/armv7-a/arm_gicv2.c
sunghan-chang: 1. How about writing full words there to avoid misunderstanding?
  ` * reset..... (Shared.... Interrupt - SPI only) to cpu 0. `

os/arch/arm/src/armv7-a/arm_gicv2.c
sunghan-chang: could you add this in the function description?

os/arch/arm/src/armv7-a/arm_gicv2.c
sunghan-chang: @abhishek-samsung Is this enough for your query?
@ritesh55555 1. Could you check my comment above?
2. and releated Abhishek query, when CONFIG_SMP_NCPUS is 24 for example, this line is ok?

os/arch/arm/src/armv7-a/arm_gicv2.c
```diff
@@ -508,6 +508,78 @@ int up_prioritize_irq(int irq, int priority)
...
+ *   set the affinity of a given irq (shared peripheral interrupts - SPI only) to given cpu set
+ *
+ * Input Parameters:
+ *   irq: irq id
+ *   cpuset: cpu set
+ *   
+ *   Example usage: 
+ *   for cpu 1    -> up_set_irq_affinity(10, 0x10)
+ *   for cpu 0    -> up_set_irq_affinity(10, 0x01)
+ *   for cpu 0, 1 -> up_set_irq_affinity(10, 0x11)
```
sunghan-chang, line 524: This means any cpu 0 or 1 will get the interrupt? or both will get the same irq at the same time?
If any cpu 0 or 1 gets, how does the system determine to assign which cpu?

os/arch/arm/src/armv7-a/arm_gicv2.c
sunghan-chang: @ritesh55555 Could you check the comment above?

ritesh55555: regarding this, I followed the up_prioritize_api() format and added both the condition. 
Also if CONFIG_DEBUG is not enable in production , then DEBUGASSERT will not work. So one more if condition is necessary.

sunghan-chang: Then, how about removing DEBUGASSERT?
If the config is disabled, yes, you're right.
But if the config is enabled, it will be duplicated. SVACE will say always true.

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: 1. Only one cpu will get the interrupt at a time.
2. Default cpu selection method is round robin. The gic keeps track of last cpu that handled the interrupt and next interrupt goes to the next available cpu. 
But if one of the cpu is already handling other interrupt, it will go to another cpu. If both are busy, it will be in pending state.

when i verified this, i observed that interrupts were divided equally between both core.

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: done

os/arch/arm/src/armv7-a/arm_gicv2.c
ritesh55555: 1. yes, just waste some memory maximum 3 bytes. Else we can use uint32_t always.
2. anyway, i have removed debugassert line. But if a cpu_set of uint32_t will be passed, the upper 24 bits will be ignored.

kishore-sn (approved):

PR #6653, opened 2025-01-29
arch/armv7-a: fix assert in up_cpu_resume() with common pause API

neel-samsung:
This patch fix the assert issue happened on product with commit ("os/arch/armv7-a: add api to pause and resume all CPUs") If signaling task is scheduled on the current CPU, then we don't need to resume the other CPUS as we don't pause them before.

This patch is tested on both public and ecode. We are not able to reproduce the reported assert with the patch.


kishore-sn (approved):

ritesh55555 (approved):
looks good to merge

sunghan-chang (approved):

PR #6654, opened 2025-01-31
arch/arm/armv7-a: Modify to printing LR when NULL executing fault

ewoodev:


ewoodev:
The PR is from #6643.

os/binfmt/libelf/libelf_sections.c
```diff
@@ -268,6 +268,23 @@ void elf_show_all_bin_section_addr(void)
...
+bool elf_is_text_addr(void *addr)
+{
+	int bin_idx;
+	void *start_addr;
+	void *end_addr;
+	for (bin_idx = 0; bin_idx <= CONFIG_NUM_APPS; bin_idx++) {
+		if (g_bin_addr_list[bin_idx].text_addr != 0) {
+			start_addr = g_bin_addr_list[bin_idx].text_addr;
+			end_addr = start_addr + g_bin_addr_list[bin_idx].text_size;
+			if (start_addr < addr && addr < end_addr) {
```
kishore-sn, line 281: I think it must be `start_addr <= addr`

sunghan-chang:
@kishore-sn @r-prabu @abhishek-samsung Could you review this?

os/binfmt/libelf/libelf_sections.c
ewoodev: Thanks, Updated

os/binfmt/libxipelf/xipelf.c
```diff
@@ -257,3 +257,21 @@ void *elf_find_text_section_addr(int bin_idx)
...
+bool elf_is_text_addr(void *addr)
+{
+	int bin_idx;
+	void *start_addr;
+	void *end_addr;
+	for (bin_idx = 0; bin_idx <= CONFIG_NUM_APPS; bin_idx++) {
+		if (g_bin_addr_list[bin_idx].text_addr != 0) {
+			start_addr = g_bin_addr_list[bin_idx].text_addr;
+			end_addr = start_addr + g_bin_addr_list[bin_idx].text_size;
+			if (start_addr < addr && addr < end_addr) {
```
abhishek-samsung, line 270: similar function for elf checks <= and >= . Please match it here too.

ewoodev: thank you, Updated

os/arch/arm/src/armv7-a/arm_prefetchabort.c
```diff
@@ -80,7 +84,7 @@ static inline void print_prefetchabort_detail(uint32_t *regs, uint32_t ifar, uin
 	lldbg_noarg("\n");
 	_alert("#########################################################################\n");
 	_alert("PANIC!!! Prefetch Abort at instruction : 0x%08x\n",  regs[REG_PC]);
-	_alert("PC: %08x IFAR: %08x IFSR: %08x\n", regs[REG_PC], ifar, ifsr);
+	_alert("PC: %08x LR: %08x IFAR: %08x IFSR: %08x\n", regs[REG_PC], regs[REG_LR], ifar, ifsr);
```
kishore-sn, line 87: How about printing LR - 4 here?

os/arch/arm/src/armv7-m/up_usagefault.c
```diff
@@ -117,7 +117,7 @@ int up_usagefault(int irq, FAR void *context, FAR void *arg)
 	if (cfsr & INVPC) {
 		/* As PC value might be invalid use LR value to determine if
 		 * the crash occurred in the kernel space or in the user space */
-		system_exception_location = regs[REG_R14];
+		system_exception_location = regs[REG_R14] - 4;
```
kishore-sn, line 120: The above comment is wrong. INVPC does not mean that PC value is invalid. It means that there was an integrity check failure at the time of return from exception. The integrity check might fail due to various reasons. But, in any case, the PC value does not become invalid. So, I think we need to remove the whole if-statement.

os/arch/arm/src/armv8-m/up_usagefault.c
```diff
@@ -120,7 +120,7 @@ int up_usagefault(int irq, FAR void *context, FAR void *arg)
 	if (cfsr & INVPC) {
 		/* As PC value might be invalid use LR value to determine if
 		 * the crash occurred in the kernel space or in the user space */
-		system_exception_location = regs[REG_R14];
+		system_exception_location = regs[REG_R14] - 4;
```
kishore-sn, line 123: Same as above. We can remove the complete if-statement.

PR #6655, opened 2025-02-03
os/board/rtl8730e: Modify TX retry count API

LiuZhuo-Realtek:
- Modify implementation for obtaining WiFi TX Retry count
- Add a counter to break out from while loop to prevent stuck in infinite loop


hk-gwak (approved):
Done aging test(for 7 days, 2 times) with this commit and looks working well.

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_host_api_ext.c
```diff
@@ -1098,12 +1099,18 @@ unsigned int wifi_get_tx_retry(int idx)
 	DCache_Clean((u32)tx_retry_cnt, sizeof(tx_retry_cnt));
 	ipc_send_message(IPC_AP_TO_LP, IPC_A2L_WIFI_FW_INFO, &ipc_req_msg); 
 
-	while (1) {
+	while (try_cnt) {
 		DCache_Invalidate((u32)flag, sizeof(flag));
+		try_cnt --;
+		DelayUs(2);
```
hk-gwak, line 1105: Sorry for late comment, but
need to consider try_cnt == 1 case.
In that case, if (try_cnt == 0) clause will be entered unconditionally, even data is revealed.

jwei5: Thank you for the review, we will fix it on the next PR.

PR #6649, opened 2025-01-23
os/board/rtl8730e: Add BLE disconnect call back API trigger

ziyik:
-Add application disconnect API trigger for connection fail case
-Add application disconnect API trigger for pairing fail case

PR #6656, opened 2025-02-05
framework/ble_manager: Call disconnected callback when connection or pairing failed in client

giwon-nam:
- In the case where the LL connection establishment fails or pairing fails while entering the BLE_CLIENT_CONNECTING state after a connect request from the client, neither the connected callback nor the disconnected callback is called, resulting in continued waiting. Therefore, I modified it so that when connection or pairing fails, the disconnected callback is called to allow reconnection attempts.

Related PR : PR #6649


hs36-kim (approved):
LGTM
connecting상태의 ctx고려하도록 수정

PR #6657, opened 2025-02-05
ai-soc/ndp120: Add support for BT mic

cvejlbo:
Adds sample support for use of BT mic via `BT_MIC_SUPPORT`
Only tested with regular PDM mic.


Taejun-Kwon:
Thanks, let us test this.

PR #6658, opened 2025-02-06
os/kernel: Support silent reboot

jeongarmy:
- Support silent reboot driver for ioctl (lock, unlock, delay, checkmode)
- Set silent mode based on reboot reason
- Support periodic silent reboot when all the following conditions are satisfied.
  - Random time in (00:00 ~ 5:00 am) after 7 days from system boot.
  - No lock

### DIFF
--- a/os/drivers/silent_reboot/Make.defs
+++ b/os/drivers/silent_reboot/Make.defs
@@ -1,0 +1,25 @@
+##########################################################################
+#
+# Copyright 2024 Samsung Electronics All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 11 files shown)


os/drivers/silent_reboot/Make.defs
```diff
@@ -0,0 +1,25 @@
+##########################################################################
+#
+# Copyright 2024 Samsung Electronics All Rights Reserved.
```
sunghan-chang, line 3: 2025

os/kernel/init/os_bringup.c
```diff
@@ -264,6 +267,11 @@ static inline void os_do_appstart(void)
 {
 	int pid;
 
+#ifdef CONFIG_SILENT_REBOOT
+	silent_reboot_initialize();
+#endif
+	silent_reboot_driver_register();
```
sunghan-chang, line 273: Maybe, this should be in the conditional.

os/kernel/silent_reboot/silent_reboot.c
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return false;
+}
+
+static void silent_timer_callback(int argc, uint32_t cmd)
+{
+	switch (cmd) {
+		case SILENT_START:
+		case SILENT_RESTART:
```
sunghan-chang, line 182: should be the same indentation with switch

sunghan-chang (approved):

PR #6660, opened 2025-02-06
Flusing log uart

ewoodev:


os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1377,6 +1380,43 @@ int up_getc(void)
...
+ ****************************************************************************/
+void up_flush_console(void)
+{
+#ifdef CONFIG_UART4_SERIAL_CONSOLE
+	uint16_t nbyte;
+	irqstate_t flags = enter_critical_section();
+
+	/* To avoid duplicated calling up_flush_console(). */
+	if (g_log_flush_running) {
+		return;
```
zhongnuo-tang, line 1405: we should have leave_critical_section here.

ewoodev: Thank you, Updated!

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1377,6 +1380,44 @@ int up_getc(void)
...
+	if (g_log_flush_running) {
+		leave_critical_section(flags);
+		return;
+	}
+	g_log_flush_running = true;
+
+	do {
+		while (!LOGUART_Ready());
+		nbyte = uart_xmitchars(&CONSOLE_DEV);
+	} while (nbyte);
```
abhishek-samsung, line 1413: How does this work? Currently it looks like we still rely on uart interrupts to send data. But interrupts are disabled. Wont it be better to use lldbg path instead?

ewoodev: The IRQ is calling uart_xmitchars()

abhishek-samsung: I meant uart_xmitchars path is through drivers (will call the drivers send function and it also has some signalling logic to wakeup other threads), but at this point, we are no longer using the driver's path to print logs. Instead we use lldbg, so I asked the above question, why not use up_putc instead of driver send.

ewoodev: we need signalling logic for reloading

os/arch/arm/src/amebasmart/amebasmart_serial.c
ziyik: Since we have enter critical here, there won't be duplicate calling of up_flush_console(), right?

ewoodev: This prevents the same cpu-thread from calling this function several times.

sunghan-chang:
@amandeep-samsung Could you check armv8-m for the same? This PR has only armv7-a.

PR #6662, opened 2025-02-10
os/drivers/input/touchscreen.c: Add IOCTL for enabling and disabling touch

abhinav-s235:
Added two new IOCTL commands TSIOC_ENABLE and TSIOC_DISABLE to enable and disable the touch functionality in the touch driver.

### DIFF
--- a/apps/examples/touchscreen/touchscreen_main.c
+++ b/apps/examples/touchscreen/touchscreen_main.c
@@ -202,6 +202,42 @@ static int touchsceen_resume(void)
 	return OK;
 }
 
+static int touchscreen_enable(void)
+{
+	int ret = OK;
+	int fd = open(TOUCH_DEV_PATH, O_RDWR);
+	if (fd < 0) {
+		printf("Fail to open %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+		return ERROR;
+	}
+
+	ret = ioctl(fd, TSIOC_ENABLE, NULL);
+	if (ret != OK) {
+		printf("Fail to TSIOC_ENABLE %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+	}
+
+	close(fd);
+	return ret;
+}
+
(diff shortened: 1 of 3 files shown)


namanjain7:
LGTM

ewoodev (approved):

apps/examples/touchscreen/touchscreen_main.c
```diff
@@ -202,6 +202,38 @@ static int touchsceen_resume(void)
...
+		printf("Fail to open %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+		return ERROR;
+	}
+
+	if (ioctl(fd, TSIOC_ENABLE, NULL) != OK) {
+		printf("Fail to TSIOC_ENABLE %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+	}
+
+	close(fd);
+	return OK;
```
sunghan-chang, line 218: When we get a failure on the open, we return ERROR.
But when we get a fail on the ioctl, we return OK.
How about using a local variable `ret`?

apps/examples/touchscreen/touchscreen_main.c
```diff
@@ -202,6 +202,42 @@ static int touchsceen_resume(void)
...
+	int fd = open(TOUCH_DEV_PATH, O_RDWR);
+	if (fd < 0) {
+		printf("Fail to open %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+		return ERROR;
+	}
+
+	ret = ioctl(fd, TSIOC_ENABLE, NULL);
+	if (ret != OK) {
+		printf("Fail to TSIOC_ENABLE %s, errno:%d\n", TOUCH_DEV_PATH, get_errno());
+	}
```
sunghan-chang, line 217: Need to set the `ret` to ERROR

sunghan-chang (approved):

PR #6664, opened 2025-02-12
os/arch/arm/src/amebasmart: support multiple watchdogs

zhongnuo-tang:
1. WDG2 was originally for KM4, but since KM4 does not use watchdog, it can be used by CA32
    2. CA32 can use two watchdogs now, WDG4 and WDG2, both are non-secure watchdogs.
    3.  "/dev/watchdog1" for WDG2, and "/dev/watchdog0" for WDG4.
    4. up_wdog_init will use WDG4 if CONFIG_WATCHDOG_FOR_IRQ is enabled.


sunghan-chang:
What about reboot reason for this? We should know which watchdog causes a reset.

zhongnuo-tang:
> What about reboot reason for this? We should know which watchdog causes a reset.

Yes, do you think we should add new reboot reason in reboot_reason.h? or within amebasmart_reboot_reason.c, we will printout the which watchdog causes the reset similar to system reset. eg "lldbg("Reboot reason: REBOOT_SYSTEM_WATCHDOG_WDG2\n");

zhongnuo-tang:
Test results:
up_reboot_reason_get_hw_value: Reboot reason: WDG2 reset
up_initialize: [Reboot Reason] : 54

up_reboot_reason_get_hw_value: Reboot reason: WDG4 reset
up_initialize: [Reboot Reason] : 54

sunghan-chang:
@zhongnuo-tang Thank you for update. Let me discuss it internally and get back to you.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/wdt_api.c
```diff
@@ -100,17 +96,17 @@ void watchdog_init(uint32_t timeout_ms)
 	WDG_InitStruct.EICNT = timeout_ms > 100 ? 100 : timeout_ms / 2;
 	WDG_InitStruct.EIMOD     = ENABLE;
 
-	WDG_Init(WDGDev, &WDG_InitStruct);
+	WDG_Init(WDGArray[WDG_id], &WDG_InitStruct);
```
jeongarmy, line 99: I think we need to check WDG_id with bound of WDGArray to prevent out of memory.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/wdt_api.c
```diff
@@ -32,13 +32,8 @@
...
   * @{
   */
-#if defined (ARM_CORE_CM4)
-WDG_TypeDef *WDGDev = WDG2_DEV;
-IRQn_Type WdgIrqNum = WDG2_IRQ;
-#elif defined (ARM_CORE_CA32)
-WDG_TypeDef *WDGDev = WDG4_DEV;
-IRQn_Type WdgIrqNum = WDG4_IRQ;
-#endif
+WDG_TypeDef *WDGArray[] = {WDG2_DEV, WDG4_DEV};
```
jeongarmy, line 35: How about using max value of this array size like WDGArray[XXX_MAX] =

os/board/rtl8730e/src/component/mbed/hal_ext/wdt_api.h
```diff
@@ -44,21 +44,28 @@ extern "C" {
 ///@name Ameba Common
 ///@{
 
+typedef enum {
+  AMEBASMART_WDG2,
+  AMEBASMART_WDG4,
+  AMEBASMART_WDGMAX,
```
jeongarmy, line 50: oh we can use it for checking boundary

zhongnuo-tang: Thanks, updated to check for boundary.

PR #6665, opened 2025-02-12
build/configs/rtl8730e: Add BLE debug file generation script command

ziyik:
-Add BLE debug file (APP.trace) generation script command -APP.trace is used to decode BLE debug log


hs36-kim (approved):
LGTM

PR #6668, opened 2025-02-13
os/board/rtl8730e/src: support BOD and enable BOR

zhongnuo-tang:
1. enable brownout reset with default fall and rise threshold set
2. BOR fall threshold: 2.72V, BOR rise threshold: 2.83V
3. value should be configured according different voltage drop scenarios


zhongnuo-tang:
When BOR happens, reboot reason will be as follows. 
up_initialize: [Reboot Reason] : 252
REBOOT_SYSTEM_BOD_RESET      = REBOOT_BOARD_SPECIFIC3, /* Brownout reset */

os/arch/arm/src/amebasmart/Kconfig
```diff
@@ -75,6 +75,25 @@ config AMEBASMART_USBDEVICE
 	bool "AMEBASMART_USBDEVICE"
 	default n
 
+config AMEBASMART_BOR
+	bool "AMEBASMART_BOR"
+	default y
+
+if AMEBASMART_BOR
+config AMEBASMART_Threshold_Fall
```
sunghan-chang, line 83: Let's use upper characters all in the config name.

zhongnuo-tang: Changed to all upper case, thank you.

sunghan-chang (approved):

PR #6669, opened 2025-02-13
os/board/rtl8730e: Fix issue of stuck in wifi tx retry count API

jwei5:
- During aging test, code was stuck in while loop in wifi_get_tx_retry() due to "flag" variable not updated
- Resolve issue by: a) Optimizing wifi_get_tx_retry API to fix issue of "flag" variable not updated b) Add timeout counter in while loop to prevent getting stuck in infinite loop (added in PR #6655)
- Fix logic of while loop when try_cnt is == 1

This PR addresses the issue of wifi_netmgr_utils_get_signal_quality() not returning during aging test.
The cause came from wifi_get_tx_retry(), as there was a infinite while loop that only breaks when flag[0] is not zero.
To resolve this, a counter was added in the while loop so that the loop will still break if the variable remains zero (PR #6655).
As flag was previously a local variable, the address might change at every function call, which could cause cache coherency issues if the address became unaligned. This might be the reason the API got stuck during the aging test. To prevent this issue, these variables are aligned and set as static global variables so that the address is fixed at compile time. The API is optimized to only require one variable tx_retry_cnt, as one of the unused indexes in tx_retry_cnt can be used as flag[0].


hk-gwak:
Done aging test for 80 hours and it worked well.

hk-gwak (approved):

PR #6672, opened 2025-02-17
os/board/rtl8730e: Add BLE API for add service after BLE Init

ziyik:
-Allow BLE combo init without service
-Add API to add service after BLE init


j11591 (approved):
LGTM

PR #6673, opened 2025-02-17
os/board/rtl8730e: Change printf to debug_print in BLE client

yeetee179:
- changed printf to debug_print in rtk_gcs_client.c


hs36-kim (approved):
LGTM

PR #6674, opened 2025-02-17
 os/board/rtl8730e/src: implement recovery mechanism for PG flow using IWDG

zhongnuo-tang:
1. IWDG will be used to recovery, HW enabled using OTP.
2. IWDG ensures KM0 is working properly, KM0 and KM4 wakeup properly, AP power on properly. Once AP power on, watchdog in CA32 will resume counting and take over
3. Backup register will be used to record which core triggers the reset of IWDG during PG wakeup flow


zhongnuo-tang:
Simulation of wakeup fail from PG
up_reboot_reason_get_hw_value: Reboot reason: IWDG reset, CA32 wakeup failed

up_reboot_reason_get_hw_value: Reboot reason: IWDG reset, KM4 wakeup failed

zhongnuo-tang:
Implementation is as follows.

1. in rtl8730e_boot.c, IPC request will be sent to KM0 to enable IWDG, with timeout of 60s. 
2. IWDG can remain enable or disable during PG. If it is enable during PG, it will trigger wakeup, which we do not want it as wakeup source, so in our implementation it will be disabled in PG state.
3. IWDG once enabled, cannot be disabled in active state. Refer to our usermanual, IWDG's clock source is a no-calibration clock, with a margin of error of 50%, so it will be refreshed every 30s in KM0 during active state to ensure no timeout.
4. During wakeup flow, once hardware PMC wakes LP (KM0), IWDG will resume its counter and start counting, if it successfully wake to KM0, the timer will refresh IWDG every 30s.
5. KM0->KM4->CA32, so if KM4 is not able to be waken up or CA32's power is not able to power on during wakeup flow, it will enter a while loop, and prevent KM0 from refreshing the IWDG.
6. any of the three cases happen,(KM0 not wakeup, KM4 not wakeup, or CA32 not power on) it will cause KM0 not refreshing IWDG and reset happen after 60s.

I did not add CONFIG_PM macro to make it only enable for PM, because i think KM0 should have a watchdog to ensure it is running, if there is any issue with KM0, we should do reset too.

sunghan-chang:
@zhongnuo-tang Could you resolve the conflict? And please share your test result.

zhongnuo-tang:
> @zhongnuo-tang Could you resolve the conflict? And please share your test result.

Hi @sunghan-chang , this PR will need to be modified since now IWDG is enabled by HW.
The timeout can no longer be set by user, so it will be default 4s.
PG recovery mechanism flow will be the same, only changes will be timeout value.
After modification, I will share the test result again.

zhongnuo-tang:
> @zhongnuo-tang Could you resolve the conflict? And please share your test result.

Hi @sunghan-chang, I have removed the control using IPC to enable SW IWDG because IWDG can only be enabled either HW/SW, and once enabled it cannot be disabled anymore. 

Below is the test results:
Simulated scenario where KM4 core fails to power on properly during wake-up.
up_reboot_reason_get_hw_value: Reboot reason: IWDG reset, KM4 wakeup failed
up_initialize: [Reboot Reason] : 246

Simulated scenario where CA32 core fails to power on properly during wake-up.
up_reboot_reason_get_hw_value: Reboot reason: IWDG reset, CA32 wakeup failed
up_initialize: [Reboot Reason] : 246

Build error from ci/circleci looks not related to this PR.

os/arch/arm/src/amebasmart/amebasmart_reboot_reason.c
```diff
@@ -112,6 +118,16 @@ static reboot_reason_code_t up_reboot_reason_get_hw_value(void)
 		}
 		/* KM0: IWDG reset*/
 		else if (boot_reason & AON_BIT_RSTF_IWDG) {
+			/* KM0: IWDG is used in KM0 to ensure KM0 is working properly, KM4 is wakeup properly from PG, CA32 power is power-on from PG
+			* backup register will be used to record the case that triggered IWDG reboot
+			*/
+			if (boot_reason_reg2 == 0x2) {
```
sunghan-chang, line 124: Is this right to compare with the value? not bit masking?

os/arch/arm/src/amebasmart/amebasmart_reboot_reason.c
```diff
@@ -112,6 +118,16 @@ static reboot_reason_code_t up_reboot_reason_get_hw_value(void)
...
+				BKUP_Write(BKUP_REG2, 0);
+				lldbg("Reboot reason: IWDG reset, KM4 wakeup failed\n");
+			} else if (boot_reason_reg2 == 0x3) {
+				BKUP_Write(BKUP_REG2, 0);
```
sunghan-chang, line 128: We initialize the `BKUP_REG1` at the `up_reboot_reason_init`. How about doing this there?

zhongnuo-tang: BKUP_REG1 is used by Samsung for abort/asset reason code.. That is why we used BKUP_REG2 for specific board reason that happen in KM4 or secure image.

os/arch/arm/src/amebasmart/amebasmart_reboot_reason.c
zhongnuo-tang: When designing amebasmart's reboot reason, I made reference to amebadD for up_reboot_reason_write which writes the reboot reason code directly into BKUP register instead of bit wise operation.
So for amebasmart's km4 fault or secure fault, it followed the same implementation. As i checked in reboot_reason.h, there is no multiple reboot reason code.

Dane-Kang:
Hi @ewoodev 
Here is the description for IWDG process.
- Default KM0 already have a thread for feed IWDG .
- This PR, Added below implementation : 
    + If KM4 or CA32 cannot wake up, KM0 will not feed IWDG and cause reset.
    + KM0 check the register flag which indicate to finish wake up. 
    + If KM4 core and CA32 didn't set the flag , KM0 will not refresh IWDG and cause to trigger reset.
Until so far, @zhongnuo-tang  tested long run test with PG-Wake up with a few seconds interval.
And, now he will start to long run test with more high stress PG scenario (PG->wake up in 500ms or 300ms)

zhongnuo-tang:
Hi @sunghan-chang  @ewoodev 
The stress test is running for 3 days, and no issue identified. 
PG->wakeup with 100ms HW timer and removal the tash_pm_timedsuspend(TASH_PM_TIMEDSUSPEND_TIME_IN_MS) tash_main.

sunghan-chang:
@zhongnuo-tang Could you confirm the km4 is latest or not? If yes, let me merge.

Dane-Kang:
> @zhongnuo-tang Could you confirm the km4 is latest or not? If yes, let me merge.

@sunghan-chang , I just confirmed KM4 version in this PR is latest .

And Please remember that KM4 image need to change if changes the IWDG timeout value from default value(4s).
(For this, i informed via email)

sunghan-chang (approved):

PR #6678, opened 2025-02-18
os/board/rtl8730e: Only disable SWD when PA_14 is in use

ziyik:
-Only disable SWD when PA_14 is in use
-Prevent all LCD model to disable SWD


sunghan-chang (approved):

sunghan-chang:
@anjana348 @namanjain7 Could you check the PR?

anjana348:
> @anjana348 @namanjain7 Could you check the PR?

@sunghan-chang We need to check on st7701SN 4.3 inch LCD , right?

@ziyik - This change is for all LCD except 3.9inch st7701 lcd?

namanjain7:
We have verified the PR in ST7701SN lcd.
The lcd test is working fine.

PR #6679, opened 2025-02-19
os/drivers/lcd, os/board/rtl8730e: modify power on/off sequence

anjana348:
Backlight on/off sequence updated as per correct order

It is verified with power off/on test

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -344,16 +344,18 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 		lcddbg("Power exceeds CONFIG_LCD_MAXPOWER %d", CONFIG_LCD_MAXPOWER);
 		return -1;
 	}
+	if (power == priv->power) {
+		return OK;
+	}
 	while (sem_wait(&priv->sem) != OK) {
 		ASSERT(get_errno() == EINTR);
 	}
 	if (power == 0) {
+		priv->config->backlight(power);
 		lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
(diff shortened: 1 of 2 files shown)


sunghan-chang:
@anjana348 Could you write details in the commit description?
"What is correct order for on and off"

And could you leave verification result? Is this tested?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -344,27 +344,27 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
...
+		priv->config->power_off();
 	} else {
 		/* The power on must operate only when LCD is OFF */
 		if (priv->power == 0) {
 			priv->config->power_on();
 			/* We need to send init cmd after LCD IC power on */
 			send_init_cmd(priv, lcd_init_cmd_g);
 		}
-		lcm_setting_table_t display_on_cmd = {0x29, 0, {0x00}};
-		send_cmd(priv, display_on_cmd);
```
sunghan-chang, line 365: Could you let me know why this is removed?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -344,27 +344,27 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 		lcddbg("Power exceeds CONFIG_LCD_MAXPOWER %d", CONFIG_LCD_MAXPOWER);
 		return -1;
 	}
+	if (power == 0 && priv->power == 0) {
```
sunghan-chang, line 347: What about other value? 
For example, power = 10, priv->power = 10.
I suggest
```
if (power == priv->power) {
   return OK;
}
```

sunghan-chang (approved):

PR #6684, opened 2025-02-19
Initial commits of mnemofs

abhishek-samsung:


os/fs/inode.h
```diff
@@ -0,0 +1,384 @@
+/****************************************************************************
```
Taejun-Kwon, line 1: I think this file added wrongly. let's remove this.

PR #6692, opened 2025-02-20
os/board/rtl8721csm: Sync the BLE APIs table

ziyik:
-Adding NULL to sync BLE APIs table
-To prevent APIs mismatch


hs36-kim (approved):
LGTM
for TP1x chipset

PR #6695, opened 2025-02-24
os/board/rtl8730e: Add BLE TX power setting API

yeetee179:
- added ble_server_set_adv_tx_power to set TX power for multi adv


hs36-kim (approved):
LGTM

PR #6697, opened 2025-02-25
os/driver/input/ist415: Change to power off when touch disabling

ewoodev:
Change to tunr off ist415 power when touch is disabled from apps for reduce power consumption.

### DIFF
--- a/os/drivers/input/ist415.c
+++ b/os/drivers/input/ist415.c
@@ -534,6 +534,11 @@ static int ist415_suspend_device(struct touchscreen_s *upper)
 		ASSERT(get_errno() == EINTR);
 	}
 
+	if (dev->forcedoff) {
+		sem_post(&dev->sem);
+		return EPERM;
+	}
+
 	dev->suspend = true;
 
 	if (dev->knockknock) {
@@ -568,6 +573,11 @@ static int ist415_resume_device(struct touchscreen_s *upper)
 		ASSERT(get_errno() == EINTR);
 	}
 
+	if (dev->forcedoff) {
+		sem_post(&dev->sem);
+		return ESHUTDOWN;
+	}
+
 	dev->suspend = false;
 
 	if (dev->knockknock) {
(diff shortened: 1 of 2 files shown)


IMJuBBang:
Please add the ist415_forced_release function when you turn off the ist415 power.

os/drivers/input/ist415.c
```diff
@@ -568,6 +573,11 @@ static int ist415_resume_device(struct touchscreen_s *upper)
 		ASSERT(get_errno() == EINTR);
 	}
 
+	if (dev->forcedoff) {
+		sem_post(&dev->sem);
+		return EPERM;
```
Taejun-Kwon, line 578: EPERM is tightly related to permission, so I think ESHUTDOWN or ENXIO is good for this situation, I prefer ESHUTDOWN

os/drivers/input/ist415.c
```diff
@@ -666,7 +692,10 @@ static void ist415_lockup_work(struct ist415_dev_s *dev)
 		ist415_disable(dev);
 		ist415_forced_release(dev);
 		ist415_reset(dev, false);
-		if (dev->suspend) {
+
+		if (dev->forcedoff) {
+			ist415_power_off(dev);
+		} else if (dev->suspend) {
```
Taejun-Kwon, line 698: if both of forceoff & suspend never set as true, I think you can define enum regarding state and use dev->state instead of these 3 states.

os/drivers/input/ist415.c
```diff
@@ -568,6 +573,11 @@ static int ist415_resume_device(struct touchscreen_s *upper)
 		ASSERT(get_errno() == EINTR);
 	}
 
+	if (dev->forcedoff) {
```
sunghan-chang, line 576: How about checking this at start of the function? Then, you don't need to take the sem.

os/drivers/input/ist415.c
```diff
@@ -534,6 +534,11 @@ static int ist415_suspend_device(struct touchscreen_s *upper)
...
```
sunghan-chang, line 537: same as previous

sunghan-chang (approved):

PR #6698, opened 2025-02-26
os/tools: fix condition for checking max l2 pages needed for binary

abhishek-samsung:
we need only 4 l2 pages at maximum for each app. 2 pages for the start and end of read only region, and two pages for the start and end of write-read region. Hence, maximum number is 4.


sunghan-chang (approved):

samsung-singh (commented):
fixed condition for checking max l2 pages needed for binary good to merge.

PR #6699, opened 2025-02-28
os/board/rtl8730e/src/compoent: add critical section to GPIO write bit

zhongnuo-tang:
1. With SMP/AMP, there might be race condition when both Core 0 and Core 1 attempt to modify the GPIO simultaneously, causing the the value pin to be abnormal
2. enter_critical_section is required to ensure that only one core can modify the GPIO register at any given time, we can prevent simultaneous access to the GPIO data register, thus eliminating the race condition.

### DIFF
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/rom_common/ameba_gpio_rom.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/rom_common/ameba_gpio_rom.c
@@ -369,6 +369,9 @@ void GPIO_WriteBit(u32 GPIO_Pin, u32 Pin_State)
 	u8 pin_num;
 	u32 RegValue;
 
+#ifdef CONFIG_PLATFORM_TIZENRT_OS
+	irqstate_t flags = enter_critical_section();
+#endif
 	port_num = PORT_NUM(GPIO_Pin);
 	pin_num = PIN_NUM(GPIO_Pin);
 	GPIO = GPIO_PORTx[port_num];
@@ -380,6 +383,9 @@ void GPIO_WriteBit(u32 GPIO_Pin, u32 Pin_State)
 		RegValue |= BIT(pin_num);
 	}
 	GPIO->PORT[0].GPIO_DR = RegValue;
+#ifdef CONFIG_PLATFORM_TIZENRT_OS
+	leave_critical_section(flags);
+#endif
 }
 
 /**


sunghan-chang:
@zhongnuo-tang Is this the SPI issue?

jsdosa:
@zhongnuo-tang With this logic added, will there be any performance impact?

zhongnuo-tang:
> @zhongnuo-tang Is this the SPI issue?

Yes, as SPI uses manual CS control using GPIO.

zhongnuo-tang:
> @zhongnuo-tang With this logic added, will there be any performance impact?

it will not impact the performance, GPIO write bit only consists ~16 instructions.

jeongarmy:
Have you verified that the SPI issue related to resource fs has been resolved?
Please add commit message about resolving problems and verification.

sunghan-chang:
@zhongnuo-tang Is it possible to support and to change to auto cs instead of GPIO control?

os/board/rtl8730e/src/component/soc/amebad2/fwlib/rom_common/ameba_gpio_rom.c
```diff
@@ -369,6 +369,9 @@ void GPIO_WriteBit(u32 GPIO_Pin, u32 Pin_State)
 	u8 pin_num;
 	u32 RegValue;
 
+#ifdef CONFIG_PLATFORM_TIZENRT_OS
```
sunghan-chang, line 372: Could you let me know why we need this conditional?
Other OS does not need?

zhongnuo-tang: Hi, other OS also needs it, just that our default SDK freertos is using a different API for enter critical, so i think it is better to differentiate it.

ZhenBei-Sin: Suggestion:
If the issue occurs with SMP/AMP, using this macro CONFIG_SMP to wrap is more appropriate.
CONFIG_AMP is already covered since it depends on CONFIG_SMP [=y].

sunghan-chang: @zhongnuo-tang @ZhenBei-Sin Could you let me know final your opinion for this?

zhongnuo-tang: Hi @sunghan-chang , keep it unchanged.

sunghan-chang (approved):

PR #6702, opened 2025-03-04
apps/examples/lcd_test/: Add bimap shade and lcdtest power on off feature

namanjain7:
apps/examples/lcd_test/: Add light and dark shade to bitmap and power off and power on feature via terminal in lcd_test

### DIFF
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -56,6 +56,8 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
 #include <errno.h>
 
 #define LCD_DEV_PATH "/dev/lcd%d"
@@ -423,6 +425,21 @@ static void test_fps(void)
 	}
 }
 
+bool is_valid_power(char *power)
+{
+	int power_val_size = strlen(power);
+
+	if (power_val_size < 1 || power_val_size > 3) {		/* Length of Power val should be 1, 2, or 3 */
+		return false;
+	}
+	for (int i = 0; i < power_val_size; i++) {
+		if (!isdigit(power[i])) {	/* If not a digit */
+			return false;
+		}
+	}
+	return true;
+}
+
 #ifdef CONFIG_BUILD_KERNEL
 int main(int argc, FAR char *argv[])
 #else
@@ -441,6 +458,21 @@ int lcd_test_main(int argc, char *argv[])
 		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
 		return ERROR;	
 	}
+
+	/* LCD Power test */
+	if (argc >= 2 && !strncmp(argv[1], "power", 5)) {
+		if (argc > 2 && is_valid_power(argv[2])) {
+			ioctl(fd, LCDDEVIO_SETPOWER, atoi(argv[2]));
+		} else {
+			printf("ERROR: Value of power should be int in range [0, 100]\n");
+			printf("Usage: lcd_test power <value>\n");
+			printf("0 --> LCD Power OFF\n");
+			printf("100 --> LCD Power ON\n");
+		}
+		close(fd);
+		return OK;
+	}
+
 	while (count < 5) {
 		test_put_area_pattern();
 		test_quad();
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -235,6 +235,7 @@ static void test_bit_map(void)
         uint8_t *lcd_data = (uint8_t *)malloc(len);
         if (lcd_data == NULL) {
                 printf("malloc failed for lcd data : %d\n", len);
+		close(fd);
                 return;
         }
 	area.planeno = 0;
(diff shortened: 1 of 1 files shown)


apps/examples/lcd_test/example_lcd.c
```diff
@@ -244,15 +257,35 @@ static void test_bit_map(void)
 	area.col_end = xres - 1;
 	area.stride = 2 * xres;
 	area.data = lcd_data;
-	uint16_t color;
+	uint16_t colorl, colord, colornormal;	// stores light and dark color pixel value
+	bool islight = true;
 	for (int y = 0; y < yres / SIZE * 2; y++) {
 		for (int x = 0; x < xres / SIZE; x++) {
-			color = generate_color_code(rand() % 31, rand() % 31, rand() % 31);
+			colornormal = generate_color_code(rand() % 31, rand() % 31, rand() % 31, true, false);
```
ewoodev, line 264: Could you please add annotation about codes?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -202,12 +202,22 @@ static void test_put_area_pattern(void)
 	sleep(3);
 }
 
-static unsigned short generate_color_code(int red, int green, int blue)
+static unsigned short generate_color_code(int red, int green, int blue, bool islight)
 {
+	int R, G, B;
```
ewoodev, line 208: Could you please update commit comment also?

namanjain7: I have added a description if it in commit comment.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -262,6 +264,65 @@ static void test_bit_map(void)
...
+	area.col_end = xres - 1;
+	area.stride = 2 * xres;
+	area.data = lcd_data;
+	int pixel_index = 0;
+	uint16_t color;
+
+	for (int y = 0; y < yres; y++) {
+		for (int x = 0; x < xres; x++) {
+			pixel_index = ((y * xres) + x ) * 2;
+			if (x < xres/2) {
```
jeongarmy, line 304: Please apply coding rules

jeongarmy: Please add spaces to before and after the symbols

apps/examples/lcd_test/example_lcd.c
```diff
@@ -381,8 +442,16 @@ int lcd_test_main(int argc, char *argv[])
 		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
 		return ERROR;	
 	}
+
+	if (argc > 1 && !strncmp(argv[1], "power", 6)) {
```
jeongarmy, line 446: I think it should be 5, not 6

jeongarmy: And the condition for argc should be `argc > 2` or `argc == 3`

apps/examples/lcd_test/example_lcd.c
```diff
@@ -262,6 +264,65 @@ static void test_bit_map(void)
...
 }
 
+static void test_quad(void)
+{
+	int fd = 0;
+	int p = 0;
+	char port[20] = {'\0'};
+	struct lcddev_area_s area;
+	size_t len;
+	sprintf(port, LCD_DEV_PATH, p);
```
jeongarmy, line 274: Please use snprintf

ewoodev (approved):

apps/examples/lcd_test/example_lcd.c
```diff
@@ -441,6 +442,12 @@ int lcd_test_main(int argc, char *argv[])
 		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
 		return ERROR;	
 	}
+
+	if (argc > 2 && !strncmp(argv[1], "power", 5)) {
+		ioctl(fd, LCDDEVIO_SETPOWER, atoi(argv[2]));
```
sunghan-chang, line 447: If non-integer is given to argv[2], what happen? No issue? Could you confirm that?

sunghan-chang: @namanjain7 Non- integer value should be invalid input, but it will working as 0. Then, it looks strange.

namanjain7: I have added a condition.
the condition will check if power is int between [0, 100].

apps/examples/lcd_test/example_lcd.c
```diff
@@ -56,6 +56,8 @@
 #include <sys/types.h>
 #include <fcntl.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
```
namanjain7, line 60: the ctype is for isdigit() function used to check if power is valid or not.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -381,8 +462,25 @@ int lcd_test_main(int argc, char *argv[])
 		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
 		return ERROR;	
 	}
+
+	/* LCD Power test */
+	if (argc >= 2 && !strncmp(argv[1], "power", 5)) {
+		if (argc > 2 && is_valid_power(argv[2])) {
```
namanjain7, line 468: This extra argc condition is required.
Usage: lcd_test power <value>

if user enters: lcd_test power
and dones't specify value. Then it should give error. This argc is handling this.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -363,6 +425,25 @@ static void test_fps(void)
...
+	int power_val_size = strlen(power);
+
+	if (power_val_size < 1 || power_val_size > 3) {		/* Length of Power val should be 1, 2, or 3 */
+		return false;
+	}
+	for (int i = 0; i < power_val_size; i++) {
+		if (!isdigit(power[i])) {	/* If not a digit */
+			return false;
+		}
+	}
```
sunghan-chang, line 439: Is this necessary? If we call atoi directly, what happen? What can we prevent any wrong operation?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -363,6 +425,25 @@ static void test_fps(void)
...
+	int power_val = atoi(power);
+	if (power_val < 0 || power_val > 100) {	/* Value of power should be in range [0, 100] */
+		return false;
+	}
+	return true;
```
ewoodev, line 444: the lcd_setpower is already returning error if power val input is over than CONFIG_LCD_MAXPOWER.
so, we don't need to check this value here.

apps/examples/lcd_test/example_lcd.c
ewoodev: if we enter input not integer char, the atoi will return 0. it will make lcd off.
for avoid it, this checking digital looks good.

namanjain7:
fixed.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -363,6 +425,22 @@ static void test_fps(void)
...
+
+	if (power_val_size < 1 || power_val_size > 3) {		/* Length of Power val should be 1, 2, or 3 */
+		return false;
+	}
+	for (int i = 0; i < power_val_size; i++) {
+		if (!isdigit(power[i])) {	/* If not a digit */
+			return false;
+		}
+	}
+	int power_val = atoi(power);
```
ewoodev, line 440: Please remove un-used value

sunghan-chang (approved):

PR #6703, opened 2025-03-05
os/drivers/lcd/: Set LCD to black screen when LCD power on

namanjain7:
After LCD is power on, the LCD displays the last displayed data on screen. This commit now sets the screen to black color after LCD power ON


sunghan-chang:
@namanjain7 We will apply PR #6713 for this. Let's close this.

PR #6705, opened 2025-03-06
os/board/rtl8730e/src/component: Add critical section in GDMA channel…

zhongnuo-tang:
… allocate

1. In SMP, the same channel might be allocated to both cores if they are access it simultaneously
2. add critical section to ensure only one core can access at one time

### DIFF
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_gdma_ram.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_gdma_ram.c
@@ -730,7 +730,9 @@ GDMA_ChnlAlloc(u32 GDMA_Index, IRQ_FUN IrqFun, u32 IrqData, u32 IrqPriority)
 	u8 ValTemp = 0;
 
 	assert_param(IS_GDMA_Index(GDMA_Index));
-
+#ifdef CONFIG_SMP
+	irqstate_t flags = enter_critical_section();
+#endif
 	/* Match idle channel. */
 	ValTemp = HAL_READ8(SYSTEM_CTRL_BASE_LP, REG_LSYS_BOOT_REASON_SW + 3);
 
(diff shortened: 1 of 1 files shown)


os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_gdma_ram.c
```diff
@@ -730,7 +730,9 @@ GDMA_ChnlAlloc(u32 GDMA_Index, IRQ_FUN IrqFun, u32 IrqData, u32 IrqPriority)
 	u8 ValTemp = 0;
 
 	assert_param(IS_GDMA_Index(GDMA_Index));
-
+#ifdef CONFIG_SMP
+	irqstate_t flags = enter_critical_section();
```
sunghan-chang, line 734: 1. Is this required enter_critical_section? not dma lock like sem or mutex?
2. Is there any issue from this?

zhongnuo-tang: 1. I think mutex/sem should also be fine, the purpose is to prevent multi-core accessing this at the same time. Let me think about how to modify it.
2. GDMA channel is not dedicated to each peripheral, so it will search for available channel and allocate it to the respective peripheral during initialization. There is a chance where multi core access, the same free channel is allocatecd to different peripherals and this will cause issue.

zhongnuo-tang: Hi @sunghan-chang ,
After discussion, I still think enter_critical_section should be used, because channel allocate/free might be called in ISR, so sem or mutex shouldnt be used..

sunghan-chang:
@zhongnuo-tang Could you let us know how verify (test)?

PR #6706, opened 2025-03-07
binary_manager: Support new APIs for partition swap

jeongarmy:
1. binary_manager_get_inactive_info_all : Get info of all registered binaries in inactive(non-running) partition

2. binary_manager_swap_partition : Update bootparam data to swap opposite partition if all valid binaries exist in opposite partition


os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -422,3 +425,96 @@ void binary_manager_set_bpidx(uint8_t index)
...
+	if (ret < 0) {
+		bmdbg("Fail to find valid kernel binary, %d\n", ret);
+		goto send_response;
+	}
+	update_bp_data.active_idx ^= 1;
+
+#ifdef CONFIG_RESOURCE_FS
+	/* Update bootparam if valid resource binary exists */
+	ret = binary_manager_check_resource_update(false);
+	if (ret < 0) {
```
sunghan-chang, line 475: When we get failure at somewhere, kernel update or resource update or user update, can we know which gets failure?

PR #6707, opened 2025-03-07
(DO NOT MERGE) os/arch/arm/src: runtime CPU1 power up/down in SMP/AMP environment

lhenry-realtek:
This commit introduces experimental testing feature of powering down/up CPU1 at runtime via tash commands.

To test, use flat_dev configurations and turn on the SMP example then build:
* Application Configuration > Examples > SMP Example [*]

Run the following commands to turn off the cpu, then on: TASH>cpu1off
TASH>cpu1on

To view cpu state:
TASH>cpustate

Test Cases:
* Default: 
TASH>smp will run and all threads exit properly

* CPU1 turned off: TASH>cpu1off
TASH>smp will hang due to the reason below

* CPU1 turned back on: TASH>cpu1off
TASH>cpu1on
TASH>smp will run and all threads exit properly

Notes:
* SCU startup flow may not be correct and requires further investigation. cp15_invalidate_dcache_all was moved to invalidate after setting cache bit, but this may not be correct! Without this change, arm_enable_smp(1) will cause a prefetch abort in CPU1 during runtime startup!
* Also, the OS does not currently know that CPU1 is dead when powered down and will enter deadlock if attempting to schedule a task on CPU1, work needs to be done on this end
* There is a minimum time that the CPU requires to initialize all its function! Rapid on/off cycles will LIKELY cause the cpu to become unstable!!!!

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_smp.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_smp.c
@@ -22,6 +22,13 @@ extern void _boot(void);
 extern void vPortRestoreTaskContext(void);
 #endif
 
+/* from psci.h aff_info_state_t */
+typedef enum {
+	AFF_STATE_ON = 0,
+	AFF_STATE_OFF = 1,
+	AFF_STATE_ON_PENDING = 2
+} aff_info_state_t;
+
 /*-----------------------------------------------------------*/
 
 void rtk_core1_power_on(void)
@@ -106,6 +113,83 @@ bool vPortGateOtherCore(void)
 	}
 	return true;
 #endif
+}
+
+void vPortSecondaryStart(void)
+{
+	int count = 10;
+	int state;
+
+	/* take the lock, as we are about to put cpu1 idle task back active */
+	sched_lock();
+	
+	/* turn on CPU1's power domain, require time to stabilize */
+	rtk_core1_power_on();
+	DelayMs(1);
+
+	/* renable scu and invalidate tags on all processors */
+	arm_enable_smp(0);
+
+	/* cold-boot the core */
+	lldbg("Booting secondary core... \n");
+	BaseType_t err = psci_cpu_on(SECONDARY_CORE_ID, (unsigned long)__cpu1_start);
+	DEBUGASSERT(err >= 0);
+
+	/* await for PSCI to report cpu state */
+	do {
+		state = psci_affinity_info(1, 0);
+		if (state == AFF_STATE_ON) {
+			break;
+		}
+
+		DelayUs(50);
+	} while (count--);
+
+	if (count <= 0) {
+		lldbg("Secondary core boot timeout affinfo: %d\n", state);
+	} else {
+		lldbg("Secondary core booted?... %d\n", err);
+		up_set_cpu_state(SECONDARY_CORE_ID, CPU_RUNNING);
+		
+		/* fire SGI1 to kick the freshly booted CPU into TizenRT Idle Task */
+		os_smp_start();
+
+		/* set the idle task state of cpu1 back to running, after it has booted */
+		os_set_cpu_idle_task(SECONDARY_CORE_ID, TSTATE_TASK_RUNNING);
+	}
+
+	/* unlock the scheduler */
+	sched_unlock();
+}
+
+void smp_full_powerdown_cpu(void)
+{
+	int count = 10;	// 10 * 50us = 500us
+	int state;
+
+	/* take the lock, as we are about to kill cpu1 */
+	sched_lock();
+
+	/* send SGI4 to make core enter halt state (not complete powerdown yet) */
+	up_cpu_haltcore(SECONDARY_CORE_ID);
+
+	/* release the lock, cpu1 is no longer part of SMP and the idle task is marked inactive */
+	sched_unlock();
+	while (up_get_cpu_state(SECONDARY_CORE_ID) != CPU_HALTED);
+
+	/* wait for PSCI to report that CPU1 is now safe to power off */
+	do {
+		state = psci_affinity_info(1, 0);
+		if (state == AFF_STATE_OFF) {
+			/* turn off CPU1's power domain */
+			rtk_core1_power_off();
+			return;
+		}
+
+		DelayUs(50);
+	} while (count--);
+
+	lldbg("Secondary core shutdown complete: %d\n", up_get_cpu_state(SECONDARY_CORE_ID));
 }
 
 void vPortWakeOtherCore(void)
--- a/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
+++ b/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
@@ -87,12 +87,35 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 
 	/* Set secondary core to hotplug state */
 	int cpu = this_cpu();
-	up_set_cpu_state(cpu, CPU_HOTPLUG);
-	/* Save the tcb state */
+	
 	struct tcb_s *rtcb = this_task();
-	arm_savestate(rtcb->xcp.regs);
-	rtcb->task_state = TSTATE_TASK_ASSIGNED;
+	if (up_get_cpu_state(cpu) == CPU_HALTED) {
+		/* 
+			clear GIC APR0 as when this gets set during restore by ATF boot flow,
+			When this register is set, it cause GIC to not serve other requests as it thinks there is an
+			active ISR being served currently.
+			
+			On powerup, this register is restored by ATF from previous saved state. 
+			If hard reset, then state is re-initialized, otherwise it will save the state when psci_cpu_off
+		*/
+		putreg32(0x00000000, GIC_ICCNSAPR1);	/* clear SGI1 nonsecure APR0 */
+		putreg32(0x00000000, GIC_ICCAPR1);	/* clear SGI1 secure APR0 */
+	} else {
+		up_set_cpu_state(cpu, CPU_HOTPLUG);
+		/* Save the tcb state */
+		arm_savestate(rtcb->xcp.regs);
+	}
+
+	/* mark this cpu's idle task as inactive to prevent the scheduler from putting things on it */
+	rtcb->task_state = TSTATE_TASK_INACTIVE;
 	CURRENT_REGS = NULL;
+
+	/* 
+		manually clear SGI4 interrupt active
+		As psci_cpu_off is a noreturn function it will trap the CPU in a WFI loop 
+		this means that it will never exit back to arm_decodeirq where EOIR will be set
+	*/
+	putreg32(irq, GIC_ICCEOIR);
 
 	/* MINOR: PSCI is a general interface for controlling power of cortex-A cores 
 	   It is one of the feature in arm-trusted-firmware (ie. Trustzone-A)
@@ -101,7 +124,7 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 	   booting the secondary core, thus we can ignore the return value here first
 	*/
 	/* Shut down the secondary core */
-   	(void)psci_cpu_off(); 
+   	(void)psci_cpu_off();
 	return OK;
 }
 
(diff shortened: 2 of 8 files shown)


os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -87,10 +87,24 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
...
+		/* 
+			clear GIC APR0 as when this gets set during restore by ATF boot flow,
+			When this register is set, it cause GIC to not serve other requests as it thinks there is an
+			active ISR being served currently.
+			
+			On powerup, this register is restored by ATF from previous saved state. 
+			If hard reset, then state is re-initialized, otherwise it will save the state when psci_cpu_off
+		*/
+		putreg32(0x00000000, GIC_ICCNSAPR1);	/* clear SGI1 nonsecure APR0 */
+		putreg32(0x00000000, GIC_ICCAPR1);	/* clear SGI1 secure APR0 */
```
eunwoo-nam, line 102: Hello henry 
shoud not reqiured clear SGI ARP when hotplug?

os/arch/arm/src/armv7-a/arm_scu.c
```diff
@@ -187,14 +187,16 @@ void arm_enable_smp(int cpu)
...
-		cp15_invalidate_dcache_all();
+		
 #ifdef CONFIG_ARCH_CORTEXA32
 		regval = arm_get_sctlr();
 		regval |= SCTLR_C;
 		ARM_DSB();
 		arm_set_sctlr(regval);
 		ARM_ISB();
 #endif
+		cp15_invalidate_dcache_all();
```
eunwoo-nam, line 198: it seem there is no calling arm_enable_smp(1);
is it okay?

lhenry-realtek: Hi @ewoodev 

arm_enable_smp(1) should be called through early assembly init https://github.com/Samsung/TizenRT/blob/24092db7875684d8470968620ef91f9f45d31445/os/arch/arm/src/armv7-a/arm_cpuhead.S#L460-L463

https://github.com/Samsung/TizenRT/blob/24092db7875684d8470968620ef91f9f45d31445/os/arch/arm/src/amebasmart/amebasmart_cpuboot.c#L133-L139

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
lhenry-realtek: Hi @ewoodev 

Based on the last finding, this is the issue summary

SGI APR/NSAPR require to be cleared because when target CPU (CPU1) execute psci_cpu_off, this happens inside SGI4 handler. This means that CPU1 is now in WFI loop waiting to power down, while in IRQ context

The regular IRQ flow will clear GIC_EOIR in arm_decodeirq:
https://github.com/Samsung/TizenRT/blob/24092db7875684d8470968620ef91f9f45d31445/os/arch/arm/src/armv7-a/arm_gicv2.c#L389-L395

However because the CPU1 is WFI, the EOIR will never complete. This means that the GIC will still think that CPU1 is in interrupt, even though it has already resumed after. This will block CPU1 from further interrupts, such as waiting for SGI1 to boot into CPU1_IDLE from BL33

os/arch/arm/src/armv7-a/arm_scu.c
eunwoo-nam: I got it, thank you for explain

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
eunwoo-nam: the reason why there is no clear register in hotplug is because all the registers are cleared when entering and exiting PG?

lhenry-realtek: Yes, so there are 2 scenario,

1) PG, which will hotplug CPU1, then prep CPU0 for power down
For this one, the GIC state should be reset when CPU0 first powers up, this is as you describe

2) runtime hotplug, which only hotplug CPU1, but CPU0 is still active
For this one, because CPU0 is still active the GIC registers are still retained, so when CPU1 is hotplugged it is not properly cleaned due to ISR flow not complete

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -108,6 +108,57 @@ bool vPortGateOtherCore(void)
...
+	/* fire SGI1 to kick the freshly booted CPU into TizenRT Idle Task */
+	os_smp_start();
+}
+
+void smp_full_powerdown_cpu(void)
+{
+	irqstate_t flags = enter_critical_section();
+#ifdef CONFIG_CPU_HOTPLUG
+	/* Notify secondary core to migrate task to primary core and enter wfi*/
+	up_set_cpu_state(SECONDARY_CORE_ID, CPU_HALTED);
```
anjana348: Is it needed to call up_set_cpu_state(cpu, CPU_HALTED) twice ? It is also called in below function up_cpu_haltcore(SECONDARY_CORE_ID);

lhenry-realtek: Probably not required, anyway this is demonstration code, the final outcome will take these into account

lhenry-realtek:
Hi @ewoodev 

The example is now able to power down and up CPU1 multiple times without deadlock
Previously, when 2nd loop of power down and up is run, CPU1 will get deadlocked. 

This is due to psci_cpu_off behavior being noreturn, trapped in WFI loop. As a result, EOIR is never written to, meaning the interrupt is never acknowledged.
This causes CPU1 to stop responding to interrupts as the GIC thinks CPU1 is still serving SGI1/4.

https://github.com/Samsung/TizenRT/blob/145c56a89bec45dbedc779ad918414e7d2497aa3/os/arch/arm/src/armv7-a/arm_cpuhotplug.c#L113-L118

The Idle task of CPU1 is marked inactive/running when CPU1 is hotplugged and set back to running
https://github.com/Samsung/TizenRT/blob/145c56a89bec45dbedc779ad918414e7d2497aa3/os/arch/arm/src/armv7-a/arm_cpuhotplug.c#L109-L110
https://github.com/Samsung/TizenRT/blob/145c56a89bec45dbedc779ad918414e7d2497aa3/os/arch/arm/src/amebasmart/amebasmart_smp.c#L147-L149


SMP example is able to run after CPU1 is powered off and on again.
It is expected behavior for SMP example to be stuck when CPU1 is OFF while it is run, the OS currently lacks detection of CPU1 being powered down

eunwoo-nam:
Hello @lhenry-realtek 
yes, right we plan to make changes sched for cpu1 off state.

Thank you for checking

we will make hal layer about this. 
Until then, we would like to discuss more about this in here.

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
eunwoo-nam: can we apply clear register in hotplug state during PG also?
because i think, 1. case also we should have clear register, but in the case, entering/exiting PG is covering that.

and, i think we have two option.

1) Clear register in any CPU1 off situation
2) move pcsi cpu off to cpu1 idle loop.

Could you please let me know your opinion?

lhenry-realtek: I think it should be fine since the register clearing is for resetting the state of CPU1 at the GIC, but need to test with PG

For your options, what do you mean by (2)? Are you going to make CPU1 hotplug in idle task, then startup only if CPU1 has work?

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -108,6 +108,66 @@ bool vPortGateOtherCore(void)
...
+	sched_lock();
+
+	/* send SGI4 to make core enter halt state (not complete powerdown yet) */
+	up_cpu_haltcore(SECONDARY_CORE_ID);
+
+	/* release the lock, cpu1 is no longer part of SMP and the idle task is marked inactive */
+	sched_unlock();
+	while (up_get_cpu_state(SECONDARY_CORE_ID) != CPU_HALTED);
+	lldbg("Secondary core shutdown?... %d\n", up_get_cpu_state(SECONDARY_CORE_ID));
+
```
anjana348: Hi henry, 
Shouldn't we add psci_affinity_info check here before powering off as done in vPortSecondaryOff() ? There is no delay as well after called psci_cpu_off here

lhenry-realtek: Hi, it may be added without issue. 

There were no effects observed when initially drafting the POC code for this feature

anjana348: We tried running it with flat build , it works fine but with loadable build, the code gets stuck on cpu on so we tried adding psci_affinity_info check here then it was working with loadable build as well

lhenry-realtek: Ok, I will modify the commit later to put in the check

amandeep-samsung: @lhenry-realtek : regarding cpu states, we are preparing to submit PR for them.
https://github.com/Samsung/TizenRT/pull/7007

we have added HALTED state for cpu as done in this PR.
psci_cpu_off()   --> hotplug
rtk power off   --> HALT  

is it ok ?

lhenry-realtek: > @lhenry-realtek : regarding cpu states, we are preparing to submit PR for them. #7007
> 
> we have added HALTED state for cpu as done in this PR. psci_cpu_off() --> hotplug rtk power off --> HALT
> 
> is it ok ?

Hi, technically following general CPU terminology of CPU C-States, it should be swapped around

image

Halt = CPU is still powered on, but no longer taking part in the OS, it is in low power state ready to turn off (psci_cpu_off)
Hotplug == Soft Power Off = CPU power domain is turned off (rtk power off)

amandeep-samsung: so, it means we need to reverse definition and meaning for HOTPLUG and HALT. We will change now accordingly

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -108,6 +108,78 @@ bool vPortGateOtherCore(void)
...
+	/* cold-boot the core */
+	lldbg("Booting secondary core... \n");
+	BaseType_t err = psci_cpu_on(SECONDARY_CORE_ID, (unsigned long)__cpu1_start);
+	DEBUGASSERT(err >= 0);
+
+	/* await for PSCI to report cpu state */
+	do {
+		state  = psci_affinity_info(1, 0);
+		if (state == 1) {
```
abhinav-s235: Shouldn't it be (state == 0)?
I think it should wait until CPU is ON (state == 0)

os/arch/arm/src/amebasmart/amebasmart_smp.c
```diff
@@ -108,6 +108,78 @@ bool vPortGateOtherCore(void)
...
+
+	/* take the lock, as we are about to put cpu1 idle task back active */
+	sched_lock();
+	
+	/* turn on CPU1's power domain, require time to stabilize */
+	rtk_core1_power_on();
+	DelayMs(1);
+
+	/* renable scu and invalidate tags on all processors */
+	arm_enable_smp(0);
```
amandeep-samsung, line 131: @lhenry-realtek ; as it is for secondary CPU (cpu id 1) enable, so arm_enable_smp(1) should be in place of arm_enable_smp(0). right ?

lhenry-realtek: Hi, the reason why 0 is used is because we want to invalidate the SCU state and reinitialize it, ensuring all data from the past CPU caches are cleared to a clean state

The SCU enters an unknown state if CPU1 is powered back on, since the caches on CPU1 are now invalid / stale data

If you check the API, only when cpu=0 is used, extra steps are performed

https://github.com/Samsung/TizenRT/blob/35c8c8b443a76ec8d9e327e737f2de40d9120d00/os/arch/arm/src/armv7-a/arm_scu.c#L163

https://github.com/Samsung/TizenRT/blob/35c8c8b443a76ec8d9e327e737f2de40d9120d00/os/arch/arm/src/armv7-a/arm_scu.c#L174-L179

PR #6709, opened 2025-03-10
driver/input: Add ist415 tmode command

IMJuBBang:
- Add tmode command.
- tmode is a command that disables Alive and Interrupt to connect the Tool for CS noise testing.


ewoodev:
The touch disable and enable code is already in  https://github.com/Samsung/TizenRT/blob/f59895f481cde509edc10d91d9d0333363141e59/apps/examples/touchscreen/touchscreen_main.c#L205-L239

but it's disabling is touch ic power off.

is thie command for testing case which touch ic is still running but touch irq only disabling?

ewoodev (approved):

sunghan-chang (approved):

PR #6676, opened 2025-02-18
os/arch/arm/src/amebasmart: remove sleep callback for mipi

zhongnuo-tang:
1. MIPI and LCDC resume flow is now handled in lcd power on and backlight on, and this will be controlled by the application layer, so remove duplicate resume flow in amebasmart_mipi_dsi_host_initialize.

PR #6710, opened 2025-03-10
os/board/rtl8730e/src: modify LCD power on/off flow

zhongnuo-tang:
1. There will be cases when LCD power off is called, but module do not enter PG. In this case, the lcdc and mipi drivers need not to re-initialize when LCD power on is called.
2. When LCD power on, we need to switch MIPI to cmd mode to send the lcd init cmd.
3. Fix UI coordinate offset shift gradually in lcd power on/off long run test

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -140,7 +140,6 @@ static void rtl8730e_lcd_power_on(void)
 {
 	GPIO_WriteBit(MIPI_GPIO_RESET_PIN, PIN_HIGH);
 	DelayMs(120);
-	amebasmart_mipi_dsi_host_reinitialize();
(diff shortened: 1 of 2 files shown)


zhongnuo-tang:
Hi @ewoodev ,
Following the production steps in email, after commenting out pm_start and related code, i can reproduce the shift in coordinate within 5mins. With this PR, it is not reproduced in 1hr+ test.
I will run the long run test with pm_start included, could you also help to run this in the product code side to see if this fixes the issue. 
If this PR works,  #6676 is no longer needed.

sunghan-chang:
@neel-samsung Could you verify this?

neel-samsung:
> @neel-samsung Could you verify this?

I performed long run test with PM enabled. But, I was not able to reproduce the LCD screen shift issue with this PR.

sunghan-chang (approved):

PR #6711, opened 2025-03-11
os/arch/arm/src/amebasmart: switch to SPI non-dma when called in idle…

zhongnuo-tang:
… thread

1. When CONFIG_AMEBASMART_SPI_DMA enabled, assert happens if amebasmart_spi_exchange is called in idle thread
2. add condition use spi polling mode if called in idle thread


zhongnuo-tang:
Verified with emulator using produce code. Without this PR, module will hang in entering PG.
sleep -> ls -> hi-bixby -> wakeup -> hi-bixby -> sleep -> ...

sunghan-chang:
@zhongnuo-tang This is pending due to LCD and Hang issues.

PR #6713, opened 2025-03-12
os/drivers/lcd: Use command mode to turn screen to black after power on

anjana348:
Command mode keeps the LCD in black frame which is needed just after power on.
 APP will send its buffer after power on and then mipi mode will be changed to video

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -41,6 +41,13 @@
 #define CONFIG_LCD_MAXPOWER 100
 #endif
 
+enum lcd_mode_e {
+        LCD_ON = 0,
+        LCD_OFF = 1
+};
+
+typedef enum lcd_mode_e lcd_mode_t;
+
 extern const uint8_t lcd_logo_raw_data[]; // Buffer containing only logo
 static uint8_t *lcd_init_fullscreen_image = NULL; // Buffer containing full screen data with logo on specific position
 
@@ -111,6 +118,7 @@ struct mipi_lcd_dev_s {
 	//u8 *BackupLcdImgBuffer;
 	int fb_alloc_count;
 	uint8_t power;
+	lcd_mode_t lcdonoff;
 	sem_t sem;
 	struct mipi_lcd_config_s *config;
 };
@@ -135,9 +143,9 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 	msg.tx_buf = cmd_addr;
 	msg.tx_len = payload_len;
 	msg.flags = 0;
-	priv->config->lcd_mode_switch(false);
+	priv->config->mipi_mode_switch(CMD_MODE);
 	transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
-	priv->config->lcd_mode_switch(true);
+	priv->config->mipi_mode_switch(VIDEO_MODE);
 	if (transfer_status != OK) {
 		lcddbg("Command %x not sent \n", cmd);
 	}
@@ -255,6 +263,10 @@ static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord
 	}
 	priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
 #endif
+	if (priv->lcdonoff == LCD_OFF) {
+                priv->config->mipi_mode_switch(VIDEO_MODE);
+                priv->lcdonoff = LCD_ON;
+        }
 	sem_post(&priv->sem);
 	return OK;
 }
@@ -360,10 +372,10 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 		/* The power on must operate only when LCD is OFF */
 		if (priv->power == 0) {
 			priv->config->power_on();
+			priv->config->mipi_mode_switch(CMD_MODE);
+			priv->lcdonoff = LCD_OFF;
 			/* We need to send init cmd after LCD IC power on */
-			priv->config->lcd_mode_switch(false);
 			send_init_cmd(priv, lcd_init_cmd_g);
-			priv->config->lcd_mode_switch(true);
 		}
 		priv->config->backlight(power);
 	}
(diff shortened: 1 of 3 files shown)


os/drivers/lcd/mipi_lcd.c
```diff
@@ -498,6 +508,7 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	}
 	priv->config->backlight(CONFIG_LCD_MAXPOWER);
 	priv->power = CONFIG_LCD_MAXPOWER;
+	priv->mipimode = VIDEO;
```
zhongnuo-tang, line 511: Hi, i think it should be cmd mode here, lcd mode is not change to video mode yet at this step. initialization will only change to video mode at rtl8730e_enable_lcdc. 
how about other places that call priv->config->lcd_mode_switch(true/false), should we also change priv->mipimode accordingly?

anjana348: @zhongnuo-tang  - yes, the video mode is being set in rtl8730e_enable_lcdc which is called after mipi_lcdinitialize, but they all are part of initialization sequence only so setting it just before calling the function doesn't make any difference 

And setting priv->mipimode to other places was not needed as we are setting it true and false in same function and it is synchronized with putarea call using semaphore.

Still if you think of any scenario which can cause issue, please let me know

ewoodev: how about changing to the same name as lcdoff?

anjana348: @ewoodev - which name priv->mipimode ?  It is for letting us know if command mode is set or video mode. It should be lcdoff ?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -111,6 +116,7 @@ struct mipi_lcd_dev_s {
 	//u8 *BackupLcdImgBuffer;
 	int fb_alloc_count;
 	uint8_t power;
+	uint8_t mipimode;
```
Taejun-Kwon, line 119:
```
typedef enum mipi_mode_e mipi_mode_t;
...
uint8_t power;
mipi_mode_t mode;
...
```

os/drivers/lcd/mipi_lcd.c
```diff
@@ -44,6 +44,11 @@
 extern const uint8_t lcd_logo_raw_data[]; // Buffer containing only logo
 static uint8_t *lcd_init_fullscreen_image = NULL; // Buffer containing full screen data with logo on specific position
 
+enum {
+	CMD = 0,
+	VIDEO = 1
```
sunghan-chang, line 49: `CMD` and `VIDEO` are too common names.
Let me suggest 
```
enum mipi_mode_e {
    MIPI_CMD_MODE = 0,
    MIPI_VIDEO_MODE =1
};
```

And please find Taejun's comment below.

os/drivers/lcd/mipi_lcd.c
ewoodev: If the name is mipimode, we need to apply this to below also.
```
static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
...
	priv->config->lcd_mode_switch(false);
	transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
	priv->config->lcd_mode_switch(true);
```

But This is needed in only lcd on.
So, To avoid confusion, Please add it in send_cmd or change name.

anjana348: priv->config->lcd_mode_switch(false);  ---> this is setting mipi cmd mode
transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
priv->config->lcd_mode_switch(true);     ----> this is setting mipi video mode
So there was no need as both are done in same function and this function call is synchronized in power off
send_cmd is to send any command to LCD
send_init_cmd is for sending intialization command sequence.

ewoodev: According to your response, we should not name the variable as mipimode.

send_cmd also changes the mipi mode, but we do not add the priv->mipimode variable here. In this case, the variable priv->mipimode is only used to delay changing the cmd mode to video mode until the next frame arrives in the lcd on situation, not for mipimode.

Therefore, the variable name should be changed to lcd on/off or lcd_init/lcd_running instead of mipimode.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -244,6 +252,10 @@ static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord
 	row_end += 1;
 	col_start += 1;
 	col_end += 1;
+	if (priv->lcdonoff == LCD_OFF) {
```
ewoodev, line 255: I think, the code should be after lcd_put_area.

1. lcd on & cmd mode & init cmd
2. new frame flusing
3. lcd video mode

anjana348: @ewoodev - This sequence was given by realtek on mail (dane.kang).
 Are you saying it to modify like
if (priv->lcdonoff == LCD_OFF) {
                priv->config->lcd_mode_switch(true);
        }
priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
if (priv->lcdonoff == LCD_OFF) {
                priv->lcdonoff = LCD_ON;
        }

Or 
move complete code like
priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
if (priv->lcdonoff == LCD_OFF) {
                priv->config->lcd_mode_switch(true);
                priv->lcdonoff = LCD_ON;
        }

Better to confirm with realtek regarding side effect

sunghan-chang: @zhongnuo-tang Could you confirm @anjana348 's comment?

ewoodev: This can output the previous frame in a short time.
it may not be visible, but the moment the video mode makes that tte previous frame is allowed to be output.

Could you please let me know your opinion.

sunghan-chang: @zhongnuo-tang This is urgent. Could you check @ewoodev 's reply?

Dane-Kang: Hi @sunghan-chang , @ewoodev ,
Yes, it can be run with below flow 
Set cmd mode -> flush the data -> switch mode to video.

sunghan-chang: @anjana348 Realtek confirmed. Is it applied in the new update?

anjana348: @sunghan-chang Yes 
@Dane-Kang Please confirm latest commit again

Dane-Kang: LGTM

sunghan-chang (approved):

PR #6714, opened 2025-03-13
Silent reboot check

anjana348:
During initialization if silent reboot is detected,
LCD backlight is off
Touch is disabled

By default LCD backlight is set to 0,

It has been verified for LCD and touch both

### DIFF
--- a/os/drivers/input/touchscreen.c
+++ b/os/drivers/input/touchscreen.c
@@ -52,6 +52,7 @@
 #include <tinyara/fs/fs.h>
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/input/touchscreen.h>
+#include <tinyara/silent_reboot.h>
 
 /****************************************************************************
  * Pre-Processor Definitions
@@ -144,13 +145,6 @@ static int touch_open(FAR struct file *filep)
 		}
 		priv->tp_buf.size = CONFIG_TOUCH_BUFSIZE;
 
-		if (priv->ops && priv->ops->touch_enable) {
-			priv->ops->touch_enable(priv);
-		} else {
-			kmm_free(priv->tp_buf.buffer);
-			touch_semgive(&priv->sem);
-			return -EINVAL;
-		}
 	}
 	priv->crefs++;
 	DEBUGASSERT(priv->crefs > 0);
@@ -174,7 +168,6 @@ static int touch_close(FAR struct file *filep)
 	DEBUGASSERT(priv->crefs > 0);
 	priv->crefs--;
 	if (priv->crefs == 0) {
-		priv->ops->touch_disable(priv);
 #if !defined(CONFIG_DISABLE_POLL)
 		/* Check if this file is registered in a list of waiters for polling.
 		* For example, when task A is blocked by calling poll and task B try to terminate task A,
@@ -500,6 +493,8 @@ void touch_report(struct touchscreen_s *dev, struct touch_sample_s *data)
 
 int touch_register(const char *path, struct touchscreen_s *dev)
 {
+	int ret;
+	bool is_silent_mode;
 	sem_init(&dev->sem, 0, 1);
 	sem_init(&dev->pollsem, 0, 1);
 	sem_init(&dev->waitsem, 0, 0);
@@ -512,7 +507,14 @@ int touch_register(const char *path, struct touchscreen_s *dev)
 	}
 	touch_semgive(&dev->pollsem);
 #endif
-	int ret = register_driver(path, &g_touchdev_fileops, 0666, dev);
+	is_silent_mode = silent_reboot_is_silent_mode();
+	if (!is_silent_mode) {
+		if (dev->ops && dev->ops->touch_enable) {
+                        dev->ops->touch_enable(dev);
+                }
+	}
+
+	ret = register_driver(path, &g_touchdev_fileops, 0666, dev);
 	if (ret < 0) {
 		kmm_free(dev);
 		sem_destroy(&dev->sem);
--- a/os/drivers/lcd/lcd_dev.c
+++ b/os/drivers/lcd/lcd_dev.c
@@ -69,6 +69,7 @@
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/lcd/lcd_dev.h>
 #include <tinyara/pm/pm.h>
+#include <tinyara/silent_reboot.h>
 
 #define MAX_NO_PLANES 3
 /****************************************************************************
@@ -291,6 +292,7 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		/* To ensure setpower operation from power saving,
 		 * pm_suspend is calling first. */
 		if (old_power == 0 && new_power > 0) {
+			silent_reboot_lock();
 			(void)pm_suspend(priv->pm_domain);
 		}
 
@@ -302,6 +304,8 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		/* To ensure setpower operation from power saving,
 		 * pm_resume is calling after setpower. */
 		if (old_power > 0 && new_power == 0) {
+			silent_reboot_unlock();
+			silent_reboot_delay(3600);
 			(void)pm_resume(priv->pm_domain);
 		}
 #else
@@ -461,16 +465,20 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_info->pm_domain = pm_domain_register("LCD");
 #endif
 
-	lcd_init_put_image(dev);
 	sem_init(&lcd_info->sem, 0, 1);
-	if (dev->setpower) {
+	bool is_silent_mode;
+	is_silent_mode = silent_reboot_is_silent_mode();
+	if (!is_silent_mode) {
+		ASSERT(dev->setpower);
 		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
 		if (ret != OK) {
 			goto cleanup;
 		}
 #ifdef CONFIG_PM
-		(void)pm_suspend(lcd_info->pm_domain);
+                (void)pm_suspend(lcd_info->pm_domain);
 #endif
+		silent_reboot_lock();
+		lcd_init_put_image(dev);
 	}
 	if (lcd_info->dev->getplaneinfo) {
 		lcd_info->dev->getplaneinfo(lcd_info->dev, 0, &lcd_info->planeinfo);	//plane no is taken 0 here
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -42,8 +42,8 @@
 #endif
 
 enum lcd_mode_e {
-        LCD_ON = 0,
-        LCD_OFF = 1
+	LCD_ON = 0,
+	LCD_OFF = 1
 };
 
 typedef enum lcd_mode_e lcd_mode_t;
@@ -391,7 +391,9 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 		if (priv->power == 0) {
 			priv->config->power_on();
 			/* We need to send init cmd after LCD IC power on */
-			send_init_cmd(priv, lcd_init_cmd_g);
+			if (send_init_cmd(priv, lcd_init_cmd_g) != OK) {
+				lcddbg("ERROR: LCD Init sequence failed\n");
+			}
 		}
 		priv->config->backlight(power);
 	}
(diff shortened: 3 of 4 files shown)


os/drivers/lcd/lcd_dev.c
```diff
@@ -464,7 +465,11 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_init_put_image(dev);
 	sem_init(&lcd_info->sem, 0, 1);
 	if (dev->setpower) {
-		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
+		if (silent_reboot_is_silent_mode()) {
```
jeongarmy, line 468: Please check usage of this function in touchscreen.c and change this

anjana348: done

os/drivers/lcd/lcd_dev.c
```diff
@@ -464,7 +465,13 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_init_put_image(dev);
 	sem_init(&lcd_info->sem, 0, 1);
 	if (dev->setpower) {
-		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
+		bool is_silent_mode = false;
```
jeongarmy, line 468: I think we can remove checking `ret`
```
		bool is_silent_mode = false;
		silent_reboot_is_silent_mode(&is_silent_mode);
		if (is_silent_mode) {
			ret = dev->setpower(dev, 0);
		} else {
			ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
		}
```

os/drivers/input/touchscreen.c
```diff
@@ -512,7 +515,14 @@ int touch_register(const char *path, struct touchscreen_s *dev)
 	}
 	touch_semgive(&dev->pollsem);
 #endif
-	int ret = register_driver(path, &g_touchdev_fileops, 0666, dev);
+	silent_reboot_is_silent_mode(&is_silent_mode);
```
jeongarmy, line 518: I changed type of this function. Please check below commit and update it.
https://github.com/Samsung/TizenRT/commit/3007c01245b3266dc92c3063faaaad2d51a3dec4
```
bool is_silent_mode;
is_silent_mode = silent_reboot_is_silent_mode();
```

os/drivers/lcd/lcd_dev.c
```diff
@@ -464,7 +467,13 @@ int lcddev_register(struct lcd_dev_s *dev)
 	lcd_init_put_image(dev);
 	sem_init(&lcd_info->sem, 0, 1);
 	if (dev->setpower) {
-		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
+		bool is_silent_mode;
+		is_silent_mode = silent_reboot_is_silent_mode();
+		if (is_silent_mode) {
+			ret = dev->setpower(dev, 0);
```
ewoodev, line 473: If this case, don't call pm_suspend()

The pm_susepnd must be called only lcd turning on state.

If you make lcd turn off state, don't call pm_suspend or, need to release pm_suspend using pm_resume

ewoodev: How about
```
sem_init(&lcd_info->sem, 0, 1);
is_silent_mode = silent_reboot_is_silent_mode();
if (!is_silent_mode) {
	lcd_init_put_image(dev);
	if (dev->setpower) {
		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
		if (ret != OK) {
			goto cleanup;
		}
#ifdef CONFIG_PM
		(void)pm_suspend(lcd_info->pm_domain);
#endif
	}
}
```

if you change default power 0 in mipi_lcd, we don't need to call setpower 0

os/drivers/input/touchscreen.c
```diff
@@ -512,7 +515,14 @@ int touch_register(const char *path, struct touchscreen_s *dev)
...
 	touch_semgive(&dev->pollsem);
 #endif
-	int ret = register_driver(path, &g_touchdev_fileops, 0666, dev);
+	is_silent_mode = silent_reboot_is_silent_mode();
+	if (is_silent_mode) {
+		if (dev->ops && dev->ops->touch_disable) {
+			dev->ops->touch_disable(dev);
+		}
+	}
+
```
ewoodev, line 524: If you're going to expect this, we need to remove touch_enable/touch_disalbe in open/close

anjana348: @ewoodev - you means if silent mode on, we need to add check in touch open/close as well for enabling /disabling

os/drivers/input/touchscreen.c
```diff
@@ -512,7 +515,14 @@ int touch_register(const char *path, struct touchscreen_s *dev)
 	}
 	touch_semgive(&dev->pollsem);
 #endif
-	int ret = register_driver(path, &g_touchdev_fileops, 0666, dev);
+	is_silent_mode = silent_reboot_is_silent_mode();
+	if (is_silent_mode) {
+		if (dev->ops && dev->ops->touch_disable) {
+			dev->ops->touch_disable(dev);
```
jeongarmy, line 521: When touch enable is called?
I think we should enable touch after lcd ON

os/drivers/input/touchscreen.c
ewoodev: I think, it's better that remove touch enable in open and disable in close.
and if not silent mode, call priv->ops->touch_enable(priv) in touch_register,

os/drivers/lcd/mipi_lcd.c
```diff
@@ -526,8 +526,8 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	} else {
 		lcddbg("ERROR: LCD Init sequence failed\n");
 	}
-	priv->config->backlight(CONFIG_LCD_MAXPOWER);
-	priv->power = CONFIG_LCD_MAXPOWER;
+	priv->config->backlight(0);
+	priv->power = -1;
```
ewoodev, line 530: Why we need to check -1 condition?

os/drivers/lcd/lcd_dev.c
ewoodev: And if not silent mode, we need to add silent_reboot_lock(); here when lcd on

os/drivers/lcd/mipi_lcd.c
anjana348: We need to set inital value for priv->power
as we check condition in lcd_setpower(){
if (power == priv->power) {
                return OK;
        }
}
To execute for first time, if we set it to 0, it will return without doing anything means no power off will happen.
You can suggest better way.

ewoodev: 1. mipi_lcd init -> lcd power 0
2. lcd_dev if (!is_silent_mode) call lcd power 100
    - if silent_mode, we don't call anyting.

How about above?

anjana348: Still, in mipi_lcd_init
if we set priv->power = 0, it will just set value to 0 but actual power off will not happen
which should be executed. This condition needs to have some intial value and some new value, if both are same it will not run even once
lcd_setpower(){
if (power == priv->power) {
return OK;
}
}

ewoodev: we can remove `mipi_mode_switch_to_video(true);` in `rtl8730e_lcdc_initialize()`
and we can remove `priv->config->backlight(CONFIG_LCD_MAXPOWER);` in mipi_lcdinitialize().

I mean, The all lcd code init state is turn off and lcd_dev silent reboot controls lcd power.

0. mipi ini mode is cmd mode in `rtl8730e_lcdc_initialize()`
1. send lcd init cmd in `mipi_lcdinitialize()`
2. set the `priv->power = 0` and `priv->lcdonoff = LCD_OFF` in `mipi_lcdinitialize()`
3. don't change mipi mode to video in `rtl8730e_lcdc_initialize()`
4. Check the silent mode first in `lcddev_register` 
- silent_mode == false :  
  ->  call `lcd_init_put_image(dev);` 
  ->  call `ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);`
  ->  call `(void)pm_suspend(lcd_info->pm_domain);`
  ->  call `silent_reboot_lock();`
- silent_mode == true:
  -> don't call anything.

ewoodev: silent boot should be more fast than normal boot.

anjana348: If we remove  mipi_mode_switch_to_video(true); in rtl8730e_lcdc_initialize(), we will not see logo ever since 
on power on we don't switch to video mode untill first putarea call comes.
There is no way currently to distinguish whether power on is called after power off or is it power on after intialization.
We need to add something to distinguish between these two ways of power on . Any suggestion ?

os/drivers/input/touchscreen.c
ewoodev: Application will turn on when silent mode

os/drivers/lcd/mipi_lcd.c
```diff
@@ -524,9 +532,8 @@ FAR struct lcd_dev_s *mipi_lcdinitialize(FAR struct mipi_dsi_device *dsi, struct
 	} else {
 		lcddbg("ERROR: LCD Init sequence failed\n");
 	}
-	priv->config->backlight(CONFIG_LCD_MAXPOWER);
-	priv->power = CONFIG_LCD_MAXPOWER;
-	priv->lcdonoff = LCD_ON;
+	priv->power = 0;
```
ewoodev, line 535: Could you please LCD IC reset GPIO? 
It should be also OFF

os/drivers/lcd/lcd_dev.c
```diff
@@ -461,16 +465,21 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+		bool is_silent_mode;
+		is_silent_mode = silent_reboot_is_silent_mode();
+		if (!is_silent_mode) {
+			ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
+			lcd_init_put_image(dev);
+#ifdef CONFIG_PM
+                (void)pm_suspend(lcd_info->pm_domain);
+#endif
+			silent_reboot_lock();
+		}
```
ewoodev, line 479: minor, I think, the lcd_init_put_image() is not releated to the power ops.
how about below?
```suggestion
	bool is_silent_mode;
	is_silent_mode = silent_reboot_is_silent_mode();
	if (!is_silent_mode) {
		if (dev->setpower) {
			ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
			(void)pm_suspend(lcd_info->pm_domain);
			silent_reboot_lock();
		}
		lcd_init_put_image(dev);
	}
```

anjana348: lcd_init_put_image() now changes mode to video mode so we should only call it when power is turned on in intialization
If power on didn't execute and we call it, then it can create issue
As app will call power on in that case, it will not be in command mode and sending init commands might encounter issue

ewoodev: dev->setpower looks mandatory. So, How about chagne to ASSERT(dev->setpower)?

os/drivers/lcd/lcd_dev.c
```diff
@@ -461,16 +465,20 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+	if (!is_silent_mode) {
+		ASSERT(dev->setpower);
 		ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
+#ifdef CONFIG_PM
+                (void)pm_suspend(lcd_info->pm_domain);
+#endif
+		silent_reboot_lock();
+		lcd_init_put_image(dev);
 		if (ret != OK) {
 			goto cleanup;
```
ewoodev, line 480: if it calls here, we need to add pm_reusm() and silent_reboot_unlock into cleanup:
Let's move it before pm_suspend calling

os/drivers/input/touchscreen.c
```diff
@@ -144,13 +145,6 @@ static int touch_open(FAR struct file *filep)
 		}
 		priv->tp_buf.size = CONFIG_TOUCH_BUFSIZE;
 
-		if (priv->ops && priv->ops->touch_enable) {
```
jeongarmy, line 147: When touch is enabled when silent mode booting?
As I know, when knockknock event happens, it turns on LCD and enabled touch.
@ewoodev Is it right?

os/drivers/lcd/lcd_dev.c
```diff
@@ -291,6 +292,7 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		/* To ensure setpower operation from power saving,
 		 * pm_suspend is calling first. */
 		if (old_power == 0 && new_power > 0) {
+			silent_reboot_lock();
```
jeongarmy, line 295: when lcd on, touch should be on again

os/drivers/lcd/lcd_dev.c
```diff
@@ -302,6 +304,8 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		/* To ensure setpower operation from power saving,
 		 * pm_resume is calling after setpower. */
 		if (old_power > 0 && new_power == 0) {
+			silent_reboot_unlock();
```
jeongarmy, line 307: when lcd off, touch should be off

os/drivers/lcd/lcd_dev.c
ewoodev: we don't control touch in lcd driver. it is turn on/of by application

jeongarmy: okay. thank you for correct

ewoodev (approved):
Could you please leave verify result?

jeongarmy (approved):
LGTM

PR #6715, opened 2025-03-14
os/board/rtl8730e/src/component/wifi: remove spin_lock and irqsave, a…

zhongnuo-tang:
…nd use enter_critical_section instead

1. irqsave and spin_lock was previously to reduce operations for optimization, but it will be safer to use enter critical section.
2. since enter critical section has macro to differentiate SMP and non-SMP case, we just call it for both cases.

### DIFF
--- a/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_msg_queue.c
+++ b/os/board/rtl8730e/src/component/wifi/inic/inic_ipc_msg_queue.c
@@ -56,10 +56,6 @@ inic_ipc_ex_msg_t g_inic_ipc_ex_msg __attribute__((aligned(64)));
 #else
 static inic_ipc_ex_msg_t g_inic_ipc_ex_msg = {0};
 #endif
-/* FreeRTOS: #if defined(configNUM_CORES) && (configNUM_CORES > 1)*/
-#if defined(CONFIG_SMP_NCPUS) && (CONFIG_SMP_NCPUS > 1)
-static spinlock_t ipc_msg_queue_lock = {0};
-#endif
 #ifdef CONFIG_PLATFORM_TIZENRT_OS
 struct task_struct ipc_msgQ_wlan_task;
 #endif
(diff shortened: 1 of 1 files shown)


sunghan-chang (approved):
@zhongnuo-tang Could you let us know how did you verify this?

zhongnuo-tang:
> @zhongnuo-tang Could you let us know how did you verify this?
This is mainly used by wifi trx between CA32 and wifi driver (km4), we verified by testing wifi throughput and did not notice any issues with the changes. Moreover, as discussed in the email with Ritesh, since enter_critical_section has taken into account various cases for SMP, we think it would be safer to stay consistent with TizenRT's implementation instead of calling directly into spinlock and irqsave.  
However, there isn't any issues related to this PR, it is part of tidying up our code.

sunghan-chang (approved):

sunghan-chang:
@kishore-sn @r-prabu Could you review this?

os/board/rtl8730e/src/component/wifi/inic/inic_ipc_msg_queue.c
```diff
@@ -73,27 +73,18 @@ struct task_struct ipc_msgQ_wlan_task;
  */
 static sint enqueue_ipc_msg_node(struct ipc_msg_node *p_node, _queue *p_queue)
 {
-#if defined(CONFIG_SMP_NCPUS) && (CONFIG_SMP_NCPUS > 1)
```
kishore-sn, line 76: Please delete ipc_msg_queue_lock variable at line 61 as it is no longer used.

zhongnuo-tang: Done.

sunghan-chang (approved):

sunghan-chang:
@kishore-sn Review is done?

Sambhavi-1 (approved):
using enter_critical_section instead of spin_lock, irqsave

kishore-sn (approved):

PR #6717, opened 2025-03-15
os/board/rtl8730e: Optimize FTL Write Protection Process

ziyik:
-Remove Status Bits read and backup
-Check if key is injected, else do not set or clear write protection


hs36-kim (approved):
LGTM

PR #6718, opened 2025-03-17
framework/ble_manager : Fix the issue of server config being passed as NULL value

giwon-nam:
- Resolved the issue where the config passed to the driver in the function that calls the newly added server config after ble_manager init is NULL.
 - Set the profile count value to 0 in g_server_null_config


hs36-kim (approved):
LGTM

PR #6719, opened 2025-03-17
os/board/rtl8730e: Fix deadlock issue when entering PG

jwei5:
- API _netlib_getipv4addr may call into sem_wait/sem_post due to socket(), causing context switch in pm_idle
- API will directly call into netdev_ifrioctl, remove call into socket as it is unnecessary


j11591 (approved):
LGTM

PR #6721, opened 2025-03-17
arch/arm/arm7-a: Add to check condition hotplug in cpu_pause

ewoodev:
The device hang occurs when cpu pause is attempted for a CPU in hotplug state.

Therefore, Check the status of the CPU that intends to pause and add an assertion if it is in a hotplug state. Additionally, since there is a CPU pause in up_assert, it is necessary to return in the abort_mode state to prevent the assertion from being called redundantly.


os/arch/arm/src/armv7-a/arm_cpupause.c
```diff
@@ -331,6 +332,16 @@ int up_cpu_pause(int cpu)
 		return OK;
 	}
 
+	/* If the pause target cpu is in CPU_HOTPLUG, g_cpu_paused spi_lock is
```
neel-samsung, line 335: could you please check the typo ```spi_lock``` ?

PR #6723, opened 2025-03-18
os/arch/arm/src/amebasmart: refactor peripheral suspend flow memory management

lhenry-realtek:
This commit refactors peripheral object to remove kmm_free in suspend flow

The memory region is instead reused by memset to 0 and then reinitialized

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2c.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2c.c
@@ -200,6 +200,8 @@ struct amebasmart_i2c_priv_s {
 	struct amebasmart_trace_s trace[CONFIG_I2C_NTRACE];
 #endif
 
+	uint8_t initialized;		/* Initialization status */
+
 	uint32_t status;			/* End of transfer SR2|SR1 status */
 };
 
@@ -923,7 +925,16 @@ static int amebasmart_i2c_init(FAR struct amebasmart_i2c_priv_s *priv)
 	/* Power-up and configure GPIOs */
 	DEBUGASSERT(priv);
 	DEBUGASSERT(!priv->i2c_object);
-	priv->i2c_object = (i2c_t *)kmm_malloc(sizeof(i2c_t));
+
+	if (!priv->initialized) {
+		priv->i2c_object = (i2c_t *)kmm_malloc(sizeof(i2c_t));
+		DEBUGASSERT(priv->i2c_object);
+		priv->initialized = true;
+	} else {
+		/* memset 0 if not required to allocate */
+		memset(priv->i2c_object, 0, sizeof(i2c_t));
+	}
+
 	DEBUGASSERT(priv->i2c_object);
 	i2c_init(priv->i2c_object, priv->config->sda_pin, priv->config->scl_pin);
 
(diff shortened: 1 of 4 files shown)


ewoodev:
Hello @lhenry-realtek,
We only need to avoid using kmm_Xalloc and kmm_free calls with sem in PG entry situations.
So, bsp peri code using static or allock doesn't matter. 

How about this seperate init/deinit and suspend/resume?
I think, we don't need option using configuration.

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -203,6 +203,9 @@ struct amebasmart_i2c_priv_s {
 	uint32_t status;			/* End of transfer SR2|SR1 status */
 };
 
+#define AMEBASMART_I2C_MAX 3
+static u8 g_i2c_inited[AMEBASMART_I2C_MAX];
```
edwakuwaku, line 207: How about add one more var in amebasmart_i2c_priv_s to indicate the initialised status (eg. bool is_initialized), then just use priv->is_initialized as reference.
In that case, you don't need to depend on i2c_index_get() to locate the port number as well.

ewoodev (commented):
Could you please solve confict? and check edwakuwaku's comments?

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
```diff
@@ -60,7 +60,7 @@ void i2c_send_restart(I2C_TypeDef *I2Cx, u8 *pBuf, u8 len, u8 restart);
   * @retval 1: I2C1 Device.
   * @retval 2: I2C2 Device.
   */
-static uint32_t i2c_index_get(PinName sda)
+uint32_t i2c_index_get(PinName sda)
```
ewoodev, line 63: is this changes needed?

os/arch/arm/src/amebasmart/amebasmart_i2c.c
```diff
@@ -923,7 +925,16 @@ static int amebasmart_i2c_init(FAR struct amebasmart_i2c_priv_s *priv)
...
 	DEBUGASSERT(!priv->i2c_object);
-	priv->i2c_object = (i2c_t *)kmm_malloc(sizeof(i2c_t));
+
+	if (!priv->initialized) {
+		priv->i2c_object = (i2c_t *)kmm_malloc(sizeof(i2c_t));
+		DEBUGASSERT(priv->i2c_object);
+		priv->initialized = true;
+	} else {
+		/* memset 0 if not required to allocate */
+		memset(priv->i2c_object, 0, sizeof(i2c_t));
```
sunghan-chang, line 935: The `i2c_t` has just two members with 32 bit size of each. Then, just allocating 0 or null will be much faster than memset which is working with 8 bit. How do you think?

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -242,6 +242,9 @@ static const struct amebasmart_i2s_config_s amebasmart_i2s3_config = {
 	.txenab = 0,
 };
 #endif
+
+static u8 g_i2s_inited[I2S_NUM_MAX];
```
sunghan-chang, line 246: Is it impossible to follow the same way with i2c?

sunghan-chang (approved):

sunghan-chang:
@lhenry-realtek Because I are much much urgent, let me merge it now. Please check my comments at next Monday.

sunghan-chang:
@lhenry-realtek This PR is not working.. When i2cuninitialize is called, then even we call i2cinitialize, i2c is not working. Because of your new variable, other necessary operation in i2cinitialize is not called. 
Did you really verify this??

PR #6724, opened 2025-03-18
os/arch/armv7-a: Modify cpu pause-resume during signal handling

neel-samsung:
When we perform signaling a task and the task is running on some other CPU, we need to pause the other CPU only. We don't need to pause all other CPUs. This patch fixes it.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_schedulesigaction.c
+++ b/os/arch/arm/src/armv7-a/arm_schedulesigaction.c
@@ -266,13 +266,13 @@ void up_schedule_sigaction(struct tcb_s *tcb, sig_deliver_t sigdeliver)
 
 			else {
 				/* If we signaling a task running on the other CPU, we have
-				 * to PAUSE the other CPUs.
+				 * to PAUSE the other CPU.
 				 */
 
 				if (cpu != me) {
-					/* Pause the other CPUs */
+					/* Pause the CPU */
 
-					up_cpu_pause_all();
+					up_cpu_pause(cpu);
 
 					/* Now tcb on the other CPU can be accessed safely */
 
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@neel-samsung Could you add what problem is with current code? If we pause all cpus, what happen?

neel-samsung:
> @neel-samsung Could you add what problem is with current code? If we pause all cpus, what happen?

No, there would not be any problem but, It's just why to pause all other CPUs if we can make it work by pausing the required one.

neel-samsung:
@sunghan-chang 

Latest Code: https://github.com/Samsung/TizenRT/blob/3007c01245b3266dc92c3063faaaad2d51a3dec4/os/arch/arm/src/armv7-a/arm_schedulesigaction.c#L271-L275

commit: https://github.com/Samsung/TizenRT/commit/a738838503917f46c0272bc44019d67c34b3b938
The above case was improperly handle in the above commit.

Code before commit:
https://github.com/Samsung/TizenRT/blob/54212f05b177be12082953aa443e9fe4085c7791/os/arch/arm/src/armv7-a/arm_schedulesigaction.c#L262-L265

sunghan-chang (approved):
Could you add how to verify this change?

neel-samsung:
> Could you add how to verify this change?

Let us verify it on LCD.

kripal-samsung (approved):

edwakuwaku:
> > Could you add how to verify this change?
> 
> Let us verify it on LCD.

There are only two cores in RTL8730E, I don't think it is possible to see any difference reflected from this change.
Maybe you can try it on imx6 qemu, it has 4 cores.

neel-samsung:
@sunghan-chang Onboarding and LCD/Touch test is working fine with the latest washer release.

kishore-sn (approved):

PR #6725, opened 2025-03-18
os/: Resolve lcd init cmd send stuck

namanjain7:
During LCD initialization, if a command send gets fail, then it gets stuck. This commit resolves this issue by adding changes:
1) Change sequence of setpower. lcd power on after mode switch will add additional delay of 120ms which can be utilized for send_init_cmd.
2) if a command fails, then add a retry mechanism. The count of retry can be changed from config.
3) add 5s timeout for the send_cmd_done to prevent stuck in infinite loop.


build/configs/rtl8730e/loadable_ext_ddr/defconfig
```diff
@@ -614,6 +614,7 @@ CONFIG_LCD_XRES=480
 CONFIG_LCD_YRES=800
 CONFIG_LCD_LOGO=y
 CONFIG_LCD_LOGO_52_340=y
+CONFIG_LCD_SEND_CMD_RETRY_COUNT=3
```
sunghan-chang, line 617: Let's split a commit for build configuration change.

os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -303,6 +303,7 @@ static int amebasmart_mipi_transfer(FAR struct mipi_dsi_host *dsi_host, FAR cons
 #endif
 	FAR struct amebasmart_mipi_dsi_host_s *priv = (FAR struct amebasmart_mipi_dsi_host_s *)dsi_host;
 	struct mipi_dsi_packet packet;
+	int cnt = 5000;
```
sunghan-chang, line 306: Let's exclude this change. I will merge this with PR #6722 .

os/drivers/lcd/Kconfig
```diff
@@ -177,6 +177,15 @@ config LCD_LOGO_52_340
 endchoice
 endif
 
+if LCD_ST7701 || LCD_ST7701SN || CONFIG_LCD_ST7785
```
sunghan-chang, line 180: Is this conditional required?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -136,6 +137,10 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 	msg.tx_len = payload_len;
 	msg.flags = 0;
 	priv->config->lcd_mode_switch(false);
+	DelayMs(100);
+	while (max_cmd_retry_count && mipi_dsi_transfer(priv->dsi_dev, &msg) != OK) {	/* Send command */
+		max_cmd_retry_count--;
+	}
 	transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
```
sunghan-chang, line 144: Maybe `mipi_dsi_transfer` is duplicated with line 141.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -178,11 +184,14 @@ static int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table
 			transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
 			if (transfer_status != OK) {
 				lcddbg("Command %x not sent \n", cmd);
-				return transfer_status;
+				max_cmd_retry_count--;
+				continue;
```
sunghan-chang, line 188: What about line 169? There is another `mipi_dsi_transfer`.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -359,9 +368,9 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 	} else {
 		/* The power on must operate only when LCD is OFF */
 		if (priv->power == 0) {
-			priv->config->power_on();
 			/* We need to send init cmd after LCD IC power on */
 			priv->config->lcd_mode_switch(false);
+			priv->config->power_on();
```
sunghan-chang, line 373: Before power_on, lcd_mode_switch is working?

sunghan-chang (commented):
You add retry logic at lcd driver, not in mipi transfer function. Is this better? 
If yes, how about making new function like lcd send cmd?
inside of the new function, we can add retry logic, then, the code will be simple.

os/drivers/lcd/Kconfig
namanjain7: yes, it's for mipi specific LCD only.

os/drivers/lcd/mipi_lcd.c
namanjain7: yes, it's working.

os/drivers/lcd/Kconfig
sunghan-chang: Then,
```
config LCD_INTERFACE_MIPI
	bool
	---help---
		LCD is connected with MIPI

config LCD_ST7701
	bool "ST7701 LCD Driver"
	select LCD_INTERFACE_MIPI
	---help---
		Enable LCD Driver, ST7701 with mipi.

if LCD_INTERFACE_MIPI
config LCD_SEND_CMD_RETRY_COUNT
     .....
     ---help---
          need to write why this is necessary in mipi only   <<<--- check this as well
```

os/drivers/lcd/mipi_lcd.c
```diff
@@ -117,6 +117,17 @@ struct mipi_lcd_dev_s {
 
 static struct mipi_lcd_dev_s g_lcdcdev;
 
+static int send_cmd_retry(struct mipi_lcd_dev_s *priv, struct mipi_dsi_msg* msg)
```
sunghan-chang, line 120: This function is not in the conditional like
```
#ifdef CONFIG_LCD_...RETRY
static int send_cmd_retry(....)
{
}
#endif
```
So, you should add some more code like
```
#if !defined(CONFIG_LCD_.._RETRY)
#define CONFIG_LCD....RETRY 1
#endif
```

sunghan-chang: BUT `CONFIG....RETRY 1 ` can be understandable that we will retry once.
Undefined the config means that we will not retry.
When you make modification, you should check this as well.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -177,8 +192,11 @@ static int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table
 			msg.flags = 0;
 			transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
 			if (transfer_status != OK) {
-				lcddbg("Command %x not sent \n", cmd);
-				return transfer_status;
+				transfer_status = send_cmd_retry(priv, &msg);
```
sunghan-chang, line 195: 1. What about line 178?
2. Can we change line 193? instead of adding 195 here?

os/drivers/lcd/mipi_lcd.c
sunghan-chang: hmmmm looks strange. Let's talk this later.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -48,6 +48,10 @@ enum lcd_mode_e {
 
 typedef enum lcd_mode_e lcd_mode_t;
 
+#if !defined(CONFIG_LCD_SEND_CMD_RETRY_COUNT)
+#define CONFIG_LCD_SEND_CMD_RETRY_COUNT 3
+#endif
```
sunghan-chang, line 53: Because the config is undefined, how about 1 (no retry) instead of 3?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -185,8 +209,15 @@ static int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table
 			msg.flags = 0;
 			transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
 			if (transfer_status != OK) {
-				lcddbg("Command %x not sent \n", cmd);
-				return transfer_status;
+				if (CONFIG_LCD_SEND_CMD_RETRY_COUNT > 0) {
```
sunghan-chang, line 212: same as previous

os/drivers/lcd/mipi_lcd.c
```diff
@@ -147,8 +162,17 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 	transfer_status = mipi_dsi_transfer(priv->dsi_dev, &msg);
 	priv->config->mipi_mode_switch(VIDEO_MODE);
 	if (transfer_status != OK) {
-		lcddbg("Command %x not sent \n", cmd);
+		/* Retry command send */
+		if (CONFIG_LCD_SEND_CMD_RETRY_COUNT > 0) {
+			transfer_status = send_cmd_retry(priv, &msg);
```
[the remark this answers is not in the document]
sunghan-chang: Because of my comment above, I think you made this.
```
sunghan-chang
BUT CONFIG....RETRY 1  can be understandable that we will retry once.
Undefined the config means that we will not retry.
When you make modification, you should check this as well.
```

How about below for this?
```
#if !defined(config)
#define config (0)
#endif

... send_cmd_retry(..)
{
   int max = config;
   int ret;
   ret = mipi transfer();
   while (ret != OK && max) {
      ret = mipi_transfer();
      max--;
   }
   return  ret; 
}
```

sunghan-chang (approved):

PR #6726, opened 2025-03-18
os/arch/arm/src/amebasmart: declare send_cmd_done flag as volatile

zhongnuo-tang:
1. send_cmd_done flag is updated in ISR, add volatile to prevent compiler from optimizing out memory reads, ensures the latest value is always read from memory.
2. since it is just a flag, use u8 instead of u32.

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_mipi.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_mipi.c
@@ -46,7 +46,7 @@
 #endif
 
 static u32 vo_freq;
-static u32 send_cmd_done = 0;
+static volatile u8 send_cmd_done = 0;
 
 struct amebasmart_mipi_dsi_host_s {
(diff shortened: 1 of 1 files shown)


sunghan-chang:
@zhongnuo-tang Is there any issue related this PR? or just fixing embedded hidden(?) issue before coming out?

zhongnuo-tang:
> @zhongnuo-tang Is there any issue related this PR? or just fixing embedded hidden(?) issue before coming out?

Yes, without volatile, the while(send_cmd_done != 1) loop in amebasmart_mipi_transfer might not exit because the compiler could optimize it by using a cached value instead of the latest update from the ISR. 
So, I'm afraid this might be one possible reason that cause mipi stuck issue.

sunghan-chang (approved):

sunghan-chang:
@anjana348 Could you review this?

PR #6727, opened 2025-03-18
os/arch/arm/src/armv7-a: Fix assert logs for irq crash

kishore-sn:
Fix wrong logs when a crash occurs in irq context:
1. The intial logs print the task name and pid instead of irq num
2. The irq stack caclulation is wrong and hence the logs indicate that stack pointer if out of allocated stack
3. Even though the SP is actually valid, the proper stack data is not printed. So, it is not possible to get the correct stack trace.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -392,7 +392,8 @@ static void up_dumpstate(struct tcb_s *fault_tcb, uint32_t asserted_location)
 	stacksize = (uint32_t)fault_tcb->adj_stack_size;
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
 #ifdef CONFIG_SMP
-	istackbase = (uint32_t)arm_intstack_alloc(),
+	/* Initialize istackbase based on the interrupt stack size and proper alignment value (~7) */
+	istackbase = ((uint32_t)arm_intstack_alloc() + (CONFIG_ARCH_INTERRUPTSTACK & ~7));
 #else
 	istackbase = (uint32_t)&g_intstackbase,
 #endif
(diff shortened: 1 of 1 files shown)


os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -520,10 +520,17 @@ void check_heap_corrupt(struct tcb_s *fault_tcb)
 
 static inline void print_assert_detail(const uint8_t *filename, int lineno, struct tcb_s *fault_tcb, uint32_t asserted_location)
 {
+	int cpu = up_cpu_index();
 	lldbg_noarg("===========================================================\n");
 	lldbg_noarg("Assertion details\n");
 	lldbg_noarg("===========================================================\n");
-	lldbg("Assertion failed CPU%d at file: %s line %d task: %s pid: %d\n", up_cpu_index(), filename, lineno, LOG_TASK_NAME, fault_tcb->pid);
+	lldbg_noarg("Assertion failed CPU%d at file: %s line %d", cpu, filename, lineno);
+	if (g_irq_num[cpu] >= 0) {
```
sunghan-chang, line 528: What does it mean `0`?

sunghan-chang (commented):
Is this verified?

kishore-sn:
> Is this verified?

Yes. I have verified it for crash in normal task and crash in irq.

os/arch/arm/src/armv7-a/arm_assert.c
edwakuwaku: Why not use up_interrupt_context()?

kishore-sn: We can use up_interrupt_context(). But i also want to print the irq number in the next line. The irq number is stored in g_irq_num[cpu]. So, I used the same to check the irq context.

edwakuwaku: It seems not very straightforward to realise this for whoever view the code. If we were to keep the same, perhaps adding some comment would be better.

kishore-sn: I added comment.

sunghan-chang (dismissed):

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -392,7 +392,7 @@ static void up_dumpstate(struct tcb_s *fault_tcb, uint32_t asserted_location)
 	stacksize = (uint32_t)fault_tcb->adj_stack_size;
 #if CONFIG_ARCH_INTERRUPTSTACK > 7
 #ifdef CONFIG_SMP
-	istackbase = (uint32_t)arm_intstack_alloc(),
+	istackbase = ((uint32_t)arm_intstack_alloc() + (CONFIG_ARCH_INTERRUPTSTACK & ~7));
```
r-prabu, line 395: can we have some comment or meaningful macro for it?

r-prabu (approved):

Sambhavi-1 (approved):
irq crash fixed looks good

r-prabu (approved):

PR #6729, opened 2025-03-18
os/board/rtl8730e: Rename BLE Set Server Config API

ziyik:
-Rename add_profile API to set_server_config API
-Remove not used variable


hs36-kim (approved):
LGTM

PR #6730, opened 2025-03-18
os/board/rtl8730e: fix condition check when setting status bits for ftl

yeetee179:
- added bitwise AND to status bits, so that it only checks the necessary bits for flash protection


sunghan-chang:
@yeetee179 Could you write more details at commit description? What was the issue? Why it happens? why modify like this?

sunghan-chang:
@yeetee179 I think you modified just two lines.

```
from
if (FLASH_STATUS_BITS == backup_state) { /* State is enable */

to
if (FLASH_STATUS_BITS == (backup_state & FLASH_STATUS_BITS)) { /* State is enable */
```
Right?

yeetee179:
> @yeetee179 Could you write more details at commit description? What was the issue? Why it happens? why modify like this?

More information is added to the commit

yeetee179:
> @yeetee179 I think you modified just two lines.
> 
> ```
> from
> if (FLASH_STATUS_BITS == backup_state) { /* State is enable */
> 
> to
> if (FLASH_STATUS_BITS == (backup_state & FLASH_STATUS_BITS)) { /* State is enable */
> ```
> 
> Right?

yea just 2 lines to for the conditional check

hs36-kim (approved):
LGTM

PR #6733, opened 2025-03-19
os/drivers/lcd/: Refactor lcd send command

namanjain7:
1) If config CONFIG_LCD_SEND_CMD_RETRY_COUNT is undefined, then no retry if command send gets failed.
2) Reduce code repetition of mipi_dsi_transfer function call. Shifted this call to send_cmd_transfer function.
3) Add new config CONFIG_LCD_INTERFACE_MIPI which speficies if LCD is connected to MIPI.


os/drivers/lcd/Kconfig
```diff
@@ -104,6 +104,12 @@ config LCD_MAXPOWER
 		backlight setting.  Board-specific logic may place restrictions on this
 		value.
 
+config LCD_INTERFACE_MIPI
+	bool "MIPI support for LCD"
+	default y
+	---help---
+		LCD is connected with MIPI
```
sunghan-chang, line 111: Let's remove `"MIPI support for LCD"` and set it as default n as below
```
config LCD_INTERFACE_MIPI
	bool
	default n
	---help---
		LCD is connected with MIPI
```

If there is no `" "` after type, user can't control it.
I think that new config is H/W specific value so that no need to set by user.
Just `select LCD_I....._MIPI` at each LCD config is enough.

os/drivers/lcd/Kconfig
```diff
@@ -128,6 +134,7 @@ config LCD_ST7701
 
 config LCD_ST7701SN
 	bool "ST7701SN LCD Driver"
+	select LCD_INTERFACE_MIPI
```
sunghan-chang, line 137:
```
choice
	prompt "Choose LCD"
	default LCD_ST7789
	---help---
		LCD ST7789 is 2.8 inch with SPI
		LCD ST7785 is 2.8 inch with MIPI
		LCD ST7701 is 3.9 inch with MIPI
		LCD ST7701SN is 4.3 inch with MIPI
```
You should add this line at each LCD which is connected with MIPI.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -129,13 +129,25 @@ struct mipi_lcd_dev_s {
...
+			while (max_cmd_retry_count && transfer_status != OK) {
+				transfer_status = mipi_dsi_transfer(priv->dsi_dev, msg);
+				max_cmd_retry_count--;
+			}
+			if (transfer_status != OK) {
+				lcddbg("Command %x not sent after retry\n", cmd);
+			}
+		} else {
+			lcddbg("Command %x not sent\n", cmd);
+		}
```
sunghan-chang, line 150: I think `if (CONFIG_...._COUNT > 0)` is unnecessary. The while conditional covers it.
Could you check it?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -129,13 +129,25 @@ struct mipi_lcd_dev_s {
 
 static struct mipi_lcd_dev_s g_lcdcdev;
 
-static int send_cmd_retry(struct mipi_lcd_dev_s *priv, struct mipi_dsi_msg* msg)
+static int send_cmd_transfer(struct mipi_lcd_dev_s *priv, struct mipi_dsi_msg* msg)
```
sunghan-chang, line 132: `send` and `transfer` looks the same.

os/drivers/lcd/Kconfig
namanjain7: ok. I have changed the Kconfig.

os/drivers/lcd/Kconfig
namanjain7: done.

os/drivers/lcd/mipi_lcd.c
namanjain7: changed the name to: send_to_mipi_dsi

sunghan-chang (approved):
Could you leave how to verify?

namanjain7:
Kconfig changes: verified using menuconfig manually and checking .config if changes applied or not.
cmd retry changes: verified using example_lcd test app, so init function is working fine. Delay is increased to 140ms in mode switch in #6725 PR. So, lcd won't get stuck again like previously.

PR #6734, opened 2025-03-19
os/driver/lcd: modify lcd power on off flow

zhongnuo-tang:
1. when lcd power off, it switches to cmd mode, we should keep it in cmd mode, and set lcdonoff as LCD_OFF
2. So that when lcd power on, we do not need to switch to cmd again to send init cmd table
3. move video mode switch out from rtl8730e_enable_lcdc so that when wake up from PG, it stays in cmd mode
4. when lcd power off, we need to check if there is pending framedone interrupt in LCDC before disable irq, and if there is, sem_post, to prevent g_next_frame_block being hold during lcd power off, and when power on again, it could not send next frame. Eg: fps_test() is sending frames very fast, so it will hit the case where it sem_wait at rtl8730e_lcd_put_area and at the same time lcd power off.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -134,6 +134,10 @@ static void rtl8730e_lcd_init(void)
 static void rtl8730e_lcd_power_off(void)
 {
 	InterruptDis(lcdc_irq_info.num);
+	if (lcdc_nextframe == 1) {
+		lcdc_nextframe = 0;
+		sem_post(&g_next_frame_block);
+	}
 	GPIO_WriteBit(MIPI_GPIO_RESET_PIN, PIN_LOW);
 }
 static void rtl8730e_lcd_power_on(void)
@@ -236,7 +240,6 @@ static void rtl8730e_enable_lcdc(void)
 {
 	LCDC_Cmd(pLCDC, ENABLE);
 	while (!LCDC_CheckLCDCReady(pLCDC)) ;
-	mipidsi_mode_switch(true);
 }
 
 void rtl8730e_mipidsi_underflowreset(void)
@@ -318,6 +321,7 @@ void rtl8730e_lcdc_initialize(void)
 	LcdcInitValues(config);
 	rtl8730e_lcd_init();
 	rtl8730e_enable_lcdc();
+	mipidsi_mode_switch(true);
 
 	if (lcddev_register(dev) < 0) {
 		lcddbg("ERROR: LCD driver register fail\n");
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -62,7 +62,7 @@ static void rtl8730e_lcd_put_area(u8 *lcd_img_buffer, u32 x_start, u32 y_start, 
 static void rtl8730e_enable_lcdc(void);
 static void rtl8730e_register_lcdc_isr(void);
 static void rtl8730e_control_backlight(u8 level);
-FAR void mipidsi_mode_switch(bool do_enable);
+FAR void mipi_mode_switch_to_video(bool do_enable);
 FAR void mipidsi_acpu_reg_clear(void);
 FAR struct mipi_dsi_host *amebasmart_mipi_dsi_host_initialize(struct lcd_data *config);
 FAR struct mipi_dsi_device *mipi_dsi_device_register(FAR struct mipi_dsi_host *host, FAR const char *name, int channel);
@@ -159,12 +159,12 @@ static void rtl8730e_gpio_reset(void)
 static void rtl8730e_mipi_mode_switch(mipi_mode_t mode)
 {
 	if (mode == CMD_MODE) {
-		mipidsi_mode_switch(false);
+		mipi_mode_switch_to_video(false);
 		MIPI_DSI_INT_Config(MIPI, DISABLE, ENABLE, FALSE);
 		DelayMs(140);
 	} else {
 		MIPI_DSI_INT_Config(MIPI, DISABLE, DISABLE, FALSE);
-		mipidsi_mode_switch(true);
+		mipi_mode_switch_to_video(true);
 	}
 }
 
@@ -284,7 +284,7 @@ u32 rtl8730e_hv_isr(void *Data)
 			InterruptRegister((IRQ_FUN)rtl8730e_mipidsi_underflowreset, mipi_irq_info.num, (u32)mipi_irq_info.data, mipi_irq_info.priority);
 			InterruptEn(mipi_irq_info.num, mipi_irq_info.priority);
 			mipidsi_acpu_reg_clear();
-			mipidsi_mode_switch(false);
+			mipi_mode_switch_to_video(false);
 			MIPI_DSI_INT_Config(MIPI, ENABLE, ENABLE, ENABLE);
 		}
 	}
@@ -321,7 +321,7 @@ void rtl8730e_lcdc_initialize(void)
 	LcdcInitValues(config);
 	rtl8730e_lcd_init();
 	rtl8730e_enable_lcdc();
-	mipidsi_mode_switch(true);
+	mipi_mode_switch_to_video(true);
 
 	if (lcddev_register(dev) < 0) {
 		lcddbg("ERROR: LCD driver register fail\n");
--- a/os/arch/arm/src/amebasmart/amebasmart_mipi.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_mipi.c
@@ -270,7 +270,7 @@ static void amebasmart_register_interrupt(void)
 	InterruptEn(mipi_irq_info.num, mipi_irq_info.priority);
 }
 
-void mipidsi_mode_switch(bool do_enable){
+void mipi_mode_switch_to_video(bool do_enable){
 	if(do_enable){
 		MIPI_DSI_Mode_Switch(g_dsi_host.MIPIx, ENABLE);
 	}
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -165,9 +165,7 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 	msg.tx_buf = cmd_addr;
 	msg.tx_len = payload_len;
 	msg.flags = 0;
-	priv->config->mipi_mode_switch(CMD_MODE);
 	transfer_status = send_to_mipi_dsi(priv, &msg);
-	priv->config->mipi_mode_switch(VIDEO_MODE);
 	return transfer_status;
 }
 
@@ -382,6 +380,8 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 	}
 	if (power == 0) {
 		priv->config->backlight(power);
+		priv->config->mipi_mode_switch(CMD_MODE);
+		priv->lcdonoff = LCD_OFF;
 		lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
 		send_cmd(priv, display_off_cmd);
 		/* The power off must operate only when LCD is ON */
@@ -390,8 +390,6 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 		/* The power on must operate only when LCD is OFF */
 		if (priv->power == 0) {
 			priv->config->power_on();
-			priv->config->mipi_mode_switch(CMD_MODE);
-			priv->lcdonoff = LCD_OFF;
 			/* We need to send init cmd after LCD IC power on */
 			send_init_cmd(priv, lcd_init_cmd_g);
 		}


sunghan-chang:
@anjana348 Could you review this?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -253,6 +258,7 @@ void rtl8730e_mipidsi_underflowreset(void)
 		lcddbg("ERROR: LCDC_CTRL 0x%x\n", pLCDC->LCDC_CTRL);
 
 		rtl8730e_enable_lcdc();
+		mipidsi_mode_switch(true);
```
sunghan-chang, line 261: Regardless of this PR, `mipi mode switch true` is difficult to understand in code readability. Could you rename it with another commit?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -318,7 +324,7 @@ void rtl8730e_lcdc_initialize(void)
 	LcdcInitValues(config);
 	rtl8730e_lcd_init();
 	rtl8730e_enable_lcdc();
-
+	mipidsi_mode_switch(true);
```
anjana348, line 327: @zhongnuo-tang - why is this removed from rtl8730e_enable_lcdc() and called separately?

anjana348:
@zhongnuo-tang - Regarding setting black color frame after power on, we have observed that, when we call putarea and switch to video mode, it still first shows previous frame and then new frame. 
It seems when we enable video mode, it shows previous frame untill the LCDC interrupt comes and when interrupt comes, it shows next frame although we have set next frame data in DMA buffer before switching to video mode.
Please check this and let us know solution
In Public PR, we can reproduce it
Set logo in intialization (Change this to some visible frame for testing as it is black for now) 
Keep LCD power off after initialization
Power on 
Set new frame in putarea
---> Here we will see logo frame and then new frame when switched to video mode

zhongnuo-tang:
> @zhongnuo-tang - Regarding setting black color frame after power on, we have observed that, when we call putarea and switch to video mode, it still first shows previous frame and then new frame. It seems when we enable video mode, it shows previous frame untill the LCDC interrupt comes and when interrupt comes, it shows next frame although we have set next frame data in DMA buffer before switching to video mode. Please check this and let us know solution In Public PR, we can reproduce it Set logo in intialization (Change this to some visible frame for testing as it is black for now) Keep LCD power off after initialization Power on Set new frame in putarea ---> Here we will see logo frame and then new frame when switched to video mode

Hi in rtl8730e_lcd_put_area, the new configurable will only be applied when LCDC_TrigerSHWReload happens, before this 'old data' will be read, maybe the reason why it shows the previous frame because it is blocked by sem_wait and not able to reload the LCDC. I suggest you can test with this PR, it will take care of the sem_post when lcd power off. 
        sem_wait(&g_next_frame_block);
	lcdc_nextframe = 1;
	LCDC_TrigerSHWReload(pLCDC);

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
zhongnuo-tang: This is because of the following, before this PR
1. ,when lcd off, cmd mode is used to send display off, then it switch back to video mode --> this is not necessary
2. when lcd on, it switch to cmd mode again to send init cmd table because 1. set it to video mode

so this PR will reduce the unnecessary switch to video mode after lcd off. But since PG wakeup flow will call rtl8730_enable_lcdc, if we do not remove this and call separately, when wakeup from PG, it will be in video mode and this will not work because i have remove the switch to cmd mode before send init table cmd in lcd power on.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
zhongnuo-tang: mipidsi_mode_switch(true) --> video mode
mipidsi_mode_switch(false) --> cmd mode
this has been taken care below, maybe i will just change it to call rtl8730e_mipi_mode_switch instead of rename mipidsi_mode_switch? 
static void rtl8730e_mipi_mode_switch(mipi_mode_t mode)
{
	if (mode == CMD_MODE) {
		mipidsi_mode_switch(false);
		MIPI_DSI_INT_Config(MIPI, DISABLE, ENABLE, FALSE);
		DelayMs(140);
	} else {
		MIPI_DSI_INT_Config(MIPI, DISABLE, DISABLE, FALSE);
		mipidsi_mode_switch(true);
	}
}

sunghan-chang: mipi_mode_switch_to_video(true)
mipi_mode_switch_to_video(false)

or

mipi_mode_switch(VIDEO_MODE)
mipi_mode_switch(CMD_MODE)

I suggest renaming like above.
Current code is impossible to change what. I should go into the function and should check code again.

sunghan-chang:
@zhongnuo-tang Need to rebase the PR due to conflicts

ewoodev:
@anjana348, @zhongnuo-tang 
This PR looks to improve user responsiveness, we can reduce the time it takes to turn on the LCD by 140ms.
we are changing mipi mode to cmd during lcd turn on. so lcd turn on time is effected 120ms + 140ms = 260ms.
If we change mipi mode to cmd when turning off, we can reduce lcd turn on time.

sunghan-chang:
@zhongnuo-tang @anjana348 Let's modify and review this PR with top priority. The real product has a performance issue.

anjana348:
@zhongnuo-tang 
https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531

This needs to be changed with your PR
priv->lcdonoff = LCD_ON;
--->
priv->lcdonoff = LCD_OFF;

zhongnuo-tang:
> @zhongnuo-tang
> 
> https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531
> 
> This needs to be changed with your PR priv->lcdonoff = LCD_ON; ---> priv->lcdonoff = LCD_OFF;

any reason for this? even though video mode is remove from rtl8730_enable_lcdc, it is still being called during intialization right after it

anjana348:
> > @zhongnuo-tang
> > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531
> > 
> > This needs to be changed with your PR priv->lcdonoff = LCD_ON; ---> priv->lcdonoff = LCD_OFF;
> 
> any reason for this? even though video mode is remove from rtl8730_enable_lcdc, it is still being called during intialization right after it

Ohh yes, it will work that way.
And what about 
https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/lcd_dev.c#L467

lcddev_register - It will be called after switching to video mode but send_init_cmd  will be called in power on without being switched to command mode. It works but is it right?
If we don't call power off

zhongnuo-tang:
> > > @zhongnuo-tang
> > > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531
> > > 
> > > This needs to be changed with your PR priv->lcdonoff = LCD_ON; ---> priv->lcdonoff = LCD_OFF;
> > 
> > 
> > any reason for this? even though video mode is remove from rtl8730_enable_lcdc, it is still being called during intialization right after it
> 
> Ohh yes, it will work that way. And what about
> 
> https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/lcd_dev.c#L467
> 
> lcddev_register - It will be called after switching to video mode but send_init_cmd will be called in power on without being switched to command mode. It works but is it right? If we don't call power off

yes, it is alright as it is default in cmd mode.

anjana348:
> > > > @zhongnuo-tang
> > > > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531
> > > > 
> > > > This needs to be changed with your PR priv->lcdonoff = LCD_ON; ---> priv->lcdonoff = LCD_OFF;
> > > 
> > > 
> > > any reason for this? even though video mode is remove from rtl8730_enable_lcdc, it is still being called during intialization right after it
> > 
> > 
> > Ohh yes, it will work that way. And what about
> > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/lcd_dev.c#L467
> > 
> > lcddev_register - It will be called after switching to video mode but send_init_cmd will be called in power on without being switched to command mode. It works but is it right? If we don't call power off
> 
> yes, it is alright as it is default in cmd mode.

But we are changing it 
rtl8730e_lcd_init();
	rtl8730e_enable_lcdc();
mipidsi_mode_switch(true); ---------> here it is changed to video mode
	if (lcddev_register(dev) < 0) {
		lcddbg("ERROR: LCD driver register fail\n");
		return;
	}

zhongnuo-tang:
> > > > > @zhongnuo-tang
> > > > > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/mipi_lcd.c#L531
> > > > > 
> > > > > This needs to be changed with your PR priv->lcdonoff = LCD_ON; ---> priv->lcdonoff = LCD_OFF;
> > > > 
> > > > 
> > > > any reason for this? even though video mode is remove from rtl8730_enable_lcdc, it is still being called during intialization right after it
> > > 
> > > 
> > > Ohh yes, it will work that way. And what about
> > > https://github.com/Samsung/TizenRT/blob/eee6439811990f7b5c1177514b31377825daf638/os/drivers/lcd/lcd_dev.c#L467
> > > 
> > > lcddev_register - It will be called after switching to video mode but send_init_cmd will be called in power on without being switched to command mode. It works but is it right? If we don't call power off
> > 
> > 
> > yes, it is alright as it is default in cmd mode.
> 
> But we are changing it rtl8730e_lcd_init(); rtl8730e_enable_lcdc(); mipidsi_mode_switch(true); ---------> here it is changed to video mode if (lcddev_register(dev) < 0) { lcddbg("ERROR: LCD driver register fail\n"); return; }

send_init_cmd is called in mipi_lcdinitialize which is before we switch to video mode here, so it will still be cmd mode.
https://github.com/Samsung/TizenRT/blob/96708006dd47d1dfb16362fcfe0e05b0132e8a82/os/drivers/lcd/mipi_lcd.c#L534

sunghan-chang:
@zhongnuo-tang @anjana348 Hmm.. Is there any way to check current mode (command mode? video mode?)
Current code has too strong dependency with previous code operation. If we move or add some logic, then, all working mode could be corrupted.

sunghan-chang:
@zhongnuo-tang @anjana348 I don't think current code is perfectly ok.

zhongnuo-tang:
> @zhongnuo-tang @anjana348 Hmm.. Is there any way to check current mode (command mode? video mode?) Current code has too strong dependency with previous code operation. If we move or add some logic, then, all working mode could be corrupted.

There isn't way to check current mode, i think it is because only video mode is used in the current implementation for rendering of lcd., and cmd mode is only used in power on/off when sending either display off cmd or init cmd table, so it is not that of an issue now. Anyway i think if necessary, it can be implemented similar to LCD_ON/OFF.

zhongnuo-tang:
> @zhongnuo-tang @anjana348 I don't think current code is perfectly ok.

Are you referring to this PR or the general lcd driver code?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -133,7 +133,13 @@ static void rtl8730e_lcd_init(void)
 
 static void rtl8730e_lcd_power_off(void)
 {
+	volatile u32 IntId = LCDC_GetINTStatus(pLCDC);
```
edwakuwaku, line 136: where is IntId used?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
zhongnuo-tang: Hi @sunghan-chang, new commit to rename done.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
zhongnuo-tang: removed

sunghan-chang:
@anjana348 When you review done, could you leave approval?
@zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?

anjana348:
> @anjana348 When you review done, could you leave approval? @zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?

We can measure performance similar to fps_test
I tried locally using this PR
Power off + Power on sequence
500 iterations ~ 230seconds
which can be divided as
Delay after GPIO pin high in power on - 120ms * 500(iterations) ~ 60seconds
Delay after command mode switch in power off - 140ms * 500 ~ 70seconds
Time taken in sending init command sequence in power on over 500 iterations  ~ 100seconds

sunghan-chang:
> > @anjana348 When you review done, could you leave approval? @zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?
> 
> We can measure performance similar to fps_test I tried locally using this PR Power off + Power on sequence 500 iterations ~ 230seconds which can be divided as Delay after GPIO pin high in power on - 120ms * 500(iterations) ~ 60seconds Delay after command mode switch in power off - 140ms * 500 ~ 70seconds Time taken in sending init command sequence in power on over 500 iterations ~ 100seconds

@anjana348 Could you compare it with and without this PR?

anjana348:
> > > @anjana348 When you review done, could you leave approval? @zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?
> > 
> > 
> > We can measure performance similar to fps_test I tried locally using this PR Power off + Power on sequence 500 iterations ~ 230seconds which can be divided as Delay after GPIO pin high in power on - 120ms * 500(iterations) ~ 60seconds Delay after command mode switch in power off - 140ms * 500 ~ 70seconds Time taken in sending init command sequence in power on over 500 iterations ~ 100seconds
> 
> @anjana348 Could you compare it with and without this PR?

As we can also see theoretically , In this PR, command mode is not switched in power on so without this PR
500 iterations ~ 300second (230 + 70) 
70 seconds extra time to switch to command mode in power on which is removed here, rest performance is same

edwakuwaku:
> > > > @anjana348 When you review done, could you leave approval? @zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?
> 
> > > 
> 
> > > 
> 
> > > We can measure performance similar to fps_test I tried locally using this PR Power off + Power on sequence 500 iterations ~ 230seconds which can be divided as Delay after GPIO pin high in power on - 120ms * 500(iterations) ~ 60seconds Delay after command mode switch in power off - 140ms * 500 ~ 70seconds Time taken in sending init command sequence in power on over 500 iterations ~ 100seconds
> 
> > 
> 
> > @anjana348 Could you compare it with and without this PR?
> 
> 
> 
> As we can also see theoretically , In this PR, command mode is not switched in power on so without this PR
> 
> 500 iterations ~ 300second (230 + 70) 
> 
> 70 seconds extra time to switch to command mode in power on which is removed here, rest performance is same
> 
> 

Do you mean saving up 4.2frames per iteration based on 30fps?

anjana348:
> > > > > @anjana348 When you review done, could you leave approval? @zhongnuo-tang @anjana348 We need to measure performance with this PR. Do you have any suggestion for that?
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > We can measure performance similar to fps_test I tried locally using this PR Power off + Power on sequence 500 iterations ~ 230seconds which can be divided as Delay after GPIO pin high in power on - 120ms * 500(iterations) ~ 60seconds Delay after command mode switch in power off - 140ms * 500 ~ 70seconds Time taken in sending init command sequence in power on over 500 iterations ~ 100seconds
> > 
> > 
> > > 
> > 
> > 
> > > @anjana348 Could you compare it with and without this PR?
> > 
> > 
> > As we can also see theoretically , In this PR, command mode is not switched in power on so without this PR
> > 500 iterations ~ 300second (230 + 70)
> > 70 seconds extra time to switch to command mode in power on which is removed here, rest performance is same
> 
> Do you mean saving up 4.2frames per iteration based on 30fps?

this is not FPS performance , it is just for power on/off , not putarea calls.
So for a single power on operation, this PR has removed 140ms time to switch to command mode, which was clear from code also - DelayMS(140)

sunghan-chang (approved):

sunghan-chang:
@ewoodev @anjana348 Please leave approval when you review done by today.

PR #6736, opened 2025-03-21
build/tools/amebasmart: Modify KM4 IPC print

jwei5:
- Add condition to check TizenRT reboot reason before doing IPC print
- IPC print will be disabled if TizenRT reboot reason changed (eg. while handling assert)


sunghan-chang:
@jwei5 could you confirm the log comes out again when the reboot reason is initialized?

jeongarmy (approved):
LGTM

sunghan-chang:
@jwei5 Could you add description about re-flushing with reboot reason initialization in the commit?

jwei5:
> @jwei5 Could you add description about re-flushing with reboot reason initialization in the commit?

done

sunghan-chang (approved):

sunghan-chang:
@jwei5 I don't know which PR has a change on km0_km4_app.bin. Could you check and let the owner know rebasing?

PR #6737, opened 2025-03-21
kernel/mem_leak_checker: Disable km4 log print when mem leak checking

jeongarmy:
A thread of mem leak checker has the highest priority to check all heaps. At this time, log thread can not run and print logs from km4 because of mem leak checker. Then some error logs will be shown like below.
WARN: KM4 logbuf full, dropped log!
WARN: KM4 logbuf full, dropped log!

So we set reboot reason temporaily while checking mem leaks to avoid km4 log print.

### DIFF
--- a/os/kernel/debug/mem_leak_checker.c
+++ b/os/kernel/debug/mem_leak_checker.c
@@ -32,6 +32,9 @@
 #include <binary_manager/binary_manager.h>
 #include <arch/chip/memory_region.h>
 #include <tinyara/binfmt/elf.h>
+#if defined(CONFIG_ARCH_BOARD_RTL8730E)
+#include <tinyara/reboot_reason.h>
+#endif
 
 /****************************************************************************
  * Pre-processor Definitions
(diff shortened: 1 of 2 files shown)


os/include/tinyara/reboot_reason.h
```diff
@@ -46,7 +46,9 @@ typedef enum {
 	REBOOT_NETWORK_WIFICORE_WATCHDOG   = 80, /* Wi-Fi Core Watchdog Reset */
 	REBOOT_NETWORK_WIFICORE_PANIC      = 81, /* Wi-Fi Core Panic */
 
-	REBOOT_BOARD_SPECIFIC1             = 250, /* Board Specific Reboot Reason */
+	REBOOT_BOARD_SPECIFIC6             = 246, /* Board Specific Reboot Reason */
```
sunghan-chang, line 49: Let's split a commit for this.

os/kernel/debug/mem_leak_checker.c
```diff
@@ -377,6 +380,11 @@ int run_mem_leak_checker(int checker_pid, char *bin_name)
 	int broken_cnt = 0;
 	struct mm_heap_s *heap = NULL;
 
+#if defined(CONFIG_ARCH_BOARD_RTL8730E)
+	/* Set reboot reason temporaily to avoid km4 log print */
+	WRITE_REBOOT_REASON(REBOOT_TEMP_BLOCK_KM4_LOG);
```
ewoodev, line 385: How about make up_pause_subcore_logging()?

The KM4 is related to only rtl8730e chip.

PR #6743, opened 2025-03-26
os/drivers/lcd/mipi_lcd.c/: Fix LOGO when fill data in buffer interchange MSB and LSB bits

namanjain7:
The issue was that LOGO couldn't be displayed properly due to MSB and LSB bits were interchanged. This commit  fix this bug. Increase in sharpness in LOGO


sunghan-chang (approved):

jeongarmy (approved):
lgtm

PR #6748, opened 2025-04-01
Add PM ops structure

neel-samsung:


os/pm/pm_idle.c
```diff
@@ -53,6 +53,25 @@ static clock_t stime;
...
+ *
+ * Input Parameters:
+ *   ops: pm operations to use.
+ *
+ * Returned Value:
+ *   None
+ *
+ ****************************************************************************/
+
+void pm_set_ops(const struct platform_pm_ops *ops)
```
ewoodev, line 70: Let's merge wtih void pm_initialize(ops)

os/pm/pm.h
```diff
@@ -144,6 +144,8 @@ struct pm_global_s {
  * Public Data
  ****************************************************************************/
 
+static const struct platform_pm_ops *pm_ops;
```
ewoodev, line 147: How about include it into g_pmglobals??

neel-samsung: yes, It's better to declare it into g_pmglobals as we have only one PM for the system.
Let me change it.

edwakuwaku:
Since you are updating pm structure, suggest to also create an enum for dvfs range instead of int. Since the implementation for different SoC might varies, it will be better to have a unified range level. For example, full perf (100%), mid perf(60%), low perf(30%); definition can be formalised later. Then, when dvfs is applied from either upper layer or kernel mechanism, it can just simply pass in the desired range, and BSP should do the rest.
You may create another PR if necessary.

os/include/tinyara/pm/pm.h
```diff
@@ -264,6 +266,25 @@ struct pm_callback_s {
...
+ *
+ * @adjust_dvfs: It adjusts the DVFS level according to the division level.
+ *  It helps by reducing the frequency of operation of the core.
+ *
+ * @set_timer: Set the wakeup timer
+ */
+
+struct platform_pm_ops {
+	void (*sleep)(pm_wakehandler_t handler);
+	void (*adjust_dvfs)(int div_lvl);
```
ritesh55555, line 284: do we need to apply CONFIG_PM_DVFS check here?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -107,7 +116,8 @@ void up_idle(void)
 #ifdef CONFIG_PM
 void arm_pminitialize(void)
 {
+
```
ritesh55555, line 119: redundant space

os/include/tinyara/pm/pm.h
neel-samsung: @ewoodev could you also let me know your opinion?

os/pm/pm.h
```diff
@@ -138,6 +138,9 @@ struct pm_global_s {
 
 	/* Indicates Board is Ready to State Change */
 	bool is_running;
+
+	/* PM ops structure */
+	struct platform_pm_ops *pm_ops;
```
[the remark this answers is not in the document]
neel-samsung: yes, we should prefer ops as we have shifted it to ```pm_global_s```. Let me change.

os/include/tinyara/pm/pm.h
neel-samsung: yes, I also think, It would be better to separate it.

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -40,6 +40,18 @@
 #endif
 
 #ifdef CONFIG_PM
+
+struct platform_pg_ops system_pg_ops = {
```
ewoodev, line 44: minor, how about like?
```suggestion
struct pm_sleep_ops rtl8730e_sleep_ops = {
```

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -40,6 +40,18 @@
...
+	.sleep = up_pm_board_sleep,
+	.set_timer = up_set_pm_timer,
+};
+
+#ifdef CONFIG_PM_DVFS
+struct platform_cg_ops system_cg_ops = {
```
ewoodev, line 50: struct pm_clock_ops rtl8730e_clock_ops = {

ewoodev (approved):

samsung-singh (approved):
added PM ops struct and pm clock ops struct this change is good to merge.

Sambhavi-1 (approved):
Added pm_sleep and pm_clock ops struct

samsung-agnel (approved):

kishore-sn (approved):

sunghan-chang (approved):

PR #6749, opened 2025-04-01
os/mm/mm_heap: Optimize alloc fail recovery operation.

ewoodev:
If security level is high, we must not print memory address and dumps. Therefore, we added secure state checking in mfdbg and head_dbg.

However, there are many memory dumps. Even if we add secure state checking to mfdbg adn heap_dbg, checking the secure state may cause delays in recovery when outputting every log lines.

Therefore, add to check secure state when the first alloc function is called, and if the secure state is active, do not call any memory printing.


os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -97,6 +99,18 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 {
 	irqstate_t flags = enter_critical_section();
 
+	/* If secure state, do not print memory usage and address infomation */
```
sunghan-chang, line 102: Don't you need to move this after cpu pause (line 119)?

sunghan-chang: hmmm... little bit duplicated and missing conditional.
How about below?
```
void mm_manage_alloc_fail(...)
{
    enter_critical_section();
    cpu pause all
    abort_mode = true;
    write reboot reason ();

    if (!IS_SECURE_STATE()) {   <---- only if is added, rest are existed codes.
       heap messages
    }

   cpu resume all
   panic();
   leave_critical_section();
}
```

ewoodev: we don't need to call cpu pause. 
the cpu pause is for only log mixing

ewoodev: I removed duplicated code.

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -109,12 +109,6 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	abort_mode = true;
```
sunghan-chang, line 109: When the secure level is high, this is not called. Is it ok?
@r-prabu @kishore-sn Could you review this?

os/mm/mm_heap/mm_manage_allocfail.c
```diff
@@ -191,6 +185,30 @@ void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size
 	 */
 	up_cpu_resume_all();
 #endif
+}
+
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
+void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size_t size, size_t align, int heap_type, mmaddress_t caller)
+#else
+void mm_manage_alloc_fail(struct mm_heap_s *heap, int startidx, int endidx, size_t size, size_t align, int heap_type)
+#endif
```
sunghan-chang, line 194: Regardless of this PR, how about using just one api?
When the config is disabled, we can give null or 0 for additional argu.

@r-prabu @kishore-sn How do you think?

r-prabu: I think this is a good suggestion so that maintainability will be easy.

os/mm/mm_heap/mm_manage_allocfail.c
r-prabu: Not sure about it. With the secure state are we restricting only the console logs or log dump as well? If we are only restricting the console logs then will this change wont affect the logdump to these alloc fail logs?

ewoodev: If security level is high, we don't print alloc fail log, so, we will call PANIC(). 
The setting abort_mode is in the up_assert().

os/mm/mm_heap/mm_manage_allocfail.c
ewoodev: How about raise new PR about this suggestion?

os/mm/mm_heap/mm_manage_allocfail.c
sunghan-chang: https://github.com/Samsung/TizenRT/commit/327e46f86bb3a4f5599e22435d708ca4bafb4396

This is conflict with this commit. The variable has too many usages.

@kishore-sn @r-prabu @neel-samsung @ewoodev

seokhun-eom24:
LGTM

allen-kim-sec (approved):
looks good

PR #6750, opened 2025-04-02
os/board/rtl8730e/src/ fix audio not played when last audio buffer is…

zhongnuo-tang:
… small

1. during the last buffer, we should ensure all dma pages have been tx complete before callback to application to terminate it. if it is not the last buffer, we need not to wait for all dma pages to be tx to prevent latency of updating the data from application that might lead to noise. Here we assume the last buffer is not aligned with page*size.
2. remove duplicate memcpy of data to dma link list in i2s_send_page
3. remove unnecessary memset of ptx_buf to 0 when remaining bytes equals to tx_size, change condition from less than equals to less than.

Verified with test audio:
soundplayer /mnt/audio_cut_10ms_raw.raw 1 48000 1
soundplayer /mnt/audio_cut_2ms_raw.raw 1 48000 1
soundplayer /mnt/audio_cut_5ms_raw.raw 1 48000 1
soundplayer /mnt/audio_cut_raw.raw 1 48000 1

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -407,7 +407,7 @@ static int amebasmart_i2s_tx(struct amebasmart_i2s_s *priv, struct amebasmart_bu
 			ptx_buf = i2s_get_tx_page(priv->i2s_object);
 			i2s_enable(priv->i2s_object);
 			if (ptx_buf) {
-				if ((apb->nbytes - apb->curbyte) <= tx_size) {
+				if ((apb->nbytes - apb->curbyte) < tx_size) {
 					tx_size = apb->nbytes - apb->curbyte;
 					memset(ptx_buf, 0, I2S_DMA_PAGE_SIZE); /* Clear ptx_buf to prevent sending old data since we are sending less than I2S_DMA_PAGE_SIZE */
 					memcpy((void *)ptx_buf, (void *)&apb->samp[apb->curbyte], tx_size);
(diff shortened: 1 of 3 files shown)


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -559,7 +559,16 @@ static void i2s_tx_schedule(struct amebasmart_i2s_s *priv, int result)
 		/* Start next transfer */
 		amebasmart_i2s_tx(priv, bfcontainer);
 	} else if ((priv->apb_tx->nbytes - priv->apb_tx->curbyte) <= 0) {
-		ameba_i2s_pause(priv->i2s_object);
+		if (i2s_dma_tx_done(I2S_DMA_PAGE_NUM) == OK ){
+			/*if all dma tx done, we should pause i2s dma to mute it*/
+			ameba_i2s_pause(priv->i2s_object);
+		} else if (priv->apb_tx->nbytes < I2S_DMA_PAGE_SIZE * I2S_DMA_PAGE_NUM) {
```
Taejun-Kwon, line 565: I think if remained data is greater than 1 slot of dma buffer, it should send to tx immediately.
Otherwise this will occur underrun if next i2s_send delayed. It should consume remained data if there's remained data.

zhongnuo-tang: If there is pending container to send,  !sq_empty(&priv->tx.pend) will send to tx immediately.
When we reached here, there should not be any remain data in the current container because in i2s_transfer_tx_handleirq is checks for remaining data in container, and only callback when all data are moved to dma.

Taejun-Kwon: ah, ok I agree with you. Let us test something more, and I will merge if everything work fine. thanks.

Taejun-Kwon (approved):

PR #6753, opened 2025-04-02
external/resample: Add config for media resampler

aadotverma:
1. Add CONFIG_MEDIA_RESAMPLER for external/resampler.
2. It will be n by default.
3. If CONFIG_MEDIA is y, then CONFIG_MEDIA_RESAMPLER will also be set
to y automatically.


sunghan-chang (dismissed):

framework/src/media/Kconfig
```diff
@@ -9,6 +9,7 @@ config MEDIA
 	depends on AUDIO
 	select HAVE_CXX
 	select HAVE_CXXINITIALIZE if BUILD_FLAT
+	select MEDIA_RESAMPLER
```
Taejun-Kwon, line 12: I don't mean this, it should be removed here.

external/resample/Kconfig
```diff
@@ -0,0 +1,11 @@
+#
+# For a description of the syntax of this configuration file,
+# see kconfig-language at https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt
+#
+
+config MEDIA_RESAMPLER
+	bool "Media Resampler"
```
Taejun-Kwon, line 7: depend on MEDIA here, so resampler can be enabled when user enable CONFIG_MEDIA

build/configs/rtl8730e/loadable_ext_ddr/defconfig
```diff
@@ -329,8 +329,8 @@ CONFIG_SECOND_FLASH_START_ADDR=0x0A000000
 CONFIG_RTL8730E_BOARD_REVISION=8
 CONFIG_RAM_DDR=y
 # CONFIG_RAM_PSRAM is not set
-CONFIG_USERFS_MNTPT="/mnt"
-CONFIG_USERFS_EXT_MNTPT="/ext"
+CONFIG_USERFS_MNTPT="/mnt0"
```
vibhor-m, line 332: we should not do this change. this is not related to us

framework/src/media/Kconfig
vibhor-m: I think this required.

Taejun-Kwon (approved):

PR #6760, opened 2025-04-09
arch/arm: fix variable conflict by adding new one for recursive abort

neel-samsung:
This patch adds new variable abort_state for recursive abort to avoid conflict with existing variable abort_mode.

Test results:
=========================================================== Loading location information
=========================================================== elf_show_all_bin_section_addr: [common] Text Addr : 0xe161010, Text Size : 3112960 elf_show_all_bin_section_addr: [app1] Text Addr : 0xe459030, Text Size : 1417216

print_dataabort_detail: #########################################################################
print_dataabort_detail: PANIC!!! Data Abort at instruction : 0x0e001340
print_dataabort_detail: PC: 0e001340 DFAR: 00000000 DFSR: 0000080d
print_dataabort_detail: #########################################################################

boardctl: Board will Reboot now. pid: 20
ROM:[V1.1]

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -144,6 +144,9 @@ extern int g_irq_num[CONFIG_SMP_NCPUS];
 #define LOG_TASK_NAME   "N/A"
 #endif
 
+#define NORMAL_STATE 0
+#define ABORT_STATE 1
+
 /****************************************************************************
  * Public Variables
  ****************************************************************************/
@@ -152,6 +155,9 @@ char assert_info_str[CONFIG_STDIO_BUFFER_SIZE] = { '\0', };
 /****************************************************************************
  * Private Data
  ****************************************************************************/
+
+/* Variable to check the recursive abort */
+static int state = NORMAL_STATE;
 
 /****************************************************************************
  * Private Functions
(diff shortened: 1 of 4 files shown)


os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -108,6 +108,9 @@
 
 bool abort_mode = false;
 
+/* Variable to check the recursive abort */
+static bool abort_state  = false;
```
r-prabu, line 112: check if need to use volatile keyword for this?

samsung-singh (approved):
fixed variable conflict by adding new variable for recursive abort this change is good to merge.

os/arch/arm/src/armv7-a/arm_assert.c
sunghan-chang, line 112: Abort state is true or false? Looks strange. If you want to name it state, then let's use definitions.
NORMAL STATE 0
ABORT STATE 1

neel-samsung: okay, but we are using it in normal computation. So, we won't need the keyword.

neel-samsung: Done

r-prabu (approved):

Sambhavi-1 (approved):
variable conflict fixed by adding new variable for recursive abort

kishore-sn (approved):

PR #6761, opened 2025-04-10
os/board/rtl8730e/src: add timeout for i2c read

zhongnuo-tang:
1. add timeout to break from infinite loop when waiting for i2c rx flag in case error occur in middle of rx
2. in case where i2c stucked in trap state, we should only re-try limited times to prevent infinite loop.

### DIFF
--- a/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
+++ b/os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
@@ -273,7 +273,8 @@ int rtk_i2c_read(i2c_t *obj, int address, char *data, int length, int stop)
 {
 	/* To avoid gcc warnings */
 	(void) stop;
-
+	int data_read = 0;
+	int cnt = 10;
 	if (i2c_target_addr[obj->i2c_idx] != address) {
 		i2c_prepare(obj, address);
 	}
@@ -281,15 +282,21 @@ int rtk_i2c_read(i2c_t *obj, int address, char *data, int length, int stop)
 	if (!master_addr_retry) {
 		I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length);
 	} else {
-		while (0 == I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length)) {
+		while ((data_read = I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length)) == 0) {
 			/* Wait for i2c enter trap state from trap_stop state*/
 			DelayUs(100);
 
 			i2c_prepare(obj, address);
+			/* we will re-try 10 times, if it still failed to read any data, we will break it to prevent infinite loop*/
+			if (cnt == 0) {
+				break;
+			}
+			cnt--;
+
 		}
 	}
 
-	return length;
+	return data_read;
 }
 
 /**
--- a/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
+++ b/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
@@ -50,11 +50,11 @@
   */
 
 #include "ameba_soc.h"
-
+#include <errno.h>
 static const char *TAG = "I2C";
 
 #define I2C_SPEED_FINE_TUNE 0
-
+#define I2C_TIMEOUT 2000
 const I2C_DevTable I2C_DEV_TABLE[3] = {
 #ifdef ARM_CORE_CM4
 	{I2C0_DEV, I2C0_IRQ},
@@ -617,9 +617,17 @@ u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 
 	/* Write in the DR register the data to be sent */
 	for (cnt = 0; cnt < len; cnt++) {
-		InTimeoutCount = 1000;
+		InTimeoutCount = I2C_TIMEOUT; /*2s*/
 
-		while ((I2C_CheckFlagState(I2Cx, I2C_BIT_TFNF)) == 0);
+		while ((I2C_CheckFlagState(I2Cx, I2C_BIT_TFNF)) == 0) {
+			DelayUs(1);
+			if (InTimeoutCount == 0) {
+				I2C_Cmd(I2Cx, DISABLE);
+				I2C_Cmd(I2Cx, ENABLE);
+				return -EAGAIN;
+			}
+			InTimeoutCount--;
+		}
 
 		if (cnt >= len - 1) {
 			/*generate stop signal*/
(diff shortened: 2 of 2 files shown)


os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
```diff
@@ -273,23 +273,30 @@ int rtk_i2c_read(i2c_t *obj, int address, char *data, int length, int stop)
...
+	int cnt = 10;
 	if (i2c_target_addr[obj->i2c_idx] != address) {
 		i2c_prepare(obj, address);
 	}
 
 	if (!master_addr_retry) {
 		I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length);
 	} else {
-		while (0 == I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length)) {
+		while ((data_read = I2C_MasterRead(obj->I2Cx, (unsigned char *)data, length)) == 0) {
```
sunghan-chang, line 285: Is there any chance to be less size than the length?
Regardless of the length, if i2c gives some data, is it ok to return without gathering all datas?

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -717,6 +718,14 @@ u32 I2C_MasterRead(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 				I2C_ClearAllINT(I2Cx);
 				return cnt;
 			}
+			DelayUs(1);
+			if (timeout == 0) {
+				RTK_LOGW(TAG, "I2C_MasterRead timeout - resetting I2C\n");
```
sunghan-chang, line 723: could you let me know which Tizen Lite api is called by RTK_LOGW?

zhongnuo-tang: DiagVSprintf

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
sunghan-chang: Oh.. I saw this partially. (I did not check the change in ameba_i2c.c). It's ok.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
zhongnuo-tang: do you have any suggestion which API to use?

sunghan-chang: As far as I know, Realtek has other API to flash logs into SW buffer, not into uart hw. I don't know it. Could you check it in Realtek internally?

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -717,6 +718,14 @@ u32 I2C_MasterRead(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
...
+				I2C_Cmd(I2Cx, DISABLE);
+				I2C_Cmd(I2Cx, ENABLE);
```
sunghan-chang, line 725: When all datas are read, you don't call these disable and enable. Is this necessary when we don't get full data only? If we call this, what happen?

zhongnuo-tang: We don't have to call it when all data are read, this is normally called when we want to change slave addr, or when it enter a trap state. When we are stuck in waiting fifo flag, it is more likely that it enter a trap state so required a I2C reset, so to prepare for next read. 
But even if we call it when getting all data, it should not create any issue.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
zhongnuo-tang: I see, let me check this, but i heard from Eunwoo that printf/dbg should not be used as it will be used in idle_pm and will cause context switching..

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -701,9 +701,10 @@ void I2C_MasterReadDW(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 u32 I2C_MasterRead(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 {
 	u32 cnt = 0;
-
+	u32 timeout;
 	/* read in the DR register the data to be received */
 	for (cnt = 0; cnt < len; cnt++) {
+		timeout = 2000; /*timeout 2ms*/
```
sunghan-chang, line 707: Is this proper value? If the len is much big, then, does not this value cause timeout at normal working?

zhongnuo-tang: Yeah, because timeout will reset to 2000 (2ms) every time it enter the loop, so it is 2ms for each data, not for the entire data length.

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
ewoodev: How about remove log and return error?

zhongnuo-tang: I think it would be better. let me change it

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -715,8 +715,15 @@ u32 I2C_MasterRead(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
...
 			if (I2C_GetRawINT(I2Cx) & I2C_BIT_TX_ABRT) {
 				I2C_ClearAllINT(I2Cx);
-				return cnt;
+				return -1;
+			}
+			DelayUs(1);
+			if (timeout == 0) {
+				I2C_Cmd(I2Cx, DISABLE);
+				I2C_Cmd(I2Cx, ENABLE);
+				return -1;
```
ewoodev, line 724: How about return `-ETIMEDOUT`

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -715,8 +715,15 @@ u32 I2C_MasterRead(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 		while ((I2C_CheckFlagState(I2Cx, I2C_BIT_RFNE)) == 0) {
 			if (I2C_GetRawINT(I2Cx) & I2C_BIT_TX_ABRT) {
 				I2C_ClearAllINT(I2Cx);
-				return cnt;
+				return -1;
```
ewoodev, line 718: what is the error?
Let's seperate error type.
If this error is needed try again, Please return -EAGAIN or 0

zhongnuo-tang: I have changed all to -EAGAIN

os/board/rtl8730e/src/component/soc/amebad2/fwlib/ram_common/ameba_i2c.c
```diff
@@ -617,31 +617,38 @@ u32 I2C_MasterWriteBrk(I2C_TypeDef *I2Cx, u8 *pBuf, u32 len)
 
 	/* Write in the DR register the data to be sent */
 	for (cnt = 0; cnt < len; cnt++) {
-		InTimeoutCount = 1000;
+		InTimeoutCount = 2000; /*2s*/
```
edwakuwaku, line 620: define a macro instead of hardcoding

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/i2c_api.c
```diff
@@ -273,23 +273,30 @@ int rtk_i2c_read(i2c_t *obj, int address, char *data, int length, int stop)
 {
 	/* To avoid gcc warnings */
 	(void) stop;
-
+	int data_read = 0;
+	int cnt = 10;
```
edwakuwaku, line 277: why 10?

ewoodev (approved):

sunghan-chang (approved):

PR #6762, opened 2025-04-10
mm/mm_heap: Add to print size of pending free in heapinfo logs

ewoodev:
If the CPU idle state is not present and lpwork is not running, the delayed free operation of sched_free may not function properly.

To check this, Add to display the size of any delayed free that has not yet been released in heapinfo command.

### DIFF
--- a/os/mm/mm_heap/mm_heapinfo_parse_heap.c
+++ b/os/mm/mm_heap/mm_heapinfo_parse_heap.c
@@ -107,13 +107,16 @@ static inline int check_watchdog_state(int wd_fd, bool *is_watchdog_running)
  ****************************************************************************/
 void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 {
+	irqstate_t flags;
 	struct mm_allocnode_s *node;
+	void *address;
 	size_t mxordblk = 0;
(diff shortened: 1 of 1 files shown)


os/mm/mm_heap/mm_heapinfo_parse_heap.c
```diff
@@ -252,6 +256,40 @@ void heapinfo_parse_heap(FAR struct mm_heap_s *heap, int mode, pid_t pid)
 		}
 		mm_givesemaphore(heap);
 	}
+
+	delayfree_resource = 0;
+	flags = enter_critical_section();
+	for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
```
edwakuwaku, line 262: why only this particular code piece is needed to be in critical zone?

eunwoo-nam: https://github.com/Samsung/TizenRT/blob/f669e8692cdbbda79cd44b2670673a720c355c54/os/mm/mm_heap/mm_malloc.c#L101-L124

Please refer above code lines.
If mm_heapinfo_parse_heap.c printing is excuting and other core is free delaylist, the information can be wrong or the data abort can be occured.

PR #6769, opened 2025-04-11
Timeout peripherals UART and LCDC

zhongnuo-tang:
1. remove busy-wait in serial_putc and serial_getc
2. add timeout for lcdc enable, and retry if enabled failed


os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -237,26 +237,41 @@ static void rtl8730e_control_backlight(uint8_t level)
...
 	LCDC_Cmd(pLCDC, ENABLE);
-	while (!LCDC_CheckLCDCReady(pLCDC)) ;
+	while (!LCDC_CheckLCDCReady(pLCDC)) {
+		DelayMs(1);
+		if (timeout == 0) {
+			return FAIL;
+		}
+		timeout--;
+	}
+	return OK
```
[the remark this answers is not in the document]
zhongnuo-tang: added

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -237,25 +237,31 @@ static void rtl8730e_control_backlight(uint8_t level)
...
 	LCDC_Cmd(pLCDC, ENABLE);
-	while (!LCDC_CheckLCDCReady(pLCDC)) ;
+	while (!LCDC_CheckLCDCReady(pLCDC)) {
+		DelayMs(1);
+		if (timeout == 0) {
+			return FAIL;
+		}
+		timeout--;
+	}
+	return OK;
```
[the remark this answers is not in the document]
zhongnuo-tang: added assert

sunghan-chang:
@zhongnuo-tang Could you let me know how do you verify the changes?
For 3rd, could you try malloc with small size again and again? When it has no memory for allocation, it will print all nodes.
At that case, we can get no `\n` issue easily.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/serial_api.c
```diff
@@ -741,7 +740,6 @@ void serial_putc(serial_t *obj, int c)
 {
 	PMBED_UART_ADAPTER puart_adapter = &(uart_adapter[obj->uart_idx]);
 
-	while (!serial_writable(obj));
```
sunghan-chang, line 744: Serial driver checks `uart_txready`. It looks different with this `serial_writable`. 
The `uart_txready` returns sw status, `serial_writable` does hw status.
Do you think sw status can guarantee serial status instead of hw status?

UART issue will be critical. Please let me know your test way and result.

zhongnuo-tang: I have made this changes with reference to tp1x where it fixes uart data loss during flash operation.

https://github.com/Samsung/TizenRT/commit/8ee1b95bb635454c31dbbc646eb747a6c46d585d#diff-ce9702b589ceb0508b8098f5ce272259dd14285ee17095d7e34a3aa6bd05779f

I will perform some stress test and share you the results.

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1291,6 +1291,14 @@ void up_serialinit(void)
 void up_lowputc(char ch)
 {
 #ifdef CONFIG_UART4_SERIAL_CONSOLE
+	u32 CounterIndex = 0;
+	/*check if there is space in fifo*/
+	while(!LOGUART_Ready()) {
+		CounterIndex++;
+		if (CounterIndex >= CHAR_TIMEOUT) {
+			break;
```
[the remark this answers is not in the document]
zhongnuo-tang: Yes the timeout value is verified, it was originally from ameba_log_uart.c. 
https://github.com/Samsung/TizenRT/commit/ed36644b8fc35347885599cbce71777742e5d1d0#diff-ef625aa06185fea6a4dea8e1350e0840bd185230c645d6d71172bb71399b7283

If timeout happen, the char will be discarded and this will only happen in low level dbg, where data are putting into fifo directly, and as i check for the caller of up_putc, there are no checking for status, so there is no re-sending of bytes..

In printf/dbg, it is handled by irq and there is a check for fifo with uart_txready, it will only calls into LOGUART_PutChar when there is available fifo.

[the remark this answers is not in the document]
zhongnuo-tang: Hi @sunghan-chang ,
I have verified this PR with a loopback test sending 160bytes of data in a while loop. There is no data loss in a long run test. 

The UART TX FIFO has a depth of 16 bytes. In the TX interrupt handler (TxIrq), data is transmitted in 16-byte chunks, matching the FIFO size. This ensures optimal use of the FIFO and minimizes the number of interrupts required.

The TX interrupt is only triggered when the TX FIFO is completely empty. This guarantees that when uart_xmitchars() is called in response to the interrupt, the FIFO has enough space to accept up to 16 new bytes of data.

For RX case, rtl8730e_up_rxavailable already ensures that there is available data in fifo before calling receive.

os/board/rtl8730e/src/component/mbed/targets/hal/rtl8730e/serial_api.c
zhongnuo-tang: Hi @sunghan-chang,

I have verified this PR with a loopback test sending 160bytes of data in a while loop. There is no data loss in a long run test.
The UART TX FIFO has a depth of 16 bytes. In the TX interrupt handler (TxIrq), data is transmitted in 16-byte chunks, matching the FIFO size. This ensures optimal use of the FIFO and minimizes the number of interrupts required.

The TX interrupt is only triggered when the TX FIFO is completely empty. This guarantees that when uart_xmitchars() is called in response to the interrupt, the FIFO has enough space to accept up to 16 new bytes of data.

For RX case, rtl8730e_up_rxavailable already ensures that there is available data in fifo before calling receive.

zhongnuo-tang: Update:
currently running longrun with below setting
UART loopback test + smartfs powercut running concurrently.
No data loss so far.

sunghan-chang: @zhongnuo-tang Could you update the longrun test status?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -38,7 +38,7 @@
 #define PIN_LOW 0
 #define PIN_HIGH 1
 #define LCD_LAYER 0
-
+#define LCD_ENABLE_TIMEOUT 5000;
```
sunghan-chang, line 41: Could you write unit of 5000 ? in ms?

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1399,14 +1399,6 @@ int up_getc(void)
  ****************************************************************************/
 int up_putc(int ch)
 {
-	u32 CounterIndex = 0;
```
sunghan-chang: Regardless of this PR,, this #else case (non USE_SERIALDRIVER) is unnecessary.
up_putc and up_getc are perfectly the same with #if case.

os/arch/arm/src/amebasmart/amebasmart_serial.c
```diff
@@ -1343,8 +1351,6 @@ int up_lowgetc(void)
  ****************************************************************************/
 int up_putc(int ch)
 {
-	/*check if there is space in fifo*/
-	while(!LOGUART_Ready());
 	/* Check for LF */
 
 	if (ch == '\n') {
```
sunghan-chang, line 1356:
```
void LOGUART_PutChar(u8 c)
{
	LOGUART_TypeDef *UARTLOG = LOGUART_DEV;
	u32 CounterIndex = 0;
	u32 CPUID = SYS_CPUID();

	UARTLOG->LOGUART_UART_THRx[LOG_UART_IDX_FLAG[CPUID].idx] = c;

	if (c == KB_ASCII_LF) {
		UARTLOG->LOGUART_UART_THRx[LOG_UART_IDX_FLAG[CPUID].idx] = KB_ASCII_CR;
	}
}
```
What's the meaning of the if conditional? Does it append `\r`? or replace `\n` to `\r`?
If it's the append, then this line looks duplicated.

zhongnuo-tang: Hi @sunghan-chang ,
It is append. This API will also be called from printf/dbg, if CR-LF is handled properly at all level log printing, I will remove this line.

os/arch/arm/src/amebasmart/amebasmart_serial.c
zhongnuo-tang: I will check why is this condition added, if unnecessary i will remove in a separable PR.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
zhongnuo-tang: added

sunghan-chang (approved):

PR #6770, opened 2025-04-11
os/board/rtl8730e: Revise BLE while loop condition

ziyik:
-Added 3 second timeout for while
-Prevent hang forever when GAP state not change


hs36-kim:
@ziyik 
if the return case, need to reboot?

ziyik:
> @ziyik if the return case, need to reboot?

No, just the action is fail not need to reset.

hs36-kim (approved):
LGTM

giwon-nam (approved):
https://github.com/Samsung/TizenRT/pull/6800
Build error was addressed with this PR.

PR #6774, opened 2025-04-14
drivers/input: Add suspending PM when a touch input.

ewoodev:
Touch is a user interaction, so when a touch is detected, the power management is delayed
 for a certain period to prevent entering power sleep mode.

Furthermore, when I2C is used for checking the touch IC alive, the PM is also delayed
 to avoid entering power sleep mode and blocking the I2C.

### DIFF
--- a/os/drivers/input/ist415.c
+++ b/os/drivers/input/ist415.c
@@ -56,6 +56,7 @@
 #include <tinyara/wdog.h>
 #include <tinyara/wqueue.h>
 
+#include <tinyara/pm/pm.h>
 #include <tinyara/input/touchscreen.h>
 #include <tinyara/input/ist415.h>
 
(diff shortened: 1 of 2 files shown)


sunghan-chang:
@neel-samsung @anjana348 Could you review this?

os/drivers/input/ist415.c
```diff
@@ -756,6 +759,8 @@ static void touch_interrupt(struct ist415_config_s *config)
 	if (ret == OK && sem_cnt < 1) {
 		sem_post(&dev->wait_irq);
 	}
+
+	(void)pm_timedsuspend(dev->pm, 1000);
```
anjana348, line 763: If pm_timedsuspend is only needed for touch event processing then it can be put inside if condition. Right?
if (ret == OK && sem_cnt < 1) {
		sem_post(&dev->wait_irq);
	}

os/drivers/input/ist415.c
```diff
@@ -1274,11 +1279,21 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
 		ist415dbg("Fail to alloc ist415 wdog\n");
 	}
 
+	dev->pm = pm_domain_register("IST415");
```
anjana348, line 1282: It should be inside CONFIG_PM

ewoodev: even if CONFIG_PM is disabled, pm_domain_register returns 0.
So, It will not be problem.

os/drivers/input/ist415.c
ewoodev: I think, touch event generating and pm wake locking is seperatly
If sem_cnt < 1, we need to refresh wake lock timer. otherwise, if the touch event thread does not receive the event, it may fall into power saving.

sunghan-chang (approved):

PR #6775, opened 2025-04-14
build/configs/rtl8730e: fix floader for UART operation and linux download script

lhenry-realtek:
Add back missing command handling for UART download causing download hang when UART is used instead of USB download

Fix download script for audio/flat* due to improper handling of the loop variable causing the script to skip to end of download without doing anything

PR #6782, opened 2025-04-16
build/tools/amebasmart: bugfix for floader loadable config Flash Image Invalid

lhenry-realtek:
Address an issue where when "loadable" configs are selected, the image is burned, however encounter Flash Image Invalid

Root Cause: 
Top 8KB of BL1 is missing due to invalid address calculation of internal vs external offsets, causing image to not boot 
This issue is not present when **"flat"** configs are selected, and a workaround was to burn BL1 again after burning ALL

Resolution: 
Fix the address check in the floader to correctly determine internal vs external flash addresses

Verification:
loadable_ext_psram config
PSRAM+NOR UART, USB Boot OK
DDR+NOR (AI Lite) UART, USB Boot OK
DDR+NOR (AI Lite+W) UART, USB Boot OK
DDR+NAND (AI Lite+W) UART, USB Boot OK (due to NAND FTL missing, does not boot to shell, but success)


sunghan-chang:
@ritesh55555 Could you check this? But when you verified previous PR #6775 , did not you do that with loadable?
@lhenry-realtek Could you let us know when we can meet the issue?

sunghan-chang:
@lhenry-realtek What about the windows tool? No issue with that? (no partially cut off in windows tool?)

sunghan-chang:
@lhenry-realtek I mean this is only for the linux script?

sunghan-chang:
@lhenry-realtek Sorry but last question. Is this the side effect of the PR #6775 ? or inherent(?) hidden(?) issue?

PR #6784, opened 2025-04-17
build/tools/amebasmart: Fix mdns discover fail after pm wake

jwei5:
- Multicast packets not received after pm wake, causing mDNS discovery to fail
- Fix issue by restoring a register after wake


hs36-kim (approved):
LGTM
I heard that the patch to fix receiving multicase packet has been tested.

PR #6786, opened 2025-04-18
os/drivers/input/ist415.c: Fix memory leaks in ist415_initialize() on error paths

abhinav-s235:
The ist415_initialize () function had several error return paths where allocated resources such as semaphores, watchdog timers, and kernel threads were not properly released. This could lead to memory leaks on initialization failure. This commit ensures that all allocated resources are correctly cleaned up in all failure scenarios.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_ist415.c
+++ b/os/board/rtl8730e/src/rtl8730e_ist415.c
@@ -186,6 +186,7 @@ void rtl8730e_ist415_initialize(void)
 	if (ist415_initialize(TOUCH_DEV_PATH, i2c, dev) < 0) {
 		touchdbg("ERROR: Touch driver register fail\n");
 		up_i2cuninitialize(i2c);
+	} else {
+		touchdbg("Touch driver register success\n");
 	}
-	touchdbg("Touch driver register success\n");
 }
--- a/os/drivers/input/ist415.c
+++ b/os/drivers/input/ist415.c
@@ -1241,8 +1241,7 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
 	dev->pid = kernel_thread("ist415_isr", CONFIG_IST415_WORKPRIORITY , 2048, (main_t)ist415_event_thread, (FAR char *const *)parm);
 	if (dev->pid < 0) {
 		ist415dbg("Fail to create kernel thread\n");
-		kmm_free(dev);
-		return ERROR;
+		goto cleanup_sems_mem;
 	}
 
 	/* Init System Power */
@@ -1258,8 +1257,7 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
 	ret = ist415_get_info((void *)dev);
 	if (ret) {
 		ist415dbg("Fail to get info\n");
-		kmm_free(dev);
-		return ret;
+		goto cleanup_task;
 	} else {
 		ist415vdbg("Get info success\n");
 	}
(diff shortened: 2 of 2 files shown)


os/drivers/input/ist415.c
```diff
@@ -1270,14 +1269,26 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
...
-		kmm_free(dev);
 		ist415_disable(dev);
-		return ERROR;
+		goto cleanup_task_wd;
 	}
 	upper->ops = &g_ist415_ops;
 	upper->priv = dev;
 	dev->upper = upper;
+	ret = touch_register(path, upper);
+	if(ret < 0) {
```
ewoodev, line 1279:
```suggestion
	if (ret < 0) {
```

os/drivers/input/ist415.c
```diff
@@ -1270,14 +1269,26 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
...
+		ist415_disable(dev);
+		goto cleanup_task_wd;
```
ewoodev, line 1281:
```suggestion
		kmm_free(upper);
		goto cleanup_task_wd;
```
we need to release upper.

ewoodev: Could you please add fail log here?

os/drivers/input/ist415.c
```diff
@@ -1270,14 +1269,26 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
...
-	ist415dbg("Touch Driver registered Successfully\n");
-	return touch_register(path, upper);
+/* Error Handling Releasing all allocated resources */
+cleanup_task_wd:
+	wd_delete(dev->wdog);
+	task_delete(dev->pid);
+cleanup_sems_mem:
+	sem_destroy(&dev->sem);
+	sem_destroy(&dev->wait_irq);
+	kmm_free(dev);
```
ewoodev, line 1293: The touch driver registering fail is critical issue.
So, How about add DEBUGASSERT here for detect early?

os/drivers/input/ist415.c
abhinav-s235: fixed

os/drivers/input/ist415.c
abhinav-s235: fixed

ewoodev: Could you please add fail log here?

abhinav-s235: > Could you please add fail log here?

Its already there in touchscreen.c function
should I add it here also?

ewoodev (approved):

os/drivers/input/ist415.c
ewoodev: okay, touch_regiseter is printing fail case log,
we don't need to add log here!

os/drivers/input/ist415.c
```diff
@@ -1270,14 +1269,27 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
...
+		goto cleanup_task_wd;
 	}
 	upper->ops = &g_ist415_ops;
 	upper->priv = dev;
 	dev->upper = upper;
+	ret = touch_register(path, upper);
+	if (ret < 0) {
+		ist415_disable(dev);
+		kmm_free(upper);
+		goto cleanup_task_wd;
```
sunghan-chang, line 1282: I don't wanna keep several locations to recover allocated resources at failure.
Could you move all free, disable, delete at label?

abhinav-s235: fixed

os/board/rtl8730e/src/rtl8730e_ist415.c
```diff
@@ -186,6 +186,7 @@ void rtl8730e_ist415_initialize(void)
 	if (ist415_initialize(TOUCH_DEV_PATH, i2c, dev) < 0) {
 		touchdbg("ERROR: Touch driver register fail\n");
 		up_i2cuninitialize(i2c);
-	}
+	} else {
 	touchdbg("Touch driver register success\n");
+	}
```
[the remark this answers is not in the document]
abhinav-s235: In original code success log was being printed in case of failure

PR #6791, opened 2025-04-21
os/kernel/sched: Fix assert fail in remove readytorun

kishore-sn:
Assert fail is observed at below lines of code.

	ntcb = (FAR struct tcb_s *)rtcb->flink;
	DEBUGASSERT(ntcb != NULL);

However, we need to check the flink only if the list is runnable. Otherwise the list can be empty and we dont need to check flink.


r-prabu (approved):
Changes for fixing the assert in remove readytorun state is fine to take.

sunghan-chang (approved):

Sambhavi-1 (approved):
Assertion fail in remove readytorun is fixed

PR #6792, opened 2025-04-21
os/include/tinyara/net/if : Fix the enum values so they do not change

giwon-nam:
- The enum values are currently defined as common / server / client / advertise / scan, so when a new API is added, the enum values are inserted in between. As a result, the enum values change every time a new API is added, causing compatibility issues with previous versions. To resolve this, the enum values have been modified to have fixed values.


hs36-kim (approved):
LGTM

PR #6794, opened 2025-04-21
os/drivers/compression: To decompress the compressed file

Sambhavi-1:
Previously this driver exposed ioctls for user to compress or decompress data from a buffer.
This PR adds new ioctls which will allow user to decompress data directly from a file. 

COMPIOC_FDCMP_INIT: Pass file path and init the decomp framework in kernel
COMPIOC_FDCMP_GET_BUFSIZE: Get the buf size required to hold decompressed data. User needs to allocate buffer of this size
COMPIOC_FDCMP_DECOMPRESS: Pass buffer and recieve decompressed data
COMPIOC_FDCMP_DEINIT: Cleanup the kernel decomp framework


os/drivers/compression/compress.c
```diff
@@ -23,10 +23,12 @@
 #include <tinyara/fs/fs.h>
 #include <tinyara/fs/ioctl.h>
 #include <tinyara/compression.h>
+#include <tinyara/binfmt/compression/compress_read.h>
```
kishore-sn, line 26: Please change the commit title and body as follows and format it properly.

```
os/drivers/compression: Implemention for file decompression

Previously this driver exposed ioctls for user to compress or decompress data from a buffer.
This PR adds new ioctls which will allow user to decompress data directly from a file. 

COMPIOC_FDCMP_INIT: Pass file path and init the decomp framework in kernel
COMPIOC_FDCMP_GET_BUFSIZE: Get the buf size required to hold decompressed data. User needs to allocate buffer of this size
COMPIOC_FDCMP_DECOMPRESS: Pass buffer and recieve decompressed data
COMPIOC_FDCMP_DEINIT: Cleanup the kernel decomp framework
```

os/drivers/compression/compress.c
```diff
@@ -87,9 +95,12 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
 static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 {
 	int ret = -ENOSYS;
+	unsigned int size;
+	off_t filelen;
 	struct compress_header *comp_info = (struct compress_header *)arg;
+	struct file_decomp_data_s *data;
 
-	if (comp_info == NULL && cmd != COMPIOC_GET_COMP_TYPE) {
+	if (comp_info == NULL && !((cmd == COMPIOC_GET_COMP_TYPE) || (cmd==COMPIOC_GET_COMP_HEADER) || (cmd ==COMPIOC_UNINIT))) {
```
kishore-sn, line 103: This condition seems wrong and unnecessary. Please change it as below.

```
if (comp_info == NULL) {
    return -EINVAL;
}
```

Also, add a default case in the case statement and return EINVAL from there.

os/include/tinyara/fs/ioctl.h
```diff
@@ -474,6 +474,10 @@
 #define COMPIOC_GET_COMP_TYPE      _COMPIOC(0x0002)
 #define COMPIOC_GET_COMP_NAME      _COMPIOC(0x0003)
 #define COMPIOC_DECOMPRESS	   _COMPIOC(0x0004)
+#define COMPIOC_INIT               _COMPIOC(0x0005)
+#define COMPIOC_GET_COMP_HEADER    _COMPIOC(0x0006)
+#define COMPIOC_FILE_DECOMPRESS    _COMPIOC(0x0007)
+#define COMPIOC_UNINIT             _COMPIOC(0x0008)
```
kishore-sn, line 480: Please change the ioctl names as below:

```
COMPIOC_FDCMP_INIT
COMPIOC_FDCMP_GET_BUFSIZE
COMPIOC_FDCMP_DECOMPRESS
COMPIOC_FDCMP_DEINIT
```

os/drivers/compression/compress.c
```diff
@@ -120,6 +131,36 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		if (ret == ENOMEM) {
 			bcmpdbg("Output buffer allocated is not sufficient\n");
 		}
+	case COMPIOC_INIT:
+		data = (struct file_decomp_data_s *)malloc(sizeof(struct file_decomp_data_s));
+		data->fd = open((char *)arg, "r");
```
kishore-sn, line 136: Please do a null check for `arg` before using it.

os/drivers/compression/compress.c
```diff
@@ -120,6 +131,36 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		filep->f_priv = data;
+		if (data->fd < 0) {
+			int errval = get_errno();
+			berr("Failed to open file ERROR = %d\n", errval);
```
kishore-sn, line 140: Let us print the file path in the error statement.

os/drivers/compression/compress.c
```diff
@@ -120,6 +131,36 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		break;
+	case COMPIOC_GET_COMP_HEADER:
+		data = filep->f_priv;
+		data->compression_header = get_compression_header();
+		ret = data->compression_header->binary_size;
+		break;
+	case COMPIOC_FILE_DECOMPRESS:
+		data = filep->f_priv;
+		size = compress_read(data->fd, 0, (uint8_t *)arg, data->compression_header->binary_size, 0);
+		lldbg("size after decompression %d \n",size);
```
kishore-sn, line 153: Let us remove this log.

os/include/tinyara/fs/ioctl.h
Sambhavi-1: Changed as per suggestion

os/drivers/compression/compress.c
Sambhavi-1: done

os/drivers/compression/compress.c
Sambhavi-1: Done

os/drivers/compression/compress.c
Sambhavi-1: Changes done

apps/examples/testcase/le_tc/compression/tc_compress_read.c
```diff
@@ -305,6 +305,51 @@ static void tc_comp_name(void)
...
+ ****************************************************************************/
+static void tc_comp_file_decomp(void)
+{
+	int ret;
+	fd = get_compfd();
+
+	/*If FAIL : Failed to open testcase driver*/
+	TC_ASSERT_GEQ("get_compfd", fd, OK);
+
+	const char *file_path = "/res/dummy_comp.txt";
```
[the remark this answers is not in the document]
Sambhavi-1: Done .Also added the comment

apps/examples/testcase/le_tc/compression/tc_compress_read.c
```diff
@@ -305,6 +305,51 @@ static void tc_comp_name(void)
...
+
+	// initialising decompression framework in kernel
+	ret = ioctl(fd, COMPIOC_FDCMP_INIT, file_path);
+	TC_ASSERT_GEQ("compress_init", ret, OK);
```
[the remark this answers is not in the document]
Sambhavi-1: Okay

apps/examples/testcase/le_tc/compression/tc_compress_read.c
```diff
@@ -305,6 +305,51 @@ static void tc_comp_name(void)
...
+	ret = ioctl(fd, COMPIOC_FDCMP_GET_BUFSIZE, NULL);
+	TC_ASSERT_GEQ("get_compression_header", ret, OK);
+	uint8_t *dst_buffer = (uint8_t *)malloc(ret);
+
+	// Test decompression method
+	ret = ioctl(fd, COMPIOC_FDCMP_DECOMPRESS, dst_buffer);
+	if(ret < 0) {
+		free(dst_buffer);
+	}
+	TC_ASSERT_GEQ("compress_read", ret, OK);
```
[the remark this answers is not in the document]
Sambhavi-1: Done

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,52 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+	case COMPIOC_FDCMP_INIT:
+		if((char *)arg == NULL) {
+			return -EINVAL;
+		}
+		data = (struct file_decomp_data_s *)malloc(sizeof(struct file_decomp_data_s));
+		data->fd = open((char *)arg, "r");
+		filep->f_priv = data;
+		if (data->fd < 0) {
+			berr("Failed to open file = %s\n", (char *)arg);
+			return -EINVAL;
```
kishore-sn, line 150: If there is a failure in opening file, then we need to free all resources that were allocated before returning error. This is because, if init itself returns error, then user will not call deinit.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,52 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		ret = compress_init(data->fd, 0, &filelen);
```
kishore-sn, line 152: We need to check the value of ret. 
If it is ok, then we need to return OK to user. 
Else, we again need to free the resources and return an error code which can be understood by user.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,52 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		data = filep->f_priv;
+		data->compression_header = get_compression_header();
+		ret = data->compression_header->binary_size;
+		break;
+	case COMPIOC_FDCMP_DECOMPRESS:
+		if((uint8_t *)arg == NULL) {
+			return -EINVAL;
+		}
+		data = filep->f_priv;
+		size = compress_read(data->fd, 0, (uint8_t *)arg, data->compression_header->binary_size, 0);
```
kishore-sn, line 164: Need to check if compress read can return an error. If so, we cannot assume that it is the size. 
We will need to handle the error code and free the resources.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,74 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+	case COMPIOC_FDCMP_INIT:
+		if((char *)arg == NULL) {
+			return -EINVAL;
+		}
+		data = (struct file_decomp_data_s *)malloc(sizeof(struct file_decomp_data_s));
+		data->fd = open((char *)arg, "r");
+		filep->f_priv = data;
+		if (data->fd < 0) {
+			berr("Failed to open file = %s\n", (char *)arg);
+			free(data->compression_header);
```
kishore-sn, line 150: We are not allocating the memory for compression_header. It is allocated and freed inside the compression code. So, we must never free this. Please fix this in all the places below also.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,74 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			free(data);
+			filep->f_priv = NULL;
+			return -EINVAL;
+		}
+		ret = compress_init(data->fd, 0, &filelen);
+		if(ret != OK) {
+			compress_uninit();
+			free(data->compression_header);
+			free(data);
+			filep->f_priv = NULL;
```
kishore-sn, line 160: Let us set `ret = -ERROR` explicitly here.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,74 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		break;
+	case COMPIOC_FDCMP_GET_BUFSIZE:
+		data = filep->f_priv;
+		data->compression_header = get_compression_header();
+		ret = data->compression_header->binary_size;
+		if(ret < 0) {
```
kishore-sn, line 167: Let us change this to `if (ret <= 0) {`

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		break;
+	case COMPIOC_FDCMP_GET_BUFSIZE:
+		data = filep->f_priv;
+		data->compression_header = get_compression_header();
+		ret = data->compression_header->binary_size;
+		if(ret <= 0) {
+			compress_uninit();
+			free(data);
+			filep->f_priv = NULL;
```
kishore-sn, line 169: Let us print the value of ret and then set ret = -ERROR.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			return -EINVAL;
+		}
+		data = filep->f_priv;
+		size = compress_read(data->fd, 0, (uint8_t *)arg, data->compression_header->binary_size, 0);
+		ret = OK;
+		if(size != data->compression_header->binary_size) {
+			compress_uninit();
+			free(data);
+			filep->f_priv = NULL;
+			ret = -1;
```
kishore-sn, line 183: `-1` -> `-ERROR`

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		break;
+	case COMPIOC_FDCMP_DEINIT:
+		compress_uninit();	
+		data = filep->f_priv;
+		free(data);
```
kishore-sn, line 189: We can directly do `free(filep->f_priv);`

kishore-sn (approved):

kripal-samsung (approved):

sunghan-chang:
@vibhor-m Could you check this PR for your usages?

sunghan-chang:
@Sambhavi-1 The title `os/drivers/compression: Implementation for file decompression` looks too normal(?). I think `drivers/compression: add new ioctl to decompress data directly from a file` is better to know what the PR is.

Same as 2nd commit. The title says too common thing in compression. I suggest `add new tcs for decompression data directly from a file`

vibhor-m:
> @vibhor-m Could you check this PR for your usages?

I found the 4 IOCTLs are fine and sufficient to use a compressed file.

sunghan-chang (approved):

os/drivers/compression/compress.c
```diff
@@ -87,14 +95,17 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
 	struct compress_header *comp_info = (struct compress_header *)arg;
+	struct file_decomp_data_s *data;
 
-	if (comp_info == NULL && cmd != COMPIOC_GET_COMP_TYPE) {
-		return -EINVAL;
-	}
 	/* Handle built-in ioctl commands */
 	switch (cmd) {
 	case COMPIOC_COMPRESS:
+		if(comp_info == NULL) {
```
sunghan-chang, line 106: if (

add a space

os/drivers/compression/compress.c
```diff
@@ -105,6 +116,9 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		ret = CONFIG_COMPRESSION_TYPE;
 		break;
 	case COMPIOC_GET_COMP_NAME:
+		if((char *)arg == NULL) {
```
sunghan-chang, line 119: same as previous

os/drivers/compression/compress.c
```diff
@@ -105,6 +116,9 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			return -EINVAL;
+		}
 		switch (CONFIG_COMPRESSION_TYPE) {
 			case LZMA_TYPE:
```
sunghan-chang, line 123: regardless of this PR, here is a violation.

switch
case

should be the same indentation

Let's modify this with new commit.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
 		ret = decompress_block(comp_info->output_buffer, &comp_info->output_size, comp_info->input_buffer, &comp_info->input_size);
 		if (ret == ENOMEM) {
 			bcmpdbg("Output buffer allocated is not sufficient\n");
 		}
+#ifdef CONFIG_COMPRESSED_BINARY
+	case COMPIOC_FDCMP_INIT:
+		if((char *)arg == NULL) {
+			return -EINVAL;
+		}
+		data = (struct file_decomp_data_s *)malloc(sizeof(struct file_decomp_data_s));
```
sunghan-chang, line 145: Need to check data null here

sunghan-chang: malloc is right? or kmalloc is right?

apps/examples/testcase/le_tc/compression/tc_compress_read.c
```diff
@@ -305,6 +305,46 @@ static void tc_comp_name(void)
...
+	// to get buffer size require to hold decompressed data
+	ret = ioctl(fd, COMPIOC_FDCMP_GET_BUFSIZE, NULL);
+	TC_ASSERT_GT("get_compression_header", ret, OK);
+	uint8_t *dst_buffer = (uint8_t *)malloc(ret);
+
+	// Test decompression method
+	ret = ioctl(fd, COMPIOC_FDCMP_DECOMPRESS, dst_buffer);
+	TC_ASSERT_EQ_CLEANUP("compress_read", ret, OK, free(dst_buffer));
+
+	ret = ioctl(fd, COMPIOC_FDCMP_DEINIT, NULL);
```
sunghan-chang, line 341: This looks not thread safe. Is it?

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
 		break;
  	case COMPIOC_DECOMPRESS:
+		if(comp_info == NULL) {
+			return -EINVAL;
+		}
 		ret = decompress_block(comp_info->output_buffer, &comp_info->output_size, comp_info->input_buffer, &comp_info->input_size);
 		if (ret == ENOMEM) {
 			bcmpdbg("Output buffer allocated is not sufficient\n");
 		}
+#ifdef CONFIG_COMPRESSED_BINARY
```
sunghan-chang, line 140: Could you let me know why this is in the config?

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+	case COMPIOC_FDCMP_INIT:
+		if((char *)arg == NULL) {
+			return -EINVAL;
+		}
+		data = (struct file_decomp_data_s *)malloc(sizeof(struct file_decomp_data_s));
+		data->fd = open((char *)arg, "r");
+		filep->f_priv = data;
+		if (data->fd < 0) {
```
sunghan-chang, line 148: line 147 and 148 should be exchanged. Then, no need to call line 151 at failure.

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			bcmpdbg("Failed to open file = %s\n", (char *)arg);
+			free(data);
+			filep->f_priv = NULL;
+			return -EINVAL;
+		}
+		ret = compress_init(data->fd, 0, &filelen);
+		if(ret != OK) {
+			free(data);
+			filep->f_priv = NULL;
```
sunghan-chang, line 157: should close the opened fd as well

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+			ret = ERROR;
+		}
+		break;
+	case COMPIOC_FDCMP_GET_BUFSIZE:
+		data = filep->f_priv;
```
sunghan-chang, line 162: No need to verify the filep?
same as other lines which use the filep

os/drivers/compression/compress.c
```diff
@@ -116,10 +130,70 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		data->compression_header = get_compression_header();
+		ret = data->compression_header->binary_size;
+		if(ret <= 0) {
+			bcmpdbg("Failed to get buffer size = %d\n", ret);
+			compress_uninit();
+			free(data);
+			filep->f_priv = NULL;
```
sunghan-chang, line 169: 1. I can't find closing opened fd at anywhere.
2. Is it right to free all resources here even it gets failure at some operation? not in deinit?
3. If we get failure at get buffer or decompress, what is our guidance in ioctl sequence next?

apps/examples/testcase/le_tc/compression/tc_compress_read.c
kishore-sn: Hi Chang.. 
We are storing all the data related compression in fd->fpriv.
So, as long as each thread opens the driver and makes ioctl calls, then it is safe. 
But if a open driver fd is shared accross threads, then it is not safe. It is application's responsibility to prevent this.

os/drivers/compression/compress.c
kishore-sn: This config was included because all compress_xxx apis used below are under this config. 
But I think this is not correct way. 
@Sambhavi-1 
Please change this. Instead of using CONFIG_COMPRESSED_BINARY config here, let us remove this config from all the compress_xxx apis. Instead move all the compress_xxx apis to CONFIG_COMPRESSION. This is because, previously, these apis were used only for binary compression. But, now they will be used for file compression also.

os/drivers/compression/compress.c
kishore-sn: @Sambhavi-1 Please use kmalloc for allocating any data which will ONLY be used inside kernel. In fact its better to use kzalloc so that the data is set to zero.

os/drivers/compression/compress.c
kishore-sn: @Sambhavi-1 We need to close fd in all failure cases. And also in deinit case.

os/drivers/compression/compress.c
kishore-sn: Hi @Sambhavi-1 and @sunghan-chang ,
The call flow for success case will be like below:
init -> getsize -> decompress -> deinit

But, it there is failure at any of the intermediate steps, then our previous plan was that app will not make any further calls. So, in each case, when there is a failure, we were trying to release all resources. 
But, I think this is not the best approach. I think below method needs to be followed by app. And we need to modify our code accordingly. 

APP CODE
```
while (retry) {
    if (init != OK) goto error;
    if (get size != OK) goto error;
    if (decompress != OK) goto error:
    else deinit, return
error:
    deinit
}
```

sunghan-chang: @Sambhavi-1 @kishore-sn We should guide how to call new APIs in success and in failure.

neel-samsung (commented):
some checks are failing, please check.

os/drivers/compression/compress.c
```diff
@@ -120,12 +120,12 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 			return -EINVAL;
 		}
 		switch (CONFIG_COMPRESSION_TYPE) {
-			case LZMA_TYPE:
```
kishore-sn, line 109: Please add folder path in commit title. 
Change the commit title to : Fix indentation issues

os/drivers/compression/compress.c
kishore-sn: @Sambhavi-1 
Please add below to first commit. 

```
App shall use the ioctls as follows to perform the decompression.

while (retry) {
    if (init != OK) goto error;
    if (get size != OK) goto error;
    if (decompress != OK) goto error:
    else deinit, return
error:
    deinit
}

If each of the above steps is successful, the decompression will be done and the data will
be stored in the buffer provided by the app at decompress ioctl. However, if there is an error
at any step, then app needs to call deinit ioctl. 
'''

**Review by samsung-singh — APPROVED** [review]
decompressing the compressed file in os drivers compression module this change is good to merge.

**Comment by sunghan-chang** [comment]
drivers/compress: correct switch case indentation violation

**Comment by sunghan-chang** [comment]
Let's remove `to` at the title. I mean
apps/examples/testcase: To add new  -> apps/examples/testcase: add new

File: os/drivers/compression/compress.c (hunk 28 above)
Comment (sunghan-chang) [reply os/drivers/compression/compress.c]: That should be written in the code as well.
When we use new APIs, we don't check the commit description. We just check API description usually.

File: os/drivers/compression/compress.c [hunk 30]
```diff
@@ -105,21 +116,86 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		data = (struct file_decomp_data_s *)kmm_zalloc(sizeof(struct file_decomp_data_s));
+		if (!data) {
+			bcmpdbg("Memory allocation fail for data storing\n");
+			return -ENOMEM;
+		}
+		data->fd = open((char *)arg, "r");
+		if (data->fd < 0) {
+			bcmpdbg("Failed to open file = %s\n", (char *)arg);
+			return -EINVAL;
```
sunghan-chang, line 152: data is a local variable. If you don't free it here, it will be a leak.

os/drivers/compression/compress.c
Sambhavi-1: Done

os/drivers/compression/compress.c
Sambhavi-1: okay..got it

os/drivers/compression/compress.c
```diff
@@ -105,21 +116,86 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		DEBUGASSERT(filep);
+		filep->f_priv = data;
```
sunghan-chang, line 155: What does the `filep` have? It's the compression path `/dev/compress`. No?
This is the point that I told not thread safe. 
If two different threads want decompress two different files, Is this working?
@kishore-sn @Sambhavi-1

os/include/tinyara/fs/ioctl.h
```diff
@@ -474,6 +474,10 @@
 #define COMPIOC_GET_COMP_TYPE      _COMPIOC(0x0002)
 #define COMPIOC_GET_COMP_NAME      _COMPIOC(0x0003)
 #define COMPIOC_DECOMPRESS	   _COMPIOC(0x0004)
+#define COMPIOC_FDCMP_INIT         _COMPIOC(0x0005)
+#define COMPIOC_FDCMP_GET_BUFSIZE  _COMPIOC(0x0006)
+#define COMPIOC_FDCMP_DECOMPRESS   _COMPIOC(0x0007)
+#define COMPIOC_FDCMP_DEINIT       _COMPIOC(0x0008)
```
sunghan-chang, line 480: It looks strange in name `FDCMP`.
It's for file, not fd. Right?
And line 474~476 use `COMP`, not `CMP`. Let's use the same word.

os/drivers/compression/compress.c
```diff
@@ -105,21 +116,86 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+		}
+		DEBUGASSERT(filep);
+		data = filep->f_priv;
+		size = compress_read(data->fd, 0, (uint8_t *)arg, data->compression_header->binary_size, 0);
+		ret = OK;
+		if (size != data->compression_header->binary_size) {
+			ret = ERROR;
+		}
+		break;
+	case COMPIOC_FDCMP_DEINIT:
```
sunghan-chang, line 183: Please add some code comment on each new ioctl commands to describe what they do.

os/drivers/compression/compress.c
Sambhavi-1: Done

os/drivers/compression/compress.c
Sambhavi-1: done

os/drivers/compression/compress.c
Sambhavi-1: okay

os/drivers/compression/compress.c
Sambhavi-1: usage instruction added both in commit and code

os/drivers/compression/compress.c
Sambhavi-1: Done

os/include/tinyara/fs/ioctl.h
Sambhavi-1: Now I have changed the name to FCOMP like COMPIOC_FCOMP_INIT. Is it ok?

os/drivers/compression/compress.c
```diff
@@ -72,6 +80,26 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
  * 	 COMPIOC_DECOMPRESS - for decompression of input data
  * 	 COMPIOC_GET_COMP_TYPE - for getting compression type
  * 	 COMPIOC_GET_COMP_NAME - for getting compression type in string
+ *   COMPIOC_FCOMP_INIT - for init the decompression framework in kernel
+ *   COMPIOC_FCOMP_GET_BUFSIZE - for getting the buffer size require to hold decompressed data
+ *   COMPIOC_FCOMP_DECOMPRESS - for getting decompressed data in buffer
+ *   COMPIOC_FCOMP_DEINIT - for cleaning the kernel decompression framework
+ * 
+ *   App shall use the ioctls as follows to perform the decompression.
```
kishore-sn, line 88: Add this line in the commit title also. ` App shall use the ioctls as follows to perform the decompression.`

kishore-sn (approved):

os/drivers/compression/compress.c
```diff
@@ -105,21 +136,86 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		ret = CONFIG_COMPRESSION_TYPE;
 		break;
 	case COMPIOC_GET_COMP_NAME:
+		if ((char *)arg == NULL) {
```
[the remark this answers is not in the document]
sunghan-chang: Could you let me know the reason why you use it differently for the same argument?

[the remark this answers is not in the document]
kishore-sn: Hi Chang, 
In case of ioctl, the arg parameter can be used to pass any type of argument. Sometimes we use it to pass a string. At other times we use it to pass pointer to some data structure. This is normal behavior for ioctl.

sunghan-chang (approved):

sunghan-chang:
@kishore-sn Let's merge.

PR #6797, opened 2025-04-23
driver/input/ist415: Patch semaphore issue.

IMJuBBang:
- Fixed an issue where sem_wait and sem_post did not match depending on the conditions.

### DIFF
--- a/os/drivers/input/ist415.c
+++ b/os/drivers/input/ist415.c
@@ -765,6 +765,7 @@ static void touch_interrupt(struct ist415_config_s *config)
 static int ist415_event_thread(int argc, char **argv)
 {
 	struct ist415_dev_s *dev;
+	bool check = false;
 
 	DEBUGASSERT(argc == 2);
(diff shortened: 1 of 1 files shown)


os/drivers/input/ist415.c
```diff
@@ -795,7 +795,10 @@ static int ist415_event_thread(int argc, char **argv)
 
 		dev->irq_working = false;
 		dev->lower->ops->irq_enable(dev->lower);
-		sem_post(&dev->sem);
+
+		if ((dev->ready == true) && (dev->calib == false)) {
```
ewoodev, line 799: the ready can be changed during ist415_process_event

How about make below

bool check = (dev->ready == true) && (dev->calib == false)

if (check) {
   sem_wait
}

...

if (check) {
    sem_post
}

IMJuBBang: That's a good idea, too.

ewoodev (approved):

sunghan-chang (approved):

PR #6798, opened 2025-04-23
arch/amebasmart, board/rtl8730e: Fix i2c sem corruption issue

ewoodev:
since we are calling up_i2cuninitialize and up_i2cinitialize during ist415 TSC power on/off sequence for recovery I2C operation.

but the up_i2cuninitialize has sem_destroy
so the sem is destoryed, but a crash is occurring by accessing the sem from another thread.

we just need i2c operation recovery, therefore replace up_i2cuninitialze to amebasmart_i2c_deinit

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_ist415.c
+++ b/os/board/rtl8730e/src/rtl8730e_ist415.c
@@ -124,16 +124,18 @@ static void rtl8730e_ist415_disable_irq(struct ist415_config_s *dev)
 static void rtl8730e_ist415_power_off(struct ist415_config_s *dev)
 {
 	struct rtl8730e_ist415_s *priv = (struct rtl8730e_ist415_s *)dev->priv;
-	up_i2cuninitialize(priv->i2c);	/* Workaround: IC20 write hang issue */
+	struct amebasmart_i2c_priv_s *priv_i2c = (struct amebasmart_i2c_priv_s *)priv->i2c;
+	amebasmart_i2c_deinit(priv_i2c);
 	GPIO_WriteBit(IST415_GPIO_RESET_PIN, PIN_LOW);
 }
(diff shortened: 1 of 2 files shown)


os/board/rtl8730e/src/rtl8730e_ist415.c
```diff
@@ -133,7 +133,7 @@ static void rtl8730e_ist415_power_on(struct ist415_config_s *dev)
 	struct rtl8730e_ist415_s *priv = (struct rtl8730e_ist415_s *)dev->priv;
 	GPIO_WriteBit(IST415_GPIO_RESET_PIN, PIN_HIGH);
 	DelayMs(1);  /* Wait for stable voltage before i2c commands issued */
-	priv->i2c = up_i2cinitialize(IST415_I2C_PORT);	/* Workaround: IC20 write hang issue */
+	amebasmart_i2c_init(priv->i2c);
```
zhongnuo-tang, line 136: amebasmart_i2c_init takes in amebasmart_i2c_priv_s, but here we passed down i2c_dev_s, there is not casting in amebasmart_i2c_init

ewoodev: Hello @zhongnuo-tang, the code is in BSP
sorry, I just raise this changes for solve quickly this issue.
I think, your changing is better about this issue . 

Could you guide kindly how to make better this? or you can raise your change.

zhongnuo-tang: how about changing to below
```	
struct amebasmart_i2c_priv_s *priv_i2c = (struct amebasmart_i2c_priv_s *)priv->i2c;
amebasmart_i2c_init(priv_i2c);
```

PR #6799, opened 2025-04-23
kernel/clock: Add clock_getinittime API

ewoodev:
The clock_getinittime returns system initalized time

### DIFF
--- a/os/include/tinyara/clock.h
+++ b/os/include/tinyara/clock.h
@@ -341,6 +341,22 @@ int clock_systimespec(FAR struct timespec *ts);
  */
 
 /****************************************************************************
+ * Name: clock_getinittime
+ *
+ * Description:
+ *   This function returns initalized system time.
+ *
+ * Parameters:
+ *   tp - Location to return the time
+ *
+ * Return Value:
+ *   OK (0) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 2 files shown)


os/include/tinyara/clock.h
```diff
@@ -340,6 +340,23 @@ int clock_systimespec(FAR struct timespec *ts);
  * @endcond
  */
 
+/****************************************************************************
+ * Name: initialize_system_time
```
sunghan-chang, line 344: Need to check

os/include/tinyara/clock.h
```diff
@@ -340,6 +340,23 @@ int clock_systimespec(FAR struct timespec *ts);
...
+ *   This function returns initalized system time.
+ *
+ * Parameters:
+ *   tp - Location to return the time
+ *
+ * Return Value:
+ *   OK (0) on success; a negated errno value on failure.
+ *
+ ****************************************************************************/
+#ifdef CONFIG_INIT_SYSTEM_TIME
```
sunghan-chang, line 356: This is valid with this? I think it should not.

ewoodev: I got it, updated, Could you please reivew again?

os/include/tinyara/clock.h
ewoodev: Chagned

sunghan-chang (approved):

PR #6802, opened 2025-04-24
os/include: Remove conditional directives use from all enums.

kripal-samsung:
Remove conditional directives use from all enums in os/include

### DIFF
--- a/apps/system/utils/utils_ps.c
+++ b/apps/system/utils/utils_ps.c
@@ -77,16 +77,10 @@ static const char *utils_statenames[] = {
 	"INACTIVE",
 	"WAITSEM ",
 	"WAITUNBLOCK",
-#ifndef CONFIG_DISABLE_SIGNALS
 	"WAITSIG ",
-#endif
-#ifndef CONFIG_DISABLE_MQUEUE
 	"MQNEMPTY",
 	"MQNFULL ",
-#endif
-#ifdef CONFIG_PAGING
 	"WAITPAGEFILL "
-#endif
 };
 
 static const char *utils_ttypenames[4] = {
--- a/os/fs/procfs/fs_procfsproc.c
+++ b/os/fs/procfs/fs_procfsproc.c
@@ -313,13 +313,9 @@ static const char *g_statenames[] = {
 	"Inactive",
 	"Semaphore wait",
 	"To be unblocked wait"
-#ifndef CONFIG_DISABLE_MQUEUE
 	"Signal wait",
-#endif
-#ifndef CONFIG_DISABLE_MQUEUE
 	"MQ not empty wait",
 	"MQ no full wait"
-#endif
 };
 
 static const char *g_ttypenames[4] = {
--- a/os/include/tinyara/sched.h
+++ b/os/include/tinyara/sched.h
@@ -225,16 +225,10 @@ enum tstate_e {
 	TSTATE_TASK_INACTIVE,		/* BLOCKED      - Initialized but not yet activated */
 	TSTATE_WAIT_SEM,			/* BLOCKED      - Waiting for a semaphore */
 	TSTATE_WAIT_FIN,		/* BLOCKED		- Waiting to be unblocked by fin */
-#ifndef CONFIG_DISABLE_SIGNALS
 	TSTATE_WAIT_SIG,			/* BLOCKED      - Waiting for a signal */
-#endif
-#ifndef CONFIG_DISABLE_MQUEUE
 	TSTATE_WAIT_MQNOTEMPTY,		/* BLOCKED      - Waiting for a MQ to become not empty. */
 	TSTATE_WAIT_MQNOTFULL,		/* BLOCKED      - Waiting for a MQ to become not full. */
-#endif
-#ifdef CONFIG_PAGING
 	TSTATE_WAIT_PAGEFILL,		/* BLOCKED      - Waiting for page fill */
-#endif
 	NUM_TASK_STATES				/* Must be last */
 };
 typedef enum tstate_e tstate_t;
--- a/os/kernel/init/os_start.c
+++ b/os/kernel/init/os_start.c
@@ -172,7 +172,9 @@ volatile dq_queue_t g_readytorun;
  * always the CPU's IDLE task.
  */
 
+#ifdef CONFIG_SMP
 volatile dq_queue_t g_assignedtasks[CONFIG_SMP_NCPUS];
+#endif
 
 /* g_running_tasks[] holds a references to the running task for each cpu.
  * It is valid only when up_interrupt_context() returns true.
@@ -317,12 +319,18 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
 		&g_waitingforfin,
 		TLIST_ATTR_PRIORITIZED
 	},
+#ifndef CONFIG_DISABLE_SIGNALS
 	{                                              /* TSTATE_WAIT_SIG */
 		&g_waitingforsignal,
 		0
-	}
+	},
+#else
+	{                                             
+		NULL,
+		0
+	},
+#endif
 #ifndef CONFIG_DISABLE_MQUEUE
-	,
 	{                                              /* TSTATE_WAIT_MQNOTEMPTY */
 		&g_waitingformqnotempty,
 		TLIST_ATTR_PRIORITIZED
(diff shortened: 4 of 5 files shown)


os/fs/procfs/fs_procfsproc.c
```diff
@@ -307,13 +307,9 @@ static const char *g_statenames[] = {
 	"Inactive",
 	"Semaphore wait",
 	"To be unblocked wait"
-#ifndef CONFIG_DISABLE_MQUEUE
```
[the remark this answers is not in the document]
kripal-samsung: Ok sure

os/include/tinyara/sched.h
```diff
@@ -643,9 +635,7 @@ struct tcb_s {
 
 	/* POSIX Named Message Queue Fields ****************************************** */
 
-#ifndef CONFIG_DISABLE_MQUEUE
```
neel-samsung, line 646: Do we need to remove this conditional also? I don't think, we should. Please check it.

os/kernel/init/os_start.c
```diff
@@ -320,25 +312,20 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
...
-#endif
-#ifdef CONFIG_PAGING
-	,
+	},
 	{                                              /* TSTATE_WAIT_PAGEFILL */
 		&g_waitingforfill,
 		TLIST_ATTR_PRIORITIZED
 	}
-#endif
+
```
neel-samsung, line 328: Please remove extra space.

os/kernel/init/os_start.c
```diff
@@ -194,31 +194,23 @@ volatile dq_queue_t g_waitingforsemaphore;
...
 
-#ifndef CONFIG_DISABLE_MQUEUE
 volatile dq_queue_t g_waitingformqnotfull;
-#endif
 
 /* This is the list of all tasks that are blocking waiting for a page fill */
 
-#ifdef CONFIG_PAGING
 volatile dq_queue_t g_waitingforfill;
-#endif
```
kishore-sn, line 221: Let us not modfiy the conditions for any variables or actual code. Only for enums declarations, we will remove the conditionals.

apps/system/utils/utils_ps.c
```diff
@@ -77,16 +77,10 @@ static const char *utils_statenames[] = {
 	"INACTIVE",
 	"WAITSEM ",
 	"WAITUNBLOCK",
-#ifndef CONFIG_DISABLE_SIGNALS
```
kishore-sn, line 80: Please make a separate commit for this file. 
All files in os and lib folders can be made into single commit. 
But for any other folders, we need to make separate commit.

kishore-sn (approved):

Sambhavi-1 (approved):
Conditional directive are removed from all enums looks good

os/include/tinyara/sched.h
```diff
@@ -225,16 +225,10 @@ enum tstate_e {
 	TSTATE_TASK_INACTIVE,		/* BLOCKED      - Initialized but not yet activated */
 	TSTATE_WAIT_SEM,			/* BLOCKED      - Waiting for a semaphore */
 	TSTATE_WAIT_FIN,		/* BLOCKED		- Waiting to be unblocked by fin */
-#ifndef CONFIG_DISABLE_SIGNALS
```
sunghan-chang, line 228: @kripal-samsung @kishore-sn  I saw that you modified the `g_tasklisttable` before and removed it now. Is it ok not to modity?
```
	{                                              /* TSTATE_WAIT_SIG */
		&g_waitingforsignal,
		0
	}
#ifndef CONFIG_DISABLE_MQUEUE
	,
	{                                              /* TSTATE_WAIT_MQNOTEMPTY */
		&g_waitingformqnotempty,
		TLIST_ATTR_PRIORITIZED
	},
	{                                              /* TSTATE_WAIT_MQNOTFULL */
		&g_waitingformqnotfull,
		TLIST_ATTR_PRIORITIZED
	}
#endif
#ifdef CONFIG_PAGING
	,
	{                                              /* TSTATE_WAIT_PAGEFILL */
		&g_waitingforfill,
		TLIST_ATTR_PRIORITIZED
	}
#endif
#ifdef CONFIG_SIG_SIGSTOP_ACTION
	,
	{                                              /* TSTATE_TASK_STOPPED */
		&g_stoppedtasks,
		0                                            /* See tcb->prev_state */
	},
#endif
};
```
1. If we disable `CONFIG_DISABLE_MQUEUE` and access the list with `TSTATE_WAIT_PAGEFILL`, what happen? Invalid access (out of the list)?
2. Need to check what `g_stoppedtasks` is. This enum does not have matched value.

kishore-sn: hi Chang..

neel-samsung (approved):

samsung-singh (approved):
Removed conditional directives use from all enums in os include this change is good to merge.

os/include/tinyara/sched.h
kishore-sn: @kripal-samsung 
Let us remove the configs from enum in sched.h and also from the lists array in os_start.
Then all the lists will always be defined. Based on whether the config is defined or not, the tasks will get added to the lists. 

Also, pls remove g_stoppedtasks as we dont have corresponding enum for it.

kishore-sn (dismissed):

os/kernel/init/os_start.c
```diff
@@ -194,31 +194,23 @@ volatile dq_queue_t g_waitingforsemaphore;
 
 /* This is the list of all tasks that are blocked waiting for a signal */
 
-#ifndef CONFIG_DISABLE_SIGNALS
 volatile dq_queue_t g_waitingforsignal;
```
sunghan-chang, line 197: To support enabled CONFIG_DISABLE_SIGNALS, you define the queue always regardless of the config.
But when the config is enabled, this will not be used.

When the config is enabled, if we give null at the g_tasklisttable, what happen?
I mean "is there any problem on usages and memory?"

kishore-sn: Since we are deleting the config check in tstate enum, if we add null in this array, then it will cause problem.

r-prabu (approved):

os/kernel/init/os_start.c
sunghan-chang: When CONFIG_DISABLE_SIGNALS is enabled and enum has no config checking, scheduler will not access the queue. Because the config is enabled (the functionality is disabled).
But if we don't match the number of enum and queue list, normal operation can have mismatched.
To resolve it, we have two options.
1. this PR - just enable all lists
2. my suggestion - keep config checking at the queue list, but add null if the config is enabled.

I told you 2nd option. I think the scheduler will not access the list (null) when the config is enabled (function is disabled). It means no issue with 2nd option. Could you correct me something wrong in my opinion?

os/kernel/init/os_start.c
```diff
@@ -330,21 +330,25 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
 	{                                              /* TSTATE_WAIT_MQNOTFULL */
 		&g_waitingformqnotfull,
 		TLIST_ATTR_PRIORITIZED
+	},
+#else
+	{                                             
+		NULL,
+		0
```
sunghan-chang, line 337: `#ifndef` has two lists, but `#else` has just one list. One more list should added with null value.

os/kernel/init/os_start.c
```diff
@@ -330,21 +330,25 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
 	{                                              /* TSTATE_WAIT_MQNOTFULL */
 		&g_waitingformqnotfull,
 		TLIST_ATTR_PRIORITIZED
+	},
```
sunghan-chang, line 333: line 321 should be applied like this.
Because signal also has the config to disable
```
#ifndef CONFIG_DISABLE_SIGNALS
volatile dq_queue_t g_waitingforsignal;
#endif
```

sunghan-chang: line 175, g_assignedtasks should be in CONFIG_SMP

os/kernel/init/os_start.c
```diff
@@ -317,33 +319,46 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
...
 	{                                              /* TSTATE_WAIT_PAGEFILL */
 		&g_waitingforfill,
 		TLIST_ATTR_PRIORITIZED
-	}
-#endif
-#ifdef CONFIG_SIG_SIGSTOP_ACTION
-	,
-	{                                              /* TSTATE_TASK_STOPPED */
-		&g_stoppedtasks,
-		0                                            /* See tcb->prev_state */
```
sunghan-chang, line 346: This is removed. Is this intended?

os/kernel/init/os_start.c
```diff
@@ -317,33 +319,46 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
...
+	},
+#else
+	{                                             
+		NULL,
+		0
+	},
+	{                                             
+		NULL,
+		0
 	}
```
sunghan-chang, line 350: Here is no comma even line 341 has.

os/kernel/init/os_start.c
```diff
@@ -317,33 +319,46 @@ const struct tasklist_s g_tasklisttable[NUM_TASK_STATES] =
...
-	}
-#endif
-#ifdef CONFIG_SIG_SIGSTOP_ACTION
-	,
-	{                                              /* TSTATE_TASK_STOPPED */
-		&g_stoppedtasks,
-		0                                            /* See tcb->prev_state */
+	},
+#else
+	{
```
sunghan-chang, line 359: Even line 353 has a comma, here does not.

Let's use the same format.

os/kernel/init/os_start.c
kishore-sn: We dont have an enum or list for it. So, we removed it.

kishore-sn (approved):

r-prabu (approved):

PR #6803, opened 2025-04-24
Svace fix

r-prabu:
Fixed issues reported by svace tool
1. kernel/binary_manager: 
-  make sure to increment the load_cnt on binary_manager_load success. 
-  proper error message on the debug print when binary_manager_update_bpinfo fails.

2. kernel/debug: 
- Added NULL pointer check in the heap_check to validate tcb.

### DIFF
--- a/os/kernel/binary_manager/binary_manager_load.c
+++ b/os/kernel/binary_manager/binary_manager_load.c
@@ -506,7 +506,7 @@ static int loadingall_thread(int argc, char *argv[])
 	for (bin_idx = 1; bin_idx <= bin_count; bin_idx++) {
 		if (BIN_LOAD_PRIORITY(bin_idx, BIN_USEIDX(bin_idx)) == BINARY_LOADPRIO_HIGH) {
 			ret = binary_manager_load(bin_idx);
-			if (ret > 0) {
+			if (ret == BINMGR_OK) {
 				load_cnt++;
 			}
 		}
--- a/os/kernel/binary_manager/binary_manager_resource.c
+++ b/os/kernel/binary_manager/binary_manager_resource.c
@@ -99,7 +99,8 @@ int binary_manager_mount_resource(void)
 
 #ifdef CONFIG_USE_BP
 	binmgr_bpdata_t *bp_data;
-	if (binary_manager_update_bpinfo() != BINMGR_OK) {
+	ret = binary_manager_update_bpinfo();
+	if (ret != BINMGR_OK) {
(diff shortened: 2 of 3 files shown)


os/kernel/binary_manager/binary_manager_load.c
```diff
@@ -506,7 +506,7 @@ static int loadingall_thread(int argc, char *argv[])
 	for (bin_idx = 1; bin_idx <= bin_count; bin_idx++) {
 		if (BIN_LOAD_PRIORITY(bin_idx, BIN_USEIDX(bin_idx)) == BINARY_LOADPRIO_HIGH) {
 			ret = binary_manager_load(bin_idx);
-			if (ret > 0) {
+			if (ret == 0) {
```
kishore-sn, line 509: How about checking BINMGR_OK or OK instead of zero?

os/kernel/sched/sched_free.c
```diff
@@ -132,8 +132,8 @@ void sched_ufree(FAR void *address)
 	} else {
 		/* No.. just deallocate the memory now. */
 
-		kumm_free(address);
 		kumm_givesemaphore(address);
+		kumm_free(address);
```
kishore-sn, line 136: I am not sure if this change is correct. If we give up the sem before free, then it might result in context switch. 
The logic of this api is such that we first check if the sem is available. If it is avaialable, we want to free the memory immediately without a context switch. If the sem is not avaialable, then we add the memory chunk to a list to be freed later.

kishore-sn: Also, can you please paste the svace error msg in the commit description?

sunghan-chang:
@kishore-sn When your reviews are resolved, let's merge.

sunghan-chang:
@r-prabu Any update?

os/kernel/sched/sched_free.c
r-prabu: @kishore-sn I am not sure if this is a good idea to free the memory & operate on that memory. If you feel context switching is the problem then can add  a critical section for this operation for these 2 API's?

r-prabu: reverted this change

kishore-sn (approved):

kripal-samsung (approved):

Sambhavi-1 (approved):
svace issue got fixed looks ok

samsung-singh (approved):
this PR fixes Svace and added some sanity check, good to merge.

PR #6804, opened 2025-04-25
os/board/rtl8730e: Modify wifi porting layer to prep for concurrent

jwei5:
- Certain parts of wifi porting code requires to be changed if concurrent mode is supported in the future
- Main point is to ensure wlan index on the application side is matched to the correct interface in wifi driver
- Use preprocessor CONFIG_RTK_CONCURRENT_MODE to set the right code flow based on whether concurrent mode is supported

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_enet.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_enet.c
@@ -68,8 +68,6 @@
 #include "rtw_wifi_constants.h"
 #include "wifi_intf_drv_to_app_basic.h"
 
-//for concurrent mode, not supported now
-//#define RTK_CONCURRENT_MODE
 #ifdef CONFIG_AMEBASMART_WIFI
 
 struct netdev *ameba_nm_dev_wlan0 = NULL;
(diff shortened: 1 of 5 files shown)


hk-gwak:
Hi @jwei5 , I have a question. No need the changes in
‎os/board/rtl8730e/src/component/os/tizenrt/rtk_wifi_utils.h
‎os/board/rtl8730e/src/component/wifi/wifi_interactive_mode.c
these files, like in other PR?

jwei5:
Hi @hk-gwak , the main changes for this PR is to match the wlan interface on TizenRT to the correct interface in the driver during concurrent mode. These changes are necessary as long as concurrent mode is supported.
Yes you are right, some of the other files such as wifi_interactive_mode.c might need to be changed.
I didn't make the changes to these files in this PR because I saw that they were already changed in the previous bridge mode test branches 068686a4b04c8bbd8247ed142f6792578200f426

Will there be a new test branch for Samsung to verify bridge mode? My plan was to see how bridge mode is implemented on TizenRT (such as state handling) and review our RTK porting layer. If more changes are needed, we will raise another PR. Will this approach be ok?

hk-gwak:
We are going to apply most of changes in 068686a4b04c8bbd8247ed142f6792578200f426. But for applying most of the changes in 068686a4b04c8bbd8247ed142f6792578200f426, looks like TizenRT side change need to be applied first( ex) wifi.h). I think this is kind of thing that you mentioned(
`
My plan was to see how bridge mode is implemented on TizenRT (such as state handling) and review our RTK porting layer
`
).
So let me apply this PR first, then we will apply TizenRT side change(such as state handling), and will request another PR for RTK side like
`
	//LWNL_EVT_CONCURRENT_JOINED and LWNL_EVT_CONCURRENT_LEFT as placeholder
	case 6:
		trwifi_post_event(ameba_nm_dev_wlan1, LWNL_EVT_CONCURRENT_JOINED, NULL, 0);
		break;
	case 7:
		trwifi_post_event(ameba_nm_dev_wlan1, LWNL_EVT_CONCURRENT_LEFT, NULL, 0);
		break;
`
in rtk_netmgr.c

jwei5:
Understood, we will review our porting layer again after the changes for TizenRT side has been applied. Thanks 👍

hk-gwak (approved):
LGTM

PR #6806, opened 2025-04-25
os/kernel/task: Kernel thread must not inherit app's heap, etc..

kishore-sn:
If a kernel thread gets created from an app thread, then it must not inherit the app uheap, uspace and memory protection data.

### DIFF
--- a/os/arch/arm/src/common/up_restoretask.c
+++ b/os/arch/arm/src/common/up_restoretask.c
@@ -83,7 +83,7 @@ void up_restoretask(struct tcb_s *tcb)
 #endif
 
 #ifdef CONFIG_ARCH_USE_MMU
-		if (tcb->app_id && tcb->pgtbl != mmu_l1_pgtable()) {
+		if (tcb->app_id && tcb->pgtbl && tcb->pgtbl != mmu_l1_pgtable()) {
 			cp15_wrttb((uint32_t) tcb->pgtbl | TTBR0_RGN_WBWA | TTBR0_IRGN0);
 			cp15_invalidate_tlbs();
 		}
--- a/os/kernel/task/task_setup.c
+++ b/os/kernel/task/task_setup.c
@@ -497,28 +497,36 @@ static int thread_schedsetup(FAR struct tcb_s *tcb, int priority, start_t start,
 		up_initial_state(tcb);
 
 #ifdef CONFIG_APP_BINARY_SEPARATION
-		/* Copy the parent task ram details to this task */
+		/* Copy some attributes from parent task to this task */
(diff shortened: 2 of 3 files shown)


samsung-singh (approved):
this pr fixes the bugs, avoids inheriting user heap into kernel thread, this change is good to merge.

Sambhavi-1 (approved):
Fix for inheriting apps heap by kernel thread

os/kernel/task/task_setup.c
```diff
@@ -497,28 +497,36 @@ static int thread_schedsetup(FAR struct tcb_s *tcb, int priority, start_t start,
...
-		tcb->uheap = rtcb->uheap;
 		tcb->app_id = rtcb->app_id;
+
+		/* If a kernel thread is getting created due to some
+		 * call from a user task or pthread, then the kernel thread
+		 * will not inherit the some attributes related to heap, uspace,
+		 * memory protection, etc..
+		 */
+		if ((tcb->flags & TCB_FLAG_TTYPE_MASK) != TCB_FLAG_TTYPE_KERNEL) {
+			tcb->uspace = rtcb->uspace;
```
sunghan-chang, line 510: These members will be set when it's the user thread. Could you verify all members which are set in the if statement are used after null or 0 checking?
I mean, 
`uspace` is used when it's not null in all codes?
same in uheap
same in pgtbl
and so on

kishore-sn: Thanks Chang. I verified and added null checks where  they are required.

kripal-samsung (approved):

os/arch/arm/src/common/up_restoretask.c
```diff
@@ -83,7 +83,7 @@ void up_restoretask(struct tcb_s *tcb)
 #endif
 
 #ifdef CONFIG_ARCH_USE_MMU
-		if (tcb->app_id && tcb->pgtbl != mmu_l1_pgtable()) {
+		if (tcb->app_id && tcb->pgtbl && tcb->pgtbl != mmu_l1_pgtable()) {
```
sunghan-chang, line 86: It looks unnecessary. Previous covers this new addition as well in this case..

kishore-sn: No Chang. 
Previously we only checked `tcb->pgtbl != mmu_l1_pgtable()`.
This condition will be true even if pgtbl is NULL. But in NULL case, we must not execute the if statement. 
So, I think both checks are necessary.

sunghan-chang (approved):

sunghan-chang:
@r-prabu Let's merge this now.

PR #6808, opened 2025-05-02
os/kernel/pthread: Pthread mutex wait must not reutrn EINTR

kishore-sn:
According to posix standard, if a pthread is waiting to aquire mutex and gets interrupted by a signal, then it is supposed to execute the signal and then continue to wait for the mutex.

Previously, our code allowed to pass an argument which decided whether an EINTR error is retured or not. Since this is not posix compliant behavior, this commit removes this argument.

Kernel tc has be verified with this commit.


sunghan-chang (approved):

samsung-agnel (approved):

kripal-samsung (approved):

r-prabu (commented):
Pthread mutex wait changes to remove the EINTR return looks fine to take.

PR #6810, opened 2025-05-04
tools/trap: add missing "self" keyword to the variable in trap tool

neel-samsung:
This patch fixes issue of trap not working by adding the required self keyword to the variable in logparser.py


kishore-sn (approved):

sunghan-chang (approved):
This means the TRAP is not working now??

kripal-samsung (approved):

r-prabu (approved):
changes for the  "self" keyword to the variable in trap too is good to take

Sambhavi-1 (commented):
missing self keyword added in trap tool

PR #6812, opened 2025-05-06
os/fs/procfs: make ps cmd to show next task as running instead of tash

neel-samsung:
This patch shows the tash as waiting for semaphore and next ready to run tcb as running for that cpu.

Test result:

  PID | PRIO | FLAG |  TYPE   | NP |  STATUS  | CPU | NAME
------|------|------|---------|----|----------|-----|----
    0 |    0 | FIFO | KTHREAD | N  | RUNNING  |   0 | CPU0 IDLE
    1 |    0 | FIFO | KTHREAD | N  | RUNNING  |   1 | CPU1 IDLE
    2 |  201 | RR   | KTHREAD |    | WAITSIG  |   0 | hpwork
    3 |   50 | RR   | KTHREAD |    | WAITSIG  |   0 | lpwork
    5 |  200 | RR   | KTHREAD |    | MQNEMPTY |   0 | km4_log_task
    6 |  107 | RR   | KTHREAD |    | WAITSEM  |   0 | inic_msg_q_task
    7 |  104 | RR   | KTHREAD |    | WAITSEM  |   0 | inic_host_rx_tasklet
    8 |  103 | RR   | KTHREAD |    | WAITSEM  |   0 | inic_ipc_api_host_task
   10 |  105 | RR   | KTHREAD |    | WAITSEM  |   0 | LWIP_TCP/IP
   11 |  100 | RR   | KTHREAD |    | WAITSEM  |   0 | netmgr_event_handler
   12 |  200 | RR   | KTHREAD |    | MQNEMPTY |   0 | log_dump
   13 |  203 | RR   | KTHREAD |    | MQNEMPTY |   0 | binary_manager
   16 |  180 | RR   | TASK    |    | WAITSIG  |   0 | app1
   17 |  100 | RR   | TASK    |    | WAITSIG  |   0 | uwork
   18 |  125 | RR   | TASK    |    | WAITSEM  |   0 | tash
   19 |  100 | RR   | TASK    |    | WAITSEM  |   0 | wifi msg handler
   20 |  100 | RR   | TASK    |    | WAITSEM  |   0 | ble msg handler

### DIFF
--- a/os/fs/procfs/fs_procfsproc.c
+++ b/os/fs/procfs/fs_procfsproc.c
@@ -98,6 +98,12 @@
  */
 
 #define STATUS_LINELEN 128
+#define TASH_TASK_NAME "tash"
+
+/* Modify TASH_TASK_NAME_LENGTH along with TASH_TASK_NAME
+ * whenever we change the name of tash tcb
+ */
+#define TASH_TASK_NAME_LENGTH 4
 
 /****************************************************************************
  * Private Types
@@ -374,6 +380,9 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 	uint32_t total_cpuload = 0;
 	uint32_t total_active = 0;
 #endif
+	uint8_t state;
+	bool is_tash_running = false;
+	pid_t tash_pid = -1;
 
 	remaining = buflen;
 	totalsize = 0;
@@ -409,10 +418,32 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 #else
 	ppid = -1;
 #endif
+
+	state = tcb->task_state;
+
+	/* If tash is currently running tcb, then show tash as waiting
+	 * for semaphore and next ready to run tcb as running.
+	 */
+#if CONFIG_TASK_NAME_SIZE > 0
+	struct tcb_s *rtcb = sched_gettcb(getpid());
+	if (!strncmp(rtcb->name, TASH_TASK_NAME, TASH_TASK_NAME_LENGTH + 1)) {
+		is_tash_running = true;
+		tash_pid = rtcb->pid;
+	}
+
+	if (is_tash_running) {
+		if (tcb->blink != NULL && tcb->blink->pid == tash_pid) {
+			state = TSTATE_TASK_RUNNING;
+		} else if (tcb->pid == tash_pid) {
+			state = TSTATE_WAIT_SEM;
+		}
+	}
+#endif
+
 #ifdef CONFIG_SMP
-	linesize = snprintf(procfile->line, STATUS_LINELEN, "%d %d %d %d %d %d %d %d %d %d", tcb->pid, ppid, tcb->sched_priority, tcb->flags, tcb->task_state, tcb->adj_stack_size, peak_stack, curr_heap, peak_heap, tcb->cpu);
+	linesize = snprintf(procfile->line, STATUS_LINELEN, "%d %d %d %d %d %d %d %d %d %d", tcb->pid, ppid, tcb->sched_priority, tcb->flags, state, tcb->adj_stack_size, peak_stack, curr_heap, peak_heap, tcb->cpu);
 #else
-	linesize = snprintf(procfile->line, STATUS_LINELEN, "%d %d %d %d %d %d %d %d %d %d", tcb->pid, ppid, tcb->sched_priority, tcb->flags, tcb->task_state, tcb->adj_stack_size, peak_stack, curr_heap, peak_heap, 0);
+	linesize = snprintf(procfile->line, STATUS_LINELEN, "%d %d %d %d %d %d %d %d %d %d", tcb->pid, ppid, tcb->sched_priority, tcb->flags, state, tcb->adj_stack_size, peak_stack, curr_heap, peak_heap, 0);
 #endif
 	copysize = procfs_memcpy(procfile->line, linesize, buffer, buflen, &offset);
 	totalsize += copysize;


os/fs/procfs/fs_procfsproc.c
```diff
@@ -409,10 +410,17 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 #else
 	ppid = -1;
 #endif
+	if (tcb->blink != NULL && tcb->blink->pid == getpid()) {
+		state = TSTATE_TASK_RUNNING;
+	} else if (tcb->pid == getpid()) {
+		state = TSTATE_WAIT_SEM;
+	} else {
+		state = tcb->task_state;
+	}
```
sunghan-chang, line 419: I think you consider the TASH only. This function can be called by other thread. (No limitation) Then, this will give wrong information. How about directly checking thread name - TASH for this?

samsung-singh (approved):
now ps cmd shows next task as running instead of tash this change is good to merge.

samsung-agnel (approved):

ritesh55555 (approved):
can be merged

ritesh55555 (approved):
changes in ps command

os/fs/procfs/fs_procfsproc.c
```diff
@@ -409,10 +411,23 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
...
+	if (tcb->blink != NULL && !strcmp(tcb->blink->name,TASH)) {
+		state = TSTATE_TASK_RUNNING;
+	} else if (!strcmp(tcb->name,TASH)) {
+		state = TSTATE_WAIT_SEM;
+	} else {
+#endif
+		state = tcb->task_state;
+#if CONFIG_TASK_NAME_SIZE > 0
+	}
+#endif
```
sunghan-chang, line 425: This causes strcmp twice always. It looks not good at performance.
What we want is just 1 case. When TASH is running to execute the ps command.
Then, how about checking `this_task` first, if it's NOT the TASH, no need to check this logic.
If it's the TASH, we can use two flags to check 1st and 2nd conditionals.
When they are detected, we don't need to check it anymore.

This is just my suggestion. I think we can think better logic.

os/fs/procfs/fs_procfsproc.c
```diff
@@ -98,6 +98,7 @@
  */
 
 #define STATUS_LINELEN 128
+#define TASH "tash"
```
sunghan-chang, line 101: How about `TASH_THREAD_NAME`?

os/fs/procfs/fs_procfsproc.c
neel-samsung: Thank you for the suggestion.

If we check ```this_task``` and if it's not TASH, we need to check the logic if the tcb is the next ready-to-run or not. If it is then we need to change the state to running. So, we would need to check for TASH along with tcb is not next-ready-to-run. This would also require strncmp twice ( in best case).

os/fs/procfs/fs_procfsproc.c
neel-samsung: yes, It looks good.
Can we use ```TASH_TASK_NAME``` ?

os/fs/procfs/fs_procfsproc.c
```diff
@@ -409,10 +411,23 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
 #else
 	ppid = -1;
 #endif
+
+#if CONFIG_TASK_NAME_SIZE > 0
+	if (tcb->blink != NULL && !strcmp(tcb->blink->name,TASH)) {
```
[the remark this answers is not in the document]
neel-samsung: okay, Let me take care.

os/fs/procfs/fs_procfsproc.c
sunghan-chang: Sorry I don't know why we should change the state of next ready to run thread to running when running taka is not the TASH. Could you elaborate it more?

neel-samsung: @sunghan-chang Please review it

os/fs/procfs/fs_procfsproc.c
```diff
@@ -98,6 +98,8 @@
  */
 
 #define STATUS_LINELEN 128
+#define TASH_TASK_NAME "tash"
+#define TASH_TASK_NAME_LENGTH strlen(TASH_TASK_NAME)
```
sunghan-chang, line 102: Let's use 4 instead of strlen. But need to leave notice to modify this together with NAME string when name is changed.

os/fs/procfs/fs_procfsproc.c
```diff
@@ -409,10 +412,26 @@ static ssize_t proc_entry_stat(FAR struct proc_file_s *procfile, FAR struct tcb_
...
+#if CONFIG_TASK_NAME_SIZE > 0
+	struct tcb_s *rtcb = sched_gettcb(getpid());
+	if (strncmp(rtcb->name, TASH_TASK_NAME, TASH_TASK_NAME_LENGTH + 1)) {
+		state = tcb->task_state;
+	} else if (tcb->blink != NULL && !strncmp(tcb->blink->name, TASH_TASK_NAME, TASH_TASK_NAME_LENGTH + 1)) {
+		state = TSTATE_TASK_RUNNING;
+	} else if (!strncmp(tcb->name, TASH_TASK_NAME, TASH_TASK_NAME_LENGTH + 1)) {
+		state = TSTATE_WAIT_SEM;
+	} else {
+#endif
```
jeongarmy, line 425: How about this?
```
#if CONFIG_TASK_NAME_SIZE > 0
 if (....) {
 } else if {
 ...
 } else
#endif
 {
     state = tcb->task_state;
 }
 ```

Comment (neel-samsung) [reply os/fs/procfs/fs_procfsproc.c:425]: Case 1 is similar to mine. no logic change but just a code format change. looks good.

Case 2 : It would cause multiple strncmp checks when current_tcb is not TASH. We should not miss the below condition:
```if (strncmp(rtcb->name, TASH_TASK_NAME, TASH_TASK_NAME_LENGTH + 1))```

Comment (neel-samsung) [reply os/fs/procfs/fs_procfsproc.c:425]: Let us go with Case 1. It makes more sense and easy to interpret.

@jeongarmy Pls review it

Comment (sunghan-chang) [reply os/fs/procfs/fs_procfsproc.c:425]: looks strange..
```
struct tcb_s *rtcb = this_task();
if (!strncmp(TASH)) {
    is_tash_running = true;
    tash_pid = rtcb->pid;
}

if (is_tash_running) {
    // check pid for TASH, not strncmp everytime
}
```

sunghan-chang (dismissed):

sunghan-chang (approved):

PR #6815, opened 2025-05-08
os/arch/arm/src: SMP/AMP migrate systick control to running core during pause

lhenry-realtek:
This PR demonstrates functional switching of systick count when pausing a core.

As systick is counted by each individual core's timer PPI, it will trigger independently of each other. Currently, this will cause the system tick to have multiple count and affect OS timing.

To avoid multiple counting, only one core will be allowed to increment the common OS tick counter

If Core0 happens to be paused, then the other core should take over, and then return control when Core0 resumes

Please check with flat_dev_ddr config

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
@@ -60,10 +60,6 @@
 #  error SYSTICK_RELOAD exceeds the range of the RELOAD register
 #endif
 
-#if defined(CONFIG_SMP) && CONFIG_SMP_NCPUS > 1
-volatile uint32_t g_active_system_timer_cpu;
-#endif
-
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
@@ -79,15 +75,6 @@ volatile uint32_t g_active_system_timer_cpu;
 
 int up_timerisr(int irq, uint32_t *regs)
 {
-#if defined(CONFIG_SMP) && CONFIG_SMP_NCPUS > 1
-    /* Process timer interrupt, only if the active timer is set to this cpu's PPI */
-    if (g_active_system_timer_cpu == up_cpu_index()) {
-      sched_process_timer();
-    }
-#else
-    sched_process_timer();
-#endif
-
     /* Clear interrupt */
     uint32_t delta_ticks;
     uint64_t last_cycle;
@@ -103,6 +90,13 @@ int up_timerisr(int irq, uint32_t *regs)
 #else
 	  delta_ticks = 1;
 #endif
+
+    u32 ticks_to_process = delta_ticks;
+    while (ticks_to_process > 0) {
+      /* process missing ticks */
+      sched_process_timer();
+      ticks_to_process--;
+    }
 
     arm_arch_timer_set_compare(last_cycle + delta_ticks * SYSTICK_RELOAD);
     return 0;
--- a/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
@@ -60,6 +60,10 @@
 #  error SYSTICK_RELOAD exceeds the range of the RELOAD register
 #endif
 
+#if defined(CONFIG_SMP) && CONFIG_SMP_NCPUS > 1
+volatile uint32_t g_active_system_timer_cpu;
+#endif
+
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
@@ -75,8 +79,14 @@
 
 int up_timerisr(int irq, uint32_t *regs)
 {
-    /* Process timer interrupt */
+#if defined(CONFIG_SMP) && CONFIG_SMP_NCPUS > 1
+    /* Process timer interrupt, only if the active timer is set to this cpu's PPI */
+    if (g_active_system_timer_cpu == up_cpu_index()) {
+      sched_process_timer();
+    }
+#else
     sched_process_timer();
+#endif
 
     /* Clear interrupt */
     uint32_t delta_ticks;
--- a/os/arch/arm/src/armv7-a/arm_cpugating.c
+++ b/os/arch/arm/src/armv7-a/arm_cpugating.c
@@ -38,9 +38,12 @@
 #include "gic.h"
 #include "sched/sched.h"
 #include "barriers.h"
+#include "arch_timer.h"
 
 #ifdef CONFIG_CPU_GATING
 static volatile uint32_t g_cpugating_flag[CONFIG_SMP_NCPUS];
+
+extern volatile uint32_t g_active_system_timer_cpu;
 
 /****************************************************************************
  * Public Functions
@@ -74,6 +77,11 @@ void up_do_gating(void)
 			SP_WFE();
 		}
 		irqrestore(PrevIrqStatus);
+
+		/* transfer systick control back to cpu0 if cpu0 exit gating. cpu1 do not need */
+		if (cpu == 0) {
+			g_active_system_timer_cpu = cpu;
+		}
 	}
 
 }
@@ -117,6 +125,9 @@ void up_cpu_gating(int cpu)
 
 	/* Fire SGI for cpu to enter gating */
 	arm_cpu_sgi(GIC_IRQ_SGI3, (1 << cpu));
+
+	/* after gating other CPU, this cpu is the active timer, since the other one will be gated */
+	g_active_system_timer_cpu = this_cpu();
 }
 
 #endif /* CONFIG_CPU_GATING */
--- a/os/arch/arm/src/armv7-a/arm_cpugating.c
+++ b/os/arch/arm/src/armv7-a/arm_cpugating.c
@@ -43,8 +43,6 @@
 #ifdef CONFIG_CPU_GATING
 static volatile uint32_t g_cpugating_flag[CONFIG_SMP_NCPUS];
 
-extern volatile uint32_t g_active_system_timer_cpu;
-
 /****************************************************************************
  * Public Functions
  ****************************************************************************/
@@ -55,6 +53,9 @@ void up_set_gating_flag_status(uint32_t CoreID, uint32_t val)
 	ARM_DSB();
 	/* Flag already reach 0 */
 	if (!g_cpugating_flag[CoreID]) {
+		if (this_cpu() != 0) {
+			up_timer_disable();
+		}
 		SP_SEV();
 	}
 }
@@ -77,11 +78,6 @@ void up_do_gating(void)
 			SP_WFE();
 		}
 		irqrestore(PrevIrqStatus);
-
-		/* transfer systick control back to cpu0 if cpu0 exit gating. cpu1 do not need */
-		if (cpu == 0) {
-			g_active_system_timer_cpu = cpu;
-		}
 	}
 
 }
@@ -127,7 +123,9 @@ void up_cpu_gating(int cpu)
 	arm_cpu_sgi(GIC_IRQ_SGI3, (1 << cpu));
 
 	/* after gating other CPU, this cpu is the active timer, since the other one will be gated */
-	g_active_system_timer_cpu = this_cpu();
+	if (cpu == 0 && g_cpugating_flag[cpu] != 0) {
+		up_timer_enable();
+	}
 }
 
 #endif /* CONFIG_CPU_GATING */
--- a/os/arch/arm/src/armv7-a/arm_cpustart.c
+++ b/os/arch/arm/src/armv7-a/arm_cpustart.c
@@ -54,6 +54,7 @@
 #include "gic.h"
 #include "sched/sched.h"
 #include "smp.h"
+#include "arch_timer.h"
 
 #ifdef CONFIG_SMP
 
@@ -110,6 +111,9 @@ int arm_start_handler(int irq, void *context, void *arg)
 	struct tcb_s *tcb = this_task();
 
 	svdbg("CPU%d Started\n", this_cpu());
+
+	/* start generic timer PPI for this core */
+	up_timer_initialize();
 
 #ifdef CONFIG_SCHED_INSTRUMENTATION
 	/* Notify that this CPU has started */
--- a/os/arch/arm/src/amebasmart/amebasmart_smp.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_smp.c
@@ -168,6 +168,10 @@ void smp_init(void)
 			continue;
 		}
 		err = psci_cpu_on(xCoreID, (unsigned long)__cpu1_start);
+
+		/* Wait for CPU1 to boot from ATF to SGI1 WFI loop, before os_smp_start is called */
+		DelayUs(50);
+
 		/* If we failed to boot secondary core here, it will be very likely
 		issue is coming from ATF/kernel flow, and that operation is irreversible
 		(ie. There is no way we can restore the secondary core to come
--- a/os/arch/arm/src/armv7-a/arm_cpupause.c
+++ b/os/arch/arm/src/armv7-a/arm_cpupause.c
@@ -52,8 +52,11 @@
 #include "up_internal.h"
 #include "gic.h"
 #include "sched/sched.h"
+#include "arch_timer.h"
 
 #ifdef CONFIG_SMP
+
+extern volatile uint32_t g_active_system_timer_cpu;
 
 /****************************************************************************
  * Private Data
@@ -363,6 +366,9 @@ int up_cpu_pause(int cpu)
 	spin_lock(&g_cpu_paused[cpu]);
 	spin_unlock(&g_cpu_paused[cpu]);
 
+	/* after taking the spinlocks, this cpu is the active timer, since the other one will be paused */
+	g_active_system_timer_cpu = this_cpu();
+
 	/* On successful return g_cpu_wait will be locked, the other CPU will be
 	 * spinning on g_cpu_wait and will not continue until g_cpu_resume() is
 	 * called.  g_cpu_paused will be unlocked in any case.
@@ -418,6 +424,11 @@ int up_cpu_resume(int cpu)
 		spin_lock(&g_cpu_resumed[cpu]);
 
 		spin_unlock(&g_cpu_resumed[cpu]);
+
+		/* transfer systick control back to cpu0 if cpu0 is being resumed */
+		if (cpu == 0) {
+			g_active_system_timer_cpu = cpu;
+		}
 	}
 
 	return OK;
--- a/os/arch/arm/src/armv7-a/arm_cpupause.c
+++ b/os/arch/arm/src/armv7-a/arm_cpupause.c
@@ -56,8 +56,6 @@
 
 #ifdef CONFIG_SMP
 
-extern volatile uint32_t g_active_system_timer_cpu;
-
 /****************************************************************************
  * Private Data
  ****************************************************************************/
@@ -366,8 +364,13 @@ int up_cpu_pause(int cpu)
 	spin_lock(&g_cpu_paused[cpu]);
 	spin_unlock(&g_cpu_paused[cpu]);
 
-	/* after taking the spinlocks, this cpu is the active timer, since the other one will be paused */
-	g_active_system_timer_cpu = this_cpu();
+	/* Check if we are pausing cpu0.
+	* In this case, after pause request is being handled by cpu0,
+	* we need to enable timer irq on current cpu
+	*/
+	if (cpu == 0) {
+		up_timer_enable();
+	}
 
 	/* On successful return g_cpu_wait will be locked, the other CPU will be
 	 * spinning on g_cpu_wait and will not continue until g_cpu_resume() is
@@ -427,7 +430,7 @@ int up_cpu_resume(int cpu)
 
 		/* transfer systick control back to cpu0 if cpu0 is being resumed */
 		if (cpu == 0) {
-			g_active_system_timer_cpu = cpu;
+			up_timer_disable();
 		}
 	}
 


edwakuwaku:
why do we need to pass the tickisr to secondary core when primary core is paused? Isn't the tick will be compensated during next cycle of tickisr happens (ie. after primary core is resumed). Also, current implementation will cause resource wastage, since h/w has to always react to secondary cpu tickisr but nothing is being done in f/w.
In https://github.com/Samsung/TizenRT/blob/master/os/arch/arm/src/armv7-a/arm_gicv2.c#L151, all the cores bringup will do this, which all the irq for SGI and PPI will be assigned, I think it should be possible to explicitly disable specific irq num for specific core through the GIC cpu interface. Just need to figure out the correct offset for bit masking through https://github.com/Samsung/TizenRT/blob/master/os/arch/arm/src/armv7-a/gic.h...

edwakuwaku:
It's clear to me for point 1.
For point 2, the reason to ask this is because, the current implementation has direct interaction between bsp and arch layer code which shouldn't be allowed in such hierarchal OS. Some interface should be created from arch layer, and then invoke from bsp layer. 
Maybe @sunghan-chang can comment this better...

sunghan-chang:
@kishore-sn @lhenry-realtek @edwakuwaku If we set the irq affinity to both (any core) instead of this PR, what happen?

edwakuwaku:
> @kishore-sn @lhenry-realtek @edwakuwaku If we set the irq affinity to both (any core) instead of this PR, what happen?

In gic, there are distributor and cpu based interface, for shared peripheral interface, it is controlled through distributor, for SGI and PPI, it is through cpu based interface, each core has its own independent set of registers... I don't know what do you mean by what will happen, it is based on user control to decide the behaviour...

sunghan-chang:
@kishore-sn @lhenry-realtek It looks we should merge this ASAP. Could you check it ASAP?

> Agreed that this is just a "quick hack" to demo the flow, there should be a proper api to handle the systicks and timers that's not just via direct access of extern variable
> 
> Maybe @kishore-sn has some suggestion to modify the system timer APIs for this, and also to handle more than 2 core situation

kishore-sn:
Hi all, 
We need following cases to work with this PR:
1. cpu1 pauses cpu0. cpu1 goes into sleep. Cpu1 must wake up after sleep duration.
2. Cpu1 gates cpu0. Cpu1 goes to sleep. After sleep duration Cpu1 must wake up and continue.
3. During normal operation, the system timer must count properly. If we call sleep(30), then wake up must happen after 30 secs.
4. cpu1 calls cpu pause for cpu0, at that time, system gets assert.
5. cpu1 calls cpu gating for cpu0, at that time system get assert.
6. Large flash access operation is happening on cpu1 (so cpu0 gets gated). After the flash operation, the system time must be proper.


When I checked with this PR, 
- cases 1, 3, 4 and 5 are passing
- case 2 is still failing. When cpu0 goes into gating and cpu1 goes to sleep, then system gets hanged. 
- I am yet to verify case 6. I will update in sometime.

kishore-sn:
> it is using up_cpu

I meant usleep in all cases.

kishore-sn:
> Hi all, We need following cases to work with this PR:
> 
> 1. cpu1 pauses cpu0. cpu1 goes into sleep. Cpu1 must wake up after sleep duration.
> 2. Cpu1 gates cpu0. Cpu1 goes to sleep. After sleep duration Cpu1 must wake up and continue.
> 3. During normal operation, the system timer must count properly. If we call sleep(30), then wake up must happen after 30 secs.
> 4. cpu1 calls cpu pause for cpu0, at that time, system gets assert.
> 5. cpu1 calls cpu gating for cpu0, at that time system get assert.
> 6. Large flash access operation is happening on cpu1 (so cpu0 gets gated). After the flash operation, the system time must be proper.
> 
> When I checked with this PR,
> 
> * cases 1, 3, 4 and 5 are passing
> * case 2 is still failing. When cpu0 goes into gating and cpu1 goes to sleep, then system gets hanged.
> * I am yet to verify case 6. I will update in sometime.

Case 6 is also failing. 
Test procedure: 
1. Run below steps on cpu1
2. Allocate 5mb buffer with dummy data
3. Open file /mnt/test.txt
4. Write 5mb buffer to file
5. Measure time taken using clock_gettime API

Actual time taken is around 50 to 60secs. But time calculated using clock_gettime is 10secs.

kishore-sn:
> Agreed that this is just a "quick hack" to demo the flow, there should be a proper api to handle the systicks and timers that's not just via direct access of extern variable
> 
> Maybe @kishore-sn has some suggestion to modify the system timer APIs for this, and also to handle more than 2 core situation

For more than 2 cores scenario, I am not sure how pause- resume will work. Is it possible for core A to pause core B and then core C to resume core B? 
In any case, I think it is better to do currently implementation assuming two cores only. 
However, in order to keep the code of arch layer separate from bsp, I suggest below method. This method does not use global variable. There will be no changes in amebasmart_timerisr.c.

```
diff --git a/os/arch/arm/src/armv7-a/arm_cpupause.c b/os/arch/arm/src/armv7-a/arm_cpupause.c
index fcee79f1f..970ea2086 100644
--- a/os/arch/arm/src/armv7-a/arm_cpupause.c
+++ b/os/arch/arm/src/armv7-a/arm_cpupause.c
@@ -52,6 +52,7 @@
 #include "up_internal.h"
 #include "gic.h"
 #include "sched/sched.h"
+#include "arch_timer.h"

 #ifdef CONFIG_SMP

@@ -363,6 +364,14 @@ int up_cpu_pause(int cpu)
        spin_lock(&g_cpu_paused[cpu]);
        spin_unlock(&g_cpu_paused[cpu]);

+       /* Check if we are pausing cpu0.
+        * In this case, after pause request is being handled by cpu0,
+        * we need to enable timer irq on current cpu
+        */
+       if (cpu == 0) {
+               up_timer_enable();
+       }
+
        /* On successful return g_cpu_wait will be locked, the other CPU will be
         * spinning on g_cpu_wait and will not continue until g_cpu_resume() is
         * called.  g_cpu_paused will be unlocked in any case.
@@ -418,6 +427,11 @@ int up_cpu_resume(int cpu)
                spin_lock(&g_cpu_resumed[cpu]);

                spin_unlock(&g_cpu_resumed[cpu]);
+
+               /* transfer systick control back to cpu0 if cpu0 is being resumed */
+               if (cpu == 0) {
+                       up_timer_disable();
+               }
        }

        return OK;
```

kishore-sn:
> The following tests have been run with this commit:
> 
> Pause Cases
> 
> * cpu1 pause cpu0, then cpu1 aborts
> * board enter PG, cpu1 pause cpu0, then cpu1 aborts
> * cpu0 pause cpu1, then cpu0 aborts
> * board enter PG, cpu0 pause cpu1, then cpu0 aborts
> 
> Gate Cases
> 
> * cpu1 gate cpu0, then cpu1 aborts
> * board enter PG, cpu1 gate cpu0, then cpu1 aborts
> * cpu0 gate cpu1, then cpu0 aborts
> * board enter PG, cpu0 gate cpu1, then cpu0 aborts
> 
> Normal Cases
> 
> * both cpu active, cpu0 abort / cpu1 abort
> * board enter PG, cpu0 abort / cpu1 abort
> 
> @kishore-sn it was unclear what you meant by "sleep", so we assumed it to be PG.
> 
> During testing of cases with PG we caught a issue where during wakeup, CPU1 was stuck waiting in WFI for SGI1, but CPU0 was already paused. I suspect that os_smp_start (send sgi1) was called, but CPU1 was not fully initialized (configure gic, irq) to receive it, so the IRQ was not handled properly.
> 
> Inserting this delay helps to resolve both gate and pause cases using the extern variable method. I will check with your method of using up_timer_enable/disable shortly
> 
> For the last test case of long flash write, do you have test code? I tried to use the following, but it did not show a large time in order of 50-60seconds:
> 
> ```
> static pthread_addr_t run_smp_gate_operation(pthread_addr_t parameter)
> {
> 	struct timespec before;
> 	struct timespec after;
> 
> 	u8* buf;
> 
> 	buf = malloc(1024*1000*5);
> 	for (int i = 0; i < 1024*1000*5; i++) {
> 		buf[i] = rand() & 0xFF;
> 	}
> 
> 	(void)clock_gettime(CLOCK_REALTIME, &before);
> 
> 	/* do flash op */
> 	int fd = open("/mnt/test.txt", O_RDWR | O_CREAT);
> 	write(fd, buf, 1024*1000*1);
> 	close(fd);
> 
> 	(void)clock_gettime(CLOCK_REALTIME, &after);
> 	printf("BEFORE: (%lu sec, %lu nsec)\n",
> 		   (unsigned long)before.tv_sec, (unsigned long)before.tv_nsec);
> 	printf("AFTER:  (%lu sec, %lu nsec)\n",
> 		   (unsigned long)after.tv_sec, (unsigned long)after.tv_nsec);
> }
> ```
> 
> ```
> #define CPU_ZERO(s) do { *(s) = 0; } while (0)
> #define CPU_SET(c,s) do { *(s) |= (1 << (c)); } while (0)
> int test_smp_gate(int argc, char** argv) {
> 
> 	pthread_attr_t attr;
> 	int ret = pthread_attr_init(&attr);
>  
> 	CPU_ZERO(&attr.affinity);
> 	CPU_SET(1, &attr.affinity);
>  
> 	pid_t thread_id;
>  
> 	ret = pthread_create(&thread_id, &attr, run_smp_gate_operation, NULL);
> 
> 	return 0;
> }
> ```

Hi,
I am trying similar method. I retried now with your latest code. 
But still I am facing the issue. 
There is not much difference for small file size. 
For example, while writing 1MB of data, it takes around 6 secs. This is working well.
But, when I try to write 5MB data, then it prints 47secs. But the actual time taken is 248secs. 
My userfs partition size is 6144KB. 
Please retry on your side with large file write operation.

Dane-Kang:
> Hi @kishore-sn
> 
> We have managed to reproduce the tick loss issue. I suspect that this tick loss is due to interrupts being disabled on CPU0, while the flash operation is happening, and then it is restored, but not enough ticks were added back
> 
> Currently checking on this

Hi @kishore-sn , @sunghan-chang ,
@lhenry-realtek trying to found solution for fix tick loss now so far but we haven't found one yet...
We're trying our best to fix this issue ASAP, so I hope you understand current status..
Thank you.

sunghan-chang:
@lhenry-realtek When we execute file operation in cpu0, time tick is much delayed. In 5 mins, 2~3 mins are delayed. When this is applied, is this fixed as well?

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -94,6 +91,13 @@ int up_timerisr(int irq, uint32_t *regs)
 	  delta_ticks = 1;
 #endif
 
+    u32 ticks_to_process = delta_ticks;
+    while (ticks_to_process > 0) {
+      /* process missing ticks */
+      sched_process_timer();
+      ticks_to_process--;
+    }
+
```
ewoodev, line 100: I think, this code can make duplicated compensate missing tick.
https://github.com/Samsung/TizenRT/blob/d4a807c961546331a8957fd4c1e9aa364808026b/os/pm/pm_wakehandler.c#L64-L77

pm_wakehandler is conpensating missing tick.
but this code looks conpensating tick irq loss.

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -94,6 +91,13 @@ int up_timerisr(int irq, uint32_t *regs)
 	  delta_ticks = 1;
 #endif
 
+    u32 ticks_to_process = delta_ticks;
```
ewoodev, line 94: this PR change should be applied regardless PM. right?

ewoodev: The delta_ticks is updated in only CONFIG_PM
If CONFIG_PM is disabled, `delta_ticks = 1`. is it okay?

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
ewoodev:
```
void up_pm_board_sleep(void (*wakeuphandler)(clock_t, pm_wakeup_reason_code_t))
{
...

	/* Arch timer is running at 50Mhz */
	arm_arch_timer_set_compare(arm_arch_timer_count() + 50000);
	arm_arch_timer_int_mask(0);
}
```
the timer is setted latest count in above code, so duplicating compensation is not ocurred right?

well, then why did we need below code for CONFIG_PM?
```
#ifdef CONFIG_PM
	  if (arm_arch_timer_count() < last_cycle) {
		  return -1;
	  } else {
		  delta_ticks = (uint32_t)((arm_arch_timer_count() - last_cycle) / SYSTICK_RELOAD) + 1;
	  }
#else
```

kishore-sn:
> 5MB flash op (DDR) @sunghan-chang
> 
> ```
> (13:59:01.450) 	System Time: 19 May 2025, 00:00:01 [s] UTC Hardware RTC Support
> (13:59:01.450) TASH>><NUL>
> (13:59:03.341) date
> (13:59:03.459) date
> (13:59:03.459) 
> (13:59:03.459) May 19 00:00:03 2025 Mon UTC
> (13:59:03.459) TASH>><NUL>
> (13:59:07.559) smpgate 0 x
> (13:59:07.680) smpgate 0 x
> (13:59:07.680) 
> (13:59:07.680) TASH>><NUL>argc: 3
> (13:59:07.680) target cpu: 0
> (13:59:07.680) flash gate example on cpu0
> (13:59:07.838) 
> (13:59:07.838) START
> (13:59:07.838) up_timerisr: arm_arch_timer_count: 357331925
> (13:59:19.038) 
> (13:59:19.038) END
> (13:59:19.038) BEFORE: (1747612807 sec, 672000000 nsec)
> (13:59:19.089) AFTER:  (1747612818 sec, 873000000 nsec)
> (13:59:19.089) up_timerisr: arm_arch_timer_count: 917370723
> (13:59:24.536) date
> (13:59:24.623) date
> (13:59:24.623) 
> (13:59:24.623) May 19 00:00:24 2025 Mon UTC
> (13:59:24.623) TASH>><NUL>
> ```
> 
> ```
> Time based on PC UART log time
> START: 13:59:07.838
> END: 13:59:19.038
> delta = ~11.2s
> 
> Time based on clock_gettime()
> BEFORE: (1747612807 sec, 672000000 nsec)
> AFTER:  (1747612818 sec, 873000000 nsec)
> delta = ~11.201s 
> 
> Time based on arm system counter:
> START tick (t1) = 357331925
> END tick (t2) = 917370723
> delta = (t2-t1)/SYSTICK_RELOAD +1
>       = 11201.999979996 ms
>       = 11.202s
> ```

@lhenry-realtek 
You mentioned that 5MB is taking 3 minutes. But, above results show 11 secs. Is this correct?

os/arch/arm/src/armv7-a/arm_cpustart.c
```diff
@@ -111,6 +112,9 @@ int arm_start_handler(int irq, void *context, void *arg)
 
 	svdbg("CPU%d Started\n", this_cpu());
 
+	/* start generic timer PPI for this core */
+	up_timer_initialize();
```
jeongarmy, line 116: It is called in up_initialize(). Please check it is not problem

jeongarmy: Please check whether it is ok although it is called multiple times or add conditions to avoid duplicated calls.

sunghan-chang:
@lhenry-realtek @kishore-sn @ewoodev @jeongarmy  This must be merged by today. Let's do final review and verify the PR ASAP.

os/arch/arm/src/armv7-a/arm_cpustart.c
```diff
@@ -111,6 +112,9 @@ int arm_start_handler(int irq, void *context, void *arg)
...
+
```
kishore-sn, line 117: I think we must call up_timer_disable(); here.

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
ewoodev: Okay, sometime we are test non-CONFIG_PM case.
Please support non-CONFIG_PM case also, you can raise new PR after merge this

os/arch/arm/src/armv7-a/arm_cpustart.c
kishore-sn: When I tested your latest code, the flash write is still causing issue. 
When the write is happening on cpu0, then time is ok (11 secs).
But when running on cpu1, the time shows 28secs.

```
TASH>>hello 0
TASH>>Hello, World!!
Running file write test on cpu 0
BEFORE: (1747612937 sec, 136000000 nsec)
AFTER:  (1747612948 sec, 763000000 nsec)
Time taken = 11 secs


TASH>>hello 1
TASH>>Hello, World!!
Running file write test on cpu 1
BEFORE: (1747612934 sec, 801000000 nsec)
AFTER:  (1747612962 sec, 222000000 nsec)
Time taken = 28 secs
```

sunghan-chang:
> @lhenry-realtek When I tested your latest code, the flash write is still causing issue. When the write is happening on cpu0, then time is ok (11 secs). But when running on cpu1, the time shows 28secs.
> 
> ```
> TASH>>hello 0
> TASH>>Hello, World!!
> Running file write test on cpu 0
> BEFORE: (1747612937 sec, 136000000 nsec)
> AFTER:  (1747612948 sec, 763000000 nsec)
> Time taken = 11 secs
> 
> 
> TASH>>hello 1
> TASH>>Hello, World!!
> Running file write test on cpu 1
> BEFORE: (1747612934 sec, 801000000 nsec)
> AFTER:  (1747612962 sec, 222000000 nsec)
> Time taken = 28 secs
> ```

@kishore-sn Could you measure them without the PR? I wanna know before and after.

kishore-sn:
@lhenry-realtek 
There seems to be some issue in your implementation in the gating file. I modified it and now the timing values in both cpus are coming properly. Below are the test results. (5MB file size and 6144KB partition size)
This test is performed with CONFIG_PM=y, but pm_resume not called. So, PM is not active. 

```
TASH>>hello 1
TASH>>Hello, World!!
Running file write test on **cpu 1**
BEFORE: (1747612804 sec, 102000000 nsec)
AFTER:  (1747612815 sec, 713000000 nsec)
Time taken = 11 secs           <<<<<<<<<<<<< Fresh file

TASH>>hello 1
TASH>>Hello, World!!
Running file write test on **cpu 1**
BEFORE: (1747612855 sec, 424000000 nsec)
AFTER:  (1747613512 sec, 45000000 nsec)
Time taken = 657 secs        <<<<<<<<<<<<<<< Overwrite file


TASH>>hello 0
TASH>>Hello, World!!
Running file write test on **cpu 0**
BEFORE: (1747612871 sec, 180000000 nsec)
AFTER:  (1747612882 sec, 818000000 nsec)
Time taken = 11 secs        <<<<<<<<<<<<< Fresh file

TASH>>hello 0
TASH>>Hello, World!!
Running file write test on **cpu 0**
BEFORE: (1747612887 sec, 820000000 nsec)
AFTER:  (1747613548 sec, 76000000 nsec)
Time taken = 661 secs         <<<<<<<<<<<<<<< Overwrite file

```

Please find my changes in the below patch. 
```
diff --git a/os/arch/arm/src/armv7-a/arm_cpugating.c b/os/arch/arm/src/armv7-a/arm_cpugating.c
index 4b3732fdf..753bb7049 100644
--- a/os/arch/arm/src/armv7-a/arm_cpugating.c
+++ b/os/arch/arm/src/armv7-a/arm_cpugating.c
@@ -53,6 +53,10 @@ void up_set_gating_flag_status(uint32_t CoreID, uint32_t val)
        ARM_DSB();
        /* Flag already reach 0 */
        if (!g_cpugating_flag[CoreID]) {
+                if (this_cpu() != 0) {
+                        up_timer_disable();
+                }
+
                SP_SEV();
        }
 }
@@ -75,11 +79,6 @@ void up_do_gating(void)
                        SP_WFE();
                }
                irqrestore(PrevIrqStatus);
-
-               /* transfer systick control back to cpu0 if cpu0 exit gating. cpu1 do not need */
-               if (cpu == 0) {
-                       up_timer_enable();
-               }
        }
 
 }
@@ -124,10 +123,10 @@ void up_cpu_gating(int cpu)
        /* Fire SGI for cpu to enter gating */
        arm_cpu_sgi(GIC_IRQ_SGI3, (1 << cpu));
 
-       /* after gating other CPU, this cpu is the active timer, since the other one will be gated */
-       if (cpu == 0) {
-               up_timer_enable();
-       }
+        if (cpu == 0 && g_cpugating_flag[cpu] == 1) {
+                up_timer_enable();
+        }
+
 }

```

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -94,6 +91,13 @@ int up_timerisr(int irq, uint32_t *regs)
...
+    while (ticks_to_process > 0) {
```
kishore-sn, line 95: Does this work well if CONFIG_PM is enabled, but pm_resume is not called?

os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -117,6 +121,11 @@ void up_cpu_gating(int cpu)
 
 	/* Fire SGI for cpu to enter gating */
 	arm_cpu_sgi(GIC_IRQ_SGI3, (1 << cpu));
+
+	/* after gating other CPU, this cpu is the active timer, since the other one will be gated */
+	if (cpu == 0 && g_cpugating_flag[cpu] == 1) {
```
kishore-sn, line 126: Minor change. `g_cpugating_flag[cpu] != 0`

kishore-sn:
> > @kishore-sn it was unclear what you meant by "sleep", so we assumed it to be PG.
> 
> @lhenry-realtek, When i say sleep, I mean calling "sleep(30)" or usleep(500) in the test app.
> 
> ```
> test_pthread{
>     pause cpu0;
>     usleep(500) <<<< hang
>     xxx
>     xxx
> }
> ```

Hi, 
I had reported this issue previously. This issue is still present with latest code. 
If cpu0 gates cpu1 and then cpu0 goes into sleep, it is able to wake up.
But, if cpu1 gates cpu0 and cpu1 goes into sleep, it is not waking up. Device hangs.

```
TASH>>hello 0
TASH>>Hello, World!!
Before gating cpu1
Sleep on cpu0
Wake up cpu1 and cpu0
Time taken = 5 secs

TASH>>hello 1
TASH>>Hello, World!!
Before gating cpu0   <<<<<<<<<< Hang
```

kishore-sn:
I am not sure why this issue is happening. We are enabling the timer PPI for cpu1 when gating cpu0. 
And sleep -> nanosleep seems to use the system timer. 
@lhenry-realtek you can find my test code on my private fork at https://github.com/kishore-sn/TizenRT/commit/76c897abafa6d7ebc300f6b0cfefd5a48a1ee510

os/arch/arm/src/armv7-a/arm_cpugating.c
kishore-sn: Looks like this change is not yet added. Can you pls add?

kishore-sn:
> Hi @kishore-sn
> 
> I tested your code and reproduced the issue
> 
> From T32 output CPU0 seems to be stuck in gating waiting for flag to unset, CPU1 tick ISRs are triggering and system timer is updated
> 
> But the scheduler is not re-scheduling the thread back on CPU1 so the thread cant run, even if it has been unblocked after nanosleep wdog expired
> 
> The observation is as such: gate CPU0 CPU1 run nanosleep ---> thread status from RUNNING ==> WAIT_SIG and then thread blocked to wakeup by wdog timer ticks ISR... CPU1 wdog wake ---> thread unblocked ---> thread status from WAIT_SIG ==> PENDING CPU1 switches to idle task CPU1 never adds thread back to g_readytorun so the scheduler cannot pre-empt CPU1 idle to run the thread
> 
> g_cpu_schedlock is being held by something in CPU1, so CPU1 is unable to restore the pending tasks while CPU0 is gated

@lhenry-realtek 
Thank you for your analysis. We can take this up in a different PR. 
Let us merge this PR now.

sunghan-chang (approved):

sunghan-chang:
@lhenry-realtek This PR has 7 commits. All are necessary in history? All are different topics? If not, could you clean them up?

jeongarmy (approved):
looks good

PR #6817, opened 2025-05-09
os/semaphore: Check semcount for mutex 

kishore-sn:


os/kernel/semaphore/sem_trywait.c
```diff
@@ -132,6 +132,10 @@ int sem_trywait(FAR sem_t *sem)
 
 		saved_state = enter_critical_section();
 
+		if ((sem->flags & FLAGS_SIGSEM) != 0) {
+			DEBUGASSERT(sem->semcount < 2);
+		}
```
sunghan-chang, line 137: I don't know this is right. When we use a sem for signal purpose, we should not meet the semcount 2 ever? 
I don't think we can guarantee and it's a wrong state.

neel-samsung (approved):

samsung-singh (approved):
added sanity check for sem count for mutex in os/semaphore module this change is good to merge.

kripal-samsung (approved):

Sambhavi-1 (approved):
Sanity check added for semcount for mutex

os/kernel/semaphore/sem_trywait.c
kishore-sn: Ok. I added a new flag to be used for mutex only.

os/kernel/semaphore/sem_trywait.c
```diff
@@ -132,6 +132,10 @@ int sem_trywait(FAR sem_t *sem)
 
 		saved_state = enter_critical_section();
 
+		if ((sem->flags & FLAGS_SEM_MUTEX) != 0) {
```
sunghan-chang, line 135: Do you need to check it in wait as well? Because you add it in post, it should not be 2 in normal case.
Do you intend to detect by memory corruption?

os/kernel/pthread/pthread_mutexinit.c
```diff
@@ -156,6 +156,7 @@ int pthread_mutex_init(FAR pthread_mutex_t *mutex,
 		if (status != OK) {
 			ret = get_errno();
 		}
+		mutex->sem.flags |= FLAGS_SEM_MUTEX;
```
sunghan-chang, line 159: 1. Is it ok to set it regardless of the status value?
2. Regardless of this PR, this function looks strange.. 
    1. This line and sem_setprotocol are executed regardless of result of sem_init.
    2. The sem_setprotocol overwrites the status value (sem_init result is removed..)

sunghan-chang (commented):
Don't you need to add this in the binary_manager?

os/kernel/semaphore/sem_trywait.c
kishore-sn: Actually, we are not sure why the semcount is becoming 2 for mutex case. So I added this check in all the semaphore API's so that whenever a sem api is called, if the semcount has increased, then we can catch the issue immediately.

kishore-sn:
> Don't you need to add this in the binary_manager?

Added in binary manager. Thank you.

sunghan-chang (commented):
@jeongarmy Could you review and leave the approval?

r-prabu (dismissed):

jeongarmy:
I think it cannot cover when mutex is initialized by PTHREAD_MUTEX_INITIALIZER(). Please check this.

r-prabu (approved):

PR #6819, opened 2025-05-13
Add irqlock status to the irq_procfs and ps command output [ Under Verification ]

neel-samsung:


samsung-agnel (approved):

ritesh55555 (commented):
added irqlock status in ps command logs

sunghan-chang (commented):
Is it possible to know each thread irq status as well?

sunghan-chang:
@neel-samsung Could you add sched lock status on the ps with new PR?

neel-samsung:
> Is it possible to know each thread irq status as well?

Let me check and get back to you.

build/configs/rtl8730e/audio/defconfig
```diff
@@ -1269,6 +1269,7 @@ CONFIG_ARCH_HAVE_HEAPCHECK=y
 CONFIG_DEBUG_MM_HEAPINFO=y
 # CONFIG_DEBUG_CHECK_FRAGMENTATION is not set
 # CONFIG_DEBUG_IRQ is not set
+CONFIG_DEBUG_IRQ_INFO=y
```
kishore-sn, line 1272: Is it ok to set CONFIG_DEBUG_IRQ_INFO=y without setting CONFIG_DEBUG_IRQ ?

neel-samsung:
> > Is it possible to know each thread irq status as well?
> 


> > Is it possible to know each thread irq status as well?
> 
> @sunghan-chang I have checked it. We have `irqcount` variable in tcb struct but is only enabled under SMP case. So, we don't keep track of it in No SMP case. Can we just print `irqcount` in smp case and `NA` for No SMP case. As, we will use ps cmd mostly in smp case to debug ?

Please correct me if I am wrong and Let me know your opinion.

sunghan-chang:
> > Is it possible to know each thread irq status as well?
> 
> @sunghan-chang I have checked it. We have `irqcount` variable in tcb struct but is only enabled under SMP case. So, we don't keep track of it in No SMP case. Can we just print `irqcount` in smp case and `NA` for No SMP case. As, we will use ps cmd mostly in smp case to debug ?

Can we apply it in no smp case as well?

sunghan-chang:
@neel-samsung If it has any reason why they should be in dependency, please let me know.

neel-samsung:
> @neel-samsung If it has any reason why they should be in dependency, please let me know.

No, there is not any dependency of SMP on IRQCOUNT.
Please review the PR for the same: https://github.com/Samsung/TizenRT/pull/6838

samsung-singh (approved):
Added irqlock status to the irq_procfs and ps command output, this change is good to merge.

os/fs/procfs/fs_procfs.c
```diff
@@ -140,8 +140,9 @@ static const struct procfs_entry_s g_procfsentries[] = {
 	{"fs/smartfs**", &smartfs_procfsoperations},
 #endif
 
-#if defined(CONFIG_DEBUG_IRQ_INFO) && !defined(CONFIG_FS_PROCFS_EXCLUDE_IRQS)
-	{"irqs", &irqs_operations},
+#ifndef CONFIG_FS_PROCFS_EXCLUDE_IRQS
```
r-prabu, line 143: why this condition compilation is modified to remove the checking of  CONFIG_DEBUG_IRQ_INFO? With this commit is CONFIG_DEBUG_IRQ_INFO not required?

os/kernel/irq/irq_procfs.c
```diff
@@ -139,18 +172,77 @@ const struct procfs_operations irqs_operations = {
...
+ *   Analyse relpath to find the directory reference entry it represents,
+ *   if any.
+ *
+ ****************************************************************************/
 
+static int irqs_find_dirref(FAR const char *relpath, FAR struct irqs_dir_s *dir)
+{
+	const char *str;
+	/* Function to check path is starting with given name */
+	int checkStart(char *name, bool isDirectory) {
```
r-prabu, line 199: Not a good practice to define a function inside a function. better to move the "checkStart" function outside this function scope and call here/

PR #6821, opened 2025-05-19
os/pm: prevent PM if cpu is already paused

lhenry-realtek:
In an unlikely case where secondary CPU was already paused by another operation, PM will hang as the pause request was already handled, so up_cpu_pausereq is 0

An additional preventive check is added to ensure the secondary cpu is actually paused, so that PM should be aborted until the secondary cpu is resumed to re-enter idle task

While checking #6815 case of Pause CPU1 then Abort on CPU0 in PM, this case was caught where the board hang due to infinite loop. Normally this case should not happen, but in the event that it does due to improper API use, this fix should help abort PM and fix the hang @kishore-sn @sunghan-chang

### DIFF
--- a/os/pm/pm_idle.c
+++ b/os/pm/pm_idle.c
@@ -137,6 +137,9 @@ void pm_idle(void)
 						pmdbg("Sleep abort! CPU%d\n", cpu);
 						goto EXIT;
 					}
+
+					/* If the target cpu is already paused it will not be in idle thread, so PM should abort */
+					ASSERT(!up_is_cpu_paused(cpu));
 				}
 
 				tcb = current_task(cpu);


sunghan-chang:
@ewoodev @kishore-sn @neel-samsung Could you review this ASAP?

os/pm/pm_idle.c
```diff
@@ -137,6 +137,12 @@ void pm_idle(void)
 						pmdbg("Sleep abort! CPU%d\n", cpu);
 						goto EXIT;
 					}
+
+					/* If the target cpu is already paused it will not be in idle thread, so PM should abort */
+					if (up_is_cpu_paused(cpu)) {
+						pmdbg("Sleep abort! CPU%d is paused\n", cpu);
+						goto EXIT;
+					}
```
ewoodev, line 145: I think, `ASSERT(!up_is_cpu_paused(cpu))` is better.

enter_critical_section() is required for cpu pause
so if enter_critical_section and cpu pause are called and this checking here, the system is already corruped

kishore-sn:
Hi @lhenry-realtek , 
I think we must handle this in a slightly different manner. 
If a cpu has already been paused, then we dont need to do the gating. 
For this I raised a different PR: https://github.com/Samsung/TizenRT/pull/6822
Please check it and let me know your opinion on it. If we take the above pr, then your current pr might not be required. 
Can you please test your current scenario with my PR?

PR #6822, opened 2025-05-19
arch/armv7-a: Change cpu gating logic when cpu is already paused

kishore-sn:
If this cpu has already been paused, then we
will not perform gating. However, we will set
the gating flag as if gating is performed. This
is because, the main intention of both gating and
pausing the cpu is to prevent the cpu from running. Since this is already handled and cpu is in pause state, we can perform critical opertion assuming gating is done. At a future point of time, the cpu will get resumed by the code which had initially paused it

NOTE: This only works for 2 cpu case, in case of more cpus, we need to redesign the pause and gating logic such that only the cpu or the task which called pause is allowed to call resume.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_cpugating.c
+++ b/os/arch/arm/src/armv7-a/arm_cpugating.c
@@ -115,6 +115,33 @@ void up_cpu_gating(int cpu)
 {
 	DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
 
+	/* If this cpu has already been paused, then we 
+	 * will not perform gating. However, we will set
+	 * the gating flag as if gating is performed. This
+	 * is because, the main intention of both gating and 
+	 * pausing the cpu is to prevent the cpu from running.
+	 * Since this is already handled and cpu is in pause state,
+	 * we can perform critical operation assuming gating is done.
+	 * At a future point of time, the cpu will get resumed
+	 * by the code which had initially paused it
+	 */
+
+	/* NOTE: This only works for 2 cpu case, in case of more cpus, 
+	 * we need to redesign the pause and gating logic such that only
+	 * the cpu or the task which called pause is allowed to call resume.
+	 */
+
+	if (up_is_cpu_paused(cpu)) {
+		g_cpugating_flag[cpu] = 2;
+		return;
+	} else if (up_cpu_pausereq(cpu)) {
+		/* On the other hand, if a pause request is pending, it 
+		 * has to be handled first and then the caller must retry
+		 * gating request
+		 */
+		return;
+	}
+
 	/* Fire SGI for cpu to enter gating */
 	arm_cpu_sgi(GIC_IRQ_SGI3, (1 << cpu));
 }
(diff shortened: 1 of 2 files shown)


kripal-samsung (approved):

jeongarmy (approved):
LGTM

os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -115,6 +115,27 @@ void up_cpu_gating(int cpu)
 {
 	DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
 
+	/* If this cpu has already been paused, then we 
+	 * will not perform gating. However, we will set
+	 * the gating flag as if gating is performed. This
+	 * is because, the main intention of both gating and 
+	 * pausing the cpu is to prevent the cpu from running.
+	 * Since this is already handled and cpu is in pause state,
+	 * we can perform critical opertion assuming gating is done.
```
sunghan-chang, line 124: typo
operation

sunghan-chang:
@ewoodev Could you review and leave the approval?

os/arch/arm/src/armv7-a/arm_cpugating.c
```diff
@@ -115,6 +115,27 @@ void up_cpu_gating(int cpu)
...
+	 * At a future point of time, the cpu will get resumed
+	 * by the code which had initially paused it
+	 */
+
+	/* NOTE: This only works for 2 cpu case, in case of more cpus, 
+	 * we need to redesign the pause and gating logic such that only
+	 * the cpu or the task which called pause is allowed to call resume.
+	 */
+
+	if (up_is_cpu_paused(cpu)) {
```
ewoodev, line 134: don't we need up_cpu_pausereq(cpu)?

Sambhavi-1 (approved):
changes for not gating cpu if it is paused looks ok

os/arch/arm/src/armv7-a/arm_cpugating.c
kishore-sn: Added. Thanks.

kishore-sn:
@lhenry-realtek 
I have a question. 
In this PR, if cpu is already paused, then we set gating flag to 2 and return from the gating api, without performing any gating operation. 
But, in os/arch/arm/src/amebasmart/amebasmart_smp.c, we have an additional check for gating as below. 

`while ((up_get_gating_flag_status(ulCoreID) == 1) || CA32_GET_STANDBYWFE(ca32->CA32_C0_CPU_STATUS) != BIT(ulCoreID)) {`

I think the second condition in the while loop will cause the device to hang if we take this PR. 
Is the second condition required?

kishore-sn:
> @lhenry-realtek I have a question. In this PR, if cpu is already paused, then we set gating flag to 2 and return from the gating api, without performing any gating operation. But, in os/arch/arm/src/amebasmart/amebasmart_smp.c, we have an additional check for gating as below.
> 
> `while ((up_get_gating_flag_status(ulCoreID) == 1) || CA32_GET_STANDBYWFE(ca32->CA32_C0_CPU_STATUS) != BIT(ulCoreID)) {`
> 
> I think the second condition in the while loop will cause the device to hang if we take this PR. Is the second condition required?

I added check for cpu paused in the while loop to prevent hang.

jeongarmy (approved):
lgtm

r-prabu (commented):
cpu gatting changes when the cpu is already paused are good.

PR #6825, opened 2025-05-19
build/tools/amebasmart/gnu_utility: Modify AES GCM buffer allocation

ziyik:
-Modify AES GCM handle input and output method
-Do not allocate AES GCM input and output, use pass-in buffer


sunghan-chang:
@jylee9613 Could you review this?
@ziyik Sorry but I can't get the point exactly. Could you elaborate it more? Why do you make this PR? How is the usages changed?

ziyik:
> @jylee9613 Could you review this? @ziyik Sorry but I can't get the point exactly. Could you elaborate it more? Why do you make this PR? How is the usages changed?

Prior to this PR, AES GCM would allocate 2 secure buffers to handle encryption or decryption. These buffers would hold the plaintext and ciphertext during encryption or decryption.
It would copy the data from the non-secure incoming buffers and save it. Once encryption or decryption is complete, the output is copied to the incoming buffers.
This PR removes the allocation of secure buffers and SE will use the incoming buffers directly.

sunghan-chang:
> > @jylee9613 Could you review this? @ziyik Sorry but I can't get the point exactly. Could you elaborate it more? Why do you make this PR? How is the usages changed?
> 
> Prior to this PR, AES GCM would allocate 2 secure buffers to handle encryption or decryption. These buffers would hold the plaintext and ciphertext during encryption or decryption. It would copy the data from the non-secure incoming buffers and save it. Once encryption or decryption is complete, the output is copied to the incoming buffers. This PR removes the allocation of secure buffers and SE will use the incoming buffers directly.

Could you write it in the commit description?

build/tools/amebasmart/gnu_utility/bl2_bl32.bin
jylee9613, line 1: Q1. bl2_bl32.bin was removed last year. Is it needed??
Q2. As I understood, this PR uses incoming input/output buffer directly in TZ. Then don't we need to change our TizenRT public API's input/output structure? If it is needed, please let me know.

ziyik: Q1. Yes, it's not needed. In [5918](https://github.com/Samsung/TizenRT/pull/5918), we explained the cause.
Q2. There is no need to modify TizenRT public API's input/output structure, original structure already passed in non-secure data buffer.

jylee9613: Thank you for answer my question! Then please remove the bl2_bl32.bin

sunghan-chang:
@jylee9613 Review is done?

jylee9613 (approved):

sunghan-chang (approved):

PR #6832, opened 2025-05-23
tools/trap: Optimize trap tool

samsung-agnel:
Removes unnecessary calls to nm command which were adding 4+ mins to execution time. The output from these calls are not used nor is there any effect on execution.


sunghan-chang (approved):

r-prabu (commented):
Optimization changes are fine for the trap tool.

PR #6834, opened 2025-05-26
os/pm: add pm_wakehandler to sleep ops

ritesh55555:
This commit adds pm_wakehandler api to sleep ops structure of PM. Previously pm_wakehandler was passed as a functional pointer to bsp sleep api. But now it is added as a parameter for sleep ops structure.

### DIFF
--- a/os/include/tinyara/pm/pm.h
+++ b/os/include/tinyara/pm/pm.h
@@ -277,8 +277,9 @@ struct pm_callback_s {
  */
 
 struct pm_sleep_ops {
-	void (*sleep)(pm_wakehandler_t handler);
+	void (*sleep)(void);
 	void (*set_timer)(unsigned int delay_us);
+	void (*wakehandler)(clock_t missing_tick, pm_wakeup_reason_code_t wakeup_src);
 };
 
 /*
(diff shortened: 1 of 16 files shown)


os/include/tinyara/pm/pm.h
```diff
@@ -277,8 +277,9 @@ struct pm_callback_s {
  */
 
 struct pm_sleep_ops {
-	void (*sleep)(pm_wakehandler_t handler);
+	void (*sleep)(void);
 	void (*set_timer)(unsigned int delay_us);
+	void (*wakehandler)(clock_t missing_tick, pm_wakeup_reason_code_t wakeup_src);
```
neel-samsung, line 282: Please add the functioning of  newly added call backs to above comment description.
```
/*
 * struct pm_sleep_ops - Callbacks for managing platform dependent PM power gating operations.
 *
 * @sleep: It makes the board enter the system sleep state. It also
 *  performs the necessary operation before sleep and immediately after
 *  wake up. This callback is mandatory.
 *
 * @set_timer: Set the wakeup timer
 *
 * @wakehandler :  .......................
 */

ewoodev:
we can remove wakeup handler when we organize enter PG flow.
if we can control secondery core enabling/disable, we can make wakeup flow in os side.
we don't need to provide callback address to BSP for recovery OS feature when wakeup.

PR #6826, opened 2025-05-21
os/pm: remove CPU gating from pm_idle

ritesh55555:
We are already inside critical section in pm_idle. So, there won't be any chance of context switching on other core. Hence, this patch removes the CPU gating from pm_idle.

PR #6835, opened 2025-05-27
os/arch/arm/src/amebasmart: fix race condition in systick ISR when PM

lhenry-realtek:
During PM, there is a race condition where pm is about to hotplug other CPU in critical section and then other CPU tick ISR triggers. Due to sched_process_loop in the tick ISR, it may re-enter critical section and acquire a spinlock that will never be released in this state

This commit adds a flag to let the target CPU disable its own timer and prevent further tick ISRs from entering scheduler loop when hotplug is about to fire. Because hotplug is effectively turning CPU offline, the PM resume boot flow will re-enable the timer and then lost ticks will be compensated.

This commit addresses the hang issue for #6826 for gating removal in pm_idle

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_timerisr.c
@@ -34,6 +34,7 @@
 #include "up_internal.h"
 #include "gic.h"
 #include "arch_timer.h"
+#include "barriers.h"
 
 /*----------------------------------------------------------------------------
   Clock Variable definitions
@@ -59,6 +60,29 @@
 #if SYSTICK_RELOAD > 0x00ffffff
 #  error SYSTICK_RELOAD exceeds the range of the RELOAD register
 #endif
+
+#ifdef CONFIG_SMP
+static volatile u8 systimer_status[CONFIG_SMP_NCPUS];
+
+void up_systimer_pause(u8 cpu)
+{
+  systimer_status[cpu] = 1;
+  ARM_DSB();
+}
+
+void up_systimer_resume(u8 cpu)
+{
+  systimer_status[cpu] = 0;
+  ARM_DSB();
+}
+
+u8 up_systimer_is_paused(u8 cpu)
+{
+  return systimer_status[cpu];
+}
+#endif
+
+void up_timer_disable(void);
 
 /****************************************************************************
  * Private Functions
@@ -91,6 +115,24 @@ int up_timerisr(int irq, uint32_t *regs)
 	  delta_ticks = 1;
 #endif
 
+
+#ifdef CONFIG_SMP
+    /* 
+       If this CPU is about to be hotplugged and a tick ISR is still triggering, prevent it from entering any critical sections.
+       
+       A deadlock condition may otherwise occur when pm_idle is processing and has entered critical section on CPU0
+       but when CPU1 tick ISR fires during this time, the scheduler processing loop (sched_process_timer())
+       tries to acquire a spinlock and fails
+
+       This behavior is easily reproducible when performing many HW TIMER sleeps in sequence (e.g LCD) and may be
+       classified as a race condition, since there are often several sleep-wakeup successful iterations before deadlock happens
+    */
+    if (up_systimer_is_paused(up_cpu_index())) {
+      up_timer_disable();
+      goto skip_sched;
+    }
+#endif
+
     u32 ticks_to_process = delta_ticks;
     while (ticks_to_process > 0) {
       /* process missing ticks */
@@ -98,6 +140,7 @@ int up_timerisr(int irq, uint32_t *regs)
       ticks_to_process--;
     }
 
+skip_sched:
     arm_arch_timer_set_compare(last_cycle + delta_ticks * SYSTICK_RELOAD);
     return 0;
 }
@@ -123,9 +166,16 @@ void up_timer_initialize(void)
   /* Attach the timer interrupt vector */
   irq_attach(ARM_ARCH_TIMER_IRQ, (xcpt_t)up_timerisr, NULL);
 
-  //arm_arch_timer_count
-	arm_arch_timer_set_compare(arm_arch_timer_count() + SYSTICK_RELOAD);
-	arm_arch_timer_enable(1);
+#ifdef CONFIG_SMP
+  /* reset the hotplug status bit on startup */
+  up_systimer_resume(up_cpu_index());
+#endif
+
+  /* Only enable the timer for CPU0 on startup, CPU1's will be enabled when pause/gating takes place  */
+  if (up_cpu_index() == 0) {
+    arm_arch_timer_set_compare(arm_arch_timer_count() + SYSTICK_RELOAD);
+    arm_arch_timer_enable(1);
+  }
 
   /* And enable the timer interrupt at the GIC */
   up_prioritize_irq(ARM_ARCH_TIMER_IRQ, 224);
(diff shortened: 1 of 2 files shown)


sunghan-chang:
@kishore-sn Could you review this?

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -60,6 +60,22 @@
...
+void up_timer_set_hotplug_status(u8 cpu, u8 status)
+{
+  timer_hotplug_status[cpu] = status;
+}
+
+u8 up_timer_get_hotplug_status(u8 cpu)
+{
+  return timer_hotplug_status[cpu];
+}
+#endif
```
kishore-sn, line 83: When we are using global variables with smp, would it be better to add memory barriers in the set and get apis?

lhenry-realtek: Yes, it will be added to align with other APIs like hotplug

kishore-sn (approved):

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -123,9 +157,16 @@ void up_timer_initialize(void)
...
+#ifdef CONFIG_SMP
+  /* reset the hotplug status bit on startup */
+  up_timer_set_hotplug_status(up_cpu_index(), 0);
+#endif
+
+  /* Only enable the timer for CPU0 on startup, CPU1's will be enabled when pause/gating takes place  */
+  if (up_cpu_index() == 0) {
+    arm_arch_timer_set_compare(arm_arch_timer_count() + SYSTICK_RELOAD);
+    arm_arch_timer_enable(1);
+  }
```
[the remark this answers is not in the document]
eunwoo-nam: Hello @lhenry-realtek 
Do you plan merge this PR first? and will you fix rootcouse of pop up cpu1 tick irq?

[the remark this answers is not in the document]
lhenry-realtek: Hi @ewoodev 

As discussed with dane, this is a corner case where the tick ISR can run before SGI3 hotplug

Based on https://nuttx.apache.org/docs/latest/implementation/critical_sections.html#smp-critical-sections, it is possible for CPU1 to still be running even though CPU0 has already entered critical section. There is a tiny amount of time that the tick ISR on CPU1 can still fire (since it happens in microseconds) before SGI3 comes in

Next, the tick IRQ priority is configured as 224 
https://github.com/Samsung/TizenRT/blob/75216f58663eabe20747c015054e2f3043105a28/os/arch/arm/src/amebasmart/amebasmart_timerisr.c#L131
while SGI3 priority is configured as 128 https://github.com/Samsung/TizenRT/blob/75216f58663eabe20747c015054e2f3043105a28/os/arch/arm/src/armv7-a/arm_gicv2.c#L172

This means that in the event SGI3 comes in, then tick IRQ triggers, tick IRQ will preempt SGI3 and cause deadlock

There are 2 possible scenario:

1. hotplug CPU0
2. CPU0 enter critical
3. CPU0 send SGI3 to CPU1
4. CPU1 receives SGI3 and runs normally -> no issue, CPU1 becomes hotplugged
OR
5. CPU1 receives SGI3 first, but then tick ISR comes in, so tick ISR runs first before SGI3 -> deadlock

As long as CPU1 is prevented from entering critical section inside sched_process_timer for the short time until SGI3 completes on CPU1, the deadlock condition can be avoided with this PR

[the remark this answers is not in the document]
eunwoo-nam: thank you for explain~ 👍👍

os/pm/pm_idle.c
```diff
@@ -164,6 +164,8 @@ void pm_idle(void)
 		/* Send signal to shutdown other cores here */
 		for (cpu = 1; cpu < CONFIG_SMP_NCPUS; cpu++) {
 			if (up_get_cpu_state(cpu) == CPU_RUNNING) {
+				/* Ensure that secondary CPU systick PPI stops firing to prevent deadlock condition */
+				up_timer_set_hotplug_status(cpu, 1);
```
eunwoo-nam: ✔ The up_timer_set_hotplug_status is never called without up_cpu_hotplug(). 
✔ So, we don't need to call sepeatliy up_timer_set_hotplug_status and up_cpu_hotplug. 

✨How about move calling up_timer_set_hotplug_status into up_cpu_hotplug()? 👀

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
```diff
@@ -60,6 +61,23 @@
 #  error SYSTICK_RELOAD exceeds the range of the RELOAD register
 #endif
 
+#ifdef CONFIG_SMP
+static volatile u8 timer_hotplug_status[CONFIG_SMP_NCPUS];
+
+void up_timer_set_hotplug_status(u8 cpu, u8 status)
```
eunwoo-nam: The actual behavior of this function is that it does not execute the OS timer handler even when the timer triggers.
When viewed within this file, the function correctly sets the state of the timer_hotplug_status[cpu] variable, so the function name up_timer_set_hotplug_status is appropriate. 👍
However, from the OS perspective, this API is called with the intention of stopping or pausing the timer. ✨

From the OS viewpoint, the API name should be designed to be universally applicable across multiple chipsets. Therefore, How about rename this API to something like up_timer_pause?. 😊

lhenry-realtek: Sure it will be renamed to fit the style

lhenry-realtek: As this is a getter/setter, proposed name is up_timer_set_pause / get_pause

os/pm/pm_idle.c
lhenry-realtek: The original reason why we put it inside amebasmart_timerisr is because it is chip-specific change to do the timer toggling

Do you think other chipsets may also require this feature

eunwoo-nam: I don't know the other chipset yet. 😥
but the calling up_timer_set_hotplug_status is pm_idle of OS layer. 

so, if the function should be called in os layer, in up_cpu_hotplug() is better. 😊

If you want to limit this to within the amebasmart code, we have to call bsp api in up_cpu_hotplug and call this in it.

In the current structure, one should be added as bsp api calling in os side.
The position is timer pause due to hotplug, so the hotplug function would be good.

lhenry-realtek: Okay, in that case let's move it inside, in future if required to move it back into bsp-specific then we make a separate commit

os/arch/arm/src/amebasmart/amebasmart_timerisr.c
eunwoo-nam: Thank you for your contribution. 😊
The text I reviewed may pertain to the implementation of the OS HAL API.

In Tizen Lite,  the use of up_timer_pause() / up_timer_resume() format. is more common compared than up_timer_set_pause(0, 1). ✨

If you find it burdensome to modify the OS HAL API content, I will proceed with this in a subsequent PR.

lhenry-realtek: There is one more ambiguity, that is systick timer and regular timer may be confused due to naming, are there any recommended naming? up_pause_systimer?

eunwoo-nam: up_pause_systimer looks good 👍👍

jeongarmy (approved):
lgtm

PR #6836, opened 2025-05-27
Add littlefs corrupt

gidori98:


apps/system/utils/fscmd.c
```diff
@@ -1419,67 +1425,152 @@ static int tash_format(int argc, char **args)
...
+	if (argc >= 2 && strncmp(args[1], "internal", strlen(args[1]) + 1) == 0) {
+		ret = corrupt_filesystem(FS_BLOCK_MINOR_PRIMARY);
+	} else if (argc >= 2 && strncmp(args[1], "external", strlen(args[1]) + 1) == 0) {
+		ret = corrupt_filesystem(FS_BLOCK_MINOR_SECONDARY);
+	} else {
+		printf("Usage: corrupt [internal | external] \n");
+		ret = ERROR;
+	}
+#else
+	printf("BCH should be enabled to format file system\n");
```
Taejun-Kwon, line 1501: 1. below has added in line 1474 already, hence I think below condition can wrap entire of corrupt + format functions.
```
#ifndef CONFIG_DISABLE_ENVIRON
#if !defined(CONFIG_DISABLE_MOUNTPOINT) && \
	!defined(CONFIG_DISABLE_PSEUDOFS_OPERATIONS) && defined(CONFIG_BCH)
```
2. then #else part can be removed.

os/fs/driver/mtd/little.c
```diff
@@ -258,6 +258,13 @@ static int little_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 			fdbg("Update Format Info Finished. after reboot, fs will be formatted\n");
 		}
 		break;
+	case BIOC_CORRUPTION:
+		fdbg("Update Corrupt Info started\n");
+		ret = lfs_reserve_corrupt(dev->lfs);
+		if (ret == OK) {
+			fdbg("Update Corrupt Info Finished. after reboot, fs will be formatted\n");
```
Taejun-Kwon, line 265: fs will be formatted -> mount will be failed

os/fs/driver/mtd/little.c
```diff
@@ -258,6 +258,13 @@ static int little_ioctl(FAR struct inode *inode, int cmd, unsigned long arg)
 			fdbg("Update Format Info Finished. after reboot, fs will be formatted\n");
 		}
 		break;
+	case BIOC_CORRUPTION:
+		fdbg("Update Corrupt Info started\n");
```
Taejun-Kwon, line 262: 'Corrupt super block started' is better option IMO

os/fs/littlefs/lfs_vfs.c
```diff
@@ -901,6 +903,10 @@ static int littlefs_bind(FAR struct inode *driver, FAR const void *data, FAR voi
 	if (ret < 0) {
 		/* Auto format the device if -o autoformat */
 
+		if (ret == LFS_ERR_SIG) {
+			return LFS_ERR_SIG;
```
Taejun-Kwon, line 907: I think goto errout_with_fs is correct. If it return here, then state is sem_wait, thread will wait signal forever.

os/fs/littlefs/littlefs/lfs.h
```diff
@@ -104,6 +104,7 @@ enum lfs_error {
     LFS_ERR_NOMEM       = -12,  // No more memory available
     LFS_ERR_NOATTR      = -61,  // No data/attr available
     LFS_ERR_NAMETOOLONG = -36,  // File name too long
+	LFS_ERR_SIG = -10,		// Signiture crpt__fs
```
Taejun-Kwon, line 107: so mount is failed because super block has corrupted. then I think LFS_ERR_CORRUPT is enough to handle it.

Taejun-Kwon (commented):
you should not apply coding rule for lfs.c & lfs.h
Because if we apply our coding rule, then it is really hard to update code from littlefs git.
Until it become stable & and until we fully understand littlefs, let's keep their coding rule.

os/fs/littlefs/lfs_vfs.c
gidori98: At first, I did it you commented, But If use "goto errout_with_fs", semaphore of fs and fs are free. So, when I enter command "format external(littlefs)", can not access fs. finally occur assert.

PR #6840, opened 2025-05-28
os/board/rtl8730e/src: replace gating with pause in flash op

lhenry-realtek:
This commit removes gating in place of using pause to halt other cores while XIP is taking place

The main benefit is to reduce complexities of handling both pause and gating at the same time, along with being more portable

Since up_cpu_pause_all can also pause core0, this means that unlike gating, pause is more flexible and that write operations can be done on both Core0 & Core1

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_smp.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_smp.c
@@ -73,48 +73,6 @@ void rtk_core1_power_off(void)
 	HAL_WRITE32(SYSTEM_CTRL_BASE_HP, REG_HSYS_HP_PWC, val);
 }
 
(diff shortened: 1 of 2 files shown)


kishore-sn (approved):

sunghan-chang:
> we are verfing flash operartion. if verifiication is done on realtek side, pleare remove label "do not merge"

@lhenry-realtek Could you check it ASAP?

sunghan-chang:
> > we are verfing flash operartion. if verifiication is done on realtek side, pleare remove label "do not merge"
> 
> @lhenry-realtek Could you check it ASAP?

@lhenry-realtek Oh, already removed. Thank you!

PR #6841, opened 2025-05-28
dbuild, build/configs: Add device selection menu

ewoodev:
When we download, we either predefined a default port for each board or manually typed commands like ./dbuild.sh download port=ttyACM0 all.

However, depending on the connection state of the computer, the board might be detected on a port other than the default one.

Therefore, instead of hardcoding the default port in the script, we will modify it to display and allow the user to select the connected device port every time during the download process.

like below:
```
==================================================
  "Port Selected:"
==================================================
  "1. ttyACM0"
  "2. ttyUSB0"
  "3. ttyUSB1"
  "x. Exit"
==================================================
2
```

### DIFF
--- a/build/configs/common_download.sh
+++ b/build/configs/common_download.sh
@@ -46,9 +46,15 @@ if [ "$1" == "port" ]; then
 	PORT="$2"
 	shift 2
 else
+	if [ -f "${OS_PATH}/.download" ]; then
+		source "${OS_PATH}/.download"
+	fi
+
 	echo ==================================================
-	echo "  \"Port Selected:\""
+	echo "  \"Port Selected: [$PREV_PORT]\""
 	echo ==================================================
+	echo "  *If you keep this port, please just enter"
+	echo ""
 	devices=()
 	count=1
 	for dev in /dev/ttyACM* /dev/ttyUSB*; do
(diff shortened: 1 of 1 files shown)


os/dbuild.sh
```diff
@@ -465,7 +465,7 @@ function DOWNLOAD()
 {
 	# Currently supports ALL only, later this will have a menu
 	pushd ${OSDIR} > /dev/null
-	docker run --rm ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged ${DOCKER_IMAGE}:${DOCKER_VERSION} ${BUILD_CMD} download $1 $2 $3 $4 $5 $6
+	docker run --rm -it ${DOCKER_OPT} -v ${TOPDIR}:/root/tizenrt -w /root/tizenrt/os --privileged ${DOCKER_IMAGE}:${DOCKER_VERSION} ${BUILD_CMD} download $1 $2 $3 $4 $5 $6
```
jeongarmy, line 468: Is there any reason to add '-it' option for download?
I think it is not necessary because there is no input after running docker container.

ewoodev: common_download.sh is called in docker
In this change, I am trying to select a port by receiving user input within common_download.sh

jeongarmy: Yes, you're right. I missed.

jeongarmy (approved):
LGTM

jeongarmy (approved):
Looks good

PR #6845, opened 2025-05-30
external/onert-micro: updated to version 2.0

vibhor-m:
framework/aifw: OneRTM.cpp updated files to use version 2.0 namespace and classes.


pillip8282 (commented):
the number of files is too many. all these files are required?
It can degrade git performance.

vibhor-m:
> the number of files is too many. all these files are required? It can degrade git performance.

OneRT-Mirco 2.0 has around 844 files. This PR also including deleting files on OneRT-Micro 1.0.

PR #6846, opened 2025-05-30
build/configs: Add batch download layer for binary files

ewoodev:


lhenry-realtek:
@ewoodev 
is it expected behavior for bootparam to be flashed even though it is not specified on the command line

e.g "download kernel" expect only kernel flashed, but both kernel and bootparam are flashed together

build/configs/rtl8730e/rtl8730e_download.sh
```diff
@@ -93,101 +88,79 @@ function pre_download()
...
-			echo "Start USB $4 erase in Flash"
-			echo "==================================="
-			./upload_image_tool_linux "erase" $1 $2 $3
-			echo "Complete USB download"
-		else
-			return;
-		fi
-	fi
+		shift 3
+	done
```
seokhun-eom24, line 180: I think add echos like `board_download_batch()` would be also great for symmetry.
```
echo ""
echo "==================================="
echo "Start USB erase in Flash"
echo "==================================="
```

PR #6849, opened 2025-06-02
Optimize mm_free_delaylist calls and add background processing

priyanshscpp:
# Memory Management Optimization: Delay List Handling

This update introduces significant enhancements to the memory management subsystem, targeting performance optimization and long-term memory safety.

---

### 1. Optimized `mm_free_delaylist` Calls

- The `mm_malloc` function now calls `mm_free_delaylist` **only** when a memory allocation fails.
- This change avoids unnecessary delay list flushing during successful allocations.
- Result: **Reduced overhead** and improved performance for fast-path memory operations.

### 2. Optional Background Delay List Cleanup

- A new configurable feature, enabled via `CONFIG_MM_DELAYED_FREE_WORKER`, introduces a background worker thread to handle delayed frees.
- The worker periodically flushes the delay list, even when memory allocations are infrequent.
- This helps prevent **unbounded growth** of the delay list in scenarios with frequent deallocations in critical contexts.

---

## ⚙️ Configuration

To enable the background cleanup worker, add the following to your configuration:

```c
#define CONFIG_MM_DELAYED_FREE_WORKER 1
@sanbeam @bhupathiranjith @kimusan @jkivilin

### DIFF
--- a/apps/examples/hello/README.txt
+++ b/apps/examples/hello/README.txt
@@ -1,7 +1,7 @@
 examples/hello
 ^^^^^^^^^^^^^^
 
-  This is the mandatory, "Hello, World!!" example.  It prints "Hello, World!!" third.
+  This is the mandatory, "Hello, World!!" example.  It prints "Hello, World!!" first.
   Really useful only for bringing up new TinyAra architectures.
 
   Configs (see the details on Kconfig):
--- a/os/mm/mm_heap/mm_initialize.c
+++ b/os/mm/mm_heap/mm_initialize.c
@@ -63,6 +63,11 @@
 
 #include <tinyara/sched.h>
 #include <tinyara/mm/mm.h>
+#include "mm_node.h" /* For mm_delayed_free_worker prototype */
+
+#ifdef CONFIG_MM_DELAYED_FREE_WORKER
+#include <tinyara/kthread.h>
+#endif
 
 /****************************************************************************
  * Pre-processor Definitions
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: os/mm/mm_heap/mm_initialize.c [hunk 1]
```diff
@@ -283,5 +288,23 @@ int mm_initialize(FAR struct mm_heap_s *heap, FAR void *heapstart, size_t heapsi
 	heapinfo_update_group_info(INVALID_PROCESS_ID, HEAPINFO_INVALID_GROUPID, HEAPINFO_INIT_INFO);
 #endif
 #endif
+
+#ifdef CONFIG_MM_DELAYED_FREE_WORKER
+	/* If the heap was initialized successfully, create the delayed free worker thread. */
+	if (ret == OK) {
+		int worker_pid;
+		worker_pid = kthread_create("mm_delayed_free", CONFIG_SCHED_PRIORITY_DEFAULT,
```
eunwoo-nam, line 296: We have two types of delay-free operations:

1. sched_free: This function allows the free operation to be delayed and executed in the LPWORK context if it is called in an IRQ context or if the mm lock is held.
2. free on isr: If the free operation is performed in an IRQ context, it is added to the mm_delaylist and subsequently freed during the next malloc operation.

The key difference between the two is that the first option explicitly delays the free operation if the mm lock is held, while the second option waits if the mm lock is acquired.

Since we are already utilizing lpwork for delayed free operations in the first scenario, it is necessary to evaluate whether creating a separate kernel thread for the second scenario is advisable. If the load on lpwork is not full, it appears sufficient to handle the second scenario using lpwork as well.

apps/examples/hello/README.txt
```diff
@@ -1,7 +1,7 @@
 examples/hello
 ^^^^^^^^^^^^^^
 
-  This is the mandatory, "Hello, World!!" example.  It prints "Hello, World!!" third.
+  This is the mandatory, "Hello, World!!" example.  It prints "Hello, World!!" first.
```
eunwoo-nam, line 4: This changes is not same agenda,
Let's sepeate commit.

ewoodev (commented):

You can refer to the commit rule on the link.

os/mm/mm_heap/mm_initialize.c
eunwoo-nam: I previously mentioned that there could be a possibility of delay-free failure due to the high load on LPWORK, but even if allocation failure occurs in such cases, since delay-free is performed at that specific moment.

So, How about integrate this with the existing LPWORK delay-free mechanism without creating an additional kernel thread?

you can find it in os/wqueue/kwqueue/kwork_lpthread.c

PR #6850, opened 2025-06-05
build/tools/amebasmart/image_tool: support download with new jig

zhongnuo-tang:
1. The USB-to-serial chip in new jig uses CH347F which display as ttyACM instead of ttyUSB (FTDI), so the condition to check for USB/UART cannot use dev port name.
2. Use Vendor ID to determine whether it is USB or UART download. VID for USB is '0bda'.
3. Add retry mechanism when reading VID because if chip is IWDG enabled, it might reboot while reading VID, and this will cause VID not able to read.
4. Proceed with USB download if port VID = '0bda' else force to use UART download.
4. map /run/udev into the docker container from the VM, so that docker container can read the external copy of the udev, to read for Vendor ID.


zhongnuo-tang:
Verify with both latest public and PR6846 (variable name USB_DOWNLOAD change to IS_BOARD_SUPPORTED_BATCH)
All is working for both UART/USB download.

ewoodev (approved):

zhongnuo-tang:
> Hello @zhongnuo-tang do this changes include iwdg retry?


the original retry in the linux tool is unchanged.
this PR include a new iwdg retry at download script reading for VID.
	
	while [ $VID_FOUND -ne 1 ] && [ $RETRY_COUNT -lt $MAX_RETRIES ]; do
		VENDOR_ID=$(udevadm info -a -n "$TTYDEV" | grep -m1 'ATTRS{idVendor}' | sed -nE 's/.*"([0-9a-fA-F]+)".*/\1/p')

		if [ -n "$VENDOR_ID" ]; then
			VID_FOUND=1
			echo "Found VID: $VENDOR_ID"
			break
		else
			echo "VID not found, retrying..."
			((RETRY_COUNT++))
			sleep 1
		fi
	done

jeongarmy (approved):
lgtm

PR #6853, opened 2025-06-10
os/drivers/lcd/: Add SW rotation for ST7785 LCD

namanjain7:
Since enabling HW rotation is creating issues in LCD, so this PR is enables SW rotation for this LCD.


os/drivers/lcd/mipi_lcd.c
```diff
@@ -71,7 +71,7 @@ static void lcd_rotate_buffer(short int* src, short int* dst)
 	short int *psrc;
 	short int *pdst;
 
-#if defined(CONFIG_LCD_LANDSCAPE)
```
Taejun-Kwon: This not allowed, this is kind of abstract layer, if any device support LANDSCAPE, then it should enable CONFIG_LCD_LANDSCAPE in defconfig. hence if you already enabled CONFIG_LCD_LANDSCAPE in above defconfig, you should not add CONFIG_LCD_ST7785 here.

namanjain7: Hello Sir,
The issue with ST7785 LCD is that it supports HW rotation, but when we enable HW rotation using CONFIG_LCD_LANDSCAPE config, then it shows tearing as LCD doesn't support Landscape in video mode, which is our current requirement.
Due to this, we need to forcefully enable SW rotation for this particular LCD while disabling Hardware rotation (Landscape mode).

namanjain7:
@ewoodev 
One way is this code that I implemented in this PR.
Other way is to make changes in st7785.h file, we will set Portrait mode in st7785.h while enabling CONFIG_LCD_LANDSCAPE globally.
So, it might create confusion as settings are different in st7785.h and global.

Other way to implement this is to make different config CONFIG_LCD_SW_LANDSCAPE_ROTATION. So, there will be two different configs for HW rotation and SW rotation, so globally it might create problems.

os/drivers/lcd/mipi_lcd.c
Taejun-Kwon: Then You know answer already.
1. Add some kind of CONFIG_LCD_SW_ROTATE to KConfig in os/driver/lcd
2. Enable CONFIG_LCD_SW_ROTATE  to def config.
3. Use CONFIG_LCD_SW_ROTATE  instead of CONFIG_LCD_ST7785

eunwoo-nam: the row cal is replaced in st7785.h if you enable CONFIG_LCD_LANDSCAPE.
it appres to better remove CONFIG_LCD_LANDSCAPE in st7785.h and enable SW ROTATE configs without this changes

I mean, if st7785 dosen't support hw rotate, let's remove hw rotate setting in st7785.h and use sw rotate

namanjain7:
Required changes done

namanjain7:
requested changes done.

os/include/tinyara/lcd/st7785.h
```diff
@@ -42,30 +42,17 @@
...
-#define LCD_CASET2 0x3F
-#define LCD_RASET1 0x00
-#define LCD_RASET2 0xEF
-#endif
 
-#if defined(CONFIG_LCD_PORTRAIT)
+#if defined(CONFIG_LCD_PORTRAIT) || defined(CONFIG_LCD_LANDSCAPE) || defined(CONFIG_LCD_RLANDSCAPE)
 #define LCD_ORIENTATION 0x00
 #elif defined(CONFIG_LCD_RPORTRAIT)
 #define LCD_ORIENTATION 0xc0
```
eunwoo-nam: It appears that you intend to maintain the RPORTRAIT HW rotation.
If CONFIG_LCD_RPORTRAIT=y and CONFIG_LCD_SW_ROTATION=y are applied in the current modification, it is likely that the display will be rotated twice, resulting in an output without rotation.

how about below? 
```suggestion
#if defined(CONFIG_LCD_SW_ROTATION) || defined(CONFIG_LCD_PORTRAIT)
#define LCD_ORIENTATION 0x00
#elif defined(CONFIG_LCD_RPORTRAIT)
#define LCD_ORIENTATION 0xc0
#else
#error Not support
#endif
```

eunwoo-nam: The `0xc0` has tearing.
The `0xc0` rotates the orientation to RPROTRAIT. however, since the fourth bit, which determines the data read direction, remains 0, tearing occurs.
Could you please confirm to 0xd0?

namanjain7: both 0xc0 and oxd0 has tearing.
should I remove the RPORTRAIT?

namanjain7: 0xD0 doesn't show any tearing. I have replaced 0xC0 with 0xD0 for RPORTRAIT

namanjain7: changed.

PR #6856, opened 2025-06-13
syu645b : Set equalizer using json file

gidori98:
Previously, the set_equalizer was set using the syu645bscripts.h file, but Now it was changed to be set using the json file.

### DIFF
--- a/os/drivers/audio/syu645b.c
+++ b/os/drivers/audio/syu645b.c
@@ -1067,7 +1067,6 @@ static int parse_equalizer_metadata_json(void){
 		ret = -EIO;
 		goto err_with_fd;
 	}
-	
 	gJSON = cJSON_Parse(buffer);
 	if (!gJSON) {
 		meddbg("Failed to parse equalizer json file.\n");
@@ -1091,8 +1090,7 @@ static int dec_to_hex_string(uint32_t decimal, char *hexadecimal) {
 		int mod = decimal % 16;
 		if (mod < 10) {
 			hexadecimal[pos] = 48 + mod;
-		}
-		else {
+		} else {
 			hexadecimal[pos] = 65 + (mod - 10);
 		}
 		decimal = decimal / 16;
@@ -1122,8 +1120,7 @@ static int string_to_script(char *string, t_codec_init_script_entry *script) {
 			if (ptr != end_ptr) {
 				if (cnt == 0) {
 					script->addr = val;
-				}
-				else {
+				} else {
 					script->val[cnt-1] = val;	
 				}
 				ptr = end_ptr;
@@ -1183,6 +1180,7 @@ static int syu645b_set_equalizer(FAR struct syu645b_dev_s *priv, uint32_t preset
 
 	/* Not find proper key, execute default script*/
 	if (!search){
+		cJSON_Delete(gJSON);
 		return -EINVAL;
 	}
 
@@ -1216,6 +1214,7 @@ static int syu645b_set_equalizer(FAR struct syu645b_dev_s *priv, uint32_t preset
 	if (!priv->mute) {
 		syu645b_setvolume(priv);
 	}
+	cJSON_Delete(gJSON);
 }
 
 /****************************************************************************


os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
 	syu645b_setmute(priv, true);
 }
 
+static int parse_equalizer_metadata_json(void){
+	struct stat jsonFileStat;
+	if (stat(PROD_EQUALIZE_ENCODE_JSON_PATH, &jsonFileStat) != 0) {
+		meddbg("Failed to fetch equalizer json file information.\n");
```
Taejun-Kwon: 1. driver uses auddbg.
2. When you use file operation, and if result is not OK then you should always print out errno.
```
auddbg("Failed to fetch json file : %d\n", errno);

File: os/drivers/audio/syu645b.c [hunk 2]
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		return -1;
+	}
+	char *buffer = (char *)malloc(jsonFileStat.st_size + 1);
+	if (!buffer) {
+		meddbg("Failed to allocate memory to hold equalizer json file content.\n");
```
Taejun-Kwon: Log displayed through the uart, and most of h/w we using has 16 bytes of uart.
Hence long length of log will make more interrupt.
'failed to alloc' is enough here.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	
+	if (!gJSON) {
+		meddbg("Failed to parse equalizer json file.\n");
+		close(fd);
+		free(buffer);
+		return -1;
+	}
+	close(fd);
+	free(buffer);
+	return 0;
```
Taejun-Kwon: please consider style as below.
```
	int ret = OK;
	char *buffer = (char *)malloc(jsonFileStat.st_size + 1);
	if (!buffer) {
		meddbg("Failed to allocate memory to hold equalizer json file content.\n");
		return ERROR;
	}
	int fd = open(PROD_EQUALIZE_ENCODE_JSON_PATH, O_RDONLY);
	if (fd == -1) {
		meddbg("Failed to open equalizer json.\n");
		goto err_with_buffer;
	}
	ssize_t bytesRead = read(fd, buffer, jsonFileStat.st_size);
	if (bytesRead == -1) {
		meddbg("Failed to read from equalizer json.\n");
		goto err_with_fd;
	}
	...
	return OK;
err_with_fd:
	close(fd);
err_with_buffer:
	free(buffer);
	return ret;
	
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
 	syu645b_setmute(priv, true);
 }
 
+static int parse_equalizer_metadata_json(void){
+	struct stat jsonFileStat;
+	if (stat(PROD_EQUALIZE_ENCODE_JSON_PATH, &jsonFileStat) != 0) {
+		meddbg("Failed to fetch equalizer json file information.\n");
+		return -1;
```
Taejun-Kwon: let's use OK or ERROR

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	}
+	int fd = open(PROD_EQUALIZE_ENCODE_JSON_PATH, O_RDONLY);
+	if (fd == -1) {
+		meddbg("Failed to open equalizer json.\n");
+		free(buffer);
+		return -1;
+	}
+	ssize_t bytesRead = read(fd, buffer, jsonFileStat.st_size);
+	if (bytesRead == -1) {
```
Taejun-Kwon: when you use read, you should check return value & requested size.
you should compare bytesRead & jsonFileStat.st_size.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
 	syu645b_setmute(priv, true);
 }
 
+static int parse_equalizer_metadata_json(void){
+	struct stat jsonFileStat;
```
Taejun-Kwon: naming rule of 'C' is 'do_what' or noun type of variable. and use lower character.
here you can define it as 'json_stat' or just simply 'st' or 'stat'

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	if (stat(PROD_EQUALIZE_ENCODE_JSON_PATH, &jsonFileStat) != 0) {
+		meddbg("Failed to fetch equalizer json file information.\n");
+		return -1;
+	}
+	char *buffer = (char *)malloc(jsonFileStat.st_size + 1);
```
Taejun-Kwon: if json file become long, this will consume huge size of memory.
We should use static buffer and handle this in loop.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+
+	/* Parse equlizer json to gJSON */
+	ret = parse_equalizer_metadata_json();
+	if (ret) {
+		printf("[TEST] goto default_script parse_equalizer_metadata_json\n");
+		goto default_script;
+	}
+
+	/* Convert preset hex to string for key */
+	char c_preset[11];
```
Taejun-Kwon: as possible as we can, we avoid to use constant value.
use define simiar to below
```
#define EQUALIZER_PRESET_LEN 11 //preset code use 8 digit. and 3more digit is required '0', 'x', '\0'
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	ret = dec_to_hex_string(preset, c_preset);
```
Taejun-Kwon, line 1169: You should consider which is more cheaper operation.
I think 'dec_to_hex_string' is more cheaper than parse_equalizer_metadata_json. so you should check dec_to_hex_string first. This approach is always need to be considered, even if you use if state as below.
if (cheaper || another)

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			syu645b_setvolume(priv);
+			return 0;
+		}
+	}
+	/* Not find proper key, execute default script*/
+	else {
+		goto default_script;
+	}
+
+default_script:
```
Taejun-Kwon: During init, we call set_equalizer(0);
I think if we provide function like a set_basic_equalizer, then set_equalizer(0) will not execute above logic.
Also we can remove 'goto default_script'
goto is very useful actually but to prevent human error, we should use very rarely.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	ret = dec_to_hex_string(preset, c_preset);
+	if (ret) {
+		goto default_script;
+	}
+
+	/* Executes given script through i2c to configure SYU645B device. */
+	cJSON *search = cJSON_GetObjectItemCaseSensitive(gJSON, c_preset);
+
+	/* Find proper key */
+	if (search){
```
Taejun-Kwon: Early return improve readability.
```
if (!search) {
	goto default_script;
}

cJSON *key;
t_codec_init_script_entry script;
...
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		cJSON *key;
+		t_codec_init_script_entry script;
+		
+		cJSON_ArrayForEach(key, search){
```
Taejun-Kwon: cJSON_ArrayForEach(key, search) {

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			if (strcmp(key->string, "_comment") == 0) {
+				printf("[TEST] find _comment\n");
+				continue;
+			}
+			printf("[TEST] script | %s : %s\n", key->string, key->valuestring);
```
Taejun-Kwon: auvdbg. DEBUG_AUDIO_VERBOSE disabled in default case.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		if (mod < 10) {
+			hexadecimal[pos] = 48 + mod;
+		}
+		else {
+			hexadecimal[pos] = 65 + (mod - 10);
+		}
+		decimal = decimal / 16;
+		pos--;
+
+		if (decimal == 0) {
```
Taejun-Kwon, line 1099:
```
if ((decimal == 0) && (pos > 1)) {
        for(;;)
}
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	int pos = 9;
+	hexadecimal[0] = '0';
+	hexadecimal[1] = 'x';
+	hexadecimal[10] = '\0';
+	while (1) {
+		int mod = decimal % 16;
+		if (mod < 10) {
+			hexadecimal[pos] = 48 + mod;
+		}
+		else {
```
Taejun-Kwon: coding rule

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			printf("[TEST] script | %s : %s\n", key->string, key->valuestring);
+
+			ret = string_to_script(key->valuestring, &script);
+			if (ret) {
+				goto default_script;
+			}
+
+			if (script.addr == 0x08 && script.type == SYU645B_REG_D_2BYTE) {
+				priv->max_volume = script.val[0];
+				printf("[TEST] priv->max_volume update %d \n", priv->max_volume);
```
Taejun-Kwon: change printf to audvdbg

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			}
+
+			reg[0] = script.addr;
```
Taejun-Kwon: I think you can remove addr, each of script line has address value in offset 0 of array.
also I think you can update syu645b_scrhipt.h

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+				printf("%x ", script.val[i]);
+			}
+			printf("\n");
+
+			ret = i2c_write(dev, syu645b_i2c_config, (uint8_t *)reg, script.type);
+			if (ret < script.type) {
+				auddbg("Error, cannot write to reg addr 0x%x, ret = %d\n", script.addr, ret);
+				break;
+			}
+			if (script.delay > 0) {
```
Taejun-Kwon: delay also not required.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	printf("[TEST] goto default script\n");
 	syu645b_exec_i2c_script(priv, t_codec_dq_preset_default_script, sizeof(t_codec_dq_preset_default_script) / sizeof(t_codec_init_script_entry));
 	priv->max_volume = 0x9F;
 
 	/* And then setVolume Again */
 	if (!priv->mute) {
 		syu645b_setvolume(priv);
+		return 0;
 	}
+	return 0;
```
Taejun-Kwon: this is void type function. you can change type to int.

os/drivers/audio/syu645b.c
Taejun-Kwon: or simply 
```
if (preset == 0) {
    goto default_script;
}
```

apps/examples/hello/hello_main.c
```diff
@@ -67,6 +67,14 @@ int main(int argc, FAR char *argv[])
 int hello_main(int argc, char *argv[])
 #endif
 {
-	printf("Hello, World!!\n");
+	if (argc != 3) {
+		char *preset = argv[1];
+		char *endptr;
+    	long result;
```
aadotverma: Please check indentation.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	ssize_t bytesRead = read(fd, buffer, jsonFileStat.st_size);
+	if (bytesRead == -1) {
+		meddbg("Failed to read from equalizer json.\n");
+		close(fd);
+		free(buffer);
+		return -1;
+	}
+	
+	gJSON = cJSON_Parse(buffer);
+
```
aadotverma: Please remove empty line before if

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		}
+		else {
+			hexadecimal[pos] = 65 + (mod - 10);
+		}
+		decimal = decimal / 16;
+		pos--;
+
+		if (decimal == 0) {
+			if (pos > 1) {
+				for(int i=2; i<=pos; i++){
```
aadotverma: for (int i = 2; i <= pos; i++) {

Apply coding rule

os/drivers/audio/syu645b.c
```diff
@@ -50,6 +53,10 @@
 #define BYTE_TO_BIT_FACTOR           8
 #define SEC_TO_MSEC_FACTOR	     1000
 
+#define PROD_EQUALIZE_ENCODE_JSON_PATH "/mnt/script.json"
+
+#define SYU645B_REG_DATA_TYPE_MAX 21
```
aadotverma: SYU645B_REG_DATA_TYPE_MAX is also used in enum reg_data_type

It can lead to conflict. Please check.

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	/* Initialize for i2c write */
+	uint8_t reg[SYU645B_REG_DATA_TYPE_MAX];
+	FAR struct i2c_dev_s *dev = priv->i2c;
+	FAR struct i2c_config_s *syu645b_i2c_config = &(priv->lower->i2c_config);
+	
+	int ret = 0;
+
+	/* Parse equlizer json to gJSON */
+	ret = parse_equalizer_metadata_json();
+	if (ret) {
```
aadotverma: if (ret != OK) { 

improves readability

os/drivers/audio/syu645b.c
```diff
@@ -1031,19 +1040,201 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			}
+
+			if (script.addr == 0x08 && script.type == SYU645B_REG_D_2BYTE) {
+				priv->max_volume = script.val[0];
+				printf("[TEST] priv->max_volume update %d \n", priv->max_volume);
+			}
+
+			reg[0] = script.addr;
+			for (int i = 0; i < script.type-1; i++) {
+				reg[i+1] = script.val[i];
```
aadotverma: reg[i + 1] =

os/drivers/audio/syu645b.c
Taejun-Kwon:
```
if () {
} else {
}
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	if (!gJSON) {
+		meddbg("Failed to parse equalizer json file.\n");
+		ret = ERROR;
+		goto err_with_fd;
+	}
+
+err_with_fd:
+	close(fd);
+err_with_buffer:
+	close(buffer);
```
Taejun-Kwon: free, not close

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		syu645b_set_basic_equalizer(priv);
+		return;
+	}
+	if (parse_equalizer_metadata_json() == ERROR) {
+		syu645b_set_basic_equalizer(priv);
+		return;
+	}
+
+	/* Executes given script through i2c to configure SYU645B device. */
+	cJSON *search = cJSON_GetObjectItemCaseSensitive(gJSON, c_preset);
```
Taejun-Kwon, line 1179: gJSON will hold memory. can you check 'mem_leak' ?

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+				}
+			}
+			break;
+		}
+	}
+
+	return OK;
+}
+
+static int string_to_script(char *string, t_codec_init_script_entry *script){
```
Taejun-Kwon: this function always return OK

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	/* Find proper key */
+	cJSON *key;
+	t_codec_init_script_entry script;
+	
+	cJSON_ArrayForEach(key, search){
+		if (strcmp(key->string, "_comment") == 0) {
+			continue;
+		}
+
+		ret = string_to_script(key->valuestring, &script);
```
Taejun-Kwon: remove line from 1203 to 1207, it will never be happened.

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			syu645b_set_basic_equalizer(priv);
+			return;
+		}
+
+		if (script.addr == 0x08 && script.type == SYU645B_REG_D_2BYTE) {
+			priv->max_volume = script.val[0];
+		}
+
+		reg[0] = script.addr;
+		for (int i = 0; i < script.type-1; i++) {
```
Taejun-Kwon: coding rule, 
```
script.type - 1
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+			reg[i+1] = script.val[i];
+		}
+
+		ret = i2c_write(dev, syu645b_i2c_config, (uint8_t *)reg, script.type);
+		if (ret < script.type) {
```
Taejun-Kwon: ret != script.type

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		ptr++;
+	}
+
+	script->delay = 0;
+	script->type = cnt;
+
+	return OK;
+}
+
+static int syu645b_set_basic_equalizer(FAR struct syu645b_dev_s *priv){
```
Taejun-Kwon: 1. coding rule
```
static int syu645b_set_basic_equalizer(FAR struct syu645b_dev_s *priv) {
```
2. Line 1353 also should call this function, but should not set volume at that time. hence I suggest below
```
static int syu645b_set_basic_equalizer(FAR struct syu645b_dev_s *priv, bool set_volume)
```
3. Function is int type should return some error or OK

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	FAR struct i2c_dev_s *dev = priv->i2c;
+	FAR struct i2c_config_s *syu645b_i2c_config = &(priv->lower->i2c_config);
+	
+	int ret = 0;
+	char c_preset[EQUALIZER_PRESET_LEN];
 
+	/* Convert preset hex to string for key & Parse equlizer json to gJSON */
+	if (dec_to_hex_string(preset, c_preset) == ERROR) {
+		syu645b_set_basic_equalizer(priv);
+		return;
```
Taejun-Kwon: I think syu645b_set_equalizer need to be changed to int.
here it should return -EINVAL (because preset is not valid)
hmm and I think we already set basic eq value in syu645b_initialize, syu645b_set_basic_equalizer can be removed.

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
 	syu645b_setmute(priv, true);
 }
 
+static int parse_equalizer_metadata_json(void){
+	int ret = OK;
+
+	struct stat st;
+	if (stat(PROD_EQUALIZE_ENCODE_JSON_PATH, &st) != 0) {
+		auddbg("Failed to fetch json file : %d\n", errno);
+		ret = ERROR;
```
Taejun-Kwon:
```
int ret = OK;
ret = stat(PROD_EQUALIZE_ENCODE_JSON_PATH, &st);
if (ret != OK) {
   return ret; // Perhaps flash file system will return -ENOENT
}
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		return ret;
+	}
+
+	// Todo: If the script file is large, you can use a lot of memory when you read it at once, so should set a static size so that only a certain part of it can be read and used.
+	char *buffer = (char *)malloc(st.st_size + 1);
+	if (!buffer) {
+		meddbg("Failed to allocate memory to hold equalizer json file content.\n");
+		return ERROR;
```
Taejun-Kwon: -ENOMEM

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	}
+	int fd = open(PROD_EQUALIZE_ENCODE_JSON_PATH, O_RDONLY);
+	if (fd == -1) {
+		meddbg("Failed to open equalizer json.\n");
+		ret = ERROR;
```
Taejun-Kwon: same as above. (stat)
file system will return correct error value.

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		meddbg("Failed to read from equalizer json.\n");
+		ret = ERROR;
+		goto err_with_fd;
+	}
+	
+	gJSON = cJSON_Parse(buffer);
+	
+	if (!gJSON) {
+		meddbg("Failed to parse equalizer json file.\n");
+		ret = ERROR;
```
Taejun-Kwon: -EIO

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	int fd = open(PROD_EQUALIZE_ENCODE_JSON_PATH, O_RDONLY);
+	if (fd == -1) {
+		meddbg("Failed to open equalizer json.\n");
+		ret = ERROR;
+		goto err_with_buffer;
+	}
+	ssize_t bytesRead = read(fd, buffer, st.st_size);
+	if (bytesRead == -1 || bytesRead != st.st_size) {
+		meddbg("Failed to read from equalizer json.\n");
+		ret = ERROR;
```
Taejun-Kwon: -EIO

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+
+err_with_fd:
+	close(fd);
+err_with_buffer:
+	close(buffer);
+	return ret;
+}
+
+static int dec_to_hex_string(uint32_t decimal, char *hexadecimal){
+	if (decimal > EQUALIZER_PRESET_MAX) {
```
Taejun-Kwon: EQUALIZER_PRESET_MAX is max value of uint32_t, so This will not be happened foreve.

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	
+	int ret = 0;
+	char c_preset[EQUALIZER_PRESET_LEN];
 
+	/* Convert preset hex to string for key & Parse equlizer json to gJSON */
+	if (dec_to_hex_string(preset, c_preset) == ERROR) {
+		syu645b_set_basic_equalizer(priv);
+		return;
+	}
+	if (parse_equalizer_metadata_json() == ERROR) {
```
Taejun-Kwon: so,
```
ret = parse_equalizer_metadata_json()
if (ret != OK) {
    return ret;
}
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1041,187 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		syu645b_set_basic_equalizer(priv);
+		return;
+	}
+
+	/* Executes given script through i2c to configure SYU645B device. */
+	cJSON *search = cJSON_GetObjectItemCaseSensitive(gJSON, c_preset);
+
+	/* Not find proper key, execute default script*/
+	if (!search){
+		syu645b_set_basic_equalizer(priv);
```
Taejun-Kwon:
```
if (!search) {
    return -EINVAL;
}
```

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1040,180 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+		else {
+			hexadecimal[pos] = 65 + (mod - 10);
+		}
+		decimal = decimal / 16;
+		pos--;
+
+		if (decimal == 0) {
+			if (pos > 1) {
+				for (int i = 2; i <= pos; i++) {
+					hexadecimal[pos] = '0';
```
Taejun-Kwon: for loop looks something wrong. pos is always same

os/drivers/audio/syu645b.c
```diff
@@ -1031,15 +1040,180 @@ static void syu645b_hw_reset_config(FAR struct syu645b_dev_s *priv)
...
+	/* Not find proper key, execute default script*/
+	if (!search){
+		return -EINVAL;
+	}
+
+	/* Find proper key */
+	cJSON *key;
+	t_codec_init_script_entry script;
+	
+	cJSON_ArrayForEach(key, search){
```
Taejun-Kwon: blank before '{'

Taejun-Kwon (approved):

PR #6859, opened 2025-06-19
os/kernel/sched: addition of panic timer in case of device hang.

samsung-singh:
whenever device gets hang panic timer waits for 2 minutes then assert to get device logs.


neel-samsung:
@samsung-singh How about splitting it into two different commits ?
1. add config
2. add panic timer

os/kernel/sched/sched_processtimer.c
```diff
@@ -258,6 +258,20 @@ void sched_process_timer(void)
...
+	static int timer_count = 0;
+	int pid = getpid();
+	
+	if (last_pid != pid) {
+		last_pid = pid;
+		timer_count = 0;
+	} else if (++timer_count >= CONFIG_IDLE_PANIC_TIMER * 1000) {
+		PANIC();
+	}
+#endif
```
eunwoo-nam: 1. This change can only detect cases where a high-priority thread continuously holds and does not release the CPU. Please add comments regarding this limitation.

2. This implementation didn't consider SMP  environments.

3. how about to separate this functionality into new function?

build/configs/rtl8730e/audio/defconfig
```diff
@@ -1716,3 +1716,5 @@ CONFIG_SECURITY_AUTH=y
 CONFIG_SECURITY_CRYPTO=y
 CONFIG_SECURITY_KEYMGR=y
 CONFIG_SECURITY_SS=y
+
+CONFIG_IDLE_PANIC_TIMER=120
```
eunwoo-nam: If you add new config, you have to add it in Kconfig also.

os/kernel/sched/sched_processtimer.c
```diff
@@ -258,6 +258,20 @@ void sched_process_timer(void)
...
+
+#if defined(CONFIG_IDLE_PANIC_TIMER)
+	static int last_pid = -1;
+	static int timer_count = 0;
+	int pid = getpid();
+	
+	if (last_pid != pid) {
+		last_pid = pid;
+		timer_count = 0;
+	} else if (++timer_count >= CONFIG_IDLE_PANIC_TIMER * 1000) {
```
samsung-agnel: The 1000 is hardcoded; can we instead compute it from USEC_PER_TICK or other symbols?

build/configs/rtl8730e/flat_apps/defconfig
```diff
@@ -1559,3 +1559,5 @@ CONFIG_SECURITY_AUTH=y
 CONFIG_SECURITY_CRYPTO=y
 CONFIG_SECURITY_KEYMGR=y
 CONFIG_SECURITY_SS=y
+
+CONFIG_IDLE_PANIC_TIMER=120
```
samsung-agnel: Is it possible to include unit of time in the name for this setting like SEC/MSEC/USEC?

os/kernel/sched/sched_processtimer.c
samsung-singh: updated

build/configs/rtl8730e/audio/defconfig
samsung-singh: updated.

build/configs/rtl8730e/loadable_ext_psram/defconfig
```diff
@@ -1780,3 +1780,5 @@ CONFIG_SECURITY_AUTH=y
 CONFIG_SECURITY_CRYPTO=y
 CONFIG_SECURITY_KEYMGR=y
 CONFIG_SECURITY_SS=y
+
+CONFIG_IDLE_PANIC_TIMER=120
```
sunghan-chang: When we run the menuconfig, this will be changed. Let's add this by menuconfig.

os/kernel/Kconfig
```diff
@@ -268,6 +268,13 @@ config PREALLOC_TIMERS
 		pool of preallocated timer structures to minimize dynamic allocations.  Set to
 		zero for all dynamic allocations.
 
+config CONFIG_IDLE_PANIC_TIMER
```
sunghan-chang: We don't use config definition like `config CONFIG_XXXX`.
If you do this like above, you should use the config in code like `CONFIG_CONFIG_XX`, not `CONFIG_XX`.

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,38 @@ static inline void sched_process_scheduler(void)
...
+void idle_panic_timer(void)
+{
+	static int last_pid = -1;
+	static int timer_count = 0;
+	int pid = getpid();
+	
+	if (last_pid != pid) {
+		last_pid = pid;
+		timer_count = 0;
+	} else if (++timer_count >= CONFIG_IDLE_PANIC_TIMER * TICK_PER_SEC) {
```
sunghan-chang: Please check my previous comment. There is no CONFIG_XX in Kconfig. You should use CONFIG_CONFIG_XX. But it looks not correct.

build/configs/rtl8730e/loadable_ext_ddr/defconfig
```diff
@@ -1849,3 +1849,5 @@ CONFIG_SECURITY_AUTH=y
 CONFIG_SECURITY_CRYPTO=y
 CONFIG_SECURITY_KEYMGR=y
 CONFIG_SECURITY_SS=y
+
+CONFIG_IDLE_PANIC_TIMER=120
```
sunghan-chang: Let's split the commit into two commits.
1. .c and Kconfig change. (Logic implementation)
2. defconfig change (apply the change in configuration)

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,38 @@ static inline void sched_process_scheduler(void)
 #  define sched_process_scheduler()
 #endif
 
+/************************************************************************
+ * Name:	idle_panic_timer
+ *
+ * Description:
+ *  This function check if the system is idle for a long time. It
```
sunghan-chang, line 222: 1. If we keep the idle running, can we say it's hang?
2. PM goal is keeping idle running as longer as. 1st looks strange.
3. How about hang detector?

sunghan-chang:
@samsung-singh `whenever device gets hang panic timer waits for 2 minutes then assert
to get device logs.`
Need to elaborate more. What does the `hang panic timer`? What does it mean `the timer waits for 2 mins`? What is the concept or thought in this commit?

os/kernel/Kconfig
kishore-sn: Let us change the naming of this config as HANG_DETECT_WAIT_SECS . 
Description: If a task is running for more than HANG_DETECT_WAIT_SECS without a context switch, then we assume that the device has hanged and we try to print the debug logs and restart the device.

os/kernel/sched/sched_processtimer.c
kishore-sn: It is possible for the device to run idle thread for any duration. This check must be applied only for other threads. We also need to take care of idle thread for all cpus in smp case. So, let us add a condition in the "else if" statement. 

`pid != up_cpu_index() && ++timer_count >= CONFIG_IDLE_PANIC_TIMER * TICK_PER_SEC`

os/kernel/sched/sched_processtimer.c
```diff
@@ -258,6 +290,11 @@ void sched_process_timer(void)
 	{
 		clock_timer();
 	}
+
+#if defined(CONFIG_IDLE_PANIC_TIMER)
```
kishore-sn: Since this feature is to be used only for kernel debugging, I think we need to check for CONFIG_DEBUG flag here. 
In the real product, this feature must get disabled. The product team has its own feature to detect system hang and restart the device after 3 mins.

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,38 @@ static inline void sched_process_scheduler(void)
...
+{
+	static int last_pid = -1;
+	static int timer_count = 0;
+	int pid = getpid();
+	
+	if (last_pid != pid) {
+		last_pid = pid;
+		timer_count = 0;
+	} else if (++timer_count >= CONFIG_IDLE_PANIC_TIMER * TICK_PER_SEC) {
+		PANIC();
```
kishore-sn: Let us use DEBUGPANIC_INFO instead of PANIC and print the reason for calling panic. `PANIC called because device hang detection kicked in after %d secs`

os/kernel/Kconfig
samsung-singh: updated.

os/kernel/sched/sched_processtimer.c
samsung-singh: included idle process case.

kishore-sn:
@samsung-singh 
Please resolve the build errors.

os/kernel/sched/sched_processtimer.c
```diff
@@ -258,6 +295,9 @@ void sched_process_timer(void)
 	{
 		clock_timer();
 	}
+
+	idle_panic_timer();
```
kishore-sn: I think its better to rename the api as "detect_system_hang()".
Also, I think its better to either call the api at the entry of sched_process_timer or after sched_process_scheduler().

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,43 @@ static inline void sched_process_scheduler(void)
...
+ *  continuously holds the CPU. SMP case is not considered
+ *
+ * Inputs:
+ *   None
+ *
+ * Return Value:
+ *   None
+ *
+ ************************************************************************/
+void idle_panic_timer(void)
```
kishore-sn: This api must be static type.

samsung-singh: updated

os/kernel/sched/sched_processtimer.c
```diff
@@ -247,6 +284,8 @@ static inline void sched_process_scheduler(void)
 
 void sched_process_timer(void)
 {
+	detect_system_hang();
```
r-prabu, line 287: Do we need to have conditional compilation to execute only on non smp case?

kishore-sn: Hang detection must work in both cases. In fact it is more important in smp case.

neel-samsung:
Is this approach tested? and it must work for all kind of hangs.

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,43 @@ static inline void sched_process_scheduler(void)
...
+static void detect_system_hang(void)
+{
+#if defined(CONFIG_DEBUG) && defined(CONFIG_HANG_DETECT_WAIT_SECS)
+	static int last_pid = -1;
+	static int timer_count = 0;
+	int pid = getpid();
+	
+	if (last_pid != pid) {
+		last_pid = pid;
+		timer_count = 0;
```
eunwoo-nam, line 247: How about check both cpu0 and cpu1 ?

```c
for (int cpu = 0; cpu < CONFIG_SMP_NCPUS; cpu++) {
  if (last_cpu[cpu] != g_assignedtasks[cpu]) {
      last_cpu[cpu] = g_assignedtasks[cpu];
  } else if ((struct tcb_s *)g_assignedtasks[cpu]->pid != cpu && ++timer_count[cpu] >= CONFIG_HANG_DETECT_WAIT_SECS * TICK_PER_SEC) {
     /* detected hang */;
  }
}
```

os/kernel/sched/sched_processtimer.c
```diff
@@ -215,6 +215,43 @@ static inline void sched_process_scheduler(void)
...
+	} else if (pid != up_cpu_index() && ++timer_count >= CONFIG_HANG_DETECT_WAIT_SECS * TICK_PER_SEC) {
```
eunwoo-nam, line 248: it's not good to calculate it in every tick isr.
Let's fix value in build time
```#define MAX_HANG_DETECT_WAIT_TICK   (CONFIG_HANG_DETECT_WAIT_SECS * TICK_PER_SEC)```

PR #6861, opened 2025-06-23
os/board/rtl8730e: Limit max STA connection in SoftAP

jwei5:
- Set maximum STA connections allowed in SoftAP mode to 3

### DIFF
--- a/os/board/rtl8730e/src/component/wifi/driver/include/autoconf_8730e.h
+++ b/os/board/rtl8730e/src/component/wifi/driver/include/autoconf_8730e.h
@@ -32,7 +32,11 @@
 
 /************************* Default Values of User Configure *****************************/
 /* Upper limit of STAs connected with SoftAP, more STAs connected will cost more heap*/
+#ifdef CONFIG_PLATFORM_TIZENRT_OS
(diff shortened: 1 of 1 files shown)


os/board/rtl8730e/src/component/wifi/driver/include/autoconf_8730e.h
```diff
@@ -32,7 +32,11 @@
 
 /************************* Default Values of User Configure *****************************/
 /* Upper limit of STAs connected with SoftAP, more STAs connected will cost more heap*/
+#ifdef CONFIG_PLATFORM_TIZENRT_OS
+#define AP_STA_NUM	3
+#else
 #define AP_STA_NUM	12
```
hk-gwak, line 38: Before this PR, the max number of station is 12? I thought there was no max number.

hk-gwak (approved):
LGTM

os/board/rtl8730e/src/component/wifi/driver/include/autoconf_8730e.h
jwei5: Yes, there is always a max number of stations supported for SoftAP

PR #6865, opened 2025-06-24
os/mm : Replace ARCH_GET_RET_ADDRESS api with __builtin_return_address api for debugging purpose

aniket1samsung:
Replaces ARCH_GET_RET_ADDRESS api with __builtin_return_address api with level zero returning the address for current function.

PR #6867, opened 2025-06-24
Replaces ARCH_GET_RET_ADDRESS api with  __builtin_return_address api

aniket1samsung:
This PR includes the changes from previously closed PR #6865 and incorporates the requested review changes.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -580,11 +580,6 @@ static inline void print_assert_detail(const uint8_t *filename, int lineno, stru
 
 void up_assert(const uint8_t *filename, int lineno)
 {
-	/* ARCH_GET_RET_ADDRESS should always be
-	 * called at the start of the function */
-
-	size_t kernel_assert_location = 0;
-	ARCH_GET_RET_ADDRESS(kernel_assert_location);
 
 	irqstate_t flags = enter_critical_section();
 
@@ -608,7 +603,7 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = (uint32_t)user_assert_location;
 		user_assert_location = 0x0;
 	} else {
-		asserted_location = kernel_assert_location;
+		asserted_location = (uint32_t)GET_RETURN_ADDRESS();
 	}
 
 #ifdef CONFIG_SMP
--- a/os/arch/arm/src/armv8-m/up_assert.c
+++ b/os/arch/arm/src/armv8-m/up_assert.c
@@ -589,11 +589,6 @@ void dump_all_stack(void)
 
 void up_assert(const uint8_t *filename, int lineno)
 {
-	/* ARCH_GET_RET_ADDRESS should always be
-	 * called at the start of the function */
-
-	size_t kernel_assert_location = 0;
-	ARCH_GET_RET_ADDRESS(kernel_assert_location)
 
 	irqstate_t flags = irqsave();
 
@@ -629,7 +624,7 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = (uint32_t)user_assert_location;
 		user_assert_location = 0x0;
 	} else {
-		asserted_location = (uint32_t)kernel_assert_location;
+		asserted_location = (uint32_t)GET_RETURN_ADDRESS();
 	}
 
 #ifdef CONFIG_SECURITY_LEVEL
--- a/os/include/tinyara/compiler.h
+++ b/os/include/tinyara/compiler.h
@@ -69,6 +69,10 @@
 #define CONFIG_CPP_HAVE_VARARGS 1	/* Supports variable argument macros */
 #define CONFIG_CPP_HAVE_WARNING 1	/* Supports #warning */
 
+/* GCC macro to retrieve current function's return address as mmaddress_t for debugging. */
+
+#define GET_RETURN_ADDRESS()  __builtin_return_address(0)
+
 /* Intriniscs */
 
 #define CONFIG_HAVE_FUNCTIONNAME 1	/* Has __FUNCTION__ */
@@ -240,6 +244,7 @@
 
 /* Pre-processor */
 
+#undef GET_RETURN_ADDRESS
 #define CONFIG_CPP_HAVE_VARARGS 1	/* Supports variable argument macros */
 #define CONFIG_CPP_HAVE_WARNING 1	/* Supports #warning */
 
@@ -457,6 +462,7 @@
 
 #else
 
+#undef  GET_RETURN_ADDRESS
 #undef  CONFIG_CPP_HAVE_VARARGS
 #undef  CONFIG_CPP_HAVE_WARNING
 #undef  CONFIG_HAVE_FUNCTIONNAME
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -64,6 +64,7 @@
 #include <debug.h>
 #include <stdint.h>
 #include <tinyara/mm/heap_regioninfo.h>
+#include <tinyara/compiler.h>
 #ifdef CONFIG_HEAPINFO_USER_GROUP
 #include <tinyara/mm/heapinfo_internal.h>
 #endif
@@ -229,29 +230,6 @@ typedef uint16_t mmsize_t;
 #else
 typedef size_t mmsize_t;
 #define MMSIZE_MAX SIZE_MAX
-#endif
-
-#if defined(CONFIG_ARCH_MIPS)
-/* Macro gets return address of malloc API. */
-#define ARCH_GET_RET_ADDRESS(caller_retaddr) \
-	do { \
-		asm volatile ("sw $ra, %0" : "=m" (caller_retaddr)); \
-	} while (0);
-#elif defined(CONFIG_ARCH_ARM)
-/* This macro should be placed at the first of the function.
- * If no, it is better to use "__builtin_return_address(0)" instead.
- */
-#define ARCH_GET_RET_ADDRESS(caller_retaddr) \
-	do { \
-		asm volatile ("mov %0,lr\n" : "=r" (caller_retaddr));\
-	} while (0);
-#elif defined(CONFIG_ARCH_XTENSA)
-#define ARCH_GET_RET_ADDRESS(caller_retaddr) \
-	do { \
-		asm volatile ("mov %0,a0\n" : "=&r" (caller_retaddr));\
-	} while (0);
-#else
-#error Unknown CONFIG_ARCH option, malloc debug feature wont work.
 #endif
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
--- a/os/kernel/semaphore/sem_post.c
+++ b/os/kernel/semaphore/sem_post.c
@@ -200,8 +200,7 @@ int sem_post(FAR sem_t *sem)
 {
 	irqstate_t saved_state;
 	int ret = ERROR;
-	size_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr);
+	size_t caller_retaddr = (size_t)GET_RETURN_ADDRESS();
 
 	/* Make sure we were supplied with a valid semaphore. */
 
--- a/os/mm/kmm_heap/kmm_calloc.c
+++ b/os/mm/kmm_heap/kmm_calloc.c
@@ -118,8 +118,7 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 	void *ret;
 	struct mm_heap_s *kheap;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
 #endif
 	if (heap_index > HEAP_END_IDX || heap_index < HEAP_START_IDX) {
 		mdbg("kmm_calloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, HEAP_END_IDX);
@@ -161,8 +160,7 @@ FAR void *kmm_calloc(size_t n, size_t elem_size)
 		return NULL;
 	}
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
 #endif
 	return kheap_calloc(n, elem_size
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
--- a/os/mm/kmm_heap/kmm_malloc.c
+++ b/os/mm/kmm_heap/kmm_malloc.c
@@ -134,8 +134,7 @@ void *kmm_malloc_at(int heap_index, size_t size)
 	void *ret;
 	struct mm_heap_s *kheap;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
 #endif
 	if (heap_index > HEAP_END_IDX || heap_index < HEAP_START_IDX) {
 		mdbg("kmm_malloc_at failed. Wrong heap index (%d) of (%d)\n", heap_index, HEAP_END_IDX);
@@ -180,8 +179,7 @@ void *kmm_malloc_at(int heap_index, size_t size)
 FAR void *kmm_malloc(size_t size)
 {
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
 #endif
 	if (size == 0) {
 		return NULL;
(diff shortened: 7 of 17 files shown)


os/.download
samsung-agnel, line 1: This file is not relevant to this change.

os/app_size_file
samsung-agnel, line 1: Exclude

samsung-agnel (commented):
Maybe redo this. Too many minor issues.

os/mm/umm_heap/umm_xalloc_user_at.c
```diff
@@ -43,8 +43,8 @@ void *calloc_user_at(struct mm_heap_s *heap, size_t n, size_t elem_size)
 		return NULL;
 	}
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr;
-	builtin_retaddr = (mmaddress_t)__builtin_return_address(0);
+	mmaddress_t caller_retaddr=0;
```
samsung-agnel: As mentioned in earlier comment, why not declare and assign caller_retaddr in the same line?

`mmaddress_t caller_retaddr = (mmaddress_t)__builtin_return_address(0);`

samsung-agnel (commented):
Can we combine both commits into 1?

sunghan-chang:
@aniket1samsung You miss many comments in previous PR. Please check them.
And in you description,
```
In existing code in os/mm module, we are using  ARCH_GET_RET_ADDRESS  to get the return address (calling function address) for debugging purpose.
In this I have replaced ARCH_GET_RET_ADDRESS(caller_retaddr) with  __builtin_return_address with level 0 in all places in mm module for debugging purpose.
```
Previous way and new way both are the same in the purpose. No need to modify this in purpose.
1. Previous has limitation to get proper address. (must be used at start of a function), new one does not have.
2. New one only has a limitation (working on specific toolchain), but we support the docker so that no issue on this.
That's why we make this change.

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -112,8 +112,7 @@ static void *kheap_calloc(size_t n, size_t elem_size)
  *   The address of the allocated memory (NULL on failure to allocate)
  *
  ************************************************************************/
-
- #if CONFIG_KMM_NHEAPS > 1
```
kishore-sn: It looks like you have made this commit mainly to fix the unnecessary changes from previous commits and resolve previous comments. We generally dont make new commit for resolving comments in previous commits. Please squash both commits into a single commit.

os/.download
```diff
@@ -0,0 +1 @@
+PREV_PORT=ttyUSB1
```
kishore-sn: Pls remove this file from commit.

os/app_size_file
```diff
@@ -0,0 +1,2 @@
+   text	   data	    bss	    dec	    hex	filename
+ 334631	    974	   7864	 343469	  53dad	../build/output/bin/common
```
kishore-sn: Pls remove this file from commit.

os/mm/umm_heap/umm_zalloc.c
```diff
@@ -177,12 +176,14 @@ FAR void *zalloc(size_t size)
...
 	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	caller_retaddr = (mmaddress_t)__builtin_return_address(0);
 #endif
 
 	if (size == 0) {
 		return NULL;
 	}
+
+
```
kishore-sn: ARCH_GET_RET_ADDRESS is also used in assert.c file of all architectures. You can replace those also. 
Finally, it has been declared in os/include/tinyara/mm/mm.h. We can remove that declaration as it is no longer required.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -580,11 +580,8 @@ static inline void print_assert_detail(const uint8_t *filename, int lineno, stru
 
 void up_assert(const uint8_t *filename, int lineno)
 {
-	/* ARCH_GET_RET_ADDRESS should always be
-	 * called at the start of the function */
 
-	size_t kernel_assert_location = 0;
-	ARCH_GET_RET_ADDRESS(kernel_assert_location);
+	size_t kernel_assert_location = (size_t)__builtin_return_address(0);
```
sunghan-chang: When we use the `ARCH_GET_RET_ADDRESS`, we should get the address at start of the function. That's why we define the local variable - kernel_assert_location and use it line 611 only.
But with the new function, we don't need to save it here. I mean no need to use the variable in this purpose.

sunghan-chang:
@aniket1samsung `Removed declarations of ARCH_GET_RET_ADDRESS and removed unnecessary files.` Could you rewrite the title? The main change which is important in history point of view in this commit is not removing `ARCH_GET_RET_ADDRESS`. Using __builtin... instead of ARCH_... is important I think.

os/include/tinyara/mm/mm.h
```diff
@@ -231,28 +231,6 @@ typedef size_t mmsize_t;
...
-		asm volatile ("mov %0,lr\n" : "=r" (caller_retaddr));\
-	} while (0);
-#elif defined(CONFIG_ARCH_XTENSA)
-#define ARCH_GET_RET_ADDRESS(caller_retaddr) \
-	do { \
-		asm volatile ("mov %0,a0\n" : "=&r" (caller_retaddr));\
-	} while (0);
-#else
-#error Unknown CONFIG_ARCH option, malloc debug feature wont work.
-#endif
```
eunwoo-nam, line 255: __biltin_return_address(0) is more convenient
However, we need to make sure that __biltin_return_address(0) is supported by all compilers.
That's why we use macros.
If new chipset and new compilers do not support, we have to replace with macro again.

kishore-sn: @aniket1samsung Let us revert this file.

os/mm/kmm_heap/kmm_malloc.c
```diff
@@ -180,13 +179,11 @@ void *kmm_malloc_at(int heap_index, size_t size)
...
 {
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = (mmaddress_t)__builtin_return_address(0);
 #endif
 	if (size == 0) {
 		return NULL;
 	}
-
```
kishore-sn: Pls revert this change.

os/mm/umm_heap/umm_calloc.c
```diff
@@ -79,14 +79,12 @@
  *   The address of the allocated memory (NULL on failure to allocate)
  *
  ************************************************************************/
-
```
kishore-sn: Pls revert.

os/mm/umm_heap/umm_zalloc.c
```diff
@@ -80,14 +80,12 @@
  *   The address of the allocated memory (NULL on failure to allocate)
  *
  ************************************************************************/
-
```
kishore-sn: Pls revert all un-related changes like adding or removing empty lines.

kishore-sn:
@aniket1samsung 
Please change the commit title as below:

`os/mm: Use builtin_return_address api to get caller address`

os/include/tinyara/mm/mm.h
sunghan-chang: What's the direction of this, then?
Keeping the macro and just converting it to builtin function for ARM only?

os/mm/umm_heap/umm_realloc.c
```diff
@@ -89,9 +89,7 @@ void *realloc_at(int heap_index, void *oldmem, size_t size)
 {
 	void *ret;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = (mmaddress_t)__builtin_return_address(0);
```
eunwoo-nam: We need to abstract this to at least some extent.

While it is supported by GCC and Clang, it may not be supported by certain other compilers (e.g., MSVC).

Looking at the usage of __builtin_return_address(0) in NuttX and Linux, we can observe that it is abstracted using a #define directive.
https://github.com/apache/nuttx/blob/5594c2e8875730f2f65b9fab57806e4c3c046379/include/nuttx/compiler.h#L171
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/instruction_pointer.h#n7

To ensure future scalability and compatibility, it is necessary to abstract this functionality like below
```
#define GET_RETURN_ADDRESS   (mmaddress_t)__builtin_return_address(0);
```

os/include/tinyara/mm/mm.h
kishore-sn: We cannot convert the same macro for using builtin apis. If we dont have gcc compiler, then we need this definition of macro. Also, it needs to be called in the first line of the function. So, I think its better to maintain the `ARCH_GET_RET_ADDRESS` definitions. They can be used in future if required. For the current implementation, we can add another macro in compiler.h file as below. 


```
#ifdef __GNUC__
#define GET_RETURN_ADDRESS   (mmaddress_t)__builtin_return_address(0);
#else
#undef GET_RETURN_ADDRESS   
#endif
```

We can modify our code as below:

```
mmaddress_t caller_retaddr;
#ifndef GET_RETURN_ADDRESS
ARCH_GET_RET_ADDRESS(caller_retaddr);
#else 
caller_address = GET_RETURN_ADDRESS
#endif 
```

Also, like below:
```
#ifndef GET_RETURN_ADDRESS
size_t kernel_assert_location = 0;
ARCH_GET_RET_ADDRESS(kernel_assert_location)
#endif
...
...
...
#ifndef GET_RETURN_ADDRESS
assert_location = kernel_assert_location;
#else 
assert_location = GET_RETURN_ADDRESS
#endif
```

@sunghan-chang & @ewoodev 
Please let us know if this is ok.

os/kernel/semaphore/sem_post.c
```diff
@@ -214,7 +212,7 @@ int sem_post(FAR sem_t *sem)
 		saved_state = enter_critical_section();
 
 		/* Perform the semaphore unlock operation. */
-		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x, caller address = 0x%x", sem, caller_retaddr);
+		ASSERT_INFO(sem->semcount < SEM_VALUE_MAX, "sem = 0x%x, caller address = 0x%x", sem,__builtin_return_address(0));
```
sunghan-chang: 1. need a space after `,` for the coding rule
2. @kishore-sn @ewoodev left comments for that. What I understand is that we should keep and use the defined macro instead of using the builtin function for different architecture. Is that right? Could you confirm it? @ewoodev @kishore-sn

neel-samsung:
@aniket1samsung Some of the GitHub TCs are failing. Please check.

os/include/tinyara/compiler.h
```diff
@@ -62,6 +62,12 @@
 
 /* GCC-specific definitions *************************************************/
 
+#if defined(__GNUC__) && defined(CONFIG_DEBUG_MM_HEAPINFO)
```
kishore-sn: This is not related to heapinfo only. It can be used by anyone. So, let us remove that condition.

kishore-sn: Also, we can club our definition with below definitions under gnuc

os/kernel/semaphore/sem_post.c
```diff
@@ -200,8 +200,12 @@ int sem_post(FAR sem_t *sem)
 {
 	irqstate_t saved_state;
 	int ret = ERROR;
-	size_t caller_retaddr = 0;
+	size_t caller_retaddr=0;
```
kishore-sn: Pls add space before and after '='

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -161,8 +165,12 @@ FAR void *kmm_calloc(size_t n, size_t elem_size)
 		return NULL;
 	}
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr =0;
```
kishore-sn: Pls take care of spacing in all places.

kishore-sn (dismissed):

kishore-sn:
@sunghan-chang @ewoodev 
We have added macro as per your suggestion. 
Please check and let us know if this can be merged.

os/arch/arm/src/armv8-m/up_assert.c
```diff
@@ -629,7 +631,11 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = (uint32_t)user_assert_location;
 		user_assert_location = 0x0;
 	} else {
-		asserted_location = (uint32_t)kernel_assert_location;
+		#ifndef GET_RETURN_ADDRESS
```
eunwoo-nam: Sorry for confusing
i mean, we need make wrappping __builtin_return_address(0) for expend new compiler.
we don't need to keep ARCH_GET_RET_ADDRESS()

our almost build is using GNU compiler, let's use __builtin_return_address(0)
and if new compiler is comming, let's add to branch using new way to getting caller address.

kishore-sn: Hi Eunwoo.. 
Our current implementation also supports adding new compilers.

```
#ifdef gnuc
#define GET_RETURN_ADDRESS __builtin_return_address(0)
#else if clang
#define GET_RETURN_ADDRESS __something_else__
#else 
#undef GET_RETURN_ADDRESS 
#endif
```

Only in the last case, if any compiler does not support getting return address, then we use ARCH_GET_RET_ADDRESS().

However, if you feel that ARCH_GET_RET_ADDRESS() has to be completely removed, then we can do that. Please confirm this.

kishore-sn: @ewoodev 
Please let me know your opinion, so that we can close this PR.

r-prabu (approved):
Changes for the getting the return address with the built-in function is fine.

os/arch/arm/src/armv8-m/up_assert.c
eunwoo-nam:
```c
#ifdef gnuc
#define RETURN_ADDRESS() __builtin_return_address(0)
#else if clang
#define GET_RETURN_ADDRESS() __something_else__
#else 
#undef GET_RETURN_ADDRESS 
#endif
```
it looks good

Let's remove ARCH_GET_RET_ADDRESS(kernel_assert_location)

samsung-singh (approved):
Replaced ARCH_GET_RET_ADDRESS api with __builtin_return_address api in os module this change is good to merge.

ewoodev (approved):

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -608,7 +603,7 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = (uint32_t)user_assert_location;
 		user_assert_location = 0x0;
 	} else {
-		asserted_location = kernel_assert_location;
+			asserted_location = (uint32_t)GET_RETURN_ADDRESS();
```
eunwoo-nam: Please check indentation

os/include/tinyara/compiler.h
```diff
@@ -240,6 +244,7 @@
 
 /* Pre-processor */
 
+#define GET_RETURN_ADDRESS()
```
kishore-sn: Since we dont have a method to get return address for this compiler, let us set this as `#undef GET_RETURN_ADDRESS`

os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -118,8 +118,7 @@ void *kmm_calloc_at(int heap_index, size_t n, size_t elem_size)
 	void *ret;
 	struct mm_heap_s *kheap;
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = 0;
-	ARCH_GET_RET_ADDRESS(caller_retaddr)
+	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
```
sunghan-chang, line 121: The definition is defined in os/include/tinyara/compiler.h.
Don't you need to include it to use?

sunghan-chang (approved):

kishore-sn (approved):

os/mm/kmm_heap/kmm_calloc.c
kishore-sn: > The definition is defined in os/include/tinyara/compiler.h. Don't you need to include it to use?

Added compiler.h to mm.h and mm.h is included in all src files where return address macro is used.

kishore-sn:
@sunghan-chang 
All review comments have been addressed. Kindly check and approve.

os/include/tinyara/mm/mm.h
```diff
@@ -231,29 +232,6 @@ typedef size_t mmsize_t;
...
-	} while (0);
-#elif defined(CONFIG_ARCH_ARM)
-/* This macro should be placed at the first of the function.
- * If no, it is better to use "__builtin_return_address(0)" instead.
- */
-#define ARCH_GET_RET_ADDRESS(caller_retaddr) \
-	do { \
-		asm volatile ("mov %0,lr\n" : "=r" (caller_retaddr));\
-	} while (0);
-#elif defined(CONFIG_ARCH_XTENSA)
```
sunghan-chang, line 248: What about xtensa? We have esp32 boards in supported boards.

sunghan-chang: Eunwoo checked it worked with GNUC

sunghan-chang:
@kishore-sn Let's merge

Sambhavi-1 (approved):
This PR uses builtin_return_address api in place of ARCH_GET_RET_ADDRESS api looks goos

PR #6871, opened 2025-07-01
external/libcxx-test: Fix build issue in C++14

pcs1265:
Issue:
When upgrading to C++14, the build of `Map` test failed due to missing `TemplateConstructor` and comparison operators for `PrivateConstructor`. These were likely excluded previously because TizenLite merges all tests into a single executable, causing linker errors ("multiple definitions").

Fix:
Resolve ODR violation by declaring missing operators as `inline` and `static` functions in the shared header.

Reference:
The approach is inspired by LLVM’s test structure (https://github.com/llvm/llvm-project/tree/main/libcxx/test).

This commit makes libcxx-test can be built in C++14.

### DIFF
--- a/external/libcxx-test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp
+++ b/external/libcxx-test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp
@@ -35,6 +35,14 @@
 #include "test_macros.h"
 #include "libcxx_tc_common.h"
 
+struct TemplateConstructor
+{
+    template <typename T>
+    TemplateConstructor (const T&) {}
+};
+
+static bool operator < (const TemplateConstructor&, const TemplateConstructor&) { return false; }
+
 int tc_libcxx_containers_map_modifiers_erase_iter(void)
 {
     {
--- a/external/libcxx-test/std/containers/associative/multimap/multimap.modifiers/erase_iter.pass.cpp
+++ b/external/libcxx-test/std/containers/associative/multimap/multimap.modifiers/erase_iter.pass.cpp
@@ -35,6 +35,14 @@
 #include "test_macros.h"
 #include "libcxx_tc_common.h"
 
+struct TemplateConstructor
+{
+    template <typename T>
+    TemplateConstructor (const T&) {}
+};
+
+static bool operator < (const TemplateConstructor&, const TemplateConstructor&) { return false; }
+
 int tc_libcxx_containers_multimap_modifiers_erase_iter(void)
 {
     {
--- a/external/libcxx-test/std/containers/unord/unord.map/unord.map.modifiers/erase_const_iter.pass.cpp
+++ b/external/libcxx-test/std/containers/unord/unord.map/unord.map.modifiers/erase_const_iter.pass.cpp
@@ -36,9 +36,18 @@
 #include <string>
 #include <cassert>
 #include "libcxx_tc_common.h"
-
 #include "test_macros.h"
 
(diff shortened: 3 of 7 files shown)


external/libcxx-test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp
```diff
@@ -34,6 +34,7 @@
 #include <cassert>
 #include "test_macros.h"
 #include "libcxx_tc_common.h"
+#include "template_constructor.hpp"
```
eunwoo-nam: `typedef TemplateConstructor T` is used in this file. 

https://github.com/llvm/llvm-project/blob/67be4fe3d5fd986a3149de3806bcf2c92320015e/libcxx/test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp#L21-L26

but the struct TemplateConstructor is missing as above.
so you add it 👍👍👍

the llvm is using only `bool operator<`. but `template_constructor.hpp` has `==` operater also. is it okay?

external/libcxx-test/support/template_constructor.hpp
```diff
@@ -0,0 +1,43 @@
...
+    template<typename T>
+    TemplateConstructor (const T&) {}
+};
+
+//declare as inline function to prevent linking error (multiple definition error)
+inline bool operator<(const TemplateConstructor&, const TemplateConstructor&) { return false; }
+inline bool operator==(const TemplateConstructor&, const TemplateConstructor&) { return false; }
+struct Hash {
+  std::size_t operator()(const TemplateConstructor&) const { return 0; }
+};
```
eunwoo-nam: Is this the file form LLVM?

It may not be good for maintenance to separate this into separate files. 👀
Because it's different from the LLVM open source code structure.

external/libcxx-test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp
pcs1265: The equal operator is also used in `erase_const_iter.pass.cpp`, so I combined it into one file.
However, as you mentioned in the comment below, it is different from the LLVM code.
It would be better to follow the LLVM structure.

external/libcxx-test/support/template_constructor.hpp
pcs1265: No, it's not.
I just updated it. Please check again.

pcs1265: No, it's not.
I just updated it. Please check again.

pcs1265: I decided to use `static` function in `*_iter.pass.cpp`.
In this case, declaring functions as `static` seems more suitable.

external/libcxx-test/std/containers/associative/map/map.modifiers/erase_iter.pass.cpp
```diff
@@ -35,6 +35,14 @@
 #include "test_macros.h"
 #include "libcxx_tc_common.h"
 
+struct TemplateConstructor
+{
+    template<typename T>
+    TemplateConstructor (const T&) {}
+};
+
+static bool operator<(const TemplateConstructor&, const TemplateConstructor&) { return false; }
```
jeongarmy: Please add space between keywords

pcs1265: Added. Thank you for reviewing.

PR #6872, opened 2025-07-02
libs/libc/math: Implement nearbyint and its variations

pcs1265:
Some libraries use `nearbyint` and `nearbyintf` from <cmath>, but there were no implementations, causing linker errors:
- nearbyint() << undefined reference
- nearbyintf() << undefined reference

This change adds implementations that map to `rint`, `rintf` and `rintl` respectively, since `nearbyint` is equivalent to `rint` but doesn't raise floating-point exceptions.

Note: This is a temporary solution using existing `rint` functions. For a complete solution, we should consider:
1. Adding proper `fenv` support
2. Implementing FPU acceleration


jeongarmy (approved):
lgtm

PR #6874, opened 2025-07-07
os/kernel: Rename function name in sched_unlock

seokhun-eom24:
- Fix the function name from undefined function `sched_reassess_timer` to correct function `sched_timer_reassess` in `sched_unlock.c`.
- `CONFIG_SMP` not defined case has right function name. (line 275)


jeongarmy (approved):
lgtm

PR #6875, opened 2025-07-07
os/kernel/irq: Correct the input parameter comment in irq_cpu_locked.

seokhun-eom24:
- Correct the input parameter comment from rtcb to cpu in irq_cpu_locked.
- Write a description of input parameter `cpu`.


jeongarmy (approved):
lgtm

PR #6876, opened 2025-07-08
external/libcxx: Prevent C99 macro leakage in <cmath>

pcs1265:
Problem:
- Including `<cmath>` in C++ caused C99 macros (`isfinite`/`isinf`/`isnan`) from math.h to leak into global namespace.
- This led to compilation errors as macros overrode std functions: `std::isnan(NAN)` → `std::((NAN) != (NAN))` (invalid syntax).

Solution:
- Undefine C99 macros.
- Define `std::isfinite`, `std::isinf`, `std::isnan` as functions.

Result:
- Functions in std namespace work as defined by C++ standard.
- No macro interference in C++ compilation paths.

### DIFF
--- a/external/include/libcxx/cmath
+++ b/external/include/libcxx/cmath
@@ -61,10 +61,20 @@
 #include <tinyara/compiler.h>
 
 #include <math.h>
+#include <type_traits>
 
 //***************************************************************************
 // Namespace
 //***************************************************************************
+
+//These functions should be defined in `std` namespace as functions, not a macros.
+//To prevent leaking C99 macros into the C++ codes, they should be undefined.
+#undef isfinite
+#undef isinf
+#undef isnan
+
+// C++ linkage is required here to support overloaded functions (e.g., isinf, isnan).
+extern "C++" {
 
 namespace std
 {
@@ -94,6 +104,21 @@ namespace std
   using ::sqrtf;
   using ::tanf;
   using ::tanhf;
+
+  bool isinf(float __x)
+  {
+    return (((__x) == INFINITY) || ((__x) == -INFINITY));
+  }
+
+  bool isfinite(float __x)
+  { 
+    return (!(isinf(__x)) && (__x != NAN));
+  }
+
+  bool isnan(float __x)
+  { 
+    return ((__x) != (__x));
+  }
 #endif
 
 #ifdef CONFIG_HAVE_DOUBLE
@@ -124,6 +149,21 @@ namespace std
   using ::tanh;
   using ::gamma;
   using ::lgamma;
+
+  bool isinf(double __x)
+  {
+    return (((__x) == INFINITY) || ((__x) == -INFINITY));
+  }
+
+  bool isfinite(double __x)
+  { 
+    return (!(isinf(__x)) && (__x != NAN));
+  }
+
+  bool isnan(double __x)
+  { 
+    return ((__x) != (__x));
+  }
 #endif
 
 #ifdef CONFIG_HAVE_LONG_DOUBLE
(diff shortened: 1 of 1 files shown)


external/include/libcxx/cmath
```diff
@@ -66,6 +66,24 @@
 // Namespace
 //***************************************************************************
 
+//These functions should be defined in `std` namespace as functions, not a macros.
+//To prevent leaking C99 macros into the C++ codes, they should be undefined.
+#ifdef __cplusplus
```
pcs1265: Since this library(libcxx) is C++ library, Checking `__cplusplus` seems to not be necessary.
The code is working without checking `__cplusplus`.
Should I remove this?

external/include/libcxx/cmath
```diff
@@ -66,6 +66,24 @@
...
+  #ifdef isfinite
```
seokhun-eom24: `#ifdef` can be removed. It's your choice.

https://stackoverflow.com/questions/9090536/when-do-we-need-ifdef-before-undef

external/include/libcxx/cmath
pcs1265: removed.

external/include/libcxx/cmath
pcs1265: Accepted.
Thank you.

external/include/libcxx/cmath
```diff
@@ -152,8 +216,56 @@ namespace std
...
+    #else
+      return ((__x) != (__x));
+    #endif
+  }
 #endif
 
+  // For integral types
+  template<typename T>  
+  typename std::enable_if<std::is_integral<T>::value, bool>::type  
+  isinf(T) { return false; }  // Integers can't be infinite
```
jeongarmy: In cppreference, integer value needs to be treated as double value. Please check Note in the below link and give your opinion.

`The additional overloads are not required to be provided exactly as (A). They only need to be sufficient to ensure that for their argument num of integer type, std::isinf(num) has the same effect as std::isinf(static_cast<double>(num)).`
https://en.cppreference.com/w/cpp/numeric/math/isinf

pcs1265: I didn't noticed that rule so I just followed implementation of `libstdc++` ([Link](https://github.com/gcc-mirror/gcc/blob/75111d0594772d1e6dc12c1b9fe7808961ae14b0/libstdc%2B%2B-v3/include/c_global/cmath#L1177))
but when I checked `libc++` also, it is implementing `isinf` as you said.
It would be better following `libc++`'s method.

jeongarmy: What do current our libcxx and nuttx use? libc++ or libstdc++?

pcs1265: Our libcxx is a port of libc++(from [llvm-project/libcxx](https://github.com/llvm/llvm-project/tree/def7bbbe17c6d0302a0cccf67355e36c45fcb224/libcxx)).

So I thought it would be better to follow libc++.

external/include/libcxx/cmath
```diff
@@ -94,6 +104,33 @@ namespace std
   using ::sqrtf;
   using ::tanf;
   using ::tanhf;
+
+  bool isinf(float __x)
+  { 
+    #if __has_builtin(__builtin_isinf)
```
jeongarmy: Please remove space in front of #if, #else, #endif

external/include/libcxx/cmath
```diff
@@ -152,8 +192,44 @@ namespace std
...
+  bool isnan(long double __x)
+  { 
+    return ((__x) != (__x));
+  }
 #endif
 
+  // For integral types
+  template<typename T>  
+  typename std::enable_if<std::is_integral<T>::value, bool>::type  
+  isinf(T __x) { return isinf(static_cast<double>(__x)); }
```
jeongarmy: Is it ok when CONFIG_HAVE_DOUBLE is disabled?

pcs1265: Hmm.
The code has to be complicated to comply with the cppreference that integers should be treated as `double` type for `isinf`
(cppreference does not assume the case where there is no `double` - libc++ too)

Semantically, the results of `isfinite`, `isnan`, and `isinf` for integers are fixed (`isinf: false`, `isnan: false`, `isfinite: true`).

To prevent excessive code complexity,
I think it would be reasonable to return a predefined boolean like libstdc++.

external/include/libcxx/cmath
```diff
@@ -94,6 +104,21 @@ namespace std
   using ::sqrtf;
   using ::tanf;
   using ::tanhf;
+
+  bool isinf(float __x)
+  {
+    return (((__x) == INFINITY) || ((__x) == -INFINITY));
```
jeongarmy, line 110: We declares these same with macros in math.h. So is it possible to change them like below?
1. Make isinf, isfinite, isnan macros to function in libc
2. add `using ::XXX` here without undef XXX

pcs1265: Then the macros will be deleted in `math.h`.

But these macros are defined for overloading-like feature in C.
As they are macros, There are no type constraints. (They can accept all types: float, double, long double)

Making `isinf`, `isfinite`, `isnan` to functions cannot handle this feature.

jeongarmy: okay. I got it

jeongarmy (approved):
lgtm

PR #6881, opened 2025-07-09
build/tools/amebasmart/gnu_utility: Fix AES GCM AAD NULL and Mismatch…

ziyik:
… Error

- Fix AAD NULL error, allow AAD length to be 0, and AAD is NULL.
- Fix decryption success when AAD is mismatched


jylee9613 (approved):
I confirmed with below testcases.
https://github.com/Samsung/TizenRT/pull/6889

Thank you:)

jeongarmy (approved):
lgtm

PR #6882, opened 2025-07-10
os/arch/arm/src/amebasmart: support mipi read

zhongnuo-tang:
1. add mipi receive cmd decode for rx isr
2. add error handling if mipi entered invalid state

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -170,7 +170,50 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 
 	return transfer_status;
 }
+/* rx_len is the maximum return packet size*/
+static int set_return_packet_len(struct mipi_lcd_dev_s *priv, u8 rx_len)
+{
+	int transfer_status = OK;
+	struct mipi_dsi_msg msg;
+	msg.channel = rx_len;
+	msg.type = MIPI_DSI_SET_MAXIMUM_RETURN_PACKET_SIZE;
+	msg.tx_len = 0;
+	msg.flags = 0;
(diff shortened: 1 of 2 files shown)


zhongnuo-tang:
Hi @anjana348 , please check this PR

os/drivers/lcd/mipi_lcd.c
```diff
@@ -170,7 +170,50 @@ static int send_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t command)
 
 	return transfer_status;
 }
+/* rx_len is the maximum return packet size*/
+static int set_return_packet_len(struct mipi_lcd_dev_s *priv, u8 rx_len)
```
anjana348, line 174: what is use of this function? We are setting msg.channel = rx_len ? and no rx_buf is passed? 
There is no command passed. Can you please check it again if it works

zhongnuo-tang: This is a send cmd, it sets the maximum return packet size so that we can read long response.

the msg.type = MIPI_DSI_SET_MAXIMUM_RETURN_PACKET_SIZE and msg.channel = rx_len;

You can refer to the datasheet.
image

I have verified it is working, you can try reading with cmd '0x4', it suppose to return 3 bytes, without this function, you will only get 1 byte return. 
image

similarly to read cmd '0xA1', you can only get all the 5 bytes when u call this function to set rx_len = 5.
image

anjana348: ok, got it. Thanks for your explanation.

anjana348:
> Hi @anjana348 , please check this PR

I verified it, this looks fine.

ewoodev (approved):

jeongarmy (approved):
lgtm

PR #6883, opened 2025-07-10
framework/wifi_manager : remove ip reset in dhcpc close

hk-gwak:
Remove ip reset in dhcpc close
- Even dhcpc closed, softap ip needs to be maintained. Because even AP disconnected, STA can still be connected by bridge mode.
- Performing SOFTAP ip reset in dhcpc close is odd.


j11591 (approved):
LGTM

PR #6886, opened 2025-07-14
os/drivers/lcd : read Register ID for LCD

anjana348:
Support for Read mipi read command

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_mipi.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_mipi.c
@@ -49,7 +49,9 @@
 
 static u32 vo_freq;
 static volatile u8 send_cmd_done = 0;
-
+static volatile u8 receive_cmd_done = 1;
+static volatile bool rx_data_rdy = FALSE;
+static uint8_t rx_data_buff[5]; /*long response have 5 bytes in payload, short response have 1 or 2 bytes in payload */
 struct amebasmart_mipi_dsi_host_s {
 	struct mipi_dsi_host dsi_host;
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/amebasmart/amebasmart_mipi.c
```diff
@@ -182,15 +184,18 @@ static void amebasmart_mipi_init_helper(FAR struct amebasmart_mipi_dsi_host_s *p
 	MIPI_DSI_INT_Config(priv->MIPIx, DISABLE, ENABLE, FALSE);
 }
 
-static void amebasmart_mipidsi_send_cmd(MIPI_TypeDef *MIPIx, u8 cmd, u8 payload_len, const u8 *para_list)
+static void amebasmart_mipidsi_send_cmd(MIPI_TypeDef *MIPIx, u8 cmd, u8 payload_len, const u8 *para_list, u8 cmd_type)
 {
 	u32 word0, word1, addr, idx;
 	u8 cmd_addr[128];
+	if (MIPI_LPTX_IS_READ(cmd_type)) {
+		receive_cmd_done = 0;
```
pcs1265, line 192: What happens if the read command fails?
It looks like if the read fails, it gets set to 0 here and never gets set to 1 again.

send read cmd -> read fail -> send write cmd -> write fail (`receive_cmd_done` is stuck at `0`)

seokhun-eom24: I think you should write comment here too: https://github.com/Samsung/TizenRT/pull/6882

PR #6887, opened 2025-07-15
external/libcxx: Fix iostream in loadable build

pcs1265:
Problem:
- The `iostream` static objects did not initialized in loadable build.
- So, `std::cout` did not work in loadable build (works only in flat build).

Why:
- If `__start_std_streams` have `init_priority`, its constructor did not be included to `ctors`.
- As the `__start_std_streams` declared as static variable, its constructor has to be called before C++ app starts.
- But since its constructor did not be included to `ctors`, it was not called.

Solution:
- I assume there is some reason why `__start_std_streams` has `init_priority`.
- So remove the `__attribute__((init_priority))` from `__start_std_streams` in loadable build only.

Result:
- The `iostream` objects initialize correctly in loadable build.
- `std::cout` works normally in loadable build.

### DIFF
--- a/external/libcxx/iostream.cxx
+++ b/external/libcxx/iostream.cxx
@@ -58,7 +58,11 @@ static mbstate_t mb_wcerr;
 _ALIGNAS_TYPE (ostream)  _LIBCPP_FUNC_VIS char clog[sizeof(ostream)];
 _ALIGNAS_TYPE (wostream) _LIBCPP_FUNC_VIS char wclog[sizeof(wostream)];
 
+#ifdef CONFIG_BINFMT_CONSTRUCTORS
+ios_base::Init __start_std_streams;
+#else
 ios_base::Init __start_std_streams __attribute__ ((init_priority (500)));
+#endif
 
 ios_base::Init::Init()
 {


external/libcxx/iostream.cxx
```diff
@@ -58,7 +58,11 @@ static mbstate_t mb_wcerr;
 _ALIGNAS_TYPE (ostream)  _LIBCPP_FUNC_VIS char clog[sizeof(ostream)];
 _ALIGNAS_TYPE (wostream) _LIBCPP_FUNC_VIS char wclog[sizeof(wostream)];
 
+#ifdef CONFIG_BINFMT_CONSTRUCTORS
+ios_base::Init __start_std_streams;
```
seokhun-eom24: Is there any risk of changing the sequence that is initialized because there is no priority?

external/libcxx/iostream.cxx
```diff
@@ -58,7 +58,11 @@ static mbstate_t mb_wcerr;
 _ALIGNAS_TYPE (ostream)  _LIBCPP_FUNC_VIS char clog[sizeof(ostream)];
 _ALIGNAS_TYPE (wostream) _LIBCPP_FUNC_VIS char wclog[sizeof(wostream)];
 
+#ifdef CONFIG_BINFMT_CONSTRUCTORS
```
jeongarmy: I think we need to add contidions about loadable build because both are declared as static variable for constructor.

external/libcxx/iostream.cxx
jeongarmy: Is there any solution to set the priority after app bringup?

external/libcxx/iostream.cxx
pcs1265: Then the condition statement will be like this.
`#if defined(BINFMT_LOADABLE) && defined(CONFIG_BINFMT_CONSTRUCTORS)`
Is this correct?

external/libcxx/iostream.cxx
pcs1265: There is no way to specify initialization priority of static variables in C++ standard.
`init_priority` is GCC specific feature.

init priority is determined by the declare position in a compiling unit. and init priority between other compiling unit is undefined in C++.

if the init priority is affecting in initialization of static objects
We can explicitly declare the first object in depending object.
```
struct A {  
    A() {  
        std::ios_base::Init init;  // Forcing the std IO streams to be initialized.
        std::cout << "A initialized\n";  
    }  
};  

static A a;
```

Using `init_priority` is not recommended by C++.

pcs1265: Hmm. after the app bring up..
I will research about it soon.

jeongarmy: Please check effects of priority setting. I want to know how priority works in tizenrt and affects io operations. If it doesn't affect, we can ignore priority setting as your changes.

jeongarmy: `Using init_priority is not recommended by C++.`
Could you leave reference or something about this? Then we can remove line 64 simply.

pcs1265: I think I found the exact problem.
Before consider deleting `init_priority`, I focused on why init_priority won't work in loadable build.

If a static variable has `init_priority`, it will be located at `.init_array.XXXXX`(XXXXX is the priority. ex: .init_array.00500).
and It should be sorted by the priority and linked properly to `.init_array`(`.ctors`) of user binary.

but our user binary linker script has no declaration about it.
```
  .ctors :
    {
      _sctors = . ;
      *(.ctors)       /* Old ABI:  Unallocated */
      *(.init_array)  /* New ABI:  Allocated */
      _ectors = . ;
    } > uflash
```

Since the location of priortized constructors is undefined, `.init_array.XXXXX` sections became to orphan sections in common binary between `.data` and `.bss` section.
```
 .data.g_dhcpd_tid
                0x000000006078034c        0x2 ../build/output/libraries/libexternal.a(dhcpd_lwip.o)
 *(.gnu.linkonce.d*)
                0x000000006078034e                _edata = .

.igot.plt       0x0000000060780350        0x0 load address 0x000000000e1afeca
 .igot.plt      0x0000000060780350        0x0 /root/tizenrt/os/userspace/up_userspace_common.o

.init_array.00500
                0x0000000060780350        0x4 load address 0x000000000e1afeca
 .init_array.00500
                0x0000000060780350        0x4 ../build/output/libraries/libcxx.a(iostream.o)

.bss            0x0000000060780358     0x1868 load address 0x000000000e1afece
                0x0000000060780358                _sbss = .
 *(.bss)
```

The `.init_array.XXXXX` section should be located in .ctors like below.
```
.ctors          0x000000000e1afb74        0xc
                0x000000000e1afb74                _sctors = .
 *(.ctors)
 *(SORT_BY_NAME(.init_array.*))
 .init_array.00500
                0x000000000e1afb74        0x4 ../build/output/libraries/libcxx.a(iostream.o)
 *(.init_array)
 .init_array    0x000000000e1afb78        0x4 ../build/output/libraries/libapps.a(helloxx_main.o)
 .init_array    0x000000000e1afb7c        0x4 /bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/lib/thumb/v8-a+simd/hard/libsupc++.a(eh_alloc.o)
                0x000000000e1afb80                _ectors = .

```

To support prioritized initialization of static variable, linker script of user binary should be modified.

pcs1265: And libc++(LLVM) is using `init_priority` in their code(libc++ priority = 100, our priority = 500).
so I think we can keep it now.

jeongarmy (approved):
lgtm

PR #6888, opened 2025-07-15
Sync master -> TDM_DEV

namanjain7:
Latest code is taken from master and synced with TDM_DEV branch for new board.
Extra changes added:
1) new defconfig for mems sensor. (mems_audio)
2) Fix build error in amebasmart files.
3) Removed unnecessary build files located in image tool folder.

This PR doesn't include new driver changes. 
This PR is only for sync and removal of unnecessary files and fix build error.

Kindly use mems_audio to build for mems sensor.


jeongarmy:
How about removing TDM_DEV branch and making TDB_DEV (branch out) with the latest codes?

namanjain7:
> How about removing TDM_DEV branch and making TDB_DEV (branch out) with the latest codes?

Hello @jeongarmy ,
yes, it would be better if we want to review code properly before merging.
but, I dont' think that I have the permission to create new branch in Samsung/TizenRT.
so, what should I do?

PR #6889, opened 2025-07-18
apps/examples/testcase/ta_tc/security/utc: Add GCM testcases

jylee9613:
Add GCM testcases of AAD and TAG validation.
1. Check AAD mismatch
2. Check TAG mismatch
3. Encrypt/decrypt success if AAD is null


jeongarmy (approved):
lgtm

PR #6892, opened 2025-07-22
external/libcxx: Fix linkage error and implementation error of math

pcs1265:
This resolves the linkage error(multiple definitions) when `cmath` header is included multiple times.

and
The original implementation returned true for NaN due to flawed comparison logic.  
This replaces `(x != NAN)` with `!isnan(x)` to correctly handle NaN values.

### DIFF
--- a/external/include/libcxx/cmath
+++ b/external/include/libcxx/cmath
@@ -110,14 +110,14 @@ namespace std
     return (((__x) == INFINITY) || ((__x) == -INFINITY));
   }
 
-  inline bool isfinite(float __x)
-  { 
-    return (!(isinf(__x)) && (__x != NAN));
-  }
-
   inline bool isnan(float __x)
   { 
     return ((__x) != (__x));
+  }
+
+  inline bool isfinite(float __x)
+  { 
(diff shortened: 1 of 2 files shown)


jeongarmy:
Wasn't it occured because there was only one place where it is used in the public? Could you check why are results of public and ecode different?

pcs1265:
> Wasn't it occured because there was only one place where it is used in the public? Could you check why are results of public and ecode different?

I think so.
There is no error when building in public.
I guess the link error occurs in ecode due to integration between packages during the final link.

external/include/libcxx/cmath
```diff
@@ -150,20 +151,21 @@ namespace std
...
 
-  bool isfinite(double __x)
+  inline bool isnan(double __x)
   { 
-    return (!(isinf(__x)) && (__x != NAN));
+    return ((__x) != (__x));
   }
-
-  bool isnan(double __x)
+
```
jeongarmy: Please remove all empty spaces in this file.

jeongarmy (approved):
LGTM

PR #6894, opened 2025-07-24
drivers/compression: fix decompress ioctl

seokhun-eom24:
Add a break statement to the COMPIOC_DECOMPRESS case in comp_ioctl.
This fix ensures proper handling of decompression requests.

### DIFF
--- a/os/drivers/compression/compress.c
+++ b/os/drivers/compression/compress.c
@@ -158,6 +158,7 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		if (ret == ENOMEM) {
 			bcmpdbg("Output buffer allocated is not sufficient\n");
 		}
+		break;
 	case COMPIOC_FCOMP_INIT:
 		if ((char *)arg == NULL) {
 			return -EINVAL;


jeongarmy (approved):
lgtm

os/drivers/compression/compress.c
```diff
@@ -158,6 +158,7 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		if (ret == ENOMEM) {
```
eunwoo-nam, line 158: error return valuse must be negative.
Could you please check return values of decompress_block and change to negative?

os/drivers/compression/compress.c
```diff
@@ -158,6 +158,7 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
 			bcmpdbg("Output buffer allocated is not sufficient\n");
 		}
+		break;
```
Taejun-Kwon, line 161: it's not related to this pr but please check code line where return just ERROR, it is wrong. each of ioctl should return one of errno.

Taejun-Kwon, line 161: it's not related to this pr but please check code line where return just ERROR, it is wrong. each of ioctl should return one of errno.

os/drivers/compression/compress.c
seokhun-eom24: In the decompress_block function the return of ENOMEM indicates that the output buffer was insufficient, resulting in only a portion of the data being decompressed while the remainder remained compressed. 
In the LZMA algorithm this scenario is designed to return OK, whereas the MINIZ algorithm returns a Z_BUF_ERROR.
Would it be appropriate to modify this to return an error instead?

os/drivers/compression/compress.c
seokhun-eom24: Thank you for review. I'll make another PR for it.

Sambhavi-1 (approved):
PR fix the required decompression reqin comp_ioctl

os/drivers/compression/compress.c
seokhun-eom24: I made a new PR from your comment.
https://github.com/Samsung/TizenRT/pull/6911

os/drivers/compression/compress.c
seokhun-eom24: I fixed it from new PR.
https://github.com/Samsung/TizenRT/pull/6911

PR #6895, opened 2025-07-24
build/tools/amebasmart: Fix hardfault when turn on bridge mode with CSI

ZivLow:
- Fix hardfault issue about turning on bridge mode with CSI
- Crash was caused by accessing a null pointer when concurrent mode was started while csi is running. Add condition to handle such case
- This change does not affect wifi STA mode

Verification:
Verified with TizenLite5.
Working TizenLite5 bridge between Mobile phone and AP (intenetm utube)


hk-gwak (approved):
Verified with TizenLite5, too.

PR #6897, opened 2025-07-24
os/kernel/log_dump: fix log_dump_save buffer full handling

seokhun-eom24:
**log_dump**
Previously when the entire uncompressed buffer was full, log dumps were skipped. 
This commit ensures that when compression is available, the entire buffer is compressed in such scenarios.
Modify `compress_full_bufs` function to compress all uncompressed buffer.

**log_dump example**
1. Add logging with sched_lock to verify log saving function.
2. Add functions to handle decompression and printing of decompressed logs

### DIFF
--- a/apps/examples/log_dump/log_dump_main.c
+++ b/apps/examples/log_dump/log_dump_main.c
@@ -22,15 +22,27 @@
 
 #include <tinyara/config.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <sys/ioctl.h>
 #include <tinyara/log_dump/log_dump.h>
+#include <tinyara/compression.h>
+
+/****************************************************************************
+ * Private Function Prototypes
+ ****************************************************************************/
+
+static int decompress_log_data(unsigned char *comp_buf, int total_read, unsigned char **decomp_buf, size_t *decomp_size);
+
+static void print_decompressed_data(unsigned char *decomp_buf, size_t decomp_size);
 
 /****************************************************************************
  * Pre-processor Definitions
  ****************************************************************************/
 
 #define READ_BUFFER_SIZE	1024	/* user configurable read size */
+#define COMP_DRVPATH		"/dev/compress"
 
 /****************************************************************************
  * log_dump_main
@@ -46,6 +58,9 @@ int log_dump_main(int argc, char *argv[])
 	int total_read = 0;
 	char buf[READ_BUFFER_SIZE];
 	int fd = OPEN_LOGDUMP();
+	unsigned char *comp_buf = NULL;
+	unsigned char *decomp_buf = NULL;
+	size_t decomp_size = 0;
 
 	if (fd < 0) {
 		printf("Please ensure that procfs is mounted\n");
@@ -56,6 +71,7 @@ int log_dump_main(int argc, char *argv[])
 	 * To test start, intentionally add to stop. */
 	if (STOP_LOGDUMP_SAVE(fd) < 0) {
 		printf("Failed to stop log dump, errno %d\n", get_errno());
+		ret = ERROR;
 		goto errout;
 	}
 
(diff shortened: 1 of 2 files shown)


os/kernel/log_dump/log_dump.c
```diff
@@ -515,6 +515,21 @@ int log_dump_save(char ch)
 			return compress_full_bufs();
 		}
 	}
+	else {
+		// This else statement is entered only when enitre uncompressed buffer is full.
+		// If compression is available, reset the cursor and compress the entire uncompressed buffer.
+		// Otherwise, discard the character.
+		if (sched_self()->sched_priority > CONFIG_LOG_DUMP_PRIO || sched_lockcount()) {
```
eunwoo-nam: Could you plaesa check below condition?
```
static int compress_full_bufs(void)
		if (IS_KMM_LOCKED()) {
			return LOG_DUMP_MEM_FAIL;
		}
```

seokhun-eom24: Fixed. Thank you.

os/kernel/log_dump/log_dump.c
```diff
@@ -515,6 +515,21 @@ int log_dump_save(char ch)
 			return compress_full_bufs();
 		}
 	}
+	else {
+		// This else statement is entered only when enitre uncompressed buffer is full.
+		// If compression is available, reset the cursor and compress the entire uncompressed buffer.
+		// Otherwise, discard the character.
+		if (sched_self()->sched_priority > CONFIG_LOG_DUMP_PRIO || sched_lockcount() || IS_KMM_LOCKED()) {
+			return LOG_DUMP_OK;
```
eunwoo-nam: it should be returning LOG_DUMP_FAIL

os/kernel/log_dump/log_dump.c
```diff
@@ -515,6 +515,21 @@ int log_dump_save(char ch)
...
+		}
+
+		uncomp_curbytes = 0;	/* reset cursor */
```
eunwoo-nam: it already 0, we don't need to set again

os/kernel/log_dump/log_dump.c
```diff
@@ -515,6 +515,21 @@ int log_dump_save(char ch)
...
+		int temp = uncomp_idx++;
+		if (uncomp_idx == CONFIG_LOG_DUMP_NUMBUFS) {
+			uncomp_idx = 0;
+		}
```
eunwoo-nam: this code is changing buffer index,
but all buffer is full state. so do not change it.
it will make wrong log orders

os/kernel/log_dump/log_dump.c
seokhun-eom24: Yes that's right.
I wrote the code to explicitly reset cursor, but I will delete it.

os/kernel/log_dump/log_dump.c
seokhun-eom24: In the `compress_full_bufs` function, due to the `while (i != uncomp_idx)` condition `uncomp_idx` is not compressed and always enter else case.
So I wrote code to update `uncomp_idx` in `log_dump_save`.
However it will make wrong log orders as you mentioned.
I'll solve this problem with alternative approach.

os/kernel/log_dump/log_dump.c
seokhun-eom24: Before applying this PR the else case returned OK. So write code like this.
But I'll modify it to return FAIL like your comment.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -79,20 +101,129 @@ int log_dump_main(int argc, char *argv[])
...
 
 	CLOSE_LOGDUMP(fd);
 
+	if (decompress_log_data(comp_buf, total_read, &decomp_buf, &decomp_size) < 0) {
+		goto errout;
+	}
+
+	print_decompressed_data(decomp_buf, decomp_size);
+	
+	if (comp_buf) {
```
jeongarmy: free should be errout as well.

seokhun-eom24: Fixted to free in errout case too.

apps/examples/log_dump/log_dump_main.c
```diff
@@ -79,20 +101,134 @@ int log_dump_main(int argc, char *argv[])
...
 
 errout:
 	CLOSE_LOGDUMP(fd);
+	if (comp_buf) {
+		free(comp_buf);
+	}
+	if (decomp_buf) {
+		free(decomp_buf);
+	}
 	return -1;
```
eunwoo-nam: we can remove duplicated code
```c
	ret = decompress_log_data(comp_buf, total_read, &decomp_buf, &decomp_size);
	if (ret != OK) {
		goto errout;
	}


errout:
	CLOSE_LOGDUMP(fd);
	if (comp_buf) {
		free(comp_buf);
	}
	if (decomp_buf) {
		free(decomp_buf);
	}
	return ret;
```

eunwoo-nam: we can remove duplicated code like below

```c
	ret = decompress_log_data(comp_buf, total_read, &decomp_buf, &decomp_size);
	if (ret != OK) {
		goto errout;
	}


errout:
	CLOSE_LOGDUMP(fd);
	if (comp_buf) {
		free(comp_buf);
	}
	if (decomp_buf) {
		free(decomp_buf);
	}
	return ret;
```

seokhun-eom24: Updated. Thank you.

jeongarmy (approved):
lgtm

PR #6899, opened 2025-07-24
os/kernel/log_dump: Add crash log flash save

Kimhyunjong121:
Developed a new feature to store logs during assert situations. It only operates when a crash occurs in the user space.

The following modifications are required:
1. Avoid using double vsnprintf.
2. Check available space in the file system before wrting.
3. Check available heap memory before compression.
4. Store compressed data in the heap instead of the data section.
5. Add a unique identifier or delimiter to the file name.
6. Implement a retry count mechanism.
7. Ensure logs from functions like __alert() are also captured.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_assert.c
+++ b/os/arch/arm/src/armv7-a/arm_assert.c
@@ -611,6 +611,12 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = kernel_assert_location;
 	}
 
+#ifdef CONFIG_BINMGR_RECOVERY
+	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
+		/* start assert log saving */
+		set_assert_situation();
+	}
+#endif
 #ifdef CONFIG_SMP
 	/* Pause all other CPUs to avoid mix up of logs while printing assert logs */
 	up_cpu_pause_all();
@@ -654,6 +660,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
 		/* Recover user fault through binary manager */
 		binary_manager_recover_userfault();
+		state = NORMAL_STATE; 
 	} else
 #endif
 	{
--- a/os/include/syslog.h
+++ b/os/include/syslog.h
@@ -262,6 +262,8 @@ int lowsyslog(int priority, FAR const char *format, ...);
  */
 int lowvsyslog(int priority, FAR const char *format, va_list ap);
 
+
+
 #else
 
 #ifdef CONFIG_CPP_HAVE_VARARGS
--- a/os/include/tinyara/log_dump/log_dump_internal.h
+++ b/os/include/tinyara/log_dump/log_dump_internal.h
@@ -64,4 +64,59 @@ int log_dump_get_size(void);
  ****************************************************************************/
 int log_dump(int argc, char *argv[]);
 
+/****************************************************************************
+ * Description:
+ *   This is used to store logs that are output via lldbg() during a user crash situation into a buffer.
+ *
+ ****************************************************************************/
+void assert_log_to_buffer(FAR const char *fmt, va_list ap);
+
+
+/****************************************************************************
+ * Description:
+ *   This is used to enable assert handling.
+ *
+ ****************************************************************************/
+int set_assert_situation();
+
+/****************************************************************************
+ * Description:
+ *   This is used to unable assert handling.
+ *
+ ****************************************************************************/
+int set_notassert_situation();
+
+/****************************************************************************
+ * Description:
+ *   This is used to save not compressed assert log to file.
+    file name fixed: /mnt/assert_logsave.txt -> needs to be modified.
+ ****************************************************************************/
+int assert_log_to_file();
+
+/****************************************************************************
+ * Description:
+ *   This is used to compress assert log.
+    it's stored in the data section, but it needs to be changed to be stored in the heap section.
+ ****************************************************************************/
+void assert_log_compress();
+
+/****************************************************************************
+ * Description:
+ *   This is used to save compressed assert log to file.
+    file name fixed: /mnt/assert_logsave_zip -> needs to be modified.
+ ****************************************************************************/
+int compressed_assert_log_to_file();
+
+/****************************************************************************
+ * Description:
+ *   This is used to read compressed assert log from file to buf.
+ ****************************************************************************/
+int compressed_assert_log_read();
+
+/****************************************************************************
+ * Description:
+ *   This is used to decompress buf and print decompressed log.
+ ****************************************************************************/
+int decompress_and_print();
+
 #endif							/* __LOG_DUMP_INTERNAL_H */
--- a/os/kernel/binary_manager/binary_manager_recovery.c
+++ b/os/kernel/binary_manager/binary_manager_recovery.c
@@ -346,6 +346,13 @@ void binary_manager_recovery(int bin_idx)
 		goto reboot_board;
 	}
 #endif
+	//for now, these must be executed in this exact order
+	set_notassert_situation();
+	assert_log_to_file();
+	assert_log_compress();
+	compressed_assert_log_to_file();
+	compressed_assert_log_read();
+	decompress_and_print();
 	/* Create loader to reload binary */
 	ret = binary_manager_execute_loader(LOADCMD_RELOAD, bin_idx);
 	if (ret == OK) {
--- a/os/kernel/log_dump/log_dump.c
+++ b/os/kernel/log_dump/log_dump.c
@@ -70,6 +70,9 @@
 #define LOG_DUMP_COMP_CHECK_RETRY_CNT 		(CONFIG_LOG_DUMP_HANG_CHECK_SEC * USEC_PER_SEC / LOG_DUMP_COMP_CHECK_RETRY_USEC)
 #endif
 
+ #define ASSERT_LOG_BUF_SIZE 8192 /* Buffer size for logs printed during an assert */
+
+
 static bool is_started_to_save;
 
 /****************************************************************************
@@ -118,6 +121,17 @@ static unsigned char *last_comp_block;
 static size_t last_comp_block_size;
 static int last_comp_block_ptr;
 static bool compress_last_block;
+
+/* Structures used to save compressed log data*/
+static unsigned char compressed_buf[ASSERT_LOG_BUF_SIZE]; 
+static unsigned int compressed_buf_size;
+static char assert_log[ASSERT_LOG_BUF_SIZE];
+static int assert_situation=0;
+static int assert_log_index=0;
+static int log_len=0;
+
+/* Structures used to decompress and print compressed log data*/
+static char tm[ASSERT_LOG_BUF_SIZE];
 
 /****************************************************************************
  * Private Functions
@@ -569,6 +583,14 @@ size_t log_dump_read(FAR char *buffer, size_t buflen)
 	return ret;
 }
 
+
+
+
+
+
+
+
+
 int log_dump(int argc, char *argv[])
 {
 	struct mq_attr attr;
@@ -612,4 +634,135 @@ int log_dump(int argc, char *argv[])
 			compress_full_block = false;
 		}
 	}
+}
+
+
+void assert_log_init(){
+	assert_situation=0;
+	assert_log_index=0;
+}
+int assert_log_save(char ch){
+	if(assert_log_index<ASSERT_LOG_BUF_SIZE) assert_log[assert_log_index++]=ch;
+}
+int get_assert_log_save_size(){
+	return log_len;
+}
+int set_assert_situation(){
+	lldbg("\n\n\n===============================FROM THIS, lldbg LOG WILL BE SAVED!!================================\n\n\n\n");
+	assert_situation=1;
+}
+int set_notassert_situation(){
+	lldbg("\n\n\n=============================================TO THIS===============================================\n\n\n");
+	
+	assert_situation=0;
+}
+int get_assert_log(char *buf,int size){
+	int i;
+	for(i=0;i<get_assert_log_save_size;i++){
+		if(i>=size) break;
+		buf[i]=assert_log[i];
+	}
+	return i;
+}
+void append_assert_log(const char *fmt, va_list ap){
+	if (log_len>=ASSERT_LOG_BUF_SIZE - 1){
+		return ;
+	}
+	size_t remain=ASSERT_LOG_BUF_SIZE - log_len-1;
+	int written = vsnprintf(assert_log + log_len, remain+1,fmt,ap);
+	if(written<0) return;
+	if((size_t)written > remain){
+		log_len = ASSERT_LOG_BUF_SIZE -1 ;
+	}
+	else{
+		log_len +=(size_t)written;
+	}
+	assert_log[log_len]='\0';
+}
+void assert_log_to_buffer(FAR const char *fmt, va_list ap){
+	if(assert_situation){
+		va_list ap2;
+		va_copy(ap2,ap);
+		append_assert_log(fmt,ap2);
+		va_end(ap2);
+	}
+}
+
+
+
+int assert_log_to_file(){
+	lldbg("\n\n\n===============assert log file saving start==================\n\n\n");
+	lldbg("\nlog_save_size is %d\n",get_assert_log_save_size());
+	int fd=open("/mnt/assert_logsave.txt",O_WRONLY|O_CREAT|O_TRUNC);
+	if(fd<0){
+		lldbg("file open failed\n");
+		return -1;
+	}
+	lldbg("file open success\n");
+	int log_size=write(fd,assert_log,strlen(assert_log));
+	close(fd);
+	if(log_size<0){
+		lldbg("file write failed\n");
+		return -1;
+	}
+	lldbg("file write success, count:%d\n",log_size);
+	lldbg("\n\n\n===============assert log file saving end=====================\n\n\n");
+	return log_size;
+}
+void assert_log_compress(){
+
+	lldbg("\n\n================compress start==============\n\n");
+	compressed_buf_size=ASSERT_LOG_BUF_SIZE;
+	compress_block(compressed_buf,&compressed_buf_size,assert_log,ASSERT_LOG_BUF_SIZE);
+	lldbg("compressed buf size: %d\n",compressed_buf_size);
+	lldbg("\n====================compress end =====================\n");
+		
+}
+int compressed_assert_log_to_file(){
+	lldbg("\n\n================compress file write start==============\n\n");
+	int fd=open("/mnt/ssert_logsave_zip",O_WRONLY|O_CREAT|O_TRUNC);
+	if(fd<0){
+		lldbg("file open failed\n");
+		return -1;
+	}
+	lldbg("file open success\n");
+	int ret=write(fd,compressed_buf,compressed_buf_size);
+	if(ret<0){
+		lldbg("file write failed\n");
+		return -1;
+	}
+	close(fd);
+	lldbg("write count: %d\n",ret);
+	lldbg("\n\n=====================compress file write end ===================\n\n");
+	return ret;
+}
+int compressed_assert_log_read(){
+	lldbg("\n========================== compressed_assert_log_read start======================\n");
+	for(int i=0;i<ASSERT_LOG_BUF_SIZE;i++)compressed_buf[i]=0;
+	int fd=open("/mnt/assert_logsave_zip",O_RDONLY);
+		if(fd<0){
+			lldbg("file open failed\n");
+			return -1;
+		}
+		lldbg("file open success\n");
+	int ret=read(fd,compressed_buf,sizeof(compressed_buf));
+	if(ret<0){
+		lldbg("file read failed\n");
+		return -1;
+	}
+	lldbg("file read successed, count:%d\n",ret);
+	close(fd);
+	lldbg("\n========================== compressed_assert_log_read end========================\n");
+	compressed_buf_size=ret;
+	return ret;
+
+}
+int decompress_and_print(){
+		lldbg("\n========================== compressed data decompress and print start======================\n");
+		int tm_size=ASSERT_LOG_BUF_SIZE;
+		decompress_block(tm,&tm_size,compressed_buf,&compressed_buf_size);
+		lldbg("decompressed size: %d\n",tm_size);
+		tm[ASSERT_LOG_BUF_SIZE-1]='\0';
+		lldbg("\n%s\n",tm);
+		lldbg("\n========================== compressed data decompress and print end======================\n");
 }


os/kernel/log_dump/log_dump.c
```diff
@@ -119,6 +122,17 @@ static size_t last_comp_block_size;
...
+/* Structures used to save compressed log data*/
+static unsigned char compressed_buf[ASSERT_LOG_BUF_SIZE]; 
+static unsigned int compressed_buf_size;
+static char assert_log[ASSERT_LOG_BUF_SIZE];
+static int assert_situation=0;
+static int assert_log_index=0;
+static int log_len=0;
+
+/* Structures used to decompress and print compressed log data*/
+static char tm[ASSERT_LOG_BUF_SIZE];
```
eunwoo-nam: It appears that this feature is not utilizing the same buffer as the existing log_dump.
 In that case, please proceed with adding a completely new file.

os/include/tinyara/log_dump/log_dump_internal.h
```diff
@@ -64,4 +64,59 @@ int log_dump_get_size(void);
  ****************************************************************************/
 int log_dump(int argc, char *argv[]);
 
+/****************************************************************************
+ * Description:
+ *   This is used to store logs that are output via lldbg() during a user crash situation into a buffer.
+ *
+ ****************************************************************************/
+void assert_log_to_buffer(FAR const char *fmt, va_list ap);
```
eunwoo-nam: which code is calling this?

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -611,6 +611,12 @@ void up_assert(const uint8_t *filename, int lineno)
 		asserted_location = kernel_assert_location;
 	}
 
+#ifdef CONFIG_BINMGR_RECOVERY
+	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
+		/* start assert log saving */
+		set_assert_situation();
```
eunwoo-nam: our crash code is started on fault isr like arm_dataabort.c
the fault isr logs also matter.

And do we need to make start log saving point?

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -346,6 +346,13 @@ void binary_manager_recovery(int bin_idx)
 		goto reboot_board;
 	}
 #endif
+	//for now, these must be executed in this exact order
+	set_notassert_situation();
+	assert_log_to_file();
+	assert_log_compress();
+	compressed_assert_log_to_file();
+	compressed_assert_log_read();
+	decompress_and_print();
```
eunwoo-nam: the functions are calling here only,
we don't need to call theae seperatly.

How about call onec api here?

```c
log_dump_crash_save_to_file()
```

os/kernel/log_dump/log_dump.c
```diff
@@ -613,3 +635,134 @@ int log_dump(int argc, char *argv[])
...
+}
+void assert_log_compress(){
+
+	lldbg("\n\n================compress start==============\n\n");
+	compressed_buf_size=ASSERT_LOG_BUF_SIZE;
+	compress_block(compressed_buf,&compressed_buf_size,assert_log,ASSERT_LOG_BUF_SIZE);
+	lldbg("compressed buf size: %d\n",compressed_buf_size);
+	lldbg("\n====================compress end =====================\n");
+		
+}
```
eunwoo-nam: The code is exculed during crash recovery.

So, we need check time for compress.
If It is long, not compress is better.

for it, How about make configuration compress or not.

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -346,6 +346,13 @@ void binary_manager_recovery(int bin_idx)
...
 	/* Create loader to reload binary */
 	ret = binary_manager_execute_loader(LOADCMD_RELOAD, bin_idx);
```
eunwoo-nam, line 360: we need to make option to system reboot here.

ewoodev (commented):
Please check coding rule

os/kernel/log_dump/log_dump.c
```diff
@@ -613,3 +635,134 @@ int log_dump(int argc, char *argv[])
...
+int assert_log_to_file(){
+	lldbg("\n\n\n===============assert log file saving start==================\n\n\n");
+	lldbg("\nlog_save_size is %d\n",get_assert_log_save_size());
+	int fd=open("/mnt/assert_logsave.txt",O_WRONLY|O_CREAT|O_TRUNC);
+	if(fd<0){
+		lldbg("file open failed\n");
+		return -1;
+	}
+	lldbg("file open success\n");
+	int log_size=write(fd,assert_log,strlen(assert_log));
```
Taejun-Kwon: below is better code.
```
	int len = strlen(assert_log);
	int ret = write(fd, assert_log, len);
	close(fd);
	if (ret != len) {
		lldbg("file write failed ret : %d len : %d errno : %d\n", ret, len, errno);
		return -1;
	}
```

os/kernel/log_dump/log_dump.c
```diff
@@ -613,3 +635,134 @@ int log_dump(int argc, char *argv[])
...
+	int fd=open("/mnt/ssert_logsave_zip",O_WRONLY|O_CREAT|O_TRUNC);
+	if(fd<0){
+		lldbg("file open failed\n");
+		return -1;
+	}
+	lldbg("file open success\n");
+	int ret=write(fd,compressed_buf,compressed_buf_size);
+	if(ret<0){
+		lldbg("file write failed\n");
+		return -1;
```
Taejun-Kwon: file should be closed here.

os/kernel/log_dump/log_dump.c
Kimhyunjong121: I conducted an experiment where, after printing the entire log, I measured the time it takes to save the log to a file using clock(). The measurement was repeated multiple times.

1. When the log size was 5370 bytes:
-With compression-
run       run1  run2  run3   run4   run5   mean
time(s) 0.09 0.129  0.042  0.091  0.134  0.0972
compressed size: 1638 bytes

-Without compression-
run       run1  run2  run3   run4   run5   mean
time(s) 0.037  0.033 0.082 0.357 0.034  0.095


2. When the log size was 27971 bytes (It was randomly generated using rand()):
-With compression-
run                        run1   run2     run3    run4   run5   mean
time(s)                  0.046   0.374   0.046  0.136  0.093  0.139
comp size(bytes)  3596    4483    3523   3593  3591   

-Without compression-
run       run1  run2  run3   run4   run5   mean
time(s) 0.883 0.355 0.58    0.269  0.079 0.256


When the log size is small, the time taken is similar whether compression is used or not. However, as the log size increases, using compression becomes faster overall. It seems that the time spent on writing dominates more than the time spent on compression. Therefore, it may not be necessary to create a separate configuration for this.

os/kernel/log_dump/log_dump.c
```diff
@@ -613,3 +635,134 @@ int log_dump(int argc, char *argv[])
...
+	lldbg("\n\n================compress start==============\n\n");
+	compressed_buf_size=ASSERT_LOG_BUF_SIZE;
+	compress_block(compressed_buf,&compressed_buf_size,assert_log,ASSERT_LOG_BUF_SIZE);
+	lldbg("compressed buf size: %d\n",compressed_buf_size);
+	lldbg("\n====================compress end =====================\n");
+		
+}
+int compressed_assert_log_to_file(){
+	lldbg("\n\n================compress file write start==============\n\n");
+	int fd=open("/mnt/ssert_logsave_zip",O_WRONLY|O_CREAT|O_TRUNC);
```
seokhun-eom24: Use macro for `/mnt/ssert_logsave_zip` would be better.

os/kernel/log_dump/log_dump.c
```diff
@@ -613,3 +635,134 @@ int log_dump(int argc, char *argv[])
...
+}
+int assert_log_save(char ch){
+	if(assert_log_index<ASSERT_LOG_BUF_SIZE) assert_log[assert_log_index++]=ch;
+}
+int get_assert_log_save_size(){
+	return log_len;
+}
+int set_assert_situation(){
+	lldbg("\n\n\n===============================FROM THIS, lldbg LOG WILL BE SAVED!!================================\n\n\n\n");
+	assert_situation=1;
```
seokhun-eom24: You can use `assert_situation = true;`

os/include/syslog.h
```diff
@@ -262,6 +262,8 @@ int lowsyslog(int priority, FAR const char *format, ...);
  */
 int lowvsyslog(int priority, FAR const char *format, va_list ap);
 
+
```
seokhun-eom24: Unnecessary file changes

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -610,7 +614,14 @@ void up_assert(const uint8_t *filename, int lineno)
 	} else {
 		asserted_location = kernel_assert_location;
 	}
-
+#ifdef CONFIG_CRASH_LOG_SAVE
+#ifdef CONFIG_BINMGR_RECOVERY
```
eunwoo-nam: Let's check CONFIG_CRASH_LOG_SAVE only here.

you already made dependency BINMGR_RECOVERY

os/include/tinyara/crashlog_writer/crashlog_writer.h
```diff
@@ -0,0 +1,57 @@
+/****************************************************************************
+ *
+ * Copyright 2022 Samsung Electronics All Rights Reserved.
```
seokhun-eom24: 2022 -> 2025.
Apply it to other new files as well.

os/kernel/crashlog_writer/Kconfig
```diff
@@ -0,0 +1,64 @@
...
+        This saves the crash log.
+
+if CRASH_LOG_SAVE
+
+config COMPRESSION_FLAG
+    bool "COMPRESSION_FLAG"
+    default y
+    ---help---
+        Controls whether compression is applied.
+        if the value is 0 compression is not performed.
```
seokhun-eom24: It should be `n` instead of `0` since it's a boolean value.

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+		if (compressed_buf == NULL) {
+			lldbg("kmm_malloc() failed\n");
+			return -1;
+		}
+  		unsigned int compressed_buf_size = malloc_size;
+		ret = read(fd, compressed_buf, compressed_buf_size);
+		if (ret < 0) {
+			lldbg("read failed\n");
+			free(compressed_buf);
+			return -1;
```
seokhun-eom24: You can use `ERROR` instead of -1. `OK` for 0.

```c
// defined in <sys/types.h>
#define ERROR -1
#define OK 0
```

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+ ****************************************************************************/
+
+/****************************************************************************
+ * Private Variables
+ ****************************************************************************/
+static unsigned char g_crash_log[CONFIG_CRASH_LOG_BUF_SIZE];
+static int g_start_to_save = 0;
+static int g_crash_log_index = 0;
+static char g_filename[MAX_FILENAME_LEN];
+#ifndef REBOOT_AFTER_LOG_SAVE
```
jeongarmy: CONFIG_ is missing

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+* Name: is_valid_filename
+*
+* Description:
+*   This function checks whether a file in mnt/crashlog/ is a log save file based on its filename.
+*
+*************************************************************************************/
+static int is_valid_filename(const char *name, int *out_number) 
+{
+    char expected_prefix[16];
+    snprintf(expected_prefix, sizeof(expected_prefix), "crashlog_");
```
jeongarmy: Please define and use prefix value
#define XXX_PREFIX "crashlog_"

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -345,6 +349,17 @@ void binary_manager_recovery(int bin_idx)
 		bmlldbg("Failed to deactivate binary, bin idx %d\n", bin_idx);
 		goto reboot_board;
 	}
+#endif
+#ifdef CONFIG_CRASH_LOG_SAVE
```
jeongarmy: Please use defined values, not hard-coded values for better readability.

```
#define XXXX_ENABLE 1
#define XXXX_DISABLE 0
```

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+}
+
+/*************************************************************************************
+* Name: save_crash_log
+*
+* Description: 
+*   Saves the crash log to a file. 
+*   If the flag is 1, the log is saved in compressed form; if it is 0, it is saved uncompressed.
+*************************************************************************************/
+char *save_crash_log(int flag)
```
jeongarmy: If flag is used for compression only, we'd better to change it to `bool compression`.
If there will be more options for logging, we can support this with flag setting like below.
For examples,

#define FLAGS_COMPRESS (1 << 0)
#define FLAGS_XXXXX         (1 << 1)

caller
flag |= FLAGS_COMPRESS;
save_crash_log(flag);


save_crash_log(int flag)
{

if (sem->flags & FLAGS_COMPRESS) {
      ret = crash_log_compress();
} else {
     ret = crash_log_to_file();
}
}

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+		double elapsed_time = (double)(end - start) / CLOCKS_PER_SEC;
+		lldbg("\n\nElapsed time: %.6f seconds\n\n", elapsed_time);
+		int fd = open("/mnt/crashlog/result", O_CREAT | O_WRONLY | O_APPEND);
+		if (fd < 0) {
+			return g_filename;
+		}
+		char buf[100];
+#ifdef CONFIG_COMPRESSION_FLAG
+		snprintf(buf, sizeof(buf), "C: %d -> %d ", log_size, ret);
+		write(fd, buf, strlen(buf));
```
jeongarmy: Please add result checking and error handling.

os/kernel/crashlog_writer/crashlog_writer.c
```diff
@@ -0,0 +1,482 @@
...
+		if (ret < 0) {
+			lldbg("read failed\n");
+			free(compressed_buf);
+			return -1;
+		}
+		int tmp_buf_size = buf_size;
+		decompress_block(buf, &tmp_buf_size, compressed_buf, &compressed_buf_size);
+		free(compressed_buf);
+		return tmp_buf_size;
+	}
```
eunwoo-nam: we need to remove log file after read

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -345,6 +349,17 @@ void binary_manager_recovery(int bin_idx)
...
+	set_store_to_buffer_flag(0); //stop saving log to buf
+#ifdef CONFIG_COMPRESSION_FLAG
+	save_crash_log(1);
+#else
+	save_crash_log(0);
+#endif
```
eunwoo-nam: 1. If we select compress or not using CONFIG_COMPRESSION_FLAG, we don't need to have dynamic check and branch compress or not.
but, if it is for checking memory not enough state, we need to have branch dynamically.

2. and binary manager don't need to know about commpress or not, Let's move these into `save_crash_log`
```c
void save_crash_log()
{
#ifdef CONFIG_CRASH_LOG_COMPRESS
    /* check memory largest chunk */
    if (memory enough) {
        /* compress log */
        /* saving compressed log into fs */
    } 
#ifdef CONFIG_CRASH_LOG_SUPPORT_MEMORY_FULL
    else {
        /* just saving log into fs */
    }
#endif
#else
/* just saving log into fs */
#endif
}
```

lib/libc/stdio/lib_lowoutstream.c
```diff
@@ -76,6 +79,11 @@
 static void lowoutstream_putc(FAR struct lib_outstream_s *this, int ch)
 {
 	DEBUGASSERT(this);
+#ifdef __KERNEL__
+#ifdef CONFIG_CRASH_LOG_SAVE
+	crash_log_to_buffer(ch);
```
eunwoo-nam: 1. This feature covers the storage of all logs output in lldbg as well as crash logs
2. This feature supports log saving to fs with compress.
 
How about module naming like ll_log_dump?

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -610,7 +614,14 @@ void up_assert(const uint8_t *filename, int lineno)
...
+	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
+		/* start assert log saving */
+		set_store_to_buffer_flag(1);
```
eunwoo-nam: the fault handler start point might not be here. like `arm_dataabort.c` 
the fault function also have log, the info matters

so in my opinion, we need to set default save on, and we need to stop and start between specific section menually

lib/libc/stdio/lib_lowoutstream.c
```diff
@@ -76,6 +79,11 @@
 static void lowoutstream_putc(FAR struct lib_outstream_s *this, int ch)
 {
 	DEBUGASSERT(this);
+#ifdef __KERNEL__
+#ifdef CONFIG_CRASH_LOG_SAVE
```
eunwoo-nam: How about below

```c
#if defined(__KERNEL__) && defined(CONFIG_CRASH_LOG_SAVE)
void crash_log_to_buffer(ch);
#else
#define crash_log_to_buffer(ch) (void)
```

This makes it possible to specify in the api header that it can only be used in kernel.

os/kernel/log_dump/lowlog_dump_save_ch.c
```diff
@@ -0,0 +1,103 @@
...
+*   init global buf
+*
+*************************************************************************************/
+void lowlog_dump_init_buf()
+{
+	for (int i = 0; i < CONFIG_LOWLOG_DUMP_BUF_SIZE; i++) {
+		g_lowlog[i] = '\0';
+	}
+	g_lowlog_index = 0;
+}
```
eunwoo-nam: do we need this? the bss is already 0.

if clear is needed, let's do in save to flash

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,337 @@
...
+	write(fd, buf, strlen(buf));
+	snprintf(buf, sizeof(buf), " (%s)\n", filenm);
+	write(fd, buf, strlen(buf));
+	close(fd);
+#ifndef CONFIG_LOWLOG_DUMP_REBOOT
+	lowlog_dump_init();
+#endif
+#endif
+	return OK;
+}
```
eunwoo-nam: Please add end of line

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,337 @@
...
+
+/*************************************************************************************
+* Name: lowlog_dump_save_uncomp
+*
+* Description:
+*   Saves the log stored in the buffer to a file.
+*************************************************************************************/
+static int lowlog_dump_save_uncomp(char *filenm)
+{
+	lldbg("\n\n\n===================================crash log file saving start==========================================\n\n\n");
```
eunwoo-nam: the log is info, Let's replace to ldpllvdbg

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,337 @@
...
+	char new_filename[MAX_FILENAME_LEN];
+	int ret = lowlog_dump_get_next_filename(new_filename);
+	if (ret < 0) {
+		lldbg("get_next_filename() failed\n");
```
eunwoo-nam: Let's replace ldplldbg

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,337 @@
...
+* Description:
+*   This function generates the filename for a lowlog file
+*   ex)
+*   if compression is enabled -> lowlog_dumpXXX
+*   if compression is disabled -> lowlog_dump_compXXX
+*************************************************************************************/
+static int lowlog_dump_get_next_filename(char *new_filename) 
+{
+    DIR *dir = opendir(DIR_PATH);
+    if (!dir) {
```
eunwoo-nam: the coding rule indentation Tab Size:4

os/arch/arm/src/armv7-a/arm_dataabort.c
```diff
@@ -114,7 +114,7 @@ static inline void print_dataabort_detail(uint32_t *regs, uint32_t dfar, uint32_
 
 #ifdef CONFIG_PAGING
 uint32_t *arm_dataabort(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
-{
+{
```
eunwoo-nam: please revert unnecessary change

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -610,7 +610,6 @@ void up_assert(const uint8_t *filename, int lineno)
 	} else {
 		asserted_location = kernel_assert_location;
 	}
-
```
eunwoo-nam: please revert unnecessary change

os/arch/arm/src/armv7-a/arm_dataabort.c
```diff
@@ -199,7 +199,10 @@ uint32_t *arm_dataabort(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
 
 #include "section_config.h"
 SRAMDRAM_ONLY_TEXT_SECTION uint32_t *arm_dataabort(uint32_t *regs, uint32_t dfar, uint32_t dfsr)
-{
+{	
+#ifdef CONFIG_LOWLOG_DUMP
+	lowlog_dump_set_store_flag(1);
```
eunwoo-nam: it looks missed remove, pelease check

os/kernel/log_dump/lowlog_dump_pause.c
```diff
@@ -0,0 +1,111 @@
...
+
+/*************************************************************************************
+* Name: lowlog_dump_pause
+*
+* Description: 
+*   pause saving lldbg to buf.
+*************************************************************************************/
+void lowlog_dump_pause(void)
+{
+	ldpllvdbg("\n\n\n==========================================================TO THIS!!==========================================================\n\n\n\n");
```
eunwoo-nam: what's mean `TO THIS`???

os/arch/arm/src/armv7-a/arm_dataabort.c
eunwoo-nam: do we need this here?

os/kernel/log_dump/lowlog_dump.h
```diff
@@ -0,0 +1,72 @@
...
+ ****************************************************************************/
+typedef struct {
+    int number;
+    char name[MAX_FILENAME_LEN];
+} FileEntry;
+
+/****************************************************************************
+ * Function Prototypes
+ ****************************************************************************/
+int lowlog_dump_get_retry_count(void);
```
eunwoo-nam: it looks not be removed.

eunwoo-nam: it looks not removed.

os/arch/arm/src/armv7-a/arm_assert.c
```diff
@@ -654,6 +653,7 @@ void up_assert(const uint8_t *filename, int lineno)
 	if (IS_FAULT_IN_USER_SPACE(asserted_location)) {
 		/* Recover user fault through binary manager */
 		binary_manager_recover_userfault();
+		state = NORMAL_STATE;
```
eunwoo-nam: Please revert unnecessary change

os/kernel/log_dump/lowlog_dump.h
```diff
@@ -0,0 +1,71 @@
...
+#include <tinyara/config.h>
+
+#ifdef CONFIG_LOWLOG_DUMP
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <debug.h>
```
eunwoo-nam, line 35: Please remove unnecessary including header

os/kernel/binary_manager/binary_manager_recovery.c
```diff
@@ -345,6 +349,12 @@ void binary_manager_recovery(int bin_idx)
 		bmlldbg("Failed to deactivate binary, bin idx %d\n", bin_idx);
 		goto reboot_board;
 	}
+#endif
+#ifdef CONFIG_LOWLOG_DUMP
+	lowlog_dump_save();
+#ifdef CONFIG_LOWLOG_DUMP_REBOOT
+	binary_manager_reset_board(REBOOT_SYSTEM_BINARY_RECOVERYFAIL);
```
eunwoo-nam, line 356: Reboot Reason should not be overwritten here.
in the time, we already writed reason reason about cause of crash.
```c
/* Do not overwrite reboot reason here */ 
boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
```

os/kernel/log_dump/lowlog_dump_ch.c
```diff
@@ -0,0 +1,104 @@
...
+*
+* Description: 
+*   Stores the crash log into the buffer.
+*   This is used to store logs that are output via lldbg() during a user crash situation into a buffer.
+*************************************************************************************/
+void lowlog_dump_save_ch(char ch)
+{
+	if (lowlog_dump_get_store_flag() && g_lowlog_index < CONFIG_LOWLOG_DUMP_BUF_SIZE) {
+		g_lowlog[g_lowlog_index++] = ch;
+	}
```
eunwoo-nam, line 102: In the previous review, we decided to turn on the log save from the start of booting.
this is because we need to call the start function to all asset occurrence codes in order to turn on log storage in case of asset occurrence.

However, for this to work properly, the g_lowlog buffer must be implemented as a ring buffer.
This is already full due to boot log, so it will not be able to save the assert log.

```suggestion
  if (lowlog_dump_get_store_flag()) {
      g_lowlog[g_lowlog_tail] = ch;
      g_lowlog_tail = (g_lowlog_tail + 1) & BUFFER_MASK;

      if (g_lowlog_tail == g_lowlog_head) {
          g_lowlog_head = (g_lowlog_head + 1) & BUFFER_MASK;
      }
  }
```

os/kernel/log_dump/lowlog_dump_pause.c
```diff
@@ -0,0 +1,110 @@
...
+*
+* Description:
+*   init global variables.
+*
+*************************************************************************************/
+void lowlog_dump_init()
+{
+    lowlog_dump_init_buf();
+    lowlog_dump_init_flag();
+}
```
eunwoo-nam, line 61: the function is not called anywhere
If it is unnecessary, please remove.

os/kernel/log_dump/lowlog_dump_read.c
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    closedir(dir);
+
+	snprintf(most_recent_file_name, MAX_FILENAME_LEN, "%s/%s", DIR_PATH, temp_name);
+	ldpllvdbg("most_recent_file_name is %s\n",most_recent_file_name);
+	lowlog_dump_read(most_recent_file_name, buf, buf_size);
+    return OK;
+}
```
eunwoo-nam, line 142: please add last line

os/kernel/log_dump/lowlog_dump_resume.c
```diff
@@ -0,0 +1,51 @@
...
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <debug.h>
+
+#include <tinyara/clock.h>
+#include "lowlog_dump.h"
+#include <dirent.h>
+#include <ctype.h>
+#include <fcntl.h>
```
eunwoo-nam, line 36: Please remove unnecessary including header

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+        if (lowlog_dump_is_valid_filename(entry->d_name, &num) == 0K) {
+            strncpy(entries[count].name, entry->d_name, sizeof(entries[count].name));
+            entries[count].number = num;
+            count++;
+            if (num > max_number) {
+				max_number = num;
+			}
+			if (num < min_number) {
+				min_number = num;
+			}
```
eunwoo-nam, line 130: Your logic seems to be trying to prevent repeated crash storage using max number.

if you do not need min number, please remove it

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+					return ERROR;
+				}
+			}
+		}
+		else {
+			ldpllvdbg("opendir failed\n");
+			return ERROR;
+		}
+    }
+    FileEntry entries[CONFIG_LOWLOG_DUMP_MAX_FILES];
```
eunwoo-nam, line 114: we don't need to store each file datas using array

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+    }
+    else if (name_len > pre_filename_uncomp_len && !strncmp(name, PRE_FILENAME_UNCOMP, pre_filename_uncomp_len)) {
+        for (int i = pre_filename_uncomp_len; i < name_len; i++) {
+            if (!isdigit(name[i])) {
+                return ERROR;
+            }
+        }
+        *out_number = atoi(name + pre_filename_uncomp_len);
+        return OK;
+    }
```
eunwoo-nam, line 80:
```c
#ifdef CONFIG_LOWLOG_DUMP_COMPRESS
	ret = lowlog_dump_compress_and_save(filenm);
#else
	ret = lowlog_dump_save_uncomp(filenm);
#endif
```
If compression is determined by CONFIG_LOWLOG_DUMP_COMPRESS, we do not need to check both of these here.

```suggestion
#ifdef CONFIG_LOWLOG_DUMP_COMPRESS
#define PRE_FILENAME "lowlog_dump_comp"
#else
#define PRE_FILENAME "lowlog_dump"
#endif
    int pre_filename_len = strlen(PRE_FILENAME);
    int name_len = strlen(name);
    if (name_len > pre_filename_len && !strncmp(name, PRE_FILENAME, pre_filename_len)) {
        for (int i = pre_filename_len; i < name_len; i++) {
            if (!isdigit(name[i])) {
                return ERROR;
            }
        }
        *out_number = atoi(name + pre_filename_len);
        return OK;
    }

File: os/kernel/log_dump/lowlog_dump_save.c [hunk 46]
```diff
@@ -0,0 +1,318 @@
...
+	}
+	char tmp[64];
+	snprintf(new_filename, MAX_FILENAME_LEN, "%s/", DIR_PATH);
+
+	int path_len = strlen(new_filename);
+#ifdef CONFIG_LOWLOG_DUMP_COMPRESS
+	snprintf(tmp, MAX_FILENAME_LEN - path_len - 1, "%s%d", PRE_FILENAME_COMP, max_number + 1);
+#else
+	snprintf(tmp, MAX_FILENAME_LEN - path_len - 1, "%s%d", PRE_FILENAME_UNCOMP, max_number + 1);
+#endif
```
eunwoo-nam, line 149: please check my above comment

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+		ldpllvdbg("file write failed\n");
+		close(fd);
+		return ERROR;
+	}
+	close(fd);
+	snprintf(filenm, MAX_FILENAME_LEN, "%s", new_filename);
+	ldpllvdbg("write count: %d\n", ret);
+	ldpllvdbg("\n\n==========================================================compress file write end===========================================================\n\n");
+	return ret;
+}
```
eunwoo-nam, line 221: the two function are almost same

we can reduce duplicated code like below

```c
static int lowlog_dump_save_to_file(char *log_buf, size_t size)
{
    /* file save codes */
}

#ifdef CONFIG_LOWLOG_DUMP_COMPRESS
static int lowlog_dump_save_to_file_with_comp(char *log_buf, size_t size)
{
  /* compress codes */
  return lowlog_dump_save_to_file(compressed_buf, compressed_buf_size);
}
#endif

#ifdef CONFIG_LOWLOG_DUMP_COMPRESS
	ret = lowlog_dump_save_to_file_with_comp(lowlog_dump_get_buf(), lowlog_dump_get_size());
#else
	ret = lowlog_dump_save_to_file(lowlog_dump_get_buf(), lowlog_dump_get_size());
#endif
```

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+/*************************************************************************************
+* Name: lowlog_dump_compress_and_save
+*
+* Description: 
+*   Compresses the log stored in the buffer and save to file.
+*************************************************************************************/
+static int lowlog_dump_compress_and_save(char *filenm)
+{
+	ldpllvdbg("\n\n=====================================================compress start=====================================================\n\n");
+    int malloc_size = lowlog_dump_get_size() / 3;
```
eunwoo-nam, line 232: how did you decide compress buffer malloc size?

os/kernel/log_dump/lowlog_dump_save.c
```diff
@@ -0,0 +1,318 @@
...
+#else
+	snprintf(buf, sizeof(buf), "U: %d ", ret);
+	write(fd, buf, strlen(buf));
+#endif
+	snprintf(buf, sizeof(buf), "in %.6fs", elapsed_time);
+	write(fd, buf, strlen(buf));
+	snprintf(buf, sizeof(buf), " (%s)\n", filenm);
+	write(fd, buf, strlen(buf));
+	close(fd);
+#endif
```
eunwoo-nam: This appears to be a debugging code for calculating log storage time.
If yes, please remove this code.

os/kernel/binary_manager/binary_manager_recovery.c
ewoodev:
```c
/* Do not overwrite reboot reason here */
boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
```

PR #6906, opened 2025-07-25
Support obtaining BSSID info in rtk_netmgr

jwei5:
Add support for obtaining BSSID info in rtk_netmgr of rtl8721csm and rtl8730e


hk-gwak (approved):
LGTM

PR #6913, opened 2025-07-30
drivers/compression: implement comp_open, comp_close

seokhun-eom24:
Implement comp_open() and comp_close() with reference counting and semaphore.
Implement comp_semtake() and comp_semgive() utility functions.
Update file operations to use the new open/close functions. 
Add comments for compression driver functions.
Handle compress_register errors in os_bringup.


os/drivers/compression/compress.c
```diff
@@ -71,6 +78,86 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
+ *   filep - Pointer to file structure
+ *
+ * Return Value:
+ *   OK on success, negative error code on failure
+ ****************************************************************************/
+static int comp_open(FAR struct file *filep)
+{
+	struct comp_dev_s *dev = filep->f_inode->i_private;
+
+	if (!dev) {
```
Taejun-Kwon: EINVAL is not proper value here.
You can add DEBUGASSERT(dev != NULL), comp_dev_s must be added when driver registered.

os/drivers/compression/compress.c
```diff
@@ -71,6 +78,86 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
+		return -EINVAL;
+	}
+
+	comp_semtake(&dev->exclsem);
+	if (dev->refs) {
+		comp_semgive(&dev->exclsem);
+		return -EMFILE; // We allow only 1 ref count;
```
eunwoo-nam: lzma and miniz allow multiple calling.
we don't need to block this driver

os/drivers/compression/compress.c
```diff
@@ -42,17 +44,22 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
-	comp_ioctl              /* ioctl */
+	comp_ioctl                  /* ioctl */
 #ifndef CONFIG_DISABLE_POLL
 	, 0                         /* poll */
 #endif
 };
 
+struct comp_dev_s {
+	bool refs;
+	sem_t exclsem;
```
Taejun-Kwon: I think you can add 'int running' here.
Then you can use it as below.
```
	case COMPIOC_FCOMP_INIT:
		if ((char *)arg == NULL) {
			return -EINVAL;
		}
		if (dev->running) {
			return -EBUSY;
		}
		...
		break;
	case COMPIOC_FCOMP_DEINIT:
		if (!dev->running) {
			return -EACCES; //or -EIO
		}
```
Also instead of refs, I think 'reserved' is better name.

seokhun-eom24:
> VFS layer already check validation of fileep in fs_ioctl, so you can remove all DEBUGASSERT(filep) from ioctl.

I checked. Thank you.

os/drivers/compression/compress.c
seokhun-eom24: Fixed.

os/drivers/compression/compress.c
seokhun-eom24: Good idea.
I will refer to the code when I edit it later.

os/drivers/compression/compress.c
seokhun-eom24: Thanks. 
I'll fix it and modify compress_read.c too.

os/drivers/compression/compress.c
```diff
@@ -42,17 +44,22 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
 	0,                          /* seek */
-	comp_ioctl              /* ioctl */
+	comp_ioctl                  /* ioctl */
 #ifndef CONFIG_DISABLE_POLL
 	, 0                         /* poll */
 #endif
 };
 
+struct comp_dev_s {
+	sem_t sem;
```
jeongarmy: Please include header for semaphore

seokhun-eom24: Thank you for check.

os/drivers/compression/compress.c
```diff
@@ -71,6 +79,84 @@ static ssize_t comp_write(FAR struct file *filep, FAR const char *buffer, size_t
...
+ *   OK on success, negative error code on failure
+ ****************************************************************************/
+static int comp_open(FAR struct file *filep)
+{
+	struct comp_dev_s *dev = filep->f_inode->i_private;
+
+	DEBUGASSERT(dev != NULL);
+
+	comp_semtake(&dev->sem);
+	if (dev->crefs == (uint8_t)255) {
```
jeongarmy: Please change this to UINT8_MAX or use defined value like COMP_REF_MAX

seokhun-eom24: Fixed using COMP_CREF_MAX.

os/drivers/compression/compress.c
```diff
@@ -229,14 +314,33 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
  *
+ * Return Value:
+ *   OK on success, negative error code on failure
  ****************************************************************************/
-void compress_register(void)
+int compress_register(void)
 {
-	(void)register_driver(COMP_DRVPATH, &compress_fops, 0666, NULL);
+	int ret;
+	struct comp_dev_s *dev = (struct comp_dev_s *)kmm_malloc(sizeof(struct comp_dev_s));
```
jeongarmy: crefs initialization is missing.
Please change kmm_malloc to kmm_zalloc or initialize crefs to 0

seokhun-eom24: I'll fix this. Thank you.

os/drivers/compression/compress.c
```diff
@@ -239,5 +269,6 @@ static int comp_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
  ****************************************************************************/
 void compress_register(void)
 {
+	initialize_comp_ctx_list();
```
jeongarmy: I think we need to set flag to check initialization. If compress_register is called multiple times during running by mistake, an assert or error seems to be occured because of dq_init.

seokhun-eom24: Fixed to avoid double initialization.

seokhun-eom24:
> @seokhun-eom24 Could you leave how to verify the change in commit description?

I'll add test case for compression drivers concurrently with new commit.

seokhun-eom24:
Hello, @kishore-sn 
Can you review this PR?

PR #6914, opened 2025-07-30
build/configs/rtl8730e: Fix incorrect NM override

pcs1265:
The `NM` variable was defined twice, with the second definition overriding the first one.
However, the first definition using `$(CROSSDEV)nm` is correct.

When `CONFIG_ARMV7A_TOOLCHAIN_GNU_EABI` is enabled, `NM` should resolve to `arm-none-eabi-nm`.
Since `ARCROSSDEV` is not defined for ARMv7-A, the second definition incorrectly sets `NM` to just `nm`.

This removes the redundant override to ensure the correct `NM` variable is set.


seokhun-eom24 (approved):
LGTM

PR #6915, opened 2025-08-01
os/board/rtl8721csm: modify ftl write flow

yeetee179:
- previously the secure will call ftl_non_secure_save_to_storage to write encrypted data to flash, this causes fault
- to fix the fault, now non-secure will call ftl_write_to_storage to write encrypted data to flash


hs36-kim (approved):
LGTM

PR #6917, opened 2025-08-04
fs/littlefs : Apply thread-safe logic for littlefs

Taejun-Kwon:
Enable LFS_LOCK / LFS_UNLOCK


os/fs/littlefs/littlefs/lfs.c
```diff
@@ -6570,7 +6570,7 @@ int lfs_check_format(lfs_t *lfs, const struct lfs_config *cfg)
 	}
 cleanup:
 	lfs_deinit(lfs);
-    LFS_UNLOCK(lfs->cfg);
+    LFS_UNLOCK(cfg);
```
sunghan-chang, line 6573: please check indentation

Taejun-Kwon: this is littlefs code, not ours I will keep this until it become stable we pulled 2 times from littlefs github already.

PR #6919, opened 2025-08-05
os/arch/arm/src/amebasmart: add boundary check for mipi read to preve…

zhongnuo-tang:
…nt buffer overflow

1. add boundary check based on msg len to prevent buffer overflow.
2. replace busy-wait with sem_timedwait


ewoodev (approved):

jeongarmy (approved):
lgtm

PR #6921, opened 2025-08-06
apps/examples/lcd_test/example_lcd.c: Add lcd stress test

abhinav-s235:
Implemented a stress test inside the LCD test application to check the stability of the LCD during continuous operations.
-Thread1 : Continuously switches the frame colour between red and blue to simulate rapid screen updates. -Thread2 : Continuously turns the lcd on and off to simulate frequent power cycling.

Usage:
tash>> lcd_test stress_test start <mode>                                     ----Start stress test
tash>> lcd_test stress_test stop                                                     --- Stop stress test

<mode>
0 : Run only power cycle stress test
1 : Run only frame change stress test
2 : Run both tests simultaneously

### DIFF
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -75,6 +75,7 @@
 #define NOPIXELS 200
 static int xres;
 static int yres;
+static bool g_terminate;
 
 #ifdef CONFIG_EXAMPLE_LCD_FPS_TEST
 #define EXAMPLE_LCD_FPS_TEST CONFIG_EXAMPLE_LCD_FPS_TEST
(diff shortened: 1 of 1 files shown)


apps/examples/lcd_test/example_lcd.c
```diff
@@ -407,8 +417,8 @@ static void test_fps(void)
 	
 	close(fd_rtc);
 	close(fd_lcd);
-	free(lcd_data_red);
-	free(lcd_data_blue);
+	free(area_blue.data);
+	free(area_red.data);
```
eunwoo-nam: It looks good to change it to be used for common use.👍👍👍

It is better to match the pair of alloc and free.
how about make release_frame_buffer like below?
And the prepare_frame_buffer need not be limited to red and blue.

```c
static int prepare_frame_buffer(struct lcddev_area_s *area, uint_16 color)
{

	frame_buf = malloc(len);
	for (int i = 0; i < len - 1; i += 2) {
		frame_buf[i] = (color & 0xFF00) >> 8;
		frame_buf[i + 1] = color & 0x00FF;
	}

}

void release_frame_buffer(struct lcddev_area_s *area) 
{
	free(area.data);
}

prepare_frame_buffer(&area_red);
prepare_frame_buffer(&area_blue);

release_frame_buffer(&area_red);
release_frame_buffer(&area_blue);
```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -440,6 +450,93 @@ bool is_valid_power(char *power)
...
+
+static void frame_change_test()
+{
+	int fd = 0;
+	int p = 0;
+	int ret;
+	char port[20] = { '\0' };
+	struct lcddev_area_s area_red;
+	struct lcddev_area_s area_blue;
+	prepare_frame_buffer(&area_red, &area_blue);
```
eunwoo-nam: It is duplicated allocation

apps/examples/lcd_test/example_lcd.c
```diff
@@ -440,6 +450,93 @@ bool is_valid_power(char *power)
...
+	if (ret != OK) {
+		printf("ERROR: prepare_frame_buffer failed\n");
+		close(fd);
+		return;
+	}
+
+	bool is_red = true;
+	while (!g_terminate) {
+		if (is_red) {
+			ioctl(fd, LCDDEVIO_PUTAREA, (unsigned long)(uintptr_t)&area_red);
```
eunwoo-nam: Minor, it's better to print error
```suggestion
			if (ioctl(fd, LCDDEVIO_PUTAREA, (unsigned long)(uintptr_t)&area_red) != OK) {
				printf("Fail to lcd putarea, errno: %d, get_errno());
			}
```

apps/examples/lcd_test/example_lcd.c
abhinav-s235: changes made as suggested

apps/examples/lcd_test/example_lcd.c
abhinav-s235: fixed

apps/examples/lcd_test/example_lcd.c
abhinav-s235: Done

apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,6 +445,104 @@ bool is_valid_power(char *power)
...
+		return;
+	}
+	while (!g_terminate) {
+		ioctl(fd, LCDDEVIO_SETPOWER, power);
+		if (power == 100) {
+			power = 0;
+		} else {
+			power = 100;
+		}
+		usleep(10000); /*Sleep for 100 ms*/
```
seokhun-eom24: Fix to match with code and comment.

`usleep(100000); /*Sleep for 100 ms*/`
or
`usleep(10000); /*Sleep for 10 ms*/`

apps/examples/lcd_test/example_lcd.c
```diff
@@ -470,6 +576,31 @@ int lcd_test_main(int argc, char *argv[])
...
 
+	/* Stress test*/
+	if (argc >= 2 && !strncmp(argv[1], "stress_test", 12)) {
+		if (argc > 2) {
+			if (!strncmp(argv[2], "start", 6)) {
+				if (argc > 3) {
+					struct fb_videoinfo_s vinfo;
+					ioctl(fd, LCDDEVIO_GETVIDEOINFO, (unsigned long)(uintptr_t)&vinfo);
+					xres = vinfo.xres;
+					yres = vinfo.yres;
```
eunwoo-nam: Minor

the xres and yres global value are setted from several code.
Actually the value is same because we use only 1 lcd driver.
but it is multiple access value, 
So it is better to remove global value and replace to internal value. 
because the example apps can be guidence.

It looks only use global xrex, yrex
```c
static int prepare_frame_buffer(struct lcddev_area_s *area, uint16_t color, xrex, yrex)

```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -321,66 +322,73 @@ static void test_quad(void)
...
+	free(area->data);
+}
+
+static int prepare_frame_buffer(struct lcddev_area_s *area, uint16_t color)
+{
+	size_t len;
+	len = xres * yres * 2 + 1;
+	uint8_t *lcd_data = (uint8_t *)malloc(len);
+	if (lcd_data == NULL) {
+		printf("malloc failed for lcd data red : %d\n", len);
```
sunghan-chang: "red" means?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,6 +445,104 @@ bool is_valid_power(char *power)
 	return true;
 }
 
+static void power_cycle_test()
```
sunghan-chang: (void)

apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,6 +445,104 @@ bool is_valid_power(char *power)
...
+			power = 0;
+		} else {
+			power = 100;
+		}
+		usleep(10000); /*Sleep for 100 ms*/
+	}
+	close(fd);
+}
+
+static void frame_change_test()
```
sunghan-chang: same as previous

apps/examples/lcd_test/example_lcd.c
```diff
@@ -470,6 +576,31 @@ int lcd_test_main(int argc, char *argv[])
...
+	if (argc >= 2 && !strncmp(argv[1], "stress_test", 12)) {
+		if (argc > 2) {
+			if (!strncmp(argv[2], "start", 6)) {
+				if (argc > 3) {
+					struct fb_videoinfo_s vinfo;
+					ioctl(fd, LCDDEVIO_GETVIDEOINFO, (unsigned long)(uintptr_t)&vinfo);
+					xres = vinfo.xres;
+					yres = vinfo.yres;
+					g_terminate = false;
+					stress_test(atoi(argv[3]));
```
sunghan-chang: 1. Don't you need to verify `argv[3]` first?
2. I can't find any usages message for new command. Let's add it.

abhinav-s235: - is_valid_stress_test_arg() has been added to verify argv[3]
-  show_usage() function has been added to show usages message

apps/examples/lcd_test/example_lcd.c
abhinav-s235: fixed

apps/examples/lcd_test/example_lcd.c
abhinav-s235: fixed

apps/examples/lcd_test/example_lcd.c
abhinav-s235: changes made as suggested

apps/examples/lcd_test/example_lcd.c
abhinav-s235: fixed

apps/examples/lcd_test/example_lcd.c
abhinav-s235: Thanks for pointing that out, I have removed the extra word.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -321,66 +322,73 @@ static void test_quad(void)
 	free(lcd_data);
 }
 
+static void release_frame_buffer(struct lcddev_area_s *area)
+{
+	free(area->data);
```
jeongarmy, line 362: Isn't it needed to check NULL before calling free?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,37 +445,147 @@ bool is_valid_power(char *power)
...
+	printf("usage: lcd_test <command args(optional)>\n");
+	printf("    basic             : Execute basic lcd_test\n");
+	printf("    power <value>     : Sets the brightness to given value\n");
+	printf("    stress_test <start> <mode> | <stop> : Start or stop stress test, <mode>: 0 = power cycle test only, 1 = frame change test only, 2 = both test simultaneously\n");
+}
+
+static void stress_test(int num)
+{
+	switch (num) {
+		case 0: {
```
sunghan-chang: last comment,
switch and case should be the same indentation
```
	switch (num) {
	case 0: 
		....
		break;
```

abhinav-s235: fixed

apps/examples/lcd_test/example_lcd.c
abhinav-s235: NULL check has been added

sunghan-chang (approved):

apps/examples/lcd_test/example_lcd.c
```diff
@@ -321,66 +322,76 @@ static void test_quad(void)
 	free(lcd_data);
 }
 
+static void release_frame_buffer(struct lcddev_area_s *area)
+{
+	if (area->data != NULL) {
```
sunghan-chang: I think she does not mean checking `area->data` with null.
I think, it's `area`. If it's null, then `free(area->data);` will be null dereferencing.

But if there is no chance to be null for area, no need to check it.

abhinav-s235: Yes this scenario(null dereferencing) will not arise

ewoodev (approved):

allen-kim-sec (approved):
it just test application for LCD... lgtm

PR #6922, opened 2025-08-08
apps/examples/lcd_test: Add lcd get info test

anjana348:
Add test for LCD_GETINFO ioctl for LCD width mm, height mm, DPI

TASH>>lcd_test lcdinfo
TASH>>=== LCD demo ===
LCD HEIGHT MM: 58
LCD WIDTH MM: 43
LCD SIZE INCH: 2.842563
LCD DPI: 140.718063

### DIFF
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -61,6 +61,7 @@
 #include <errno.h>
 
 #define LCD_DEV_PATH "/dev/lcd%d"
+#define LCD_DEV_PORT 0
 
 #define RED   0xF800
 #define WHITE 0xFFFF
@@ -86,7 +87,6 @@ static void putarea(int x1, int x2, int y1, int y2, int color)
 	struct lcddev_area_s area;
 	char port[20] = { '\0' };
 	int fd = 0;
-	int p = 0;
 	size_t len;
 	len = xres * yres * 2 + 1;
 	uint8_t *lcd_data = (uint8_t *)malloc(len);
@@ -94,7 +94,7 @@ static void putarea(int x1, int x2, int y1, int y2, int color)
 		printf("malloc failed for lcd data : %d\n", len);
 		return;
 	}
-	sprintf(port, LCD_DEV_PATH, p);
+	snprintf(port, sizeof(port) / sizeof(port[0]), LCD_DEV_PATH, LCD_DEV_PORT);
 	fd = open(port, O_RDWR | O_SYNC, 0666);
 	if (fd < 0) {
 		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
@@ -113,7 +113,9 @@ static void putarea(int x1, int x2, int y1, int y2, int color)
 	}
 
 	area.data = lcd_data;
-	ioctl(fd, LCDDEVIO_PUTAREA, (unsigned long)(uintptr_t)&area);
+	if (ioctl(fd, LCDDEVIO_PUTAREA, (unsigned long)(uintptr_t)&area) < 0) {
+		 printf("Fail to call LCD PUTAREA(errno %d)", get_errno());
+	}
 	close(fd);
 	free(lcd_data);
 }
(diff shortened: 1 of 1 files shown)


apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,6 +437,36 @@ bool is_valid_power(char *power)
 	return true;
 }
 
+static int lcd_get_info(void)
+{
+	int ret = OK;
+	int fd = 0;
+	int p = 0;
+	char port[20] = {'\0'};
+	snprintf(port, sizeof(port) / sizeof(port[0]), LCD_DEV_PATH, p);
```
seokhun-eom24: Why was `snprintf` used instead of `sprintf` like other functions?

anjana348: Previously Ms Jeong suggested to use snprintf instead of sprintf, you will see new code using snprintf in example_lcd.c also

apps/examples/lcd_test/example_lcd.c
```diff
@@ -469,7 +499,10 @@ int lcd_test_main(int argc, char *argv[])
 		close(fd);
 		return OK;
 	}
-
+	/* LCD get Info */
+	if (argc == 2 && !strncmp(argv[1], "lcdinfo", 5)) {
```
jeongarmy: 5 looks wrong.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -437,6 +437,36 @@ bool is_valid_power(char *power)
...
+static int lcd_get_info(void)
+{
+	int ret = OK;
+	int fd = 0;
+	int p = 0;
+	char port[20] = {'\0'};
+	snprintf(port, sizeof(port) / sizeof(port[0]), LCD_DEV_PATH, p);
+	fd = open(port, O_RDWR | O_SYNC, 0666);
+	if (fd < 0) {
+		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, fd);
```
jeongarmy: Please print get_errno instead of fd.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -469,7 +499,10 @@ int lcd_test_main(int argc, char *argv[])
 		close(fd);
 		return OK;
 	}
-
+	/* LCD get Info */
+	if (argc == 2 && !strncmp(argv[1], "lcdinfo", 7)) {
```
sunghan-chang: Is there any help message for usages?

anjana348: Added

apps/examples/lcd_test/example_lcd.c
```diff
@@ -484,6 +503,56 @@ int lcd_test_main(int argc, char *argv[])
...
+	return OK;
+}
+
+static int power_test(int power)
+{
+	int count = 0;
+	int fd = 0;
+	int p = 0;
+	char port[20] = { '\0' };
+	sprintf(port, LCD_DEV_PATH, p);
```
jeongarmy: Please use snprintf instead of sprintf.
Line 485 as well.

jeongarmy: And how about defining port number?
Currently all tests use same port path with `p` which is set as 0.
It looks I think you can change them like below.
```
#define LCD_DEV_PORT 0

In each functions,
snprintf(port, sizeof(port) / sizeof(port[0]), LCD_DEV_PATH, LCD_DEV_PORT);
```

apps/examples/lcd_test/example_lcd.c
```diff
@@ -484,6 +503,56 @@ int lcd_test_main(int argc, char *argv[])
...
+{
+
+	if (argc <= 1 || !strncmp(argv[1], "-h", 2) || !strncmp(argv[1], "--help", 6)) {
+		show_usage();
+		return OK;
+	}
+
+	/* LCD Power test */
+	if (argc >= 2 && !strncmp(argv[1], "power", 5)) {
+		if (argc > 2 && is_valid_power(argv[2])) {
```
jeongarmy: `argc > 2` is duplicated.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -484,6 +494,55 @@ int lcd_test_main(int argc, char *argv[])
...
+	int count = 0;
+	int fd = 0;
+	char port[20] = { '\0' };
+	snprintf(port, sizeof(port) / sizeof(port[0]), LCD_DEV_PATH, LCD_DEV_PORT);
+	fd = open(port, O_RDWR | O_SYNC, 0666);
+	if (fd < 0) {
+		printf("ERROR: Failed to open lcd port : %s error:%d\n", port, get_errno());
+		return ERROR;
+	}
+	ioctl(fd, LCDDEVIO_SETPOWER, power);
```
jeongarmy: Isn't it needed to check result?

apps/examples/lcd_test/example_lcd.c
```diff
@@ -484,6 +494,55 @@ int lcd_test_main(int argc, char *argv[])
...
+
+	/* LCD Power test */
+	if (argc >= 2 && !strncmp(argv[1], "power", 5)) {
+		if (is_valid_power(argv[2])) {
+			return power_test(atoi(argv[2]));
+		} else {
+			printf("ERROR: Value of power should be int in range [0, 100]\n");
+			printf("Usage: lcd_test power <value>\n");
+			printf("0 --> LCD Power OFF\n");
+			printf("100 --> LCD Power ON\n");
```
jeongarmy, line 586: Please add return here

apps/examples/lcd_test/example_lcd.c
```diff
@@ -484,6 +494,55 @@ int lcd_test_main(int argc, char *argv[])
...
+		}
+	}
+	if (argc == 2) {
+		if (!strncmp(argv[1], "lcdinfo", 7)) {
+			return lcd_get_info();
+		} else if (!strncmp(argv[1], "basic", 5)) {
+			return lcd_basic_test();
+		}
```
jeongarmy, line 595: please add `show_usage`

apps/examples/lcd_test/example_lcd.c
seokhun-eom24: I think we need to verify `argc > 2` as previously done.
Otherwise, accessing `argv[2]` will result in an error when `argc == 2`.

apps/examples/lcd_test/example_lcd.c
```diff
@@ -476,14 +557,48 @@ int lcd_test_main(int argc, char *argv[])
...
+#endif
+{
+
+	if (argc <= 1 || !strncmp(argv[1], "-h", 2) || !strncmp(argv[1], "--help", 6)) {
+		show_usage();
+		return OK;
+	}
+
+	/* LCD Power test */
+	if (argc >= 2 && !strncmp(argv[1], "power", 5)) {
```
seokhun-eom24: You can use string length+ 1 for exact comparison.
This can ensure exact string matching rather than partial matching (e.g. powerful).
```suggestion
	if (argc >= 2 && !strncmp(argv[1], "power", 6)) {
```
and apply this to other `strncmp` too.

jeongarmy (approved):
lgtm

PR #6923, opened 2025-08-11
os/drivers/lcd: Add Kconfig option for lcd initialization

seokhun-eom24:
Add Kconfig options for LCD init code.
If one config is selected use that init code, otherwise read id from lcd and choose right lcd init code.
Add new Holitech st7785 LCD initialization code with detailed comments for each command.
Seperate Holitech and Techshine LCD initialization code.
Refactor check_lcd_vendor_send_init_cmd to use proper conditional compilation based on configuration.

Set loadable_ext_ddr defconfig to use lcd pwm inversion.
Set loadable_ext_ddr_st7785 defconfig to use Holitech init sequence.


os/drivers/lcd/st7785.c
```diff
@@ -48,6 +49,19 @@ int check_lcd_vendor_send_init_cmd(struct mipi_lcd_dev_s *priv)
...
+		return send_init_cmd(priv, lcd_init_cmd_g);
+	#elif defined(CONFIG_LCD_STATIC_INIT_AVD)
+		return send_init_cmd(priv, lcd_init_cmd_g_avd);
+	#elif defined(CONFIG_LCD_STATIC_INIT_HOLITECH)
+		return send_init_cmd(priv, lcd_init_cmd_g_holitech);
+	#else
+		lcddbg("No static initialization type selected\n");
+		return ERROR;
+	#endif
+#endif
```
eunwoo-nam:
```suggestion
#elif defined(CONFIG_LCD_STATIC_INIT)
	/* Use static init code */
#ifdef CONFIG_LCD_STATIC_INIT_DEFAULT
	return send_init_cmd(priv, lcd_init_cmd_g);
#elif defined(CONFIG_LCD_STATIC_INIT_AVD)
	return send_init_cmd(priv, lcd_init_cmd_g_avd);
#elif defined(CONFIG_LCD_STATIC_INIT_HOLITECH)
	return send_init_cmd(priv, lcd_init_cmd_g_holitech);
#else
#error "No static initialization type selected"
#endif
#endif
```

Please check the indentation and the configure is static init so build time error is better.

seokhun-eom24: Thank you.
I corrected the indentation and fixed to throw error in build time.

os/drivers/lcd/st7785.c
```diff
@@ -48,6 +49,18 @@ int check_lcd_vendor_send_init_cmd(struct mipi_lcd_dev_s *priv)
...
+#ifdef CONFIG_LCD_STATIC_INIT_DEFAULT
+	return send_init_cmd(priv, lcd_init_cmd_g);
+#elif defined(CONFIG_LCD_STATIC_INIT_AVD)
+	return send_init_cmd(priv, lcd_init_cmd_g_avd);
+#elif defined(CONFIG_LCD_STATIC_INIT_HOLITECH)
+	return send_init_cmd(priv, lcd_init_cmd_g_holitech);
+#else
+#error "No static initialization type selected"
+#endif
+#endif
```
jeongarmy: I think it is ok when user configures this by using menuconfig but it can be missed by changing defconfig directly.
How about making build error when init configuration is not set?
I mean
```
#else
#error "No static initialization type selected"
#endif
#else
#error "No configuration for LCD initialization"
#endif
```

seokhun-eom24: I added else case to throw build error when LCD initialization is not set.

os/drivers/lcd/Kconfig
```diff
@@ -145,22 +164,59 @@ config LCD_ST7785
...
+choice
+	prompt "Static Initialization code"
+	default LCD_STATIC_INIT_DEFAULT
+	---help---
+		Select predefined LCD initialization sequence
+
+config LCD_STATIC_INIT_DEFAULT
+	bool "Default initialization"
+	---help---
+		Use default initialization sequence
```
jeongarmy: What code is the default init code for which LCD? Is it necessary?

seokhun-eom24: The `LCD_STATIC_INIT_DEFAULT` config exists for other LCDs.
Config like `CONFIG_LCD_STATIC_INIT_AVD_002701` are options for only st7785, so the default option is left to allow other LCDs to select it when `CONFIG_LCD_STATIC_INIT` is set.
In the current st7785, this config performs the same function as `CONFIG_LCD_STATIC_INIT_HLT_A196`.
However, it was created as an option for future scalability.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -235,10 +235,12 @@ static void rtl8730e_control_backlight(uint8_t level)
 		/* TO-DO: Move LCD IC Power ON flow */
 		InterruptEn(lcdc_irq_info.num, lcdc_irq_info.priority);
 	}
-#if defined(CONFIG_LCD_ST7701SN)
+#if defined(CONFIG_LCD_PWM_INVERSION)
```
jeongarmy, line 243: It is obvious that the backlight turns off when the PWM value is 0.
Even if user don't set a pwm mode additionally, it seems that it operates in the normal mode by default.
So I think you can remove a configuration for the basic operation and handle it when PWM_INVERSION is enabled only for better management.

#if defined(CONFIG_LCD_PWM_INVERSION)
...
#else
...
#endif

seokhun-eom24: You're right.
`CONFIG_LCD_PWM_NORMAL` is a default case, so it's more natural not to set it.
Only `CONFIG_LCD_PWM_INVERSION` has been left as a config.

os/drivers/lcd/Kconfig
eunwoo-nam:
```
	prompt "Static Initialization code"
	default LCD_STATIC_INIT_AVD_0027NNB01
```
you can select default setting using above code

os/include/tinyara/lcd/st7785.h
```diff
@@ -118,31 +120,56 @@ static const lcm_setting_table_t lcd_init_cmd_g[] = {
...
-        {0x030000, lcd_init_cmd_g_avd},            // AVD B
-        {0x484c54, lcd_init_cmd_g_holitech},  // Holitech
-        {0x020000, lcd_init_cmd_g_holitech},  // Holitech
-#endif /* CONFIG_LCD_ST7785_TYPEB */
+        {0x8A8A85, lcd_init_cmd_g_avd_002701},  // AVD B
+        {0x010000, lcd_init_cmd_g_avd_002701},  // AVD B
+        {0x030000, lcd_init_cmd_g_avd_002701},  // AVD B
+        {0x484c54, lcd_init_cmd_g_hlt_a196},    // HLT
+        {0x020000, lcd_init_cmd_g_hlt_a196},    // HLT
+        {0x030000, lcd_init_cmd_g_tec},         // TEC
```
eunwoo-nam: {0x030000, lcd_init_cmd_g_avd_002701}, // AVD B
{0x030000, lcd_init_cmd_g_tec}, // TEC
the two id are conflicted

os/drivers/lcd/Kconfig
```diff
@@ -116,6 +116,13 @@ config LCD_FLUSH_THREAD
 	---help---
 		Separate thread for LCD which will start processing next frame after previous is sent to lcd driver.
 
+config LCD_PWM_INVERSION
```
sunghan-chang, line 119: This config is defined in drivers/lcd, not in board/rtl8730e.
But you use this in board. Please check that we can move the code into drivers.

os/drivers/lcd/Kconfig
```diff
@@ -116,6 +116,13 @@ config LCD_FLUSH_THREAD
...
+	bool "LCD PWM inversion mode"
```
sunghan-chang: Let's add `backlight` at description.

os/include/tinyara/lcd/st7785.h
```diff
@@ -76,30 +75,33 @@
...
+        {REGFLAG_DELAY, 120, {}},               /* Delayms (120) */
+        {0x29, 0, {0x00}},                      /* Display On */
+        {REGFLAG_DELAY, 10, {}},                /* Delayms (10) */
         {REGFLAG_END_OF_TABLE, 0, {}},
 };
 #endif
 
-#if defined(CONFIG_LCD_ST7785_TYPEA)
-static const lcm_setting_table_t lcd_init_cmd_g[] = {
+#if defined(CONFIG_LCD_STATIC_INIT_AVD_0027NNB01)
```
sunghan-chang: Don't you need this LCD type in dynamic init?

os/include/tinyara/lcd/st7785.h
seokhun-eom24: Thank you. Fixed.

os/drivers/lcd/Kconfig
seokhun-eom24: Added.

os/include/tinyara/lcd/st7785.h
seokhun-eom24: Yes, we need `CONFIG_LCD_STATIC_INIT_AVD_0027NNB01` for id `0x8A8A85`.
So I fixed it.
Thank you.

os/drivers/lcd/Kconfig
seokhun-eom24: I removed all static, dynamic, and default configs.  
For the init code config, if multiple options are selected, the ID will be read and chosen accordingly.
If only one config is selected, that specific configuration will be applied.

os/drivers/lcd/Kconfig
seokhun-eom24: Option 1 is relocating the `PWM_INVERSION` config to the board's Kconfig, as it is unconventional to configure the inversion of the LCD's PWM within the board Kconfig.

Option 2 is passing the inversion status as an argument in `rtl8730e_control_backlight` or storing it in `g_rtl8730e_config_dev_s`. However both approaches are challenging to implement. 

Because during the boot process `rtl8730e_lcdc_initialize` calls `rtl8730e_control_backlight` to initialize the backlight to the off state before the driver is registered, making it impossible to determine the PWM inversion status without referencing `PWM_INVERSION` config.

eunwoo-nam: Could you please seperate commit about this?

os/drivers/lcd/st7785.c
```diff
@@ -29,6 +29,22 @@ int check_lcd_vendor_send_init_cmd(struct mipi_lcd_dev_s *priv)
...
+	return send_init_cmd(priv, lcd_init_cmd_g_avd_nb01);
+#elif CONFIG_LCD_INIT_HLT_A196
+	return send_init_cmd(priv, lcd_init_cmd_g_hlt_a196);
+#elif CONFIG_LCD_INIT_TEC_181A
+	return send_init_cmd(priv, lcd_init_cmd_g_tec_181a);
+#endif
+#elif NUM_LCD_VENDERS == 0
+#error "No LCD init code defined"
+#endif
+
```
eunwoo-nam: if NUL_LCD_VENDERS == 1, we don't need to include checking id code. 
Let's exclude it like below

```c

#if NUM_LCD_VENDERS == 1
~~~
#elif NUM_LCD_VENDERS > 1 
~~~
#else
#error "No LCD init code defined"
#endif


```

os/drivers/lcd/Kconfig
seokhun-eom24: > Could you please seperate commit about this?

I seperated to pwm relative and id relative commit.

os/drivers/lcd/st7785.c
seokhun-eom24: Fixed.

eunwoo-nam (approved):

jeongarmy (approved):
lgtm

allen-kim-sec (approved):
looks good to me

PR #6926, opened 2025-08-13
amebasmart: Fix code flow issue in KM4

jwei5:
- Fix code flow issue in wifi driver that may cause KM4 to be stuck
- Add logs in IPC API to aid debugging of IPC in the future


jwei5:
This issue was found to be caused by a variable in the driver that was being set to FALSE too early. Because this variable is checked when handling interrupts, it could cause interrupts to not be handled in KM4 if it was set incorrectly due to corner cases, causing KM4 to be stuck. It has been resolved by correcting the code flow in our driver.

hk-gwak:
> This issue was found to be caused by a variable in the driver that was being set to FALSE too early. Because this variable is checked when handling interrupts, it could cause interrupts to not be handled in KM4 if it was set incorrectly due to corner cases, causing KM4 to be stuck. It has been resolved by correcting the code flow in our driver.

Thank you for the details. Is this fix for the stuck issue when receiving the deauth packet by wifi jamming continuously?

jwei5:
> > This issue was found to be caused by a variable in the driver that was being set to FALSE too early. Because this variable is checked when handling interrupts, it could cause interrupts to not be handled in KM4 if it was set incorrectly due to corner cases, causing KM4 to be stuck. It has been resolved by correcting the code flow in our driver.
> 
> Thank you for the details. Is this fix for the stuck issue when receiving the deauth packet by wifi jamming continuously?

Yes, this PR is to fix that issue

PR #6927, opened 2025-08-18
os/kernel/: Add null check to some kernel api

Sambhavi-1:
For some syscalls, there are security issues because null check is not performed in corresponding kernel apis. This pr adds null check

### DIFF
--- a/os/kernel/clock/clock_getres.c
+++ b/os/kernel/clock/clock_getres.c
@@ -107,6 +107,10 @@ int clock_getres(clockid_t clock_id, struct timespec *res)
 {
 	int ret = OK;
 
+	if (!res) {
+		return -EINVAL;
+	}
+
 	svdbg("clock_id=%d\n", clock_id);
 
 	/* Only CLOCK_REALTIME is supported */
--- a/os/kernel/clock/clock_gettime.c
+++ b/os/kernel/clock/clock_gettime.c
@@ -60,7 +60,6 @@
 
 #include <stdint.h>
 #include <time.h>
-#include <assert.h>
 #include <errno.h>
 #include <debug.h>
 
@@ -114,8 +113,11 @@ int clock_gettime(clockid_t clock_id, struct timespec *tp)
 	uint32_t carry;
 	int ret = OK;
 
+	if (!tp) {
+		return -EINVAL;
+	}
+
 	svdbg("clock_id=%d\n", clock_id);
-	DEBUGASSERT(tp != NULL);
 
 #ifdef CONFIG_CLOCK_MONOTONIC
 	/* CLOCK_MONOTONIC is an optional under POSIX: "If the Monotonic Clock
--- a/os/kernel/environ/env_getenvironptr.c
+++ b/os/kernel/environ/env_getenvironptr.c
@@ -92,6 +92,11 @@ FAR char *get_environ_ptr(size_t *envsize)
 	FAR struct task_group_s *group;
 	char *ret;
 
+	if (!envsize) {
+		set_errno(EINVAL);
+		return NULL;
+	}
+
 	group = ptcb->group;
 	if (group && group->tg_envp) {
 		*envsize = group->tg_envsize;
--- a/os/kernel/pthread/pthread_keycreate.c
+++ b/os/kernel/pthread/pthread_keycreate.c
@@ -98,7 +98,7 @@ int pthread_key_create(FAR pthread_key_t *key, pthread_destructor_t destructor)
 	DEBUGASSERT(group);
 
 	if (!key) {
-		return EINVAL;
+		return -EINVAL;
 	}
 
 	/* Find free key */
--- a/os/kernel/sched/sched_waitpid.c
+++ b/os/kernel/sched/sched_waitpid.c
@@ -208,7 +208,6 @@ pid_t waitpid(pid_t pid, int *stat_loc, int options)
 
 	DEBUGASSERT(stat_loc);
 
-
 	/* None of the options are supported */
 
 	if (options != 0) {
@@ -316,6 +315,10 @@ pid_t waitpid(pid_t pid, int *stat_loc, int options)
 	int ret;
 
 	DEBUGASSERT(stat_loc);
+	if (!stat_loc) {
+		set_errno(EINVAL);
+		return ERROR;
+	}
 
 	/* None of the options are supported */
 


os/kernel/pthread/pthread_join.c
```diff
@@ -123,6 +123,10 @@ int pthread_join(pthread_t thread, FAR pthread_addr_t *pexit_value)
 {
 	int ret;
 
+	DEBUGASSERT(pexit_value);
+	if (!pexit_value) {
+		return EINVAL;
+	}
```
kishore-sn: Since this pointer is not being dereferenced, we can remove this check.

os/kernel/sched/sched_waitpid.c
```diff
@@ -207,6 +207,9 @@ pid_t waitpid(pid_t pid, int *stat_loc, int options)
 	int ret;
 
 	DEBUGASSERT(stat_loc);
+	if (!stat_loc) {
+		return EINVAL;
+	}
```
kishore-sn: This is not required as we are doing a null check at line 248 before using it.

os/kernel/sched/sched_waitpid.c
```diff
@@ -316,6 +319,9 @@ pid_t waitpid(pid_t pid, int *stat_loc, int options)
 	int ret;
 
 	DEBUGASSERT(stat_loc);
+	if (!stat_loc) {
+		return EINVAL;
```
kishore-sn: We need to set_errno and return ERROR.

os/kernel/clock/clock_gettime.c
```diff
@@ -116,6 +116,9 @@ int clock_gettime(clockid_t clock_id, struct timespec *tp)
 
 	svdbg("clock_id=%d\n", clock_id);
 	DEBUGASSERT(tp != NULL);
+	if (!tp) {
+		return EINVAL;
```
kishore-sn: return -EINVAL

os/kernel/clock/clock_getres.c
```diff
@@ -107,6 +107,10 @@ int clock_getres(clockid_t clock_id, struct timespec *res)
 {
 	int ret = OK;
 
+	if (!res) {
+		return EINVAL;
```
kishore-sn: return -EINVAL
or 
set_errno(EINVAL)
return ERROR

os/kernel/clock/clock_getres.c
```diff
@@ -107,6 +107,10 @@ int clock_getres(clockid_t clock_id, struct timespec *res)
 {
 	int ret = OK;
```
kishore-sn, line 109: Please add DEBUGASSERT also here.

os/kernel/sched/sched_waitpid.c
Sambhavi-1: Done

os/kernel/clock/clock_gettime.c
Sambhavi-1: Done

os/kernel/clock/clock_getres.c
Sambhavi-1: Done

os/kernel/environ/env_getenvironptr.c
```diff
@@ -92,6 +92,11 @@ FAR char *get_environ_ptr(size_t *envsize)
 	FAR struct task_group_s *group;
 	char *ret;
 
+	DEBUGASSERT(envsize);
+	if (!envsize) {
+		return -EINVAL;
+	}
```
seokhun-eom24: How about this for a pointer type return?
```suggestion
	if (!envsize) {
	    set_errno(EINVAL);
	    return NULL;
	}
```

Sambhavi-1: Resolved

kishore-sn (approved):

samsung-singh (approved):
added null check to few kernel apis this is to required for debug purposes, this change is good to merge.

kripal-samsung (approved):

os/kernel/clock/clock_getres.c
```diff
@@ -107,6 +108,11 @@ int clock_getres(clockid_t clock_id, struct timespec *res)
...
 
+	DEBUGASSERT(res);
```
sunghan-chang: I don't think we should make the assert in this case. Just error return is enough. Let's remove this line.

os/kernel/clock/clock_gettime.c
```diff
@@ -116,6 +116,9 @@ int clock_gettime(clockid_t clock_id, struct timespec *tp)
 
 	svdbg("clock_id=%d\n", clock_id);
 	DEBUGASSERT(tp != NULL);
```
sunghan-chang: same here.

os/kernel/clock/clock_gettime.c
```diff
@@ -116,6 +116,9 @@ int clock_gettime(clockid_t clock_id, struct timespec *tp)
...
+	if (!tp) {
```
sunghan-chang, line 116: Much minor comment.
You check the variable before svdbg for clock_id at previous function.
Let's do that with same sequence.

os/kernel/pthread/pthread_keycreate.c
```diff
@@ -97,8 +97,9 @@ int pthread_key_create(FAR pthread_key_t *key, pthread_destructor_t destructor)
 
 	DEBUGASSERT(group);
 
+	DEBUGASSERT(key);
```
sunghan-chang: If the group is null, it should not be happened. When it's null, then it means something wrong. That's why we should call assert here.
But for key, I think just error return is enough.

os/kernel/clock/clock_getres.c
Sambhavi-1: okay

os/kernel/pthread/pthread_keycreate.c
Sambhavi-1: okay...done

os/kernel/clock/clock_getres.c
```diff
@@ -58,6 +58,7 @@
 
 #include <stdint.h>
 #include <time.h>
+#include <assert.h>
```
sunghan-chang: You removed added DEBUGASSERT so that this inclusion is unnecessary.

os/kernel/environ/env_getenvironptr.c
```diff
@@ -92,6 +92,12 @@ FAR char *get_environ_ptr(size_t *envsize)
 	FAR struct task_group_s *group;
 	char *ret;
 
+	DEBUGASSERT(envsize);
```
sunghan-chang: Let's remove this for the same reason I left.

sunghan-chang:
@kishore-sn After modifying my two comments, let's merge.

kishore-sn (approved):

PR #6928, opened 2025-08-18
os/include/tinyara/log_dump: Reduce stack size of log_dump

seokhun-eom24:
log_dump uses global variables or static allocations, so it does not appear to consume additional stack space. 
Its current peak stack usage is approximately 7K against a 16K allocation. 
Reducing this allocation to 10K would free roughly 6K of stack memory while retaining a 3K buffer.


pcs1265 (approved):
LGTM

sunghan-chang (approved):

PR #6929, opened 2025-08-18
fs/littlefs: apply block/MTD driver in littlefs

aashish-l:
Add support to littlefs VFS layer to choose the driver based on INODE. 
For MTD driver select the MTD defined functions otherwise use block device.

Check build: Build is fine.

### DIFF
--- a/os/fs/littlefs/lfs_vfs.c
+++ b/os/fs/littlefs/lfs_vfs.c
@@ -446,7 +446,17 @@ static int littlefs_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	fs = inode->i_private;
 	drv = fs->drv;
 	DEBUGASSERT(drv && drv->i_private);
-	return MTD_IOCTL((FAR struct mtd_dev_s *)drv->i_private, cmd, arg);
+
+	if (INODE_IS_MTD(drv)) {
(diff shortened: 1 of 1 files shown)


os/fs/littlefs/lfs_vfs.c
```diff
@@ -446,7 +446,17 @@ static int littlefs_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
...
+	if (INODE_IS_MTD(drv)) {
+		ret = MTD_IOCTL(drv->u.i_mtd, cmd, arg);
+	} else {
+		if (drv->u.i_bops->ioctl != NULL) {
+			ret = drv->u.i_bops->ioctl(drv, cmd, arg);
+		} else {
+			ret = -ENOTTY;
+		}
+	}
+return ret;
```
seokhun-eom24, line 459: Need to add indent.

os/fs/littlefs/lfs_vfs.c
```diff
@@ -817,12 +833,20 @@ static int littlefs_sync_block(FAR const struct lfs_config *c)
 	int ret;
```
seokhun-eom24, line 833: I think set some default value is safer.
This function will return trash value in current code.

PR #6932, opened 2025-08-20
apps/system/utils/heapinfo: Default to printing heapinfo for all binaries

pcs1265:
Modify heapinfo behavior to print heap information for all binaries when executed without specifying a binary.
  
  If the user uses the -k, -b, or -d options, it is assumed that binary is specified.

### DIFF
--- a/apps/system/utils/utils_heapinfo.c
+++ b/apps/system/utils/utils_heapinfo.c
@@ -196,6 +196,14 @@ static void heapinfo_show_taskinfo(void)
 #endif
 }
 
+static void heapinfo_show_binname(bool init_flag, char *heap_name) {
+	if (init_flag != true) {
+		printf("\n****************************************************************\n");
+		printf("     %s HEAP INFORMATION\n", heap_name);
+		printf("****************************************************************\n");
+	}
+}
+
 int utils_heapinfo(int argc, char **args)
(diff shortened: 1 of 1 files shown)


apps/system/utils/utils_heapinfo.c
```diff
@@ -196,13 +196,22 @@ static void heapinfo_show_taskinfo(void)
 #endif
 }
 
+static void heapinfo_show_binname(bool init_flag, char *heap_name) {
+	if (init_flag != true) {
+		printf("\n****************************************************************\n");
+		printf("     %s HEAP INFORMATION\n", heap_name);
+		printf("****************************************************************\n");
+	}
+}
```
seokhun-eom24, line 205: Please ensure that curly braces are placed on the following line.
```suggestion
static void heapinfo_show_binname(bool init_flag, char *heap_name)
{
	if (init_flag != true) {
		printf("\n****************************************************************\n");
		printf("     %s HEAP INFORMATION\n", heap_name);
		printf("****************************************************************\n");
	}
}
```

Sambhavi-1 (commented):
Changes to print heapinfo for all binaries as default looks ok

jeongarmy (approved):
lgtm

PR #6938, opened 2025-08-25
os/board/rtl8730e: modify malloc to kmm_malloc

yeetee179:
- fixed assert in loadable config caused by malloc


giwon-nam (approved):
Indicate / Notify Normal Operation Check has been completed.

PR #6942, opened 2025-08-25
os/board/rtl8730e: BLE remove unused variable and correct memory allocation

ziyik:
…cation

-Remove unused structure variable
-Correct memory allocation to prevent memory fault


sunghan-chang:
@giwon-nam @hs36-kim Could you check this?

giwon-nam (approved):
- Verified the removal of all unused `tizenrt_ble_service_tbl` entries.
- Verified the clear modification of the creation/removal timing after changing the `ble_tizenrt_attrs` array, declared within the function, to static.

PR #6945, opened 2025-08-28
apps/system/utils/stackmonitor : Add config to toggle dead thread stack info

pcs1265:
- Add CONFIG_STACKMONITOR_DISABLE_DEAD_THREAD (bool, default n).
- When enabled, dead thread stack infos are not stored, reducing RAM consumption.
- When disabled, the stack monitor retains stack info of threads that have exited.

### DIFF
--- a/apps/system/utils/utils_stackmonitor.c
+++ b/apps/system/utils/utils_stackmonitor.c
@@ -259,6 +259,7 @@ static void *stackmonitor_daemon(void *arg)
 		printf("|------------");
 #endif
 		printf("\n");
+#if !defined(CONFIG_STACKMONITOR_DISABLE_DEAD_THREAD)
 		stkmon_print_inactive_list();
 		printf("------|----------|----------");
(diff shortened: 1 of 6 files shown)


seokhun-eom24 (approved):

apps/system/utils/utils_stackmonitor.c
```diff
@@ -259,6 +259,7 @@ static void *stackmonitor_daemon(void *arg)
 		printf("|------------");
 #endif
 		printf("\n");
+#ifndef CONFIG_STACKMONITOR_DISABLE_DEAD_THREAD
```
jeongarmy: Could you change it to `#if !defined(XX)` for readability?

gidori98:
LGTM!

apps/system/utils/utils_stackmonitor.c
pcs1265: Resolved.

PR #6949, opened 2025-08-29
os/board/rtl8730e: fix task size before calling kernel_thread

yeetee179:
- kernel_thread takes in task stack size in bytes, no need to multiply by word size

### DIFF
--- a/os/board/rtl8730e/src/component/bluetooth/os/tizenrt/osif_tizenrt.c
+++ b/os/board/rtl8730e/src/component/bluetooth/os/tizenrt/osif_tizenrt.c
@@ -115,7 +115,6 @@ bool osif_task_create(void **pp_handle, const char *p_name, void (*p_routine)(vo
 	task_info[0] = itoa((int) p_routine, routine_addr, 16);
 	task_info[1] = itoa((int) p_param, param_addr, 16);
 	task_info[2] = NULL;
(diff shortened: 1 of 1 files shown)


os/board/rtl8730e/src/component/bluetooth/os/tizenrt/osif_tizenrt.c
```diff
@@ -115,7 +115,6 @@ bool osif_task_create(void **pp_handle, const char *p_name, void (*p_routine)(vo
 	task_info[0] = itoa((int) p_routine, routine_addr, 16);
 	task_info[1] = itoa((int) p_param, param_addr, 16);
 	task_info[2] = NULL;
-	stack_size = stack_size * sizeof(uint32_t);
```
jeongarmy, line 118: It looks good change because it makes mistakes.
But this change makes all stack sizes of threads created by this function 4 times smaller than before.
Did you verify all operations related to this change?
Is it okay to not change the stack size of the part using `osif_task_create`?
Please test all related operations to avoid stack overflow later.

hs36-kim: I had checked that.
image
It seems that it is okay to delete it because it was confirmed to be a code that was included because of freeRTOS that allocates in word units.

PR #6950, opened 2025-09-01
os/kernel/debug: Move and Rename dead thread stackinfo config

pcs1265:
The option `CONFIG_STACKMONITOR_DISABLE_DEAD_THREAD` was introduced in PR #6945,
but its placement under the Stackmonitor section and its name do not reflect its real purpose.
Saving dead‑thread stack information is a generic kernel debug feature, not a Stackmonitor‑specific one.

This commit includes:
* Move the option from the Stackmonitor section to the Debug Features section.
* Rename it to `CONFIG_DEBUG_DISABLE_DEAD_THREAD_STACKINFO`.

### DIFF
--- a/os/syscall/syscall_lookup.h
+++ b/os/syscall/syscall_lookup.h
@@ -87,73 +87,29 @@ SYSCALL_LOOKUP(sem_post,                  1, STUB_sem_post)
 SYSCALL_LOOKUP(sem_timedwait,             2, STUB_sem_timedwait)
 SYSCALL_LOOKUP(sem_trywait,               1, STUB_sem_trywait)
 SYSCALL_LOOKUP(sem_wait,                  1, STUB_sem_wait)
-
-#ifdef CONFIG_PRIORITY_INHERITANCE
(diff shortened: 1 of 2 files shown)


seokhun-eom24 (approved):
LGTM

os/Kconfig.debug
```diff
@@ -1277,6 +1277,15 @@ config STACK_COLORATION
 
 		Only supported by a few architectures.
 
+config DEBUG_DISABLE_DEAD_THREAD_STACKINFO  
+	bool "Disable dead threads stackinfo"  
+	default n
+	depends on DEBUG
```
jeongarmy: I think it should have a dependency on stack coloration.
Please add `if STACK_COLORATION` or `depends on STACK_COLORATION`

pcs1265: As the `ENABLE_STACKMONITOR` selects `STACK_COLORATION`, I thought it is redundant.
but for double check purpose, it does not matter I think.

Resolved.

pcs1265: I didn't push the commit. Sorry...

PR #6952, opened 2025-09-02
arch, board: Remove deprecated and unused PM code blocks

ewoodev:
This commit removes old and incompatible PM related code.

The power management (PM) logic in TizenRT has undergone significant refactoring, rendering these older code blocks obsolete. They are no longer compatible with the current PM architecture and are not supported in recent versions of Tizen Lite.

This change reduces maintenance overhead by eliminating dead code and improves code clarity by removing outdated logic, which helps prevent developer confusion.
This change is a pure cleanup and does not introduce any functional modifications to the current, active PM system.


os/arch/arm/src/amebad/amebad_idle.c
```diff
@@ -91,31 +90,6 @@
  * Public Functions
  ****************************************************************************/
 
-#ifdef CONFIG_PM
```
seokhun-eom24, line 94: I think we should delete line 84 in this file.
And also os/arch/arm/src/amebalite/amebalite_idle.c line 84.

os/board/stm32l4r9ai-disco/src/stm32_buttons.c
```diff
@@ -62,12 +62,6 @@
 
 /* Button Power Management */
 
-#ifdef CONFIG_PM
```
pcs1265, line 65: Line 57 in this file can be deleted too.

os/arch/arm/src/amebad/amebad_idle.c
eunwoo-nam: Updated

os/board/stm32l4r9ai-disco/src/stm32_buttons.c
eunwoo-nam: Updated

PR #6953, opened 2025-09-02
Fix pm api used without pm domain registered during boot

ewoodev:


os/drivers/input/ist415.c
```diff
@@ -1261,6 +1261,12 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
 		goto cleanup_sems_mem;
 	}
 
+	dev->pm_domain = pm_domain_register("IST415");
+	if (dev->pm_domain < 0) {
+		ist415dbg("Fail to register pm domain\n");
```
sunghan-chang: Regardless of this commit, Is it possible to know why it fails with this log?

seokhun-eom24 (approved):
LGTM

os/drivers/input/ist415.c
```diff
@@ -1261,6 +1261,12 @@ int ist415_initialize(const char *path, struct i2c_dev_s *i2c, struct ist415_con
...
+		goto cleanup_wd;
```
kishore-sn: Please change to cleanup_task

eunwoo-nam: oh right!, Updated. thank you 👍👍

os/drivers/input/ist415.c
eunwoo-nam: yes, we can get reason using get_errno(), I added, thank you 😊😊

PR #6954, opened 2025-09-02
os/.gitignore: Add gitignore ".download", "app_size_file" files

ewoodev:
the .download file is generated during bianry download script and app_size_file is generated in mkverifyappsize.py os.system('size ' + file_path + ' > ' + APP_SIZE_FILE)

So two files are generated by script.
Therefore, Add gitignore

### DIFF
--- a/os/.gitignore
+++ b/os/.gitignore
@@ -19,3 +19,5 @@
 /.appSpec
 /.bininfo
 /dutils_output_*.txt
+/.download
+/app_size_file


sunghan-chang:
@ewoodev If we delete them at end of download, what happen?

seokhun-eom24 (approved):
LGTM

eunwoo-nam:
> @ewoodev If we delete them at end of download, what happen?

This is not the data required during the download process.

To improve the inconvenience of selecting a port each time, we save the previously used port and allow the selected port to be chosen simply by pressing Enter during the next download.

PR #6956, opened 2025-09-04
Refactor PM memory struct

ewoodev:


eunwoo-nam:
If this PR is merged, it will cause a code error related to the incorrect use of the PM domain during booting. 
Therefore, #6953 PR must be merged first.

PR #6958, opened 2025-09-04
os/: Add PWM frequency set for ST7785 MIPI LCD

namanjain7:
Add PWM frequency set for MIPI based LCD ST7785, initial value is set to 100, can be changed from Kconfig

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -234,6 +234,10 @@ static void rtl8730e_gpio_init(void)
 	pwmout_init(&g_rtl8730e_config_dev_s.pwm_led, GPIO_PIN_BACKLIGHT);
 	lcdvdbg("initial pwm read %f\n", pwmout_read(&g_rtl8730e_config_dev_s.pwm_led));
 #endif
+
+#if defined(LCD_BACKLIGHT_PWM_FREQ_KHZ)
+	pwmout_period_us(&g_rtl8730e_config_dev_s.pwm_led, 1000 / LCD_BACKLIGHT_PWM_FREQ_KHZ);
+#endif
 }
 
 static void rtl8730e_control_backlight(uint8_t level)
--- a/os/include/tinyara/lcd/st7785.h
+++ b/os/include/tinyara/lcd/st7785.h
@@ -202,4 +202,6 @@ static const lcd_vendor_map_t g_lcd_vendors[] = {
 
 #define NUM_LCD_VENDERS (AVD_NB01 + AVD_2701 + HLT_A196 + TEC_181A)
 
+#define LCD_BACKLIGHT_PWM_FREQ_KHZ 20
+
 #endif	/* __DRIVERS_LCD_ST7785_H */


build/configs/rtl8730e/loadable_ext_ddr_st7785/defconfig
```diff
@@ -621,6 +621,7 @@ CONFIG_LCD_INIT_HLT_A196=y
 # CONFIG_LCD_INIT_TEC_181A is not set
 CONFIG_LCD_XRES=240
 CONFIG_LCD_YRES=320
+CONFIG_LCD_BACKLIGHT_PWM_FREQUENCY=100
```
sunghan-chang: Let's split the commit for code change and defconfig change.

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -229,6 +229,14 @@ static void rtl8730e_gpio_init(void)
...
 	lcdvdbg("initial pwm read %f\n", pwmout_read(&g_rtl8730e_config_dev_s.pwm_led));
 #endif
+
+#if defined(CONFIG_LCD_ST7785)
+#if CONFIG_LCD_BACKLIGHT_PWM_FREQUENCY > 0
+	pwmout_period_us(&g_rtl8730e_config_dev_s.pwm_led, 1000 / CONFIG_LCD_BACKLIGHT_PWM_FREQUENCY);
+#else
+#error value LCD_BACKLIGHT_PWM_FREQUENCY is invalid
+#endif
+#endif
```
sunghan-chang: Is this ST7785 specific functionality? Is this necessary here? We should split common part and lcd specific part in file.

os/drivers/lcd/Kconfig
```diff
@@ -222,6 +222,14 @@ config LCD_LOGO_52_340
 endchoice
 endif
 
+if LCD_ST7785
+config LCD_BACKLIGHT_PWM_FREQUENCY
+	int "LCD Backlight PWM Frequency"
```
sunghan-chang: Could you write the unit of the frequency? in Hz? in MHz?

build/configs/rtl8730e/loadable_ext_ddr_st7785/defconfig
sunghan-chang: 100 is a proper value? @ewoodev Could you confirm?

os/drivers/lcd/Kconfig
```diff
@@ -222,6 +222,14 @@ config LCD_LOGO_52_340
 endchoice
 endif
 
+if LCD_ST7785
```
sunghan-chang: Could you let me know why this is under LCD_ST7785? No way to set this in other LCDs?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
namanjain7: Hello  Mr. chang, 
as per Mr. nam, This pwm set is only for st7785 LCD.
so pwmout_period_us can't be called for other LCD's and this funciton is defined in low level driver, so we have to call it from rtl8730e_mipi_lcdc.c file, and not from driver.

os/drivers/lcd/Kconfig
namanjain7: as per Mr. nam, This pwm set is only for st7785 LCD.

os/drivers/lcd/Kconfig
namanjain7: it's in kHz, I have added in Kconfig help and message.

build/configs/rtl8730e/loadable_ext_ddr_st7785/defconfig
namanjain7: done

os/drivers/lcd/Kconfig
eunwoo-nam: Hello @namanjain7 
yes, right, the st7785 lcd is only reqired pwm 20khz setting from vender.

However, this topic should be addressed differently.

The LCD backlight is essentially a separate device from the LCD driver, serving as a light source. 
(the LCD IC and backlight IC are hardware-separated)
Therefore, this discussion is more appropriate not to limit to ST7785. If you were adding a frequency setting value for backlight PWM in Kconfig for tuning , it would be better to unrelated to ST7785.

Additionally, we are looking to improve the overall LCD structure. I think, this should also be a goal for inprove the LCD structure.
I suggest separating the backlight control PWM. Furthermore, I suggest that the LCD driver receives app LCD on/off requests as enable/disable commands to control the PWM driver.
It can support various types of LCDs by separating device drivers for each hw.

build/configs/rtl8730e/loadable_ext_ddr_st7785/defconfig
eunwoo-nam: The vender guidence is 10khz ~ 30khz
So, Let's set 20khz. it is verifiing on other system.

os/drivers/lcd/Kconfig
namanjain7: Got it.
Currently this lcd set pwm freq will be done during init.
If the application is changing the freq in middle of LCD running, then it's beneficial to add ioctl support for it.
If not then we can add this in st7785.c file (In check lcd vendor and send init command), which will be lcd specific.
what do you think? @ewoodev

seokhun-eom24 (commented):
As you noted in the previous defconfig, it would be helpful to add an explanation in the commit message describing how the value 20 was derived, so that when the value needs to be changed later we know how to calculate and set it.

os/drivers/lcd/Kconfig
eunwoo-nam: The application never change pwm freq. 
✔ the app should not be allowed to control direct lcd hw. 
✔ app doesn't know hw each hw spec. 

So, current lcd init struct, st7785.c is better. 👀✨

namanjain7:
> As you noted in the previous defconfig, it would be helpful to add an explanation in the commit message describing how the value 20 was derived, so that when the value needs to be changed later we know how to calculate and set it.

Hello Mr. Eom,
The value 20 is vendor specific, given by vendor.

os/include/tinyara/lcd/st7785.h
```diff
@@ -202,4 +202,6 @@ static const lcd_vendor_map_t g_lcd_vendors[] = {
 
 #define NUM_LCD_VENDERS (AVD_NB01 + AVD_2701 + HLT_A196 + TEC_181A)
 
+#define LCD_BACKLIGHT_PWM_FREQUENCY_VALUE 20
```
eunwoo-nam: How about LCD_BACKLIGHT_PWM_FREQ_KHZ?

the naming can represent  khz unit

namanjain7:
os/: Add PWM frequency set for ST7785 MIPI LCD

  Add PWM frequency set for MIPI based lcd, enabled for st7785, initial value is set to 20, can be changed in respective .h defined lcd file
  App is not allowed to change the LCD_BACKLIGHT_PWM_FREQ_KHZ value. This value is given by vendor.

PR #6960, opened 2025-09-05
/os/mm: Unify mm apis like mm_alloc, mm_realloc, mm_memalign, mm_zalloc, mm_calloc

rish-sg:
Here unifying means merging the two conditional definitions of apis into one as per the requirement to remove #ifdef CONFIG_DEBUG_MM_HEAPINFO in mm apis.
For ex :
As-is:
FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_retaddr)
#else
FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size)
#endif
{
...
}

To-Be: 

FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_retaddr)
{
...
}

### DIFF
--- a/os/include/tinyara/mm/mm.h
+++ b/os/include/tinyara/mm/mm.h
@@ -242,9 +242,9 @@ typedef size_t mmsize_t;
 #endif
 
 /* typedef is used for defining size of address space */
+typedef void *mmaddress_t;             /* 32 bit address space */
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-typedef void *mmaddress_t;             /* 32 bit address space */
 
 #define SIZEOF_MM_MALLOC_DEBUG_INFO \
 	(sizeof(mmaddress_t) + sizeof(pid_t) + sizeof(uint16_t))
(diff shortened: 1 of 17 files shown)


os/include/tinyara/mm/mm.h
```diff
@@ -543,69 +536,43 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem);
 void kmm_free(FAR void *mem);
 #endif
 
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-
 /* Functions contained in mm_realloc.c **************************************/
+/* unified mm_realloc api */
```
sunghan-chang: Is this necessary?

os/mm/umm_heap/umm_malloc.c
```diff
@@ -229,16 +222,12 @@ FAR void *malloc(size_t size)
...
 	/* User supports a single heap on app separation */
-	ret = mm_malloc(BASE_HEAP, size
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-			, caller_retaddr
-#endif
-			);
+	ret = mm_malloc(BASE_HEAP, size, caller_retaddr);
 	if (ret == NULL) {
 		mm_manage_alloc_fail(BASE_HEAP, HEAP_START_IDX, HEAP_END_IDX, size, 0, USER_HEAP
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
```
sunghan-chang: You should check this as well for the same purpose.

os/mm/umm_heap/umm_malloc.c
```diff
@@ -229,16 +222,12 @@ FAR void *malloc(size_t size)
 	}
 
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
+	caller_retaddr = GET_RETURN_ADDRESS();
```
sunghan-chang: This should be the same in both enabled and disabled CONFIG_BUILD_KERNEL.
Yes, original code does not have it in enabled. But it's wrong.
Let's set this at  line 190.

os/mm/umm_heap/umm_malloc.c
sunghan-chang: same in other APIs

sunghan-chang (commented):
Please find our commit rules at the Wiki.
The title should have the format like
```
path: title

description
```
We don't allow empty description either.

PR #6961, opened 2025-09-08
os/mm : Unify mm apis like mm_malloc, mm_calloc, mm_realloc, mm_zalloc, mm_memalign

rish-sg:
Description:
Contains changes for the unification of conditional definition of mm apis for the requirement to remove #ifdef CONFIG_DEBUG_MM_HEAPINFO in mm apis.
For ex:
As is:
FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_retaddr)
#endif
FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size) 
{
...
}

To be:
FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_retaddr) 
{
...
}

### DIFF
--- a/os/mm/kmm_heap/kmm_calloc.c
+++ b/os/mm/kmm_heap/kmm_calloc.c
@@ -67,22 +67,15 @@
 /****************************************************************************
  * Private Functions
  ****************************************************************************/
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
+
 static void *kheap_calloc(size_t n, size_t elem_size, mmaddress_t retaddr)
-#else
-static void *kheap_calloc(size_t n, size_t elem_size)
-#endif
 {
 	int heap_idx;
 	void *ret;
 	struct mm_heap_s *kheap = kmm_get_baseheap();
 
 	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
-		ret = mm_calloc(&kheap[heap_idx], n, elem_size
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-				, retaddr
-#endif
-				);
+		ret = mm_calloc(&kheap[heap_idx], n, elem_size, retaddr);
 		if (ret != NULL) {
 			return ret;
 		}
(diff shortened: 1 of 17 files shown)


os/mm/kmm_heap/kmm_calloc.c
```diff
@@ -67,22 +67,15 @@
...
 	for (heap_idx = HEAP_START_IDX; heap_idx <= HEAP_END_IDX; heap_idx++) {
-		ret = mm_calloc(&kheap[heap_idx], n, elem_size
-#ifdef CONFIG_DEBUG_MM_HEAPINFO
-				, retaddr
-#endif
-				);
+		ret = mm_calloc(&kheap[heap_idx], n, elem_size, retaddr);
 		if (ret != NULL) {
 			return ret;
 		}
```
sunghan-chang, line 81: Please check the `mm_manage_alloc_fail` as well in each allocation API.
In this file, line 85.

os/mm/kmm_heap/kmm_memalign.c
```diff
@@ -144,16 +142,14 @@ FAR void *kmm_memalign(size_t alignment, size_t size)
 	if (size == 0) {
 		return NULL;
 	}
+
+	mmaddress_t caller_retaddr = NULL;	/* for generalising the call to mm_memalign api */
 #ifdef CONFIG_DEBUG_MM_HEAPINFO
-	mmaddress_t caller_retaddr = GET_RETURN_ADDRESS();
+	caller_retaddr = GET_RETURN_ADDRESS();
 #endif
```
sunghan-chang, line 145: Usually we define the config, then it will do like below
1. assign null to the variable
2. reassign new address to the variable again

To avoid unnecessary assignment for performance, how about below?
```
mmaddress_t caller_retaddr;   (at start of the function)

if (size == 0) {
  ...
}

#ifdef CONFIG_....
caller... = GET_...();
#else
caller = NULL;
#endif
```

sunghan-chang:
@rish-sg Could you leave how to verify the change in the commit description? (not in the PR description)

os/mm/kmm_heap/kmm_memalign.c
rish-sg: We can do it the way you've suggested but since `mmaddress_t caller_retaddr` is a stack variable, I don't think the NULL assignment would affect the performance as such.

sunghan-chang (approved):

PR #6965, opened 2025-09-12
os/board/rtl8730e: Increase maximum number of APs in scan results

jwei5:
- Max number of APs in wifi scan reuslts is set to 64 by default
- Use parameter max_ap_record_num to increase the max number of APs in wifi scan results to 100 in both wifi_netmgr_utils_scan_ap() and wifi_netmgr_utils_scan_multi_ap()


hk-gwak (approved):
LGTM

PR #6966, opened 2025-09-15
pm, arch: Refactor entering pm sleep

ewoodev:
This commit refactors entering pm sleep architecture to improve maintainability and add ease and efficiency of new chipset porting.

- Refactor pm_idle() logic into smaller, modular functions
- Remove pm_wakehandler and merge it into pm_idle
- Add skeleton function to clarify the pm sleep entring

The new architecture provides better separation of concerns and makes it easier to add new PM features in the future.

Review focus
- refactoring structure for pm_idle
- side effect


os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -66,23 +84,25 @@ struct pm_clock_ops rtl8730e_clock_ops = {
  *
  ****************************************************************************/
 
-void up_pm_board_sleep(void (*wakeuphandler)(clock_t, pm_wakeup_reason_code_t))
+static int up_pm_board_sleep(void)
 {
 	/* mask sys tick interrupt*/
 	arm_arch_timer_int_mask(1);
```
anjana348, line 90: @lhenry-realtek  Can we move arm_arch_timer_int_mask(1) inside up_timer_disable ? 
As up_timer_disable is used from various places, will it cause any side effect?

Similarly for up_timer_enable

lhenry-realtek: Hi, please check the calling core, this api is core-specific.

If cpu0 calls it, then cpu0 timer will pause/enable, same for cpi1

anjana348: ok so can we create two functions if want to move timer logic from bsp to os(pm_idle.c) like
 void up_timer_disable(void)
{
  arm_arch_timer_enable(0);
}

// NEW: Specific function for preparing CPU0 for deep sleep.
void up_timer_disable_for_master_sleep(void)
{
  arm_arch_timer_int_mask(1);
  arm_arch_timer_enable(0);
}

eunwoo-nam: > Hi, please check the calling core, this api is core-specific.
> 
> If cpu0 calls it, then cpu0 timer will pause/enable, same for cpi1

Hello @lhenry-realtek 

which api is core-sepecific??
is arm_arch_timer_int_mask for only primary core?

Could you please confirm below question?
```
void up_timer_disable()
{
	arm_arch_timer_int_mask(1);  <-------------- can be called cpu1 during cpu pause?????
	arm_arch_timer_enable(0);
}

void up_timer_enable()
{
	arm_arch_timer_set_compare(arm_arch_timer_count() + SYSTICK_RELOAD); 
	arm_arch_timer_enable(1);
	// arm_arch_timer_set_compare(arm_arch_timer_count() + 50000);  <--------------  SYSTICK_RELOAD == 50000 ??
	arm_arch_timer_int_mask(0);    <-------------- can be called cpu1 during cpu pause?????
}
```

lhenry-realtek: All the APIs inside up_timer_enable and up_timer_disable are per-core

Same for mask, the reason is each core has its own set of timers, this is the PPI

For example you can observe up_timerisr, it checks up_systimer_is_paused and the argument is current cpu id, it will then proceed to disable the timer for that cpu

eunwoo-nam: yes, thank you for explain
we just wanted to know effect arm_arch_timer_int_mask api in PG flow and cpu pause.

https://github.com/Samsung/TizenRT/blob/64de270e0bebfe3539ee8fcc6ac69ef787635899/os/board/rtl8730e/src/component/soc/amebad2/fwlib/ap_core/arch_timer.h#L38-L71

if arm_arch_timer_enable(0) is called, CNTV_CTL_IMASK is cleared forced.

so even if we call arm_arch_timer_int_mask(1) first, CNTV_CTL_IMASK is cleared when arm_arch_timer_enable(0) is called. 
and clear CNTV_CTL_ENABLE. 

So actually calling arm_arch_timer_int_mask() is not neccesary.

eunwoo-nam: @lhenry-realtek 
May I know history why below code is disabled in arm_arch_timer_enable??

	//__asm__ volatile("mrc p15, 0, %0, c14, c3, 1\n\t"
	//				 : "=r"(cntv_ctl) :  : "memory");


Is it for initializing timer surely?

lhenry-realtek: Not sure, the code was already present. 
However it just seems like it's reading the value before setting the bit then write back to register, I think whoever commented it thinks it's not necessary

eunwoo-nam: Thank you for confirmation. lhenry 👍👍

firstly, Let me keep timer operation code in this PR
and change it next PR.

sunghan-chang (approved):

sunghan-chang:
@lhenry-realtek @anjana348 Could you leave approvals if you don't have any other comment?

os/pm/pm.h
```diff
@@ -358,22 +358,44 @@ void pm_metrics_update_idle(void);
 
 
 /****************************************************************************
- * Name: pm_metrics_update_wakehandler
+ * Name: pm_metrics_update_wakeup_reason
  *
  * Description:
  *   This function is called inside pm_wakehandler. It counts the frequency of wakeup
```
seokhun-eom24, line 364: I think we should update the description. 
The `pm_wakehandler` function no longer exists.
`pm_wakehandler` -> `enable_and_compensate_systick`

Also please revise the description for `pm_metrics_update_wakeup_reason` and `pm_metrics.c` file.

eunwoo-nam: oh, thank you for review

I updated

os/pm/pm_idle.c
```diff
@@ -49,6 +49,415 @@ static clock_t stime;
...
+ *
+ ****************************************************************************/
+static int disable_systick(void)
+{
+	/* TODO: move disable systick code BSP to here */
+	return OK;
+}
+
+/****************************************************************************
+ * Name: enable_and_conpensate_systick
```
neel-samsung: could you please check the typo here ```enable_and_compensate_systick``` ?

os/include/tinyara/pm/pm.h
```diff
@@ -623,6 +623,8 @@ void pm_dvfs(int div_lvl);
  *
  ****************************************************************************/
 int pm_metrics(int milliseconds);
```
r-prabu, line 625: @ewoodev how about the code size if we dont want to use the pm_metrics at all?

os/arch/arm/src/amebasmart/amebasmart_idle.c
```diff
@@ -66,23 +84,25 @@ struct pm_clock_ops rtl8730e_clock_ops = {
  *
  ****************************************************************************/
 
-void up_pm_board_sleep(void (*wakeuphandler)(clock_t, pm_wakeup_reason_code_t))
+static int up_pm_board_sleep(void)
```
Taejun-Kwon, line 87: There's no reason for int here.

os/pm/pm_idle.c
eunwoo-nam: Thank you for review, neel 👍
I fixed!

Taejun-Kwon (commented):
I think you should draw diagram to make less coupling. Currently many component tightly coupled with global var 'g_pmglobals' & many extern values. it need to be improved.

os/arch/arm/src/amebasmart/amebasmart_idle.c
eunwoo-nam: Thank you for review 😊
yes, current amebasmart chipset is not proveided return OK and ERROR.

but the board sleep operation can have return ERROR.
so we need to privide return int type HAL API.

eunwoo-nam:
> I think you should draw diagram to make less coupling. Currently many component tightly coupled with global var 'g_pmglobals' & many extern values. it need to be improved.

`g_pmglobals` global value is used in os/pm directory only.
this value is privded only using internal header os/pm/pm.h 

`sem_t regsem` and `dq_queue_t registry` 
is used in  `pm_initialize` `pm_register` `pm_unregister`  `pm_changestate`

`dq_queue_t domains`, and `uint16_t ndomains` 
is used `pm_initialize` `pm_domain_register` `pm_domain_unregister` `pm_metrics` `pm_procfs`

`dq_queue_t suspended_domains` and `int8_t state`
 is used in `pm_initialize` `pm_checkstate` `pm_suspend` `pm_resume` `pm_metrics`

`bool is_running;` 
is used  `pm_initialize` `pm_metrics` `pm_idle`

`struct pm_sleep_ops *sleep_ops`
 is used  `pm_initialize` `pm_idle`

The almost values are used in almost pm code. It's not used out of pm. so i think, it's okay to current code
but I'm ready to improve it more. If you have any idea, Could you please let me know your opinion? 😊

eunwoo-nam:
And @lhenry-realtek
we modifed amebasmart code temporary in this PR. please modify the realtek code again according to your coding rules after this PR merged 😉

eunwoo-nam:
For a more comfortable review, I separated the commit into different PRs, not the main agenda here.
https://github.com/Samsung/TizenRT/pull/6978

Taejun-Kwon:
> > I think you should draw diagram to make less coupling. Currently many component tightly coupled with global var 'g_pmglobals' & many extern values. it need to be improved.
> 
> `g_pmglobals` global value is used in os/pm directory only. this value is privded only using internal header os/pm/pm.h
> 
> `sem_t regsem` and `dq_queue_t registry` is used in `pm_initialize` `pm_register` `pm_unregister` `pm_changestate`
> 
> `dq_queue_t domains`, and `uint16_t ndomains` is used `pm_initialize` `pm_domain_register` `pm_domain_unregister` `pm_metrics` `pm_procfs`
> 
> `dq_queue_t suspended_domains` and `int8_t state` is used in `pm_initialize` `pm_checkstate` `pm_suspend` `pm_resume` `pm_metrics`
> 
> `bool is_running;` is used `pm_initialize` `pm_metrics` `pm_idle`
> 
> `struct pm_sleep_ops *sleep_ops` is used `pm_initialize` `pm_idle`
> 
> The almost values are used in almost pm code. It's not used out of pm. so i think, it's okay to current code but I'm ready to improve it more. If you have any idea, Could you please let me know your opinion? 😊

1. Since you decide control pm based on driver model, each of operation need to be maintained by private of iNode. but currently os_start & many other place control it. 
2. so instead of global variable, it should be maintained by dev (pm.c)
3. As a lower driver, amebasmart_idle should give pointer of pm operation.
4. That's why I suggest draw diagram to make this clearly. It will help you make more good structure. Currently this is some kind of mixed model, half is driver, half is pure kernel feature. you should define interface between pm driver & pm component clearly.

eunwoo-nam:
> > > I think you should draw diagram to make less coupling. Currently many component tightly coupled with global var 'g_pmglobals' & many extern values. it need to be improved.
> > 
> > 
> > `g_pmglobals` global value is used in os/pm directory only. this value is privded only using internal header os/pm/pm.h
> > `sem_t regsem` and `dq_queue_t registry` is used in `pm_initialize` `pm_register` `pm_unregister` `pm_changestate`
> > `dq_queue_t domains`, and `uint16_t ndomains` is used `pm_initialize` `pm_domain_register` `pm_domain_unregister` `pm_metrics` `pm_procfs`
> > `dq_queue_t suspended_domains` and `int8_t state` is used in `pm_initialize` `pm_checkstate` `pm_suspend` `pm_resume` `pm_metrics`
> > `bool is_running;` is used `pm_initialize` `pm_metrics` `pm_idle`
> > `struct pm_sleep_ops *sleep_ops` is used `pm_initialize` `pm_idle`
> > The almost values are used in almost pm code. It's not used out of pm. so i think, it's okay to current code but I'm ready to improve it more. If you have any idea, Could you please let me know your opinion? 😊
> 
> 1. Since you decide control pm based on driver model, each of operation need to be maintained by private of iNode. but currently os_start & many other place control it.
> 2. so instead of global variable, it should be maintained by dev (pm.c)
> 3. As a lower driver, amebasmart_idle should give pointer of pm operation.
> 4. That's why I suggest draw diagram to make this clearly. It will help you make more good structure. Currently this is some kind of mixed model, half is driver, half is pure kernel feature. you should define interface between pm driver & pm component clearly.

I glad to disscuss detail about this with you. 😀😀

Firstly, I am thinking PM layer as an between the driver and the kernel(scheduler)
The reason for this is that all drivers can call the PM's pm_suspend/pm_resume, and through pm_register, they can register callbacks to be invoked during entry into PM sleep mode.

In this process, PM call kernel API to determine the scheduling state.
Therefore, this cannot be based on the driver model.

There is a diagram, but it cannot be uploaded due to security policies.😥
To facilitate smoother discussions, I will create a channel for this purpose.

i think, the discussion you proposed is seem separate from the current PR's refactoring.
we will discussion more it and we can improve the structure more on additional PR.
(The PR changes are already big.. so it's better to seperate.)

os/include/tinyara/pm/pm.h
eunwoo-nam: enable --> KERNEL      1,695,338 bytes    2,936,832 bytes     57.73%    PASS
disable -->  KERNEL      1,693,010 bytes    2,936,832 bytes     57.65%    PASS


if we disable pm_metrics, the 2.3k flash memory is reduced.

sunghan-chang (approved):

PR #6972, opened 2025-09-18
arch: armv7-a: Fix deadlock between CPU hotplug and IRQ critical section

ewoodev:
Fix a deadlock issue that occurs after recent systick control migration commits (0192d1ef1, 96706497d, 6e46e19bc) where CPU1 receiving IRQ interrupts.

The deadlock scenario:
- CPU0 calls enter_critical_section() and requests CPU1 hotplug, then waits
- CPU1 receives tick IRQ and calls enter_critical_section(), waiting for completion
- This creates a circular dependency where both CPUs wait for each other

Solution:
when a secondery CPU is waiting for critical section entry while a hotplug request is pending, abort the hotplug operation and pm entering sleep.

Add Optimize spinlock-based synchronization for CPU hotplug requests.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
+++ b/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
@@ -28,6 +28,7 @@
 #include <stdint.h>
 #include <assert.h>
 #include <debug.h>
+#include <errno.h>
 
 #include <tinyara/arch.h>
 #include <tinyara/sched.h>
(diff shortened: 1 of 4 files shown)


os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -85,23 +170,40 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
...
+		spin_lock(&g_cpu_hotplughandled[cpu]);
+		/* Release the spinlock to synchronize with the requesting CPU. */
+		spin_unlock(&g_cpu_hotpluged[cpu]);
+
+		/* Save the tcb state */
+		struct tcb_s *rtcb = this_task();
+		arm_savestate(rtcb->xcp.regs);
+		rtcb->task_state = TSTATE_TASK_ASSIGNED;
+		CURRENT_REGS = NULL;
+
```
lhenry-realtek, line 195: Hi @ewoodev 

Please check if #6707 changes are required: clear SGI4 interrupt for CPU1 at GIC before psci_cpu_off 
Since this PR also touch hotplug, it may be applied to handle both PG and future runtime hotplug case

Refer:
https://github.com/Samsung/TizenRT/blob/145c56a89bec45dbedc779ad918414e7d2497aa3/os/arch/arm/src/armv7-a/arm_cpuhotplug.c#L113-L118

eunwoo-nam: Thank you for your review and feedback.

As you described in #6707, this appears to be the necessary code.

However, since it has not yet been verified for any issues, it is better to separate PR.
It's not effected because clear the register after board wakeup on current version.
(We have verified it)

We plan to check and apply suggestion you have made.👍👍

amandeep-samsung: Hello @ewoodev 
It seems we should add below code as told by Henry
/*  
 		manually clear SGI4 interrupt active 
 		As psci_cpu_off is a noreturn function it will trap the CPU in a WFI loop  
 		this means that it will never exit back to arm_decodeirq where EOIR will be set 
 	*/ 
 	putreg32(irq, GIC_ICCEOIR); 

This is needed code. We can verify it more before raising in new PR for this part of code.

eunwoo-nam:
The PR adds waiting receving hoplug handling.
so, the gating release is not performed. 
therefore, https://github.com/Samsung/TizenRT/pull/6826 must be merged first.

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
eunwoo-nam: It is raised https://github.com/Samsung/TizenRT/pull/6977 👍👍👍

we will merge in this page PR.

PR #6974, opened 2025-09-18
framework/src/eventloop: Optimize callback list management with dynamic allocation

pcs1265:
Replace static callback list array with pointer array to enable on-demand dynamic allocation of callback list structures.
This change improves memory efficiency by allocating actual callback list data only when needed,
rather than pre-allocating for all possible tasks.

Memory Usage Breakdown:
- thread_safe_cb_list_t structure: 24 bytes total
  - `sq_queue_t cb_list`: 8 bytes (double linked list head)
  - `el_async_t *async_handle`: 4 bytes (pointer)
  - `sem_t list_sem`: 12 bytes (semaphore with CONFIG_PRIORITY_INHERITANCE)

Memory Usage:
- Previous: 24 bytes * CONFIG_MAX_TASKS (static allocation for all tasks)
- Current: 4 bytes * CONFIG_MAX_TASKS (pointer array) + 24 bytes * tasks_using_eventloop (dynamic allocation)

Examples (tasks_using_eventloop = 10):

Case 1: CONFIG_MAX_TASKS = 128
- Previous: 24 * 128 = **3,072 bytes**
- Current: (4 * 128) + (24 * 10) = 512 + 240 = **752 bytes**
- Memory savings: 3,072 - 752 = 2,320 bytes **(75.5% reduction)**

Case 2: CONFIG_MAX_TASKS = 256
- Previous: 24 * 256 = **6,144 bytes**
- Current: (4 * 256) + (24 * 10) = 1,024 + 240 = **1,264 bytes**
- Memory savings: 6,144 - 1,264 = 4,880 bytes **(79.4% reduction)**

Index Access:
- Uses PID hash: THREAD_SAFE_CB_LIST_IDX_HASH(pid) = (pid) & MAX_PID_MASK
- Maintains O(1) access time through direct hash indexing
- Index calculation unchanged from existing implementation

Allocation Strategy:
- Previous: Static allocation at compile time for all CONFIG_MAX_TASKS entries
- Current: Static pointer array + dynamic allocation of callback list structures on-demand
- Allocation occurs in `eventloop_thread_safe_cb_list_node_init()` when first callback registered
- Deallocation happens in `eventloop_thread_safe_cb_list_node_deinit()` when no more callbacks

### DIFF
--- a/framework/src/eventloop/eventloop_async.c
+++ b/framework/src/eventloop/eventloop_async.c
@@ -51,16 +51,40 @@ struct thread_safe_cb_s {
 };
 typedef struct thread_safe_cb_s thread_safe_cb_t;
 
-static thread_safe_cb_list_t g_thread_safe_cb_list[CONFIG_MAX_TASKS];
+static thread_safe_cb_list_t *g_thread_safe_cb_list[CONFIG_MAX_TASKS];
 static sq_queue_t g_thread_safe_func_list;
 
-#define CB_LIST(idx)         g_thread_safe_cb_list[idx].cb_list
-#define ASYNC_HANDLE(idx)    g_thread_safe_cb_list[idx].async_handle
-#define LIST_SEM(idx)        g_thread_safe_cb_list[idx].list_sem
+#define CB_LIST(idx)         eventloop_get_cb_list(idx)
+#define ASYNC_HANDLE(idx)    eventloop_get_async_handle(idx)
+#define LIST_SEM(idx)        eventloop_get_list_sem(idx)
 
 /* Define for the hash operation */
 #define MAX_PID_MASK                       (CONFIG_MAX_TASKS - 1)
 #define THREAD_SAFE_CB_LIST_IDX_HASH(pid)  ((pid) & MAX_PID_MASK)
+
+static inline sq_queue_t *eventloop_get_cb_list(int idx)
+{
+	if (g_thread_safe_cb_list[idx]) {
+		return &g_thread_safe_cb_list[idx]->cb_list;
+	}
+	return NULL;
+}
+
+static inline el_async_t *eventloop_get_async_handle(int idx)
+{
+	if (g_thread_safe_cb_list[idx]) {
+		return g_thread_safe_cb_list[idx]->async_handle;
+	}
+	return NULL;
+}
+
+static inline sem_t *eventloop_get_list_sem(int idx)
+{
+	if (g_thread_safe_cb_list[idx]) {
+		return &g_thread_safe_cb_list[idx]->list_sem;
+	}
+	return NULL;
+}
 
 static thread_safe_func_t *eventloop_get_func_handle(thread_safe_callback func)
 {
@@ -126,14 +150,21 @@ static int eventloop_thread_safe_cb_list_node_init(int list_idx, thread_safe_fun
 		return EVENTLOOP_INVALID_PARAM;
 	}
 
-	ASYNC_HANDLE(list_idx) = (el_async_t *)EL_ALLOC(sizeof(el_async_t));
+	g_thread_safe_cb_list[list_idx] = (thread_safe_cb_list_t *)EL_ALLOC(sizeof(thread_safe_cb_list_t));
+	if (g_thread_safe_cb_list[list_idx] == NULL) {
+		eldbg("Failed to allocate callback list!\n");
+		return EVENTLOOP_OUT_OF_MEMORY;
+	}
+
+	g_thread_safe_cb_list[list_idx]->async_handle = (el_async_t *)EL_ALLOC(sizeof(el_async_t));
 	if (ASYNC_HANDLE(list_idx) == NULL) {
 		eldbg("Failed to allocate handle!\n");
+		EL_FREE(g_thread_safe_cb_list[list_idx]);
 		return EVENTLOOP_OUT_OF_MEMORY;
 	}
 	
-	sq_init(&CB_LIST(list_idx));
-	sem_init(&LIST_SEM(list_idx), 0, 1);
+	sq_init(CB_LIST(list_idx));
+	sem_init(LIST_SEM(list_idx), 0, 1);
 	return OK;
 }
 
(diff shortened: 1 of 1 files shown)


framework/src/eventloop/eventloop_async.c
```diff
@@ -51,17 +51,101 @@ struct thread_safe_cb_s {
...
+		if (curr->idx == idx) {
+			sem_post(&g_cb_list_sem);
+			return &curr->cb_list_data;
+		}
+		curr = (thread_safe_cb_list_node_t *)sq_next(curr);
+	}
+	sem_post(&g_cb_list_sem);
+
+	return NULL;
+}
```
eunwoo-nam: It seems like a good attempt to reduce unnecessary static memory allocation for unused PIDs and secure memory. 👍👍👍

```c
struct thread_safe_cb_list_s {
	sq_queue_t cb_list;
	el_async_t *async_handle;
	sem_t list_sem;
};
typedef struct thread_safe_cb_list_s thread_safe_cb_list_t;
```
✨Previously:  sizeof(struct thread_safe_cb_list_s) * CONFIG_MAX_TASKS which totals 24byte * CONFIG_MAX_TASKS.
✨PR change: sizeof(struct thread_safe_cb_list_s) * actual used PIDs + sizeof(g_thread_safe_cb_list)
                                 32 bytes  x actual used PIDs + 8 bytes

While there is a significant memory-saving effect, there is a concern about potential performance degradation.
The previous method was O(1), but the current change increases to O(n), which could become a performance issue if the event loop is frequently used.

✨I suggest using a PID hash, keeping the hashmap intact but changing the stored content to the address value of thread_safe_cb_list_t.
```c
static thread_safe_cb_list_t *g_thread_safe_cb_list[CONFIG_MAX_TASKS];

static int eventloop_thread_safe_cb_list_node_init(int list_idx, thread_safe_func_t *func_handle)
{
...
    g_thread_safe_cb_list[list_idx] = (thread_safe_cb_list_t *)malloc(sizeof(thread_safe_cb_list_t);

```

This can reduce memory usage while minimizing performance changes.
4 bytes * CONFIG_MAX_TASKS + sizeof(struct thread_safe_cb_list_s) * actual used PIDs

pcs1265: I implemented PID hash with pointer array - O(1) access + memory efficient.

Thank you for your suggestion.

framework/src/eventloop/eventloop_async.c
```diff
@@ -126,14 +150,19 @@ static int eventloop_thread_safe_cb_list_node_init(int list_idx, thread_safe_fun
 		return EVENTLOOP_INVALID_PARAM;
 	}
 
-	ASYNC_HANDLE(list_idx) = (el_async_t *)EL_ALLOC(sizeof(el_async_t));
+	g_thread_safe_cb_list[list_idx] = (thread_safe_cb_list_t *)EL_ALLOC(sizeof(thread_safe_cb_list_t));
+	if (g_thread_safe_cb_list[list_idx] == NULL) {
+		return EVENTLOOP_OUT_OF_MEMORY;
+	}
+
+	g_thread_safe_cb_list[list_idx]->async_handle = (el_async_t *)EL_ALLOC(sizeof(el_async_t));
```
sunghan-chang, line 159: When this gets failure, you should free the memory allocated at line 153 before return.

pcs1265: Fixed. Thank you!

framework/src/eventloop/eventloop_async.c
```diff
@@ -51,17 +51,41 @@ struct thread_safe_cb_s {
...
+#define ASYNC_HANDLE(idx)    eventloop_get_async_handle(idx)
+#define LIST_SEM(idx)        eventloop_get_list_sem(idx)
 
 /* Define for the hash operation */
 #define MAX_PID_MASK                       (CONFIG_MAX_TASKS - 1)
 #define THREAD_SAFE_CB_LIST_IDX_HASH(pid)  ((pid) & MAX_PID_MASK)
 
+static inline sq_queue_t *eventloop_get_cb_list(int idx)
+{
+	if (g_thread_safe_cb_list[idx]) {
```
jeongarmy, line 67: `idx` should be (0, CONFIG_MAX_TASKS)
I think we need to check the range of `idx` to avoid wrong access and svace issue.
Could you add a condition to check idx?

pcs1265: `THREAD_SAFE_CB_LIST_IDX_HASH` ensures that the value stays within the expected range,
and users obtain the index through it.

However, a double‑check is required to avoid svace issue, so I will add that check.

framework/src/eventloop/eventloop_async.c
```diff
@@ -144,10 +174,14 @@ static void eventloop_thread_safe_cb_list_node_deinit(int list_idx)
 		return;
 	}
 
-	ASYNC_HANDLE(list_idx)->async_cb = NULL;
-	EL_FREE(ASYNC_HANDLE(list_idx));
-	ASYNC_HANDLE(list_idx) = NULL;
-	sem_destroy(&LIST_SEM(list_idx));
+	if(g_thread_safe_cb_list[list_idx] != NULL) {
```
jeongarmy: Please add space

framework/src/eventloop/eventloop_async.c
jeongarmy: Oh, I missed that.
And caller functions already checks this range like below.
```
static int eventloop_thread_safe_cb_list_node_init(int list_idx, thread_safe_func_t *func_handle)
{
	if (list_idx < 0 || list_idx >= CONFIG_MAX_TASKS || func_handle == NULL) {
		eldbg("Invalid Parameter!\n");
		return EVENTLOOP_INVALID_PARAM;
	}
...
    sq_init(CB_LIST(list_idx));
```
https://github.com/Samsung/TizenRT/pull/6974/files#diff-381fde9827218370afee3665df535c252961922c1f3420e5d8adf49f44253c8eR148

I think if all caller functions already check this and input value is ensured as valid range, you don't need to add condition here.

framework/src/eventloop/eventloop_async.c
pcs1265: Resolved.

framework/src/eventloop/eventloop_async.c
```diff
@@ -126,14 +150,20 @@ static int eventloop_thread_safe_cb_list_node_init(int list_idx, thread_safe_fun
 		return EVENTLOOP_INVALID_PARAM;
 	}
 
-	ASYNC_HANDLE(list_idx) = (el_async_t *)EL_ALLOC(sizeof(el_async_t));
+	g_thread_safe_cb_list[list_idx] = (thread_safe_cb_list_t *)EL_ALLOC(sizeof(thread_safe_cb_list_t));
+	if (g_thread_safe_cb_list[list_idx] == NULL) {
+		return EVENTLOOP_OUT_OF_MEMORY;
```
sunghan-chang, line 156: Don't you need debug message in this case? Because allocation for async_handle has.

sunghan-chang (approved):

PR #6976, opened 2025-09-23
EOIR to be set in arm_cpuhotplug

anuj01-gupta:


anuj01-gupta:
will raise the commit again

PR #6977, opened 2025-09-23
os/arch/arm: CPU hotplug: manually clear SGI4 interrupt active

anuj01-gupta:
os/arch/arm: CPU hotplug: manually clear SGI4 interrupt active

As psci_cpu_off is a non return function it will keep the CPU in a WFI loop
this means that it will never exit back to arm_decodeirq and EOIR needs to be cleard 

This change is needed to be added before  psci_cpu_off
This change was also part of PR 6707 raised by realtek for secondary CPU off functionality.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
+++ b/os/arch/arm/src/armv7-a/arm_cpuhotplug.c
@@ -94,6 +94,13 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 	rtcb->task_state = TSTATE_TASK_ASSIGNED;
 	CURRENT_REGS = NULL;
 
+	/*
+	 manually clear SGI4 interrupt active
+	 As psci_cpu_off is a noreturn function it will trap the CPU in a WFI loop
+	 this means that it will never exit back to arm_decodeirq where EOIR will be set
+	*/
+	putreg32(irq, GIC_ICCEOIR);
+
 	/* MINOR: PSCI is a general interface for controlling power of cortex-A cores 
 	   It is one of the feature in arm-trusted-firmware (ie. Trustzone-A)
 	   Currently, the definitions are provided by chip specific arch layer


amandeep-samsung:
LGTM

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -94,6 +94,13 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 	rtcb->task_state = TSTATE_TASK_ASSIGNED;
 	CURRENT_REGS = NULL;
 
+	/*
+	 manually clear SGI4 interrupt active
+	 As psci_cpu_off is a noreturn function it will trap the CPU in a WFI loop
+	 this means that it will never exit back to arm_decodeirq where EOIR will be set
+	*/
+	putreg32(irq, GIC_ICCEOIR);
+
```
eunwoo-nam, line 103:
```c
		/* 
			clear GIC APR0 as when this gets set during restore by ATF boot flow,
			When this register is set, it cause GIC to not serve other requests as it thinks there is an
			active ISR being served currently.
			
			On powerup, this register is restored by ATF from previous saved state. 
			If hard reset, then state is re-initialized, otherwise it will save the state when psci_cpu_off
		*/
		putreg32(0x00000000, GIC_ICCNSAPR1);	/* clear SGI1 nonsecure APR0 */
		putreg32(0x00000000, GIC_ICCAPR1);	/* clear SGI1 secure APR0 */

```
in https://github.com/Samsung/TizenRT/pull/6707, GIC_ICCNSAPR1 and GIC_ICCAPR1 clearing is not required???

os/arch/arm/src/armv7-a/arm_cpuhotplug.c
```diff
@@ -94,6 +94,13 @@ int arm_hotplug_handler(int irq, void *context, void *arg)
 	rtcb->task_state = TSTATE_TASK_ASSIGNED;
 	CURRENT_REGS = NULL;
 
+	/*
+	 manually clear SGI4 interrupt active
+	 As psci_cpu_off is a noreturn function it will trap the CPU in a WFI loop
+	 this means that it will never exit back to arm_decodeirq where EOIR will be set
+	*/
+	putreg32(irq, GIC_ICCEOIR);
```
r-prabu, line 102: Better to have some safety check considerations  --> #ifdef CONFIG_DEBUG_ASSERTIONS
if (irq != GIC_IRQ_SGI4) {
    /* This should never happen as this is the SGI4 handler */
    panic("Unexpected IRQ %d in SGI4 handler\n", irq);
}
#endif

samsung-singh (approved):
manually cleared SGI4 interrupt active in os arch arm cpu hotplug module, this chage is good to merge.

eunwoo-nam:
This changes was included #7007, Let's close this pr.

PR #6979, opened 2025-09-24
os/pm: Add PMIOC_STOP to stop pm and show state in pm_procfs

seokhun-eom24:
### 1) Add PMIOC_STOP to stop pm.
Currently, pm driver only support start function with PMIOC_START.
So I added pm stop PMIOC_STOP to stop pm functionality.

### 2) Add pm stop example code
In current power example code when user input "power start" command,
"PM LONG TERM TEST END" log is printed immediately after "PM LONG TERM TEST START" log.
```
TASH>>power start
TASH>>######################### PM LONG TERM TEST START #########################
```

To fix this issue, add stop_pm_test function and call it when user input "power stop" command.
This function calls PMIOC_STOP to stop pm functionality.

### 3) Fix typo and convention in pm_start
Fix typo fucntion to function.
Fix function brace convention to enter after function name.

### 4) Add pm running state in pm_procfs
Modify pm_procfs to print whether pm is RUNNING or STOPPED.

Example
```
TASH>>cat state
PM STOPPED

* NORMAL
  SLEEP

TASH>>power start
TASH>>######################### PM LONG TERM TEST START #########################

TASH>>cat state
PM RUNNING

* NORMAL
  SLEEP

TASH>>power stop
TASH>>######################### PM LONG TERM TEST END #########################

TASH>>cat state
PM STOPPED

* NORMAL
  SLEEP
```


apps/examples/power/power_main.c
```diff
@@ -289,6 +310,13 @@ int power_main(int argc, char *argv[])
 
 		is_running = false;
 
+		pid = task_create("stop_pm_test", 100, 1024, stop_pm_test, argv + 1);
+		if (pid < 0) {
+			printf("Fail to create stop_pm_test task(errno %d)\n", get_errno());
+			is_running = true;
+			return -1;
+		}
```
eunwoo-nam: we don't need stop_pm_test task

it is requred only calling stop ioctl.
```c
	if(ioctl(fd, PMIOC_STOP, 0) < 0) {
		printf("Fail to pm start(errno %d)\n", get_errno());
		close(fd);
		return -1;
	}
```

eunwoo-nam: And how about make each testing function like below? (start / stop)

we can move log of start and stop to actually start and end entry point.

```c
static int pm_start(void) 
{
	if(ioctl(fd, PMIOC_START, 0) < 0) ~~
}

static int pm_stop(void)
{
	if(ioctl(fd, PMIOC_STOP, 0) < 0) ~~
}


....

	if (strncmp(argv[1], "start", 6) == 0) {
		...

		is_running = true;
		printf("######################### PM LONG TERM TEST START #########################\n");
		pm_stop();

	} else if (strncmp(argv[1], "stop", 5) == 0) {
		...

		pm_stop();
		printf("######################### PM LONG TERM TEST END #########################\n");
		is_running = true;
	}
```

os/pm/pm_initialize.c
```diff
@@ -99,10 +99,32 @@ const char *wakeup_src_name[PM_WAKEUP_SRC_COUNT] = {"UNKNOWN", "BLE", "WIFI", "U
...
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
+
+void pm_stop(void)
+{
+	g_pmglobals.is_running = false;
+}
```
eunwoo-nam: Looks good to me this stop fucntion.

os/include/tinyara/fs/ioctl.h
```diff
@@ -498,6 +498,7 @@
 #define PMIOC_METRICS            _PMIOC(0x0007)
 #define PMIOC_SUSPEND_COUNT      _PMIOC(0x0008)
 #define PMIOC_START              _PMIOC(0x0009)
+#define PMIOC_STOP               _PMIOC(0x0010)
```
pcs1265: `0x10` is not a next value of `0x09`.
Please change this value to `0x0A`

os/pm/pm_procfs.c
```diff
@@ -203,6 +203,8 @@ static void power_read_domains(void (*readprint)(const char *, ...))
 static void power_read_state(void (*readprint)(const char *, ...))
 {
 	enum pm_state_e pm_state;
+
+	readprint("PM %s\n\n", (g_pmglobals.is_running) ? "RUNNING" : "STOPPED");
```
Taejun-Kwon: procfs is not only for information.
sometimes application should parse data from procfs, so it shoul memcpy data to buffer.
Hence if you want reveal running state, then you should create additional directory, not here.

os/pm/pm_initialize.c
```diff
@@ -99,10 +99,32 @@ const char *wakeup_src_name[PM_WAKEUP_SRC_COUNT] = {"UNKNOWN", "BLE", "WIFI", "U
...
+ *   None.
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
+
+void pm_stop(void)
+{
+	g_pmglobals.is_running = false;
```
Taejun-Kwon: pm_ioctl doesn't use semaphore, so I think this is not 'thread-safe'

apps/examples/power/power_main.c
```diff
@@ -223,6 +223,27 @@ static int start_pm_test(int argc, char *argv[])
...
+	return 0;
+}
+
+static int stop_pm_test(int argc, char *argv[])
+{
+	int fd;
+
+	fd = open(PM_DRVPATH, O_WRONLY);
+	if (fd < 0) {
+		printf("Fail to open pm start(errno %d)", get_errno());
```
sunghan-chang: It's open pm driver, not open pm start.

apps/examples/power/power_main.c
```diff
@@ -223,6 +223,27 @@ static int start_pm_test(int argc, char *argv[])
...
+		return -1;
+	}
+
+	if(ioctl(fd, PMIOC_STOP, 0) < 0) {
+		printf("Fail to pm start(errno %d)\n", get_errno());
```
sunghan-chang: looks fail to stop pm, not start

apps/examples/power/power_main.c
```diff
@@ -223,6 +223,27 @@ static int start_pm_test(int argc, char *argv[])
...
+{
+	int fd;
+
+	fd = open(PM_DRVPATH, O_WRONLY);
+	if (fd < 0) {
+		printf("Fail to open pm start(errno %d)", get_errno());
+		return -1;
+	}
+
+	if(ioctl(fd, PMIOC_STOP, 0) < 0) {
```
sunghan-chang: space.

apps/examples/power/power_main.c
```diff
@@ -289,6 +310,13 @@ int power_main(int argc, char *argv[])
 
 		is_running = false;
 
+		pid = task_create("stop_pm_test", 100, 1024, stop_pm_test, argv + 1);
```
sunghan-chang: Could you let me know why you give `argv + 1`?
In help message, the stop command does not have any following argument.

And the task creation for start also give it, but is it right with `+1`? not `+2`?

apps/examples/power/power_main.c
```diff
@@ -289,6 +310,13 @@ int power_main(int argc, char *argv[])
...
+		if (pid < 0) {
+			printf("Fail to create stop_pm_test task(errno %d)\n", get_errno());
+			is_running = true;
```
sunghan-chang: If you set the variable to false after this creation, what happen?
I mean
 do you need to set it to false always and then revert it with failure case?

os/pm/pm_initialize.c
```diff
@@ -103,6 +103,27 @@ void pm_start(void) {
...
+ *
+ * Input parameters:
+ *   None.
+ *
+ * Returned value:
+ *    None.
+ *
+ ****************************************************************************/
+
+void pm_stop(void)
```
sunghan-chang: This file is `pm_initialize`. Let's add this with new file.
same for pm_start

os/include/tinyara/fs/ioctl.h
seokhun-eom24: Fixed. Thank you.

apps/examples/power/power_main.c
seokhun-eom24: Fixed. Thank you.
And I modified other open error message too.

apps/examples/power/power_main.c
seokhun-eom24: Fixed.

apps/examples/power/power_main.c
seokhun-eom24: Fixed this line and also in `start_pm_test`

apps/examples/power/power_main.c
seokhun-eom24: Yes, `stop_pm_test` doesn't need any argument.
So I fixed `argv + 1` to `NULL`.

`argv + 1` is the correct value to use because `argv` is a pointer to the first element of the command argument.

When the command line is something like:
```
power start -t
```
the `argv` array is initialized as:
```
argv[0] = "power"
argv[1] = "start"
argv[2] = "-t"
argv[3] = NULL
```
To pass only the part after the program name(i.e.,`{"start", "-t"}`), need to give the pointer that starts at `argv[1]`. So `argv + 1` is right.

apps/examples/power/power_main.c
seokhun-eom24: In `start_pm_test`, `is_running` should be set first and start task, 
but as you said in `stop_pm_test` reset is_running can handled after create task.
So I modified the flow to reset after create task.

apps/examples/power/power_main.c
sunghan-chang: @seokhun-eom24 You already check the argv[1] with start.
```
	if (strncmp(argv[1], "start", 6) == 0) {
		if (is_running) {
			printf("power test is already running\n");
			return 0;
		}
		is_running = true;
		pid = task_create("start_pm_test", 100, 1024, start_pm_test, argv + 1);
```
The task is created for start. Why do you need to give `start` argument to the task?

apps/examples/power/power_main.c
```diff
@@ -287,6 +308,12 @@ int power_main(int argc, char *argv[])
 			return 0;
 		}
 
+		pid = task_create("stop_pm_test", 100, 1024, stop_pm_test, NULL);
```
sunghan-chang: Is there any reason why we should do start and stop test in separate tasks?

apps/examples/power/power_main.c
```diff
@@ -223,6 +223,27 @@ static int start_pm_test(int argc, char *argv[])
...
+{
+	int fd;
+
+	fd = open(PM_DRVPATH, O_WRONLY);
+	if (fd < 0) {
+		printf("Fail to open pm driver(errno %d)", get_errno());
+		return -1;
+	}
+
+	if (ioctl(fd, PMIOC_STOP, 0) < 0) {
```
eunwoo-nam: This code can be called first than "start_pm_test" task termination

like
```
call resume(10)
######################### PM LONG TERM TEST END #########################
call suspend(10)
```

gidori98:
LGTM

apps/examples/power/power_main.c
seokhun-eom24: Thank you for suggest. 
I updated to print start and stop log inside `_pm_start` and `_pm_stop`.

os/pm/pm_procfs.c
seokhun-eom24: I updated this PR not to use procfs.
Thank you for review.

os/pm/pm_initialize.c
seokhun-eom24: I updated to protect `g_pmglobals.is_running` using semaphore.
Thank you.

apps/examples/power/power_main.c
seokhun-eom24: Right. We don't need to pass `argv + 1`, it's `start`.
I updated to pass `argv + 2`.
Thank you.

os/pm/pm_initialize.c
seokhun-eom24: I separated `pm_start`, `pm_stop`, and `pm_clock_initialize` from `pm_initialize.c` with new files.

apps/examples/power/power_main.c
seokhun-eom24: We don't need to separate tasks to stop.
I updated code.

apps/examples/power/power_main.c
seokhun-eom24: I updated to print log after `_pm_stop`.

os/pm/pm_getstatus.c
```diff
@@ -0,0 +1,59 @@
...
+int pm_getstatus(void)
+{
+	int status;
+
+	pm_lock();
+	status = g_pmglobals.is_running ? PM_STATUS_RUNNING : PM_STATUS_STOPPED;
+	pm_unlock();
+
+	return status;
+}
```
eunwoo-nam: we don't need to provide status using api.
how about add it to procfs?

seokhun-eom24: I added a new procfs `/proc/power/running`.  
The reason for creating `/power/running` separately from `/power/status` is based on Taejun-Kwon's comment.  
Procfs is not only for information but also allows applications to fetch data via memcpy, which is why the new procfs was added.

seokhun-eom24: Modified to use "/proc/power/state" together as we discussed.

PR #6980, opened 2025-09-24
os/arch/arm/src/amebasmart: Fix I2S abort during PM suspend

ziyik:
-Check I2S stuct pointer during PM suspend and resume

### DIFF
--- a/os/arch/arm/src/amebasmart/amebasmart_i2s.c
+++ b/os/arch/arm/src/amebasmart/amebasmart_i2s.c
@@ -1904,6 +1904,10 @@ errout_with_alloc:
 static void amebasmart_i2s_suspend(uint16_t port)
 {
 	struct amebasmart_i2s_s *priv = g_i2sdevice[port];
+	if (!priv) {
+		i2sinfo("I2S %d is not init\n", port);
+		return;
+	}
 
 	i2s_disable(priv->i2s_object, 1);
 #if defined(I2S_HAVE_RX) && (0 < I2S_HAVE_RX)
@@ -1922,12 +1926,14 @@ static void amebasmart_i2s_suspend(uint16_t port)
 static void amebasmart_i2s_resume(uint16_t port)
 {
 	struct amebasmart_i2s_config_s *hw_config_s;
-	struct amebasmart_i2s_s *priv;
+	struct amebasmart_i2s_s *priv = g_i2sdevice[port];	/* The I2S structure should exist after system wakeup */
 	int ret;
-	int count;
-	/* The I2S structure should exist after system wakeup */
-	priv = g_i2sdevice[port];
-	DEBUGASSERT(priv);
+
+	if (!priv) {
+		i2sinfo("I2S %d is not init\n", port);
+		return;
+	}
+
 	DEBUGASSERT(priv->i2s_object);
 	memset(priv->i2s_object, 0, sizeof(i2s_t));
 	/* Config values initialization */
@@ -1979,9 +1985,7 @@ static uint32_t rtk_i2s_resume(uint32_t expected_idle_time, void *param)
 
 	return 1;
 }
-#endif
 
-#ifdef CONFIG_PM
 void i2s_pminitialize(void)
 {
 	bsp_pm_domain_register("I2S", BSP_I2S_DRV);


os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1922,12 +1926,14 @@ static void amebasmart_i2s_suspend(uint16_t port)
...
 {
 	struct amebasmart_i2s_config_s *hw_config_s;
-	struct amebasmart_i2s_s *priv;
-	int ret;
-	int count;
-	/* The I2S structure should exist after system wakeup */
-	priv = g_i2sdevice[port];
-	DEBUGASSERT(priv);
+	struct amebasmart_i2s_s *priv = g_i2sdevice[port];	/* The I2S structure should exist after system wakeup */
+	int ret count;
```
eunwoo-nam: the count is not used. you can remove count

sunghan-chang: looks `,` is missing.

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -161,11 +161,11 @@ void board_i2s_initialize(void)
 #ifdef CONFIG_AMEBASMART_I2S
 #ifdef CONFIG_AUDIO_I2SCHAR
 	i2schar_devinit();
-#endif
 #ifdef CONFIG_PM
 	i2s_pminitialize();
 #endif
 #endif
+#endif
```
sunghan-chang, line 169: I don't get why this is changed. Why is i2s pm under audio i2s?

sunghan-chang:
@ziyik build break on flat_dev_ddr

os/board/rtl8730e/src/rtl8730e_boot.c
seokhun-eom24: i2s can also be initialized in `rtl8730e_syu645b_initialize`.
So I think the previous conditional check that `calling i2s_pminitialize` when `CONFIG_AMEBASMART_I2S` is defined is correct.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1922,12 +1926,14 @@ static void amebasmart_i2s_suspend(uint16_t port)
...
 {
 	struct amebasmart_i2s_config_s *hw_config_s;
-	struct amebasmart_i2s_s *priv;
-	int ret;
-	int count;
-	/* The I2S structure should exist after system wakeup */
-	priv = g_i2sdevice[port];
-	DEBUGASSERT(priv);
+	struct amebasmart_i2s_s *priv = g_i2sdevice[port];	/* The I2S structure should exist after system wakeup */
+	int ret, count;
```
ewoodev, line 1930: The count is not used.
How about remove it?

os/board/rtl8730e/src/rtl8730e_boot.c
ziyik: Since amebasmart_i2s_suspend() and amebasmart_i2s_resume() are based on port number, but currently when pmu_register_sleep_callback() is called, we do not pass in any param. Should we restructure i2s_pminitialize() place it in amebasmart_i2s_initialize() and called with port number pass in? So PM suspend and resume will be trigger based on i2s port which has been initialized?

seokhun-eom24: I think it would be a better design for to call `i2s_pminitialize` in `amebasmart_i2s_initialize`.
However, in that case we might need to invoke `pmu_register_sleep_callback` multiple times depending on the port, what do you think about it?

ziyik: Yes, that's what I planned to do. The callback will be triggered multiple times as well.
So pmu_register_sleep_callback() will be called during init, and pmu_unregister_sleep_callback() will be called during deinit. Is that okay?

seokhun-eom24: Ok, that looks good.

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1970,21 +1976,14 @@ static uint32_t rtk_i2s_resume(uint32_t expected_idle_time, void *param)
...
 }
-#endif
 
-#ifdef CONFIG_PM
-void i2s_pminitialize(void)
+static void i2s_pminitialize(uint16_t port)
 {
 	bsp_pm_domain_register("I2S", BSP_I2S_DRV);
-	pmu_register_sleep_callback(PMU_I2S_DEVICE, (PSM_HOOK_FUN)rtk_i2s_suspend, NULL, (PSM_HOOK_FUN)rtk_i2s_resume, NULL);
+	pmu_register_sleep_callback(PMU_I2S_DEVICE, (PSM_HOOK_FUN)rtk_i2s_suspend, (void *)port, (PSM_HOOK_FUN)rtk_i2s_resume, (void *)port);
```
eunwoo-nam:
``` c
void pmu_register_sleep_callback(u32 nDeviceId, PSM_HOOK_FUN sleep_hook_fun, void *sleep_param_ptr,
								 PSM_HOOK_FUN wakeup_hook_fun, void *wakeup_param_ptr)
{
	PSM_DD_HOOK_INFO *pPsmDdHookInfo = NULL;

	assert_param(nDeviceId < PMU_MAX);
	assert_param((sleep_hook_fun != NULL) || (wakeup_hook_fun != NULL));

	pPsmDdHookInfo = &gPsmDdHookInfo[nDeviceId];
	pPsmDdHookInfo->nDeviceId			= nDeviceId;
	pPsmDdHookInfo->sleep_hook_fun		= sleep_hook_fun;
	pPsmDdHookInfo->sleep_param_ptr		= sleep_param_ptr;
	pPsmDdHookInfo->wakeup_hook_fun		= wakeup_hook_fun;
	pPsmDdHookInfo->wakeup_param_ptr	= wakeup_param_ptr;
}
```

the pmu_register_sleep_callback has static array map for registering device.
so, if you put in port I2S2 and I2S3 using param, the port is overwriten.

seokhun-eom24:
I think the current changes have successfully achieved the purpose of the PR.

However, I’m wondering that in the current implementation all I2S ports are managed by the PM using the same cycle.
What do you think about separating the domain to allow each port to be managed individually by PM?

ziyik:
> I think the current changes have successfully achieved the purpose of the PR.
> 
> However, I’m wondering that in the current implementation all I2S ports are managed by the PM using the same cycle. What do you think about separating the domain to allow each port to be managed individually by PM?

By doing so, it will have 2 I2S domains, is it? Currently, only one domain, BSP_I2S_DRV, and driver will handle all I2S ports.
Same for other peripherals, like UART, I2C, SPI and etc. All these peripherals will have multiple channels or interfaces.
This will result in multiple domains (PM) for one peripheral.
In my opinion, the current structure is the most suitable one, with PM control domains, driver register domain, and controls its channel and interfaces. Because only the driver has full control or access to them. (as long as config is set correctly)

gidori98:
LGTM

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1888,6 +1889,10 @@ struct i2s_dev_s *amebasmart_i2s_initialize(uint16_t port)
 	//priv->i2s_num = priv->i2s_object->i2s_idx;
 	g_i2sdevice[port] = priv;
 
+#ifdef CONFIG_PM
+	i2s_pminitialize();
+#endif
```
eunwoo-nam: How about revert it?
it's duplicated calling.

ziyik: This is the only place to register the PM domain for i2s.
If this is removed, no other place will register the domain.
I have checked, and it's okay to register 2 times. It will overwrite it.
This will ensure the PM domain is registered when i2s is initialized (without depending/checking which i2s port is initialized)

sunghan-chang: @ziyik He asked revert, not remove. Previously, you called the i2s pm init at outside of the i2s init. It called the pm init once, but new change causes twice. We want to prevent duplicated calling. It needs all pm domain string comparison and unnecessary duplicated operation.

ziyik: In this case, I will need to revert rtl8730e_boot.c as well.
Otherwise, the PM domain will not be registered.
Is that okay?

eunwoo-nam: We believe that if we do not use I2S, it is appropriate not to register I2S-related PM as well.

However, in the current structure, I2S is not differentiated by port-specific PM, so we are currently registering I2S in PM and checking the initialization status of each port based on the presence of priv.

In this structure, it seems clearer for i2s_pminitialize() to be called from board_i2s_initialize(), which is the initialization process for I2S itself, rather than for each port.

Additionally, while the current code only allows overwriting, which does not cause issues with duplicate calls, it would be better to revert rather than allow duplicate calls, as understanding this history may be necessary for future modifications.

ziyik: Okay, got it.
I will revert the change, and since the PM sleep/resume has already added the check of the port initialization state (presence of priv), so it's fine.

ewoodev (dismissed):

os/arch/arm/src/amebasmart/amebasmart_i2s.c
```diff
@@ -1922,12 +1926,14 @@ static void amebasmart_i2s_suspend(uint16_t port)
...
-	struct amebasmart_i2s_s *priv;
+	struct amebasmart_i2s_s *priv = g_i2sdevice[port];	/* The I2S structure should exist after system wakeup */
 	int ret;
-	int count;
-	/* The I2S structure should exist after system wakeup */
-	priv = g_i2sdevice[port];
-	DEBUGASSERT(priv);
+
+	if (!priv) {
+		i2serr("I2S %d is not init\n", port);
```
eunwoo-nam: Could you please check, it is correct err log?
is it printed every wakeup??

If it is okay return here, don't print it using err level.
(the i2serr is enabled)

ewoodev (approved):

os/arch/arm/src/amebasmart/amebasmart_i2s.c
ziyik: You are right, it should print with i2serr here.
It's okay to return here. If it's not initialized, it shouldn't process suspend and resume.

PR #6985, opened 2025-09-29
Use system mutex APIs instead of `rtw` APIs

Jw00:
This causes conflicts during code rebase.


j11591 (approved):
rtw_mutex is for chipset vendor FW.
So changing sys_mutex is much better.

PR #6986, opened 2025-09-29
os/board/rtl8721csm: fix BLE indicate semaphore release

yeetee179:
- moved BLE indication semaphore release before calling app layer callback p_func
- to avoid semaphore not released due to action is done in app layer callback p_func


giwon-nam (approved):
Confirm the semaphore release after disconnection

PR #6987, opened 2025-09-29
os/pm: Rearchitect pm_procfs with a declarative template tree

ewoodev:
This commit completely refactors the pm_procfs.c implementation to improve readability, maintainability, and extensibility.

The previous implementation relied on a complex, procedural function (power_find_dirref) to parse relative paths. This involved manual string manipulation, nested if-else statements, and state management using level indicators, making the code difficult to understand and prone to errors when extending.

This has been replaced with a data-driven, declarative design centered around a static "template tree" (power_path_template_s).

1. Declarative Template Tree The entire /proc/power filesystem layout, including full path patterns, file/directory types, and associated handlers, is now defined in a single, static data structure. This makes the overall structure self-documenting and easy to comprehend.

2. Application of Design Patterns
- Template Method/Strategy Pattern: The generic VFS operations (power_read, power_readdir) now act as dispatchers, invoking specific function pointers (.read, .readdir) defined within each template. This decouples the core logic from the specific actions for each path.

- Composite Pattern: The template structure forms a classic tree, allowing directories (composites) and files (leaves) to be treated uniformly during path resolution.

3. Improved State Management
- Introduced power_path_priv_s to cleanly encapsulate the context of a resolved path (matched template and domain pointer).
- Optimized directory reading by tracking the last read position (domain_position), preventing redundant list traversals in readdir.

This rearchitecture significantly reduces code complexity and enhances maintainability. Adding or modifying procfs entries now primarily involves updating the static template data rather than altering complex control flow logic.

### DIFF
--- a/os/pm/pm_procfs.c
+++ b/os/pm/pm_procfs.c
@@ -90,21 +90,39 @@
 /****************************************************************************
  * Private Types
  ****************************************************************************/
+
+// Forward declaration
+struct power_path_template_s;
+struct power_file_s;
+
+/* Template structure for a node in the procfs path */
+struct power_path_template_s {
+	const char *name;
+	uint8_t dtype;
+	void (*read)(FAR struct power_file_s *priv, void (*readprint)(const char *, ...));
+	int (*readdir)(struct fs_dirent_s *dir);
+	const struct power_path_template_s *children;
+};
+
 /* This enumeration identifies all of the task/thread nodes that can be
  * accessed via the procfs file system.
  */
 
+struct power_path_priv_s {
+	const struct power_path_template_s *template; /* Matched template node */
+	struct pm_domain_s *domain_ptr;               /* Pointer to the domain structure if applicable */
+};
+
 struct power_dir_s {
-	struct procfs_dir_priv_s base; /* Base directory private data */
-	uint8_t direntry;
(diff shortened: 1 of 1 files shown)


os/pm/pm_procfs.c
```diff
@@ -177,137 +245,185 @@ const struct procfs_operations power_procfsoperations = {
...
+			if (candidate->children) {
+
+				const char *slash_pos = strchr(relpath, '/');
+				if (slash_pos) {
+					child_matched_candidate = power_find_best_match(slash_pos + 1, candidate->children, path_priv, proc_dir);
+				}
+
+				if (child_matched_candidate) {
+					candidate = child_matched_candidate;
 				}
```
seokhun-eom24: I think here's an issue where it is possible to access non-existent domain paths, such as /proc/power/domains/UARTXX when `child_matched_candidate` is NULL.

eunwoo-nam: Thank you for review 😀👍👍, Fixed.
```log
TASH>>ls proc/power/domains
/proc/power/domains:
 info
 UART/
 SPI/
 I2C/
 I2S/
 MIPI/
 LCD/
 IST415/
 NDP120/
 TASH/
TASH>>ls proc/power/domains/UARTXX
ls : stat failed
TASH>>
```

PR #6989, opened 2025-09-30
os/board/rtl8730e: fix idx_wlan used when enable bridge mode

ZivLow:
- Fix SoftAP not working when bridge mode enabled


hk-gwak (approved):
Confirmed that DNS works well after this commit.

PR #6990, opened 2025-09-30
os/drivers/lcd/mipi_lcd.c: Add runtime BMP to RGB565 conversion

abhinav-s235:
- This commit introduces a new utility to convert BMP image files to the RGB565 color format and display them on the LCD at runtime.
- Supports reading 16-bit files, including handling of `BI_RGB` and `BI_BITFIELDS` compression for 16-bit variants.

PR #6993, opened 2025-10-02
drivers/lcd: Add splash image support for lcd drivers

seokhun-eom24:
This update modifies `lcd_init_put_image` to handle splash image rendering during boot.

- Normal boot: Renders splash_normal.bmp.
- Silent boot: Renders splash_silent.bmp.
- No splash image: Turns LCD off.

The existing LCD logo rendering code is replaced with splash image handling. 
A new sample splash_normal.bmp is added for 800x480 LCD.

Implementation builds on PR #6990 (`lcd_render_bmp` function).


sunghan-chang:
@anjana348 @namanjain7 @abhinav-s235 Could you review this PR to change initial screen output?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -563,53 +559,44 @@ static int lcd_setcontrast(FAR struct lcd_dev_s *dev, unsigned int contrast)
...
+	/* Check if BMP file exists before rendering */
+	FILE *test_file = fopen(bmp_file_path, "rb");
+	if (!test_file) {
+		lcddbg("BMP file not found at %s. LCD OFF\n", bmp_file_path);
+		return ERROR;
+	}
+	fclose(test_file);
+
+	ASSERT(dev->setpower);
+	ret = dev->setpower(dev, CONFIG_LCD_MAXPOWER);
```
anjana348: Earlier we were keeping LCD off in silent reboot case, now We are powering on LCD in both cases and setting different bmp image for silent reboot case.
 And if bmp file is not present, driver registration will not be done 
has the requirement changed like above?

seokhun-eom24: We are powering on LCD in both cases and setting different bmp image for silent reboot case 
-> Right.

And if bmp file is not present, driver registration will not be done 
-> Only LCD OFF. l'll fix it in this commit.

Thank you for Checking.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -563,53 +557,52 @@ static int lcd_setcontrast(FAR struct lcd_dev_s *dev, unsigned int contrast)
...
-		lcddbg("ERROR: LCD logo data memory allocation failed\n");
+	is_silent_mode = silent_reboot_is_silent_mode();
+	if (!is_silent_mode) {
+		snprintf(bmp_file_path, sizeof(bmp_file_path), "/res/kernel/graphics/%dx%d/splash_normal.bmp", CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+	} else {
+		snprintf(bmp_file_path, sizeof(bmp_file_path), "/res/kernel/graphics/%dx%d/splash_silent.bmp", CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+	}
+
+	/* Check if BMP file exists before rendering */
+	FILE *test_file = fopen(bmp_file_path, "rb");
```
anjana348: 1. There is double checking of file which is not needed , here and also in lcd_render_bmp(). Can be removed from one place.
2. You have added an extra file "app1" in commit. Is it needed?

seokhun-eom24: > 1. There is double checking of file which is not needed , here and also in lcd_render_bmp(). Can be removed from one place.
> 2. You have added an extra file "app1" in commit. Is it needed?

1. I know this is double check, but we need to check file here for not turn on the LCD when file is not exist.
If this check is removed, LCD on -> lcd_render_bmp return error for no bmp file -> LCD off. I think this can make backlight blinking.
Do you have any good idea for about this?

2. Of course not. I'll remove it. Thank you.

seokhun-eom24: Additionally, I also think we should avoid double checking.
However, due to the current LCD driver structure, the `setpower` function performs both LCD initialization and PWM ON. So the backlight must be turned on to draw the screen.
If there is no BMP file, the backlight should not turn on, so I think this method is currently the best approach.
In the future, by improving the LCD driver to separate PWM and LCD control, it might be possible to control them separately.
What are your thoughts on the current best approach and the improvements needed in the future?
And please let me know if there's anything incorrect in what I said.

anjana348: We can seperate setpower into small functions 
power_off  
Power_on
backlight(power)

and use them in lcd_init_put_image() (not call setpower ) like 

power on ()
ret = lcd_render_bmp(dev, bmp_file_path);	// Displaying BMP image on LCD
	if (ret != OK) {
		power_off();
                return;
	} 
backlight(CONFIG_LCD_MAXPOWER)

Then we don't need to do double checking.

seokhun-eom24: Thank you for your advice.
I add new small functions, `lcd_power_off` and `lcd_power_on` for reuse.
And also removed double checking.

anjana348 (approved):
LGTM

sunghan-chang (approved):

PR #6995, opened 2025-10-13
apps/examples/rtc: Fix rtc_main build error

seokhun-eom24:
When `CONFIG_RTC_ALARM` is not defined, below error occurs during build.
```
arm-none-eabi-ld: ../build/output/libraries/libapps.a(rtc_main.o): in function `rtc_main':
/root/tizenrt/apps/examples/rtc/rtc_main.c:436: undefined reference to `show_usage'
```

This commit replace `show_usage` function out of `#ifdef CONFIG_RTC_ALARM` block.

### DIFF
--- a/apps/examples/rtc/rtc_main.c
+++ b/apps/examples/rtc/rtc_main.c
@@ -169,6 +169,28 @@ static void *rtc_test(void *arg)
 	return NULL;
 }
 
+/****************************************************************************
+ * Name: show_usage
+ ****************************************************************************/
+
+static void show_usage(void)
+{
+	printf("\nUsage:  rtc\n");
+	printf("\tor: rtc stop\n");
+	printf("\t\tStart, or Stop rtc test\n");
+#ifdef CONFIG_RTC_ALARM
+	printf("\tor rtc alarm [-r seconds] [-a hh:mm:ss] [-c]\n");
+	printf("\t\tset, or cancel rtc alarm\n");
+	printf("Where:\n");
+	printf("\t-a hh:mm:ss \n");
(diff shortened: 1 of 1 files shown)


apps/examples/rtc/rtc_main.c
eunwoo-nam: the build error could be related to argument.

seokhun-eom24: I removed the passing argument and rebuilt the project, but the same error still occurs.
Although this may not be the direct cause, it could be an issue, so I’ll fix this too.

apps/examples/rtc/rtc_main.c
```diff
@@ -407,6 +387,29 @@ int rtc_alarm_main(int argc, char *argv[])
 }
 #endif
 
+/****************************************************************************
+ * Name: show_usage
+ ****************************************************************************/
+
+static void show_usage()
```
sunghan-chang: (void)

apps/examples/rtc/rtc_main.c
eunwoo-nam, line 403: you need to move up the show_usage() function than here.

the calling show_usage() is first here.

apps/examples/rtc/rtc_main.c
seokhun-eom24: Resolved.

apps/examples/rtc/rtc_main.c
seokhun-eom24: Resolved.

sunghan-chang (approved):

PR #7000, opened 2025-10-16
os/drivers/lcd: Fix lock and config for splash image

seokhun-eom24:
Modify `lcd_init_put_image` to return whether lcd is on or off.
Only lcd is on case, call `pm_suspend` and `silent_reboot_lock`.

Add new config `CONFIG_LCD_SPLASH_IMAGE_PATH` for splash image path.
Update relative code and defconfig.


eunwoo-nam (approved):

os/drivers/lcd/mipi_lcd.c
```diff
@@ -777,10 +777,14 @@ static int lcd_render_bmp(FAR struct lcd_dev_s *dev, const char *bmp_filename)
...
  *   If there's no BMP file, turn off the LCD.
+ * 
+ * Returns:
+ *   OK if image rendering succeeded and LCD turned on,
+ *   ERROR if image rendering failed and LCD turned off
  *
  ****************************************************************************/
 
-FAR void lcd_init_put_image(FAR struct lcd_dev_s *dev)
+FAR int lcd_init_put_image(FAR struct lcd_dev_s *dev)
```
Taejun-Kwon, line 787: To make this function more generically, below is much better.
```
FAR int lcd_init_put_image(FAR struct lcd_dev_s *dev, void *buffer, size_t len)
```
OR
```
static int lcd_render_bmp(FAR struct lcd_dev_s *dev, void *buffer, size_t len)
```
Then someday application or some upper layer can use this through the ioctl...

seokhun-eom24: I'll implement that with new PR.
I think `lcd_init_put_image` handles how and what to render in booting time, and `lcd_render_bmp` can be called inside `lcd_init_put_image`.

Taejun-Kwon: sure I meaning, one of them need to be changed.

sunghan-chang (approved):

allen-kim-sec (approved):
looks good

PR #7010, opened 2025-10-21
os/drivers/lcd: Prevent multiple LCD power on/off operations

seokhun-eom24:
Add condition check to prevent multiple LCD power on/off operations.

### DIFF
--- a/os/drivers/lcd/mipi_lcd.c
+++ b/os/drivers/lcd/mipi_lcd.c
@@ -496,34 +496,58 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
  *   Power off the LCD.
  * 	 Switch to command mode and display off, power off the LCD.
  *
+ * Returns:
+ * 	 OK if power on sequence succeeded,
+ * 	 ERROR if power on sequence failed.
+ *
  ****************************************************************************/
 
-static void lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
+static int lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
 {
+	if (priv->lcdonoff == LCD_OFF) {
+		lcddbg("LCD already powered off\n");
+		return OK;
+	}
+
 	lcddbg("Switch to CMD mode\n");
 	priv->config->mipi_mode_switch(CMD_MODE);
-	priv->lcdonoff = LCD_OFF;
+
 	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
-	send_cmd(priv, display_off_cmd);
-	/* The power off must operate only when LCD is ON */
+	if (send_cmd(priv, display_off_cmd) != OK) {
+		lcddbg("ERROR: LCD power off failed\n");
+		return ERROR;
+	}
+
+	priv->lcdonoff = LCD_OFF;
 	priv->config->power_off();
+
+	return OK;
 }
 
 /****************************************************************************
  * Name:  lcd_power_on
  *
  * Description:
- *   Power on the LCD and send init command. 
- * 
+ *   Power on the LCD and send init command.
+ *
  * Assumption:
  *   This function is called when LCD is OFF.
  * 	 So we can safely assume that LCD is in command mode.
+ *
+ * Returns:
+ * 	 OK if power on sequence succeeded,
+ * 	 ERROR if power on sequence failed.
  *
  ****************************************************************************/
 
 static int lcd_power_on(FAR struct mipi_lcd_dev_s *priv)
 {
 	int retries = CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT;
+
+	if (priv->lcdonoff == LCD_ON) {
+		lcddbg("LCD already powered on\n");
+		return OK;
+	}
 	
 	lcddbg("Powering up the LCD\n");
 	priv->config->power_on();
@@ -575,7 +599,10 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 	if (power == 0) {
 		lcddbg("Powering down the LCD\n");
 		priv->config->backlight(power);
-		lcd_power_off(priv);
+		if (lcd_power_off(priv) != OK) {
+			sem_post(&priv->sem);
+			return ERROR;
+		}
 	} else {
 		/* The power on must operate only when LCD is OFF */
 		if (priv->power == 0) {
@@ -819,7 +846,10 @@ FAR int lcd_init_put_image(FAR struct lcd_dev_s *dev)
 
 	ret = lcd_render_bmp(dev, bmp_file_path);
 	if (ret != OK) {
-		lcd_power_off(priv);
+		if (lcd_power_off(priv) != OK) {
+			sem_post(&priv->sem);
+			return ERROR;
+		}
 		sem_post(&priv->sem);
 		return ret;
 	}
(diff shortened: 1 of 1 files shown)


pcs1265 (approved):
LGTM

os/drivers/lcd/mipi_lcd.c
```diff
@@ -524,6 +529,11 @@ static void lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
 static int lcd_power_on(FAR struct mipi_lcd_dev_s *priv)
```
sunghan-chang, line 552: Regardless of this PR, why do off and on APIs have different return type?

sunghan-chang, line 552: 1. Regardless of this PR, why do off and on APIs have different return type?
2. Don't you need to verify the priv for null?

seokhun-eom24: 1. Change the return type of both `lcd_power_off` and `lcd_power_on` to `int`, as power operations can fail when `send_cmd` or `check_lcd_vendor_send_init_cmd` fail.

2. Verify `priv` is not null by adding `DEBUGASSERT` to all functions to validate non-null arguments in the new commit.

abhinav-s235:
LGTM

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,57 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
+ *
  ****************************************************************************/
 
-static void lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
+static int lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
 {
+	DEBUGASSERT(priv);
+	if (priv->lcdonoff == LCD_OFF) {
+		lcddbg("ERROR: LCD already powered off\n");
+		return;
```
eunwoo-nam: it's void return type

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,57 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
+
 	lcddbg("Switch to CMD mode\n");
 	priv->config->mipi_mode_switch(CMD_MODE);
-	priv->lcdonoff = LCD_OFF;
+
 	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
-	send_cmd(priv, display_off_cmd);
-	/* The power off must operate only when LCD is ON */
+	if (send_cmd(priv, display_off_cmd) != OK) {
+		lcddbg("ERROR: LCD power off failed\n");
```
anjana348: Here there should be return

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Resolved.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Resolved.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,60 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
+	}
+
 	lcddbg("Switch to CMD mode\n");
 	priv->config->mipi_mode_switch(CMD_MODE);
-	priv->lcdonoff = LCD_OFF;
+
 	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
-	send_cmd(priv, display_off_cmd);
-	/* The power off must operate only when LCD is ON */
+	if (send_cmd(priv, display_off_cmd) != OK) {
```
Taejun-Kwon:
```
int ret = send_cmd(priv, display_off_cmd);
if (ret != OK) {
	lcddbg("ERROR: LCD power off failed ret : %d\n", ret);
}
```

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,60 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
+		lcddbg("ERROR: LCD power off failed\n");
+		return ERROR;
+	}
+
+	priv->lcdonoff = LCD_OFF;
```
Taejun-Kwon: very minor, but usually change operation and then change state is more stable

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,60 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
 
 static int lcd_power_on(FAR struct mipi_lcd_dev_s *priv)
 {
+	DEBUGASSERT(priv);
 	int retries = CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT;
+
+	if (priv->lcdonoff == LCD_ON) {
+		lcddbg("ERROR: LCD already powered on\n");
+		return ERROR;
+	}
```
Taejun-Kwon, line 559: after this, priv->lcdonoff need to be changed.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -496,34 +505,60 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
 	lcddbg("Switch to CMD mode\n");
 	priv->config->mipi_mode_switch(CMD_MODE);
-	priv->lcdonoff = LCD_OFF;
+
 	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
-	send_cmd(priv, display_off_cmd);
-	/* The power off must operate only when LCD is ON */
+	if (send_cmd(priv, display_off_cmd) != OK) {
+		lcddbg("ERROR: LCD power off failed\n");
+		return ERROR;
```
Taejun-Kwon: This will finally calls bsp function, and perhaps return value is one of value in errno.h
so you should return one of errno, not ERROR

os/drivers/lcd/mipi_lcd.c
```diff
@@ -575,7 +611,10 @@ static int lcd_setpower(FAR struct lcd_dev_s *dev, int power)
 	if (power == 0) {
 		lcddbg("Powering down the LCD\n");
 		priv->config->backlight(power);
-		lcd_power_off(priv);
+		if (lcd_power_off(priv) != OK) {
```
Taejun-Kwon: same as above comments. debug message include result & type of error

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: For now, we're setting `priv->lcdonoff = LCD_ON` when switching mipi video mode and render buffer data timing.
But this distributed state management has a risk, so I refactored it in https://github.com/Samsung/TizenRT/pull/7018.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: It's a very minor issue, but I think it would be better to fix it.
Thank you.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Thanks for pointing that out.
I modified to log so that errno can be logged.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: I modified to return errno from send_cmd.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Fixed it.

sunghan-chang (approved):

PR #7018, opened 2025-10-29
os/drivers/lcd: Refactor lcd driver less dependent to mipi

seokhun-eom24:
### 1. os/drivers/lcd: Add centralized state management to LCD driver
Add new mipi_state to replace lcdonoff field to make centralized lcd state management.
Add new function to handle this state transitions safely.
Implement safe state transition logic for power on/off sequences.

### 2. os/drivers/lcd: Add simple image library support for BMP image load
Add new image_lib.c module to handle 16-bit BMP image decoding and conversion to RGB565 format for LCD display.
Now we can load image to buffer with this common image library function.
Update Make.defs to include the new source file in LCD driver build.

### 3. os/drivers/lcd: Add config for lcd splash image
Add CONFIG_LCD_SPLASH_IMAGE to enable splash image feature.
Only when this config is enabled, image_lib.c will be compiled.
And add config for image format and silent boot support.
Modifies the code to include splash image functionality conditionally based on Kconfig settings

### 4. build/configs/rtl8730e: Update lcd config for loadable ext ddr mode
Enable CONFIG_LCD_SPLASH_IMAGE and related BMP/Silent-boot options in RTL8730E configurations

### 5. os/drivers/lcd: Add splash image ops for lcd_dev_s
Add splash image ops for lcd_dev_s struct.
This splash image feature is general for all lcd driver.
Set existing splash image load function to this ops.

### 6. os/drivers/lcd: Rename vendor ID cmd retry config to vendor init cmd retry
Rename CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT to CONFIG_LCD_SEND_VENDOR_INIT_CMD_RETRY_COUNT in Kconfig and source files.
This change makes more clear that this configuration is used to retry sending vendor specific initialization commands.

### 7. build/configs/rtl8730e: Update lcd config for init cmd retry
Update defconfig to use CONFIG_LCD_SEND_VENDOR_INIT_CMD_RETRY_COUNT instead of CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT.

### 8. os/drivers/lcd: Replace common lcd function to lcd_dev.c
Replace lcd_rotate_buffer, lcd_render_bmp, lcd_load_splash from mipi_lcd.c to lcd_dev.c
In lcddev_register, handle splash image rendering feature as a part of common lcd driver.
Add new function lcd_putarea_with_rotation to apply software rotation to buffer and putarea to LCD.

In the previous, image_load_bmp_file function rotate buffer and save into buffer, because this image was rendered without sw rotation.
Update lcddev_register to handle splash image rendering as a common LCD feature
Remove manual rotation from image_load_bmp_file by using putarea with software rotation

### 9. os/drivers/lcd: Remove CONFIG_PM conditionals from lcd_dev
Remove `#ifdef CONFIG_PM` conditions from struct lcd_s and PM function calls.
This can make struct lcd_s 4 bytes bigger when `CONFIG_PM` is not defined, but this change can remove lcd driver's PM dependency.
Fix `LCDDEVIO_SETPOWER` logic to always include power state tracking and `silent_reboot` work regardless of `CONFIG_PM`.


os/drivers/lcd/mipi_lcd.c
```diff
@@ -655,62 +661,14 @@ static int lcd_setcontrast(FAR struct lcd_dev_s *dev, unsigned int contrast)
 	return OK;
 }
 
+#ifdef CONFIG_LCD_SPLASH_IMAGE
 static int lcd_render_bmp(FAR struct lcd_dev_s *dev, const char *bmp_filename)
 {
 	int xres = LCD_XRES;
 	int yres = LCD_YRES;
 	uint8_t *fullscreen_buffer = NULL;
 	FAR struct mipi_lcd_dev_s *priv = (FAR struct mipi_lcd_dev_s *)dev;
```
eunwoo-nam: It's not mipi specific function. 
SPI LCDc might need this function.

Let's move it to lcd_dev.c

os/drivers/lcd/mipi_lcd.c
```diff
@@ -817,19 +724,26 @@ FAR int lcd_init_put_image(FAR struct lcd_dev_s *dev)
...
+	}
 	sem_post(&priv->sem);
-	priv->config->mipi_mode_switch(VIDEO_MODE);
-	priv->lcdonoff = LCD_ON;
 
 	return OK;
+#else
+	return ERROR;
+#endif
 }
```
eunwoo-nam: It's also not  mipi-dsi sepecific functionallity

If need to make switching mipi mode and controilling backlight, you can add api enable/disable/backlight to ops

os/drivers/lcd/mipi_lcd.c
```diff
@@ -316,6 +260,69 @@ int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table)
 	}
 }
 
+static int set_mipi_state(struct mipi_lcd_dev_s *priv, mipi_state_t new_state)
+{
+	int retries = CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT;
+	mipi_state_t old_state = priv->mipi_state;
+	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
```
eunwoo-nam: You can make definition about it

os/drivers/lcd/lcd_dev.c
```diff
@@ -484,13 +486,33 @@ int lcddev_register(struct lcd_dev_s *dev)
...
 #endif
-		silent_reboot_lock();
+			silent_reboot_lock();
+		}
+	} else if (test_file == NULL) {
+		lcddbg("Image file not found at %s. LCD OFF\n", splash_image_path);
+	} else {
+		lcddbg("ERROR: Failed to load splash image %s\n", splash_image_path);
 	}
+	fclose(test_file);
```
eunwoo-nam: is not needed checking CONFIG_LCD_SPLASH_IMAGE?

os/drivers/lcd/mipi_lcd.c
```diff
@@ -316,6 +260,69 @@ int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table)
 	}
 }
 
+static int set_mipi_state(struct mipi_lcd_dev_s *priv, mipi_state_t new_state)
+{
+	int retries = CONFIG_LCD_SEND_VENDOR_ID_CMD_RETRY_COUNT;
```
eunwoo-nam: i think, the config name CONFIG_LCD_SEND_VENDOR_INIT_CMD_RETRY_COUNT is better

os/drivers/lcd/mipi_lcd.c
```diff
@@ -316,6 +260,69 @@ int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table)
...
+	mipi_state_t old_state = priv->mipi_state;
+	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
+	
+	if (old_state == new_state) {
+		lcddbg("Already in requested state\n");
+		return OK;
+	}
+
+	switch(new_state) {
```
eunwoo-nam, line 229: It seems beneficial to consolidate the code related to state changes in one place. 👍👍😊

It seems you considerd case of all state swap 👍

os/drivers/lcd/mipi_lcd.c
```diff
@@ -498,15 +503,17 @@ static int lcd_getpower(FAR struct lcd_dev_s *dev)
...
+static int lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
 {
-	lcddbg("Switch to CMD mode\n");
-	priv->config->mipi_mode_switch(CMD_MODE);
-	priv->lcdonoff = LCD_OFF;
-	lcm_setting_table_t display_off_cmd = {0x28, 0, {0x00}};
-	send_cmd(priv, display_off_cmd);
-	/* The power off must operate only when LCD is ON */
-	priv->config->power_off();
+	if (priv->mipi_state == MIPI_STATE_CMD_OFF) {
```
eunwoo-nam: you make set_mipi_state api.
How about replace it to get_mipi_state

os/drivers/lcd/mipi_lcd.c
```diff
@@ -523,29 +530,14 @@ static void lcd_power_off(FAR struct mipi_lcd_dev_s *priv)
...
+		return OK;
 	}
 
-	if (retries <= 0) {
-		lcddbg("ERROR: LCD Init sequence failed\n");
+	if (set_mipi_state(priv, MIPI_STATE_CMD_ON) != OK) {
 		return ERROR;
 	}
-	
 	return OK;
```
eunwoo-nam:
```suggestion
return set_mipi_state(priv, MIPI_STATE_CMD_ON);
```

os/drivers/lcd/Kconfig
```diff
@@ -201,13 +201,43 @@ config LCD_YRES
...
+
 config LCD_SPLASH_IMAGE_PATH
 	string "LCD Splash Image Path"
 	default "/res/kernel/graphics"
+	depends on LCD_SPLASH_IMAGE
 	---help---
 		This path is used find splash image during booting.
 		Final resource path is defined like "{LCD_SPLASH_IMAGE_PATH}/{LCD_YRES}x{LCD_XRES}/splash_{normal/silent}.bmp"
 
+choice
```
anjana348, line 219: Currently format of LCD pixel data is hard coded in mipi as 16bit only
amebasmart_mipi.c
priv->MIPI_InitStruct->MIPI_VideoDataFormat = MIPI_VIDEO_DATA_FORMAT_RGB565;
So we have to make sure that if we make new config like this for 24 bit and use that without changing mipi code, it will have issue

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Thank you for suggest.
I applied it for `lcd_power_on` and `lcd_power_off`.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: I made `get_mipi_state` and use this function to get mipi state.
This approach would be more safe.
Thank you.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Yes. `CONFIG_LCD_SEND_VENDOR_INIT_CMD_RETRY_COUNT` would be more precise naming for now.
I updated Kconfig and defconfig.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I thought when `CONFIG_LCD_SPLASH_IMAGE` is not defined, `loadsplash` will return error and nothing happen for this scenario.
But wrapping with `CONFIG_LCD_SPLASH_IMAGE` and remove unnecessary function calling would be better.
Thank you.

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: I updated to use `mipi_dsi_dcs_set_display_off` that mipidsi driver's function.

os/drivers/lcd/image_lib.c
```diff
@@ -0,0 +1,213 @@
...
+
+int image_load_bmp_file(const char *filename, uint8_t *buffer, int buffer_width, int buffer_height)
+{
+	bmp_header_t header;
+	bmp_info_header_t info_header;
+	FILE *bmp_file;
+	int img_width;
+	int img_height;
+
+	bmp_file = fopen(filename, "rb");
```
Taejun-Kwon: posix(open) is faster than libc (fopen, fread)

os/drivers/lcd/lcd_dev.c
```diff
@@ -484,13 +486,35 @@ int lcddev_register(struct lcd_dev_s *dev)
 
 	sem_init(&lcd_info->sem, 0, 1);
 
-	ret = lcd_init_put_image(dev);
-	if (ret == OK) { // LCD ON
+#ifdef CONFIG_LCD_SPLASH_IMAGE
```
Taejun-Kwon: This is VFS layer, so from line 490 to 499 should be moved to 'loadsplash'. Each of lower driver (actual device) knows that path of image & resolution. Hence you can get rid of below KCONFIGs

os/drivers/lcd/lcd_dev.c
```diff
@@ -484,13 +486,35 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+		snprintf(splash_image_path, sizeof(splash_image_path), "%s/%dx%d/splash_normal.bmp", CONFIG_LCD_SPLASH_IMAGE_PATH, CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+	} else {
+		snprintf(splash_image_path, sizeof(splash_image_path), "%s/%dx%d/splash_silent.bmp", CONFIG_LCD_SPLASH_IMAGE_PATH, CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+	}
+#else
+	snprintf(splash_image_path, sizeof(splash_image_path), "%s/%dx%d/splash_normal.bmp", CONFIG_LCD_SPLASH_IMAGE_PATH, CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+#endif /* CONFIG_LCD_SPLASH_SILENT_BOOT */
+
+	// Check if splash image file exists before rendering
+	FILE *test_file = fopen(splash_image_path, "rb");
```
Taejun-Kwon: I Think below is better.
```
	FILE *test_file = fopen(splash_image_path, "rb");
	if (test_file == NULL) {
		lcddbg("Image file not found at %s. LCD OFF\n", splash_image_path);
		return -ENOENT;
	}
	if (dev->loadsplash) {
		lcddbg("ERROR: loadsplash is not supported);
		fclose(test_file);
		return -ENOSYS;
	}
	
	ret = dev->loadsplash(dev, splash_image_path);
	if (ret == OK) { // LCD ON
#ifdef CONFIG_PM
		(void)pm_suspend(lcd_info->pm_domain);
#endif
		silent_reboot_lock();
	}
```

os/drivers/lcd/image_lib.c
```diff
@@ -0,0 +1,213 @@
+/****************************************************************************
```
Taejun-Kwon, line 1: you can discuss with team internally.
IMO, this is not only related to LCD. Other module can use this in future..

os/drivers/lcd/mipi_lcd.c
```diff
@@ -316,6 +260,73 @@ int send_init_cmd(struct mipi_lcd_dev_s *priv, lcm_setting_table_t *table)
...
+				return ERROR;
+			}
+		}
+		break;
+	case MIPI_STATE_VIDEO_ON:
+		if (old_state == MIPI_STATE_CMD_ON) {
+			priv->config->mipi_mode_switch(VIDEO_MODE);
+		} else {
+			lcddbg("Powering up should be done before switching to video mode\n");
+			return ERROR;
```
eunwoo-nam: we can turn on lcd ic and change mipi mode here

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: I move this feature to `lcddev_register` in lcd_dev.c

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Right this is not dependent to mipi.
I moved this function and `lcd_rotate_buffer` to lcd_dev.c

os/drivers/lcd/lcd_dev.c
seokhun-eom24: Thank you for advice.
But I thought loading image is not dependent to specific lcd driver.
So I replaced all load image related feature to lcd_dev.c after your review.

os/drivers/lcd/image_lib.c
seokhun-eom24: Updated to use open and other libc functions.
Thank you.

os/drivers/lcd/lcd_dev.c
Taejun-Kwon: Definately, Splash image has dependecies.
1. Type of Product
2. Type of GUI guide
3. Resolution of LCD
How will you handle all of these in VFS layer?

seokhun-eom24: I understand your point.
Normally, an LCD driver doesn’t know which image it should display. Therefore, each specific LCD must provide information about itself and the splash image; the driver obtains this data through its operations and then renders the image.

For now, the splash‑image path is predefined by the TizenRT platform, so this approach works. However, we should implement a more flexible solution in the future.

os/drivers/lcd/image_lib.c
seokhun-eom24: Right. This mini image library can be used by other modules in the future.
We should discuss whether to adopt an open‑source library or develop our own implementation.
I hope we can have that discussion soon.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I applied this approach for checking test_file and call functions.
Thank you.

os/drivers/lcd/Kconfig
seokhun-eom24: First of all, thank you for pointing out an important point.

In my view, the MIPI format and the splash image format are independent.

To support new format like 24 bit, I think the two things below should be implemented.
1. A Kconfig for the MIPI format must be provided.
    - Supporting various formats through such a configuration would require a large‑scale overhaul, involving changes to the init command, buffer handling, render functions, and other parts of mipi and lcd.

2. Code that converts between the splash format and the MIPI format (e.g., rgb565 → rgb888) should be supplied in image_lib.c.
    - It would be preferable to make not only the color format but also the image format mutually compatible.

If we only need to support a configuration such as LCD_SPLASH_IMAGE_BMP_24BIT, we can simply provide conversion code in image_lib.c. 
However, supporting a configuration like MIPI_VIDEO_DATA_FORMAT_RGB888 would require a substantial amount of work.

Therefore, at this stage I believe it is realistic to either leave things as they are or to prepare the configuration so that the MIPI format is supported only with the single option MIPI_VIDEO_DATA_FORMAT_RGB565. 
If support for additional formats is required, it should be handled in a separate PR.

May I ask what you think about this approach?

os/drivers/lcd/lcd_dev.c
```diff
@@ -484,23 +546,80 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+
+	ret = image_load_bmp_file(splash_image_path, fullscreen_buffer, CONFIG_LCD_YRES, CONFIG_LCD_XRES);
+	if (ret != OK) {
+		lcddbg("ERROR: Failed to load splash image %s\n", splash_image_path);
+		goto skip_splash_image;
+	}
+	if (!lcd_info->dev->setpower) {
+		lcddbg("ERROR: LCD setpower is not supported");
+		return -ENOSYS;
+	}
```
eunwoo-nam: This lcd regester fuction is very big
how about seperate it?

os/drivers/lcd/mipi_lcd.c
seokhun-eom24: Right. That would be better than just return ERROR.
Fixed.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I seperated `lcddev_start_flush_thread` and `lcddev_display_splash` from `lcddev_register`.
Thank you for suggest.

seokhun-eom24:
Verified with 2.8'' and 4.3'' LCD.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -817,52 +711,14 @@ static int lcd_render_bmp(FAR struct lcd_dev_s *dev, const char *bmp_filename)
...
-			int dest_y = src_x;
-			int target_buffer_index = ((y_offset + dest_y) * xres + (x_offset + dest_x)) * 2;
-			uint16_t bmp_pixel = temp_row_buffer[src_x];
-			uint16_t rgb565_pixel = convert_pixel_to_rgb565(bmp_pixel, r_mask, g_mask, b_mask);
-			fullscreen_buffer[target_buffer_index] = (uint8_t)(rgb565_pixel & 0xFF);
-			fullscreen_buffer[target_buffer_index + 1] = (uint8_t)((rgb565_pixel >> 8) & 0xFF);
-		}
+	
+	if (image_load_bmp_file(bmp_filename, fullscreen_buffer, yres, xres) != OK) {
+		lcddbg("Failed to load BMP file\n");
```
anjana348: shouldn't we free buffer before returning error 
#if !defined(CONFIG_LCD_SW_ROTATION)
	if (fullscreen_buffer) {
		kmm_free(fullscreen_buffer);
	}
#endif

seokhun-eom24: Thank you for pointing out.
I fixed this in `lcd_dev.c`.
```c
	ret = lcddev_display_splash(lcd_info, fullscreen_buffer);
	if (ret != OK) {
#ifndef CONFIG_LCD_SW_ROTATION
		kmm_free(fullscreen_buffer);
#endif
		/* If ret == -ENOENT, it means the splash image is not found, so it's not error */
		if (ret != -ENOENT) {
			goto err_out;
		}
	}
```

os/drivers/lcd/mipi_lcd.c
```diff
@@ -41,95 +43,42 @@
...
 	struct mipi_lcd_config_s *config;
 };
 
 static struct mipi_lcd_dev_s g_lcdcdev;
 
+#ifdef CONFIG_LCD_SW_ROTATION
+#define NUM_OF_LCD_BUFFER	2
+extern uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER];
+extern int lcd_buffer_index;
+extern void lcd_rotate_buffer(short int* src, short int* dst);
```
eunwoo-nam: Could you please let me know why extern is used?

seokhun-eom24: `lcd_buffer`, `lcd_buffer_index`, `lcd_rotate_buffer` are common LCD variables and function.  
So I moved these into `lcd_dev.c`. Therefore, we need to use `extern` to access them here.

os/drivers/lcd/mipi_lcd.c
```diff
@@ -390,10 +354,8 @@ static int lcd_putarea(FAR struct lcd_dev_s *dev, fb_coord_t row_start, fb_coord
 #else
 	priv->config->lcd_put_area((u8 *)buffer, row_start, col_start, row_end, col_end);
 #endif
```
eunwoo-nam: If you move lcd sw rotation feature to common, this code should be moved.

seokhun-eom24: I moved every sw rotation feature from `mipi_lcd.c` to `lcd_dev.c`

os/drivers/lcd/lcd_dev.c
```diff
@@ -264,7 +346,11 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		size_t row_size = lcd_area->stride > 0 ? lcd_area->stride : cols * pixel_size;
 
 		if (priv->planeinfo[lcd_area->planeno].putarea) {
+#if defined(CONFIG_LCD_SW_ROTATION)
+			ret = lcd_putarea_with_rotation(&priv->planeinfo[lcd_area->planeno], priv->dev, lcd_area->row_start, lcd_area->row_end, lcd_area->col_start, lcd_area->col_end, lcd_area->data, row_size);
+#else
 			ret = priv->planeinfo[lcd_area->planeno].putarea(priv->dev, lcd_area->row_start, lcd_area->row_end, lcd_area->col_start, lcd_area->col_end, lcd_area->data, row_size);
+#endif
```
eunwoo-nam: How about make function for it and flusing thread function?

os/drivers/lcd/lcd_dev.c
```diff
@@ -453,62 +629,78 @@ int lcddev_register(struct lcd_dev_s *dev)
...
+err_out:
 	lcddbg("ERROR: Failed to register driver %s\n", devname);
 #if defined(CONFIG_LCD_FLUSH_THREAD)
-	task_delete(pid);
+	if (flush_thread_pid >= 0) {
+		task_delete(flush_thread_pid);
+	}
 	sem_destroy(&lcd_info->flush_buffer_sem);
 	sem_destroy(&lcd_info->copy_buffer_sem);
 	kmm_free(lcd_info->lcd_kbuffer);
```
eunwoo-nam: Since lcddev_start_flush_thread  is made into a function, it seems good to make these functions to match the pair.

os/drivers/lcd/lcd_dev.c
```diff
@@ -453,62 +629,78 @@ int lcddev_register(struct lcd_dev_s *dev)
...
-	if (lcd_info->dev->getplaneinfo) {
-		lcd_info->dev->getplaneinfo(lcd_info->dev, 0, &lcd_info->planeinfo);	//plane no is taken 0 here
-		snprintf(devname, 16, "/dev/lcd0");
-		ret = register_driver(devname, &g_lcddev_fops, 0666, lcd_info);
-		if (ret == OK) {
-			return ret;
+	ret = lcddev_display_splash(lcd_info, fullscreen_buffer);
+	if (ret != OK) {
+#ifndef CONFIG_LCD_SW_ROTATION
+		kmm_free(fullscreen_buffer);
```
eunwoo-nam: if flushing is done, we can't free this image until next flusing is comming.
because DMA is using it.

current, we never free this image.
then, static memory is better
or when next frame is comming, we can free this memory.

os/drivers/lcd/lcd_dev.c
```diff
@@ -117,10 +116,91 @@ static const struct file_operations g_lcddev_fops = {
 	lcddev_ioctl,				/* ioctl */
 };
 
+#ifdef CONFIG_LCD_SW_ROTATION
+#define NUM_OF_LCD_BUFFER	2
+uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER];
```
eunwoo-nam: Let's move it into struct lcd_s

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I made `lcddev_putarea`, so now we don't need to care about sw rotation.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I added `lcddev_stop_flush_thread` function.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: You mean when `CONFIG_LCD_SW_ROTATION` is not set and `CONFIG_LCD_SPLASH_IMAGE=y` case, fullscreen_buffer is allocated and not freed.
I updated to free when next frame is comming.
I considered to use static memory, but `320 x 240 x 2 + 1` about 15K is too big, so I choose to free when next frame is comming

os/drivers/lcd/lcd_dev.c
seokhun-eom24: Updated.

os/drivers/lcd/lcd_dev.c
```diff
@@ -83,15 +87,20 @@ struct lcd_s {
...
 	sem_t flush_buffer_sem;
 	sem_t copy_buffer_sem;
 #endif
+#ifdef CONFIG_LCD_SW_ROTATION
+	uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER];
+	int lcd_buffer_index;
+#endif
+#if defined(CONFIG_LCD_SPLASH_IMAGE) && !defined(CONFIG_LCD_SW_ROTATION)
+	uint8_t *splash_buffer;
+#endif
```
eunwoo-nam, line 103:
```suggestion
#ifdef CONFIG_LCD_SW_ROTATION
	uint8_t *lcd_buffer[NUM_OF_LCD_BUFFER];
	int lcd_buffer_index;
#else
#ifdef CONFIG_LCD_SPLASH_IMAGE
	uint8_t *splash_buffer;
#endif
#endif
```

os/drivers/lcd/lcd_dev.c
```diff
@@ -264,7 +356,7 @@ static int lcddev_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 		size_t row_size = lcd_area->stride > 0 ? lcd_area->stride : cols * pixel_size;
 
 		if (priv->planeinfo[lcd_area->planeno].putarea) {
-			ret = priv->planeinfo[lcd_area->planeno].putarea(priv->dev, lcd_area->row_start, lcd_area->row_end, lcd_area->col_start, lcd_area->col_end, lcd_area->data, row_size);
+			ret = lcddev_putarea(priv, lcd_area->planeno, lcd_area->row_start, lcd_area->row_end, lcd_area->col_start, lcd_area->col_end, lcd_area->data, row_size);
```
eunwoo-nam: How about include this codes into lcddev_putarea

os/drivers/lcd/lcd_dev.c
```diff
@@ -400,12 +490,107 @@ static void lcd_flushing_thread(int argc, char **argv)
...
+#endif /* CONFIG_LCD_SPLASH_SILENT_BOOT */
+
+	int test_fd = open(splash_image_path, O_RDONLY);
+	if (test_fd < 0) {
+		lcddbg("No splash image found at %s.\n", splash_image_path);
+		return -ENOENT;
+	}
+	close(test_fd);
+
+	ret = image_load_bmp_file(splash_image_path, fullscreen_buffer, CONFIG_LCD_YRES, CONFIG_LCD_XRES);
```
eunwoo-nam: do we need to ret?

os/drivers/lcd/lcd_dev.c
```diff
@@ -435,10 +620,11 @@ int lcddev_register(struct lcd_dev_s *dev)
 	char devname[16] = { 0, };
 	int ret;
 	struct lcd_s *lcd_info;
+#ifdef CONFIG_LCD_SPLASH_IMAGE
+	uint8_t *fullscreen_buffer = NULL;
```
eunwoo-nam: how about using lcd_info->splash_buffer?

os/drivers/lcd/lcd_dev.c
seokhun-eom24: Updated to use splash_buffer when `CONFIG_LCD_SPLASH_IMAGE=y`

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I updated to check condtion and return `-ENOSYS` in `lcddev_putarea`

os/drivers/lcd/lcd_dev.c
seokhun-eom24: Removed ret.

os/drivers/lcd/lcd_dev.c
seokhun-eom24: I updated to use `lcd_info->splash_buffer` instead of fullscreen_buffer.

PR #7027, opened 2025-10-31
apps/examples/lcd_test: Correct byte order for LCD color

seokhun-eom24:
Corrects the byte ordering in LCD data transmission by swapping the high and low bytes to match the expected format. Also fixes the GREEN color definition from 0xE007 to 0x07E0 for proper color representation.

0x07E0 -> 00000 111111 00000 -> RED (5 bits) | GREEN (6 bits) | BLUE (5 bits) 
(0x07E0 & 0xFF00) >> 8 => (0x0700) >> 8 => 0x07 -> HIGH BYTE, so it should be store in lcd_data[idx + 1] 
(0x07E0 & 0x00FF) => 0xE0 -> LOW BYTE, so it should be store in lcd_data[idx]

### DIFF
--- a/apps/examples/lcd_test/example_lcd.c
+++ b/apps/examples/lcd_test/example_lcd.c
@@ -66,7 +66,7 @@
 #define RED   0xF800
 #define WHITE 0xFFFF
 #define BLACK 0x0000
-#define GREEN 0xE007
+#define GREEN 0x07E0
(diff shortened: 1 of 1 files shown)


sunghan-chang (commented):
Is this verified with ST7785 and ST7701SN both?

sunghan-chang (approved):

PR #7082, opened 2025-12-15
os/arch/arm/amebasmart: Remove CONFIG_PM from up_timerisr

seokhun-eom24:
PM is not related with `up_timerisr`, so remove `#ifdef CONFIG_PM` from up_timerisr.
If there's missing tick whether `CONFIG_PM` is enabled or not, we should compensate it.


sunghan-chang (approved):

sunghan-chang:
@zhongnuo-tang @EangJS Could you review this?

zhongnuo-tang (approved):

EangJS (approved):

eunwoo-nam (approved):

PR #7091, opened 2025-12-22
os/arch: Add I2C timeout error log

seokhun-eom24:
Add explicit error logging using `i2cerr` when an I2C timeout occurs.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_i2c.c
+++ b/os/arch/arm/src/amebad/amebad_i2c.c
@@ -520,11 +520,14 @@ static inline int amebad_i2c_sem_waitdone(FAR struct amebad_i2c_priv_s *priv)
 
 	while (priv->intstate != INTSTATE_DONE && elapsed < timeout);
 
-	i2cinfo("intstate: %d elapsed: %d threshold: %d status: %08x\n", priv->intstate, elapsed, timeout, priv->status);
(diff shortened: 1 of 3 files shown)


eunwoo-nam (approved):

allen-kim-sec (approved):
it just append logs in case of I2C Timeout

os/arch/arm/src/amebad/amebad_i2c.c
```diff
@@ -520,11 +520,14 @@ static inline int amebad_i2c_sem_waitdone(FAR struct amebad_i2c_priv_s *priv)
 
 	while (priv->intstate != INTSTATE_DONE && elapsed < timeout);
 
-	i2cinfo("intstate: %d elapsed: %d threshold: %d status: %08x\n", priv->intstate, elapsed, timeout, priv->status);
+	if (priv->intstate != INTSTATE_DONE) {
+		i2cerr("ERROR i2c timeout: intstate: %d elapsed: %d threshold: %d status: %08x\n", priv->intstate, elapsed, timeout, priv->status);
+		ret = -ETIMEDOUT;
+	} else {
```
Taejun-Kwon, line 526: Sometimes, it returns -EAGAIN here, I think 'ret' also need to be printed here.

PR #7092, opened 2025-12-22
drivers/serial: Use lldbg in uart_register

seokhun-eom24:
To prevent idle task errno from being set to `EBADF` during boot, replace `dbg()` with `lldbg()` in `uart_register()`.
Because during `uart_register` stdout is not ready, so we have to use `lldbg`.

Registering log are printed like below.
```
up_initialize: [Reboot Reason] : 55
uart_register: Registering /dev/console
uart_register: Registering /dev/ttyS0
uart_register: Registering /dev/ttyS1
uart_register: Registering /dev/ttyS2
Logtask init ok!
```

### DIFF
--- a/os/drivers/serial/serial.c
+++ b/os/drivers/serial/serial.c
@@ -1357,6 +1357,8 @@ errout_with_sem:
 
 int uart_register(FAR const char *path, FAR uart_dev_t *dev)
 {
+	int ret;
+
 	/* Initialize semaphores */
 	sem_init(&dev->xmit.sem, 0, 1);
 	sem_init(&dev->recv.sem, 0, 1);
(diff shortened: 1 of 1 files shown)


pcs1265 (approved):
LGTM

eunwoo-nam (commented):
is it mandatory logs during boot? if it's no, please use vlldbg.

The dbg is using stdout stream which is used `/dev/console`. the console register is needed replacing to lldbg
And, the UART Register is called before `group_setupidlefiles` which is getting fd of /dev/console`, so other registering also is needed lldbg.
It's correct changs. 👍👍👍

seokhun-eom24:
> is it mandatory logs during boot? if it's no, please use vlldbg.
> 
> The dbg is using stdout stream which is used `/dev/console`. the console register is needed replacing to lldbg And, the UART Register is called before `group_setupidlefiles` which is getting fd of /dev/console`, so other registering also is needed lldbg. It's correct changs. 👍👍👍

Right. Success log is not mandatory during boot.
I seperate log for success and fail case to use `llvdbg` and `lldbg` for each case.

eunwoo-nam (approved):

eunwoo-nam (approved):

PR #7102, opened 2025-12-29
os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c: Fix sem_wait to retry on EINTR

seokhun-eom24:
Previously, `sem_wait` did not retry when interrupted by a signal (EINTR), which could cause issues.
Fixed by adding a retry loop.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
+++ b/os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
@@ -19,6 +19,7 @@
  * Included Files
  ****************************************************************************/
 #include <tinyara/config.h>
+#include <errno.h>
 #if defined(CONFIG_LCD_ST7785)
 #include <tinyara/lcd/st7785.h>
 #elif defined(CONFIG_LCD_ST7701)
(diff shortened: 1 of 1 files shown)


seokhun-eom24:
@anjana348 @namanjain7 Could you review the PR?

os/board/rtl8730e/src/rtl8730e_mipi_lcdc.c
```diff
@@ -203,7 +205,12 @@ static void rtl8730e_lcd_put_area(u8 *lcd_img_buffer, u32 x_start, u32 y_start,
 #endif
 	LCDC_LayerConfig(pLCDC, LCD_LAYER, &lcdc_init_struct.layerx[LCD_LAYER]);
 	DCache_CleanInvalidate((u32)lcd_img_buffer, LCDC_IMG_BUF_SIZE);
-	sem_wait(&g_next_frame_block);
+
+	do {
+		ret = sem_wait(&g_next_frame_block);
+		DEBUGASSERT(ret == OK || errno == EINTR);
+	} while (ret < 0);
```
eunwoo-nam: The ret is twice checking. how about below?
```suggestion
	while (sem_wait(&g_next_frame_block) != OK) {
		DEBUGASSERT(errno == EINTR);
	}
```

seokhun-eom24: Thank you for suggest.
`ret` value is useless, so your suggestion would be better.

eunwoo-nam (approved):

PR #7121, opened 2026-01-19
os/mm: Add DEBUG_MM_FREEINFO for enhanced memory free debugging

seokhun-eom24:
### 1. os/mm: Add DEBUG_MM_FREEINFO for enhanced memory free debugging
Add a new configuration option CONFIG_DEBUG_MM_FREEINFO to enable tracking of memory deallocation.
Record the address and PID of the caller that first requested the free operation, allowing for detection of double free errors.
Update mm_freenode_s, mm_delaynode_s structures and related functions to support this feature.

sizeof(struct mm_freenode_s) for 32bit architecture
case 1. NO CONFIG -> 16bytes, MM_MIN_SHIFT = 4, MIN_CHUNK_SIZE = 16
case 2. CONFIG_DEBUG_MM_HEAPINFO -> 24bytes, MM_MIN_SHIFT = 5, MIN_CHUNK_SIZE = 32
case 3. CONFIG_DEBUG_MM_HEAPINFO & CONFIG_DEBUG_MM_FREEINFO -> 32bytes, MM_MIN_SHIFT = 5, MIN_CHUNK_SIZE = 32
So there's no additional memory overhead for case 3 compare to case 2.

### 2. configs/rtl8730e: Update defconfig files about DEBUG_MM
Enable `CONFIG_DEBUG_MM_FREEINFO` and `CONFIG_DEBUG_CHECK_FRAGMENTATION`.

### 3. os/mm: Refactor mm related definitions and configurations
Add table to summarize the impact of various configurations to MM_MIN_SHIFT.
Refactor the mm related definitions to make it more readable.

### 4. os/mm: Remove CONFIG_MM_SMALL configuration
Remove the CONFIG_MM_SMALL configuration as it is no longer needed.
This configuration is for support 16-bit MCU address space, and it is not used in the current code.

### 5. build/configs: Remove CONFIG_MM_SMALL from defconfig
Remove `CONFIG_MM_SMALL` from every defconfigs.
Remove `CONFIG_MM_SMALL` section from docs.


os/include/tinyara/mm/mm.h
```diff
@@ -129,40 +129,64 @@
...
+#else /* defined(CONFIG_DEBUG_MM_HEAPINFO) || defined(CONFIG_DEBUG_MM_FREEINFO) */
+/* sizeof(struct mm_freenode_s) is 16 bytes. */
+#define MM_MIN_SHIFT    4			/* 16 bytes */
+#define MM_SHIFT_FOR_NDX 5			/* MM_MIN_SHIFT + 1 */
 #define MM_SHIFT_MASK    0xffffffe0	/* ~(1 << MM_SHIFT_FOR_NDX) */
+#endif /* defined(CONFIG_DEBUG_MM_HEAPINFO) || defined(CONFIG_DEBUG_MM_FREEINFO) */
+#define MM_MAX_SHIFT   22			/* 4 Mb */
 #endif
 
 /* All other definitions derive from these two */
```
pcs1265, line 152: It looks like to be better to place `MM_SHIFT_FOR_NDX` and `MM_SHIFT_MASK` here if they depends on `MM_MIN_SHIFT`
What do you think?

os/mm/mm_heap/mm_free.c
```diff
@@ -204,13 +208,43 @@ void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
...
+ * Name: mm_free
+ *
+ * Description:
+ *   Returns a chunk of memory to the list of free nodes, merging with
+ *   adjacent free chunks if possible.
+ *
+ ****************************************************************************/
+void mm_free(FAR struct mm_heap_s *heap, FAR void *mem)
+{
+	mm_free_internal(heap, mem, __builtin_return_address(0), getpid());
```
eunwoo-nam, line 259: we don't need to put pid argument here. 
it's increase usage stack

you can call getpid() when pid value is restored

os/mm/mm_heap/mm_malloc.c
```diff
@@ -227,6 +226,10 @@ FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_
 			remainder = (FAR struct mm_freenode_s *)(((char *)node) + size);
 			remainder->size = remaining;
 			remainder->preceding = size;
+#ifdef CONFIG_DEBUG_MM_FREEINFO
+			remainder->free_call_addr = NULL;
+			remainder->free_call_pid = 0;
+#endif
```
eunwoo-nam, line 239: It is not necessary for the purpose for which you are pushing this PR.
Because There is a low probability that the app has this address.

If you wanna clear it, how about identify it with magic number

os/include/tinyara/mm/mm.h
seokhun-eom24: I added 1 more commit for better readability including your comment.

os/mm/mm_heap/mm_free.c
```diff
@@ -86,25 +86,23 @@ static void mm_add_delaylist(FAR struct mm_heap_s *heap, FAR void *mem)
 	flags = enter_critical_section();
 
 	tmp->flink = heap->mm_delaylist[up_cpu_index()];
+	tmp->free_call_addr = free_call_addr;
+	tmp->free_call_pid = free_call_pid;
```
eunwoo-nam, line 95: how about use free node structure?

this location is same freenode->flink blink
it can be confusing align when debugging

os/include/tinyara/mm/mm.h
```diff
@@ -303,19 +304,28 @@ struct mm_freenode_s {
...
-#define SIZEOF_MM_FREENODE (SIZEOF_MM_ALLOCNODE + 2 * MM_PTR_SIZE)
+
+#define SIZEOF_MM_FREENODE \
+	(SIZEOF_MM_ALLOCNODE + 2 * MM_PTR_SIZE + SIZEOF_MM_FREE_DEBUG_INFO)
 
 #define CHECK_FREENODE_SIZE \
 	DEBUGASSERT(sizeof(struct mm_freenode_s) == SIZEOF_MM_FREENODE)
 
 struct mm_delaynode_s
 {
```
sunghan-chang: Regardless of this PR, let's fix coding rule violation.
```c
struct mm_delaynode_s {
```

seokhun-eom24: Thank you. Fixed.

os/mm/mm_heap/mm_free.c
seokhun-eom24: We need to pass `getpid()` to use same `mm_free_internal` fucntion with `mm_free_withinfo` that called from `mm_free_delaylist`.
If we split function with 3 args function and 4 args function, it'll have too many duplicated code.

os/mm/mm_heap/mm_malloc.c
seokhun-eom24: Updated to use magic number for address and pid.
`0xFEEEFEEE` is commonly used for freed heap memory.
```
#define MM_REMAINDER_FREE_CALL_ADDR ((void *)0xFEEEFEEE)
#define MM_REMAINDER_FREE_CALL_PID 	((pid_t)-1)
```

os/mm/mm_heap/mm_free.c
seokhun-eom24: I updated `mm_delaynode_s` silimar to `mm_freenode_s`. Now `free_call_addr` and `free_call_pid` are stored in right place.
```c
struct mm_delaynode_s {
	FAR struct mm_delaynode_s *flink;
	FAR struct mm_delaynode_s *blink; /* Not used, but keep structure alignment with mm_freenode_s */
#ifdef CONFIG_DEBUG_MM_FREEINFO
	mmaddress_t free_call_addr;
	pid_t free_call_pid;
	uint16_t reserved2;
#endif
};
```

eunwoo-nam (approved):

sunghan-chang (approved):

PR #7127, opened 2026-01-21
os/drivers/memory: Fix mminfo_ioctl return value correctly

seokhun-eom24:
Changed initial return value from `-ENOSYS` to `OK` for better default handling.
This patch makes `mminfo_ioctl` return `OK` for `MMINFOIOC_MNG_ALLOCFAIL` and `MMINFOIOC_GC` cases.
Before this patch it was always returning `-ENOSYS`.
And other cases returns correct return value same as before.

### DIFF
--- a/os/drivers/memory/mminfo.c
+++ b/os/drivers/memory/mminfo.c
@@ -75,7 +75,7 @@ static ssize_t mminfo_write(FAR struct file *filep, FAR const char *buffer, size
  ************************************************************************************/
 static int mminfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 {
-	int ret = -ENOSYS;
+	int ret = OK;
 	struct mallinfo mem;
 	struct mallinfo *retmem;
 #if defined(CONFIG_DEBUG_MM_HEAPINFO)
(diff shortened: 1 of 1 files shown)


os/drivers/memory/mminfo.c
```diff
@@ -150,10 +150,12 @@ static int mminfo_ioctl(FAR struct file *filep, int cmd, unsigned long arg)
 	case MMINFOIOC_MNG_ALLOCFAIL:
 		/* There is a single heap for user. So start and end indexes of heap are always 0. */
 		mm_manage_alloc_fail(BASE_HEAP, 0, 0, ((struct mm_alloc_fail_s *)arg)->size, ((struct mm_alloc_fail_s *)arg)->align, USER_HEAP, ((struct mm_alloc_fail_s *)arg)->caller);
+		ret = OK;
 		break;
 
 	case MMINFOIOC_GC:
 		sched_garbagecollection();
+		ret = OK;
```
Taejun-Kwon:
```
	ret = OK;
	switch() {
	case xxxx :
		break;
	case yyyy :
		ret = -EINVAL;
		break;
	default :
		ret = -ENOSYS;
		break;
	return ret;
```
is more clear & reduced code line

seokhun-eom24: Thank you for suggest better approach.
I updated this PR and I'll keep this advice in mind.

pcs1265 (approved):
LGTM

sunghan-chang (approved):

PR #7131, opened 2026-01-22
os/kernel/binary_manager.c: Add binary_manager_sync_bootparam

jylee9613:
All useidx and active_idx should be same in bootparam.
But sometimes partition indices of each binary are being set differently during factory process, and it causes booting failure.

This commit adds `binary_manager_sync_bootparam` to sync all binaries' partition to kernel's active partition.
(Because both BPs are corrupted, bootloader restores kernel address.)
However, in order to prevent kernel from rollback due to bootloader's restoration, `binary_manager_sync_bootparam` checks the partition which contains the latest version of kernel.

[Prerequisite]
Partition mixing in bootparam can occur only if binaries exist in both A and B partitions.

[example]
```
- Current bootparam
---------------------------------
kernel | common | app1 | resource
A      | B      | A    | B
---------------------------------

- Updated bootparam (updated by `binary_manager_sync_bootparam`)
---------------------------------
kernel | common | app1 | resource
A      | A      | A    | A
---------------------------------
```

### DIFF
--- a/apps/examples/kernel_update_test/kernel_update_test_main.c
+++ b/apps/examples/kernel_update_test/kernel_update_test_main.c
@@ -619,6 +619,16 @@ static int binary_update_run_tests(void)
 	return OK;
 }
 
+static void show_usage(void)
+{
+	printf("Usage: kernel_update <test_type>\n");
+	printf("Available test types:\n");
+	printf("  all          - Run all tests\n");
+	printf("  same_version - Run same version test\n");
+	printf("  new_version  - Run new version test\n");
+	printf("  invalid      - Run invalid binary test\n");
+}
+
 /****************************************************************************
  * binary_update_aging_test
  ****************************************************************************/
@@ -628,5 +638,39 @@ int main(int argc, FAR char *argv[])
 int kernel_update_main(int argc, char *argv[])
 #endif
 {
-	binary_update_run_tests();	
+	if (argc == 2 && !strncmp(argv[1], "all", 4)) {
+		/* Run all tests */
+		return binary_update_run_tests();
+
+	} else if (argc == 2 && !strncmp(argv[1], "same_version", 13)) {
+		/* Get info all test. */
+		binary_update_getinfo_all();
+		
+		int ret = binary_update_same_version_test();
+		if (ret != OK) {
+			printf("Same version test failed\n");
+		}
+		return ret;
+	} else if (argc == 2 && !strncmp(argv[1], "new_version", 12)) {
+		/* Get info all test. */
+		binary_update_getinfo_all();
+		
+		int ret = binary_update_new_version_test();
+		if (ret != OK) {
+			printf("New version test failed\n");
+		}
+		return ret;
+	} else if (argc == 2 && !strncmp(argv[1], "invalid", 8)) {
+		/* Get info all test. */
+		binary_update_getinfo_all();
+		
+		int ret = binary_update_invalid_binary_test();
+		if (ret != OK) {
+			printf("Invalid binary test failed\n");
+		}
+		return ret;
+	}
+	printf("Invalid test type: %s\n", argv[1]);
+	show_usage();
+	return ERROR;
 }
--- a/apps/examples/kernel_update_test/kernel_update_test_main.c
+++ b/apps/examples/kernel_update_test/kernel_update_test_main.c
@@ -45,10 +45,22 @@
 /* Kernel binary information for update test */
 #define KERNEL                  "kernel"
 
-/* Common/app1 binary names for testing */
-/* Kernel / Common / App1 binary must be set at once */
+/* Common/app1/app2 binary names for testing */
+/* Kernel / Common / App1 / App2 binary must be set at once */
 #define COMMON                   "common"
 #define APP1                     "app1"
+#define APP2                     "app2"
+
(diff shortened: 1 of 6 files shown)


os/kernel/binary_manager/binary_manager.c
```diff
@@ -151,6 +151,15 @@ int binary_manager(int argc, char *argv[])
 	}
 #endif
 
+#ifdef CONFIG_USE_BP
+	/* Check and synchronize bootparam indices at boot time */
+	ret = binary_manager_sync_bootparam_partitions();
+	if (binary_manager_sync_bootparam_partitions() != BINMGR_OK) {
```
eunwoo-nam, line 157:
```suggestion
	if (ret != BINMGR_OK) {
```

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -422,3 +423,104 @@ void binary_manager_set_bpidx(uint8_t index)
...
+	memset(bootparam, 0xff, BOOTPARAM_SIZE);
+
+	/* Get current bootparam data */
+	memcpy(&bp_data, binary_manager_get_bpdata(), sizeof(binmgr_bpdata_t));
+	kernel_active_idx = bp_data.active_idx;
+	bmdbg("Current bootparam kernel's active_idx: %u\n", kernel_active_idx);
+
+	/* Check current kernel version */
+	ret = binary_manager_check_kernel_update();
+	if (ret == BINMGR_OK) {
```
eunwoo-nam, line 462: do we need to check crc?

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -422,3 +423,104 @@ void binary_manager_set_bpidx(uint8_t index)
...
+		bmdbg("Change bootparam kernel's active_idx to the latest\n");
+	} else if (ret == BINMGR_ALREADY_UPDATED || ret == BINMGR_NOT_FOUND) {
+		bmdbg("Current kernel binary is the latest version\n");
+	} else {
+		bmdbg("Fail to check kernel update, %d\n", ret);
+		goto cleanup;
+	}
+
+	/* Check resource_active_idx with kernel's active_idx */
+#ifdef CONFIG_RESOURCE_FS
```
eunwoo-nam, line 475: if resource is broken, This function will be not performed

even If resource mount is failed, we need to call this function .

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -422,3 +423,104 @@ void binary_manager_set_bpidx(uint8_t index)
...
+	/* Get current bootparam data */
+	memcpy(&bp_data, binary_manager_get_bpdata(), sizeof(binmgr_bpdata_t));
+	kernel_active_idx = bp_data.active_idx;
+	bmdbg("Current bootparam kernel's active_idx: %u\n", kernel_active_idx);
+
+	/* Check current kernel version */
+	ret = binary_manager_check_kernel_update();
+	if (ret == BINMGR_OK) {
+		/* Update index for inactive partition */
+		bp_data.active_idx ^= 1;
```
eunwoo-nam, line 464: How about verify active_idx  0 or 1

os/kernel/binary_manager/binary_manager_bootparam.c
```diff
@@ -422,3 +423,104 @@ void binary_manager_set_bpidx(uint8_t index)
...
+		bmdbg("Fail to check kernel update, %d\n", ret);
+		goto cleanup;
+	}
+
+	/* Check resource_active_idx with kernel's active_idx */
+#ifdef CONFIG_RESOURCE_FS
+	if (bp_data.resource_active_idx != kernel_active_idx) {
+		bmdbg("Resource partition mismatch: resource_active_idx (%u) != kernel_active_idx (%u)\n", 
+			bp_data.resource_active_idx, kernel_active_idx);
+		need_update_bp = true;
```
seokhun-eom24, line 479: I'm not certain, but when I followed the code flow, it seems there could be a problem like the one below.

I think `need_update_bp = true;` should move to in line 464 after `bp_data.active_idx ^= 1;`.
Like below case, `bp_data.active_idx` will be updated, but other, app idx will be same with updated active_idx. `need_update_bp = true;` will be not called. And in this case, we don't reboot system.

```
- Current bootparam
-------------------------------------
kernel | common | app1   | resource
A(old) | B(new) | B(new) | B(new)
-------------------------------------
```

apps/examples/kernel_update_test/kernel_update_test_main.c
```diff
@@ -628,5 +786,54 @@ int main(int argc, FAR char *argv[])
...
+		int ret = test_sync_bootparam_partitions();
+		if (ret != OK) {
+			printf("Sync bootparam partitions test failed\n");
+		} else {
+			printf("[REBOOT] Success test_sync_bootparam_partitions. Board will be reboot...\n");
+			boardctl(BOARDIOC_RESET, EXIT_SUCCESS);
+		}
+		return ret;
+	}
+	printf("Invalid test type: %s\n", argv[1]);
```
seokhun-eom24, line 836: `argv[1]` can be NULL here.
Wrapping with if statement would be safe.

PR #7134, opened 2026-01-26
os/mm: Add freed size log when gc works

seokhun-eom24:
- Add freed memory size log when gc works in `mm_malloc` and `mm_memalign`.
- This feature is only available when `CONFIG_DEBUG_MM_HEAPINFO` is enabled.
- This feature is not protected by other heap operations. So it may not be 100% accurate.
- I chose no protection approach, because performance is more important than log accuracy.

[Example Log]
```
mm_malloc: Allocation failed!!! We dont have enough memory. Try to free dead task stack areas
mm_malloc: GC freed 64 bytes
```

### DIFF
--- a/os/mm/mm_heap/mm_malloc.c
+++ b/os/mm/mm_heap/mm_malloc.c
@@ -146,6 +146,9 @@ FAR void *mm_malloc(FAR struct mm_heap_s *heap, size_t size, mmaddress_t caller_
 	void *ret = NULL;
 	int ndx;
 	bool gc_done = false;
+#ifdef CONFIG_DEBUG_MM_HEAPINFO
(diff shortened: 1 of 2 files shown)


eunwoo-nam (commented):
how about printing garbege collection size log?
it's more simple than snapshot

seokhun-eom24:
> how about printing garbege collection size log? it's more simple than snapshot

I implemented it again with simpler approach. 
Can you review again?

eunwoo-nam (approved):

pcs1265 (approved):
LGTM

allen-kim-sec (approved):
LGTM

PR #7163, opened 2026-02-16
apps/examples/rtc, os/arch/arm/src/armino, os/include/tinyara: Add RTC millisecond test functionality for BK7239N

abhinav-s235:
This commit adds a new "get_ms" command to the RTC example application for testing the aon_rtc_get_ms interface on BK7239N chips. It implements the RTC_GET_MS ioctl command in the RTC driver and provides a test function that can perform multiple readings with configurable delays.


amandeep-samsung (commented):
We will re-design these changes for adding support of msec api for beken.
 - as new ioctl in driver

PR #7196, opened 2026-03-19
os/arch/arm/src, os/include/tinyara/reboot_reason.h: Add specific reboot reasons and improve error tracking

abhinav-s235:
This commit improves the reboot reason system in TizenRT by adding specific reboot reasons at system failure points where they were previously missing. This enhancement provides better error tracking and debugging capabilities. 
New reboot reasons added (61-65):
- REBOOT_SYSTEM_INTERRUPT_FAILURE (61) - Interrupt handling failures
- REBOOT_SYSTEM_SCHEDULER_FAILURE (62) - Scheduler/reprioritization failures
- REBOOT_SYSTEM_TASK_START_FAILURE (63) - Task start failures
- REBOOT_SYSTEM_PTHREAD_START_FAILURE (64) - Pthread start failures
- REBOOT_SYSTEM_STACK_OVERFLOW (65) - Stack overflow detected

### DIFF
--- a/os/arch/arm/src/armv8-m/up_doirq.c
+++ b/os/arch/arm/src/armv8-m/up_doirq.c
@@ -63,6 +63,7 @@
 #include <tinyara/arch.h>
 #include <arch/board/board.h>
 #include <tinyara/security_level.h>
+#include <tinyara/reboot_reason.h>
 
 #include "up_arch.h"
 #include "up_internal.h"
@@ -102,6 +103,9 @@ uint32_t *up_doirq(int irq, uint32_t *regs)
 #endif
 	board_led_on(LED_INIRQ);
 #ifdef CONFIG_SUPPRESS_INTERRUPTS
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	up_reboot_reason_write(REBOOT_SYSTEM_INTERRUPT_FAILURE);
+#endif
 	PANIC();
 #else
 
--- a/os/arch/arm/src/armv8-m/up_reprioritizertr.c
+++ b/os/arch/arm/src/armv8-m/up_reprioritizertr.c
@@ -61,6 +61,7 @@
 #include <sched.h>
 #include <debug.h>
 #include <tinyara/arch.h>
+#include <tinyara/reboot_reason.h>
 
 #include "sched/sched.h"
 #include "up_internal.h"
@@ -121,6 +122,9 @@ void up_reprioritize_rtr(struct tcb_s *tcb, uint8_t priority)
 		|| priority > SCHED_PRIORITY_MAX
 #endif
 	   ) {
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+		up_reboot_reason_write(REBOOT_SYSTEM_SCHEDULER_FAILURE);
+#endif
 		PANIC();
 	} else {
 		struct tcb_s *rtcb = this_task();
--- a/os/arch/arm/src/common/up_allocateheap.c
+++ b/os/arch/arm/src/common/up_allocateheap.c
@@ -70,6 +70,7 @@
 #include <tinyara/mm/mm.h>
 #include <tinyara/kmalloc.h>
 #include <tinyara/mm/heap_regioninfo.h>
+#include <tinyara/reboot_reason.h>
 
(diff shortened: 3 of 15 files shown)


amandeep-samsung (approved):
addition of reboot reasons in code having PANIC to help in debugging for reason of system reboot.

os/include/tinyara/reboot_reason.h
```diff
@@ -42,6 +42,11 @@ typedef enum {
 	REBOOT_SYSTEM_BINARY_RECOVERYFAIL  = 58, /* Binary Recovery Fail */
 	REBOOT_SYSTEM_PERIODIC_REBOOT      = 59, /* Periodic reboot */
 	REBOOT_SYSTEM_ASSERT               = 60, /* Reboot from ASSERT or PANIC */
+	REBOOT_SYSTEM_INTERRUPT_FAILURE    = 61, /* Interrupt handling failure */
+	REBOOT_SYSTEM_SCHEDULER_FAILURE    = 62, /* Scheduler/reprioritization failure */
```
seokhun-eom24, line 46: Could you add `REBOOT_SYSTEM_SCHEDULER_FAILURE`, `REBOOT_SYSTEM_INTERRUPT_FAILURE` into armv7-m and armv7-r also.

abhinav-s235: Added

os/arch/arm/src/armv8-m/up_doirq.c
```diff
@@ -102,6 +103,9 @@ uint32_t *up_doirq(int irq, uint32_t *regs)
 #endif
 	board_led_on(LED_INIRQ);
 #ifdef CONFIG_SUPPRESS_INTERRUPTS
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+	up_reboot_reason_write(REBOOT_SYSTEM_INTERRUPT_FAILURE);
+#endif
```
eunwoo-nam, line 108: The CONFIG_SUPPRESS_INTERRUPTS is config for only debugging. it can be used when board bring.

So, we don't need to add this case.

os/arch/arm/src/armv8-m/up_reprioritizertr.c
```diff
@@ -121,6 +122,9 @@ void up_reprioritize_rtr(struct tcb_s *tcb, uint8_t priority)
 		|| priority > SCHED_PRIORITY_MAX
 #endif
 	   ) {
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+		up_reboot_reason_write(REBOOT_SYSTEM_SCHEDULER_FAILURE);
+#endif
```
eunwoo-nam, line 127: If we add this to this, what benefit can we get?

I mean, In what case does this occur? 
Is there a reason why we should detect this in a specific way?

If we want to detect an asset that occurs in relation to scheduling? kernel, we'd better add it to all the panics related to scheduling.

os/arch/arm/src/common/up_allocateheap.c
```diff
@@ -171,6 +175,9 @@ void up_add_kregion(void)
 		if ((void *)&_sdata <= (void *)kregionx_end && stack_end >= (void *)kregionx_end) {
 			lldbg("ERROR: Failed to allocate kheap for ram region configuration\n");
 			lldbg("Region start = 0x%x region end = 0x%x\n_sdata = 0x%x end of stack = 0x%x\n", kregionx_start[region_cnt], kregionx_end, &_sdata, stack_end);
+#ifdef CONFIG_SYSTEM_REBOOT_REASON
+			up_reboot_reason_write(REBOOT_SYSTEM_MEMORYALLOCFAIL);
```
eunwoo-nam, line 179: Is that appropriate?
i think, it's booting error, not alloc fail

os/arch/arm/src/armv8-m/up_reprioritizertr.c
amandeep-samsung: right, this need checks all possibilities and will check and update.

os/arch/arm/src/common/up_allocateheap.c
amandeep-samsung: Mr. Nam, should we define another reboot reason for it. Like REBOOT_SYSTEM_INIT

PR #7226, opened 2026-03-31
os/, build/configs/: Segregate board specific changes from mkldscript.py

namanjain7:
Refactor mkldscript.py file. Add common code in mkldscript.py and board specific address calculation to build/configs/<board>/scripts


namanjain7:
Removed support for common_*.ld and app1_*.ld
Passing values as arguments to Makefile.unix and loadable.mk using --defsym

amandeep-samsung:
@ewoodev : please review design so we can make changes accordingly in discussed design.
 - mkldscript does not have board specific code now.
 - related files are used for board specific address calculation 
 - common_0.ld and app1_0.ld are not created now and variables are used to pass flash addresses in userspace_all.ld for XIP

loadable_apps/loadable.mk
```diff
@@ -59,11 +59,20 @@ endif
 endif
 
 ifeq ($(CONFIG_XIP_ELF),y)
+MEM_VARS := $(shell python $(TOPDIR)/tools/mkldscript.py $(CONFIG_APP1_BIN_NAME))
+$(foreach v,$(MEM_VARS),$(eval $(v)))
+APP1_LD_DEFS := \
```
eunwoo-nam: This file is common including from app1 app2.

please check app2 case

os/tools/mkldscript.py
```diff
@@ -135,143 +64,36 @@
...
-    PART_IDX = PART_IDX + 1
+if binary_type == "common":
+    print("RAM_ADD={}".format(common_ram_start))
+    print("RAM_SIZE={}".format(hex(common_ram_size)))
+elif binary_type == "app1":
+    print("RAM_ADD={}".format(app1_ram_start))
+    print("RAM_SIZE={}".format(app1_ram_size))
+elif binary_type == "app2":
+    print("RAM_ADD={}".format(app2_ram_start))
+    print("RAM_SIZE={}".format(app2_ram_size))
```
eunwoo-nam: why everytime calculate and return only one?

it's better calculating once of argument at that time.

build/configs/rtl8730e/scripts/xipelf/get_binary_address.py
```diff
@@ -0,0 +1,90 @@
...
+            else:
+                binary_start = hex(current_offset + 0x30 + signing_offset)
+                binary_size = hex(part_size - 0x30 - signing_offset)
+            
+            print("FLASH_ADD={}".format(binary_start))
+            print("FLASH_SIZE={}".format(binary_size))
+            return
+
+        current_offset = current_offset + part_size
+        index += 1
```
eunwoo-nam: This function is board specific?
i think, platform just needs only virture start address of kernel partition

loadable_apps/loadable.mk
amandeep-samsung: yes, we are adding both, 
1. app2 and also 
2. OTA index (just like before it was for both cases). We will take OTA index as argument and will fix it by defconfig as slot 0.

sunghan-chang:
@namanjain7 This is pending for long time. Please let me know status.

namanjain7:
> @namanjain7 This is pending for long time. Please let me know status.

Hello Mr. chang,
we have done the changes that need to be done.
there are some recent makefile changes, and we are merging it with our commit.
We will update ASAP when it's done.

namanjain7:
we have added dynamic ram start calculation for each partition.
the ram allocation starts from the end.
the script will be called everytime we build for common, app1, and app2.
currently, we are facing issues related to removing duplicate calculation. calculating app1 offset will require calculating common binary offset also which creates duplicate calculation. we are optimizing this.
this task is on priority and we will push the final version ASAP when done.

os/tools/get_addr.py
```diff
@@ -0,0 +1,18 @@
...
+    print("RAM_ADD={}".format(hex(ram_start)))
+    print("RAM_SIZE={}".format(hex(ram_size)))
+
+    print(sys.argv[2])
+    print(sys.argv[3])
+    print(sys.argv[4])
+    print(sys.argv[5])
+
+if __name__ == "__main__":
+    get_mem_config()
```
amandeep-samsung: please remove it

namanjain7:
we will update flash addressing related changes after more verify

os/Makefile.unix
```diff
@@ -449,6 +449,15 @@ include LibTargets.mk
...
 # is unknown to this makefile unless CONFIG_PASS1_OBJECT is defined.
 
+COMMON_BINARY_NAME=common
+MEM_VARS := $(shell python $(TOPDIR)/tools/mkldscript.py $(COMMON_BINARY_NAME))
+$(foreach v,$(MEM_VARS),$(eval $(v)))
+COMMON_LD_DEFS := \
+	--defsym __FLASH_START_ADDRESS__=$(FLASH_ADD) \
+	--defsym __FLASH_SIZE__=$(FLASH_SIZE) \
+	--defsym __RAM_START_ADDRESS__=$(RAM_ADD) \
+	--defsym __RAM_SIZE__=$(RAM_SIZE)
```
ewoodev, line 461: this code is added severel code(common app1 app2)
how about make function??

os/tools/bin_mem_layout.py
```diff
@@ -0,0 +1,291 @@
...
+import json
+import config_util as util
+import importlib.util
+import argparse
+
+os_folder = os.path.dirname(os.path.abspath(__file__)) + '/..'
+cfg_file = os.path.join(os_folder, '.config')
+build_folder = os.path.join(os_folder, '..', 'build')
+output_folder = os.path.join(build_folder, 'output', 'bin')
+saved_config_file_path = os.path.join(output_folder, 'mem_layout_cache.json')
```
ewoodev: i'd like to know why this is needed.

which file is using this file?

namanjain7: cfg_file ->used to get CONFIG_RAM_START, common app1 app2... binary size to get address.
build_folder, output_folder, saved_config_file_path -> For saving json file in output folder. This file will have config values extracted from .config file and memory values.

os/Makefile.unix
namanjain7: The common binary is build in Makefile.unix. So, this code will execute once.
For app1, app2, --> The makefile is loadable.mk
Both have different makefiles.

os/tools/mkldscript.py
namanjain7: Code modified.
Now, when script is called for common binary, then addresses for app1, app2 will also be calculated and saved in json file. When call for app1, app2 comes, it will fetch from json cache file and return it directly without calculating.

namanjain7:
All issues resolved. We have tested in tp1x, tp1x plus, and ailite boards.

os/tools/bin_mem_layout.py
```diff
@@ -0,0 +1,213 @@
...
+            with open(saved_config_file_path, 'r') as f:
+                return json.load(f)
+        except (json.JSONDecodeError, IOError):
+            pass
+    return {"configs": {}, "memory_layout": {}}
+
+def save_parameters(data):
+    ensure_output_dir()
+    with open(saved_config_file_path, 'w') as f:
+        json.dump(data, f, indent=2)
```
eunwoo-nam: Could you please let me know why file save is needed?

os/userspace/userspace_apps_xip.ld
```diff
@@ -16,6 +16,12 @@
  *
  ****************************************************************************/
 
+MEMORY
```
eunwoo-nam, line 19: build/configs/rtl8730e/scripts/xipelf/userspace_all.ld
build/configs/bk7239n/scripts/xipelf/userspace_all.ld

Do you have any reason for sperate this ld script to board specific?

os/tools/bin_mem_layout.py
namanjain7: we calculate the address of common, app1, app2, ...
The script is called multiple times for every binary (common, app1, app2).

The memory layout in RAM is: app2, app1, common, RAM_END
build order is: common, app1, app2, ....

When build for common is called, then we calculated the address for ram and flash for all binaries and saved in separate file in output directory, and then passed as arguments.

For app1, app2, we just extract values from file and pass it as arguments. We don't need to recalculate the address.

Why we didn't calculated the address for app1, app2 at the time when script is called for app1, app2?
The reason is that we need addresses of common. To avoid this, we calculated all addresses for common, app1, app2 at once and save it to file. Then we parse as argument for each binary.

os/userspace/userspace_apps_xip.ld
namanjain7: This was already separated by beken.
both the files are same. If required, we can merge them.

eunwoo-nam: please merge these

os/tools/bin_mem_layout.py
eunwoo-nam: It seems appropriate to save this Python script as it is called several times.
but I think it should be improved to call this Python script repeatedly.

We can be able to get all memory layouts from the initial call only once. 
we can make to change to the role of distributing that layout in makefile.unix.
then This Python file will get a Single Responsibility Principle that gets a simple memory layout.

It looks good to store the calculated memory layout as a product.
However, re-referencing the memory layout file back to the build does not look good.

Please let me know your opinion

os/userspace/userspace_apps_xip.ld
namanjain7: ohkk.

os/tools/bin_mem_layout.py
amandeep-samsung: @ewoodev ; yes, agreed and python script be called once. in new change, we do as
 - refactored `bin_mem_layout.py` to support a `--all` flag, which makes all 12 addresses (`COMMON_FLASH_ADD`, `APP1_FLASH_ADD`, `APP2_FLASH_ADD` etc) for all binaries in a single time from `Makefile.unix`, and remove JSON cache and file access entirely. 
  - `loadable.mk` read exported Make variables directly using passed variables from Makefile.unix.
 
there will no need for multiple calling and any intermediate temporary file.

and regarding OTA index, 
This is done for OTA index 0 which is configurable in defconfig and changed from previous hard code value.

os/userspace/userspace_apps_xip.ld
amandeep-samsung: common file created for both: build/configs/common/scripts/xipelf/userspace_all.ld

eunwoo-nam: Let's move os/userspace

os/tools/bin_mem_layout.py
```diff
@@ -0,0 +1,178 @@
...
+    args = parser.parse_args()
+
+    configs = load_configs_from_file()
+    memory_layout = calculate_memory_layout(configs)
+
+    if args.all:
+        for bin_name in ('common', 'app1', 'app2'):
+            entry = memory_layout.get(bin_name, {}).get(str(args.ota_index), {})
+            prefix = bin_name.upper()
+            print(prefix + "_FLASH_ADD="+entry.get('flash_start', '0x0'))
```
eunwoo-nam: minor, APP1_FLASH_ADDR is better

os/userspace/userspace_apps_xip.ld
eunwoo-nam: Let's rename userspace_apps_xip.ld

ewoodev (dismissed):

os/Makefile.unix
```diff
@@ -449,6 +449,24 @@ include LibTargets.mk
 # is an archive.  Exactly what is performed during pass1 or what it generates
 # is unknown to this makefile unless CONFIG_PASS1_OBJECT is defined.
 
+ifneq ($(CONFIG_ELF),y)
+# Compute all binary memory layouts in a single Python invocation
+ALL_MEM_VARS := $(shell python3 $(TOPDIR)/tools/bin_mem_layout.py --all --ota-index $(CONFIG_DEFAULT_OTA_INDEX))
```
eunwoo-nam: If we use board whch is needed OTA A/B binary, how can we make binary B?
do we need to configure CONFIG_DEFULT_OTA_INDEX and build again?

how about create config for make entire dual binary?

namanjain7: Hello @ewoodev .
yes if we want to make binary B, we need to configure default OTA index and build again.
In existing code, the binary is created only for OTA A.
OTA B is not created. But ld is generated for both OTA (common_0.ld, common_1.ld)

This PR only refactor the mkldscript.py file . currently there is no support for entire dual binary in all current defconfigs.
If required, we can add support for dual binary in separate PR.

eunwoo-nam: thanks for reply, 
i get your point. it's minor review. but, current if CONFIG_DEFULT_OTA_INDEX is not settet, the build is failed, it's default value 0, but this config is mandatory 

I just suggest, support only setting index here with config. 
when we support this secondery ota binary, let's make repeat build script for second ota linking
```
ifneq ($(CONFIG_OTA_SUPPORT_SECOND_PARTITION),y)
ACTIVE_OTA_INDEX:=0
else
ACTIVE_OTA_INDEX:=1
endif
ALL_MEM_VARS := $(shell python3 $(TOPDIR)/tools/bin_mem_layout.py --all --ota-index $(ACTIVE_OTA_INDEX))
```

os/Makefile.unix
```diff
@@ -449,6 +449,29 @@ include LibTargets.mk
 # is an archive.  Exactly what is performed during pass1 or what it generates
 # is unknown to this makefile unless CONFIG_PASS1_OBJECT is defined.
 
+ifneq ($(CONFIG_ELF),y)
+ifneq ($(CONFIG_OTA_SUPPORT_SECOND_PARTITION),y)
```
namanjain7, line 453: I have added the config for secondary OTA support.
By default it's disabled.
We can enable to switch build from OTA0 -> OTA1

namanjain7:
All requested comments fixed. Kindly review and merge.

namanjain7:
Please review and merge this PR.
All requested changes are done.

PR #7232, opened 2026-04-01
os/arch/arm: Improve CPU off/on functions with validation checks

seokhun-eom24:
Updated `up_cpu_off` and `up_cpu_on` functions to include checks for valid CPU indices and prevent operations on the current CPU. 
Added error logging for invalid CPU index and self-referential operations to improve debugging and stability. 
This changes is to prevent assertion when passing invalid parameters to `up_cpu_off` and `up_cpu_on` functions.

### DIFF
--- a/os/arch/arm/src/armv7-a/arm_cpuon.c
+++ b/os/arch/arm/src/armv7-a/arm_cpuon.c
@@ -44,6 +44,7 @@
 #include <stdint.h>
 #include <assert.h>
 #include <debug.h>
+#include <errno.h>
 
 #include <tinyara/arch.h>
 #include <tinyara/sched.h>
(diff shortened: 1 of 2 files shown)


os/arch/arm/src/armv7-a/arm_cpuon.c
```diff
@@ -177,11 +178,19 @@ int arm_start_handler(int irq, void *context, void *arg)
 
 int up_cpu_on(int cpu)
 {
-	int ret ;
+	int ret;
 
 	smpllvdbg("Starting CPU%d\n", cpu);
 
-	DEBUGASSERT(cpu >= 0 && cpu < CONFIG_SMP_NCPUS && cpu != this_cpu());
+	if (cpu < 0 || cpu >= CONFIG_SMP_NCPUS) {
```
amandeep-samsung: To avoid ON/OFF operation on CPU 0, can we change this condition as below:
if (cpu <= 0 || cpu >= CONFIG_SMP_NCPUS) 

means also include CPU 0 in condition for INVALID parameter

seokhun-eom24: Thank you for point out.
I updated it.

sunghan-chang (approved):

PR #7247, opened 2026-04-07
board/rtl8730e: Rename and fix return type of the IWDG status function

seokhun-eom24:
The `up_check_iwdg` function was called during board initialization only prints whether the IWDG is enabled,
Ant function return type was int but it doesn't return any value.
So rename this function to `up_print_iwdg_status` and fix the return type to void.

### DIFF
--- a/os/board/rtl8730e/src/rtl8730e_boot.c
+++ b/os/board/rtl8730e/src/rtl8730e_boot.c
@@ -145,7 +145,7 @@ static void board_initialize_bor(void)
 }
 #endif
 
-int up_check_iwdg(void)
+void up_print_iwdg_status(void)
 {
 	u32 Temp = HAL_READ32(SYSTEM_CTRL_BASE, REG_AON_FEN);
 
@@ -507,7 +507,7 @@ void board_initialize(void)
 	ipc_msg_loguart.rsvd = 0; /* for coverity init issue */
 	ipc_send_message(IPC_AP_TO_LP, IPC_A2L_DISLOGUART, &ipc_msg_loguart);
 	
-	up_check_iwdg();
+	up_print_iwdg_status();
 }
 #else
 #error "CONFIG_BOARD_INITIALIZE MUST ENABLE"


seokhun-eom24:
# PR #7247 — os/board/rtl8730e/src/rtl8730e_boot.c: Update IWDG check function

**This review done by codex. AI reviews can be inaccurate.**

---

**Repository**: `Samsung/TizenRT`

**Base → Head**: `master` → `pr/7247`

**HEAD Commit**: `390f29bdd2aa0a4ae51e5cbd9b01788f6c287389`

**Scope**: `rtl8730e` board boot path, specifically the IWDG diagnostic helper rename and return-type fix in `rtl8730e_boot.c`

---

## 🔎 Review Summary

| Category               | Status                     |
| ---------------------- | -------------------------- |
| Build / Compile Status | ⚠️ Not Run                 |
| Functional Correctness | ✅ No blocking issue found |
| Validation Coverage    | ⚠️ Static review only      |

**Final Verdict: ✅ Approve**

---

## 🚨 Must-Fix Issues

None.

---

## 🟡 Nice-to-Have Improvements

None.

---

## ✅ Notable Improvements

### ✔ The helper name now matches its real behavior

* `os/board/rtl8730e/src/rtl8730e_boot.c:148` now exposes the helper as `up_print_iwdg_status()`, which matches the fact that the routine only logs the IWDG state.
* This removes the misleading `check_*` naming that previously suggested a status-style return contract.

### ✔ The patch removes undefined non-void fallthrough without adding boot-path risk

* The previous `up_check_iwdg()` implementation fell off the end of a non-void function after printing the watchdog state.
* The updated `void` helper preserves the exact runtime behavior while fixing the type mismatch cleanly.

### ✔ The only in-tree caller was updated in the same patch

* `os/board/rtl8730e/src/rtl8730e_boot.c:510` now calls `up_print_iwdg_status()`, so there is no stale in-tree call site expecting the old symbol or a return value.
* Keeping the API rename and call-site update in one file keeps the early boot path easy to reason about.

---

## 🧾 Final Assessment

### Must-Fix Summary

* None.

### Nice-to-Have Summary

* None.

### Residual Risk / Test Gap

* Static review only; I did not run a board build or runtime boot validation.
* The change is narrowly scoped to a local helper and its single in-tree caller, so the remaining risk is mainly lack of build and boot confirmation rather than logic ambiguity.

---

## 📌 Final Verdict

> ✅ Approve

The updated patch resolves the earlier review concern by renaming the helper to a print-only API and changing its return type to `void`. I did not find any remaining in-tree correctness issue or behavioral regression in the touched `rtl8730e` boot path.

os/board/rtl8730e/src/rtl8730e_boot.c
eunwoo-nam: static void up_check_iwdg(void) is better

seokhun-eom24: Thank you for suggest.
I updated.

os/board/rtl8730e/src/rtl8730e_boot.c
```diff
@@ -145,7 +145,7 @@ static void board_initialize_bor(void)
 }
 #endif
 
-int up_check_iwdg(void)
+void up_check_iwdg(void)
```
sunghan-chang: Do you mean that our usage is just printing iwdg feature on/off?
Is that enough?
Then, let's change the name as well. Now, the name is `up_check_iwdg`.

sunghan-chang (commented):
Please don't use `update` at a title. 
`update` does not give any meaningful message.

os/board/rtl8730e/src/rtl8730e_boot.c
seokhun-eom24: I updated the function name and commit message meaningfully.

sunghan-chang (approved):

PR #7248, opened 2026-04-07
os/pm/pm_procfs.c: Simplify power_write function

seokhun-eom24:
Removed unnecessary variable and logic from the power_write function, which now directly returns 0.

### DIFF
--- a/os/pm/pm_procfs.c
+++ b/os/pm/pm_procfs.c
@@ -539,7 +539,6 @@ static ssize_t power_read(FAR struct file *filep, FAR char *buffer, size_t bufle
 static ssize_t power_write(FAR struct file *filep, FAR const char *buffer, size_t buflen)
 {
 	FAR struct power_file_s *priv;
-	ssize_t ret;
 
 	fvdbg("buffer=%p buflen=%d\n", buffer, (int)buflen);
 
@@ -548,14 +547,7 @@ static ssize_t power_write(FAR struct file *filep, FAR const char *buffer, size_
 	DEBUGASSERT(priv);
 
 	/* pm_procfs does not have write operation, one must code write logic here */
-	ret = 0;
-
-	/* Update the file offset */
-	if (ret > 0) {
-		filep->f_pos += ret;
-	}
-
-	return ret;
+	return 0;
 }
 
 /****************************************************************************


seokhun-eom24:
# PR #7248 — os/pm/pm_procfs.c: Simplify power_write function

**This review done by codex. AI reviews can be inaccurate.**

---

**Repository**: `Samsung/TizenRT`

**Base → Head**: `master` → `pr/7248`

**HEAD Commit**: `b376804e4a4224b11da3c42245922862f83c55a9`

**Scope**: Single-file cleanup in `os/pm/pm_procfs.c` that removes dead local state from `power_write()` and makes the existing zero-byte return explicit.

---

## 🔎 Review Summary

| Category | Status |
| -------- | ------ |
| Build / Compile Status | ⚠️ Not Run |
| Functional Correctness | ✅ No blocking issue found in the diff |
| Validation Coverage | ⚠️ Static review only |

**Final Verdict: 💬 Comment**

---

## 🚨 Must-Fix Issues

None.

---

## 🟡 Nice-to-Have Improvements

### 1. Align pm procfs write behavior with its read-only file mode

| Item | Details |
| ---- | ------- |
| **Location** | `os/pm/pm_procfs.c:424` |
| **Severity** | Medium |
| **Type** | Interface consistency |

**Problem**
`power_stat()` exposes `/proc/power/*` nodes as read-only files, but `power_open()` does not reject write access and `power_write()` still returns success with `0` bytes written. This patch keeps that behavior unchanged while touching the write path.

**Impact**

* User space can observe a silent no-op write against a node that is advertised as read-only.
* The behavior diverges from sibling procfs implementations such as `smartfs_procfs`, which reject unsupported write opens with `-EACCES`.
* Failures become harder to diagnose because the caller gets "success with no progress" instead of an explicit permission or capability error.

**Recommended Action**

* Consider a follow-up change that rejects write opens in `power_open()`, matching the read-only contract published by `power_stat()`.
* If writable procfs power nodes are intended in the future, document that contract explicitly and return a concrete error from `power_write()` until the operation is implemented.

**Expected Output**

* `open(..., O_WRONLY)` or `write()` against `/proc/power/*` should fail with an explicit error such as `-EACCES` or `-ENOSYS`.
* Behavior should be consistent with the file mode reported by `stat()` and with neighboring procfs implementations.

---

## ✅ Notable Improvements

### ✔ Removes dead state updates from a no-op write handler

* The previous implementation initialized `ret` to `0`, conditionally updated `f_pos` only when `ret > 0`, and then returned `ret`.
* Replacing that sequence with `return 0;` makes it clear that the current implementation does not consume input and does not alter file position.

---

## 🧾 Final Assessment

### Must-Fix Summary

* No blocking issues found in this cleanup patch.

### Nice-to-Have Summary

* The touched path still carries a pre-existing semantic mismatch between read-only procfs metadata and its write-side behavior.

### Residual Risk

* Static review only; no runtime or target build validation was run.
* The review did not exercise user-space write attempts against `/proc/power/*`, so the interface note above remains unvalidated in this environment.

---

## 📌 Final Verdict

> 💬 Comment

The patch itself is behavior-preserving and the simplification looks correct. I would not block merge on this change, but I would leave one follow-up comment about aligning the pm procfs write path with the read-only contract it already reports to callers.

eunwoo-nam (approved):

PR #7255, opened 2026-04-09
[Do Not Merge]Replace DEBUGASSERT with runtime null pointer checks

jijo-jacob:
To improve embedded system reliability by ensuring graceful error handling. Replace DEBUGASSERT() macro calls with proper runtime null pointer validation.

### DIFF
--- a/os/arch/arm/src/amebad/amebad_i2c.c
+++ b/os/arch/arm/src/amebad/amebad_i2c.c
@@ -852,7 +852,9 @@ static int amebad_i2c_isr(int irq, void *context, FAR void *arg)
 static int amebad_i2c_init(FAR struct amebad_i2c_priv_s *priv)
 {
 	/* Power-up and configure GPIOs */
-	DEBUGASSERT(priv);
+	if (!priv) {
+		return -ENODEV;
+	}
 	DEBUGASSERT(!priv->i2c_object);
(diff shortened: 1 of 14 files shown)


os/arch/arm/src/amebad/amebad_i2c.c
```diff
@@ -852,7 +852,9 @@ static int amebad_i2c_isr(int irq, void *context, FAR void *arg)
 static int amebad_i2c_init(FAR struct amebad_i2c_priv_s *priv)
 {
 	/* Power-up and configure GPIOs */
-	DEBUGASSERT(priv);
+	if (!priv) {
```
ewoodev, line 855: Could you please let me know why we need to change return error from DEBUGASSERT?

is priv allowed to be null?
is the BSP code handling error during booting?

ewoodev:
I'd like to provide infomation regarding this change.
The reason we made the change previously is that the kernel crashes when the argument received from the user is null.
(from syscall)

Resources like "priv" cannot be null. If it becomes null, the data, bss, or heap is corrupted.

os/arch/arm/src/amebad/amebad_i2c.c
jijo-jacob: We will change only for syscall, ioctl etc. Not during the booting.
And we also add checks to handle return error in case of syscall.

PR #7259, opened 2026-04-10
os/board/bk7239n: Report SoftAP station disconnection reason to adapter layer

jenny668-yesok:
This commit implements the reporting of disconnection reasons when a station is disconnected in SoftAP mode.

Key changes:
Enhanced the event mechanism to trigger a notification when a station is deauthenticated or dissociated from the SoftAP. 
Passes the extracted MAC and Reason Code directly to the adapter layer.


hk-gwak (approved):
Checked disconnect reason is printed like below.
> fa:4d:29:b9:59:10 disconnected from BK AP, reason(77)

PR #7265, opened 2026-04-13
os/board/bk7239n:MTU size can be set differently for each ble link

lingzhou2018:
MTU size can be set differently for each ble link


giwon-nam (approved):
Confirmed that the two issues have been resolved.
1. The issue where the MTU size is always set to 65 even when configured as 512 in the client connection.
2. The issue where the MTU size used in the client connection is continuously used in subsequent connections.

PR #7269, opened 2026-04-15
os/board/bk7239n: Fix a deadlock between BLE attr_get_data and attr_set_data

lingzhou2018:
Optimize ATT read/write buffer strategy to fix deadlocks between attr_get_data and attr_set_data


giwon-nam (approved):
I confirmed that the `ble_server_attr_get_data` function inside the attribute callback is returning normally.

PR #7271, opened 2026-04-15
os/board/bk7239n, os/arch/arm/src: adjust adapter to support single wlan interface for non-concurrent mode

jenny668-yesok:
When BK_CONCURRENT_MODE is disabled, STA and SoftAP share different VIF index but both operate on a single wlan0 interface.

Key changes:
- Guard wlan1 device registration with BK_CONCURRENT_MODE macro, so only wlan0 is registered in non-concurrent mode.
- Switch VIF index and netdev routing to wlan0-only path.
- Route SoftAP events to wlan0 and update VIF index on STA/SoftAP transitions.


hk-gwak (approved):
Checked both softAP & STA uses wlan0.

